PR #85, opened 2020-04-01
Fix Metadata::IsSet, operator==, and operator!=

### LINKED ISSUE #84: Metadata::IsSet bug
The Metadata::IsSet member function checks the value of bits beyond the size of the bits_ vector, producing undefined behavior.

This version should work as expected, but it needs to be put in and a test added.

```c++
  bool IsSet(const MetadataFlag bit) const { 
    if (bit.flag_ >= bits_.size()) {
      return false;
    }
    return bits_[bit.flag_];
  }
```

AndrewGaspar:
## PR Summary

Fixes #84 as identified by @jdolence, and then also fixes a related issues in the comparison operator. I added some new tests, too.

Also fixes up cpplint issues that were pre-existing.

## PR Checklist

- [x] Code passes cpplint
- [x] New features are documented. (no new features)

### DIFF
--- a/src/interface/Metadata.hpp
+++ b/src/interface/Metadata.hpp
@@ -21,14 +21,15 @@
 #include <tuple>
 #include <vector>
 
-/// The point of this macro is to generate code for each built-in flag using the 
+/// The point of this macro is to generate code for each built-in flag using the
 /// `PARTHENON_INTERNAL_FOR_FLAG` macro. This is to accomplish the following goals:
 /// - Generate a unique value for each flag using an enum
-/// - Wrap each value in a `MetadataFlag` type that can only be instantiated within parthenon
+/// - Wrap each value in a `MetadataFlag` type that can only be instantiated within
+///   parthenon
 /// - Make it possible to return a user-friendly string for each flag.
 ///
-/// Having this macro means you only need to add a new flag in one place and have each of these
-/// properties automatically be updated.
+/// Having this macro means you only need to add a new flag in one place and have each of
+/// these properties automatically be updated.
 #define PARTHENON_INTERNAL_FOREACH_BUILTIN_FLAG \
   /**  bit 0 is ignored */ \
   PARTHENON_INTERNAL_FOR_FLAG(Ignore) \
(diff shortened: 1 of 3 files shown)


src/interface/Metadata.hpp
```diff
@@ -237,21 +239,42 @@ class Metadata {
...
+    return bit.flag_ < bits_.size() && bits_[bit.flag_];
+  }
 
   // Operators
   bool operator==(const Metadata &b) const {
-    return PackedTuple() == b.PackedTuple();
+    auto const &a = *this;
+
+    // Check extra bits are unset
+    auto const max_bits = std::max(a.bits_.size(), b.bits_.size());
```
AndrewGaspar, line 251: Wait, this is wrong. D'oh.

tst/unit/test_metadata.cpp
```diff
@@ -30,14 +30,56 @@ TEST_CASE("Built-in flags are registered", "[Metadata]") {
...
+
+#define PARTHENON_INTERNAL_FOR_FLAG(name) \
+    if (Metadata::name != Metadata::FillGhost) REQUIRE(!m.IsSet(Metadata::name));
+    PARTHENON_INTERNAL_FOREACH_BUILTIN_FLAG
+#undef PARTHENON_INTERNAL_FOR_FLAG
+
+        REQUIRE(m.IsSet(Metadata::FillGhost));
+    }
+
+    GIVEN("Two Equivalent Metadata Structs Are Compared") {
```
jlippuner, line 71: Maybe also compare Metadata with the same flags set in different orders

tst/unit/test_metadata.cpp
```diff
@@ -30,14 +30,56 @@ TEST_CASE("Built-in flags are registered", "[Metadata]") {
...
+        Metadata a({Metadata::Cell}), b({Metadata::Face});
+
+        b.Unset(Metadata::Face);
+        b.Set(Metadata::Cell);
+
```
jlippuner, line 76: Maybe also set and unset a flat with a larger int behind it and make sure comparison still works as expected

tst/unit/test_metadata.cpp
```diff
@@ -30,14 +30,56 @@ TEST_CASE("Built-in flags are registered", "[Metadata]") {
...
+        REQUIRE(a == b);
+    }
+
+    GIVEN("Two Different Metadata Structs Are Compared") {
```
jlippuner, line 80: Maybe compare two Metadata with different numbers of flags and/or multiple flags

Yurlungur, line 80: Does this test actually catch the segfault @jdolence was seeing? i.e., #84 ? Can we compare two metadata structs where we know the flag sizes are different?

And if this test already does stress this code path, can we say that in a comment?

AndrewGaspar: This test does not, but it does test the fix I did to `Metadata::operator==`.

tst/unit/test_metadata.cpp
AndrewGaspar: This test does that, but I can do it for multiple flags.

tst/unit/test_metadata.cpp
AndrewGaspar: This test does that.

PR #105, opened 2020-04-15
[WIP] GPU Performant Container Iterations

### LINKED ISSUE #90: GPU friendly ContainerIterator
The current iterator does not return something that is usable on the GPU.  It's not entirely clear what should be returned.  A view of views?  Some other Kokkos container?  Anyone have thoughts?

@Yurlungur @forrestglines @dholladay00 perhaps this is obvious to you?

forrestglines:
Modify containers to be performant on GPU

## PR Summary

The container implementation and potentially interface need to be changed in order to be as performant as the raw arrays originally used in Athena++. This may involve providing an appropriate variable looping interface or changing the internal layout of memory in `Containers` or `ContainerCollections`.

Optimistically fixes #90 

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted

### DIFF
--- a/tst/unit/test_container_iterator.cpp
+++ b/tst/unit/test_container_iterator.cpp
@@ -39,6 +39,7 @@ using parthenon::loop_pattern_mdrange_tag;
 using parthenon::Metadata;
 using parthenon::par_for;
 using parthenon::ParArrayND;
+using parthenon::ParArray4D;
 using parthenon::Real;
 
 TEST_CASE("Can pull variables from containers based on Metadata", "[ContainerIterator]") {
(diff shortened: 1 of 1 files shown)


tst/unit/test_container_iterator.cpp
```diff
@@ -137,3 +138,156 @@ TEST_CASE("Can pull variables from containers based on Metadata", "[ContainerIte
...
+        });
+    });
+
+
+  //Make a container for testing performance
+  Container<Real> container;
+  Metadata m_in({Metadata::Independent});
+  Metadata m_out;
+  std::vector<int> scalar_block_size {N,N,N};
+  std::vector<int> vector_block_size {3,N,N,N};
```
jdolence, line 215: This is backwards, I think.  Should be
```suggestion
  std::vector<int> vector_block_size {N,N,N,3};
```

tst/unit/test_container_iterator.cpp
```diff
@@ -137,3 +138,156 @@ TEST_CASE("Can pull variables from containers based on Metadata", "[ContainerIte
...
+
+  // make some variables - 5 in all, 2 3-vectors, total 10 fields
+  container.Add("v0",m_in, scalar_block_size);
+  container.Add("v1",m_in, scalar_block_size);
+  container.Add("v2",m_in, vector_block_size);
+  container.Add("v3",m_in, scalar_block_size);
+  container.Add("v4",m_in, vector_block_size);
+
```
jdolence, line 223: You've only added 9 fields.
```suggestion
container.Add("v5",m_in, scalar_block_size);
```

tst/unit/test_container_iterator.cpp
forrestglines: I noticed this in the original test. Are we still using `LayoutRight` by default, and are we using these vectors like `{X,k,j,i}` and vectorizing across the last dimension? Then `{3,N,N,N}` is the best layout, so that consecutive elements in the vector are strided in memory. It might even make sense to use `{N,N,3,N}` so that the last dimension still vectorizes but consecutive elements are in the same page.

forrestglines: The `3` should not be used in the fastest moving direction, which needs to vectorize.

jdolence: This line doesn't control the layout.  There's still some inconsistency (which we should probably eliminate) with respect to how we declare the size of different dimensions.  When adding a variable to a container, the vector of sizes passed in is interpreted like
```c++
vector_block_size {dim1, dim2, dim3, dim4, ...}
```
This translates into a ParArrayND like
```c++
ParArrayND<Real> data("name",dim4,dim3,dim2,dim1);
```
and the layout is ultimately controlled by Kokkos, I think.  So, again, with the current code, that line needs to change.  Leaving as is produces memory errors.

forrestglines: I see, that makes much more sense. It should be `{N,N,N,3}` then if the dimensions get flipped

forrestglines:
Superseded by #125

Issue #49, opened 2020-03-13
Better associate requirements in input files with applications/drivers


jdolence:
In writing the pi example, I discovered that the input file requires things like boundary conditions and tlim to be specified, which are of course irrelevant in this application.  It's good to require things be specified, but they should be tied to the actual application being developed.

PR #108, opened 2020-04-16
Better capture application requirements through driver

jdolence:
## PR Summary

This PR is meant to allow applications to better articulate what is needed from the input deck.  This required refactoring some things, for example disassociating things like "time" from Mesh.  Introduces a mechanism for applications to specify required and desired parameters, with appropriate reporting at runtime.

This needs to wait until after #103 is merged.

Addresses #49.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


src/basic_types.hpp
```diff
@@ -33,6 +35,17 @@ using Real = double;
...
+  SimTime() = default;
+  SimTime(const Real tstart, const Real tstop, const int nmax, const int ncurr,
+          const int nout)
+      : start_time(tstart), time(tstart), tlim(tstop),
+        dt(std::numeric_limits<Real>::max()), nlim(nmax), ncycle(ncurr),
+        ncycle_out(nout) {}
+  Real start_time, time, tlim, dt;
+  int nlim, ncycle, ncycle_out;
+};
+
```
Yurlungur, line 48: I like this.

example/calculate_pi/pi.hpp
```diff
@@ -49,7 +51,14 @@ using parthenon::Globals::nranks;
...
  public:
-  CalculatePi(ParameterInput *pin, Mesh *pm, Outputs *pout) : Driver(pin, pm, pout) {}
+  CalculatePi(ParameterInput *pin, Mesh *pm) : Driver(pin, pm) {
+    InitializeOutputs();
+    AppInputs_t required;
+    AppInputs_t desired;
+    desired["Pi"].push_back("radius");
+    desired["graphics"].push_back("variables");
+    pin->CheckRequiredDesired(required, desired);
+  }
```
Yurlungur, line 61: I find it a little strange that this lives in the driver, as opposed to when the package is set in `process_packages`. I dislike that this is one extra thing that needs to be overriden per application, especially when it feels like a per-package thing instead. I think this could be done at the same time as the input file is actually read. Is there a reason not to do it that way?

jdolence:
OK, high level comment on the point of this PR.  Basically, the thought was that an application developer/user shouldn't have to specify things that aren't relevant for the particular application.  For example, setting tlim in a simulation where there is no notion of time.  Having to do so is just confusing.  With this PR, there are three ways requirements are enforced (and now a way to articulate parameters that are not required to be set, but the app developer suggests setting).

1. The Parthenon infrastructure has it's own demands.  Things like nx1, for example, must be specified.  This requirement was already captured by means of error checking in the Get* methods of the ParameterInput class.  That's sufficient, so I left that alone.

2. The Parthenon infrastructure demanded more than is always required.  I dealt with that in two ways.  Where it made sense (at least for now), I changed the required to optional arguments by moving Get* calls to GetOrAdd* calls, specifying a default.  In the case of time/cycle info, I pulled it out of the Mesh class entirely (which is relevant for all simulations) and stuck it into the EvolutionDriver (where time/cycle stuff has meaning).  Within that narrowed context, I still use the Get* methods to check for the required inputs.

3. Finally, an app developer may wish to require certain fields be specified (that either Parthenon does not require or that it provides a default for).  For this case, I've implemented a very simple CheckRequired(block_name, parameter_name) method in the ParameterInput class.  The intent is that this is used in the driver constructor, which seemed like a reasonable place for an app to check to make sure it's requirements are satisfied.

3.*  I also implemented a CheckDesired method that just issues a warning about missing fields.

To @Yurlungur's comment on why at the driver level instead of the package level: my thought was that packages articulate requirements via use of the Get* methods, much like Parthenon internals.  A package developer could, instead, use GetOrAdd* methods, and that would allow some apps to use the package with the optional defaults, while others could require those fields via the CheckRequired method.

Also important to note: It is certainly not a requirement that apps specify required and/or desired parameters.  This is an opt-in capability.

example/calculate_pi/pi.hpp
jdolence: see my book below :wink:

Yurlungur:
Hmm. I don't object to this opt-in approach, exactly. But I guess I don't quite see why this is needed over and above package-level parameter input. Are there examples where an application will have different requirements than an individual package? If so, I guess something like this is worthwhile. 

The refactoring that you did that decouples things like time was much needed.

jdolence:
> Hmm. I don't object to this opt-in approach, exactly. But I guess I don't quite see why this is needed over and above package-level parameter input. Are there examples where an application will have different requirements than an individual package? If so, I guess something like this is worthwhile.
> 
> The refactoring that you did that decouples things like time was much needed.

So (for now at least) boundary conditions is an example.  They were required (the Mesh class constructor required them), but not all applications actually need them.  Example: pi calculator.  So they are still set in the Mesh class constructor, but they have a default (they defaulted to "none" before which caused the code to die.  Now they default to reflecting.) which means they aren't required in the input file.  But an app developer might (and probably should) require them in the input if they are relevant, as they are the in the advection example.

This whole feature (the CheckRequired/CheckDesired functions) only adds like ~30 lines to the code, so it's not like there's a lot of cost here.  And I for one, plan to use them when writing physics code on top of Parthenon.

src/outputs/vtk.cpp
```diff
@@ -242,191 +242,8 @@ void VTKOutput::WriteContainer(Mesh *pm, ParameterInput *pin, bool flag) {
...
-  // increment counters
-  output_params.file_number++;
-  output_params.next_time += output_params.dt;
-  pin->SetInteger(output_params.block_name, "file_number", output_params.file_number);
-  pin->SetReal(output_params.block_name, "next_time", output_params.next_time);
-
-  return;
+void VTKOutput::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm) {
+  throw std::runtime_error(std::string(__func__) + " is not implemented");
 }
```
JoshuaSBrown, line 247: Are we removing support for these file formats or am I missing something?

src/parameter_input.cpp
```diff
@@ -876,6 +901,35 @@ void ParameterInput::ForwardNextTime(Real mesh_time) {
...
+  bool missing = true;
+  if (DoesParameterExist(block, name)) {
+    missing = (GetComment(block, name) == "# Default value added at run time");
+  }
+  if (missing) {
+    std::cout << std::endl
+              << "### WARNING in CheckDesired:" << std::endl
+              << "Parameter file missing desired field <" << block << ">/" << name
+              << std::endl
+              << std::endl;
```
JoshuaSBrown, line 929: If we are not requiring certain attributes in the input file and are using default values, I think it would make sense to also print the default values to help the user know what is going on behind the scenes.

src/outputs/outputs.hpp
```diff
@@ -109,8 +109,10 @@ class OutputType {
   void CalculateCartesianVector(ParArrayND<Real> &src, ParArrayND<Real> &dst,
                                 Coordinates *pco);
   // following pure virtual function must be implemented in all derived classes
-  virtual void WriteOutputFile(Mesh *pm, ParameterInput *pin, bool flag) = 0;
-  virtual void WriteContainer(Mesh *pm, ParameterInput *pin, bool flag) { return; }
+  virtual void WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm) = 0;
+  virtual void WriteContainer(SimTime &tm, Mesh *pm, ParameterInput *pin, bool flag) {
+    return;
+  }
```
JoshuaSBrown, line 115: Since you are already changing this, it might make sense to make all the arguments const, I can't think of any reason why writing the contents to a file should change them.

src/outputs/outputs.hpp
```diff
@@ -126,7 +128,7 @@ class OutputType {
 class HistoryOutput : public OutputType {
  public:
   explicit HistoryOutput(OutputParameters oparams) : OutputType(oparams) {}
-  void WriteOutputFile(Mesh *pm, ParameterInput *pin, bool flag) override;
+  void WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm) override;
```
JoshuaSBrown, line 131: I would also recommend changing all `override` keywords to `final` for all the children of OutputType.

src/parameter_input.cpp
Yurlungur: This is a good idea. Maybe should do the same for the `GetIfExists` family of functions.

jlippuner: Second

src/parameter_input.hpp
```diff
@@ -23,6 +23,7 @@
 // information on the Athena++ input file format.
 
 #include <cstddef>
+#include <map>
```
jlippuner, line 26: This include doesn't seem to be needed

tst/unit/test_required_desired.cpp
```diff
@@ -0,0 +1,88 @@
...
+    std::stringstream ss;
+    ss << "<block1>" << std::endl
+       << "var1 = 0   # comment" << std::endl
+       << "var2 = 1,  & # another comment" << std::endl
+       << "       2" << std::endl
+       << "<block2>" << std::endl
+       << "var3 = 3" << std::endl
+       << "# comment" << std::endl
+       << "var4 = 4" << std::endl
+       << "var_default = 5 # Default value added at run time" << std::endl;
```
jlippuner, line 43: Can we also test:
- Multiple lines with `&` continuation
- `&` continuation without a comment
- `&` in a comment
- `&` in a string, e.g. 
```
<comment>
problem = This is a quick & simple test
```

tst/unit/test_required_desired.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+    // capture all std::cout
+    std::stringstream cout_cap;
+    std::streambuf *cout = std::cout.rdbuf(cout_cap.rdbuf());
+
+    WHEN("We require a paramter that has been provided") {
+      THEN("Nothing should happen") {
+        REQUIRE_NOTHROW(in.CheckRequired("block1", "var1"));
+        REQUIRE_NOTHROW(in.CheckRequired("block2", "var4"));
+        REQUIRE_NOTHROW(in.CheckRequired("block1", "var2"));
```
jlippuner, line 56: Shouldn't we also check that there is no output in `cout_cap`?

src/parameter_input.cpp
jlippuner: Second

src/outputs/vtk.cpp
jdolence: There was some dead code here.  We can easily add this back in later if it's something we want to support.

src/outputs/outputs.hpp
jdolence: They shouldn't, but this isn't quite as simple as just adding the const qualifier to all the arguments.  Doing that makes the compiler barf because, for example, calling member functions in the ParameterInput object discards the const qualifier.  Let's put this off.

tst/unit/test_required_desired.cpp
jdolence: good catch.  forgot to deal with this case.  it is now caught and throws a runtime_error if anything but a comment follows the continuation character.

tst/unit/test_required_desired.cpp
jdolence: We could.  Could do lots of other checks too.  In general, I'm not in favor of adding a test for everything imaginable.  There is a unit test in place, and we can add to it if/when we encounter issues.

src/parameter_input.cpp
```diff
@@ -876,6 +901,35 @@ void ParameterInput::ForwardNextTime(Real mesh_time) {
...
+       << "Parameter file missing required field <" << block << ">/" << name << std::endl
+       << std::endl;
+    throw std::runtime_error(ss.str());
+  }
+}
+
+void ParameterInput::CheckDesired(std::string block, std::string name) {
+  bool missing = true;
+  if (DoesParameterExist(block, name)) {
+    missing = (GetComment(block, name) == "# Default value added at run time");
```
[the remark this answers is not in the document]
jdolence: Probably is, but this works.

src/driver/driver.cpp
```diff
@@ -12,20 +12,27 @@
...
 
 DriverStatus EvolutionDriver::Execute() {
+  InitializeBlockTimeSteps();
+  SetGlobalTimeStep();
+  pouts->MakeOutputs(pmesh, pinput, &tm);
   pmesh->mbcnt = 0;
-  while ((pmesh->time < pmesh->tlim) &&
-         (pmesh->nlim < 0 || pmesh->ncycle < pmesh->nlim)) {
-    if (Globals::my_rank == 0) pmesh->OutputCycleDiagnostics();
+  while ((tm.time < tm.tlim) && (tm.nlim < 0 || tm.ncycle < tm.nlim)) {
```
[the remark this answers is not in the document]
jdolence: done.

scripts/python/phdf.py
```diff
@@ -121,11 +121,17 @@ def __init__(self, filename):
         try:
             f = h.File(filename,'r')
             # Read in the timestep attributes
-            ts = f['Timestep']
+            ts = f['Info']
```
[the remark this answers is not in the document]
jdolence: done.

example/advection/advection.hpp
```diff
@@ -31,6 +32,7 @@ using parthenon::MultiStageBlockTaskDriver;
 using parthenon::Outputs;
 using parthenon::ParameterInput;
 using parthenon::Real;
+using parthenon::SimTime;
 using parthenon::StateDescriptor;
 using parthenon::TaskID;
 using parthenon::TaskList;
```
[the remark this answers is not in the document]
jdolence: done.

src/outputs/outputs.hpp
jdolence: not here, not now, please.

src/parameter_input.cpp
jdolence: good idea!  done.

src/basic_types.hpp
jdolence: done.

jdolence:
this one has been sitting too long.  most comments addressed.  pulling the trigger.

PR #110, opened 2020-04-16
Attempt to define more clear public interface to parthenon

### LINKED ISSUE #100: Provide parthenon::prelude, perhaps with subsets
The way Rust solves this is they have a "prelude" module for common names. So we could perhaps have something like:

```c++
namespace parthenon { namespace prelude {
// Re-exports
using Real = ::parthenon::Real;
using MeshBlock = ::parthenon::MeshBlock;
using ParArrayND = ::parthenon::PayArrayND;
} }
```

And then in the user code, you could write:
```c++
using namespace parthenon::prelude;
```

You should be careful, though, to keep that `prelude` namespace fairly lean - only things that are used by essentially ever single user. You could potentially split the preludes into something like `parthenon::task::prelude` and `parthenon::app::prelude` to distinguish being names commonly used by apps and names commonly used by task implementations.

_Originally posted by @AndrewGaspar in https://github.com/lanl/parthenon/pull/75_

AndrewGaspar:
## PR Summary

The current include files have pretty generic names, which makes it confusing when importing parthenon as a library.

This change (attempts) to define a more clear public interface. It does this by defining a new "parthenon" include directory, which is what apps should use to include parthenon names. I have added the following header files:

* parthenon.hpp - Imports the entire parthenon public interface
* app.hpp - Imports just what's needed to define the entry point to a new parthenon app, defines `app::prelude`
* driver.hpp - Imports just what's needed to define a new task driver, defines `driver::prelude`
* task.hpp - Imports just what's needed to define new tasks, defines `task::prelude`
* prelude.hpp - Minimal imports used by any parthenon app, defines `prelude`

This is a very shallow crack at this problem - a more clear delineation between "public" and "private" names could and should probably be made, but this is a good first approximation. This change does not attempt to explicitly define which names are declared by which header files, beyond that `parthenon.hpp` _should_ export every public name in parthenon.

I ported the calculate_pi example to use these new headers and the preludes. I split off the task description into a separate source/header file to demonstrate the split.

Fixes #100

## PR Checklist

- [x] Code passes cpplint
- [x] New features are documented. (See README.md)
- [x] Adds a test for any bugs fixed. Adds tests for new features. (example tests new header files)
- [x] Code is formatted


CPPLINT.cfg
```diff
@@ -33,3 +33,4 @@ filter=-whitespace/comma  # 6027 errors
 filter=-whitespace/comments  # 881 errors
 filter=-whitespace/operators  # 5240 errors
 filter=-whitespace/parens  # 413 errors
+filter=-build/namespaces
```
Yurlungur, line 36:
```suggestion
filter=-build/namespaces

```

example/calculate_pi/CMakeLists.txt
```diff
@@ -12,8 +12,9 @@
...
 add_executable(
-    parthenon-example
+    pi-example
         calculate_pi.cpp
         pi.cpp
+        tasks.cpp
 )
-target_link_libraries(parthenon-example PRIVATE parthenon)
\ No newline at end of file
+target_link_libraries(pi-example PRIVATE parthenon)
```
Yurlungur, line 20:
```suggestion
target_link_libraries(pi-example PRIVATE parthenon)

```

src/parthenon/prelude.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+namespace parthenon {
+namespace prelude {
+using ::parthenon::CellVariable;
+using ::parthenon::Container;
+using ::parthenon::MeshBlock;
+using ::parthenon::ParthenonStatus;
+using ::parthenon::Real;
+using ::parthenon::Globals::my_rank;
+using ::parthenon::Globals::nranks;
```
jlippuner, line 34: Do we also need `ParArrayND` (and associated types)?

example/calculate_pi/calculate_pi.cpp
```diff
@@ -11,13 +11,19 @@
...
+#include <parthenon/app.hpp>
 
+// Local Includes
 #include "pi.hpp"
 
+// Preludes
+using namespace parthenon::app::prelude;
+
+// Self namespace
+using namespace calculate_pi;
```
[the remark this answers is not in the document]
AndrewGaspar: Oh, I'm just trying to indicate there's a method to the madness of the organization of the includes.

src/parthenon/prelude.hpp
AndrewGaspar: Oh, yeah, probably. Not sure how this didn't come up.

AndrewGaspar:
@Yurlungur could you bring these concerns to #100? I'll wait to merge until this has been hashed out.

src/parthenon/prelude.hpp
jdolence: Much (most/all?) of the code uses the CellVariable accessors, so a ParArrayND object is never pulled out.  That may (likely will) change.

Yurlungur:
> @Yurlungur could you bring these concerns to #100? I'll wait to merge until this has been hashed out.

Done. Thanks for the MR, btw.

src/driver/driver.hpp
```diff
@@ -44,13 +44,13 @@ class Driver {
 class SimpleDriver : public Driver {
  public:
   SimpleDriver(ParameterInput *pin, Mesh *pm, Outputs *pout) : Driver(pin, pm, pout) {}
-  DriverStatus Execute() { return DriverStatus::complete; }
+  DriverStatus Execute() override { return DriverStatus::complete; }
```
JoshuaSBrown, line 47: Why not declare as final?

AndrewGaspar: I wasn't sure what the intention was, so I was just trying to match the previous behavior, but more explicitly.

JoshuaSBrown:
@AndrewGaspar what is the status on this? I see some incoming pull requests that could benefit from this.

AndrewGaspar:
I'll polish it up right now.

AndrewGaspar:
@pgrete @jdolence let me know if this is satisfactory

example/calculate_pi/calculate_pi.cpp
```diff
@@ -11,13 +11,19 @@
 // the public, perform publicly and display publicly, and to permit others to do so.
 //========================================================================================
 
-#include "parthenon_manager.hpp"
+// Parthenon Includes
+#include <parthenon/app.hpp>
```
nmsriram: Andrew, why angle braces for project specific include files? (other locations as well)

JoshuaSBrown: @nmsriram nice catch, totally missed that.

AndrewGaspar: The breakdown isn't project-specific vs. non-project-specific. The breakdown is "relative to the current file" vs. "relative to include-paths". Since `parthenon/app.hpp` is being included relative to the include-path, I use angle brackets. If I was including it relative to the current file, I'd use quotes.

nmsriram:
> This is looks good and I like the approach.
> 
> Conceptually I'm also see the package/driver abstraction similar to @jdolence so I support his request to separate at the driver <-> package/task level rather than driver/package <-> task level.
> Having said that it may be worth (but I'm not requiring it) combining the `app.hpp` and `task.hpp` into a `package.hpp`

@pgrete have the commits post-this-comment satisfied your concerns?  If so, could you please approve?

Issue #111, opened 2020-04-17
Documenting common builds


Yurlungur:
The conversation on MR #96 got me thinking that we should really document the recipes for building our code on common machines. I'm thinking another markdown file in `docs` would do nicely. As we get `parthenon` running on more and more systems, the slightly different incantation per machine will become increasingly hard to remember. I would suggest we start with the power9 nodes on `Darwin` that @jlippuner just got working.

jlippuner:
Great idea! But I want to point out that @AndrewGaspar got it working on Darwin's Power 9 nodes. I just found that we need at least CMake 3.15.

Yurlungur:
Oops yes. Sorry, @AndrewGaspar! Mis-attributed there.

AndrewGaspar:
I'm of two minds on this - I wish that building was more streamlined for common systems. And I have no opposition to even hard-coding specializations for specific platforms as long as they're correctly scoped.

But, I don't really love having a bunch of LANL specific stuff in the code, partly because it's obscure and not testable by our non-LANL compatriots, and also because it may lead to LANL developers failing to keep the code building outside those environments. EAP actually has this problem where it's difficult to build on arbitrary platforms (though part of that is just because it's big).

I don't know, I don't think my concerns are actually that strong, it's more small things. I might just submit a PR that detects Darwin+XL (and we could add additional customizations if necessary). That would be even better than a doc because it would be in the code.

Yurlungur:
I'd rather avoid having LANL-specific stuff in the code. In my experience, hard-coded, system-specific pieces of a build system rot quickly when a machine falls out of use. I am also nervous about putting machine-specific things into `CMake` since that could become hard to maintain.

That's why I advocated documentation. Of course, I'm open to better solutions. Some large projects I've worked with in the past have had system-specific configuration automated via python scripts which auto-detect what machine you're on and then read in what to do from a database of machine-specific configuration files. See, .e.g, [nubhlight](https://github.com/lanl/nubhlight/) or [simfactory](http://simfactory.org/). I don't know how easy it is to integrate something like that into `cmake`.

AndrewGaspar:
Take a look at #114 and tell me what you think.

pgrete:
I'm also in favor of documenting builds.
We did sth similar for K-Athena as there are some subtle different between different machine even if they use the same architecture, see https://gitlab.com/pgrete/kathena/-/wikis/Building%20and%20running%20k%20athena

[...]

PR #114, opened 2020-04-18
WIP: Detect Darwin Power9 + XL

AndrewGaspar:
## PR Summary
This attempts to detect Darwin Power9 and search for compatible XL configurations. Helps user to load a valid GCC if they have not already done so.

Attempts to address #111.

With this, I was able to configure with the following relevant modules:
```
$ module list
Currently Loaded Modulefiles:
                                        3) cmake/3.17.0                                 5) gcc/9.3.0
                                    4) openmpi/p9/4.0.3-xlc_16.1.1.7-xlf_16.1.1.7   6) ibm/xlc-16.1.1.7-xlf-16.1.1.7
```

And spack installed HDF5 using just:
```
cmake ..
```

## PR Checklist

- [x] Code passes cpplint (CMake only)
- [x] New features are documented. (see CONTRIBUTING.md)
- [x] Adds a test for any bugs fixed. Adds tests for new features. (no new feature)
- [x] Code is formatted (CMake only)


platform/DetectPlatform.cmake
```diff
@@ -0,0 +1,30 @@
...
+    # Running on Darwin
+    if(CMAKE_HOST_SYSTEM_PROCESSOR STREQUAL "ppc64le")
+        # On a PowerPC system
+        if(CMAKE_CXX_COMPILER_ID STREQUAL "XL")
+            # With the IBM XL compiler
+            include(platform/lanl/DarwinPower9XL.cmake)
+            return()
+        endif()
+    endif()
+endif()
```
Yurlungur, line 30: This is quite a lot of boiler plate for each include. If we end up with a lot of these, this might look kind of painful. I'm of two minds for whether or not this should be condensed. On the one hand, it's a lot of visual noise. On the other hand, the intent is clear.

platform/lanl/DarwinPower9XL.cmake
```diff
@@ -0,0 +1,91 @@
...
+execute_process(
+    COMMAND ${GCC} --version
+    OUTPUT_VARIABLE GCC_VERSION_OUTPUT
+)
+
+string(
+    REGEX MATCH "g\\+\\+ \\(GCC\\) ([0-9]+\.[0-9]+\.[0-9]+)"
+    DISCARD ${GCC_VERSION_OUTPUT})
+
+set(XL_GCC_VERSION ${CMAKE_MATCH_1})
```
JoshuaSBrown, line 45: Where does CMAKE_MATCH_1 variable come from? Is this auto generated from the REGEX MATCH in the line above?

AndrewGaspar: Yeah, that's right.

Yurlungur:
I'd love to get this into master. Do other people object to a merge?

platform/lanl/DarwinPower9XL.cmake
```diff
@@ -0,0 +1,91 @@
...
+    message(
+        FATAL_ERROR
+        "Found GCC version ${XL_GCC_VERSION} at ${GCC}. This version is not \
+        configured with your IBM XL compiler, or it is not new enough (5.4.0 \
+        or newer required). Discovered configurations for GCC versions \
+        (${XL_COMPATIBLE_GCC_VERSIONS}) in ${XL_CONFIG_ROOT}"
+    )
+endif()
+
+list(APPEND CMAKE_CXX_FLAGS -F${XL_CONFIG_FILE})
```
jlippuner, line 91: Please add newline at the end of the file

platform/DetectPlatform.cmake
AndrewGaspar: I think the problem is I don't necessarily know what the common things we'll be checking on. If it becomes clear later, with a lot more of these, what variables are checked essentially every time, we can change it to something more concise.

platform/lanl/DarwinPower9XL.cmake
AndrewGaspar: You and your newlines!!! 😬

jlippuner:
Here is what I have found so far. This is on a Power9 node on Darwin using
```
Currently Loaded Modulefiles:
  1) cmake/3.17.0
  2) ibm/xlc-16.1.1.7-xlf-16.1.1.7
  3) openmpi/p9/4.0.3-xlc_16.1.1.7-xlf_16.1.1.7
  4) clang/8.0.1
  5) hdf5-1.10.5-xl-16.1-ttt5wbw
  6) gcc/8.3.0
  7) cuda/10.1
```
and also
```
export NVCC_WRAPPER_DEFAULT_COMPILER=xlc++
```

Building with OpenMP works. This works on this branch and also my `jlippuner/small_fixes` branch, which doesn't do anything special about XL but just has the commit `don't use std::ofstream::operator= (it's deleted)`, which I pushed to this branch as well. So, this works for me:
```
CXX=xlc++ CC=xlc cmake -DKokkos_ENABLE_OPENMP=ON -DKokkos_ARCH_POWER9=ON -DKokkos_ARCH_VOLTA70=ON ..
```
It builds and all tests pass.

Building with CUDA (without OpenMP) also works, **but not on this branch.** It works on the `jlippuner/small_fixes` branch for me.
```
CXX=`pwd`/../external/Kokkos/bin/nvcc_wrapper CC=xlc cmake -DKokkos_ENABLE_CUDA=ON -DDISABLE_OPENMP=ON -DKokkos_ARCH_POWER9=ON -DKokkos_ARCH_VOLTA70=ON ..
```
On this branch, I get (for each file)
```
/projects/opt/ppc64le/p9/gcc/8.3.0/lib/gcc/powerpc64le-unknown-linux-gnu/8.3.0/../../../../include/c++/8.3.0/utility(305): error: alias template "std::_Build_index_tuple<_Num>::_IdxTuple" is not compatible with template template parameter "S"
```
On the `jlippuner/small_fixes` branch Parthenon compiles and all tests pass.

What I have not been able to get to work yet is CUDA + OpenMP. When I use the `nvcc_wrapper` compiler, CMake can no longer detect OpenMP. I'm still trying to figure this out.

jlippuner:
So I was able to build with CUDA + OpenMP now. I hit the same issue @AndrewGaspar mentioned with a `-V` flag. But it seems to only be an issue when the Kokkos `nvcc_wrapper` is used in conjunction with `xlc++`. I modified the `nvcc_wrapper` script to drop the `-V` flag and then CMake finds OpenMP with the `nvcc_wrapper` script. Everything builds and tests pass. This is on my `jlippuner/small_fixes` branch without the changes from this PR. Will also test it with this PR tomorrow.

jlippuner:
I finally figured out what the issue is with `nvcc_wrapper`. It looks like it gets confused when this gets passed (by CMake, I guess): `-Xcompiler -o out.o` Then the wrapper will consume the `-Xcompiler` flag, which takes the next argument (i.e. `-o`) and adds that to its list of `-Xcompiler` flags, but it leaves the `out.o` argument. The `out.o` argument is no consumed in the next pass but because `-o` is gone, it thinks it's a library to link against and so it sticks `-Xlinker` in front and that messes everything up.

This caused problems for me when I try to compile with OpenMP and CUDA on both my linux workstation (using gcc) and Power 9 using xlc++. The fix is to not consume the `-o` argument as part of processing `-Xlinker`. I'm pushing a fixed `nvcc_wrapper` here and I will make a pull request in Kokkos to get this fixed upstream.

jlippuner:
I also just found that it's only working for me with `cuda/10.2` loaded and not with `cuda/10.1` (which was working for me before I fixed the `nvcc_wrapper`...). It still doesn't work on this branch with the `XL` config file, but it works without any config file.

AndrewGaspar:
We may make this defunct - it seems that the old "System GCC" works "fine", but apparently the config file to use a newer GCC is broken on some configurations.

This may have issues down the road if we want newer GCC features

Yurlungur:
That's too bad. At the very least #138 does help mitigate issues like this. That said, I think this is valuable approach, so we shouldn't delete this branch, in my opinion. I'd like to keep this around as an example solution for similar problems we may hit in the future.

jlippuner:
Will revisit if needed

PR #116, opened 2020-04-21
Fix space typedefs. Add GetHostMirrorAndCopy

pgrete:
## PR Summary

Addressed two items
- Cleanup of the `DevSpace` and `HostSpace` types that incorrectly were used for both execution and memory spaces (lead to problems with the wrappers when enabling UVM). Now there are `DevExecSpace`, `DevMemSpace`, and `HostMemSpace` to distinguish between those.
- Add a `GetHostMirrorAndCopy` function to the `ParArrays`s which does what it name stands for. This will be useful, e.g., during the simulation setup in the process of Kokkosizing the remainder of the codebase.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/docs/parthenon_arrays.md
+++ b/docs/parthenon_arrays.md
@@ -80,6 +80,10 @@ my_mirror.DeepCopy(my_array);
 `ParArrayND` provides two convenience functions, `GetHostMirror()` and
 `GetDeviceMirror()` which put a mirror on the host and device
 respectively.
+In addition, `GetHostMirrorAndCopy()` creates a new and deep copies the content, e.g.,
+```C++
+auto my_host_array = my_array.getHostMirrorAndCopy();
+```
 
 ### A note on templates
 
@@ -89,7 +93,7 @@ aliases available are as follows:
 
 ```C++
 template<typename T, typename Layout=LayoutWrapper>
-using device_view_t = Kokkos::View<T******,Layout,DevSpace>;
+using device_view_t = Kokkos::View<T******,Layout,DevMemSpace>;
 
 template<typename T, typename Layout=LayoutWrapper>
 using host_view_t = typename device_view_t<T,Layout>::HostMirror;
--- a/docs/parthenon_arrays.md
+++ b/docs/parthenon_arrays.md
@@ -80,7 +80,8 @@ my_mirror.DeepCopy(my_array);
 `ParArrayND` provides two convenience functions, `GetHostMirror()` and
 `GetDeviceMirror()` which put a mirror on the host and device
 respectively.
-In addition, `GetHostMirrorAndCopy()` creates a new and deep copies the content, e.g.,
+In addition, `GetHostMirrorAndCopy()` creates a new `ParArrayND` on the host
+with identical layout and deep copies the content, e.g.,
 ```C++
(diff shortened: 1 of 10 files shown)


src/parthenon_arrays.hpp
```diff
@@ -144,6 +143,12 @@ class ParArrayNDGeneric {
     Kokkos::deep_copy(d6d_, src.Get());
   }
 
+  auto GetHostMirrorAndCopy() {
+    auto host_mirror = GetHostMirror();
+    Kokkos::deep_copy(host_mirror.Get(), d6d_);
+    return host_mirror;
+  }
+
```
Yurlungur, line 151: `GetHostMirror` returns a `ParArrayND`, not a mirror view. So if you want to mimic the `ParArrayND` interface, you need to not use `Kokkos::deep_copy`. Also you should probably use `Kokkos::create_mirror_view_and_copy`, which (I think) will be more efficient than `create_mirror_view` and then `deep_copy`. @dholladay00 might be able to correct me.

Summarizing, it should look like:

```C++
template<typename MemSpace>
auto GetMirrorAndCopy(MemSpace const &memspace) {
  auto mirror = Kokkos::create_mirror_view_and_copy(memspace, d6d_);
  return ParArrayNDGeneric<decltype(mirror)>(mirror);
}
auto GetHostMirrorAndCopy() { return GetMirrorAndCopy(Kokkos::HostSpace()); }
auto GetDeviceMirrorAndCopy() { return GetMirrorAndCopy(Kokkos::DefaultExecutionSpace()); }
```

dholladay00: I think it's cleaner, but I'm not sure that it's any more efficient since deep_copy will no-op if the views alias the same data. Its primary benefit from my perspective is when appending traits to a view. For example, if I have some inputs that are const and being passed in, I can do `View<const T**, Layout, DevMemSpace, RandomAccess> inputs = create_mirror_view_and_copy(DevMemSpace, inputs_host)` rather than having to create a non-const view, deep_copy into it, and then assign to the const random access view.

Yurlungur: Hmm. Well that's probably a reason to use `create_mirror_view_and_copy` since `ParArrayND` can in principle carry around traits.

pgrete: I see. Thanks for those suggestions. I'll update accordingly.

docs/parthenon_arrays.md
```diff
@@ -80,6 +80,10 @@ my_mirror.DeepCopy(my_array);
 `ParArrayND` provides two convenience functions, `GetHostMirror()` and
 `GetDeviceMirror()` which put a mirror on the host and device
 respectively.
+In addition, `GetHostMirrorAndCopy()` creates a new and deep copies the content, e.g.,
```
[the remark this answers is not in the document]
pgrete: Yes :D I promise the words were there in my mind!

pgrete:
I made the requested changes.
Please note that I did not introduce the `GetDeviceMirrorAndCopy` as I'm not sure there's a use case for this (we didn't have one in `K-Athena`). I'm happy to add this later if required.

pgrete:
Got two approvals. Pulling the trigger.

Issue #81, opened 2020-03-31
ability to disable exceptions


Yurlungur:
While experimenting with performance and vectorization in #65, I discovered that many loops aren't getting vectorized because of exception handling. If I try to disable exception handling, the compiler yells at me. We need a way to disable all exception handling in the code. `trhow`, `try`, and `catch` need to be disableable globally.

dholladay00:
This sounds fairly heavy, I've been curious about replacing the `nothrow` stuff in the ndarray accessors with `noexcept`. Could you see how that affects vectorization?

Yurlungur:
> This sounds fairly heavy, I've been curious about replacing the `nothrow` stuff in the ndarray accessors with `noexcept`. Could you see how that affects vectorization?

Thanks, @dholladay00 that does help. I'm still seeing some failed vectorizations inside `Kokkos` internals, though. And we can't really fix that unless we disable exception handling at the compiler level.

Yurlungur:
That said, it looks like the loops we actually care about are vectorized. So maybe it's not worth investigating further.

AndrewGaspar:
I believe Athena did, at the very least, have the ability to disable them, so it might be worth having the option, even if it's not he default.

pgrete:
Maybe @jmstone can share their Athena++ experience here with regard to this topic.
IIRC exception were enabled in the "outer" parts of the main loop and output handling, but vectorization of the physics kernels were not affected by that.

jmstone:
In Athena++ we only implemented exception handling in the initialization and I/O routines, precisely because it interfered with performance.  Moreover, you don't want to keep checking for exceptions every time the main integration loop is executed.  If the code makes it through the loop once, it likely will work for the rest of the iterations (although this is not guaranteed), and so it seems a waste to check for exceptions inside the main loop.  I would be very happy to stick to this strategy since it works well in Athena++.

What would be useful in the main loop would be a "debug mode" that will stop execution when a NaN occurs, and save the state so a debugger can be used.  This was not implemented in Athena++ but would be very useful.

AndrewGaspar:
@Yurlungur - have those exceptions been replaced by asserts?

Yurlungur:
@AndrewGaspar I think this has been cleaned up sufficiently. Exceptions are still availabe for non performance-critical code segments and the performance critical ones have been fixed thanks to #118 by @brryan

JoshuaSBrown:
bvals_refine.cpp
[...]

PR #118, opened 2020-04-21
Provide macros for error checking

brryan:
## PR Summary

Provide simple, informative macros `PARTHENON_REQUIRE(condition, message)` and `PARTHENON_FAIL(message)` as an alternative to throwing exceptions. These macros print some output that includes `__FILE__` and `__LINE__` macros, removing the need to include some identifying information in some error messages that could otherwise be ambiguous. 

Related to #81.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/src/utils/error_checking.hpp
+++ b/src/utils/error_checking.hpp
@@ -19,34 +19,37 @@
 //! \file error_checking.hpp
 //  \brief utility macros for error checking
 
+#include <Kokkos_Core.hpp>
 #include <iostream>
 
-#define PARTHENON_REQUIRE(condition, message)                                  \
-  if (!(condition)) {                                                          \
-    parthenon::ErrorChecking::require(#condition, message, __FILE__, __LINE__);\
+#define PARTHENON_REQUIRE(condition, message)                                            \
+  if (!(condition)) {                                                                    \
+    parthenon::ErrorChecking::require(#condition, message, __FILE__, __LINE__);          \
   }
 
-#define PARTHENON_FAIL(message)                                                \
-  parthenon::ErrorChecking::fail(message, __FILE__, __LINE__);                 \
+#define PARTHENON_FAIL(message)                                                          \
+  parthenon::ErrorChecking::fail(message, __FILE__, __LINE__);
 
 namespace parthenon {
 namespace ErrorChecking {
 
-inline void require(std::string const &condition, std::string const &message,
-                    std::string const &filename, int const linenumber) {
-  std::cout << "!!! PARTHENON ERROR !!!" << std::endl;
-  std::cout << "  Condition:   " << condition << std::endl;
-  std::cout << "  Message:     " << message << std::endl;
-  std::cout << "  File:        " << filename << std::endl;
-  std::cout << "  Line number: " << linenumber << std::endl;
+KOKKOS_INLINE_FUNCTION
+void require(std::string const &condition, std::string const &message,
+             std::string const &filename, int const linenumber) {
+  fprintf(stderr,
+          "!!! PARTHENON ERROR !!!\n  Condition:   %s\n  Message:     %s\n  File:       "
+          "%s\n  Line number: %i\n",
+          condition.c_str(), message.c_str(), filename.c_str(), linenumber);
+  exit(-1);
 }
 
-inline void fail(std::string const &message, std::string const &filename,
-                 int const linenumber) {
-  std::cout << "!!! PARTHENON ERROR !!!" << std::endl;
-  std::cout << "  Message:     " << message << std::endl;
-  std::cout << "  File:        " << filename << std::endl;
-  std::cout << "  Line number: " << linenumber << std::endl;
+KOKKOS_INLINE_FUNCTION
+void fail(std::string const &message, std::string const &filename, int const linenumber) {
+  fprintf(
+      stderr,
+      "!!! PARTHENON ERROR !!!\n  Message:     %s\n  File:       %s\n  Line number: %i\n",
+      message.c_str(), filename.c_str(), linenumber);
+  exit(-1);
 }
 
 } // namespace ErrorChecking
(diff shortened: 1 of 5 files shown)


src/utils/error_checking.hpp
```diff
@@ -0,0 +1,55 @@
...
+namespace parthenon {
+namespace ErrorChecking {
+
+inline void require(std::string const &condition, std::string const &message,
+                    std::string const &filename, int const linenumber) {
+  std::cout << "!!! PARTHENON ERROR !!!" << std::endl;
+  std::cout << "  Condition:   " << condition << std::endl;
+  std::cout << "  Message:     " << message << std::endl;
+  std::cout << "  File:        " << filename << std::endl;
+  std::cout << "  Line number: " << linenumber << std::endl;
```
Yurlungur, line 41: This should probably be `std::cerr` instead of `std::cout`.

src/utils/error_checking.hpp
```diff
@@ -0,0 +1,55 @@
...
+}
+
+inline void fail(std::string const &message, std::string const &filename,
+                 int const linenumber) {
+  std::cout << "!!! PARTHENON ERROR !!!" << std::endl;
+  std::cout << "  Message:     " << message << std::endl;
+  std::cout << "  File:        " << filename << std::endl;
+  std::cout << "  Line number: " << linenumber << std::endl;
```
Yurlungur, line 49: Here too.

Yurlungur:
Oh one more thought. We might want to use `fprintf` to stderr and `KOKKOS_INLINE_FUNCTION` so it will work on device.

brryan:
@Yurlungur yes the macros should call `exit`... oops. Also thanks for the work-on-device tip.

JoshuaSBrown:
Nice Brryan!

JoshuaSBrown:
I am wondering though if simply calling exit is the best solution, it might be better to throw so that we have the option to unwind the stack.

brryan:
@JoshuaSBrown Thanks! I switched from `exit()` to `runtime_error`. If this isn't ideal (historically I've been more of a C99er) let me know.

@forrestglines I agree `DEBUG` versions of these calls that can be removed by the preprocessor are desirable, I've just added these. I'm only using a local `DEBUG_MODE_` flag in `error_checking.hpp` right now, though, and I think a unit test of this file, while good, won't work well without a global parthenon DEBUG macro. Have there been thoughts already on how/whether to put in such a flag?

Yurlungur:
> @JoshuaSBrown Thanks! I switched from `exit()` to `runtime_error`. If this isn't ideal (historically I've been more of a C99er) let me know.

@JoshuaSBrown I understand the desire for a stack trace, but I would rather avoid throwing errors, or make it a switch set with a `#define`. `throw` will not work on the GPU and is also  carries a significant performance hit. I raised this issue in #81 

> @forrestglines I agree `DEBUG` versions of these calls that can be removed by the preprocessor are desirable, I've just added these. I'm only using a local `DEBUG_MODE_` flag in `error_checking.hpp` right now, though, and I think a unit test of this file, while good, won't work well without a global parthenon DEBUG macro. Have there been thoughts already on how/whether to put in such a flag?

We can perhaps implement a global `PARTHENON_DEBUG` macro. Alternatively we could use `#ifndef NDEBUG` so your code is disabled when `assert` is disabled. This would also have the advantage that `CMAKE_BUILD_TYPE=Debug` would enable your macro.

JoshuaSBrown:
> `throw` will not work on the GPU and is also carries a significant performance hit.

I thought that might be an issue.

brryan:
@JoshuaSBrown no worries, thanks for looking over this :)

@Yurlungur If you'd like me to add a `PARTHENON_DEBUG` I'm happy to do that in this PR.

Yurlungur:
@brryan that feels like it belongs in a bigger discussion. For now, use `#ifndef NDEBUG`

Yurlungur:
Looks good now, @brryan. One weird issue remaining: CI does not appear to have triggered. We might have to move your PR into a local branch so we get CI.

brryan:
@Yurlungur Thanks Jonah, I did forget to run cpplint, I just did and fixed the errors. 

Yes I have not seen the CI run for this yet.

Yurlungur:
@brryan Can you try making a branch in the lanl repository and seeing if that triggers CI? If the CI passes in the branch (and others approve), I think we can merge this in and then close the branch.

Yurlungur:
I don't think code tests are required for this and the cpplint test says formatting is fine. Some I'm going to check off the other tasks.

Yurlungur:
This looks like it's in good shape. Does anybody object to me merging once the checks pass?

nmsriram:
Is it also worth crating a PARTHENON_WARN(level, message) macro so that we can adjust verbosity of runs?  This could also be turned off for non-debug runs.

brryan:
@nmsriram I like the idea of `PARTHENON_WARN`/`PARTHENON_DEBUG_WARN` macros, just added this. Right now they only take `message` as an argument, I'm not sure what the best way to add verbosity levels would be (my preference now is to set this at compile time for performance). Others might have thoughts about this, maybe that could wait along with a `PARTHENON_DEBUG` flag for another PR?

nmsriram:
@brryan thank you so much for adding those.  Agree on waiting for another PR.

nmsriram:
@brryan, could you please identify why the checks are not passing and correct this.  If it is the case that GitHub cannot handle forks, please create a branch in the main repository and modify pull request accordingly.

Yurlungur:
> @brryan, could you please identify why the checks are not passing and correct this. If it is the case that GitHub cannot handle forks, please create a branch in the main repository and modify pull request accordingly.

@brryan may be out for a while. I'll reach out to him and see if this can be done by someone else. I can shepherd it to completion if need be.

Yurlungur:
@par-hermes format

Yurlungur:
Please approve #153. It's the same but CI triggers and passes.

Yurlungur:
Merged in as #153. Closing this one.

PR #120, opened 2020-04-22
Renamed class members with leading underscores

nmsriram:
changed all members that had a leading underscores to a trailing underscore to be consistent with C++ practices


pgrete:
Got enough approvals. Pulling the trigger.

PR #124, opened 2020-04-23
Add regression testing

JoshuaSBrown:
## PR Summary

This pull request creates a regression testing framework on which to build. It uses a python script to execute a parthenon driver and check the output. The python script is called from within the ctest framework. It uses the pi calculator as an example and provides documentation on how to add more regression tests.

There are a few points worth mentioning, there is not currently a clear means from which to pull data out of a driver that is specific to the driver (see issue https://github.com/lanl/parthenon/issues/119), as such in the calculate pi example the calculated values are output to a summary.txt file which is read in to the python script. 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


tst/regression/run_test.py
```diff
@@ -0,0 +1,130 @@
...
+            if path != "":
+                folders.append(path)
+            break
+    return folders
+
+def checkRunScriptLocation(run_test_py_path):
+  
+    """ Check that run_test is in the correct folder """
+    folders = SplitPathIntoFolders(run_test_py_path)
+    if( os.path.join(folders[2], folders[1], folders[0]) != "parthenon/tst/regression" ):
```
Yurlungur, line 46: Hmm. I see now why you wrote `SplitPathIntoFolders`. I think Python has built in functionality for this. If I recall, it's something like:
```python
if not os.path.normpath(run_test_py_path).startswith(os.path.normpath("parthenon/tst/regression"))
```
The normpaths might be unnecessary, but `run_test_py_path` does need to be a path object, not a string.

tst/regression/run_test.py
```diff
@@ -0,0 +1,130 @@
...
+  
+    """ Check that run_test is in the correct folder """
+    folders = SplitPathIntoFolders(run_test_py_path)
+    if( os.path.join(folders[2], folders[1], folders[0]) != "parthenon/tst/regression" ): 
+        error_msg = "Cannot run run_test.py, it is not in the correct directory, must be "
+        error_msg += "kept in parthenon/tst/regression"
+        raise TestError(error_msg)
+
+    """ Check that test_suites folder exists """
+    if( not os.path.isdir(os.path.join(run_test_py_path,'test_suites')) ):
```
Yurlungur, line 52: Python doesn't need the parentheses. :)
```suggestion
    if not os.path.isdir(os.path.join(run_test_py_path,'test_suites')): 
```

tst/regression/run_test.py
```diff
@@ -0,0 +1,130 @@
...
+    pass
+
+# Execute main function
+if __name__ == '__main__':
+    help_msg = ('name of the test directory, relative to test_suites/, '
+               'excluding .py.')
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--test_dir','-t',
+                        type=str,
+                        default=None,
```
Yurlungur, line 94: If `required=True`, you don't need `default=None`.

tst/regression/run_test.py
```diff
@@ -0,0 +1,130 @@
...
+                        nargs=1,
+                        required=True,
+                        help='path to input file, to pass to driver')
+
+    parser.add_argument('--mpirun',
+                        default='mpirun',
+                        # 2x MPI, Slurm, PBS/Torque, LSF, Cray ALPS
+                        choices=['mpirun', 'mpiexec', 'srun', 'qsub', 'lsrun', 'aprun'],
+                        help='change MPI run wrapper command (e.g. for job schedulers)')
```
Yurlungur, line 117: Might be nice to offer an option with no `mpi` so it just runs as `./executable`

tst/regression/CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
...
+# Note each new test must append exactly 1 item to each of the CMAKE variables
+
+function(setup_test dir arg)
+  separate_arguments(arg) 
+  add_test( NAME regression_test:${dir} COMMAND python3 "${CMAKE_CURRENT_SOURCE_DIR}/run_test.py" ${arg} --test_dir "${CMAKE_CURRENT_SOURCE_DIR}/test_suites/${dir}")
+  set_tests_properties(regression_test:${dir} PROPERTIES LABELS "regression" )        
+endfunction()
+
+list(APPEND TEST_DIRS calculate_pi)
+list(APPEND TEST_ARGS "--driver ${CMAKE_BINARY_DIR}/example/calculate_pi/parthenon-example --driver_input ${CMAKE_CURRENT_SOURCE_DIR}/test_suites/calculate_pi/parthinput.regression")
```
Yurlungur: Some of the options, such as the mpi options, might be test system dependent. So I don't love that they're hardcoded here. Maybe we could add some system-dependent properties analogous to what @AndrewGaspar did in #114? Alternatively, `python` could parse options from some declarative machine database, rather than from the command line.

@brryan  and I did something along those lines in [nubhlight](https://github.com/lanl/nubhlight)

tst/regression/run_test.py
```diff
@@ -0,0 +1,130 @@
...
+
+# Exception for unexpected behavior by individual tests
+class TestError(RuntimeError):
+    pass
+
+# Execute main function
+if __name__ == '__main__':
+    help_msg = ('name of the test directory, relative to test_suites/, '
+               'excluding .py.')
+    parser = argparse.ArgumentParser()
```
Yurlungur, line 91: Might want to add a docstring to `ArgumentParser` itself to describe what the program does from the CLI. I.e., something along the lines of
```suggestion
    parser = argparse.ArgumentParser("Automatically run a Parthenon regression test.")
```

tst/regression/utils/test_case.py
```diff
@@ -0,0 +1,162 @@
...
+class Parameters():
+    def __init__(self):
+        driver_path = ""
+        driver_input_path = ""
+        test_path = ""
+        output_path = ""
+        mpi_cmd = ""
+        mpi_opts = ""
+
+class TestCase:
```
Yurlungur, line 32: A model I like better than defining `analyze` as a function in each test script is to make `TestCase` an abstract base class, with `analyze` as an undefined member function. Then every test defines a new class which inherits from `TestCase` and defines `analyze`.

JoshuaSBrown: That's probably a better design.

tst/regression/run_test.py
JoshuaSBrown: This is an improvement, thanks!

CONTRIBUTING.md
```diff
@@ -76,4 +76,101 @@ the proper formatting.
...
+list(APPEND TEST_DIRS foo_test)
+list(APPEND TEST_ARGS "--driver location_of_foo_driver --driver_input Location_of_foo_input_deck")
+```
+
+For the calculate pi example shown above this consists of adding parameters to two cmake lists
+
+```
+list(APPEND TEST_DIRS calculate_pi)
+list(APPEND TEST_ARGS "--driver ${CMAKE_BINARY_DIR}/example/calculate_pi/parthenon-example --driver_input ${CMAKE_CURRENT_SOURCE_DIR}/test_suites/calculate_pi/parthinput.regression")
+```
```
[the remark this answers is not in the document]
JoshuaSBrown: How would you implement globbing in this case? There is no guarantee that the driver, or input deck will be located in the same general locations.

tst/regression/run_test.py
```diff
@@ -0,0 +1,130 @@
...
+""" To prevent littering up imported folders with .pyc files"""
+sys.dont_write_bytecode = True
+
+# Parthenon modules
+import utils.test_case as tc 
+
+""" Split a path into folders """
+def SplitPathIntoFolders(path):
+    folders = []
+    while 1:
```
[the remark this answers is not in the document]
JoshuaSBrown: This function has been removed.

tst/regression/CMakeLists.txt
JoshuaSBrown: Which file in that repo were you doing this?

JoshuaSBrown: Compiler dependent variables/architecture dependent variables can be passed from cmake to the python script if there is a need for it.

JoshuaSBrown: Seeing that CMake should probably know about these things in order to appropriately build the code, that would be my preference. 

Though I would need some help understanding the specifics of what needs to be passed in...?

Yurlungur: > Which file in that repo were you doing this?

The [config](https://github.com/lanl/nubhlight/blob/master/script/config.py) script reads in machine files from the [machine](https://github.com/lanl/nubhlight/tree/master/script/machine) directory and uses them to set parameters.

> Compiler dependent variables/architecture dependent variables can be passed from cmake to the python script if there is a need for it.

The big thing I'm thinking of is, for example, the `mpirun` command, which will not always be `mpirun -np N`.

Yurlungur: If it's possible to pass it into `cmake` from somewhere else, I'm happy with that solution. I just don't know very much about `cmake` so I don't think of it as a solution. :)

CONTRIBUTING.md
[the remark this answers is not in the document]
Yurlungur: We could perhaps enforce more structure. Like put a config file in a specific location, and that file specifies driver, input deck, etc locations? Something like that? Then we could glob for those config files?

Heck, the `cmake` commands could be globbed for and included and the cmake can hardcode things? Similar to what @AndrewGaspar did in #114 ?

tst/regression/CMakeLists.txt
JoshuaSBrown: So in cmake if MPI is found you have access to 
MPIEXEC_EXECUTABLE
    Executable for running MPI programs, if such exists.

Which can be used as a variable. But currently I'm dealing with a different problem with the ci. CI is crashing because it does not want me to run mpi as root which is the default for docker images.

CONTRIBUTING.md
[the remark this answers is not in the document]
JoshuaSBrown: Instead of having everything in one cmake file we would that information in a config file or multiple config files. I guess I'm not really seeing what the advantage is?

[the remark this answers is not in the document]
Yurlungur: I guess the advantage I see is that you could have each configuration feel more... self-contained and modular? It feels more natural and manageable to add a new file, that satisfies a format, than to add new code to an existing file. I'm envisioning that someday we have like 10 regression tests.

I say that since this feels less like code and more like build system infrastructure.  Of course, this could just be personal preference on my part.

tst/regression/CMakeLists.txt
Yurlungur: Ah, I see. Cool. In that case, we should provide logic in cmake for if mpi is not found. 

Calculate pi should work without mpi, so that's a solution for the short term... In the longer term, it's not great though. 

I think `openmpi` supports a flag `--allow-run-as-root` which should allow it to behave on docker? Not sure.

JoshuaSBrown:
@par-hermes format

CONTRIBUTING.md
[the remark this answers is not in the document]
JoshuaSBrown: Ok, I see your point, before doing more work on this point though I will wait to hear if anyone else has a strong opinion about it.

tst/regression/CMakeLists.txt
JoshuaSBrown: Right now I have the regression test running without mpi. I'm guessing that it would be advantageous to have the same test running in serial and in parallel.

Yurlungur: Yes, it definitely would.

JoshuaSBrown:
@par-hermes format

JoshuaSBrown:
https://github.com/lanl/parthenon/issues/128

JoshuaSBrown:
https://github.com/lanl/parthenon/issues/129

JoshuaSBrown:
TODO: Current implementation is not general enough to look at convergence criteria for slight difference in input parameters.

JoshuaSBrown:
> Thanks for putting this together and especially for making sure it's well documented!
> 
> I just pushed some changed that should make the changes work with MPI and Cuda on the CI infrastructure.
> In addition to the "allow-run-as-root" fix, the regression tests currently require UVM in order to run as we haven't ported all the Parthenon infrastructure to Kokkos yet (that's why the problem with inaccessible spaces came up).

Thanks for the help! If I could give you a round of applause I would. :)

tst/regression/CMakeLists.txt
JoshuaSBrown: This has been implemented.

CONTRIBUTING.md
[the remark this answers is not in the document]
JoshuaSBrown: @Yurlungur because this seems to fall in the preference category, and I don't want to rework something that is working I'm going to count this issue as resolved. It can be revisited later if needed.

tst/regression/test_suites/calculate_pi/calculate_pi.py
```diff
@@ -0,0 +1,94 @@
...
+sys.dont_write_bytecode = True
+
+class TestCase(utils.test_case.TestCaseAbs):
+    def Prepare(self,parameters):
+        """
+        Any preprocessing that is needed before the drive is run can be done in
+        this method
+
+        This includes preparing files or any other pre processing steps that
+        need to be implmeneted.  The method also provides access to the
```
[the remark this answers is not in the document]
JoshuaSBrown: Good catch.

CONTRIBUTING.md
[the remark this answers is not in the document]
Yurlungur: That's fine.

.gitlab-ci.yml
```diff
@@ -31,21 +31,34 @@ parthenon-cuda-short:
...
 #    - export CUDA_VISIBLE_DEVICES=0
     - mkdir build-cuda-debug
     - cd build-cuda-debug
     - cmake -DCMAKE_BUILD_TYPE=Debug -DHDF5_ROOT=/usr/local/hdf5/parallel
       -DKokkos_ENABLE_OPENMP=True -DKokkos_ARCH_WSM=True
       -DKokkos_ENABLE_CUDA=True -DKokkos_ARCH_PASCAL61=True
+      -DMPIEXEC_PREFLAGS="--allow-run-as-root"
       -DCMAKE_CXX_COMPILER=${PWD}/../external/Kokkos/bin/nvcc_wrapper
-      ../ && make -j${J} && ctest -j${J} -LE performance
+      ../ && make -j${J} && ctest -j${J} -LE 'performance|regression'
```
Yurlungur, line 44: :+1:

example/calculate_pi/pi_driver.cpp
```diff
@@ -88,6 +91,11 @@ parthenon::DriverStatus PiDriver::Execute() {
               << "PI = " << pi_val << "    rel error = " << (pi_val - M_PI) / M_PI
               << std::endl
               << std::endl;
+
+    std::fstream fs;
+    fs.open("summary.txt", std::fstream::out);
+    fs << "PI = " << pi_val << "    rel error = " << (pi_val - M_PI) / M_PI << std::endl;
```
Yurlungur, line 97: Nitpick: I don't love that the value of pi and the relative error are on the same line. I'd replace the 4 spaces with a newline. That's a little more human-readable IMHO.

tst/regression/CMakeLists.txt
```diff
@@ -0,0 +1,64 @@
...
+    separate_arguments(arg) 
+    add_test( NAME regression_mpi_test:${dir} COMMAND python ${CMAKE_CURRENT_SOURCE_DIR}/run_test.py  --mpirun ${MPIEXEC_EXECUTABLE} --mpirun_opts "${MPIEXEC_NUMPROC_FLAG} ${MPIEXEC_MAX_NUMPROCS} ${MPIEXEC_PREFLAGS}" ${arg} --test_dir ${CMAKE_CURRENT_SOURCE_DIR}/test_suites/${dir})
+    set_tests_properties(regression_mpi_test:${dir} PROPERTIES LABELS "regression,mpi" RUN_SERIAL ON )
+  else()
+    message(STATUS "MPI not found, not building regression tests with mpi")
+  endif()
+endfunction()
+
+list(APPEND TEST_DIRS calculate_pi)
+list(APPEND TEST_ARGS "--driver ${CMAKE_BINARY_DIR}/example/calculate_pi/pi-example --driver_input ${CMAKE_CURRENT_SOURCE_DIR}/test_suites/calculate_pi/parthinput.regression")
```
Yurlungur, line 45: I'm not thrilled that these are lists, making it a little hard to map dirgs to args when debugging. But I think that's just cmake and we should leave it.

tst/regression/test_suites/calculate_pi/calculate_pi.py
```diff
@@ -0,0 +1,94 @@
...
+import math
+import numpy as np
+import sys
+import os
+import utils.test_case
+
+""" To prevent littering up imported folders with .pyc files or __pycache_ folder"""
+sys.dont_write_bytecode = True
+
+class TestCase(utils.test_case.TestCaseAbs):
```
Yurlungur, line 28: :+1:

tst/regression/test_suites/calculate_pi/calculate_pi.py
```diff
@@ -0,0 +1,94 @@
...
+        contents of the string are exactly what one would type on the command
+        line run running a parthenon driver.
+
+        As an example if the following block was uncommented it would overwrite
+        any of the parameters that were specified in the parthenon input file
+        parameters.driver_cmd_line_args = ['output1/file_type=vtk',
+                'output1/variable=cons',
+                'output1/dt=0.4',
+                'time/tlim=0.4',
+                'mesh/nx1=400']
```
Yurlungur, line 51: This is fantastic. The one edge case that occurs to me is if the input file wants some weird characters like newlines. But it should be manageable with python strings.

tst/regression/test_suites/calculate_pi/calculate_pi.py
```diff
@@ -0,0 +1,94 @@
...
+        """
+        return parameters
+
+    def Analyse(self,parameters):
```
Yurlungur, line 55: One use case which I think this supports which might be useful is generating plots in `Analyse`. There are certain tests where a plot can tell us what happened if the test failed.

I don't think you need to explicitly support this in the infrastructure. The test writer can just add some code to generate a plot and save it if they think that's necessary.

tst/regression/test_suites/calculate_pi/calculate_pi.py
JoshuaSBrown: That is a good point.

example/calculate_pi/pi_driver.cpp
JoshuaSBrown: This has been addressed.

pgrete:
Any reason not merging this? Authored by LANL CS and reviewed by LANL Physics and Athena Physics.
I'll pull the trigger tomorrow morning if there are no objections.

JoshuaSBrown:
> Any reason not merging this? Authored by LANL CS and reviewed by LANL Physics and Athena Physics.
> I'll pull the trigger tomorrow morning if there are no objections.

Jonas found an issue with how the mpi tests were being run, I need to add some options for specifying the number of OpenMP and MPI threads and processors.

JoshuaSBrown:
Pulling the trigger.

Issue #123, opened 2020-04-22
ContainerIterators produce cached, unique view of views if possible.


Yurlungur:
As discussed on the call. I'll implement this once the ContainerIterator @jdolence implemented solidifies.

Yurlungur:
Related to issue #90

nmsriram:
@yurlunger I'm trying to understand the need for this.  Generating the container is expected to be a pretty inexpensive operation, so would cacheing add complexity for not a lot of gain?

Yurlungur:
@nmsriram  this is something @forrestglines  brought up in the call. The concern is that there’s a cost to allocating the outer view of views which could add up on the gpu. So what needs to be cached isn’t so much the data structure itself as the memory for the outer view.

That said, I’m not actually convinced it will be that difficult to add a caving mechanism.

PR #125, opened 2020-04-23
View of Views VariablePack for iterating on Containers

jdolence:
## PR Summary

This PR introduces a new mechanism for creating lists of variables that can be passed to kernels as a single View of Views called a VariablePack.  Like the ContainerIterator this is meant to replace, a VariablePack can be produced by a simple call to PackVariables with a container and a vector of Metadata flags.

Addresses #105.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


jdolence:
one thing @nmsriram already pointed out offline is that we need to deal with the case where no variables end up populating the VariablePack.

src/interface/container_iterator.hpp
```diff
@@ -38,9 +128,8 @@ class ContainerIterator {
 
   /// initializes the iterator with a container and a flag to match
   /// @param c the container on which you want the iterator
```
nmsriram, line 130: Should we remove class ContainerIterator, the old iterator mechanism?

Yurlungur: I vote yes. But we should be careful to integrate the new mechanism everywhere where it's used in the code. Breaking downstream is probably fine. Breaking code in parthenon itself is probably not a good idea.

nmsriram: OK, it turns out we can't eliminate the class because it satisfies a need not filled by the VariablePack class.  If you look at src/interface/update.cpp, the code accesses the flux components of the variable, which are not provided by the VariablePack class.

jdolence: Agreed about the flux stuff.  I spent a little time earlier thinking about that.  No solution coded up just yet though.

nmsriram: VariablePack could also have a fluxPack member that is a pack of the fluxes in the same order.  Problem occurs for variables that don't have a flux array.

Yurlungur: If a variable doesn't have a flux array, it can be not added to the fluxpack? Or it can be added to the fluxpack as an empty view?

nmsriram: Probably an empty view since we want to keep the indexing same for fluxes and the variable.

Yurlungur: we might need logic for checking if a view is nonempty in relevant bits of code then. Not that that's an issue.

nmsriram: Maybe an assert for debugging but otherwise I don't think we do because the flux is only accessed by physics when there is one and the physics will be accessing variables by named indices.

forrestglines:
I'm still trying to get a performance test of the new structure working, after some segfaults took me much longer than they should have to debug. Right now I'm getting preliminary results where allocating the view of views every time is finishing faster than any other method  - I have no clue how that could be possible, since it's absolutely doing more work. I might have messed something up in my testing, so it's taking longer than expected to fix that.

forrestglines:
I finally got the testing issues sorted by moving to a Volta70 card instead of the Pascal that was more convenient. These results are far more in line with what I expected.

I compared the variable pack to a 4D kokkos view and kokkos view of 3D views  using a number of configurations. I used 9 variables in all cases, testing with 9 scalars, with 3 3-vectors, 6 scalars and 1 3-vectors (vectors in the middle), one 9-vector, and with all scalars but creating the variable pack each time the kernel is ran.

image

Creating the variable pack every timestep will incur a ~95% performance loss on small grids, so we definitely want to avoid that. I'm now seeing performance loss of 20% on small grids whenever a variable pack is used. For larger grids I see a 10% loss when I mix vectors and scalars in one variable pack. Using all vectors and all scalars performs about the same. Using the variable pack with just one vector with all the variables is the fastest, for some reason faster than the View of views for an n^3 grid.

So there will be noticeable but maybe acceptable losses for 16^3 grids. It's less an issue for 64^3 and larger. However, this test was over the full domain. I need to double check this with a buffer packing like test since overhead will be more apparent there.

nmsriram:
To the physics-type folks:
For the case wherewe are using dense-material-representation, does this mean that once a block is created, the array references do not change as long as the block is 'alive'?   If this is the case, would it be worth paying the penalty once for creation and then live with the ~5% loss for the rest of the cycles rather than copying the data into a 4D allocated space?

Yurlungur:
Thanks for this @forrestglines !

> I finally got the testing issues sorted by moving to a Volta70 card instead of the Pascal that was more convenient. These results are far more in line with what I expected.

Correct me if I'm wrong: is the Pascal not the more relevant architecture, since it's much newer? What this implies to me is that new architecture features on the Pascal reduce the price we pay from variable packs and a view of views. Perhaps the thing to here is do to between the two architectures. i.e., are we getting improved performance on the Pascal compared to the Volta, as we should be? And what is the fraction of peak for each architecture?

> I'm now seeing performance loss of 20% on small grids whenever a variable pack is used. 

20% is not trivial. But I think it's certainly good enough for now. There is some price to being a framework and not a hand-tuned, hand-written single-purpose code. I think this is a good place to start and we can performance tune later.

> I finally got the testing issues sorted by moving to a Volta70 card instead of the Pascal that was more convenient. These results are far more in line with what I expected.

If I recall, you warned us this would be bad. :) That's pretty dramatic, though. I think preconstructing the packs will not be difficult, though. I'll start to think about a clean implementation.

> For the case wherewe are using dense-material-representation, does this mean that once a block is created, the array references do not change as long as the block is 'alive'? If this is the case, would it be worth paying the penalty once for creation and then live with the ~5% loss for the rest of the cycles rather than copying the data into a 4D allocated space?

@nmsriram yes, I think this is worth doing. The one complication is you need one outer allocation space per block per collection of metadata, since the thing that's killing us is the allocation of the container iterator. But I think that's pretty do-able. We know what collections of metadata we will care aobut ahead of time.

forrestglines:
> Correct me if I'm wrong: is the Pascal not the more relevant architecture, since it's much newer? What this implies to me is that new architecture features on the Pascal reduce the price we pay from variable packs and a view of views. Perhaps the thing to here is do to between the two architectures. i.e., are we getting improved performance on the Pascal compared to the Volta, as we should be? And what is the fraction of peak for each architecture?

Pascal is the older architecture, there aren't that many of them, and the difference in performance is either due to the wonky memory layout on the Pascal chips or because I was testing on a consumer GPU.

Both Pascal and Maxwell GPUs put the L1 cache in the same hard memory as texture cache in order to get more shared memory per core. This makes it a lot easier to get close to peak L1 throughput since the bandwidth of L1 was awful - the performance ceiling is lower. I think it was a failed experiment, since the Turing (the newest) and Volta chips do the same thing as the old Kepler chips by putting L1 with shared memory. I don't think we should target Pascal performance.

Yurlungur:
> > Correct me if I'm wrong: is the Pascal not the more relevant architecture, since it's much newer? What this implies to me is that new architecture features on the Pascal reduce the price we pay from variable packs and a view of views. Perhaps the thing to here is do to between the two architectures. i.e., are we getting improved performance on the Pascal compared to the Volta, as we should be? And what is the fraction of peak for each architecture?
> 
> Pascal is the older architecture, there aren't that many of them, and the difference in performance is either due to the wonky memory layout on the Pascal chips or because I was testing on a consumer GPU.
> 
> Both Pascal and Maxwell GPUs put the L1 cache in the same hardware memory as texture cache in order to leave more shared memory per core. This makes it a lot easier to get close to peak L1 throughput since the bandwidth of L1 is awful - the performance ceiling is lower. I think it was a failed experiment, since the Turing (the newest) and Volta chips do the same thing as the old Kepler chips by putting L1 with shared memory. I don't think we should target Pascal performance.

Ah, I see. Thanks for the clarification.

nmsriram:
@forrestglines can you provide me with a pointer to your test code?  I know you've posted it before, but I seem to have lost the reference.  Thanks!

forrestglines:
@nmsriram 
Oops, here it is https://github.com/forrestglines/GPU-View-of-Views-Performance

The parthenon test is here: https://github.com/forrestglines/GPU-View-of-Views-Performance/blob/master/tests/parthenon_full_domain/parthenon_full_domain.cpp

jdolence:
@forrestglines any idea why the Kokkos view of 3D views and, for example, the VariablePack of scalars are different?  After all, isn't a VariablePack just a Kokkos view of 3D views?  The only additional thing it has is a dims array.  If that's leading to the drop in performance, it would be trivial to remove.

jdolence:
perhaps it has to do with the fact that the inner views in the VariablePack are actually sliced 6-D Views.  Something with how Kokkos infers to memory layout or something (e.g. strided)?

forrestglines:
> @forrestglines any idea why the Kokkos view of 3D views and, for example, the VariablePack of scalars are different? After all, isn't a VariablePack just a Kokkos view of 3D views? The only additional thing it has is a dims array. If that's leading to the drop in performance, it would be trivial to remove.

> perhaps it has to do with the fact that the inner views in the VariablePack are actually sliced 6-D Views. Something with how Kokkos infers the memory layout or something (e.g. strided)?

I'm not sure yet. The difference is only noticeable for small kernels, which makes me think that it's an additional overhead that doesn't scale with thread count. I'm not sure what it is yet. I could try testing against a Kokkos 4D slice of a 6D array or just a ParArrayND to verify your hypothesis.

forrestglines:
I was doing the same test with the z-buffer pack test. Since it's a smaller kernel, the performance loss is more apparent. I haven't checked ParArrayND yet

image

nmsriram:
@forrestglines What is the z-buffer algorithm?  I only know z-buffers from the graphics community and I'm pretty sure you don't mean that.

forrestglines:
@nmsriram The "z buffer pack" is the kernel used during boundary communications to pack one side of the z boundaries into a contiguous buffer for MPI to send off. It's the smallest realistic kernel that we used in K-Athena, although it's worth exploring whether we can combine all 6 boundary packing kernels into a single kernel.

forrestglines:
I'm not sure about the performance results. The timings are being inconsistent, although I'm sampling each case for several hundred milliseconds over tens of thousands of kernel runs for the lower grid sizes. I'm struggling with how to estimate a standard deviation without sampling individual kernel sizes, which would introduce too much host to device latency to be reliable.

nmsriram:
> I'm not sure about the performance results. The timings are being inconsistent, although I'm sampling each case for several hundred milliseconds over tens of thousands of kernel runs for the lower grid sizes. I'm struggling with how to estimate a standard deviation without sampling individual kernel sizes, which would introduce too much host to device latency to be reliable.

@forrestglines I had same problem, had to crank up number of iterations to:
```sh
export iter=(6000 5000 4500 4000 1500 1200)
export sizes=(4 8 16 32 64 128)
for idx in {0..5}; do OMP_PROC_BIND=spread OMP_PLACES=threads ./original 12 3 ${sizes[$idx]} ${iter[$idx]}; done;

```
The number of iterations for each size were chosen such that variation was less than 1% between runs for ~100 runs.  So that I can just run once and be happy.

I also wanted to see the effect of ghosts, so I modified your code slightly to do this.  Results attached. 
[timingsWithGhost.pdf](https://github.com/lanl/parthenon/files/4560992/timingsWithGhost.pdf)

nmsriram:
@forrestglines Results above are for the effect of ghost cells.  To do ghosts I modified Forrest's code to the one attached.  Also note you have to use the branch sriramGhost in parthenon with hash '8ac42ac42787024a804114eba96d2d2d9ea20a93'

If you uncomment the '#define DOGHOST', then the code will compute with ghosts.
[parthenon_full_domain.cpp.gz](https://github.com/lanl/parthenon/files/4561008/parthenon_full_domain.cpp.gz)

jdolence:
@nmsriram @Yurlungur @forrestglines @jlippuner just FYI, I moved the VariablePack stuff into its own header, which seems more sensible.

It seems everyone agrees we're going to do this, so perhaps we should get this reviewed/merged and continue the performance work elsewhere.  Certainly subsequent PRs can improve it.  Thoughts?

Yurlungur:
> @nmsriram @Yurlungur @forrestglines @jlippuner just FYI, I moved the VariablePack stuff into its own header, which seems more sensible.

👍 

> It seems everyone agrees we're going to do this, so perhaps we should get this reviewed/merged and continue the performance work elsewhere. Certainly subsequent PRs can improve it. Thoughts?

Think that's reasonable. The performance tuning and analysis looks "good enough" to me, with some painful work ahead to squeeze the last bit of performance out. That's work we need to do, but I don't object to doing that work with this in master. 

Once we have everything on GPU and working, we may discover bigger performance bottlenecks that need our attention first anyway.

example/advection/advection_package.cpp
```diff
@@ -106,15 +107,16 @@ void PreFill(Container<Real> &rc) {
...
-      }
-    }
-  }
+  PackIndexMap imap;
+  std::vector<std::string> vars({"advected", "one_minus_advected"});
+  auto v = PackVariables<>(rc, vars, imap);
+  const int in = imap["advected"].first;
+  const int out = imap["one_minus_advected"].first;
+  par_for(
+      "advection_package::PreFill", DevExecSpace(), ks, ke, js, je, is, ie,
```
JoshuaSBrown, line 116: If I can get [this](https://github.com/lanl/parthenon/pull/77) merged soon these function calls would benefit. It would be easy to insert indices in the wrong order.

pgrete:
Quick note having a high level look at the changes: I'd be great to move all the `par_for`s to `pmb->par_for`s (i.e., use the function call without explicit `DevExecSpace()` as the `MeshBlock` associated wrapper automatically uses the execution space assigned to the `MeshBlock`.
We'll need this to  make Cuda streams and tasking work efficiently.

jdolence:
oops.  good call @pgrete.  done.

nmsriram:
@forrestglines @pgrete @jdolence turns out that if you use raw Kokkos instead of the par_for with the variablePack, the performance is identical to that of the 3DViewofViews.  Given that, I'm going to propose that this pull request be **MERGED** into the head since variablePack doesn't adversely affect performance **AND** it makes the code more reasonable.  The performance characteristics of par_for should be the target of a different merge request.

What are your thoughts?

Here is code I used:
````C++
#ifdef KOKKOSCONTAINERS
    auto var_mix_in =
        PackVariables<Real>(container_mix_in, {Metadata::Independent});
    auto var_mix_out =
        PackVariables<Real>(container_mix_out, {Metadata::Independent});
    double time_mix = kernel_timer_wrapper(n_run, n_run, [&]() {
      Kokkos::parallel_for(
          policy, KOKKOS_LAMBDA(const int &idx) {
            const int v_var = idx / n_side3;
            const int k_grid = (idx - v_var * n_side3) / n_side2;
            const int j_grid =
                (idx - v_var * n_side3 - k_grid * n_side2) / n_side;
            const int i_grid =
                idx - v_var * n_side3 - k_grid * n_side2 - j_grid * n_side;

            auto view3d_in = var_mix_in(v_var);
            auto view3d_out = var_mix_out(v_var);
            view3d_out(k_grid, j_grid, i_grid) =
                2. * view3d_in(k_grid, j_grid, i_grid);
          });
    });
#else
    double time_mix = container_test_wrapper(n_run, n_run, container_mix_in,
                                             container_mix_out);
#endif
````

Here are results:
image

**Comment by pgrete** [comment]
These are some interesting and important performance findings.
We should soon be in a position to time the code in practice on GPUs (and compare the an N VariablePack versus an N-dim variable for all kernels including launch overhead).
I think Forrest already had a closer look at the code, which is why I won't do a detailed review (unless specifically requested) and focus on the other Kokkos PR.

File: docs/outputs.md [hunk 3]
```diff
@@ -0,0 +1,27 @@
+# Outputs
```
Yurlungur, line 1: Why is this file showing as new. Maybe merge master into your branch to resolve diffs? Presumably this is in master already.

example/advection/advection_package.cpp
```diff
@@ -108,15 +109,16 @@ void PreFill(Container<Real> &rc) {
...
-  for (int k = ks; k <= ke; k++) {
-    for (int j = js; j <= je; j++) {
-      for (int i = is; i <= ie; i++) {
-        qout(k, j, i) = 1.0 - qin(k, j, i);
-      }
-    }
-  }
+  PackIndexMap imap;
+  std::vector<std::string> vars({"advected", "one_minus_advected"});
+  auto v = PackVariables<>(rc, vars, imap);
```
Yurlungur, line 114: We should be caching this. I'll get on that.

tst/unit/test_container_iterator.cpp
```diff
@@ -67,72 +82,383 @@ TEST_CASE("Can pull variables from containers based on Metadata", "[ContainerIte
...
+  Kokkos::fence();
+  double perf_time = timer.seconds();
+
+  // FIXME?
+  // Validate results?
+
+  return perf_time;
+}
+
+TEST_CASE("Container Iterator Performance", "[ContainerIterator][performance]") {
```
Yurlungur, line 283: Should we incorporate some of @forrestglines separate performance tests directly into the unit tests here?

forrestglines:
> @forrestglines Results above are for the effect of ghost cells. To do ghosts I modified Forrest's code to the one attached. Also note you have to use the branch sriramGhost in parthenon with hash '8ac42ac42787024a804114eba96d2d2d9ea20a93'

Thanks @nmsriram! I'm checking the performance again today. The previous issues I had with performance timing turned out not to be due to under sampling but due to testing on an exclusive GPU but a share CPU, although it takes a while on the machine I'm testing on to get a full GPU node.

> @forrestglines @pgrete @jdolence turns out that if you use raw Kokkos instead of the par_for with the variablePack, the performance is identical to that of the 3DViewofViews.

`par_for` should inline to almost the same raw Kokkos code except this example doesn't add loop start indices and might use a different policy.  These small changes to the kernel might be enough to make the performance difference we're seeing.

Thanks @nmsriram for checking the performance of these! I also agree that this should be merged.

nmsriram:
@forrestglines could you provide me with the launch parameters you used to reduce variation?  I could very much use those.

Also, after you confirm my numbers, could you please also click 'approve'  or even just merge this PR?  I'm very concerned that the trunk will move and then merging gets harder.

Yurlungur:
@jdolence we still need to add documentation to this.

I took a stab at adding caching to this MR as discussed in the last call and in #123. The major changes are as follows:
- The `PackVariables` calls are now methods of `container`. This was required for the caching design, but it's also more natural. Iterable objects in the C++ stdlib own their own iterators.
- Each container now owns a cache for variable packs, implemented as a std::map from the variable names packed to packs. Note that order matters. i.e., if you ask for names in a different order, the outer view of the pack orders them differently and they are treated as distinct in the cache.
- I added a performance test for this capability into `test_container_iterator.cpp` and found it quite satisfactory. Now if you call `PackVariables` every cycle, it's only 2% slower than if you call it once.

I do not think this is the final iteration of this design---even with caching, the `PackVariables` method does some logic related to names, metadata flags, etc. We should try do this as little as possible and a larger rethink about expressing these dependencies at a higher level in the algorithm might be worthwhile. That said, I think for now this is good enough for us to use it in GPU code.

Yurlungur:
I added documentation.

example/advection/advection_package.cpp
```diff
@@ -108,15 +109,16 @@ void PreFill(Container<Real> &rc) {
...
-  CellVariable<Real> &qin = rc.Get("advected");
-  CellVariable<Real> &qout = rc.Get("one_minus_advected");
-  for (int k = ks; k <= ke; k++) {
-    for (int j = js; j <= je; j++) {
-      for (int i = is; i <= ie; i++) {
-        qout(k, j, i) = 1.0 - qin(k, j, i);
-      }
-    }
-  }
+  PackIndexMap imap;
```
forrestglines, line 112: Is a variable pack needed here if the variables aren't being iterated over? Does the original solution still work? Is this just for demonstration purposes?

Yurlungur: @jdolence wrote this bit. But my guess is it's not necessary and is just for demonstration purposes. You can of course use `ParArrayND.Get<3>()` and capture with the lambda.

jdolence: yes not necessary.  it's just meant to show that you can do things like, effectively, pass in a state vector and index into it as was originally done in athena++.

jlippuner:
I'm reviewing this now

src/interface/container.cpp
```diff
@@ -147,6 +143,221 @@ Container<T> Container<T>::SparseSlice(int id) {
...
+template <typename T>
+VariablePack<T> Container<T>::PackVariables() {
+  PackIndexMap vmap;
+  return PackVariables(vmap);
+}
+
+template <typename T>
+vpack_types::VarList<T> Container<T>::MakeList_(const std::vector<std::string> &names,
+                                                std::vector<std::string> &names_out,
+                                                const std::vector<int> sparse_ids) {
```
jlippuner, line 276: Please add some developer documentation that explains what this function does. Also, I'm not a fan of non-const references, would be better to use a pointer so that the caller understands that it can be modified.

tst/unit/test_container_iterator.cpp
forrestglines: I think we have enough testing here to catch an accidental significant change in performance. More rigorous performance testing might add too much overhead for the test suite.

src/interface/variable_pack.hpp
```diff
@@ -0,0 +1,234 @@
...
+#include "interface/variable.hpp"
+
+namespace parthenon {
+
+// some convenience aliases
+namespace vpack_types {
+template <typename T>
+using VarList = std::forward_list<std::shared_ptr<CellVariable<T>>>;
+using IndexPair = std::pair<int, int>;
+using StringPair = std::pair<std::vector<std::string>, std::vector<std::string>>;
```
jlippuner, line 34: I think these need some explanation. What is do the first and second index of the `IndexPair` represent? What is the purpose for the `StringPair`?

src/interface/variable_pack.hpp
```diff
@@ -0,0 +1,234 @@
...
+  }
+  KOKKOS_FORCEINLINE_FUNCTION
+  int GetDim(const int i) {
+    assert(i > 0 && i < 5);
+    return dims_[i - 1];
+  }
+
+ protected:
+  ViewOfParArrays<T> v_;
+  std::array<int, 4> dims_;
```
jlippuner, line 63: If this goes to the device, do we need to use a `Kokkos::array` (or whatever it is)?

tst/unit/test_container_iterator.cpp
```diff
@@ -67,72 +80,401 @@ TEST_CASE("Can pull variables from containers based on Metadata", "[ContainerIte
...
+    WHEN("we set individual fields by index") {
+      PackIndexMap vmap;
+      auto v = rc.PackVariables(std::vector<std::string>({"v3", "v6"}), vmap);
+      const int iv3lo = vmap["v3"].first;
+      const int iv3hi = vmap["v3"].second;
+      const int iv6 = vmap["v6"].first;
+      par_for(
+          "Initialize variables", DevExecSpace(), 0, v.GetDim(3) - 1, 0, v.GetDim(2) - 1,
+          0, v.GetDim(1) - 1, KOKKOS_LAMBDA(const int k, const int j, const int i) {
+            v(1, k, j, i) = 1.0; // should correspond with iv3lo+1
```
forrestglines, line 157: If variables are referenced by index/name within a kernel, it seems to me that a variable pack isn't needed or isn't the best way to write the loop. There are probably instances where it makes sense but I'm not seeing.

src/interface/variable_pack.hpp
```diff
@@ -0,0 +1,234 @@
...
+};
+
+// mapping to a tuple instead of using multiple maps reduces # of lookups
+// this wouldn't be super important if lookup time was constant,
+// but std::maps are trees, not hash tables and have an O(log(N)) lookup.
+// Unfortunately, std::pair doesn't work. So I have to roll my own.
+// I have no idea why std::pair doesn't work.
+// It appears to be an interaction between caused by a std::map<key,std::pair>
+// Possibly it's a compiler bug. gcc/7.4.0
+// ~JMM
```
jlippuner, line 95: What about `std::unordered_map`? That has O(1) lookup on average (worst case O(N)).

jlippuner:
> * Each container now owns a cache for variable packs, implemented as a std::map from the variable names packed to packs. Note that order matters. i.e., if you ask for names in a different order, the outer view of the pack orders them differently and they are treated as distinct in the cache.

Could we just use a `std::set` instead of `std::vector` for the names so that order doesn't matter?

tst/unit/test_container_iterator.cpp
Yurlungur: @jdolence can comment more. But this is something we encountered when writing our own code. It's a convenient way to capture a bunch of variables in a lambda at once. I am rewriting this test a little to not implicitly assume index ordering.

src/interface/variable_pack.hpp
nmsriram: Also would 'map' well to Kokkos::unordered_map, although Kokkos::unordered_map pushes the onus of finding the entry on the user.

docs/interface/containers.md
```diff
@@ -0,0 +1,48 @@
+# Containers, Container Iterators, and Variable Packs
+
+## Containers
+
+A *container* owns *variables.* Each variable is named, and the
+container knows about various types of variable, such as whether the
+variable is for cell-centered data, face-centered data, sparse data,
+or dense data. (For more details on anonymous variables, see
+[here](Metadata.md).) A variable can be added to a container as:
```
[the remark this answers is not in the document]
Yurlungur: Yep I'll add a note.

docs/interface/containers.md
```diff
@@ -0,0 +1,48 @@
...
+```C++
+parthenon::Container.Add(name, metadata, shape)
+```
+where the name is a string, the metadata is a `std::vector` of
+metadata flags, and the shape is a `std::vector` of integers
+specifying the dimensions.
```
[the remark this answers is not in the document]
Yurlungur: good idea. I will make a note.

docs/outputs.md
```diff
@@ -0,0 +1,27 @@
...
+Parthenon allows users to select which fields are captured in the HDF5 (```.phdf```) dumps at runtime.  In the input file, include a ```<parthenon/output*>``` block, list of variables, and specify ```file_type = hdf5```.  A ```dt``` parameter controls the frequency of outputs for simulations involving evolution. A ```<parthenon/output*>``` block might look like
+```
+<parthenon/output1>
+file_type = hdf5
+variables = density, velocity, & # comments are still ok
+            energy               # notice the & continuation character
+                                 # for multiline lists
+dt = 1.0
+```
+
```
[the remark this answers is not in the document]
Yurlungur: > Maybe add that this will produce an hdf5 (.phdf) output file every 1 units of simulation time containing the density, velocity, and energy of each cell.

Will do.

> So I can add multiple output blocks with type hdf5 and each would result in a different series of files?

Don't know. I didn't write this bit. @jdolence ?

src/interface/container.cpp
```diff
@@ -147,6 +143,221 @@ Container<T> Container<T>::SparseSlice(int id) {
...
 }
 
+/// Queries related to variable packs
+/// TODO(JMM): Make sure this is thread-safe
+/// TODO(JMM): Should the vector of names be sorted to enforce uniqueness?
+template <typename T>
+VariableFluxPack<T> Container<T>::PackVariablesAndFluxes_(
+    const std::vector<std::string> &var_names, const std::vector<std::string> &flx_names,
+    const vpack_types::VarList<T> &vars, const vpack_types::VarList<T> &fvars,
+    PackIndexMap &vmap) {
```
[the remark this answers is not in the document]
Yurlungur: I feel very strongly in the other direction. I would argue that a non-const reference is the right, modern model for a variable you intend to modify. I can add developer documentation to explain what's going on here, though.

src/interface/container.hpp
```diff
@@ -313,8 +335,29 @@ class Container {
...
+  vpack_types::VarList<T> MakeList_(const std::vector<std::string> &names,
+                                    std::vector<std::string> &names_out,
+                                    const std::vector<int> sparse_ids = {});
+  vpack_types::VarList<T> MakeList_(const std::vector<MetadataFlag> &flags,
+                                    std::vector<std::string> &labels);
+  vpack_types::VarList<T> MakeList_(std::vector<std::string> &names);
+
+  // These helper functions are private scope because they assume that
+  // the names include the components of sparse variables.
+  VariableFluxPack<T> PackVariablesAndFluxes_(const std::vector<std::string> &var_names,
```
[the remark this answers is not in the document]
Yurlungur: I'll rename them `*Helper_`

src/interface/container.cpp
```diff
@@ -147,6 +143,221 @@ Container<T> Container<T>::SparseSlice(int id) {
...
+template <typename T>
+VariableFluxPack<T>
+Container<T>::PackVariablesAndFluxes(const std::vector<std::string> &var_names,
+                                     const std::vector<std::string> &flx_names,
+                                     PackIndexMap &vmap) {
+  static std::vector<std::string> all_names;
+  static std::vector<std::string> all_flux_names;
+  vpack_types::VarList<T> vars = MakeList_(var_names, all_names);
+  vpack_types::VarList<T> fvars = MakeList_(flx_names, all_flux_names);
+  return PackVariablesAndFluxes_(all_names, all_flux_names, vars, fvars, vmap);
```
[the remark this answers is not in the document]
Yurlungur: I will add some documentation. That's right. The variables for the pack are used for the key of the cache, not the variables in the container. 

The issue is that the names of the variable is actually not sufficient, because a pack may contain only a subset of sparse indices for a sparse variable. So `all_names` contains the names/labels of all the variables included in the pack, but the sparse variable is expanded to `sparse_idx` where `sparse` is the name of the variable and `idx` is the index of the specific field in the sparse variable being extracted. Making it static is a performance enhancement to do with allocation. I wanted the vector size to persist between function calls. I'm not sure it makes a difference and it is a little confusing, so I'll remove the static.

src/interface/container.cpp
Yurlungur: I prefer non-const references over pointers. I think the fact that it's not const implies it can be modified.

I will add comments explaining what this does.

Yurlungur: The explanation, by the way is:
```C++
// From a given container, extract all variables and all fields in sparse variables
// into a single linked list of variables. The sparse fields are then named
// variable_index.
// The names of the non-sparse variables and the sparse fields are then
// packed into the std::vector "expanded_names," which is used as the key for
// the pack cache.
```

src/interface/variable_pack.hpp
Yurlungur: `IndexPair` is explained in the user documentation. This is what ends up in the `PackIndexMap`. These represent the inclusive  bounds of the indices that represent a variable in the variable pack.

`StringPair` is the key of the cache for fluxpacks, because you need the names of the variables and the names of the fluxes, and order matters.

src/interface/variable_pack.hpp
Yurlungur: `std::array` appears to work in cuda lambda captures. A C-style array does not, hence the `stdlib` use.

src/interface/variable_pack.hpp
Yurlungur: This is a good idea, so I gave it a shot. Unfortunately, the key, `std::vector<std::string>` is not a hashable type but is a comparable type. So `map` works but `unordered_map` would require defining a custom hashing function. I'm going to stick with `map` for now.

We can revisit if it's an issue.

Yurlungur:
@par-hermes format

Yurlungur:
@jdolence found the bug. So merging is good when everyone approves..

Yurlungur:
Correction. @jdolence says don't merge.

jdolence:
this should be good to go now, assuming the tests pass.  @jlippuner I think @Yurlungur addressed your comments, can you please approve or speak up now if you have remaining concerns?  This needs to get in, so I'll merge soon unless there's a good reason not to.

src/interface/container.cpp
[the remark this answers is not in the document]
Yurlungur: I renamed `all_names` to `expanded_names`, which is a bit more clear, since they expand the sparse variable names to include indices.

src/interface/variable_pack.hpp
Yurlungur: I added some additional explanatory comments inline.

src/interface/container.cpp
Yurlungur: Comments added inline. The basic idea of this is that it takes a set of names of variables and/or metadata and expands that into a linked list of pointers to the variables *and* pulls out the sparse indexes for the sparse variables, optionally only small subset of sparse indices.

The `expanded_names` variable is the names of each variable associated with the elements of the list. Importantly, it expands out the sparse variables to be `sparse_name_idx`, where `idx` is the sparse index of each element of the sparse variable.

src/interface/container.cpp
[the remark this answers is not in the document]
Yurlungur: Documentation added. It's something along these lines:

This function takes the names/fluxes and lists of shared pointers of variables and either extracts the pack from the cache or makes and caches the VariableFluxPack. The parameters are:

**INPUT:**
- `var_names` are the names of variables to be added to the pack. These are assumed to be already expanded for sparse variables
- `flx_names` the same as the var names but for fluxes.
- vars` the linked list of variables
- `fvars` the linked list of fluxes
- `vmap` maps variable names to indices in the pack.

Note that this is a helper function for just the inner call. Logic to do with deciding what variables to pack is done elsewhere.

PR #126, opened 2020-04-23
Add particle framework

brryan:
## PR Summary

Some scientific algorithms use particles rather than grid-based fields as their basic data structures. Adding particles to parthenon would expand its horizons.

This is very much a work in progress, I'm creating this PR now in case others have thoughts about how a particle framework should be laid down. 

Right now I have added a map of `SwarmContainer`s to each mesh block's `real_containers` `Container_Collection`, in analogy to the existing `Container`s. Each `SwarmContainer` holds a map of `Swarm`s. One `Swarm` is intended to represent all particles of a particular species; it holds pools of memory for position and other per-particle quantities that can be enrolled with their own metadata. These pools should be resizeable/defragmentable (not implemented yet). I *think* this is a reasonable path forward for Kokkos integration but I could easily be mistaken. 

This PR also adds a particles example in `example/particles`.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


pgrete:
tagging @jmstone as IIRC there are also particle implementations based on Athena++ (not in the public version) so there may be knowledge etc to be shared

jmstone:
Yes, there is a full particle implementation in Athena++.  A lot of work was put into making it perform well (about 3x faster than the previous module in the C-version of Athena).  This places lots of constraints on the design.  In particular, whenever particle properties have to be deposited on the mesh (mass, current, momentum), this must be vectorized and must avoid gather/scatter operations from the particle array.  This is a whole other subject...

example/particles/particles.cpp
```diff
@@ -0,0 +1,166 @@
...
+using parthenon::BlockStageNamesIntegratorTask;
+using parthenon::BlockStageNamesIntegratorTaskFunc;
+using parthenon::BlockTask;
+using parthenon::CellVariable;
+using parthenon::Integrator;
+using parthenon::Metadata;
+using parthenon::Params;
+using parthenon::ParArrayND;
+using parthenon::ParthenonManager;
+
```
JoshuaSBrown: Well when this https://github.com/lanl/parthenon/pull/110 gets merged this could mostly be reduced.

example/particles/particles.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+  /*for (int k = 0; k < ncells3; k++) {
+    for (int j = 0; j < ncells2; j++) {
+      for (int i = 0; i < ncells1; i++) {
+        Real rsq = std::pow(pcoord->x1v(i), 2) + std::pow(pcoord->x2v(j), 2);
+        q(k, j, i) = (rsq < 0.15 * 0.15 ? 1.0 : 0.0);
+      }
+    }
+  }*/
```
JoshuaSBrown: Is this supposed to be commented out? Or did you mean to delete it?

jmstone:
Hi Ben.  Have you looked at how the particles are implemented in Athena, or in Matt Kunz's Pegasus code (a hybrid-PIC algorithm built on Athena)?  Also I could give you access to the particle module that has been implemented in Athena++.  There are key design features like how you deposit particles properties onto the mesh in a way that vectorizes and minimizes memory accesses, how you generalize the particle pusher (for PIC it should be a Boris-type algorithm, for dust with drag it should be a SSP RK type method, etc), and how you handle particles that are in the ghost zones with MPI (i.e. whether you evolve them separately on each MeshBlock or pass them in messages).  There are some really tricky issues in the latter feature when you use high-order stencils that overlap ghost cells with AMR.  These designs have been implemented in the existing Athena particle modules, and it would be nice (even essential) to have them in Parthenon too.  I mention this in hopes of minimizing your having to re-write code.  I'd be happy to talk more about this off-line.  Cheers, Jim.

brryan:
Hi @jmstone, I just took a quick look at how particles are implemented in athena (I should have looked before but it didn't register with me that this part was public). I don't know that I have ever seen the Pegasus source code. My takeaway now is that athena allocates a contiguous pool of memory for particle structs, whereas what I've put down in this PR is each variable belonging to a particle (x, y, etc.) is a separate pool of memory. I have/had a suspicion that the latter approach might vectorize more easily but it is still very early days over here and I could simply be mistaken. Wouldn't be much trouble to switch what I've done here. It would be great to see the athena++ implementation, thanks for offering. Are the data structures similar?

Your point about different particle types needing different algorithms is well-taken. My vision has been to provide a few generic transport operations inside parthenon, and ensure that everything necessary is exposed such that physics packages building on parthenon, like a PIC code, can provide their own integrators for a transport task. That being said, some workhorses like the Boris push would probably make sense to include in Parthenon proper. Certainly I am happy to include anything essential. The vectorized deposition issues you bring up may be more subtle than transport in terms of letting a physics package handle them gracefully -- I have not thought about this in detail.

Next on my docket is a first cut of boundaries/MPI communication, about which I also have not had any sophisticated thoughts yet. It would be great to discuss this and the other issues you bring up in more detail offline, maybe this week if you have time, thanks for suggesting. I'm enthusiastic about the possibility of re-using code rather than going through the rigmarole of a bunch of false starts on my end.

brryan:
Thanks to @jdolence and @Yurlungur for help with kokkosification... the really simple transport operation here gets ~1e8 particle pushes/second on my laptop.

JoshuaSBrown:
@brryan what is the status of this?

brryan:
@JoshuaSBrown Now that the weak linking/library install stuff is winding down I'm going to be returning to this. Talking with @Yurlungur and @jdolence, it sounds like it may be a bit premature to commit to a particular approach for next step features like boundary communication. So for the near future I'll probably be developing for this mostly in a particles application I have on gitlab; approaches that work well can then be rapidly re-developed for parthenon. Is it OK if this remains [WIP] for a while?

example/particles/particles.cpp
brryan: Gone now, just a leftover

brryan:
This is in a pretty good place now, I think. Overall, the implementation is minimal, basically just a set of memory pools (one for each particle variable: x, y, etc.) that are managed (particles added, removed, list defragmented, masks provided) by the Swarm class. As you can see in the particles example, most of the compute code (transport, sampling) is implemented rather than linked to. There will probably be a way to move at least boundary conditions into parthenon/src, but as transport and interactions can be tightly linked, it may be difficult to provide too many universally useful particle functions (e.g. `BorisPush`) in parthenon/src. 

Right now this code ONLY works on one mesh block -- there is no particle communication between blocks. However, we're in a spot where we can use this framework elsewhere (as I'm already doing) so I think merging this in would be great assuming there aren't strong objections. 

Parallel dispatch operations I've run similar to what's in the particles example here show large speedups on GPUs relative to modern Xeons (~158x faster on a V100 than  Xeon Gold). No doubt problem dependent, and I want to confirm with others that my CPU builds are optimized, but I think we are in a good place here. One efficiency I think is pretty good is that for operations that are not parallelized right now (e.g. create a list of new slots in pool memory) and done on the host, the only host-device communication is ParArrayND<bool> mask arrays, which should be a ~minimal amount of data transfer. 

@jmstone I don't know that I got access to athena++/pegasus source code but I'm very interested in how they address the issues you describe above that are still uncertain here (particle deposition, communication). Right now I've only handled grid deposition through Kokkos atomic operations (e.g. `example/particles/particles.cpp:174`). Whether or not this is acceptable I haven't studied in detail. From the Kokkos documentation I understood atomic operations to be encouraged generally. There is a design change here relative to the Athena particles framework in that each particle variable is its own memory pool, rather than having one pool of particle structs in memory. I think fits in better with the parthenon model of enrolling fields at runtime -- to enroll a whole struct would I think be a compile-time affair. Anyway, I hope this overall path I'm going down would work for you. Happy to discuss, particularly if you have concerns.

brryan:
I forgot documentation/unit tests... I'll work on these

example/particles/particles.cpp
```diff
@@ -0,0 +1,362 @@
...
+
+  // Initialize random number generator pool
+  int rng_seed = pin->GetInteger("Particles", "rng_seed");
+  pkg->AddParam<>("rng_seed", rng_seed);
+  RNGPool rng_pool(rng_seed);
+  pkg->AddParam<>("rng_pool", rng_pool);
+
+  std::string swarm_name = "my particles";
+  Metadata swarm_metadata;
+  pkg->AddSwarm(swarm_name, swarm_metadata);
```
Yurlungur, line 65: Why is `AddSwarm` separate from `AddSwarmValue` is there a need to have multiple swarms?

example/particles/particles.cpp
```diff
@@ -0,0 +1,362 @@
...
+  Real y_max = pmb->coords.x2f(jb.e + 1);
+  Real z_max = pmb->coords.x3f(kb.e + 1);
+
+  ParArrayND<Real> t("time", max_active_index + 1);
+
+  // Simple particle push: push particles half a zone width until they have
+  // traveled one integrator timestep's worth of time
+  pmb->par_for(
+      "TransportSwarm", 0, max_active_index, KOKKOS_LAMBDA(const int n) {
+        if (mask(n)) {
```
Yurlungur: Hmm. One thing worth thinking about is the limit where `mask(n) = 0` for all but like one particle. The code is still the speed of the push for that one particle, probably. I don't think that's a problem, but it might be worth highlighting in documentation or a comment.

src/interface/container_collection.cpp
```diff
@@ -17,6 +17,33 @@
 
 namespace parthenon {
 
+template <typename T>
```
Yurlungur, line 20: This isn't your job  and it shouldn't go into this PR. But we have a pattern where we repeat the same code like this for each type of variable/swarm/etc. I wonder if there's a way to make it generic.

src/interface/container_collection.hpp
```diff
@@ -19,16 +19,21 @@
...
 
 namespace parthenon {
 
 template <typename T>
 class ContainerCollection {
  public:
   ContainerCollection() {
-    containers_["base"] = std::make_shared<Container<T>>(); // always add "base" container
+    // Always add "base" containers
+    containers_["base"] = std::make_shared<Container<T>>();
```
Yurlungur, line 31: Should we rename this `VarContainer` and SwarmContainer`. It seems weird that containers of variables are special in that they don't contain swarms.

Alternatively, could the swarm live in just a normal container, but we extend the container to own swarms, in the same way as it owns `FaceVariable`s?

src/interface/state_descriptor.hpp
```diff
@@ -64,6 +65,28 @@ class StateDescriptor {
   // retrieve label
   const std::string &label() { return label_; }
 
+  bool AddSwarm(const std::string &swarm_name, Metadata &m) {
+    if (swarmMetadataMap_.count(swarm_name) > 0) {
+      throw std::invalid_argument("Swarm " + swarm_name + " already exists!");
+    }
+    swarmMetadataMap_[swarm_name] = m;
```
Yurlungur, line 90: So it looks like swarms and swarmvalues can have different metadata. How does this work? Must the swarmvalue metadata contain the swarm metadata?

src/interface/swarm.cpp
```diff
@@ -0,0 +1,345 @@
...
+#include <vector>
+
+namespace parthenon {
+
+Swarm::Swarm(const std::string label, const Metadata &metadata, const int nmax_pool_in)
+    : label_(label), m_(metadata), nmax_pool_(nmax_pool_in),
+      mask_("mask", nmax_pool_, Metadata({Metadata::Boolean})),
+      marked_for_removal_("mfr", nmax_pool_, Metadata({Metadata::Boolean})),
+      mpiStatus(true) {
+  Add("x", Metadata({Metadata::Real}));
```
Yurlungur, line 31: Oh I see. Swarms and swarm values are different because the swarm values are not all real. And this also adds many of the underlying needed values automatically.

src/interface/swarm.cpp
```diff
@@ -0,0 +1,345 @@
...
+// Particles removed: nmax_active_index is new max active index
+void Swarm::RemoveMarkedParticles() {
+  int new_max_active_index = -1; // TODO BRR this is a magic number, needed for Defrag()
+
+  auto mask_h = mask_.data.GetHostMirrorAndCopy();
+  auto marked_for_removal_h = marked_for_removal_.data.GetHostMirror();
+  marked_for_removal_h.DeepCopy(marked_for_removal_.data);
+
+  // loop backwards to keep free_indices_ updated correctly
+  for (int n = max_active_index_; n >= 0; n--) {
```
Yurlungur, line 253: At some point, it might be worth thinking about how this can be done on the device.

src/interface/swarm_container.cpp
```diff
@@ -0,0 +1,104 @@
...
+///
+/// @param label the name of the variable
+/// @param metadata the metadata associated with the particle
+void SwarmContainer::Add(const std::string label, const Metadata &metadata) {
+  printf("Adding swarm to SwarmContainer!\n");
+  if (swarmMap_.find(label) != swarmMap_.end()) {
+    throw std::invalid_argument("swarm " + label + " already enrolled during Add()!");
+  }
+
+  printf("about to make shared!");
```
Yurlungur: Are these leftover debugging comments that should be removed?

src/interface/variable.cpp
```diff
@@ -158,6 +158,24 @@ std::string EdgeVariable<T>::info() {
...
+
+  // now append size
+  snprintf(tmp, sizeof(tmp), "%d", data.GetDim(1));
+  s += std::string(tmp);
+
+  // now append flag
+  s += " : " + this->metadata().MaskAsString();
+
+  return s;
+}
```
Yurlungur: This is very C-style of you... why not this?
```suggestion
std::string ParticleVariable<T>::info() {
  std::stringstream ss;

  // first add label
  std::string s = this->label();
  s.resize(20, '.');

  // combine
  ss << s << data.GetDim(1) << ":" << this->metadata().MaskAsString();

  return ss.str();
}
```

Yurlungur (approved):
Overall this looks great! I have a few comments below. But my thinking is that this is ready to go in and that my comments can be addressed later if needed.

The main sticking point for me is the difference between `Swarm` and `SwarmVariable`. It seems weird to me that these are distinct. Having now read through the code, I understand why: `Swarm` initializes some variables automatically and does all the underlying memory management, etc. And the metadata also appears different, with `SwarmVariables` only owning type metadata and nothing else. But I wonder if it would be cleaner to implicitly have one swarm. I.e., if you add a swarm variable, a swarm that owns it is automatically created under the hood, which would simplify the API. Then the SwarmVariables would own all the metadata, and the swarm would just be a manager for the machinery per meshblock. One advantage of this is that then you could add different metadata on a per swarm variable basis. I.e., maybe only some variables should be updated within a swarm. Or is there a reason that someone might want multiple swarms?  What do you think?

A missing piece, in addition to boundaries, is all the machinery for the swarm equivalent of variable and meshblock packs. But I think that can wait until you hit that need in application code... especially since things are currently in flux on the mesh side.

example/particles/particles.cpp
brryan: `AddSwarm` adds an entire swarm of particles, and `AddSwarmValue` adds a new variable (a new pool of memory to have one scalar for each particle) to the swarm named in the `swarm_name` argument. 

Multiple swarms would be needed if you had multiple different species of particles.

example/particles/particles.cpp
brryan: This is something that would be cleaned up by calling `Defrag` -- if `mask(n) = true` for only one particle, that particle will be moved to index 0 and `max_active_index` will be 0. I've added a note about `Defrag` to the documentation.

src/interface/container_collection.cpp
brryan: I know what you mean but I don't see immediately how to make this generic. Anyway, thank you for allowing me to ignore this on this PR :)

src/interface/container_collection.hpp
brryan: I agree with you that something should be done about this. I'm not sure what the best path forward right now is (in particular I don't want to mess up how (Var)Containers already work), but both your ideas sound good to me.

That being said I'm willing to just make this change now, with appropriate parthenon wizard guidance

example/particles/particles.cpp
Yurlungur: Oh I see. You're imagining a swarm per species, as opposed to species being an integer flag? Yeah that makes sense.

src/interface/container_collection.hpp
Yurlungur: I think my preference would be to move swarms inside container. But curious what others think. @nmsriram @jdolence @pgrete

src/interface/state_descriptor.hpp
brryan: I'm not doing much with metadata at the moment, though I assume it will be needed more later (or even now?) Right now the swarmvalue metadata just says whether its Real-valued or integer-valued. The Swarm metadata doesn't do anything.

src/interface/swarm.cpp
brryan: Yes that's right. I think it's fair to say that every particle needs an `x`, `y`, `z` (if you don't own a point in space then what are you), so I just made those default fields.

src/interface/swarm.cpp
brryan: Yes I agree. There may be some game to play with the outer/inner par_for loop model used elsewhere? Some sort of recursive parallel algorithm that evaluates/sorts/merges little linked lists together into one big one.

src/interface/swarm_container.cpp
brryan: Yes sorry, I forgot to clean up this file. Fixed.

src/interface/state_descriptor.hpp
Yurlungur: Huh. Yeah, so metadata matters when you want to generically operate over some subset of particles/swarm. For example, you can use metadata to:
- Mark what particles/variables need boundary communication
- Mark what particles/variables should be evolved, and what should be sampled independently per timestep

Stuff like that. I'm just throwing ideas out off the top of my head. Maybe it's worth revisiting later after we have some experience application-wise. Just worth keeping in mind as a capability and not coding ourselves out of using it.

src/interface/variable.cpp
brryan: Looks good to me -- I actually don't remember writing this, it's possible I copied this from an old version of `Variable`.

src/interface/swarm.cpp
Yurlungur: We could also investigate `Kokkos` native data structures that have parallel-safe properties for this kind of thing. Like atomic operations, maps, lists, etc

Yurlungur: Anyway, it's for the future. I wouldn't do it now.

Yurlungur:
We talked offline. @brryan has convinced me that swarms and swarm variables should be separate.

src/interface/state_descriptor.hpp
brryan: Yes I think what you say makes sense. But I think this should probably be deferred for a subsequent PR when these things are added.

Yurlungur: Yeah that's fine.

jmstone:
Hi Ben, looks great and I don't really have any comments on the details.  Regarding the comparison to the particle module in Athena++, the latter was written for performance (on cpus), and that dictated the design.  Thus things like the deposition of particle properties on the grid and the particle integrators have to vectorize.  I am not sure the issues are the same on GPUs and therefore whether the design can change.  The only way to know is to profile and compare to other implementations.  Nice job.

brryan:
@jmstone Thanks for taking a look at this. From the performance numbers I've been getting on a monte carlo radiation app I think it's reasonable to assume that my code is not vectorizing on CPUs. If it were possible to get vectorization that would be great. I wonder though whether radiation codes are generally going to have a harder time vectorizing than plasma codes. I'm less knowledgeable about PIC but I imagine there are fewer `if` statements for things like probabilistic scattering/absorption -- my naive guess right now is that that's what's breaking vectorization for me, at least in parts of my code.

The large CPU->GPU speedup numbers I'm getting do also suggest that GPUs don't take as much of a performance hit on algorithms like what I'm using here.

src/interface/container_collection.hpp
brryan: From discussions with @Yurlungur it sounds like the best path forward is to defer this change for later due to potential issues with mismatched integrator stages for particles and fields.

src/mesh/mesh.cpp
```diff
@@ -118,9 +118,9 @@ Mesh::Mesh(ParameterInput *pin, ApplicationInput *app_in, Properties_t &properti
   }
 
   // check number of grid cells in root level of mesh from input file.
-  if (mesh_size.nx1 < 4) {
+  if (mesh_size.nx1 < 1) {
```
brryan, line 124: @Yurlungur is it OK if I do this? I wanted to be able to run one-zone problems

brryan:
@JoshuaSBrown I think this is ready to go if you wouldn't mind taking another look and approving if you're happy with everything

src/mesh/mesh.cpp
Yurlungur: I'm not sure what the full implications of this are. I think it should be fine. Others may also know better. @jdolence @jmstone ?

src/interface/swarm.hpp
```diff
@@ -0,0 +1,150 @@
...
+
+  Swarm(const std::string label, const Metadata &metadata, const int nmax_pool_in = 3);
+
+  // TODO BRR This should really be const... mask_ is managed internally
+  /// Get mask array for active particles
+  ParticleVariable<bool> &GetMask() { return mask_; }
+
+  /// Get the mask array of active particles that are marked for removal
+  ParticleVariable<bool> &GetMarkedForRemoval() { return marked_for_removal_; }
+
```
JoshuaSBrown: On first look this rubs me the wrong way. Typically you don't want to return a non const reference from a class because it breaks encapsulation. I will need to look closer to see if your use case is ok but my preference would be that these would be:
```
const ParticleVariable<bool> & GetMask() const noexcept { return mask_; }
```

The same for the GetMarked method. You could then have a separate setter if you need to asctually change the mask_. 

The other option is to simply return a copy instead of a reference. 

```
ParticleVariable<bool> GetMask() const noexcept { return mask_; }
```

though this might not be an option if ParticleVariable is large.

src/interface/swarm.hpp
```diff
@@ -0,0 +1,150 @@
...
+  /// Make a new Swarm based on an existing one
+  std::shared_ptr<Swarm> AllocateCopy(const bool allocComms = false,
+                                      MeshBlock *pmb = nullptr);
+
+  /// Add variable to swarm
+  void Add(const std::string label, const Metadata &metadata);
+
```
JoshuaSBrown: make the string a const reference.

src/interface/swarm.hpp
```diff
@@ -0,0 +1,150 @@
...
+  /// Add multiple variables with common metadata to swarm
+  void Add(const std::vector<std::string> labelVector, const Metadata &metadata);
```
JoshuaSBrown: make labelVector a const reference.

src/interface/swarm.hpp
```diff
@@ -0,0 +1,150 @@
...
+
+  /// Remote a variable from swarm
+  void Remove(const std::string label);
```
JoshuaSBrown: make const reference

src/interface/swarm.hpp
```diff
@@ -0,0 +1,150 @@
...
+
+  /// Get real particle variable
+  ParticleVariable<Real> &GetReal(const std::string label) {
```
JoshuaSBrown: make label const reference.

JoshuaSBrown: Same comment here about returning ParticleVariable as a non const reference.

src/interface/swarm.hpp
```diff
@@ -0,0 +1,150 @@
...
+    return *(realMap_.at(label));
+  }
+
+  /// Get integer particle variable
+  ParticleVariable<int> &GetInteger(const std::string label) {
+    return *(intMap_.at(label));
+  }
+
+  /// Assign label for swarm
+  void setLabel(const std::string label) { label_ = label; }
```
JoshuaSBrown: Again I would make label a const reference in both GetInterger and setLabel

src/interface/swarm.hpp
```diff
@@ -0,0 +1,150 @@
...
+
+  /// retrieve label for swarm
+  std::string label() const { return label_; }
```
JoshuaSBrown, line 106: Add noexcept

src/interface/swarm.hpp
```diff
@@ -0,0 +1,150 @@
...
+
+  /// retrieve metadata for swarm
+  const Metadata metadata() const { return m_; }
```
JoshuaSBrown, line 109: Add noexcept, you can return this as reference.

src/interface/swarm.hpp
```diff
@@ -0,0 +1,150 @@
...
+
+  /// Assign info for swarm
+  void setInfo(const std::string info) { info_ = info; }
```
JoshuaSBrown: make info const reference..

src/interface/swarm.hpp
```diff
@@ -0,0 +1,150 @@
...
+
+  /// return information string
+  std::string info() { return info_; }
```
JoshuaSBrown: Make const noexcept

src/interface/swarm.hpp
```diff
@@ -0,0 +1,150 @@
...
+  void increasePoolMax() { setPoolMax(2 * nmax_pool_); }
+
+  /// Set max pool size
+  void setPoolMax(const int nmax_pool);
+
+  /// Check whether metadata bit is set
+  bool IsSet(const MetadataFlag bit) const { return m_.IsSet(bit); }
+
+  /// Get the last index of active particles
+  int get_max_active_index() { return max_active_index_; }
```
JoshuaSBrown: make `get_max_active_index` const noexcept

src/interface/swarm.hpp
```diff
@@ -0,0 +1,150 @@
...
+
+  /// Get number of active particles
+  int get_num_active() { return num_active_; }
```
JoshuaSBrown: Make const noexcept

src/interface/swarm.hpp
```diff
@@ -0,0 +1,150 @@
...
+
+  /// Get the quality of the data layout. 1 is perfectly organized, < 1
+  /// indicates gaps in the list.
+  Real get_packing_efficiency() { return num_active_ / (max_active_index_ + 1); }
```
JoshuaSBrown: Make const noexcept.

src/interface/swarm.hpp
```diff
@@ -0,0 +1,150 @@
...
+
+  bool mpiStatus;
+
```
JoshuaSBrown, line 137: I would prefer to encapsulate this with methods but won't make an argument in this case.

src/interface/swarm.hpp
```diff
@@ -0,0 +1,150 @@
...
+  /// Mark particle for removal (usually during a parallel dispatch)
+  KOKKOS_INLINE_FUNCTION
+  void MarkParticleForRemoval(int index) { marked_for_removal_(index) = true; }
```
JoshuaSBrown: make index const

src/interface/swarm.hpp
```diff
@@ -0,0 +1,150 @@
...
+
+  /// Remove particles marked for removal and update internal indexing
+  void RemoveMarkedParticles();
+
+  /// Open up memory for new empty particles, return a mask to these particles
+  ParArrayND<bool> AddEmptyParticles(int num_to_add);
```
JoshuaSBrown: make num to add const

src/interface/swarm.cpp
```diff
@@ -0,0 +1,347 @@
...
+#include <vector>
+
+#include "bvals/cc/bvals_cc.hpp"
+#include "globals.hpp" // my_rank
+#include "mesh/mesh.hpp"
+#include "swarm.hpp"
+
+namespace parthenon {
+
+Swarm::Swarm(const std::string label, const Metadata &metadata, const int nmax_pool_in)
```
JoshuaSBrown: make label const reference

src/interface/swarm.cpp
```diff
@@ -0,0 +1,347 @@
...
+      mask_("mask", nmax_pool_, Metadata({Metadata::Boolean})),
+      marked_for_removal_("mfr", nmax_pool_, Metadata({Metadata::Boolean})),
+      mpiStatus(true) {
+  Add("x", Metadata({Metadata::Real}));
+  Add("y", Metadata({Metadata::Real}));
+  Add("z", Metadata({Metadata::Real}));
+  num_active_ = 0;
+  max_active_index_ = 0;
+
+  auto mask_h = mask_.data.GetHostMirror();
```
JoshuaSBrown, line 37: I'm pretty sure `mask_h` can be made const reference.

src/interface/swarm.cpp
```diff
@@ -0,0 +1,347 @@
...
+    mask_h(n) = false;
+    marked_for_removal_h(n) = false;
+    free_indices_.push_back(n);
+  }
+
+  mask_.data.DeepCopy(mask_h);
+  marked_for_removal_.data.DeepCopy(marked_for_removal_h);
+}
+
+void Swarm::Add(const std::vector<std::string> labelArray, const Metadata &metadata) {
```
JoshuaSBrown: make `labelArray` const reference.

src/interface/swarm.cpp
```diff
@@ -0,0 +1,347 @@
...
+  return swarm;
+}
+
+///
+/// The internal routine for allocating a particle swarm.  This subroutine
+/// is topology aware and will allocate accordingly.
+///
+/// @param label the name of the variable
+/// @param metadata the metadata associated with the particle
+void Swarm::Add(const std::string label, const Metadata &metadata) {
```
JoshuaSBrown: make `label` const reference

src/interface/swarm.cpp
```diff
@@ -0,0 +1,347 @@
...
+    realVector_.push_back(var);
+    realMap_[label] = var;
+  } else {
+    throw std::invalid_argument("swarm variable " + label +
+                                " does not have a valid type during Add()");
+  }
+}
+
+void Swarm::Remove(const std::string label) {
+  int idx, isize;
```
JoshuaSBrown: `isize` is not used here.

src/interface/swarm.cpp
```diff
@@ -0,0 +1,347 @@
...
+  bool found = false;
+
+  // Find index of variable
+  idx = 0;
```
JoshuaSBrown: Get rid of idx above and simply use this line:

`int idx = 0;` 

That will reduce the number of lines of code.

src/interface/swarm.cpp
```diff
@@ -0,0 +1,347 @@
...
+  for (auto v : intVector_) {
```
JoshuaSBrown, line 97: make `v` const reference

src/interface/variable.hpp
```diff
@@ -204,6 +204,40 @@ class EdgeVariable {
...
+    return data(std::forward<Args>(args)...);
+  }
+
+  ///< retrieve metadata for variable
+  const Metadata metadata() const { return m_; }
+
+  bool IsSet(const MetadataFlag bit) const { return m_.IsSet(bit); }
+
+  ///< retrieve label for variable
+  std::string label() { return label_; }
```
JoshuaSBrown: Make const noexcept

src/interface/variable.hpp
```diff
@@ -204,6 +204,40 @@ class EdgeVariable {
   std::string label_;
 };
 
+template <typename T>
+class ParticleVariable {
+ public:
+  /// Initialize a particle variable
+  ParticleVariable(const std::string label, const int npool, const Metadata &metadata)
```
JoshuaSBrown: make label reference

src/interface/variable.hpp
```diff
@@ -204,6 +204,40 @@ class EdgeVariable {
...
+      : data(label, npool), npool_(npool), m_(metadata), label_(label) {}
+
+  // accessors
+  KOKKOS_FORCEINLINE_FUNCTION
+  ParArrayND<T> &Get() { return data; }
```
JoshuaSBrown, line 216: Make method noexcept. Again you are returning a non const reference, this will break encapsulation.

src/interface/variable.hpp
```diff
@@ -204,6 +204,40 @@ class EdgeVariable {
...
+  template <typename... Args>
+  KOKKOS_FORCEINLINE_FUNCTION T &operator()(Args... args) { // const {
+    return data(std::forward<Args>(args)...);
+  }
+
+  ///< retrieve metadata for variable
+  const Metadata metadata() const { return m_; }
```
JoshuaSBrown, line 223: You should be able to return this as a reference. Make the method noexcept.

src/interface/swarm_container.hpp
```diff
@@ -0,0 +1,167 @@
...
+
+  //-----------------
+  // Public Methods
+  //-----------------
+  // Constructor does nothing
+  SwarmContainer() {}
+
+  ///
+  /// Set the pointer to the mesh block for this container
+  void setBlock(MeshBlock *pmb) { pmy_block = pmb; }
```
JoshuaSBrown: This method makes no sense because the `pmy_block` is a public variable so you are not enforcing encapsulation. It would be safer to move the variable to private and use methods to access it (setter/getter). This set method can be declared noexcept as it is.

src/interface/swarm_container.hpp
```diff
@@ -0,0 +1,167 @@
...
+  /// Allocate and add a variable<T> to the container
+  ///
+  /// This function will eventually look at the metadata flags to
+  /// identify the size of the first dimension based on the
+  /// topological location.  Dimensions will be taken from the metadata.
+  ///
+  /// @param label the name of the variable
+  /// @param metadata the metadata associated with the variable
+  ///
+  void Add(const std::string label, const Metadata &metadata);
```
JoshuaSBrown: make `label` const reference

src/interface/swarm_container.hpp
```diff
@@ -0,0 +1,167 @@
...
+  /// Allocate and add a variable<T> to the container
+  ///
+  /// This function will eventually look at the metadata flags to
+  /// identify the size of the first dimension based on the
+  /// topological location.  Dimensions will be taken from the metadata.
+  ///
+  /// @param labelArray the array of names of variables
+  /// @param metadata the metadata associated with the variable
+  ///
+  void Add(const std::vector<std::string> labelArray, const Metadata &metadata);
```
JoshuaSBrown: make `labelArray` a const reference.

src/interface/swarm_container.hpp
```diff
@@ -0,0 +1,167 @@
...
+  void Add(std::shared_ptr<Swarm> swarm) {
+    swarmVector_.push_back(swarm);
+    swarmMap_[swarm->label()] = swarm;
+  }
+
+  ///
+  /// Get a swarm from the container
+  /// @param label the name of the swarm
+  /// @return the Swarm if found or throw exception
+  std::shared_ptr<Swarm> &Get(std::string label) {
```
JoshuaSBrown: Make `label` const reference.

src/interface/swarm_container.hpp
```diff
@@ -0,0 +1,167 @@
...
+    if (swarmMap_.count(label) == 0) {
+      throw std::invalid_argument(std::string("\n") + std::string(label) +
+                                  std::string(" swarm not found in Get()\n"));
+    }
+    return swarmMap_[label];
+  }
+
+  std::shared_ptr<Swarm> &Get(const int index) { return swarmVector_[index]; }
+
+  int Index(const std::string &label) {
```
JoshuaSBrown: Make method const.

src/interface/swarm_container.hpp
```diff
@@ -0,0 +1,167 @@
...
+  /// Gets an array of real variables from container.
+  /// @param names is the variables we want
+  /// @param indexCount a map of names to std::pair<index,count> for each name
+  /// @param sparse_ids if specified is list of sparse ids we are interested in.  Note
+  ///        that non-sparse variables specified are aliased in as is.
+  int GetSwarms(const std::vector<std::string> &names, std::vector<Swarm> &sRet,
+                std::map<std::string, std::pair<int, int>> &indexCount);
+
+  const SwarmVector &GetSwarmVector() const { return swarmVector_; }
+  const SwarmMap &GetSwarmMap() const { return swarmMap_; }
```
JoshuaSBrown, line 115: Make methods also noexcept.

src/interface/swarm_container.hpp
```diff
@@ -0,0 +1,167 @@
...
+  /// @param label the name of the variable to be deleted
+  void Remove(const std::string label);
+
+  // Temporary functions till we implement a *real* iterator
+
+  /// Print list of labels in container
+  void Print();
+
+  // return number of stored arrays
+  int Size() { return swarmVector_.size(); }
```
JoshuaSBrown: make method const noexcept

src/interface/swarm_container.hpp
Yurlungur: I think this is an artifact of what was going on in `Container` before I cleaned up the pointers. @brryan probably copied that machinery. @brryan take a look at what's going on in container in `develop` for the new preferred way of doing this.

src/interface/swarm_container.cpp
```diff
@@ -0,0 +1,99 @@
...
+bool SwarmContainer::ReceiveBoundaryBuffers() { return true; }
+
+void SwarmContainer::ReceiveAndSetBoundariesWithWait() {}
+
+void SwarmContainer::SetBoundaries() {}
+
+void SwarmContainer::StartReceiving(BoundaryCommSubset phase) {}
+
+void SwarmContainer::ClearBoundary(BoundaryCommSubset phase) {}
+
```
JoshuaSBrown, line 89: Are you planning on implementing these later? If you really want to leave them in they should probably be marked with  [[deprecated("message")]] in the header file. Where the message explains that they are not yet implemented.

src/interface/swarm_container.hpp
```diff
@@ -0,0 +1,167 @@
...
+    return -1;
+  }
+
+  /// Gets an array of real variables from container.
+  /// @param names is the variables we want
+  /// @param indexCount a map of names to std::pair<index,count> for each name
+  /// @param sparse_ids if specified is list of sparse ids we are interested in.  Note
+  ///        that non-sparse variables specified are aliased in as is.
+  int GetSwarms(const std::vector<std::string> &names, std::vector<Swarm> &sRet,
+                std::map<std::string, std::pair<int, int>> &indexCount);
```
JoshuaSBrown: This is missing a function body. It can probably be made const, or it should simply be removed if it is not going to be implemented or used.

src/interface/swarm_container.cpp
```diff
@@ -0,0 +1,99 @@
...
+#include "globals.hpp" // my_rank
+#include "mesh/mesh.hpp"
+#include "swarm_container.hpp"
+
+namespace parthenon {
+
+void SwarmContainer::Add(const std::vector<std::string> labelArray,
+                         const Metadata &metadata) {
+  // generate the vector and call Add
+  for (auto label : labelArray) {
```
JoshuaSBrown: `label` and `labelArray` should be const reference.

src/interface/swarm_container.cpp
```diff
@@ -0,0 +1,99 @@
...
+  swarmVector_.push_back(swarm);
+  swarmMap_[label] = swarm;
+}
+
+void SwarmContainer::Remove(const std::string label) {
+  int idx, isize;
+
+  // Find index of swarm
+  isize = swarmVector_.size();
+  idx = 0;
```
JoshuaSBrown: Remove forward declartion and consolidate:

```
int isize = swarmVector_.size();
int idx=0;
```

src/interface/swarm_container.cpp
```diff
@@ -0,0 +1,99 @@
...
+  for (auto s : swarmVector_) {
```
JoshuaSBrown: `s` can be made const reference.

src/interface/swarm_container.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+void SwarmContainer::ReceiveAndSetBoundariesWithWait() {}
+
+void SwarmContainer::SetBoundaries() {}
+
+void SwarmContainer::StartReceiving(BoundaryCommSubset phase) {}
+
+void SwarmContainer::ClearBoundary(BoundaryCommSubset phase) {}
+
+void SwarmContainer::Print() {
```
JoshuaSBrown: method can be made const.

src/interface/swarm_container.cpp
```diff
@@ -0,0 +1,99 @@
...
+  std::cout << "Swarms are:\n";
+  for (auto s : swarmMap_) {
```
JoshuaSBrown: make s const reference.

src/interface/swarm_container.hpp
```diff
@@ -0,0 +1,167 @@
...
+  bool operator==(const SwarmContainer &cmp) {
+    // Test that labels of swarms are the same
+    std::vector<std::string> my_keys;
+    std::vector<std::string> cmp_keys;
+    for (auto &s : swarmMap_) {
+      my_keys.push_back(s.first);
+    }
+    for (auto &s : cmp.GetSwarmMap()) {
+      cmp_keys.push_back(s.first);
+    }
```
JoshuaSBrown: Prefer preallocation with assignment over using push_back:

```
std::vector<std::string> my_keys(swarmMap_.size());
size_t i = 0;
for( auto s : swarmMap_ ){
 my_keys[i] = std::move(s);
 ++i;
}
```

src/interface/variable.cpp
```diff
@@ -159,6 +159,20 @@ std::string EdgeVariable<T>::info() {
   return s;
 }
 
+template <typename T>
+std::string ParticleVariable<T>::info() {
```
JoshuaSBrown: make method const

JoshuaSBrown (changes requested):
I have not finished reviewing, but I think I have left you with some things to work on. You have provided a test for the swarm class which is great, I would also like to see at leas a basic unit test on the swarm container and it would really nice to get a basic regression test for the particle example as well. 

Just some general comments for the future:

1. If you are dealing with any type larger than a primitive/basic type int,double such as string struct or class it is more performant to pass them in as const references (if you are not manipulating them) In your case for the most part it won't have a big impact but it is a good habit.

2. Where possible try to label your methods as const if they are not changing the internal state of a class. This makes the intent clearer and is in general safer. 

3. The noexcept keyword doesn't have a huge impact but there are sometimes performance gains the compiler can use if it knows a method will never throw an exception. If you are simply returning a type from a class get() { return var; } you can almost always label these as noexcept.

src/interface/container_collection.hpp
```diff
@@ -57,10 +82,16 @@ class ContainerCollection {
...
+    for (auto &sc : swarmContainers_) {
+      std::cout << "SwarmContainer " << sc.first << " has:" << std::endl;
+      sc.second->Print();
+      std::cout << std::endl;
+    }
   }
 
  private:
   std::map<std::string, std::shared_ptr<Container<T>>> containers_;
+  std::map<std::string, std::shared_ptr<SwarmContainer>> swarmContainers_;
```
JoshuaSBrown, line 94: Why are you using a map and not an unordered_map here? Do you need to preserve the ordering? unordered_map will be more efficient.

src/interface/state_descriptor.hpp
```diff
@@ -135,6 +177,8 @@ class StateDescriptor {
   const std::string label_;
   std::map<std::string, Metadata> metadataMap_;
   std::map<std::string, std::vector<Metadata>> sparseMetadataMap_;
+  std::map<std::string, Metadata> swarmMetadataMap_;
+  std::map<std::string, std::map<std::string, Metadata>> swarmValueMetadataMap_;
```
JoshuaSBrown, line 228: Again if you don't need ordering unordered_map will probably be a better choice here.

src/interface/swarm.hpp
```diff
@@ -0,0 +1,150 @@
...
+  MapToParticle<Real> realMap_;
+
+  std::list<int> free_indices_;
+  ParticleVariable<bool> mask_;
+  ParticleVariable<bool> marked_for_removal_;
+};
+
+using SP_Swarm = std::shared_ptr<Swarm>;
+using SwarmVector = std::vector<SP_Swarm>;
+using SwarmMap = std::map<std::string, SP_Swarm>;
```
JoshuaSBrown: Same comment about map vs unordered_map. I just want to know the justification?

example/particles/parthinput.particles
```diff
@@ -0,0 +1,59 @@
...
+
+<parthenon/meshblock>
+nx1 = 16
+nx2 = 16
+#nx3 = 1
+
+#<parthenon/output0>
+#file_type = hdf5
+#dt = 1.0
+#variables = particle_deposition
```
JoshuaSBrown: Did you mean to leave these lines commented out?

example/particles/particles.cpp
```diff
@@ -0,0 +1,361 @@
...
+
+  pkg->EstimateTimestep = EstimateTimestep;
+
+  return pkg;
+}
+
+AmrTag CheckRefinement(Container<Real> &rc) { return AmrTag::same; }
+
+Real EstimateTimestep(std::shared_ptr<Container<Real>> &rc) {
+  //auto pmb = rc->pmy_block;
```
JoshuaSBrown: Remove this.

example/particles/particles.cpp
```diff
@@ -0,0 +1,361 @@
...
+  pkg->AddField(field_name, m);
+
+  pkg->EstimateTimestep = EstimateTimestep;
+
+  return pkg;
+}
+
+AmrTag CheckRefinement(Container<Real> &rc) { return AmrTag::same; }
+
+Real EstimateTimestep(std::shared_ptr<Container<Real>> &rc) {
```
JoshuaSBrown, line 83: I would recommend that you read through section R.37 here https://www.modernescpp.com/index.php/c-core-guidelines-passing-smart-pointer. 

Instead of passing the shared pointer as a reference I believe it is better practice to dereference the Container and pass it in. Your function should look like this:

```
Real EstimateTimestep(Container<Real> &rc) {
```

I believe you only need to pass a shared pointer into a function if you are going to be sharing ownership of the object. Let me know if this makes sense it was a little confusing to me at first.

example/particles/particles.cpp
```diff
@@ -0,0 +1,361 @@
...
+  IndexRange kb = pmb->cellbounds.GetBoundsK(IndexDomain::interior);
+  int nx_i = pmb->cellbounds.ncellsi(IndexDomain::interior);
+  int nx_j = pmb->cellbounds.ncellsj(IndexDomain::interior);
+  int nx_k = pmb->cellbounds.ncellsk(IndexDomain::interior);
+  Real dx_i = pmb->coords.dx1f(pmb->cellbounds.is(IndexDomain::interior));
+  Real dx_j = pmb->coords.dx2f(pmb->cellbounds.js(IndexDomain::interior));
+  Real dx_k = pmb->coords.dx3f(pmb->cellbounds.ks(IndexDomain::interior));
+  Real minx_i = pmb->coords.x1f(ib.s);
+  Real minx_j = pmb->coords.x2f(jb.s);
+  Real minx_k = pmb->coords.x3f(kb.s);
```
JoshuaSBrown: I think you can make most of these const references.

example/particles/particles.cpp
```diff
@@ -0,0 +1,361 @@
...
+  IndexRange ib = pmb->cellbounds.GetBoundsI(IndexDomain::interior);
+  IndexRange jb = pmb->cellbounds.GetBoundsJ(IndexDomain::interior);
+  IndexRange kb = pmb->cellbounds.GetBoundsK(IndexDomain::interior);
+
+  Real x_min = pmb->coords.x1f(ib.s);
+  Real y_min = pmb->coords.x2f(jb.s);
+  Real z_min = pmb->coords.x3f(kb.s);
+  Real x_max = pmb->coords.x1f(ib.e + 1);
+  Real y_max = pmb->coords.x2f(jb.e + 1);
+  Real z_max = pmb->coords.x3f(kb.e + 1);
```
JoshuaSBrown: Again, I think you can make most of these const or const references.

example/particles/particles.cpp
```diff
@@ -0,0 +1,361 @@
...
+
+          rng_pool.free_state(rng_gen);
+        }
+      });
+
+  return TaskStatus::complete;
+}
+
+TaskStatus TransportSwarm(MeshBlock *pmb, int stage, std::vector<std::string> &stage_name,
+                          Integrator *integrator) {
```
JoshuaSBrown: Are you using `stage_name` and `stage` anywhere in this function?

example/particles/particles.cpp
```diff
@@ -0,0 +1,361 @@
...
+            }
+          }
+        }
+      });
+
+  return TaskStatus::complete;
+}
+
+TaskStatus Defrag(MeshBlock *pmb, int stage, std::vector<std::string> &stage_name,
+                  Integrator *integrator) {
```
JoshuaSBrown: I only see `pmb` being used here?

src/interface/variable.hpp
brryan: I'll make this `const` but I'm not comfortable using `noexcept` so especially for still-under-development code that isn't performance critical here and elsewhere I'd rather leave that off in this PR. Other `label()` functions in this file aren't `noexcept` either right now

src/interface/state_descriptor.hpp
```diff
@@ -112,16 +135,35 @@ class StateDescriptor {
     return names;
   }
 
+  // retrieve all swarm names
+  std::vector<std::string> Swarms() {
+    std::vector<std::string> names;
+    names.reserve(swarmMetadataMap_.size());
```
JoshuaSBrown, line 166: Nice

src/interface/variable.hpp
brryan: We want to modify `data` directly, `ParticleVariable` is just a container for this, right? `data` is a public variable here and in the other `*Variable` classes. It looks like the other classes don't have `Get()` methods, I think they might have when I originally wrote this

src/interface/swarm_container.hpp
brryan: Yes this was just the state of the code when I copied it over. I'll update.

src/interface/variable.hpp
brryan: Are we fully confident that that this being a reference is fine? I'm worried that could turn into a memory error down the line, and `Metadata` is very lightweight, right?

src/interface/variable.hpp
JoshuaSBrown: Ok, if that's the true, then this change goes beyond the scope of your pull request.

example/particles/particles.cpp
Yurlungur: Alternatively, he could pass in the raw pointer, no?

src/interface/variable.hpp
JoshuaSBrown: That's a fair point, use your better judgement. I might have been a bit over zealous in suggesting returning references.

src/interface/variable.hpp
JoshuaSBrown: That's fine, it is just a suggestion.

src/interface/variable.hpp
brryan: Roger that, thanks for bringing this up, it's a good thing to keep in mind.

example/particles/particles.cpp
JoshuaSBrown: True, I guess it's a style choice. This is from the google style guide, they recommend using pointers if you are going to be changing the object (it's non const), which is inline with what you are saying @Yurlungur :

> Reference Arguments
>
> All parameters passed by lvalue reference must be labeled const.
>
> In C, if a function needs to modify a variable, the parameter must use a pointer, eg int foo(int *pval). In C++, the function can alternatively declare a reference parameter: int foo(int &val).
>
>Defining a parameter as reference avoids ugly code like (*pval)++. Necessary for some applications like copy constructors. Makes it clear, unlike with pointers, that a null pointer is not a possible value.
>
> References can be confusing, as they have value syntax but pointer semantics.
> 
> Within function parameter lists all references must be const:
> 
> void Foo(const string &in, string *out);
> 
> In fact it is a very strong convention in Google code that input arguments are values or const references while output arguments are pointers. See the Pass by Value section for the cases where we allow pass-by-value. Input parameters may be const pointers, but we never allow non-const reference parameters except when required by convention, e.g., swap().
> 
> However, there are some instances where using const T* is preferable to const T& for input parameters. For example:
> 
>     You want to pass in a null pointer.
>     The function saves a pointer or reference to the input.
> 
> Remember that most of the time input parameters are going to be specified as const T&. Using const T* instead communicates to the reader that the input is somehow treated differently. So if you choose const T* rather than const T&, do so for a concrete reason; otherwise it will likely confuse readers by making them look for an explanation that doesn't exist.

Yurlungur: We've used raw pointers for this elsewhere in the code, so I don't think it's unreasonable to use those here. I agree, it comes down to stylistic preference.

src/interface/swarm_container.cpp
brryan: Yes these are what I'm working on next. I'll mark them not yet implemented for now.

src/interface/swarm_container.hpp
brryan: Removed

src/interface/swarm.hpp
brryan: Discussing with @Yurlungur keeping this as a reference maintains consistency with the grid infrastructure, and since ParticleVariable is essentially holding references to data this shouldn't have an effect on encapsulation I don't think

JoshuaSBrown: So does it need to be a non const reference to be consistent with the infrastructure and not for another reason. If that's the only reason we can create an issue so it is addressed separately. If there is a reason not to return it as const beyond that then no need to do anything, @Yurlungur ?

src/interface/swarm.hpp
brryan: So I strongly agree with you that these two should be `const` methods. `mask_` and `marked_for_removal_` are both managed internally and users should not be manually editing these variables. However, it looks like there are issues with just returning these as const. For example, `GetHostMirrorAndCopy` is not a `const` method so I can't do `GetMask().GetHostMirrorAndCopy()` which I need to do to use these variables at all. I think this should be addressed, but I'm not immediately sure how to. If I work through it in the next day or two I'll update this PR

JoshuaSBrown: If you are having to touch code outside the scope of this PR I would not worry about it, just create an issue and link here, otherwise this will never get merged. If it is a simple change then ya go ahead. Most of the issues I raised are simply bringing light to issues that should be addressed but it doesn't mean they need to be addressed by you if you did not introduce them. Thanks for looking into this.

jonahm-LANL:
It’s worth discussing in another issue. There’s some benefit to returning a reference performance wise. But variables are basically referenced anyway, because kokkos views have reference semantics.


Sent from my phone. Please excuse typos.

jonahm-LANL:
We can maybe add const versions of the kokkos copy methods if that’s meaningful thing to do. It might not be, because copying obviously implies non-const behavior.


Sent from my phone. Please excuse typos.

JoshuaSBrown:
>  It might not be, because copying obviously implies non-const behavior.

@Yurlungur I don't understand this point, can you elaborate a little bit more?

Yurlungur:
> > It might not be, because copying obviously implies non-const behavior.
> 
> @Yurlungur I don't understand this point, can you elaborate a little bit more?

Copying into the mask on device from host modifies it. That's all I mean. We can probably work around it and define a cleaner insterface that exposes only something constant to the user. I'm not sure.

example/particles/particles.cpp
brryan: Oh right I forgot I don't need all these as arguments with the `AddTask` magic, thanks for pointing this out

src/mesh/mesh.cpp
brryan: @Yurlungur @jdolence I'm going to go ahead and leave this in since it seems consistent with the following
```c++
if (mesh_size.nx2 < 1)
if (mesh_size.nx3 < 1)
```
unless I hear otherwise

jonahm-LANL:
Go for it


Sent from my phone. Please excuse typos.

src/interface/swarm.hpp
brryan: #315 created, thanks

src/interface/container_collection.hpp
brryan: I just copied `map` from its usage in `containers_`; if we want to change these I would change both but I don't want to mess with that in this PR. My guess now though is that there isn't much cost associated with these data structures.

src/interface/state_descriptor.hpp
brryan: same thing above, leaving this for symmetry

src/interface/swarm.hpp
brryan: Made this change so `Swarm` can serve as an inspiration to the continuum parts of parthenon :) 

(I agree, often when I use a `map`, `unordered_map` seems to be preferred, although how it hashes entries and what guarantees on efficiency there are is still unclear to me).

example/particles/parthinput.particles
brryan: Uncommented, thanks, producing this output is part of the example

example/particles/particles.cpp
brryan: Good catch, thanks (similar issue in next function)

example/particles/particles.cpp
brryan: This is a broader issue than just this example, right? (It also shows up in the advection example, for example, and in interface/update.hpp the declaration is `Real EstimateTimestep(std::shared_ptr<Container<Real>> &rc)`. Maybe there needs to be a significant review of how `shared_ptr`s are used in the code.

JoshuaSBrown: That's fine go ahead and create an issue. This can also be addressed later. This is not a breaking issue. Consistency is probably more important at this point.

brryan: See #316

brryan:
@JoshuaSBrown @Yurlungur OK I think this is ready to go. I made a copycat PR #317 since I don't think the CI will run here as the compare branch is on my fork of the parthenon repo.

example/particles/particles.cpp
```diff
@@ -104,15 +103,14 @@ TaskStatus SetTimestepTask(std::shared_ptr<Container<Real>> &rc) {
...
+TaskStatus DestroySomeParticles(MeshBlock *pmb) {
   auto pkg = pmb->packages["particles_package"];
   auto swarm = pmb->real_containers.GetSwarmContainer()->Get("my particles");
   auto rng_pool = pkg->Param<RNGPool>("rng_pool");
 
-  auto &mask = swarm->GetMask().Get();
-  auto &marked_for_removal = swarm->GetMarkedForRemoval().Get();
+  // The swarm mask is managed internally and should always be treated as constant. This
+  // may be enforced later.
+  const auto &mask = swarm->GetMask().Get();
```
JoshuaSBrown: Nice!

JoshuaSBrown:
This is not compiling for me on the power9 nodes. 

```
/home/joshuasbrown/Software/parthenon_power9/parthenon/tst/unit/test_swarm.cpp(70): error: class "parthenon::ParArrayNDGeneric<parthenon::device_view_t<parthenon::Real, parthenon::LayoutWrapper>>" has no member "data"

/home/joshuasbrown/Software/parthenon_power9/parthenon/tst/unit/test_swarm.cpp(71): error: class "parthenon::ParArrayNDGeneric<parthenon::device_view_t<int, parthenon::LayoutWrapper>>" has no member "data"

/home/joshuasbrown/Software/parthenon_power9/parthenon/tst/unit/test_swarm.cpp(109): error: class "parthenon::ParArrayNDGeneric<parthenon::device_view_t<parthenon::Real, parthenon::LayoutWrapper>>" has no member "data"

3 errors detected in the compilation of "/tmp/tmpxft_0000e201_00000000-6_test_swarm.cpp1.ii".
make[2]: *** [tst/unit/CMakeFiles/unit_tests.dir/test_swarm.cpp.o] Error 1
make[1]: *** [tst/unit/CMakeFiles/unit_tests.dir/all] Error 2
make: *** [all] Error 2
```

And it looks like the ci is running and failing.

JoshuaSBrown:
> basic unit test on the swarm container and it would really nice to get a basic regression test for the particle example

I would still like to see something along these lines.

brryan:
> > basic unit test on the swarm container and it would really nice to get a basic regression test for the particle example
> 
> I would still like to see something along these lines.

@JoshuaSBrown Sorry, I lost track of this comment you made earlier. Talking to @Yurlungur, my guess is `SwarmContainer` is going to go away relatively soon so I think it'd just be rearranging the deck chairs on the titanic to write a unit test for that now. If it stays in long-term I'll add a unit test.

I will add a particles example regression test now though, thanks for reminding me.

example/particles/particles.cpp
```diff
@@ -317,8 +310,9 @@ TaskStatus TransportSwarm(MeshBlock *pmb, int stage, std::vector<std::string> &s
   return TaskStatus::complete;
 }
 
-TaskStatus Defrag(MeshBlock *pmb, int stage, std::vector<std::string> &stage_name,
-                  Integrator *integrator) {
+// TaskStatus Defrag(MeshBlock *pmb, int stage, std::vector<std::string> &stage_name,
+//                  Integrator *integrator) {
+TaskStatus Defrag(MeshBlock *pmb) {
```
JoshuaSBrown: Did you mean to delete the comment?

src/interface/swarm.hpp
```diff
@@ -143,7 +155,7 @@ class Swarm {
 
 using SP_Swarm = std::shared_ptr<Swarm>;
 using SwarmVector = std::vector<SP_Swarm>;
-using SwarmMap = std::map<std::string, SP_Swarm>;
+using SwarmMap = std::unordered_map<std::string, SP_Swarm>;
```
JoshuaSBrown, line 172: You don't have to do this, but anytime you are not using a vector it is a good idea to leave a comment explaining why. At least that is what I was told.

JoshuaSBrown:
Ok, what you have looks good to me, when you have the example working I'll take a last look.

example/particles/particles.cpp
brryan: Yes thanks

src/interface/swarm.hpp
brryan: Noted, although this `map<string, Object>` pattern is regularly used in non-compute parts of parthenon so I'm gonna just leave this as is at least for now

brryan:
@JoshuaSBrown I agree that a regression test for the swarm example would be great but I'll just need some time to get that set up. Is it OK if this gets merged in now without the regression test? I think I addressed your other comments. I can add that test next in a separate PR. 

I just pushed a few cleanup changes, I'm assuming the CI will be OK but I'll keep an eye on it (the other PR that runs the CI more reliably is up to date with this one).

JoshuaSBrown:
> @JoshuaSBrown I agree that a regression test for the swarm example would be great but I'll just need some time to get that set up. Is it OK if this gets merged in now without the regression test? I think I addressed your other comments. I can add that test next in a separate PR.
> 
> I just pushed a few cleanup changes, I'm assuming the CI will be OK but I'll keep an eye on it (the other PR that runs the CI more reliably is up to date with this one).

When it passes the ci feel free to merge.

JoshuaSBrown (approved):

PR #127, opened 2020-04-23
Sriram/named iterator initialization

nmsriram:
Allows creation of a ContainerIterator using a vector of variable names instead of a set of metadata flags.  The code just uses std::find to find the variables by name and adds them to the vars member.

### DIFF
--- a/tst/unit/test_container_iterator.cpp
+++ b/tst/unit/test_container_iterator.cpp
@@ -41,6 +41,39 @@ using parthenon::par_for;
 using parthenon::ParArrayND;
 using parthenon::Real;
 
+
+static void setVector( const ParArrayND<Real> &v, const Real &value) {
+  par_for(
+	  "Initialize variables", DevSpace(), 0, v.GetDim(4) - 1, 0, v.GetDim(3) - 1, 0,
+	  v.GetDim(2) - 1, 0, v.GetDim(1) - 1,
+	  KOKKOS_LAMBDA(const int l, const int k, const int j, const int i) {
+	    v(l, k, j, i) = value;
+	  });
+}
+
+static Real sumVector( const ParArrayND<Real> &v ) {
+  using policy4D = Kokkos::MDRangePolicy<Kokkos::Rank<4>>;
+  Real sum = 0.0;
+  Kokkos::parallel_reduce(
+			  policy4D({0, 0, 0, 0},
+				   {v.GetDim(4), v.GetDim(3), v.GetDim(2), v.GetDim(1)}),
+			  KOKKOS_LAMBDA(const int l, const int k, const int j, const int i,
+					Real &vsum) { vsum += v(l, k, j, i); },
+			  sum);
+  return sum;
+}
+
+static int numElements( const ParArrayND<Real> &v ) {
+  using policy4D = Kokkos::MDRangePolicy<Kokkos::Rank<4>>;
+  int sum;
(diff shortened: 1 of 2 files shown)


tst/unit/test_container_iterator.cpp
```diff
@@ -41,6 +41,39 @@ using parthenon::par_for;
...
+      });
+}
+
+static Real sumVector(const ParArrayND<Real> &v) {
+  using policy4D = Kokkos::MDRangePolicy<Kokkos::Rank<4>>;
+  Real sum = 0.0;
+  Kokkos::parallel_reduce(
+      policy4D({0, 0, 0, 0}, {v.GetDim(4), v.GetDim(3), v.GetDim(2), v.GetDim(1)}),
+      KOKKOS_LAMBDA(const int l, const int k, const int j, const int i, Real &vsum) {
+        vsum += v(l, k, j, i);
```
AndrewGaspar, line 59: This is a no-no because you captured `ParArrayND` by reference, which means it's essentially a host-pointer. You need to either make a copy - something like `auto const v_ = v;`, or change the argument to the function to take `ParArrayND` by value rather than reference.

AndrewGaspar: Same as for `setVector`.

AndrewGaspar: Someday we'll have some clang tooling from the kokkos team that can catch this.

nmsriram: Of course.  I was just trying to reduce duplicated code and didn't think through running on disjoint memory spaces.

tst/unit/test_container_iterator.cpp
```diff
@@ -107,23 +109,21 @@ TEST_CASE("Can pull variables from containers based on Metadata", "[ContainerIte
       ContainerIterator<Real> ci(rc, {Metadata::Independent});
       CellVariableVector<Real> &civ = ci.vars;
       for (int n = 0; n < civ.size(); n++) {
-        ParArrayND<Real> v = civ[n]->data;
+        ParArrayND<Real> &v = civ[n]->data;
```
AndrewGaspar, line 112: Need to get rid of this '&'

tst/unit/test_container_iterator.cpp
```diff
@@ -107,23 +109,21 @@ TEST_CASE("Can pull variables from containers based on Metadata", "[ContainerIte
...
       THEN("they should sum appropriately") {
-        using policy4D = Kokkos::MDRangePolicy<Kokkos::Rank<4>>;
         Real total = 0.0;
-        ContainerIterator<Real> ci(rc, {Metadata::Independent});
-        CellVariableVector<Real> &civ = ci.vars;
         for (int n = 0; n < civ.size(); n++) {
-          Real sum = 1.0;
-          ParArrayND<Real> v = civ[n]->data;
+          using policy4D = Kokkos::MDRangePolicy<Kokkos::Rank<4>>;
+          ParArrayND<Real> &v = civ[n]->data;
```
AndrewGaspar, line 125: Need to get rid of this '&'.

tst/unit/test_container_iterator.cpp
AndrewGaspar: hm... wait, is my intuition right here? Now I'm questioning what happens with capturing refs. Aaaah!

AndrewGaspar: This is wrong - sorry for the noise.

PR #130, opened 2020-04-28
fix CMake GLOB expressions to not match files starting with .

jlippuner:
Don't format or check copyright in files starting with `.`

## PR Checklist

- [x] Code passes cpplint
- [x] New features are documented. N/A
- [x] Adds a test for any bugs fixed. Adds tests for new features. N/A
- [x] Code is formatted

### DIFF
--- a/cmake/CheckCopyright.cmake
+++ b/cmake/CheckCopyright.cmake
@@ -14,10 +14,10 @@
 # TODO: Probably shouldn't be done at configure time - instead we should have a separate
 # build task that does this.
 
-file(GLOB_RECURSE COPYRIGHTABLE_SOURCES src/*.cpp example/*.cpp)
-file(GLOB_RECURSE COPYRIGHTABLE_HEADERS src/*.hpp example/*.hpp)
-file(GLOB_RECURSE COPYRIGHTABLE_SCRIPTS scripts/python/*.py)
-file(GLOB_RECURSE COPYRIGHTABLE_CMAKE cmake/* src/CMakeLists.txt example/CMakeLists.txt)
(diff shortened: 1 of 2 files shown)


cmake/CheckCopyright.cmake
```diff
@@ -14,10 +14,10 @@
...
 # build task that does this.
 
-file(GLOB_RECURSE COPYRIGHTABLE_SOURCES src/*.cpp example/*.cpp)
-file(GLOB_RECURSE COPYRIGHTABLE_HEADERS src/*.hpp example/*.hpp)
-file(GLOB_RECURSE COPYRIGHTABLE_SCRIPTS scripts/python/*.py)
-file(GLOB_RECURSE COPYRIGHTABLE_CMAKE cmake/* src/CMakeLists.txt example/CMakeLists.txt)
+file(GLOB_RECURSE COPYRIGHTABLE_SOURCES src/[^\.]*.cpp example/[^\.]*.cpp)
+file(GLOB_RECURSE COPYRIGHTABLE_HEADERS src/[^\.]*.hpp example/[^\.]*.hpp)
+file(GLOB_RECURSE COPYRIGHTABLE_SCRIPTS scripts/python/[^\.]*.py)
+file(GLOB_RECURSE COPYRIGHTABLE_CMAKE cmake/[^\.]* src/CMakeLists.txt example/CMakeLists.txt)
```
nmsriram, line 20: The ".cpp" should actually be "\.cpp" and ditto for .hpp since the "." by itself will match any character.

nmsriram: Never mind, I stand corrected.  Glob isn't the same as regex.
Thanks @AndrewGaspar

PR #136, opened 2020-04-29
Kokkosize Parthenon

pgrete:
## PR Summary

First goal: Get the advection example working on GPUs on a static grid.
Very much WIP at this point.

Happy to get comments/other people involved (tagging @forrestglines and @jlippuner  given previous mail exchange on this).

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


src/bvals/cc/bvals_cc.cpp
```diff
@@ -159,7 +161,8 @@ int CellCenteredBoundaryVariable::LoadBoundaryBufferToCoarser(Real *buf,
   int p = 0;
   pmb->pmr->RestrictCellCenteredValues(var_cc, coarse_buf, nl_, nu_, si, ei, sj, ej, sk,
                                        ek);
-  BufferUtility::PackData(coarse_buf, buf, nl_, nu_, si, ei, sj, ej, sk, ek, p);
+  ParArray4D<Real> coarse_buf_ = coarse_buf.Get<4>(); // auto template deduction fails
```
jdolence, line 164: Do we really need to pull out the ParArray4D?  Can't we just pass the ParArrayND?  In fact, we may *want* something more than 4D.  What if the variable is a phase space density, for example.

jdolence: Obviously, the PackData function expects something 4D, so I get why you did this.  My question is really is this necessary?  I suppose we'll want to move all this over to using VariablePacks (or something like it) to combine this stuff into fewer/bigger kernels, so maybe it's not worth thinking to hard at the moment.

pgrete: This is related to our discussion during the last call.
I prefer to use strong types in the interfaces (also over assert statements) where the implementation expects a a certain type and would be happy to refactor again if the use case changes.
My hope is that this prevents future bugs as the compiler will shout and the functions are harder to use in an unintended way.

src/bvals/cc/bvals_cc.cpp
```diff
@@ -133,7 +133,9 @@ int CellCenteredBoundaryVariable::LoadBoundaryBufferSameLevel(Real *buf,
   sk = (nb.ni.ox3 > 0) ? (pmb->ke - NGHOST + 1) : pmb->ks;
   ek = (nb.ni.ox3 < 0) ? (pmb->ks + NGHOST - 1) : pmb->ke;
   int p = 0;
-  BufferUtility::PackData(var_cc, buf, nl_, nu_, si, ei, sj, ej, sk, ek, p);
+
+  ParArray4D<Real> var_cc_ = var_cc.Get<4>(); // automatic template deduction fails
```
Yurlungur, line 137: Hmm this should not fail. I can look into it if you want. What concerns me is that if you explicitly set it to `ParArray4D` that assumes a layout. On the other hand, as we discussed on the call, if we change layouts we're dealing with more problems anyway.

src/bvals/fc/bvals_fc.cpp
```diff
@@ -161,7 +161,7 @@ int FaceCenteredBoundaryVariable::ComputeFluxCorrectionBufferSize(
 //                                                               const NeighborBlock& nb)
 //  \brief Set face-centered boundary buffers for sending to a block on the same level
 
-int FaceCenteredBoundaryVariable::LoadBoundaryBufferSameLevel(Real *buf,
+int FaceCenteredBoundaryVariable::LoadBoundaryBufferSameLevel(ParArray1D<Real> &buf_tmp,
```
Yurlungur, line 164: This is sort of a more general question: Since views have reference semantics, when do we pass them by reference and when by value? I suppose that passing by reference is still best since the metadata isn't copied over.

src/bvals/bvals_interfaces.hpp
```diff
@@ -172,7 +172,7 @@ struct BoundaryData { // aggregate and POD (even when MPI_PARALLEL is defined)
   // currently, sflag[] is only used by Multgrid (send buffers are reused each stage in
   // red-black comm. pattern; need to check if they are available)
   BoundaryStatus flag[kMaxNeighbor], sflag[kMaxNeighbor];
-  Real *send[kMaxNeighbor], *recv[kMaxNeighbor];
+  ParArray1D<Real> send[56], recv[56];
```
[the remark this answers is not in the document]
pgrete: Good catch. Copy & paste error from K-Athena.

src/bvals/cc/bvals_cc.cpp
pgrete: ParArray4D also uses the Layout specified by the LayoutWrapper so it should be agnostic to the layout (or at least in itself consistent with the other ParArrays).
Feel free to see why the automatic deduction fails.
I assume that it's related to the `device_view_t` and `host_view_t` defined in `parthenon_arrays.hpp`.

Yurlungur: > ParArray4D also uses the Layout specified by the LayoutWrapper so it should be agnostic to the layout (or at least in itself consistent with the other ParArrays).

The problem is that if you switch the layout for `ParArrayND`, then `var_cc.Get<4>()` might return an array with `LayoutTile`

pgrete: I see.
I'm starting to wonder if it would be preferable to template the majority of methods so that we can have specialization on `ParArrayND`s and `ParArray#D`s.
As far as I can tell all use cases we have in mind don't require sparse slices etc. so we could optimize the methods and kernels for a fixed layout (!= tiled).

Yurlungur: I think templating on array type is reasonable. Although it does mean you are less explicit about dimension. I don't think it's a huge deal either way at the moment, so long as we are cognizant of what we're doing.

Yurlungur: > Feel free to see why the automatic deduction fails.

I think I see what's going on.

 The type returned by `ParArrayND.Get<4>()` is:
```C++
 Kokkos::View<double****, Kokkos::LayoutRight, Kokkos::Device<Kokkos::Serial, Kokkos::HostSpace>, Kokkos::MemoryTraits<0> >
```
But `ParArray4D` is 
```C++
Kokkos::View<double*****, Kokkos::LayoutRight, Kokkos::HostSpace>
```
Fortunately, the two types are compatible. So you can set one equal to the other and do implicit casts. i.e., the following works:
```C++
void doNothingByValue(ParArray4D<Real> array) {}
ParArrayND<Real> ndarray;
auto b = ndarray.Get<4>();
doNothingByValue(b);
```
But C++ won't do implicit casts on references to variables, so *this* fails:
```C++
void doNothingByReference(ParArray4D<Real>& array) {}
ParArrayND<Real> ndarray;
auto b = ndarray.Get<4>();
doNothingByValue(b);
```
I started an issue to discuss this in more detail, #143. I don't think it's a huge deal, though. And your explicit typecasting seems like a fine solution for now.

forrestglines: On `ParArray4D` vs `ParArrayND` - `var_cc` is a single variable now which is accessed and iterated over a 4D array. I prefer the hard typing so that the compiler handles errors. However, Changing the layout so that `ParArray4D` ends up being a `LayoutTile` will probably add overhead in the form of extra integer arithmetic and register usage. It shouldn't lead to strided access patterns since the iteration pattern will change with the layout.

Long term though, these boundary functions should merge variables in one kernel, so this variable will become a `VariablePack`

pgrete: Yes, those functions are very likely to see an update anyway as they are one of the performance issues in K-Athena.
I even expect we not only need merge Variables but actually fill all buffers in a single kernel simultaneously.

Yurlungur: Right. I wouldn't be too concerned about it now. As I describe in the issue, the hard typing should work, as should passing the view into functions by value instead of reference.

Yurlungur: The reason I dug in is that I was worried `ParArrayND` had a bug related to type inference, but I think it's all working as intended.

pgrete:
@par-hermes format

pgrete:
I'm happy to report that this is not WIP any more. I just finished writing the "developer docs" and happy to receive review comments.

Pinging @AndrewGaspar and @jdolence as this may now be the basis for Kokkosizing the AMR related calls.
So far I usually compiled in debug mode with UVM off, e.g., `cmake -DCMAKE_BUILD_TYPE="Debug" -DCMAKE_CXX_FLAGS="-O0 -g" -DKokkos_ARCH_EPYC=True -DKokkos_ARCH_TURING75=True -DKokkos_ENABLE_CUDA=True -DKokkos_ENABLE_OPENMP=True -DCMAKE_CXX_COMPILER=$(pwd)/../external/Kokkos/bin/nvcc_wrapper  -DKokkos_ENABLE_CUDA_UVM=OFF`  and then ran the advection example in the cuda debugger, e.g., `cuda-gdb ./example/advection/advection-example -ex "run -i ../example/advection/parthinput.advection"` which nicely told me the next function that required conversion.

PS: The second order hydro solver in AthenaPK based on this branch is now 11% faster than K-Athena using the same algorithm.

example/advection/advection_package.cpp
```diff
@@ -287,76 +294,110 @@ TaskStatus CalculateFluxes(Container<Real> &rc) {
...
+        for (int n = 0; n < nvar; n++) {
+          if (vx > 0.0) {
+            parthenon::par_for_inner(member, ib.s, ib.e + 1, [&](const int i) {
+              x1flux(n, k, j, i) = ql(n, i) * vx;
+            });
+          } else {
+            parthenon::par_for_inner(member, ib.s, ib.e + 1, [&](const int i) {
+              x1flux(n, k, j, i) = qr(n, i) * vx;
+            });
+          }
```
Yurlungur, line 329: I understand this nested parallelism design is almost certainly required to get things to vectorize on the CPU. However, I wonder what the impacts of it are on the GPU. Do you have any insight, @forrestglines @pgrete ? 

It might be worth experimenting with disabling it for some architectures.

example/advection/redefine_parthenon_defaults.cpp
```diff
@@ -113,7 +115,8 @@ void Mesh::UserWorkAfterLoop(ParameterInput *pin, SimTime &tm) {
     IndexRange jb = pmb->cellbounds.GetBoundsJ(IndexDomain::interior);
     IndexRange kb = pmb->cellbounds.GetBoundsK(IndexDomain::interior);
 
-    // TODO(pgrete) needs to be a reduction when using parallel_for
+    // calculate error on host
+    auto q = rc.Get("advected").data.GetHostMirrorAndCopy();
```
Yurlungur, line 119: This is using the default pipeline, not the meshblock stream as you described in your README. Should I expose a meshblock version?

src/bvals/cc/bvals_cc.cpp
Yurlungur: Presumably at some point soon, we will need to take a second look at this code and use the variable packs that we developed?

src/reconstruct/dc_inline.hpp
```diff
@@ -0,0 +1,80 @@
+//========================================================================================
```
Yurlungur, line 1: Should we remove the old donor cell and PLM reconstructions and just use these?

src/bvals/cc/bvals_cc.cpp
pgrete: yes, depending on what the profiling will tell us I can also see that these function may look (completely) different when we start optimizing the code.

docs/development.md
```diff
@@ -0,0 +1,140 @@
...
+
+The following list contains a few overall design decision that are useful to keep in mind
+
+- Kokkos never hides copies, i.e., there are no hidden deep copies and all data transfer needs to be explicit.
+- Kernel launches (e.g., a parallel region) are non-blocking, i.e., the CPU continues to execute code following a kernel that is executed on a GPU.
+
+### Kokkos wrappers/abstractions
+
+- `par_for` wrappers use inclusive bounds, i.e., the loop will include the last index given
+- `AthenaArrayND` arrays by default allocate on the *device* using default precision configured
```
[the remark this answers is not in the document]
pgrete: yes, of course, copy pasted old docs ;)

docs/development.md
```diff
@@ -0,0 +1,140 @@
...
+- To create an array on the host with identical layout to the device array either use
+  - `auto arr_host = Kokkos::create_mirror(arr_dev);` to always create a new array even if the device is associated with the host (e.g., OpenMP) or
+  - `auto arr_host = Kokkos::create_mirror_view(arr_dev);` to create an array on the host if the HostSpace != DeviceSpace or get another reference to arr_dev through arr_host if HostSpace == DeviceSpace
+- `par_for` and `Kokkos::deep_copy` by default use the standard stream (on Cuda devices) and are discouraged from use. Use `mb->par_for` and `mb->deep_copy` instead where `mb` is a `MeshBlock` (explanation: each `MeshBlock` has an `ExecutionSpace`, which may be changed at runtime, e.g., to a different stream, and the wrapper within a `MeshBlock` offer transparent access to the parallel region/copy where the `MeshBlock`'s `ExecutionSpace` is automatically used).
```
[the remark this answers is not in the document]
pgrete: Yes, I think in the mid/long run that feature would be useful/required.

docs/development.md
```diff
@@ -0,0 +1,140 @@
...
+### The need for reductions within function handling `MeshBlock` data
+
+A task (often a function associated with a `MeshBlock`) in the original Athena++ code was guaranteed to only being executed using a single CPU thread at any time.
+Thus, a reduction within a single `MeshBlock`(e.g., finding a minimum value in the `MeshBlock` data or filling a buffer using an incremented `offset` pointer) did not require special care.
+This is *not* true in Parthenon any more as a parallel region may be handled by many (GPU/...) threads in parallel.
+Therefore, a `parallel_reduce` needs to be used instead of a `parallel_for`.
+
+A strong hint where this is in order are places where a single variable is incremented within a kernel or where another reduction over MPI processes follows the preceding computations.
+
+Currently, Parthenon does not provide wrappers for parallel reductions so the raw Kokkos versions are used.
```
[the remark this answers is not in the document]
pgrete: At some point this would definitely be nice, but given the few places they are currently used I would not put it a the top of the priority list right now.

docs/development.md
```diff
@@ -0,0 +1,140 @@
...
+### FAQ
+
+- What's the difference between `GetDim` and `extent`?
+
+`ParArrayND` offer `GetDim` to access the underlying array dimension.
+Here, `GetDim(0)` refers to the "first" dimension (e.g., x-direction).
+`ParArray#D`s (with `#` being 1, 2, 3, ...) are direct typedefs to `Kokkos::View`s.
+Thus, a call to `extent(0)` returns the dimension along the first index.
+Given that `ParArray#D`s are constructed using reverse indices (note the `k,j,i` order in accessing elements), `extent` and `GetDim` using the same number usually have different meaning.
+
```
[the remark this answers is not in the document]
pgrete: I expected that this was where it came from and don't think that it's a problem (but we could discuss it during the next call).
Especially as the terminology is unique (dim vs extent) there's a clear separation. I just pointed it out because I stumbled across it (despite better knowledge).

example/advection/advection_package.cpp
pgrete: The main reason for nested parallelism is not necessarily vectorization (this also worked fine in K-Athena without it both on CPUs and GPUs) but the use of temporary/scratch pad arrays (the original scrach pad arrays as MeshBlock member variables cannot be used any more due to the different threading model when using GPUs).
Ideally, all architectures will benefit from this as it allows to reduce data movement and increased data locality.

example/advection/redefine_parthenon_defaults.cpp
pgrete: See comment above. At some point this may be useful, but currently there aren't too many places (except for init and finalization where performance doesn't matter anyway/that much)

src/reconstruct/dc_inline.hpp
pgrete: I'd keep them (in a Kokkosized version, which I already have in another repo so I can add those easily) in order to allow for more flexibility on the application side (if we only keep one then it'd force the application design to use nested  loops [or not] when employing those Parthenon functionality).

docs/development.md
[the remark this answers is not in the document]
Yurlungur: Fixed it.

docs/development.md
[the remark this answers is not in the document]
Yurlungur: I started an issue, #178 , and I'll dig into this soon.

docs/development.md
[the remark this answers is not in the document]
Yurlungur: I started an issue so we can keep track of this.

example/advection/advection_package.cpp
Yurlungur: Huh. I always thought it was because the code didn't vectorize without pre-fetching. 

Is it clear that explicitly creating the scratch arrays is actually more performant on GPUs? My naive concern would be that the requirement of synchronizing these arrays may actually hurt us on high-bandwidth architectures? I don't have any justification for that concern.

forrestglines: @Yurlungur On the GPUs with scratch level 0, scratch memory has a higher bandwidth than DRAM. Just going from the roofline models we explored with K-Athena, the code appeared to be limited by the bandwidth of DRAM memory. Merging some kernels to keep things in scratch memory and out of DRAM memory should let us cross the DRAM performance boundary.  We might also implement something like this: https://devblogs.nvidia.com/finite-difference-methods-cuda-cc-part-1/ which I've seen work very well in the past. 

Synchronization happens within a team on a single CUDA SM, which I haven't seen as a problem, but it might be on other architectures. The bigger issue I've seen in the past is that the register usage for the larger kernels can grow to the point that the GPU can't run as many concurrent threads and you loose performance. We'll have to rely on profiling to determine if that's the case.

I can't speak on how this will translate for AMD or Intel GPUs. My only bet is that the graphics pipelines are similar enough that the hardware will be comparable.

Yurlungur: Thanks for the explanation, @forrestglines ! 

I guess the take-away from this is that we need to do some profiling as we move forward. But it's interesting to hear that you already made some roofline plots for K-Athena and found the limit was the DRAM bandwidth.

pgrete:
Anyone volunteering for the second review on this PR?

src/kokkos_abstraction.hpp
```diff
@@ -450,6 +468,25 @@ inline void par_for(LoopPatternSimdFor, const std::string &name, DevExecSpace ex
   Kokkos::Profiling::popRegion();
 }
 
+// 1D  outer parallel loop using Kokkos Teams
+template <typename Function>
+inline void par_for_outer(OuterLoopPatternTeams, const std::string &name,
+                          DevExecSpace exec_space, size_t scratch_size_in_bytes,
+                          const int scratch_level, const int kl, const int ku,
```
JoshuaSBrown, line 475: I would recommend using the IndexRange struct for kl and ku, so as to avoid errors associated with the order.

src/bvals/fc/bvals_fc.hpp
```diff
@@ -72,12 +72,14 @@ class FaceCenteredBoundaryVariable : public BoundaryVariable {
...
-  void SetBoundaryFromCoarser(Real *buf, const NeighborBlock &nb) override;
-  void SetBoundaryFromFiner(Real *buf, const NeighborBlock &nb) override;
+  int LoadBoundaryBufferSameLevel(ParArray1D<Real> &buf,
+                                  const NeighborBlock &nb) override;
+  void SetBoundarySameLevel(ParArray1D<Real> &buf, const NeighborBlock &nb) override;
+  int LoadBoundaryBufferToCoarser(ParArray1D<Real> &buf,
+                                  const NeighborBlock &nb) override;
+  int LoadBoundaryBufferToFiner(ParArray1D<Real> &buf, const NeighborBlock &nb) override;
+  void SetBoundaryFromCoarser(ParArray1D<Real> &buf, const NeighborBlock &nb) override;
+  void SetBoundaryFromFiner(ParArray1D<Real> &buf, const NeighborBlock &nb) override;
```
JoshuaSBrown, line 82: I would recommend using `final` over `override` in this case. There are sometimes performance benefits:

https://devblogs.microsoft.com/cppblog/the-performance-benefits-of-final-classes/

src/bvals/cc/bvals_cc.hpp
```diff
@@ -69,14 +69,16 @@ class CellCenteredBoundaryVariable : public BoundaryVariable {
...
-  int LoadBoundaryBufferToCoarser(Real *buf, const NeighborBlock &nb) override;
-  int LoadBoundaryBufferToFiner(Real *buf, const NeighborBlock &nb) override;
+  int LoadBoundaryBufferToCoarser(ParArray1D<Real> &buf,
+                                  const NeighborBlock &nb) override;
+  int LoadBoundaryBufferToFiner(ParArray1D<Real> &buf, const NeighborBlock &nb) override;
 
-  void SetBoundaryFromCoarser(Real *buf, const NeighborBlock &nb) override;
-  void SetBoundaryFromFiner(Real *buf, const NeighborBlock &nb) override;
+  void SetBoundaryFromCoarser(ParArray1D<Real> &buf, const NeighborBlock &nb) override;
+  void SetBoundaryFromFiner(ParArray1D<Real> &buf, const NeighborBlock &nb) override;
```
JoshuaSBrown, line 82: Replace override with final.

src/utils/buffer_utils.cpp
```diff
@@ -20,83 +20,104 @@
...
-          buf[offset++] = src(n, k, j, i);
-      }
-    }
-  }
+void PackData(ParArray4D<T> &src, ParArray1D<T> &buf, int sn, int en, int si, int ei,
+              int sj, int ej, int sk, int ek, int &offset, MeshBlock *pmb) {
+  int ni = ei + 1 - si;
+  int nj = ej + 1 - sj;
+  int nk = ek + 1 - sk;
+  int nn = en + 1 - sn;
```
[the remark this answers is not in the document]
pgrete: Yes, will do.

src/reconstruct/plm_inline.hpp
```diff
@@ -0,0 +1,257 @@
...
+  // compute ql_(k+1/2) and qr_(k-1/2) using limited slopes
+  Real dxp = (coords.x3f(k + 1) - coords.x3v(k)) / coords.dx3f(k);
+  Real dxm = (coords.x3v(k) - coords.x3f(k)) / coords.dx3f(k);
+  for (int n = 0; n <= nu; ++n) {
+    parthenon::par_for_inner(member, il, iu, [&](const int i) {
+      ql(n, i) = qc(n, i) + dxp * dqm(n, i);
+      qr(n, i) = qc(n, i) - dxm * dqm(n, i);
+    });
+  }
+}
```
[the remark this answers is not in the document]
pgrete: You're right. These functions could be simplified.
As with the `IndexRange` I think it should be part of a separate effort and goes beyond the goal of this feature branch.
I created another issue to keep track of it #182

src/kokkos_abstraction.hpp
pgrete: Opening another issue as all wrappers would benefit from that #185

src/bvals/fc/bvals_fc.hpp
pgrete: Good point. I'll change that.

src/mesh/amr_loadbalance.cpp
```diff
@@ -488,7 +489,7 @@ void Mesh::RedistributeAndRefineMeshBlocks(ParameterInput *pin, int ntot) {
   MPI_Request *req_send, *req_recv;
   // Step 5. allocate and start receiving buffers
   if (nrecv != 0) {
-    recvbuf = new Real *[nrecv];
+    recvbuf = new ParArray1D<Real>[nrecv];
```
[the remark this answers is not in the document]
pgrete: Do you mean an array of `std::unique_ptr` to `ParArray1D` instead of an array of `ParArray1D` pointers?

pgrete:
> I think this is great! Thanks for all the work. The only thing I don't really like is the number of variables that are passed to the pack and unpack functions which I think could be reduced by using the IndexRange struct, however, I will leave it to you to decide if you want to change anything.

Thanks for the review.
I agree that overall the IndexRange would be cleaner here, but I expect that these function are going to be touched no matter what very soon (with respect to performance optimizations), so I think it makes sense to address this there.
I'll open an issue that we keep track of it (I also had an additional question/comment with respect to the IndexStruct that I'll raise in the issue) #183.

src/mesh/amr_loadbalance.cpp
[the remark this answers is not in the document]
JoshuaSBrown: I just want to avoid having to call new and delete. 
Something like this...
```
std::unique_ptr<ParArray1D<Real>> sendbuf, recvbuf;

recvbuf = std::make_unique<ParArray1D<Real>> ( ParArray1D<Real>[nrecv] );

```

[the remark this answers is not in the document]
jdolence: why use pointers at all, smart or otherwise?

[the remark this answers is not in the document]
pgrete: Are you thinking of a 2D View?
If yes, this would "waste" memory as the buffers are of different size.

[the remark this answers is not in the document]
jdolence: oh it doesn't matter.  just ignore me and let's get this merged in! 😃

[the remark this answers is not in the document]
pgrete: Alright, let's revisit this later. I'm going to pull the trigger now.

Issue #135, opened 2020-04-29
ParArray question


pgrete:
Are the Slices of the `ParArray` correct for slices in the first and second dim?
https://github.com/lanl/parthenon/blob/496d8ac94ccaf1bfdbf73739cc0684bffc38f448/src/parthenon_arrays.hpp#L283

The first `Kokkos::ALL()` should also be an SLC0, doesn't it?
Similarly, both `Kokkos::ALL()` for the 1D slice.

Yurlungur:
You are correct. Thanks for the catch! I will submit a MR for a fix.

Yurlungur:
Addressed in issue #137

Yurlungur:
Resolved in #137. Let me know if you find more problems, @pgrete !

PR #137, opened 2020-04-29
fix pararray slice bug and add tests for previously broken behaviour

Yurlungur:
## PR Summary

In issue #135 @pgrete found a bug in the implementation of slicing for `ParArrayND`. In particular, the slices into the given dimensions were wrong. Some `Kokkos::ALL()`s should have been `SLC0`s. 

I fixed the bug and added tests for those two slice use cases to ensure that the behavior is caught if it breaks again.

## PR Checklist

- [x] Code passes cpplint
- [x] New features are documented. (no new features)
- [x] Code is formatted

### DIFF
--- a/src/parthenon_arrays.hpp
+++ b/src/parthenon_arrays.hpp
@@ -280,11 +280,11 @@ class ParArrayNDGeneric {
   }
   KOKKOS_INLINE_FUNCTION
   auto SliceD(index_pair_t slc, std::integral_constant<int, 2>) const {
-    return Slice(SLC0, SLC0, SLC0, Kokkos::ALL(), slc, Kokkos::ALL());
+    return Slice(SLC0, SLC0, SLC0, SLC0, slc, Kokkos::ALL());
   }
   KOKKOS_INLINE_FUNCTION
   auto SliceD(index_pair_t slc, std::integral_constant<int, 1>) const {
-    return Slice(SLC0, SLC0, SLC0, Kokkos::ALL(), Kokkos::ALL(), slc);
+    return Slice(SLC0, SLC0, SLC0, SLC0, SLC0, slc);
   }
 #undef SLC0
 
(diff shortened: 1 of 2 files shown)


Yurlungur:
@par-hermes format

Yurlungur:
Does anyone know why the formatting check hasn't triggered? @AndrewGaspar ?

AndrewGaspar:
Unfortunately, GitHub Actions won't run in response to commits created by another GitHub Action. It's not the end of the world for this particular case.

However, I'm surprised it's still marking it as "Required" even though it won't run. Hm...

Yurlungur:
> Unfortunately, GitHub Actions won't run in response to commits created by another GitHub Action. It's not the end of the world for this particular case.

Ah I see!
 
> However, I'm surprised it's still marking it as "Required" even though it won't run. Hm...

Maybe the block is a review, not the format check? Let's see what happens once the MR has been reviewed?

Yurlungur:
> Maybe the block is a review, not the format check? Let's see what happens once the MR has been reviewed?

Oh I see. You approved and it's still blocking. Ok---I'll just push a dummy commit to trigger it.

AndrewGaspar:
Yeah, I was checking that. Ugh...

AndrewGaspar:
I don't know a good solution to this, unfortunately.

Yurlungur:
> I don't know a good solution to this, unfortunately.

Is there a way to trigger the check from the web interface? Well in any case, I don't think it's a huge deal. If even one reviewer requests changes, then the formatting check will get triggered anyway.

AndrewGaspar:
The workaround is to basically convince GitHub something changed. So you could do something like this, I think:
```
git tag tmp
git reset --hard HEAD^
git push --force
git merge tmp
git push
```

Unideal, but would at least trigger CI... I'll make a note of it in the docs.

Yurlungur:
Maybe we can add a little script that does that? a file named `scripts/fake-git-commit.sh`?

AndrewGaspar:
Yeah, that seemed to work: https://github.com/lanl/parthenon/runs/630689253

I'll add a script.

Yurlungur:
It's a small bug fix and it's had eyes on it. Pulling the trigger.

Issue #121, opened 2020-04-22
Namespace input blocks with `Parthenon`


Yurlungur:
As discussed in the call, we want to rename all parthenon specific input blocks with a prefix `Parthenon/`.

PR #142, opened 2020-04-30
Add parthenon to input block names

jdolence:
## PR Summary

Adds a parthenon "namespace" to blocks in the input file.  For example,
```
<mesh>
```
becomes
```
<parthenon/mesh>
```

Addresses #121, which was discussed on the phone call.

## PR Checklist

- [x] Code passes cpplint
- [x] New features are documented. (Docs updated anyway,.)
- [n/a] Adds a test for any bugs fixed. Adds tests for new features. (examples still work, which should be fine.
- [x] Code is formatted


jdolence:
@jmstone @Yurlungur hoping this is what we had in mind when this came up on the phone call.

jmstone:
Looks good to me.  Can we get rid of the "Graphics" input block?  I think all the information there should be in each of the <outputN> blocks.  In addition, I think the <time> block might have to be owned by the application program, since whether the concept of "time" or "nlim" even exists depends on the physics (drivers implemented by the application).

jdolence:
@jmstone, I've updated this, so I'd appreciate another quick look.  "graphics" block is gone.  Back to output blocks like
```
<parthenon/output1>
file_type = hdf5
variables = var1, var2, & # allow line continuations
                   var3
```
I am not using the Metadata::Graphics flag anymore, but rather stash the list of variables in the OutputParameters object similar to how it originally stashed a single string to indicate the variable.  Also updated the docs to reflect this change.

Re belonging to the application: The way this is handled now is that the Parthenon provided drivers (which an application can/should derive from) instantiate the Outputs object, providing the time related info if relevant.  So I believe we've already captured this concern, but let me know if there's something about this that you think might lead to problems.

jmstone:
Yes, looks good.  But I have some more design suggestions for output blocks we should talk about, for example I think we should specify what to output by using the container name, and then the variable in the container.  We should remove options like 'prim' and 'cons' since those refer to physics and are specific to hydro anyways.  Instead one could specify the container name (conserved or primitive in the case of hydro), and then "all" or "variable_name", etc.  I have some other suggestions as well...

nmsriram:
@jmstone Is this MR good to go?  [your comment above](https://github.com/lanl/parthenon/pull/142#issuecomment-622117531) seems to indicate that it is.  Could you please hit the approve button if you're happy with it.  OR if you want changes made before this can be merged, please note that as well.

nmsriram:
> Srry, didn't realize you actually needed my approval. Yes, I think this is good to go.

@jmstone, the approval wasn't required but since this is a shared user interface and you have indicated strong preference in the past about how you want this part of the code to function I did not want this change to slip by. you unnoticed.  Thanks for responding so fast!

PR #144, opened 2020-04-30
Dummy MeshBlock initializer for timing testing

nmsriram:
## PR Summary
I have created a new constructor for Parthenon::Meshblock(int n_side, int dim) so that we can initialize containers with ghost cells within them.  This just populates the ncells[123 and ncc[123] components of the mesh block.  

## PR Checklist
- [x] Code passes cpplint


Yurlungur:
Should this be targeting @jdolence's branch, not master?

nmsriram:
oops, this was based on the wrong branch.
Will put in a new one shortly.

PR #147, opened 2020-05-01
Added option for not building example drivers

JoshuaSBrown:
## PR Summary

This is to provide the option to build the example drivers. It is in response to [issue](https://github.com/lanl/parthenon/issues/101)

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


CMakeLists.txt
```diff
@@ -19,13 +19,28 @@ include(CTest)
...
+    message(FATAL_ERROR "BUILD_EXAMPLES is OFF but ENABLE_REGRESSION_TESTS is "
+    "set to ON! Regression tests cannot be enabled without also building example "
+    "drivers (BUILD_EXAMPLES must be set to ON). If you do not want to enable "
+    "the regression tests and you also do not want to build the example drivers, "
+    "BUILD_EXAMPLES and ENABLE_REGRESSION_TESTS must both be set to OFF. NOTE: "
+    "By default regression tests are enabled.")
+
+  endif()
+endif()
+
```
nmsriram, line 43: This effectively requires you to specify TWO variables to turn off ONE feature.  I'd like to be able to turn off build_examples and not have to specify anything else.  What is the reason regression test is dependent on examples?  This seems to not be right.

JoshuaSBrown: The regression tests are using the drivers from the examples. If the example drivers are not built you can't run them. If you hide this information, it is possible that you might want to stop building the examples to save some compile time without realizing you are no longer running the regression tests. I think it is good to be explicit. The other option is to simply print a message indicating that regression tests will not be run instead of throwing an error, but this could be accidentally missed.

JoshuaSBrown:
Here is a question do we need to start documenting build flags somewhere, if so where would be an appropriate place? https://github.com/lanl/parthenon/blob/master/docs/building.md

CMakeLists.txt
nmsriram: Then make both regression tests and examples optional.  If regression testing is on, turn on both.  If regression testing is off, only turn on examples if required.  Perhaps the thing to do is to split the examples between those that are 'optional' and those that will be built when regression testing is requested.

nmsriram:
That sure looks like the right place.  However, this discussion is more appropriate under an 'issue' rather than part of this MR.

Yurlungur:
I think the answer is yes and yes.

CMakeLists.txt
Yurlungur: I would rather be more explicit: if regression tests are built and examples are not, which should be allowed, `cmake` should throw an error and fail like it does for hdf5-parallel and openmpi.

nmsriram: Except that is an actually blocking dependency. This is an arbitrary dependency.  There is no reason that we can't create a target that only builds the examples for the regression test if build_excamples is not set.  It should not be an error.  The Venn diagram has an intersection area that should be built if either is specified.  The areas outside this intersection (almost symmetric difference, but not quite) should only be built if the respective flags are specified.

JoshuaSBrown: > Then make both regression tests and examples optional. If regression testing is on, turn on both. If regression testing is off, only turn on examples if required. 

Because the defaults for all tests are and examples are currently ON, you are suggesting switching the defaults to OFF? That way you only need to use one flag to either turn on the tests or examples. @AndrewGaspar do you have an opinion on this?

> Perhaps the thing to do is to split the examples between those that are 'optional' and those that will be built when regression testing is requested.

I honestly can't think of a reason why an example driver would not be tested within a regression test.

If my understanding is correct the example drivers are primarily what is to be tested in the regression tests. If there is a subset of regression tests that will not rely on the drivers that is another thing. I don't really see where that is a problem at this stage though.

nmsriram: Based on the above It seems to me that the examples directory needs to be renamed to regression and we need to get rid of the build_examples flag.

JoshuaSBrown: This is leading to a different discussion. https://github.com/lanl/parthenon/issues/148

Yurlungur: I think the regression tests and the examples should be decoupled, at least at the level of directory names... but I wouldn't object to there being only one flag---build examples---which controls whether both examples and regression tests are built.

JoshuaSBrown: Ok, this is where we stand keep examples and regression folders separate, but possibly have a single CMake variable to enable or disable both, unless someone else wants to chime in.

nmsriram: I'd argue about the 'same flag'.  I'd say separate folders _and_ separate flags.  The most likely scenarios are that a user of the library might want examples but not regression whereas developers would likely run regression tests but may not want the examples.

JoshuaSBrown: Ok, that will require a little CMake magic, the way the integration tests are implemented in CMake is not the most ideal for this, but it should be possible to do.

Yurlungur: Thanks for going through this @JoshuaSBrown! I am happy with that solution.

tst/regression/CMakeLists.txt
```diff
@@ -22,18 +22,35 @@
...
+    foreach(ind RANGE ${list_end})
+      list(GET arg ${ind} arg2)
+      if("${arg2}" STREQUAL "--driver")
+        MATH(EXPR ind "${ind}+1")
+        list(GET arg ${ind} driver)
+        get_filename_component(driver ${driver} NAME)
+        set_property(GLOBAL APPEND PROPERTY DRIVERS_USED_IN_TESTS "${driver}" )
+      endif()
+    endforeach()
+endfunction()
```
Yurlungur, line 41: As someone with very poor cmake skills, I think I would benefit from an explanatory comment that describes exactly what this is doing. But it's not a big deal either way.

JoshuaSBrown: Good point!

JoshuaSBrown:
@nmsriram are you happy with this?

PR #149, opened 2020-05-01
[WIP] Add build instructions for Darwin Power 9

jlippuner:
This is ready to go, but it will only work once #114 is fixed and merged, so I'm marking this WIP for now.

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/docs/building.md
+++ b/docs/building.md
@@ -23,3 +23,39 @@ ctest -LE performance
 # run performance tests
 ctest -L performance
 ```
+
+## LANL Darwin - Power 9
+
+The following works on the Darwin Power 9 nodes. We use Spack to install `hdf5` with the `xl`
+compiler. Setting up Spack is beyond the scop of this documentation.
+
+```bash
+# HDF5 is optional, but recommended
+spack install hdf5%xl
+
+# load required modules and set environment
+module purge
+
+module load cmake/3.17.0
+module load ibm/xlc-16.1.1.7-xlf-16.1.1.7
+module load openmpi/p9/4.0.3-xlc_16.1.1.7-xlf_16.1.1.7
+module load gcc/8.3.0
+module load clang/8.0.1
+module load cuda/10.2
+
+spack load hdf5%xl
+
+export NVCC_WRAPPER_DEFAULT_COMPILER=xlc++
+
+# clone parthenon with submodules
+git clone --recursive https://github.com/lanl/parthenon.git
+
+# make a build directory
+mkdir build
+cd build
+
+# configure and build (if hdf5 is not available, add -DDISABLE_HDF5=ON)
+CXX=`pwd`/../cmake/kokkos_nvcc_wrapper CC=xlc cmake -DKokkos_ENABLE_CUDA=ON -DKokkos_ENABLE_OPENMP=ON -DKokkos_ARCH_POWER9=ON -DKokkos_ARCH_VOLTA70=ON ..
+make -j
+make test
+```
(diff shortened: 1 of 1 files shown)


jdolence:
@jlippuner merge in master.  I think I broke things and then fixed them yesterday.  That was my bad.  Still trying to understand how it actually happened....

docs/building.md
```diff
@@ -23,3 +23,39 @@ ctest -LE performance
...
+```bash
+# HDF5 is optional, but recommended
+spack install hdf5%xl
+
+# load required modules and set environment
+module purge
+
+module load cmake/3.17.0
+module load ibm/xlc-16.1.1.7-xlf-16.1.1.7
+module load openmpi/p9/4.0.3-xlc_16.1.1.7-xlf_16.1.1.7
```
JoshuaSBrown, line 41: So I don't see this modules available on darwin?

JoshuaSBrown: Do these also need to be spack installed?

JoshuaSBrown: Disregard.

docs/building.md
```diff
@@ -23,3 +23,39 @@ ctest -LE performance
...
 ```
+
+## LANL Darwin - Power 9
+
+The following works on the Darwin Power 9 nodes. We use Spack to install `hdf5` with the `xl`
+compiler. Setting up Spack is beyond the scop of this documentation.
+
+```bash
+# HDF5 is optional, but recommended
+spack install hdf5%xl
```
JoshuaSBrown, line 34: I had trouble getting hdf5 installed I had to use:

```
spack install hdf5@1.10.6%xl@16.1 ^openmpi@4.0.3%xl@16.1 arch=rhel7-ppc64le
```

Yurlungur: My experience with HDF5 is that it is often easier to just build it by hand if I can't get it to behave with Spack. Spack is preferred of course. However, building HDF5 is easy and fairly foolproof.

docs/building.md
```diff
@@ -23,3 +23,78 @@ ctest -LE performance
...
+
+# make a build directory
+mkdir build
+cd build
+
+# configure and build (if hdf5 is not available, add -DDISABLE_HDF5=ON)
+cmake -DCMAKE_CXX_COMPILER=$PWD/../external/Kokkos/bin/nvcc_wrapper -DKokkos_ENABLE_CUDA=ON -DKokkos_ENABLE_OPENMP=ON -DKokkos_ARCH_POWER9=ON -DKokkos_ARCH_VOLTA70=ON ..
+make -j
+make test
+```
```
JoshuaSBrown, line 63: I get several compiler errors of the form:

```
driver_types.h:1037:5: error: ‘size_t’ does not name a type
     size_t  pitch;    /**< Pitch of allocated memory in bytes */
```

When trying to compile with these options.

JoshuaSBrown: Finally, got this to build, though it was with clang.

export NVCC_WRAPPER_DEFAULT_COMPILER=clang++

JoshuaSBrown: I also have added the cmake flag -DCMAKE_CXX_FLAGS="-Xcudafe --diag_suppress=esa_on_defaulted_function_ignored", there are a ton of warning that are output by the nvcc compiler. Looking over discussions elsewhere they appear to be false positives. They are of the form:

```
warning: __host__ annotation is ignored on a 
warning: __device__ annotation is ignored on a
```

JoshuaSBrown: Turns out there is also a problem with using cmake 3.17 earlier versions will work such as 3.11

AndrewGaspar:
Superseded by #335

PR #150, opened 2020-05-03
[WIP] Uniform Cartesian-like meshes

jdolence:
## PR Summary

This PR moves us to a much lighter weight Coordinates capability that assumes constant dx1, dx2, and dx3, and drops intrinsic support for non-Cartesian coordinates.  Non-Cartesian systems should still be supported at the application level.

Should wait for #125 since this was branched from there.

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted


jdolence:
@jmstone @pgrete @Yurlungur I went ahead and did this.  If we end up deciding we don't want to go this way, so be it -- only took a couple hours to get this working.

pgrete:
Thanks for taking a shot at this @jdolence 
This will definitely simplify the kernels.

With respect to reviewing this I think it's worth waiting for #125 to be merged so that it's clearer here what's new and needs review/testing/verification.

On the last point, I think it'd be worth adding a regression test that covers all aspects of the proposed changes.
I'm particularly concerned that we break sth in the AMR infrastructure that we don't notice until few weeks down the road.

With that in mind I had thought about these changes as a multistep approach where we keep the original Coordinate class and methods around in the first step and just add the constant dx interface.
Afterwards, I'd have set up associated tests and then refactor piece by piece (starting with static meshes and then AMR).

Do you think it's reasonable aiming for a test-driven multistep approach or do you think/prefer we can make this happen in one go?

jmstone:
To be honest after thinking about this and talking to Chris, I am a little nervous about adopting the mapped coordinates approach for supporting different coordinates.  We don't know if there are any issues at the poles (where \sqrt{g} is zero, so you can't just divide and multiple variables by this factor).  We worked hard to reduce non-uniform truncation error at the poles that results in features in the solutions in the approach we adopted.  I just don't know if the mapped coordinate approach works as well.  That being said, I do think coordinates should be supported at the application level.  So just providing the coordinates class and interface for a uniform dx at the moment would be OK, since this would provide a framework for building some more complicated at the application level down the road depending on what we decide.  I think this is in ageement with Philipp's suggestion.

jdolence:
@pgrete I certainly agree about waiting to review.  I threw this up as a WIP PR just so you guys were aware it was done.

I'm not sure how to respond to the rest of your comment.  After #125 is in and the changes are clarified, we can take a good look at the modifications.  That said, this works just fine on the advection example (which certainly makes good use of the AMR), which is the closest thing we currently have to something that covers all aspects of the proposed changes.  But, of course, it doesn't cover everything.

To your last question, of course it would be reasonable to take a test-driven multistep approach, but I don't think that's at all necessary for this particular effort at this stage of development.  Move fast and break things, as they say.

pgrete:
@jmstone , @forrestglines and I spoke about the PR yesterday and we're afraid that the current suggestions here may make it very hard to decide on a different approach later (especially once we refactor the AMR routines even more in optimizing for GPUs).
Thus, we strongly favor keeping all the interfaces intact so that it's clear which pieces need to be touched and in what way (as Jim mentioned, a lot of work in Athena++ went into making this abstraction for the coordinates system implemented in Athena++).

I suggest we talk about this tomorrow during the call (I'll put this on the agenda).

jdolence:
closing.  replaced by #155.

PR #152, opened 2020-05-04
Nested Parallel For

forrestglines:
## PR Summary

Adding nested parallel for loop abstractions `par_outer_loop` and `par_inner_loop` to abstract team level parallelisms and Kokkos scratch memory.

Adds the nested parallel for wrappers `par_outer_loop` and `par_inner_loop` that abstract elements of the Kokkos hierarchical parallelism for more convenient use on 3D and 4D grids. These wrappers allow management of cache memory which is necessary for efficient reconstruction kernels.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/docs/nested_par_for.md
+++ b/docs/nested_par_for.md
@@ -1,0 +1,30 @@
+
+# Nested Parallelism
+
+The loop wrappers documented here abstracts a hierarchical parallelism model,
+which allows more fine grain control on core level vs. vector level parallelism
+and allows explicitly defined caches for tightly nested loops. These wrappers
+provide a simplified interface of the [hierarchical parallelism in
+Kokkos](https://github.com/kokkos/kokkos/wiki/HierarchicalParallelism)
+
+## `par_outer_for`
+
+`par_outer_for` abstracts the team or multicore parallelism for outer loops.
+Inside the loop body, in the lambda function provided to `par_outer_for`,
+synchronization and memory sharing between the threads in the single team is
+possible through the `member_type` team member type from Kokkos.
+
+The bytes of scratch memory for cache needed by a single team is specified via
+the `scratch_size_in_bytes`, which needs be computed using `shmem_size` from
+the Kokkos API.
+
+The argument `scratch_level` defines where the scratch memory should be
+allocated. For CUDA GPUs, `scratch_level=0` allocates the cache in the faster
+by smaller `shared` memory and `scratch_level=1` allocates the cache in the
+slower but larger `global` or on device RAM memory. For CPUs, currently
+`scratch_level` makes no difference.
+
+
+## `par_inner_for`
+
+`par_inner_for` abstracts the vector level parallelism of compute units within a team.
--- a/docs/nested_par_for.md
+++ b/docs/nested_par_for.md
@@ -15,8 +15,7 @@ synchronization and memory sharing between the threads in the single team is
 possible through the `member_type` team member type from Kokkos.
 
 The bytes of scratch memory for cache needed by a single team is specified via
-the `scratch_size_in_bytes`, which needs be computed using `shmem_size` from
-the Kokkos API.
+the `scratch_size_in_bytes`, which needs be computed using `ScratchPadXD::shmem_size`.
 
 The argument `scratch_level` defines where the scratch memory should be
 allocated. For CUDA GPUs, `scratch_level=0` allocates the cache in the faster
@@ -28,3 +27,10 @@ slower but larger `global` or on device RAM memory. For CPUs, currently
 ## `par_inner_for`
 
 `par_inner_for` abstracts the vector level parallelism of compute units within a team.
+
+## `ScratchPadXD`
+
+Data type for memory in scratch pad/cache memory. Use `ScratchPadXD::shmem_size`, which is documented in [the 
+Kokkos documentation](https://github.com/kokkos/kokkos/wiki/HierarchicalParallelism) for determining scratch pad memory needs before kernel launch.
+
+
--- a/docs/nested_par_for.md
+++ b/docs/nested_par_for.md
@@ -43,6 +43,14 @@ Data type for memory in scratch pad/cache memory. Use
 documentation](https://github.com/kokkos/kokkos/wiki/HierarchicalParallelism)
 for determining scratch pad memory needs before kernel launch.
 
+## Important usage hint
+
+In order to ensure that individual threads of a team are synchronized always call
+`team_member.team_barrier();` after an `par_for_inner` if the following execution
+depends on the results of the `par_for_inner`.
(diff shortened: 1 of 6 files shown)


forrestglines:
@par-hermes format

forrestglines:
@pgrete This is ready for use in #136, once it's been reviewed.

docs/nested_par_for.md
```diff
@@ -0,0 +1,36 @@
+
+# Nested Parallelism
+
+The loop wrappers documented here abstracts a hierarchical parallelism model,
+which allows more fine grain control on core level vs. vector level parallelism
+and allows explicitly defined caches for tightly nested loops. These wrappers
+provide a simplified interface of the [hierarchical parallelism in
+Kokkos](https://github.com/kokkos/kokkos/wiki/HierarchicalParallelism)
```
Yurlungur, line 8: Can you also add a link to the unit test where you use the par fors, so a reader can see an example call?

src/CMakeLists.txt
```diff
@@ -41,6 +41,12 @@ if (${Kokkos_ENABLE_CUDA})
   set(PAR_LOOP_LAYOUT_VALUES "MANUAL1D_LOOP;MDRANGE_LOOP;TPTTR_LOOP;TPTTRTVR_LOOP"
     CACHE STRING "Possible loop layout options.")
 
+  set(PAR_INNER_LOOP_LAYOUT "TTR_INNER_LOOP" CACHE STRING
+    "Default loop layout for par_inner_for wrapper")
+
+  set(PAR_INNER_LOOP_LAYOUT_VALUES "TTR_INNER_LOOP"
+    CACHE STRING "Possible inner loop layout options.")
+
```
Yurlungur, line 49: Perhaps we should provide some documentation in the help file explaining why TTR and simdfor are the only options? I can see a new user getting confused by this.

tst/unit/kokkos_abstraction.cpp
```diff
@@ -262,6 +262,175 @@ TEST_CASE("par_for loops", "[wrapper]") {
...
+
+  // Copy host array content to device
+  Kokkos::deep_copy(dev_u, host_u);
+
+  // Compute the scratch memory needs
+  const int scratch_level = 0;
+  size_t scratch_size_in_bytes = parthenon::ScratchPad1D<Real>::shmem_size(N);
+
+  // Compute the 2nd order centered derivative in x
+  parthenon::par_outer_for(
```
Yurlungur, line 292: For some reason, I feel like `par_for_outer` is more natural than `par_outer_for`. But of course, that's just personal preference. The name doesn't need to change.

tst/unit/kokkos_abstraction.cpp
```diff
@@ -262,6 +262,175 @@ TEST_CASE("par_for loops", "[wrapper]") {
...
+      outer_loop_pattern, "unit test Nested 3D", exec_space, scratch_size_in_bytes,
+      scratch_level, 0, N - 1, 0, N - 1,
+
+      KOKKOS_LAMBDA(parthenon::member_type team_member, const int k, const int j) {
```
Yurlungur, line 296: `parthenon::member_type` is ambiguous outside of a kokkos context, and also kind of a mouthfull. How about
`parthenon::teammbr_t` or something along those lines? For sure we can abbreviate `type` as `t`.

tst/unit/kokkos_abstraction.cpp
```diff
@@ -262,6 +262,175 @@ TEST_CASE("par_for loops", "[wrapper]") {
...
+              dev_du(k, j, i - 1) = (scratch_u(i + 1) - scratch_u(i - 1)) / 2.;
+            });
+      });
+
+  // Copy array back from device to host
+  Kokkos::deep_copy(host_du, dev_du);
+
+  Real max_rel_err = -1;
+  const Real rel_tol = 1e-12; // (forrestglines) magic number for machine tolerance - is
+                              // there something better?
```
Yurlungur, line 317: Actual machine tolerance is reported in `std::numeric_limits`. See [here](https://en.cppreference.com/w/cpp/types/numeric_limits/epsilon).

tst/unit/kokkos_abstraction.cpp
```diff
@@ -262,6 +262,175 @@ TEST_CASE("par_for loops", "[wrapper]") {
...
+          host_u(n, k, j, i) = pow((i + 1) * (j + 1) * (k + 1) * (n + 1), 2.0);
+
+  // Copy host array content to device
+  Kokkos::deep_copy(dev_u, host_u);
+
+  // Compute the scratch memory needs
+  const int scratch_level = 0;
+  size_t scratch_size_in_bytes = parthenon::ScratchPad1D<Real>::shmem_size(N);
+
+  // Compute the 2nd order centered derivative in x
```
Yurlungur, line 360: I like this as a test.

forrestglines:
> Looks good. Although I'm curious, when do we need this? I take it that you wrote this up because you need it for kokkosizing... what loops require it?

Anywhere that a 1D/2D array is used in the original code as cache within an iteration along the x direction. The reconstruction routines do this for performance. A lot of the coordinates routines in Athena++ do this as well.

Yurlungur:
> Anywhere that a 1D/2D array is used in the original code as cache within an iteration along the x direction. The reconstruction routines do this for performance. A lot of the coordinates routines in Athena++ do this as well.

Aaah I see! Yes, I was wondering how to handle those. This is the obvious solution.

tst/unit/kokkos_abstraction.cpp
forrestglines: > For some reason, I feel like `par_for_outer` is more natural than `par_outer_for`. 

I realize that `par_for_outer` more sense from a logical grouping with `par_for`, I only used `par_outer_for` because it made more sense when spoken. I'll change to `par_for_outer` and `par_for_inner` though

tst/unit/kokkos_abstraction.cpp
forrestglines: `parthenon::team_mbr_t` might work, not too long, distinguishes the abbreviation

Yurlungur: Yeah, that sounds good.

src/CMakeLists.txt
forrestglines: I realize after looking into it, a `TeamVectorRange`, which squashes `TeamThreadRange` and `ThreadVectorRange`, is more correct here. I'm making that change.

forrestglines:
>     * added KOKKOS_INLINE_FUNTION decoration as my compiler cried (despite working) "warning: calling a **host** function from a **host** **device** function is not allowed"

I'm surprised that it worked without it.

>     * fixed typos in unit test and default wrapper

Oops, good catch!

PR #153, opened 2020-05-06
Brryan/error checking

Yurlungur:
Dummy MR to trigger CI


Yurlungur:
MR #118 was already approved. Pulling the trigger.

src/utils/error_checking.hpp
```diff
@@ -0,0 +1,73 @@
...
+#define PARTHENON_REQUIRE(condition, message)                                            \
+  if (!(condition)) {                                                                    \
+    parthenon::ErrorChecking::require(#condition, message, __FILE__, __LINE__);          \
+  }
+
+#define PARTHENON_FAIL(message)                                                          \
+  parthenon::ErrorChecking::fail(message, __FILE__, __LINE__);
+
+#ifdef NDEBUG
+#define PARTHENON_DEBUG_REQUIRE(condition, message) ((void)0)
```
nmsriram, line 37: Is ((void)0) idiomatic when a macro is to expand to nothing?

Yurlungur:
> Is ((void)0) idiomatic when a macro is to expand to nothing?

Yes. It ensures that the arguments of the macro don't get automatically introduced into the code. It's how `catch2`'s `REQUIRE` and how `assert.h` are written.

PR #154, opened 2020-05-07
Kokkos example for full mesh based on calculate_pi

nmsriram:
## PR Summary
This pull request introduces a new example that can be used to measure performance using the calculate_pi example but running it using Kokkos::parallel_for and eventually par_for().

Primary file is: examples/kokkos_pi/kokkos_pi.cpp

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


example/advection/advection_package.cpp
```diff
@@ -40,8 +40,6 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
   pkg->AddParam<>("vx", vx);
   Real vy = pin->GetOrAddReal("Advection", "vy", 1.0);
```
nmsriram, line 41: I don't know where these changes came from.  I haven't touched this file intentionally.  What is the best path forward to revert it?  Or should I let it be as is?

Yurlungur:
> I don't know where these changes in example/advection/advection_package.cpp came from. I haven't touched this file intentionally. What is the best path forward to revert it? Or should I let it be as is?

Yeah this is extremely weird. Can you merge in from master? Or pull in the file directly from master?

example/kokkos_pi/kokkos_pi.cpp
```diff
@@ -0,0 +1,299 @@
...
+using parthenon::ContainerIterator;
+using parthenon::DevExecSpace;
+using parthenon::loop_pattern_mdrange_tag;
+using parthenon::MeshBlock;
+using parthenon::Metadata;
+using parthenon::MetadataFlag;
+using parthenon::par_for;
+using parthenon::ParArray4D;
+using parthenon::ParArrayND;
+using parthenon::Real;
```
Yurlungur: Can we use the package stuff that @AndrewGaspar wrote to make this a little less verbose?

example/kokkos_pi/kokkos_pi.cpp
```diff
@@ -0,0 +1,299 @@
...
+
+using View1D = Kokkos::View<Real *, Kokkos::LayoutRight, Kokkos::DefaultExecutionSpace>;
+using View2D = Kokkos::View<Real **, Kokkos::LayoutRight, Kokkos::DefaultExecutionSpace>;
+using View3D = Kokkos::View<Real ***, Kokkos::LayoutRight, Kokkos::DefaultExecutionSpace>;
```
Yurlungur: Why not use `ParArray1D`, `ParArray2D`, etc?

example/kokkos_pi/kokkos_pi.cpp
```diff
@@ -0,0 +1,299 @@
...
+
+    // allocate space for origin coordinates and create host mirror view
+    View2D xyz("xyzBlocks", 3, n_mesh3);
+    auto h_xyz = Kokkos::create_mirror_view(xyz);
+
+    // *** Kludge warning ***
+    // Since our mesh is not GPU friendly we set up a hacked up
+    // collection of mesh blocks.  The hope is that when our mesh is
+    // up to par we will replace this code with the mesh
+    // infrastructure.
```
Yurlungur, line 153: presumably once the mesh is up to par, we can just add your timing stuff directly into `calculate_pi` no?

example/kokkos_pi/kokkos_pi.cpp
```diff
@@ -0,0 +1,299 @@
...
+                const int i_grid = idx - k_grid * n_block2 - j_grid * n_block; // iops = 4
+                const Real x =
+                    xyz(0, iMesh) + dxyzCell * static_cast<Real>(i_grid); // fops = 2
+                const Real y =
+                    xyz(1, iMesh) + dxyzCell * static_cast<Real>(j_grid); // fops = 2
+                const Real z =
+                    xyz(2, iMesh) + dxyzCell * static_cast<Real>(k_grid); // fops = 2
+                const Real myR2 = x * x + y * y + z * z;                  // fops = 5
+                inOrOut(0, k_grid + NGHOST, j_grid + NGHOST, i_grid + NGHOST) =
+                    (myR2 < radius2 ? 1.0 : 0.0); // iops = 3
```
Yurlungur, line 227: it woudl be interesting to see what this looks like performance-wise with `par_for` instead of `kokkos::parallel_for`

example/kokkos_pi/kokkos_pi.cpp
```diff
@@ -0,0 +1,299 @@
...
+              });
+        }
+      });
+      // wait for work to finish, redundant due to timing loop
+      Kokkos::fence();
```
Yurlungur, line 232: since this is redundant with the timing loop, I might remove it. It might introduce extra latency we can get rid of.

example/kokkos_pi/kokkos_pi.cpp
```diff
@@ -0,0 +1,299 @@
...
+    double time_basic;
+    {
+      MeshBlock *pStart = allBlocks[0];
+      time_basic = kernel_timer_wrapper(0, n_iter, [&]() {
+        MeshBlock *pmb = pStart;
+        for (int iMesh = 0; iMesh < n_mesh3; iMesh++, pmb = pmb->next) {
+          Container<Real> &base = pmb->real_containers.Get();
+          auto inOrOut = base.PackVariables({Metadata::Independent});
+          // iops = 8  fops = 11
+          Kokkos::parallel_for(
```
Yurlungur, line 214: Because you're launching a kernel per meshblock, this is going to be incredibly slow. It'll be interesting to watch this improve as we implement streams, etc, to hide kernel launch latency.

nmsriram:
> Yeah this is extremely weird. Can you merge in from master? Or pull in the file directly from master?

I have reverted the changes.  I think it was because I started with master, merged in jdolence/iterator_musings and then merged master after josh's branch got merged in.  Somewhere along the way git forgot what the order is.

example/kokkos_pi/kokkos_pi.cpp
Yurlungur: Actually, we might be able to coalesce over blocks, since you have the blocks in a kokkos view. It would be interesting to try that and look at the timing.

example/kokkos_pi/kokkos_pi.cpp
Yurlungur: Ah I see you already did. :)

example/kokkos_pi/kokkos_pi.cpp
nmsriram: Good point.

example/kokkos_pi/kokkos_pi.cpp
```diff
@@ -0,0 +1,381 @@
...
+        lastBlock = pmb;
+      }
+    }
+  }
+  // copy our coordinates over to Device and wait for completion
+  Kokkos::deep_copy(xyz, h_xyz);
+  Kokkos::fence();
+
+  return firstBlock;
+}
```
Yurlungur, line 210: This isn't necessary, but it might be interesting:

If meshblocks can be copied to the GPU via capture, perhaps we could make a Kokkos view of them to represent a mesh rather than a tree. The advantage of such a test would be if coalescing over meshblocks offers a big performance boost.

I think for this code, that would be a different setup routine, like `setupKokkosMesh` or something.

example/kokkos_pi/kokkos_pi.cpp
```diff
@@ -0,0 +1,381 @@
...
+
+void deleteMesh(MeshBlock *firstBlock) {
+  // deletes the mesh chain starting with firstBlock
+  auto pmb = firstBlock;
+  while (pmb) {
+    auto next = pmb->next;
+    delete pmb;
+    pmb = next;
+  }
+}
```
Yurlungur, line 220: 👍

example/kokkos_pi/kokkos_pi.cpp
```diff
@@ -0,0 +1,381 @@
...
+  // up to par we will replace this code with the mesh
+  // infrastructure.
+
+  const Real dxyzCell = radius / static_cast<Real>(n_mesh * n_block);
+  auto h_xyz = Kokkos::create_mirror_view(xyz);
+
+  // Set up our mesh.
+  Metadata myMetadata({Metadata::Independent, Metadata::Cell});
+  MeshBlock *firstBlock = nullptr;
+  MeshBlock *lastBlock = nullptr;
```
Yurlungur, line 175: some day we should reconsider using raw pointers for meshblocks. Probably shouldn't be today though.

example/kokkos_pi/kokkos_pi.cpp
```diff
@@ -0,0 +1,381 @@
...
+  // This policy is over one block
+  auto policyBlock = Kokkos::RangePolicy<>(Kokkos::DefaultExecutionSpace(), 0, n_block3,
+                                           Kokkos::ChunkSize(512));
+
+  MeshBlock *pStart = firstBlock;
+  double time_basic = kernel_timer_wrapper(0, n_iter, [&]() {
+    MeshBlock *pmb = pStart;
+    for (int iMesh = 0; iMesh < n_mesh3; iMesh++, pmb = pmb->next) {
+      Container<Real> &base = pmb->real_containers.Get();
+      auto inOrOut = base.PackVariables({Metadata::Independent});
```
Yurlungur, line 248: This will be cached on a per meshblock basis.

example/kokkos_pi/kokkos_pi.cpp
```diff
@@ -0,0 +1,381 @@
...
+using View2D = Kokkos::View<Real **, Kokkos::LayoutRight, Kokkos::DefaultExecutionSpace>;
+
+// The result struct contains results of different tests
+typedef struct result_t {
+  std::string name; // The name of this test
+  Real pi;          // the value of pi calculated
+  Real t;           // time taken to run test
+  int iops;         // number of integer ops
+  int fops;         // number of floating point ops
+} result_t;
```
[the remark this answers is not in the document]
nmsriram: I went with struct because it is just a data container.  However primarily because I come from a C world. :-) 
I guess if we use it outside this file, we may change it to a class.

example/kokkos_pi/kokkos_pi.cpp
nmsriram: Right, so when we come in for the second iteration, we get the cached version, correct?  That is why I put the burn iterations for the timer as 0.

example/kokkos_pi/kokkos_pi.cpp
nmsriram: I'll try my hand at that next.  I was also thinking of playing with nested parallelism of Kokkos since we can create a parND view-of-views across multiple blocks and hand that off to the par_for.

Yurlungur: Yeah, that's a good idea.

One thing that would certainly work would be a (yuck) Kokkos view of VariablePacks, which would also achieve what we want. A `VariableBlockPack` or something.

example/kokkos_pi/kokkos_pi.cpp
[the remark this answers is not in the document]
Yurlungur: It's not important, obviously. I like the C++-style struct, since I think it gets the same idea accross with slightly fewer characters. In general I like C++-style structs for containers with all public scope and classes for things that might have private scope.

example/kokkos_pi/kokkos_pi.cpp
Yurlungur: Yep that's right :)

example/kokkos_pi/kokkos_pi.cpp
```diff
@@ -0,0 +1,381 @@
...
+
+  return perf_time;
+}
+
+static void usage(std::string program) {
+  std::cout << std::endl
+            << "    Usage: " << program << " n_block n_mesh n_iter" << std::endl
+            << std::endl
+            << "             n_block = size of each mesh block on each axis" << std::endl
+            << "              n_mesh = number mesh blocks alogn each axis" << std::endl
```
JoshuaSBrown, line 123: Typo 'alogn'

example/kokkos_pi/kokkos_pi.cpp
```diff
@@ -0,0 +1,381 @@
...
+#include "mesh/mesh.hpp"
+
+using Real = double;
+
+using parthenon::Container;
+using parthenon::DevExecSpace;
+using parthenon::MeshBlock;
+using parthenon::Metadata;
+using parthenon::par_for;
+using parthenon::Real;
```
JoshuaSBrown, line 70: I think you can get around most of these statements with:

```
#include <parthenon/driver.hpp>

using namespace parthenon::driver::prelude;
```

nmsriram: @JoshuaSBrown Fixed typo and used the package prelude.
Thanks, I should have done that from the start.

example/kokkos_pi/kokkos_pi.cpp
```diff
@@ -0,0 +1,368 @@
...
+               time_basic, niops, nfops};
+
+  // Clean up the mesh
+  deleteMesh(firstBlock);
+
+  return r;
+}
+
+result_t naiveParFor(int n_block, int n_mesh, int n_iter, double radius) {
+  // creates a mesh and rusn a basic par_for implementation for looping through blocks.
```
JoshuaSBrown, line 270: There is another typo here "rusn". :)

PR #157, opened 2020-05-11
Enforce cpplint, reintroduce UserWorkAfterLoop, add VL2 integrator, fix cmd line parsing

pgrete:
## PR Summary

As the title suggest, multiple small fixes I stumbled across when integrating with AthenaPK:
- Enforce cpplint (stage is not allowed to fail any more during CI, but also run last so that regression tests will always executed first)
- reintroduce UserWorkAfterLoop (Athena++ feature that we removed from the ParthenonManager/driver infrastructure during transition -- very helpful for regression/convergence tests)
- add VL2 integrator beta (also reintroduced from original Athena++ code)
- fix cmd line parsing to override input parameter options (the addition of prefixes, such as `parthenon/time` broke the parsing based on the first occurrence of `/`. Now works for both blocks with and without prefixes).

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/src/driver/driver.cpp
+++ b/src/driver/driver.cpp
@@ -58,6 +58,8 @@ DriverStatus EvolutionDriver::Execute() {
     }
   } // END OF MAIN INTEGRATION LOOP ======================================================
 
+  pmesh->UserWorkAfterLoop(pinput, tm);
+
   DriverStatus status = DriverStatus::complete;
 
   pouts->MakeOutputs(pmesh, pinput, &tm);
--- a/src/outputs/outputs.cpp
+++ b/src/outputs/outputs.cpp
@@ -418,9 +418,10 @@ void Outputs::MakeOutputs(Mesh *pm, ParameterInput *pin, SimTime *tm) {
   bool first = true;
   OutputType *ptype = pfirst_type_;
   while (ptype != nullptr) {
-    if ((ptype->output_params.dt >= 0.0) &&
-        (tm == nullptr || (tm->time == tm->start_time) ||
-         (tm->time >= ptype->output_params.next_time) || (tm->time >= tm->tlim))) {
+    if ((tm == nullptr) ||
+        ((ptype->output_params.dt >= 0.0) &&
+         ((tm->time == tm->start_time) || (tm->time >= ptype->output_params.next_time) ||
+          (tm->time >= tm->tlim)))) {
       if (first && ptype->output_params.file_type != "hst") {
         pm->ApplyUserWorkBeforeOutput(pin);
         first = false;
--- a/src/outputs/outputs.cpp
+++ b/src/outputs/outputs.cpp
@@ -418,8 +418,9 @@ void Outputs::MakeOutputs(Mesh *pm, ParameterInput *pin, SimTime *tm) {
   bool first = true;
   OutputType *ptype = pfirst_type_;
   while (ptype != nullptr) {
-    if (tm == nullptr || (tm->time == tm->start_time) ||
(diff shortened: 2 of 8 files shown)


src/parameter_input.cpp
```diff
@@ -397,7 +397,7 @@ void ParameterInput::ModifyFromCmdline(int argc, char *argv[]) {
 
   for (int i = 1; i < argc; i++) {
     input_text = argv[i];
-    std::size_t slash_posn = input_text.find_first_of("/"); // find "/" character
+    std::size_t slash_posn = input_text.rfind("/");         // find last "/" character
     std::size_t equal_posn = input_text.find_first_of("="); // find "=" character
```
nmsriram, line 401: This is not something that you introduced, but there should be a check to see whether the equals position is greater than slash_posn+1.   Or the slash search should be in the substring [:equal_posn].  Otherwise line 408 will crash:
````C++
    name = input_text.substr(slash_posn + 1, (equal_posn - slash_posn - 1));
````
It is possible that this is not a case that we will encounter.  If so, then feel free to ignore this statement.

File: src/driver/driver.cpp [hunk 2]
```diff
@@ -58,6 +58,8 @@ DriverStatus EvolutionDriver::Execute() {
     }
   } // END OF MAIN INTEGRATION LOOP ======================================================
 
+  pmesh->UserWorkAfterLoop(pinput, tm);
```
jdolence, line 61: It seems fine to reintroduce this for now, but @jmstone and I discussed turning ParthenonManager::PreDriver and ParthenonManager::PostDriver into something like Driver::PreExecute and Driver::PostExecute.  When that gets done, perhaps this functionality belongs there?

src/outputs/outputs.cpp
```diff
@@ -418,8 +418,9 @@ void Outputs::MakeOutputs(Mesh *pm, ParameterInput *pin, SimTime *tm) {
   bool first = true;
   OutputType *ptype = pfirst_type_;
   while (ptype != nullptr) {
-    if (tm == nullptr || (tm->time == tm->start_time) ||
-        (tm->time >= ptype->output_params.next_time) || (tm->time >= tm->tlim)) {
+    if ((ptype->output_params.dt >= 0.0) &&
```
jdolence, line 421: With this change, does the calculate_pi example still produce the hdf5 output it did before?

src/driver/driver.cpp
pgrete: Good point. I'll open an issue so that we can keep track of it.

src/parameter_input.cpp
pgrete: Good catch. Not completely unlikely that a user is going to use a `/` in the parameter file within a value.
Updated result now looks like
```
example/calculate_pi/pi-example  -i ../example/calculate_pi/parthinput.example parthenon/mesh/nx1=64/3
### PARTHENON ERROR
  Message:     '/' used as value (rhs of =) when modifying parthenon/mesh/nx1=64/3.Please update value of change logic in ModifyFromCmdline function.
  File:        /home/pgrete/src/parthenon/src/parameter_input.cpp
  Line number: 407
```

src/outputs/outputs.cpp
pgrete: Correct. I fixed the logic so that the test for `dt >=0` only occurs if the `tm` object exists.

pgrete:
@jdolence if you're happy with the latest change feel free to approve and merge.

PR #159, opened 2020-05-11
Prevent Access To Unitialized Arrays

Yurlungur:
## PR Summary

@JoshuaSBrown discovered a bug in the 2D `advection` test. When compiled with `-DCMAKE_BUILD_TYPE=Debug`, `Kokkos` bounds checking produced an error for `Kokkos::subview` and the code died.

The problem is that in two-dimensions `flux[2]` for a `Variable` is unitialized. However, `VariablePack` and `PackVariablesAndFluxes` attempt to extract a 3D slice of `Variable->flux[2]` regardless. This is an invalid operation.

This PR is intended to contain all changes required to solve this. They are as follows:
1. There needs to be error checking in `parthenon_arrays.hpp`. `ParArrayND.Get` should die with a sensible error message if a subview of an unitialized array is attempted, consistent with the `Kokkos` design.
2. `PackVariablesAndFluxes` needs to check for the dimension of the problem, and not request `flux[2]` in 2D or `flux[1]` in 1D.
3. Just as with `ParArrayND`, variable pack objects need to produce a sensible error message if an unitialized variable is indexed into from the pack.

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted


Yurlungur:
@jdolence and I are in the process of adding the required features. The `WIP` status will be removed soon.

Yurlungur:
@jdolence @JoshuaSBrown @pgrete this is ready for review. Please take a look and lets get this merged in.

tst/unit/test_container_iterator.cpp
```diff
@@ -290,6 +290,13 @@ TEST_CASE("Can pull variables from containers based on Metadata", "[ContainerIte
         REQUIRE(!intervals_intersect(imap["v3"], imap["vsparse"]));
       }
     }
+
+    WHEN("we add a 2d variable") {
+      std::vector<int> twod_block_size{16, 16, 1};
```
JoshuaSBrown, line 295: Is there any reason we would pass in something like {16, 1, 16} ?

jdolence: I don't think so.  To be honest, most development done on top of parthenon won't ever have to even specify the grid dimensions.  The intent is that developers specify an association like cell, face, node, ... and parthenon takes care of the rest.  That said, I'm actually not sure what would happen if you passed in {16,1,16}.  It might (probably) just says OK this is 3D and everything would work just fine.

Yurlungur: I don't think `{16, 1, 16}` should ever happen. That would mean `dim2=1` and `dim3=16`. I think the code assumes the trailing index is always the absent one.

JoshuaSBrown: I'm just wondering if the user will get useful feedback if they try to do something like that, if not it might be worth addressing in the future. This pull request addresses the relevant issue, so I'm happy.

Yurlungur:
All working and pretty. Pulling the trigger.

tst/unit/test_container_iterator.cpp
Yurlungur: Hmm. If it comes up, we can add a warning in `Container.Add` where this shows up.

Issue #132, opened 2020-04-28
Make python regression testing script more general


JoshuaSBrown:
Current implementation is not general enough to look at convergence criteria for slight differences in input parameters. A run method needs to be added to each test case so users have the ability to call the same executable if checking convergence criteria for slightly different inputs if needed.

AndrewGaspar:
Seems to be fixed by #165

Issue #162, opened 2020-05-12
Driver Pre/PostExecute


pgrete:
Open an issue to keep track of a comment in #157 where `pmesh->UserWorkAfterLoop(pinput, tm);` was reintroduced to `ParthenonManager`.
A more suitable place is like within a `Driver` based `Pre/PostExecute`.
Tagging @jmstone and @jdolence as they already discussed sth along those lines.

Issue #119, opened 2020-04-21
Driver specific outputs?


JoshuaSBrown:
In the original athena++ code there was a way for users to define their own inputs which could be written to the supported file formats, that functionality has since been pulled out. Refactors in the code such as the driver class also mean that a new and improved implementation is probably warranted. There is a need for such a feature, for instance in the case of regression testing it would be ideal if parameters specific to each driver were also output in the supported file formats. 

Topics I would like to address in this issue is:

1. Is there much of a need for this functionality outside of testing?
2. If so, how should such a feature be implemented and what functionality would be useful?

pgrete:
I'm not sure I completely understand the connection between the inputs  (" to define their own inputs" ) and outputs ("written to the supported file formats"/"parameters specific to each driver were also output in the supported file formats.").
Are you referring to user defined fields (such like the `one_minus_advected` field in the advection example or referring to the input parameters?

JoshuaSBrown:
Neither, take the calculate pi example, the actual value of pi is not in a user defined field, it is simply printed at the end of the run. It would probably be ideal if it was also output in the existing supported file formats.

Yurlungur:
I'm also not sure I understand. pi is a number, not a field on the grid. Are you suggesting we add the capability to add... I guess metadata? numbers and strings and the like?

JoshuaSBrown:
> I'm also not sure I understand. pi is a number, not a field on the grid. Are you suggesting we add the capability to add... I guess metadata? numbers and strings and the like?

Yes, but only if it is useful.

Yurlungur:
This would be useful, I think. But I'm not sure it belongs in the driver. We probably want the ability to register metadata as output with the IO system itself.

AndrewGaspar:
Seems like there isn't a lot of interest in this feature, so let's close it.

PR #165, opened 2020-05-13
Add convergence test infrastructure and test itself

pgrete:
## PR Summary
Introduces to option to run multiple steps in a single regression test where the Prepare function may override the input file parameters depending on stage.
As an example I added a first order regression test based on the advection example.

A sample output of the advection test looks like:
```
$ cat /home/pgrete/src/parthenon/tst/regression/test_suites/advection_convergence/output/advection-errors.dat
# Nx1  Nx2  Nx3  Ncycle  L1 max_error/L1  max_error 
32  4  4  107  1.119646e-07   1.563957e+00  1.751078e-07  
64  4  4  214  6.187221e-08   1.569180e+00  9.708863e-08  
128  4  4  427  3.258335e-08   1.570405e+00  5.116905e-08  
256  4  4  854  1.672761e-08   1.570700e+00  2.627406e-08  
512  4  4  1707  8.475962e-09   1.570772e+00  1.331381e-08  
4  32  4  107  1.119646e-07   1.563957e+00  1.751078e-07  
4  64  4  214  6.187221e-08   1.569180e+00  9.708863e-08  
4  128  4  427  3.258335e-08   1.570405e+00  5.116905e-08  
4  256  4  854  1.672761e-08   1.570700e+00  2.627406e-08  
4  512  4  1707  8.475962e-09   1.570772e+00  1.331381e-08  
4  4  32  107  1.119646e-07   1.563957e+00  1.751078e-07  
4  4  64  214  6.187221e-08   1.569180e+00  9.708863e-08  
4  4  128  427  3.258335e-08   1.570405e+00  5.116905e-08  
4  4  256  854  1.672761e-08   1.570700e+00  2.627406e-08  
4  4  512  1707  8.475962e-09   1.570772e+00  1.331381e-08  
32  32  32  107  1.996590e-07   3.250386e+00  6.489690e-07  
64  64  64  214  1.395265e-07   3.196873e+00  4.460486e-07 
```

Note that we don't get 1st order conversion in the last two rows (where advection takes place across the diagonal).
I'm currently unsure if this is expected for the given setup (problem generator) with the chosen algorithm or not. Maybe other people can chime in here.

This is currently WIP as I still would like to add
- [x] AMR test
- [x] Testing hdf5 outputs

Nevertheless, any early comment is welcome.

Addresses #132 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


pgrete:
Also note that this PR is currently based on #157 so it'll be cleaner once #157 is in master.

tst/regression/utils/test_case.py
```diff
@@ -212,18 +212,6 @@ def Analyse(self):
...
-        test_pass = False
-
-        print("*****************************************************************")
-        print("Analysing Driver Output")
-        print("*****************************************************************")
-        sys.stdout.flush()
-        test_pass = self.test_case.Analyse(self.parameters)
-
-        return test_pass
-
```
JoshuaSBrown, line 226: Good catch!

JoshuaSBrown:
I will wait till this is no longer WIP to review but this is great @pgrete thanks for doing this!

Yurlungur:
> Note that we don't get 1st order conversion in the last two rows (where advection takes place across the diagonal).
> I'm currently unsure if this is expected for the given setup (problem generator) with the chosen algorithm or not. Maybe other people can chime in here.

Hmm. We're advecting a shock, so I don't think we can expect first-order convergence. In the infinity norm (i.e., max error), I would expect no convergence at all. In the L1 norm, I think I'd expect 1/2-order convergence or something along those lines. 

That brings up another issue: We should add a smooth test that converges in the infinity norm. Maybe a wave or diffusion solve. Obviously that's beyond the scope here. But once this is in master, we should add it.

.gitlab-ci.yml
```diff
@@ -59,6 +58,7 @@ parthenon-cuda-short:
     paths:
       - build-cuda-debug/CMakeFiles/CMakeOutput.log
       - build-cuda-debug-uvm/CMakeFiles/CMakeOutput.log
+      - tst/regression/test_suites/advection_convergence/output/advection-errors.dat
```
Yurlungur, line 61: Should we generate a plot automatically as well as a dat file?

example/advection/advection_package.cpp
```diff
@@ -194,9 +194,9 @@ Real EstimateTimestep(Container<Real> &rc) {
       pmb->pcoord->CenterWidth2(k, j, is, ie, dx2);
       pmb->pcoord->CenterWidth3(k, j, is, ie, dx3);
       for (int i = is; i <= ie; i++) {
-        min_dt = std::min(min_dt, dx1(i) / std::abs(vx));
-        min_dt = std::min(min_dt, dx2(i) / std::abs(vy));
-        min_dt = std::min(min_dt, dx3(i) / std::abs(vz));
+        if (vx != 0.0) min_dt = std::min(min_dt, dx1(i) / std::abs(vx));
+        if (vy != 0.0) min_dt = std::min(min_dt, dx2(i) / std::abs(vy));
+        if (vz != 0.0) min_dt = std::min(min_dt, dx3(i) / std::abs(vz));
```
Yurlungur, line 199: This is obviously an improvement over the previous code. However, the floating-point `!=` test strikes me as strange. Not sure there's a good solution, though. Checking for a very small velocity isn't right either.

example/advection/parthinput.advection
```diff
@@ -34,25 +34,30 @@ x2max = 0.5
...
 <parthenon/meshblock>
 nx1 = 16
 nx2 = 16
-#nx3 = 1
+nx3 = 16
 
 <parthenon/time>
 tlim = 1.0
-integrator = rk2
+integrator = rk1
```
Yurlungur, line 50: Since the spatial stencil is 1st-order, I would expect 1st-order convergence with `rk2` as well as `rk1`. Does something go wrong?

example/advection/redefine_parthenon_defaults.cpp
```diff
@@ -34,11 +37,16 @@ void MeshBlock::ProblemGenerator(ParameterInput *pin) {
...
+  Real x3size = pmy_mesh->mesh_size.x3max - pmy_mesh->mesh_size.x3min;
+
   for (int k = 0; k < ncells3; k++) {
     for (int j = 0; j < ncells2; j++) {
       for (int i = 0; i < ncells1; i++) {
-        Real rsq = std::pow(pcoord->x1v(i), 2) + std::pow(pcoord->x2v(j), 2);
-        q(k, j, i) = (rsq < 0.15 * 0.15 ? 1.0 : 0.0);
+        q(k, j, i) = 1e-6 * sin(pcoord->x1v(i) / x1size * TWO_PI) *
+                     cos(pcoord->x2v(j) / x2size * TWO_PI) *
+                     cos(pcoord->x3v(k) / x3size * TWO_PI);
```
Yurlungur, line 49: Wait. I stand corrected. This isn't a shock anymore. We should be seeing 1st-order convergence.

tst/regression/test_suites/advection_convergence/advection_convergence.py
```diff
@@ -0,0 +1,183 @@
...
+        elif step == 18:
+            parameters.driver_cmd_line_args = [
+                'parthenon/meshblock/nx1=8',
+                'parthenon/meshblock/nx2=16',
+                'parthenon/meshblock/nx3=8',
+                ]
+
+        return parameters
+
+    def Analyse(self,parameters):
```
Yurlungur, line 120: Maybe we could add plotting code in here and generate a convergence plot.

Yurlungur:
> > Note that we don't get 1st order conversion in the last two rows (where advection takes place across the diagonal).
> > I'm currently unsure if this is expected for the given setup (problem generator) with the chosen algorithm or not. Maybe other people can chime in here.
> 
> Hmm. We're advecting a shock, so I don't think we can expect first-order convergence. In the infinity norm (i.e., max error), I would expect no convergence at all. In the L1 norm, I think I'd expect 1/2-order convergence or something along those lines.
> 
> That brings up another issue: We should add a smooth test that converges in the infinity norm. Maybe a wave or diffusion solve. Obviously that's beyond the scope here. But once this is in master, we should add it.

Wait I see you changed the setup to be a `cos` thing. That's good.

example/advection/redefine_parthenon_defaults.cpp
Yurlungur: Actually wait. No I see the issue. This thing has the wrong wavelength. It's not periodic in the `X1 + X2 + X3` direction. The wavelength should be `2*pi*sqrt(x1size*x1size + x2size*x2size + x3size*x3size)`. That's probably why you're not seeing convergence.

example/advection/advection_package.cpp
pgrete: There's just one representation of 0 as far as I know and the division is only a problem if the value is actually 0, so I think this should be fine here.

example/advection/redefine_parthenon_defaults.cpp
pgrete: Yes, thanks for pointing that out. This is fixed now.

tst/regression/test_suites/advection_convergence/advection_convergence.py
pgrete: I'm still undecided whether this is sth we want or not as the convergence test should mainly be a regression test of the "integrated framework" and not a the specific advection algorithm.
So from may point of view, knowing that it failed and where (the regression test prints the lines that are off in addition to the dat file as artifact) are sufficient in terms of what this test should achieve.
However, if we the test should be an example of how to produce integrated convergence plots, then I'm happy to add it.

example/advection/parthinput.advection
pgrete: Using rk2 results in worse errors (30-40%) for 1D tests and doesn't converge for the diagonal advection (error only goes down by a factor or 1.4 when doubling linear resolution).
Not entirely sure what's going on (potentially double spatial reconstruction per timestep for rk2 amplifies the first order spatial error), but given that I mainly added to to ensure that the mesh infrastructure works I'm not that concerned about rk2.

example/advection/advection_package.cpp
Yurlungur: Yes. I agree. This has code-smell, but I think it's fine.

example/advection/parthinput.advection
Yurlungur: Hmm interesting. Yes let's leave this for now. It might be worth digging into this later, though.

tst/regression/test_suites/advection_convergence/advection_convergence.py
Yurlungur: > However, if we the test should be an example of how to produce integrated convergence plots, then I'm happy to add it.

I do think this would be useful.

> So from may point of view, knowing that it failed and where (the regression test prints the lines that are off in addition to the dat file as artifact) are sufficient in terms of what this test should achieve.

Even from this point of view, I find it much easier to catch errors and understand what's going on with a plot. The tabulated output is not the easiest to read, whereas a plot will instantly tell us what is going wrong and how.

pgrete:
Alright, this is ready for review now.
The CI test currently fail for Cuda because Cuda is broken on master #167 so please neglect this "problem" for now in the review.

pgrete:
Compilation with Cuda is fixed now but the tests take ages on the device due to using UVM.
If everyone agree we either allow failure of those tests for now (from a CI point of view) or wait until the Kokkosize PR is done and then merge both.
Anyway this can be considered "done" pending any additional comments (I'll add the python plots over the next days, focusing on the Kokkos parts first now).

Yurlungur:
> If everyone agree we either allow failure of those tests for now (from a CI point of view) or wait until the Kokkosize PR is done and then merge both.

Works for me. Will github allow us to merge with failed CI though?

example/advection/parthinput.advection
```diff
@@ -51,8 +53,12 @@ integrator = rk2
 cfl = 0.45
 vx = 1.0
 vy = 1.0
+vz = 1.0
+profile = hard_sphere
```
Yurlungur, line 57: Below the profiles are enumerated 0 through 2. And here it's a string. Am I missing something? Also string would definitely be better.

example/advection/redefine_parthenon_defaults.cpp
```diff
@@ -33,11 +38,36 @@ Packages_t ParthenonManager::ProcessPackages(std::unique_ptr<ParameterInput> &pi
...
+          Real rsq = coords.x1v(i) * coords.x1v(i) + coords.x2v(j) * coords.x2v(j) +
+                     coords.x3v(k) * coords.x3v(k);
+          q(k, j, i) = 1. + amp * exp(-100.0 * rsq);
+        } else if (profile == 2) { // hard_sphere
+          Real rsq = coords.x1v(i) * coords.x1v(i) + coords.x2v(j) * coords.x2v(j) +
+                     coords.x3v(k) * coords.x3v(k);
+          q(k, j, i) = (rsq < 0.15 * 0.15 ? 1.0 : 0.0);
+        } else {
+          q(k, j, i) = 0.0;
+        }
```
Yurlungur, line 70: Nice. I like the flexibility of choosing profiles. But see my above comment about a difference between the parthinput file and this bit.

src/utils/error_checking.hpp
```diff
@@ -51,19 +51,16 @@ namespace ErrorChecking {
...
 void require(std::string const &condition, std::string const &message,
              std::string const &filename, int const linenumber) {
-  fprintf(stderr,
-          "### PARTHENON ERROR\n  Condition:   %s\n  Message:     %s\n  File:        "
-          "%s\n  Line number: %i\n",
-          condition.c_str(), message.c_str(), filename.c_str(), linenumber);
+  printf("### PARTHENON ERROR\n  Condition:   %s\n  Message:     %s\n  File:        "
+         "%s\n  Line number: %i\n",
+         condition.c_str(), message.c_str(), filename.c_str(), linenumber);
   exit(EXIT_FAILURE);
```
Yurlungur, line 57: Presumably the MR that fixes this should go in before this one.

tst/regression/test_suites/advection_convergence/advection_convergence.py
Yurlungur: I guess you decided not to add the plot? Fair enough. It's not a huge deal either way.

example/advection/redefine_parthenon_defaults.cpp
```diff
@@ -48,4 +78,129 @@ void ParthenonManager::SetFillDerivedFunctions() {
...
+    std::stringstream msg;
+    FILE *pfile;
+
+    // The file exists -- reopen the file in append mode
+    if ((pfile = std::fopen(fname.c_str(), "r")) != nullptr) {
+      if ((pfile = std::freopen(fname.c_str(), "a", pfile)) == nullptr) {
+        msg << "### FATAL ERROR in function Mesh::UserWorkAfterLoop" << std::endl
+            << "Error output file could not be opened" << std::endl;
+        PARTHENON_FAIL(msg.str());
+      }
```
JoshuaSBrown, line 180: I wanted to get your opinion on this https://github.com/lanl/parthenon/issues/119.

scripts/python/phdf_diff.py
```diff
@@ -64,51 +64,28 @@ def addPath():
...
-        Usage()
-        exit(1)
 
     # Load first file and print info
     try:
         f0 = phdf(files[0])
         if not quiet: print(f0)
     except:
-        exit(1)
+        print("whoosp")
```
JoshuaSBrown, line 87: Could we get a slightly more descriptive message?

scripts/python/phdf_diff.py
```diff
@@ -141,7 +118,7 @@ def addPath():
     if not brief and not quiet:
         print('____Comparing on a per variable basis with tolerance %.16g'%tol)
     breakOut = False
-    oneTenth = f0.TotalCells/10
+    oneTenth = f0.TotalCells//10
```
JoshuaSBrown, line 121: Was this intentional?

tst/regression/CMakeLists.txt
```diff
@@ -49,6 +52,23 @@ list(APPEND TEST_DIRS calculate_pi)
...
+
+if (ENABLE_HDF5)
+setup_test( output_hdf5 "--driver ${CMAKE_BINARY_DIR}/example/advection/advection-example \
+--driver_input ${CMAKE_CURRENT_SOURCE_DIR}/test_suites/output_hdf5/parthinput.advection \
+--num_steps 2")
+setup_test_mpi( ${NUM_MPI_PROC_TESTING} output_hdf5
+"--driver ${CMAKE_BINARY_DIR}/example/advection/advection-example \
+--driver_input ${CMAKE_CURRENT_SOURCE_DIR}/test_suites/output_hdf5/parthinput.advection \
+--num_steps 2")
+endif()
```
JoshuaSBrown, line 71: I think this would be a lot cleaner, if you just added a separate list for MPI processors and used the existing for loop below. But it would also be nice to know why you need to hard code a single mpi processor for the advection test? The variable NUM_MPI_PROC_TESTING can always be changed when building.

tst/regression/run_test.py
```diff
@@ -91,7 +92,12 @@ class TestError(RuntimeError):
                         nargs=1,
                         required=True,
                         help=desc)
-    
+
+    parser.add_argument('--output_dir','-o',
+                        type=str,
+                        help="path to simulation outputs. " +
+                             "Defaults to individual \"output\" folders in regression src.")
+
```
JoshuaSBrown, line 100: This is great!

pgrete:
@JoshuaSBrown was the force push here intentional?
I see that the branch now is four days older than the latest changes I made.

tst/regression/CMakeLists.txt
pgrete: I had to hard code using a single proc here as the tests cover the case that there's only one meshblock covering the entire mesh. Thus, using more than 1 process will fail. It's an edge case, but still worth including in the test from my point of view.

scripts/python/phdf_diff.py
pgrete: Yes, this makes sure the result is an integer (python 3).

scripts/python/phdf_diff.py
pgrete: Yes, I'll fix that. Debugging remnant :D

example/advection/redefine_parthenon_defaults.cpp
pgrete: Just left a comment there. I'm not sure completely understand. Is #119 the same as #162 ?

.gitlab-ci.yml
```diff
@@ -93,26 +89,42 @@ parthenon-cuda-peformance:
...
       -DCMAKE_CXX_COMPILER=${PWD}/../external/Kokkos/bin/nvcc_wrapper
       ../ && make -j${J} && ctest -L performance
+# run regression tests with UVM until everything is on the device
+# mitigate UVM allocations on older multi-GPU systems
+    - export CUDA_LAUNCH_BLOCKING=1
+    - export CUDA_VISIBLE_DEVICES=0
+    - cd ..
+    - mkdir build-cuda-perf-uvm
+    - cd build-cuda-perf-uvm
+    - cmake -DCMAKE_BUILD_TYPE=Release -DHDF5_ROOT=/usr/local/hdf5/parallel
```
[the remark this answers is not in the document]
pgrete: Good idea. I think this makes sense. I'll pull the test out that covers most of the integrated features and also run it in debug mode.

tst/regression/CMakeLists.txt
```diff
@@ -49,6 +50,13 @@ list(APPEND TEST_DIRS calculate_pi)
 list(APPEND TEST_ARGS "--driver ${CMAKE_BINARY_DIR}/example/calculate_pi/pi-example \
 --driver_input ${CMAKE_CURRENT_SOURCE_DIR}/test_suites/calculate_pi/parthinput.regression")
 
+# Advection test (here added manually to override default nproc for MPI runs)
+setup_test( advection_convergence "--driver ${CMAKE_BINARY_DIR}/example/advection/advection-example \
+--driver_input ${CMAKE_CURRENT_SOURCE_DIR}/test_suites/advection_convergence/parthinput.advection \
+--num_steps 18")
+setup_test_mpi( 1 advection_convergence "--driver ${CMAKE_BINARY_DIR}/example/advection/advection-example \
```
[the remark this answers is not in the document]
pgrete: See below (using only one MeshBlock for the entire mesh means we could only use one process).

src/utils/error_checking.hpp
pgrete: Yes.

example/advection/parthinput.advection
pgrete: Originally I didn't want to do string comparisons within a device kernel.
Now that all this runs on the host I can actually use the strings here.

JoshuaSBrown:
> @JoshuaSBrown was the force push here intentional?
> I see that the branch now is four days older than the latest changes I made.

I honestly have no idea how that happened. I don't recall ever pushing to this branch...

tst/regression/test_suites/advection_convergence/advection_convergence.py
pgrete: I'll add it now. I wanted to make everything else works first.

pgrete:
@JoshuaSBrown and @Yurlungur 
I just made some additional changes addressing the comments. 
From my point of view this is good to go now.

.gitlab-ci.yml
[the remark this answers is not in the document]
pgrete: I thought about it again, and am currently favoring not to running the tests in debug mode as failover mode for now.
From my point of view, the developer will have to address issues anyway if the tests fail and to it seems easier to just create the debug build in your local development environment rather than going through a lengthy log file without any way to intervene.
Do you think regression tests with a debug build are essential at this point or would you also be happy if we postpone the introduction of those tests if this is sth that turns out to be wished for/required by many developers?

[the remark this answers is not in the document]
Yurlungur: I am of two minds on this. On the one hand, I agree with @pgrete that the developer will have to make a debug build locally. On the other hand, there have been cases we've experienced recently where a debug build failed and the release build does not. 

So I lean towards running the unit tests in debug mode *every* time but only the unit tests. Integration tests and performance tests should probably just run in release mode.

.gitlab-ci.yml
```diff
@@ -42,71 +51,70 @@ parthenon-cuda-short:
...
     when: always
     expire_in: 3 days
     paths:
       - build-cuda-perf/CMakeFiles/CMakeOutput.log
-      - tst/regression/test_suites/advection_convergence/output/advection-errors.dat
+      - build-cuda-perf-uvm/tst/regression/outputs/advection_convergence/advection-errors.dat
 
-# run short suite on CPUs
-parthenon-cpu-performance:
+# run performance and regrssion on CPUs
```
Yurlungur:
```suggestion
# run performance and regression on CPUs
```

example/advection/advection_package.cpp
```diff
@@ -38,16 +39,99 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
...
+  if (profile_str.compare("wave") == 0) {
+    profile = 0;
+  } else if (profile_str.compare("smooth_gaussian") == 0) {
+    profile = 1;
+  } else if (profile_str.compare("hard_sphere") == 0) {
+    profile = 2;
+  } else {
+    PARTHENON_FAIL("Unknown profile in advection example: " + profile_str);
+  }
+  pkg->AddParam<>("profile", profile);
```
Yurlungur: Ah I see what you're up to here. Looks good overall. Can we name these profile variables though? Maybe with an enum or an enum class?
```C++
enum class InitialProfile { unknown = -1; wave = 0; gaussian = 1; sphere = 2; };
```

src/utils/error_checking.hpp
Yurlungur: make sure to merge master in, since this has been fixed in master now.

.gitlab-ci.yml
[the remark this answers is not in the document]
pgrete: Oh, definitely. That's exactly what I meant (we'll definitely do a debug build, but just run the unit tests in those and then for the regression tests [which take more time >few sec] we run in Release to get optimization)

example/advection/advection_package.cpp
pgrete: The current version is now directly using the string and no intermediate variable (maybe the browser cached an old version of the PR as this is also already updated to master).

.gitlab-ci.yml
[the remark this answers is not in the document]
Yurlungur: Cool.

example/advection/advection_package.cpp
Yurlungur: Oops my bad. Yeah I must have been reviewing an older version.

pgrete:
@par-hermes format

pgrete:
@JoshuaSBrown have I addressed all your comments? Once the pipeline passes this is good to go from my point of view.

.gitlab-ci.yml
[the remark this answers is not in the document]
JoshuaSBrown: This is fine for now.

tst/regression/CMakeLists.txt
```diff
@@ -49,6 +50,13 @@ list(APPEND TEST_DIRS calculate_pi)
 list(APPEND TEST_ARGS "--driver ${CMAKE_BINARY_DIR}/example/calculate_pi/pi-example \
 --driver_input ${CMAKE_CURRENT_SOURCE_DIR}/test_suites/calculate_pi/parthinput.regression")
```
JoshuaSBrown, line 54: I would have preferred a second list to specify the number of MPI_PROCS, e.g. something like:

```
list(APPEND TEST_MPI_PROCS 1)
```

That would be passed to setup_test_mpi, in the for loop below so you are not having to call setup_test and setup_test_mpi outside of the for loop. The HDF5 case I understand though. This is more of a stylistic preference though so non blocking.

src/mesh/domain.hpp
```diff
@@ -28,11 +28,8 @@
 namespace parthenon {
 
 struct IndexRange {
-  IndexRange() {}
-  IndexRange(int start, int end) : s(start), e(end) { assert(e >= s); }
   int s = 0; /// Starting Index (inclusive)
   int e = 0; /// Ending Index (inclusive)
-  KOKKOS_INLINE_FUNCTION int ncells() const noexcept { return e - s + 1; }
```
JoshuaSBrown, line 35: Why did you want to remove this?

tst/regression/CMakeLists.txt
pgrete: Yeah, I think we need to address the overall logic and what we want to run where and when at some point anyway and can fix is then.
For now I just made it this here "does the job".

pgrete:
Got two approvals. Pulling the trigger.

PR #168, opened 2020-05-15
Fix printf in kernels

pgrete:
## PR Summary

`fprintf` is not available on devices. Thus, we need to use `printf` (and cannot print to stderr).

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/src/utils/error_checking.hpp
+++ b/src/utils/error_checking.hpp
@@ -51,19 +51,16 @@ namespace ErrorChecking {
 KOKKOS_INLINE_FUNCTION
 void require(std::string const &condition, std::string const &message,
              std::string const &filename, int const linenumber) {
-  fprintf(stderr,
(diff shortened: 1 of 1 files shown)


pgrete:
The compiler didn't shout at me for that.
But I noticed that we also need to fix the `exit` calls as they are also not legal from within a kernel (though at least the code compiles whereas the printf resulted in compilation failure when using `-G`).

Yurlungur:
> Not really, but similarly to the `exit` calls, which are also not legal from within a kerne, the code compiles whereas the printf resulted in compilation failure when using `-G`.
> But we should definitely address this.

Hmm. We could replace them with `char[]` but it doesn't seem urgent.

Yurlungur:
@pgrete , I believe @brryan is on leave at the moment, so he might not respond to this. We can just go ahead and merge when you're ready.

PR #170, opened 2020-05-18
Joshua s brown/split container iterator test into performance and unit components

JoshuaSBrown:
## PR Summary

Because the container_iterator unit test had a performance and unit component these two parts were split. The performance component was moved to a separate performance folder. This will prevent the unit testing run time from becoming excessive.  

This partly addresses [issue](https://github.com/lanl/parthenon/issues/158).

## PR Checklist

- [x] Code passes cpplint
- [x] New features are documented
- [x] Code is formatted


tst/performance/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+## Alamos National Laboratory (LANL), which is operated by Triad National Security, LLC
+## for the U.S. Department of Energy/National Nuclear Security Administration. All rights
+## in the program are reserved by Triad National Security, LLC, and the U.S. Department
+## of Energy/National Nuclear Security Administration. The Government is granted for
+## itself and others acting on its behalf a nonexclusive, paid-up, irrevocable worldwide
+## license in this material to reproduce, prepare derivative works, distribute copies to
+## the public, perform publicly and display publicly, and to permit others to do so.
+##========================================================================================
+
+list(APPEND performance_tests_SOURCES
```
AndrewGaspar, line 19: If nobody's going to iterate over this variable, I would just change this to

```cmake
add_executable(performance_tests)
target_sources(
    performance_tests
    PRIVATE
        test_container_iterator.cpp
)
```

tst/performance/test_container_iterator.cpp
```diff
@@ -0,0 +1,282 @@
...
+    par_for(
+        "Initialize ", DevExecSpace(), 0, Nvar - 1, 0, N - 1, 0, N - 1, 0, N - 1,
+        KOKKOS_LAMBDA(const int l, const int k, const int j, const int i) {
+          raw_array(l, k, j, i) =
+              static_cast<Real>((l + 1) * (k + 1) * (j + 1) * (i + 1));
+        });
+  };
+
+  // Test performance iterating over variables (we should aim for this performance)
+  double time_raw_array = performance_test_wrapper(n_burn, n_perf, init_raw_array, [&]() {
```
AndrewGaspar, line 102: How easy would it be to, essentially, replace these `performance_test_wrapper` calls with Catch2's benchmarking macros: https://github.com/catchorg/Catch2/blob/master/docs/benchmarks.md#top

AndrewGaspar: You'd need to make sure to keep the fences, but you could get rid of the timers.

JoshuaSBrown: > How easy would it be to, essentially, replace these `performance_test_wrapper` calls with Catch2's benchmarking macros

I will look into it next week.

pgrete: > You'd need to make sure to keep the fences, but you could get rid of the timers.

Speaking of fences, I'm wondering if we actually even need more fences and/or add an option to fence between the executions within the `n_perf` loop (basically allowing to include the individual launch overhead in the timing or not).

AndrewGaspar: It would be useful as an option! But we definitely want to have a performance test that minimizes the overhead.

tst/performance/test_container_iterator.cpp
```diff
@@ -0,0 +1,286 @@
...
+
+template <typename InitFunc, typename PerfFunc>
+void performance_test_wrapper(const std::string test_name, InitFunc init_func,
+                                PerfFunc perf_func) {
+
+  BENCHMARK_ADVANCED(test_name.c_str())(Catch::Benchmark::Chronometer meter) {
+    init_func();
+    Kokkos::fence();
+    meter.measure([&]() { return perf_func(); });
+    Kokkos::fence();
```
JoshuaSBrown, line 66: Something like this:

```
init_func();
Kokkos::fence();
meter.measure([&]() { perf_func(); Kokkos::fence(); });
```

JoshuaSBrown:
@par-hermes format

JoshuaSBrown:
@pgrete So this has been reworked, unfortunately I cannot figure out how to get any of the performance metrics out of catch, so I cannot calculate the ratios that you had in your first implementation. Let me know what you think.

JoshuaSBrown:
@AndrewGaspar were they some things you wanted me to change in this pull request? 

[kernel benchmarking](https://github.com/lanl/parthenon/issues/198)

CMakeLists.txt
```diff
@@ -161,7 +162,11 @@ else()
 endif()
 
 # Build Tests and download Catch2
-if (${ENABLE_UNIT_TESTS} OR ${ENABLE_INTEGRATION_TESTS} OR ${ENABLE_REGRESSION_TESTS})
+if (${ENABLE_UNIT_TESTS} OR ${ENABLE_INTEGRATION_TESTS} OR ${ENABLE_REGRESSION_TESTS} OR ${ENABLE_PERFORMANCE_TESTS})
+
+  if (${ENABLE_PERFORMANCE_TESTS})
+    add_definitions(-DCATCH_CONFIG_ENABLE_BENCHMARKING)
```
AndrewGaspar, line 168: Instead of this, you should hang this off of `catch2_define` in `tst/CMakeLists.txt`:
```cmake
if (ENABLE_PERFORMANCE_TESTS)
    target_compile_definitions(catch2_define PUBLIC CATCH_CONFIG_ENABLE_BENCHMARKING)
endif()
```

AndrewGaspar:
Did you intentionally update the Kokkos submodule? I'd say this should only be done if updating it was necessary, and otherwise it should be done in a separate PR.

JoshuaSBrown:
> Did you intentionally update the Kokkos submodule?

That would be a no, glad you caught it.

tst/performance/CMakeLists.txt
JoshuaSBrown: Done.

tst/performance/test_container_iterator.cpp
JoshuaSBrown: We can create an [issue](https://github.com/lanl/parthenon/issues/198) for this.

CMakeLists.txt
JoshuaSBrown: Done.

CMakeLists.txt
```diff
@@ -174,7 +176,6 @@ if (${ENABLE_UNIT_TESTS} OR ${ENABLE_INTEGRATION_TESTS} OR ${ENABLE_REGRESSION_T
       # Clone the submodule
       execute_process(COMMAND git submodule update --init --force -- external/Catch2 WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
     endif()
```
AndrewGaspar, line 178: Put the newline back.

```suggestion
    endif()

```

JoshuaSBrown:
@par-hermes format

JoshuaSBrown:
@par-hermes format

AndrewGaspar:
lol it looks like catch2 and cpplint are fighting?

AndrewGaspar:
@JoshuaSBrown I pushed a fix.

JoshuaSBrown:
> @JoshuaSBrown I pushed a fix.

Thanks! For the life of me I could not figure out what was going on.

tst/performance/test_container_iterator.cpp
```diff
@@ -59,7 +59,8 @@ static const int N_kernels_to_launch_per_test = 100;
 template <typename InitFunc, typename PerfFunc>
 void performance_test_wrapper(const std::string test_name, InitFunc init_func,
                               PerfFunc perf_func) {
-  BENCHMARK_ADVANCED(test_name.c_str())(Catch::Benchmark::Chronometer meter) {
+  auto test_name_c = test_name.c_str();
+  BENCHMARK_ADVANCED(test_name_c)(Catch::Benchmark::Chronometer meter) {
```
AndrewGaspar, line 63: Oh, this change wasn't actually necessary - I was trying to figure out if this was confusing cpplint.

JoshuaSBrown:
@par-hermes format

JoshuaSBrown:
@pgrete ping.

JoshuaSBrown:
Thanks Phil! As always super helpful. 

>     1. The Kokkos abstraction unit test also includes a "performance" test (Overlapping SpaceInstances).
>        I'm still undecided on how to handle this given that for that test we definitely want to compare the results beyond benchmarking and have the test fail if the overlapping instances are not faster (i.e., so it's kind of a unit test that addresses performance).
>        At least what we should make the labels matching (here, `performance_test` is used whereas only `performance` is used for the abstraction.

I like your suggestion on the labels. We should probably discuss these further. I will create an issue. 

>     Similarly, I'm conceptually wondering to which degree we can enforce the strict separation of the tests, e.g., the test I consider "performance unit" above or once we start tracking "integrated performance" (e.g., of the advection example) it'd more be a "performance regression" test.

I think the key point is to move anything out of the unit tests, that will take more than 5s to run. Strict separation is not so important I think. I just don't want to get to the point that people cannot easily run unit tests because they start taking non negligible amounts of time.

tst/CMakeLists.txt
```diff
@@ -10,6 +10,11 @@ if(${ENABLE_INTEGRATION_TESTS})
   message(STATUS "Building integration tests.")
   add_subdirectory(integration)
 endif()
+if(${ENABLE_PERFORMANCE_TESTS})
```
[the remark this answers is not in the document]
JoshuaSBrown: That's a good point!

tst/performance/CPPLINT.cfg
```diff
@@ -0,0 +1,15 @@
...
+# for the U.S. Department of Energy/National Nuclear Security Administration. All rights
+# in the program are reserved by Triad National Security, LLC, and the U.S. Department
+# of Energy/National Nuclear Security Administration. The Government is granted for
+# itself and others acting on its behalf a nonexclusive, paid-up, irrevocable worldwide
+# license in this material to reproduce, prepare derivative works, distribute copies to
+# the public, perform publicly and display publicly, and to permit others to do so.
+#=========================================================================================
+
+filter=-readability/braces # BENCHMARK macro is misidentified as function when it's
+                           # instead a lambda
```
Yurlungur, line 15:
```suggestion
                           # instead a lambda

```

Issue #171, opened 2020-05-20
Replace `std::string` and `std::exit` in `error_checking.hpp`


Yurlungur:
Strings become `char[]` and `std::exit` should become `Kokkos::abort()`

PR #172, opened 2020-05-21
kokkosify error checking macros

Yurlungur:
## PR Summary

As discussed in issue #171 I replace `std::string` and `exit` with `char[]` and `Kokkos::abort`. This functionality should actually work on GPUs and should be performance portable thanks to the `Kokkos::abort` call.

Some things to note:
- I actually use `const char * const`, not `char[]`. I copied this from the implementation of `Kokkos::abort`. I suppose the extra const is more clear but `const char[]` would almost certainly be fine.
- The places in the code calling `PARTHENON_ABORT` and `PARTHENON_REQUIRE` used C++ strings. I had to modify them slightly to pass in a C-style string.
- I added a unit test that tests the error-checking. Kokkos abort, however, obviously kills your program. So the test fails. Therefore, I guard the test in an ifdef. To activate it, configure with `cmake -DTEST_ERROR_CHECKING`. You'll get a warning telling  you your test will fail. And it will.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/tst/unit/test_error_checking.cpp
+++ b/tst/unit/test_error_checking.cpp
@@ -1,0 +1,45 @@
+//========================================================================================
+// Parthenon performance portable AMR framework
+// Copyright(C) 2020 The Parthenon collaboration
+// Licensed under the 3-clause BSD License, see LICENSE file for details
+//========================================================================================
+// (C) (or copyright) 2020. Triad National Security, LLC. All rights reserved.
+//
+// This program was produced under U.S. Government contract 89233218CNA000001
+// for Los Alamos National Laboratory (LANL), which is operated by Triad
+// National Security, LLC for the U.S. Department of Energy/National Nuclear
+// Security Administration. All rights in the program are reserved by Triad
+// National Security, LLC, and the U.S. Department of Energy/National Nuclear
+// Security Administration. The Government is granted for itself and others
+// acting on its behalf a nonexclusive, paid-up, irrevocable worldwide license
+// in this material to reproduce, prepare derivative works, distribute copies to
+// the public, perform publicly and display publicly, and to permit others to do
+// so.
+//========================================================================================
+
+#include <catch2/catch.hpp>
+
+#include "utils/error_checking.hpp"
+
+#ifdef PARTHENON_TEST_ERROR_CHECKING
+
+TEST_CASE("Parthenon Error Checking","[ErrorChecking][Kokkos]") {
+  SECTION("PARTHENON_REQUIRE passes if condition true") {
+    PARTHENON_REQUIRE(true, "This shouldn't fail");
+  }
+  SECTION("PARTHENON_REQUIRE fails if condition false") {
+    PARTHENON_REQUIRE(false, "This should fail");
+  }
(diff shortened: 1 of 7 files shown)


Yurlungur:
@brryan probably will not see this, since he's on leave. So we should merge without his approval.

tst/unit/test_error_checking.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+#include <catch2/catch.hpp>
+
+#include "utils/error_checking.hpp"
+
+TEST_CASE("Parthenon Error Checking", "[ErrorChecking][Kokkos]") {
+  SECTION("PARTHENON_REQUIRE passes if condition true") {
+    PARTHENON_REQUIRE(true, "This shouldn't fail");
+  }
+#ifdef PARTHENON_TEST_ERROR_CHECKING
```
AndrewGaspar, line 28: ah, too bad there isn't anything equivalent AFAICT to gtest's `ASSERT_DEATH`: https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#death-tests

tst/unit/test_error_checking.cpp
```diff
@@ -0,0 +1,40 @@
...
+  SECTION("PARTHENON_REQUIRE fails if condition false") {
+    PARTHENON_REQUIRE(false, "This should fail");
```
AndrewGaspar, line 30: Do you not need to comment out each of these sections to check the subsequent one?

tst/unit/test_error_checking.cpp
Yurlungur: I know. I was frustrated by that.

tst/unit/test_error_checking.cpp
Yurlungur: You do. I don't know a way around that unfortunately. 

One possibility I guess is I could give them all different tags and then run them or not via `ctest` rather than using the `ifdef`. Not sure which is better.

AndrewGaspar: idk, unless you expect people are going to run this on occassion, I'd say you should just delete this code.

What you could potentially do is just create a non-catch2 executable that causes the failure, and then manually call `add_test`, followed by setting this test property: https://cmake.org/cmake/help/latest/prop_test/WILL_FAIL.html#prop_test:WILL_FAIL

Yurlungur: > I'd say you should just delete this code.

I wanted it in as examples of use. A bit of documentation. I'll delete it if you think I should though.

Yurlungur:
> Just to check - do we know what happens on the host when a GPU kernel fails? Does it get propagated as a host abort or something similar?

:+1: just checked. It fails with a `SIGABRT`.

tst/unit/test_error_checking.cpp
AndrewGaspar: I'll leave it up to you.

Yurlungur: :+1: I'll keep it then.

pgrete:
Looks good to me. Thanks for taking care of it.

Issue #175, opened 2020-05-23
Enforce warning free compilation?


pgrete:
While working on Kokkosizing the codebase I got the impression that the majority of changes we recently introduced were not tested if they compile with nvcc (which resulted in lots of compiler warnings when using nvcc and code that is potentially incompatible with GPUs).

Thus, I suggest that once(/at the latest) we have a fully working Kokkosize codebase we adapt the CI tests so that they fail if there are compiler warnings (similar to the style checks).
This should help to catch/prevent bugs for functionality that may currently not being used (i.e., tested) but may be done in the future, see also #174

PR #177, opened 2020-05-27
Kokkosification of AMR

jdolence:
## PR Summary

First pass at Kokkosification of the AMR components of the code.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


jdolence:
Advection example with AMR seems to work now on GPUs.  2D runs and looks correct.  3D runs, but I haven't checked the answers.

pgrete:
@jdolence I noticed that you started to work on a thread manager.
Any reason you picked that approach over using the Kokkos::PartitionManager (IIRC the plan was to port what we did in K-Athena) ?

AndrewGaspar:
@par-hermes format

pgrete:
@jdolence and @AndrewGaspar what's the status here?
As far as I can tell this looks pretty close to being done. Shall I go ahead for a detailed review and/or fix things I see directly, e.g., figuring out what goes wrong with the pipeline?

jdolence:
@pgrete @AndrewGaspar I think this is pretty close, but it's unsettling that the pipeline failed.  I haven't been able to dig in to that yet, except to see that it had something to do with the pi regression test failing.  That test runs fine on my local machine, so it might take a little digging to understand what's going wrong.

AndrewGaspar:
Yeah, I'm currently trying to reproduce the issue. Want to really get at the meat of it tomorrow but getting stuck on getting my environment set up now that we require numpy...

AndrewGaspar:
One thing I've discovered is that the issue only occurs if OpenMP is enabled for Parthenon, even if it's disabled for Kokkos.

AndrewGaspar:
The configuration of having OpenMP enabled for Parthenon but not for Kokkos seems like it's not supported in general. Regardless, OpenMP must be enabled for the issue to reproduce.

AndrewGaspar:
Pushed a fix, but I think we should have a discussion on how we want to do loop-level parallelism when doing host-parallel builds. See the Matrix channel.

src/mesh/mesh.cpp
```diff
@@ -1149,83 +1132,64 @@ void Mesh::Initialize(int res_flag, ParameterInput *pin) {
     }
 
     if (res_flag == 0) {
-#pragma omp parallel for num_threads(nthreads)
-      for (int i = 0; i < nmb; ++i) {
-        MeshBlock *pmb = pmb_array[i];
-        pmb->ProblemGenerator(pin);
-      }
+      parallel_for("Problem Generator", block_range,
```
forrestglines, line 1135: @AndrewGaspar I see now - using `parallel_for` will launch `ProblemGenerator` inside a kernel. However, that won't work on the GPU, since `ProblemGenerator` will launch its own kernels and `*pin` is a host pointer, not a device pointer.

I think the Kokkos partition paradigm - which would give us CPU threading+GPU streams - is more appropriate here and is closer to the original intent.

src/mesh/mesh.cpp
```diff
@@ -1149,83 +1132,64 @@ void Mesh::Initialize(int res_flag, ParameterInput *pin) {
...
+                   [&](int const i) { pmb_array[i]->ProblemGenerator(pin); });
     }
 
-    int call = 0;
-    // Create send/recv MPI_Requests for all BoundaryData objects
-#pragma omp parallel for num_threads(nthreads)
+    parallel_for("Create send/recv MPI_Requests for all BoundaryData objects",
```
forrestglines, line 1139: Same goes for these loops. MPI calls won't work from inside kernels. 

However, I think this function is called only once, so it's not a priority to optimize.

src/mesh/mesh.cpp
AndrewGaspar: Maybe I accidentally obfuscated it, but these `parallel_for` calls target the `Kokkos::DefaultHostExecutionSpace`.

src/mesh/mesh.cpp
AndrewGaspar: See previous comment.

src/mesh/mesh.cpp
forrestglines: I see, I missed that bit. That makes more sense

pgrete:
I made a couple of changes and the tests now pass again (and are way faster than before though proably still too long for a "short" test -- but this is a different issue).

The key changes were:
- removed the const () operator from the Variable class (as we discourage direct use of Vars in kernels and all the warning associated with that missing operator are not gone in master since #202 was merged)
- reverted the threading related changes in the mesh initialization (that actually fixed the regression tests). I'll address issue around threading in #190 

I think we still need more and/or move some fences (for multithreaded MPI runs) but from my point of view this PR should focus on getting AMR fully working on GPUs, which it does now. Threading/MPI will be addresses in more detail in #190.

Other question I had looking at the changes: 
- Should we remove the CheckRefinement function from the CalcPi example? The function is currently not used so I'm wondering whether it may be confusing to keep it as an example.
- The new `-G` flag conflicts with other default options: `ptxas warning : Conflicting options --device-debug and --generate-line-info specified, ignoring --generate-line-info option` and `nvcc warning : '--device-debug (-G)' overrides '--generate-line-info (-lineinfo)'`. What do you want to do about it? In spirit of #175 I'm in favor of reducing the warnings as much as possible.
- This similarly applies to `/builds/pgrete/parthenon/src/driver/driver.hpp(77): warning: function "parthenon::Driver::PostExecute()" is hidden by "parthenon::EvolutionDriver::PostExecute" -- virtual function override intended?`
- and `/builds/pgrete/parthenon/src/interface/variable_pack.hpp(62): warning: calling a __host__ function from a __host__ __device__ function is not allowed
          detected during instantiation of "int parthenon::VariablePack<T>::GetDim(int) [with T=parthenon::Real]" ` which both are not related to this PR.


PS: on a different machine `-G` actually results in a failed build
``
nvcc warning : '--device-debug (-G)' overrides '--generate-line-info (-lineinfo)'
/home/pgrete/src/parthenon/src/driver/driver.hpp(77): warning: function "parthenon::Driver::PostExecute()" is hidden by "parthenon::EvolutionDriver::PostExecute" -- virtual function override intended?

ptxas warning : Conflicting options --device-debug and --generate-line-info specified, ignoring --generate-line-info option
ptxas /tmp/tmpxft_0002469e_00000000-5_advection_package.ptx, line 40150; error   : Alignment of argument does not match formal parameter '_'
ptxas /tmp/tmpxft_0002469e_00000000-5_advection_package.ptx, line 40199; error   : Alignment of argument does not match formal parameter '_'
ptxas fatal   : Ptx assembly aborted due to errors
make[2]: *** [example/advection/CMakeFiles/advection-example.dir/build.make:87: example/advection/CMakeFiles/advection-example.dir/advection_package.cpp.o] Error 255
make[1]: *** [CMakeFiles/Makefile2:1643: example/advection/CMakeFiles/advection-example.dir/all] Error 2
make: *** [Makefile:141: all] Error 2
```

File: example/calculate_pi/calculate_pi.cpp [hunk 3]
```diff
@@ -149,18 +149,18 @@ TaskStatus ComputeArea(MeshBlock *pmb) {
...
                                              {kb.e + 1, jb.e + 1, ib.e + 1},
                                              {1, 1, ib.e + 1 - ib.s}),
-      KOKKOS_LAMBDA(int k, int j, int i, Real &area) {
-        area += v(k, j, i) * coords.Area(parthenon::X3DIR, k, j, i);
+      KOKKOS_LAMBDA(int k, int j, int i, Real &larea) {
+        larea += v(k, j, i) * coords.Area(parthenon::X3DIR, k, j, i);
       },
-      Kokkos::Sum<Real, parthenon::DevMemSpace>(v.Get(0, 0, 0, 0, 0, 0)));
+      area);
+  Kokkos::deep_copy(pmb->exec_space, v.Get(0, 0, 0, 0, 0, 0), area);
```
AndrewGaspar, line 163: Oh, I intentionally didn't do this so that it wouldn't fence here. Why did this need to change? I'm having a hard time understanding what the issue was.

AndrewGaspar:
> Should we remove the CheckRefinement function from the CalcPi example? The function is currently not used so I'm wondering whether it may be confusing to keep it as an example.

I think @jdolence wanted to keep it for demonstration purposes.

AndrewGaspar:
> The new -G flag conflicts with other default options: ptxas warning : Conflicting options --device-debug and --generate-line-info specified, ignoring --generate-line-info option and nvcc warning : '--device-debug (-G)' overrides '--generate-line-info (-lineinfo)'. What do you want to do about it? In spirit of #175 I'm in favor of reducing the warnings as much as possible.

Hm... I didn't see this with cuda 10.1. I would like to be able to debug device kernels, so I'd like to keep `-G`. Looking to see if I can reproduce...

AndrewGaspar:
> ptxas /tmp/tmpxft_0002469e_00000000-5_advection_package.ptx, line 40150; error : Alignment of argument does not match formal parameter ''
ptxas /tmp/tmpxft_0002469e_00000000-5_advection_package.ptx, line 40199; error : Alignment of argument does not match formal parameter ''
ptxas fatal : Ptx assembly aborted due to errors
make[2]: *** [example/advection/CMakeFiles/advection-example.dir/build.make:87: example/advection/CMakeFiles/advection-example.dir/advection_package.cpp.o] Error 255
make[1]: *** [CMakeFiles/Makefile2:1643: example/advection/CMakeFiles/advection-example.dir/all] Error 2
make: *** [Makefile:141: all] Error 2

Aaah, this is too bad. I was warned on the Kokkos slack that sometimes `-G` can crash the compiler, which is why `Kokkos::kokkos` doesn't automatically apply it for debug builds.

AndrewGaspar:
We may have to only selectively apply `-G`. :(

example/calculate_pi/calculate_pi.cpp
pgrete: There was no real "issue" (the result for the specific test didn't change), but conceptually I was confused by the original deep_copy to to `v` and then using `v` in the reduction loop (i.e., if `kb.s`, `jb.s`, and `ib,s` would be 0 then the original value stored in `v(0,0,0)` would be lost for the area summation (independent of he final reduction to the view).
Given that this should serve as an example, I think the current version is easier to follow.

pgrete:
> > Should we remove the CheckRefinement function from the CalcPi example? The function is currently not used so I'm wondering whether it may be confusing to keep it as an example.
> 
> I think @jdolence wanted to keep it for demonstration purposes.

The (properly Kokkosized) `CheckRefinement` example is still present in the advection example, where it is also used and I agree that we should definitely keep it there.

Given that the calc pi example demonstrate to use the Parthenon default refinement criteria keeping the function there may potentially be confusing (as it's not used).

pgrete:
> We may have to only selectively apply `-G`. :(

From my point of view, `-G` should not be a default option in Parthenon as long as we cannot guarantee that it'll work out of the box on any (or, let's say many) developer/user system.
We could either add a "`GPU_DEBUG`" option that by default is disabled or add a note to the docs about adding `-G` to the CMAKE_CXX_FLAGS (as it takes precedence over the lineinfo option).
I'm in favor of the latter as `-G` is sth I expect to be used on the more advanced developer site where the developer is aware of that option.
In the majority of my cases the standard `lineinfo`in Debug mode was enough for debugging/pinpointing an issue.

pgrete:
@jdolence @AndrewGaspar what's your take on removing the `-G` and the `CheckRefinement` function in the calcpi example?
And is there anything else that we should address here?
From my point of view this is looks otherwise good to go.

jdolence:
> @jdolence @AndrewGaspar what's your take on removing the `-G` and the `CheckRefinement` function in the calcpi example?
> And is there anything else that we should address here?
> From my point of view this is looks otherwise good to go.

I think this is fine to go forward.  I was waiting to hear from @AndrewGaspar before taking off WIP and trying to get merged in.

pgrete:
In anticipation of @AndrewGaspar reply I already removed those two piece from this PR so that it can be merged once we remove the WIP flag.

AndrewGaspar:
Removed WIP.

PR #180, opened 2020-05-28
Ensure that Parthenon build parameters set by host codes are evidently parthenon

nmsriram:
## PR Summary
Exposed Parthenon build parameters that host codes might want to modify.  These were also prefixed by a PARTHENON_ .  These include:

1. PARTHENON_DISABLE_MPI
1. PARTHENON_DISABLE_HDF5
1. PARTHENON_DISABLE_OPENMP
1. PARTHENON_NGHOST
1. PARTHENON_SINGLE_PRECISION

## PR Checklist
- [x] Code is formatted
- [x] Changes are documented

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,8 +14,10 @@
 cmake_minimum_required(VERSION 3.10)
 project(parthenon LANGUAGES C CXX)
 
-# Options
 include(CTest)
+
+# Options
+option(NUMBER_GHOST_CELLS "Change number of ghost cells" 2)
 option(ENABLE_UNIT_TESTS "Enable unit tests" ${BUILD_TESTING})
 option(ENABLE_INTEGRATION_TESTS "Enable integration tests" ${BUILD_TESTING})
 option(ENABLE_REGRESSION_TESTS "Enable regression tests" ${BUILD_TESTING})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,20 +16,34 @@ project(parthenon LANGUAGES C CXX)
 
 include(CTest)
 
-# Options
-option(NUMBER_GHOST_CELLS "Change number of ghost cells" 2)
+# Compile time constants
+set(PARTHENON_NGHOST 2 CACHE STRING "Number of ghost cells") 
+
+# Compile Options
+option(PARTHENON_SINGLE_PRECISION "Run in single precision" 0) 
+option(PARTHENON_DISABLE_MPI "MPI is enabled by default if found, set this to True to disable MPI" OFF)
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -14,21 +14,37 @@
...
 project(parthenon LANGUAGES C CXX)
 
-# Options
 include(CTest)
+
+# Compile time constants
+set(PARTHENON_NGHOST 2 CACHE STRING "Number of ghost cells") 
+
+# Compile Options
+option(PARTHENON_SINGLE_PRECISION "Run in single precision" 0)
```
Yurlungur, line 23: Does this parameter do anything? Or are we double everywhere?

CMakeLists.txt
```diff
@@ -14,21 +14,37 @@
 cmake_minimum_required(VERSION 3.10)
 project(parthenon LANGUAGES C CXX)
 
-# Options
 include(CTest)
+
+# Compile time constants
+set(PARTHENON_NGHOST 2 CACHE STRING "Number of ghost cells")
```
Yurlungur, line 20: Is now the time to just make this an option?
```suggestion
option(PARTHENON_NGHOST "Number of ghost cells" 2)
```

pgrete: What's the exact difference (asking out of curiosity and not because I think we should use one or the other)? I tried google but it seemed that it mostly made a difference for boolean variables (as they'd allow dependencies for `option` but not for `set`)?

pgrete:
I also noticed that these options are not documented yet. If you don't mind it'd be great to add those variables, e.g., to the build doc.

CMakeLists.txt
nmsriram: The problem with option is that it is boolean, so it is only supposed to turn ghosts on or off.  We shouldn't use an option to set a specific value.  Hence not an option.

nmsriram:
@pgrete I added a 'changes are documented' bullet to the MR and will check that when I'm done with documentation changes.  Thank you for catching that.

nmsriram:
> Thanks for taking care of this. It definitely makes the interface cleaner.
> I picked "requesting" changes as the new variables are not reflected in the CI file yet (and I don't know how to add suggestions here for a file that is currently not part of the PR).

Good point.  I will make changes requested.

nmsriram:
@pgrete I've made changes to the .gitlab-ci.yml file to reflect the new Cmake variables.  I've also documented the variables in [docs/building.md](https://github.com/lanl/parthenon/blob/933855ea83d96e44c535418ee9fce0db125fc99b/docs/building.md)

PR #186, opened 2020-06-02
Fix index shape

JoshuaSBrown:
## PR Summary

Small pull request to fix kokkos errors introduced by the IndexShape class. This is to help address this [issue](https://github.com/lanl/parthenon/issues/174)

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


JoshuaSBrown:
> I'm sad to see `GetBounds` go, but since we're not using it, it's right to remove. Thanks for resolving this issue.

I think that it can be added back without the array, but as a struct and would be fine, but might as well wait until it's needed.

Yurlungur:
> I think that it can be added back without the array, but as a struct and would be fine, but might as well wait until it's needed.

Agreed.

AndrewGaspar:
@pgrete Any thoughts? We'll merge if we don't hear anything by the end of the day.

PR #187, opened 2020-06-02
Joshua s brown/add code coverage

JoshuaSBrown:
## PR Summary

This pull request adds code coverage. It is needed to help find holes in the test suite. 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


.gitlab-ci.yml
```diff
@@ -22,10 +22,12 @@ cache:
 
 variables:
   GIT_SUBMODULE_STRATEGY: recursive
+  CODECOV_TOKEN: f3729c36-5ce7-48a3-b940-92f1cefd85a5
```
AndrewGaspar, line 25: I'm guessing this lets anybody with this token upload code coverage statistics for our repo? It's not a huge deal (seems like pretty petty trolling), but we should probably generate a new token and put it in GitLab's secret variables when this is ready to merge.

.gitlab-ci.yml
```diff
@@ -183,3 +185,25 @@ parthenon-cpu-performance_and_regression-mpi:
...
+      -DKokkos_ENABLE_OPENMP=True -DKokkos_ARCH_WSM=True
+      -DMPIEXEC_PREFLAGS="--allow-run-as-root" -DCODE_COVERAGE=ON
+      ../ && make -j${J} && make coverage && make coverage-upload
+  artifacts:
+    when: always
+    expire_in: 3 days
+    paths:
+      - build-debug-coverage/CMakeFiles/CMakeOutput.log
+  only:
+    - master
```
AndrewGaspar, line 207: Is there any way to get code coverage for every MR? Too expensive?

JoshuaSBrown: Sure.

JoshuaSBrown:
This needs to be merged first, due to the Container iterator test timing out. https://github.com/lanl/parthenon/pull/170

JoshuaSBrown:
@AndrewGaspar I could use some input on the best way to set this up. Our regression tests cover a lot more code but running them with code coverage would slow them down a lot. To get an idea of the code coverage it would make sense to run the regression tests but with significantly reduced number of iterations.

AndrewGaspar:
> To get an idea of the code coverage it would make sense to run the regression tests but with significantly reduced number of iterations.

Yeah, that's what I had in mind, too. @jdolence @pgrete Do you think we'd get virtually all the lines of code that the regression tests execute by just running the first couple iterations of the regression tests? Or would we need samples from different stages of the test?

JoshuaSBrown:
ping @jdolence @pgrete

jdolence:
I certainly think we can get away with far fewer cycles.  At a minimum, we should see how many cycles are required in the current advection test to trigger AMR.  Alternatively (if that's still too expensive), we might modify the test to make sure it triggers AMR in the first few cycles.  So the homework is to identify when AMR is triggered and evaluate whether or not we can run it that long with the coverage tools.

pgrete:
I agree. The number of samples/cycles should be irrelevant as long as we cover all features (refine, derefine, PostExecute, output, etc).

JoshuaSBrown:
Sounds good!

.gitlab-ci.yml
JoshuaSBrown: Are you meaning if not merging to master but some other branch?

JoshuaSBrown:
@AndrewGaspar should I only be looking at coverage of files in src, and ignoring the tst folder?

AndrewGaspar:
@JoshuaSBrown Yeah, I think that's the right thing to do.

JoshuaSBrown:
@AndrewGaspar @pgrete @jdolence this is ready to be merged. The coverage is at ~ 50 % for files in /src and /examples. One of the largest gaps I think has to do with fc values which can be solved by creating a test use the face fields example. https://github.com/lanl/parthenon/issues/217

JoshuaSBrown:
@pgrete thanks for the catches. 

> WARNING: build-cuda-perf-mpi/tst/regression/outputs/advection_convergence/advection-errors.png: no matching files <- for the MPI test the png is not generated any more and the dat file is also missing some lines

It looks like the python regression test script was deleting the output folder before running each case, with mpi, without mpi, with coverage etc, I will fix this by putting the output in separate folders. 

> (Maybe related) I'm not sure I fully follow the logic behind the MPI switch for the coverage test. Is my understanding correct that coverage is only run with MPI and if MPI is not found there's no coverage report? Given that some of the logic (especially the boundaries) depend on whether MPI is on or off, and whether the MeshBlocks are on the same rank or not, we could increase coverage by including those in the test, too.

Good point will add. 

> We could increase coverage by adding coverage also to the 1D and 3D advection test as we have some branches in the code that are not touched if nx3 != 1, or nx2=1 and nx3=1.

I could use some help with this right now, I have this section of logic enabled with coverage:

```
       if step <= n_res:                                                                                                           
                                                                                                                                    
            parameters.coverage_status = "both"                                                                                     
                'parthenon/mesh/nx1=%d' % lin_res[step % n_res -1],                                                                 
                'parthenon/meshblock/nx1=%d' % lin_res[step % n_res -1],                             
                'parthenon/mesh/nx2=1',                                                                                             
                'parthenon/meshblock/nx2=1',                                                                                        
                'parthenon/mesh/nx3=1',                                                                                             
                'parthenon/meshblock/nx3=1',                                                                                        
                'Advection/vy=0.0',                                                                                                 
                'Advection/vz=0.0',                                                                                                 
                ]
```

But apparently that is not triggering the branches you are talking about even though nx2 and nx3 are both 1, which section of the advection tests would be best to cover the use cases you are talking about. 

> More a general question: What best practices in terms of coverage on different devices. We only have very few place in the code that are architecture/backend dependent so I'm not really worried about it.

It's a good question, I think it would be worth creating an issue about it in case things change, but as you mentioned I'm not sure it is a huge concern at this point. I imagine most of the architecture dependent options are hidden behind kokkos abstraction layers, hopefully it will stay that way.

JoshuaSBrown:
ping @pgrete

tst/regression/test_suites/advection_convergence/advection_convergence.py
```diff
@@ -56,11 +56,14 @@ def Prepare(self,parameters, step):
...
         """
 
+
         n_res = len(lin_res)
 
         # TEST: Advection only in x-direction 
         # using nx2 = nx3 = 4 > 1 for identical errors between dimensions
         if step <= n_res:
+
+            parameters.coverage_status = "both"
```
pgrete, line 66: I think it's fine to just run the coverage for the one case as for the y-direction.
```suggestion
            if lin_res[step - 1] == 32:
                parameters.coverage_status = "both"
```

tst/regression/test_suites/advection_convergence/advection_convergence.py
```diff
@@ -73,6 +76,9 @@ def Prepare(self,parameters, step):
                 ]
         # TEST: Advection only in y-direction
         elif step <= 2*n_res:
+            # Only run coverage for 32 case
+            if lin_res[step % n_res - 1] == 32:
```
pgrete, line 80: I'd add a similar check to to enable test coverage for one case in the z-direction (i.e, where nx3 > 1) below as it's the one case not covered yet (3D, no AMR, single MeshBlock per Mesh).

JoshuaSBrown:
> Looks good! Thanks for your patience on fixing the details.

Thanks!, though it doesn't look like coverage went up with the changes.

pgrete:
> > Looks good! Thanks for your patience on fixing the details.
> 
> Thanks!, though it doesn't look like coverage went up with the changes.

True. I guess the majority of lines specific to 3D were already covered in the AMR test (which is 3d)

pgrete:
@jdolence @Yurlungur we just talked about the PR in the meeting. Does the LANL Physics side have any strong feeling about his? Otherwise, we'll go ahead and merge this after tomorrow as it got two approvals.

CONTRIBUTING.md
```diff
@@ -238,4 +238,49 @@ regression    =   1.44 sec*proc (1 test)
...
+    def Prepare(self,parameters):
+        parameters.coverage_status = "only-coverage"
+        return parameters
+```
+
+The coverage status argument can take 3 different settings:
+
+1. **only-regression** - will not run if with coverage, is the default
+2. **both** - will run with or without coverage
+3. **only-coverage** - will only run if coverage is specified
```
Yurlungur, line 266: I don't love that we have to add coverage in two separate spots. I don't see a way around this though.

JoshuaSBrown: Just to be clear, you don't need to add it in two different spots for any given test. It's just that if you use a regression test you are adding it in python and in a unit test you are adding it in the cpp test files.

PR #188, opened 2020-06-03
PreExecute/PostExecute

jdolence:
## PR Summary

Introduces PreExecute and PostExecute functions as replacements for PreDriver/PostDriver.  This allows for applications to specialize setup and takedown.

Addresses #162.

## PR Checklist

- [x] Code passes cpplint
- [x] Examples all make use of these functions, including demonstrating specialization.
- [x] Code is formatted

### DIFF
--- a/example/face_fields/face_fields_example.cpp
+++ b/example/face_fields/face_fields_example.cpp
@@ -56,7 +56,7 @@ void MeshBlock::ProblemGenerator(ParameterInput *pin) {
 }
 
 DriverStatus FaceFieldExample::Execute() {
-  PreExecute();
+  Driver::PreExecute();
   DriverUtils::ConstructAndExecuteBlockTasks<>(this);
 
   // post-evolution analysis
@@ -92,7 +92,7 @@ DriverStatus FaceFieldExample::Execute() {
   }
 
   pmesh->mbcnt = pmesh->nbtotal;
-  PostExecute();
+  Driver::PostExecute();
   return DriverStatus::complete;
 }
 
--- a/example/face_fields/face_fields_example.cpp
+++ b/example/face_fields/face_fields_example.cpp
@@ -56,6 +56,7 @@ void MeshBlock::ProblemGenerator(ParameterInput *pin) {
 }
 
 DriverStatus FaceFieldExample::Execute() {
+  PreExecute();
   DriverUtils::ConstructAndExecuteBlockTasks<>(this);
 
   // post-evolution analysis
@@ -91,6 +92,7 @@ DriverStatus FaceFieldExample::Execute() {
   }
 
   pmesh->mbcnt = pmesh->nbtotal;
+  PostExecute();
   return DriverStatus::complete;
 }
 
--- a/docs/driver.md
+++ b/docs/driver.md
@@ -4,7 +4,7 @@ Parthenon provides some basic functionality for coordinating various types of ca
 
 ## Driver
 
-```Driver``` is an abstract base class that owns pointers to a ```ParameterInput``` object, a ```Mesh``` object, and an ```Outputs``` object.  It has a single pure virtual member function called ```Execute``` that must be defined by a derived class and is intended to be called from ```main()```.  A simple example of defining an application based on inheriting from this class can be found [here](../example/calculate_pi/pi.hpp).
+```Driver``` is an abstract base class that owns pointers to a ```ParameterInput``` object, a ```Mesh``` object, and an ```Outputs``` object.  It has a single pure virtual member function called ```Execute``` that must be defined by a derived class and is intended to be called from ```main()```.  A simple example of defining an application based on inheriting from this class can be found [here](../example/calculate_pi/pi_driver.hpp).
 
 ## EvolutionDriver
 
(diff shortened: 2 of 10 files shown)


jdolence:
This one should be quick and easy.  I think it's ready to go.  Tagging @jmstone because we had discussed this a while back, @pgrete because it came up in one of his earlier PR's, and @bprather because I know he's been waiting for this.

nmsriram:
@jdolence I have a philosophical question: Why are preExecute() and postExecute() called within execute()?  I would argue that these should be called in the main() with something like:
````C++
  appDriver.preExecute(<args>);
  appDriver.Execute();
  appDriver.postExecute(<args>);
````
Otherwise they could just as well be ordinary subroutines.

**Comment by jdolence** [comment]
> @jdolence I have a philosophical question: Why are preExecute() and postExecute() called within execute()? I would argue that these should be called in the main() with something like:
> 
> ```c++
>   appDriver.preExecute(<args>);
>   appDriver.Execute();
>   appDriver.postExecute(<args>);
> ```
> 
> Otherwise they could just as well be ordinary subroutines.

I had a similar thought, so I know what you mean.  Here was my basic reasoning.

1. I suspect most applications will want the functionality provided by these functions and also probably won't need to extend/redefine them.  Sticking them at the beginning/end of parthenon-defined Execute functions (i.e. in the `EvolutionDriver`) means applications will call these functions without needing to even be aware of them, which I think is nice.
2. With the current arrangement, both `PreExecute` and `PostExecute` can be redefined in such a way as to be functions of local variables in `Execute`.  This is probably more relevant for `PostExecute`.  The pi example shows a use case for this where `PostExecute` was redefined to take the computed value of pi as an argument and print it to the screen and a file.  Moving these calls into `main` would mean stashing the value of pi in the driver as a member variable (at least this is one solution) so that `PostExecute` could have access to it.  Just seems less flexible.

I could easily be convinced I'm wrong on this, so feel free to argue back.  One thing that might help clarify the current design is better naming.  I settled on `PreExecute` and `PostExecute` before I settled on where they'd be called from.  I can't immediately think of better names, though.  Suggestions?

**Comment by jdolence** [comment]
> Looks good to me. I don't see any problems. I only have a suggestion. I would use the full scope when calling on an inherited virtual method, like you did in the PostExecute() call in the pi example, as it adds clarity and intent.
> 
> ```
> Driver::PreExecute();
> Driver::PostExecute();
> ```

Is there a specific place you'd recommend adding the scoping?  Thanks for the quick review, btw.

File: example/face_fields/face_fields_example.cpp [hunk 1]
```diff
@@ -91,6 +92,7 @@ DriverStatus FaceFieldExample::Execute() {
   }
 
   pmesh->mbcnt = pmesh->nbtotal;
+  PostExecute();
```
JoshuaSBrown, line 95: Here:

Driver::PostExecute();

jdolence:
> > 2. With the current arrangement, both `PreExecute` and `PostExecute` can be redefined in such a way as to be functions of local variables in `Execute`.  This is probably more relevant for `PostExecute`.  The pi example shows a use case for this where `PostExecute` was redefined to take the computed value of pi as an argument and print it to the screen and a file.  Moving these calls into `main` would mean stashing the value of pi in the driver as a member variable (at least this is one solution) so that `PostExecute` could have access to it.  Just seems less flexible.
> 
> I see your point, but perhaps they should not be called `preExecute` and `postExecute` since they are clearly called within `Execute`. Perhaps `prologue()` and `epilogue()` if you want to retain them in the `Execute()` function?
> 
> Perhaps `Execute()` should return a `result` class that held values that both `preExecute` and `postExecute` could change?
> 
> I realize this is a minor point and am willing to just go with what you have if you feel strongly about it.

I don't feel super strongly about these details.  I think a `result` class might be more work than it's worth to be appropriately general, so I think I'm leaning towards leaving these calls where they are.  That said, I like the idea of changing the names.  `Prologue` and `Epilogue` might do nicely.  I'll wait to see if @pgrete or @jmstone have any comments on any of this before changing things.

Thanks!

jmstone:
I like this re-organization.  But I agree that somehow we need to convey the idea these functions are for things that cannot be done at the start/end of the Driver itself, ideally through the function names.  Right now they just print diagnostic messages.  Maybe something like PreExecuteDiagnostics?  But I suppose in future they might be used for other things so this is also not ideal.  
Right now it seems the diagnostic output is specific to time-evolution drives (prints ncycles, zc/sec, etc).  Somehow we need to generalize and abstract this, because of course some drivers do not involve time-evolution.  Anyways, I approve.  We can always tweak later.

jdolence:
> I like this re-organization. But I agree that somehow we need to convey the idea these functions are for things that cannot be done at the start/end of the Driver itself, ideally through the function names. Right now they just print diagnostic messages. Maybe something like PreExecuteDiagnostics? But I suppose in future they might be used for other things so this is also not ideal.
> Right now it seems the diagnostic output is specific to time-evolution drives (prints ncycles, zc/sec, etc). Somehow we need to generalize and abstract this, because of course some drivers do not involve time-evolution. Anyways, I approve. We can always tweak later.

Thanks for taking a look @jmstone.  I agree, this isn't quite perfect yet, so we can/should revisit later.  Perhaps as this is used more it will become more clear how to modify.

One use case that's imminent is @bprather's work where he needs to modify initial conditions based on some global normalization, so it will quickly move beyond just diagnostics.

Anyway, thanks again!

PR #193, opened 2020-06-09
Change parthenon::Mesh to more modern C++ style

AndrewGaspar:
## PR Summary

I wanted to replace the manual memory management in Mesh with automatic memory management by replacing raw pointers with STL containers. That was the easy part - mostly replaced `new[]` and `delete[]` with `std::vector`. The more involved part came when I replaced the intrusive MeshBlock list with the `std::list` container. However, I think it cleaned up a lot of the code that manages that linked list (Step 7 of RedistributeAndRefineMeshBlocks is much clearer now, imo), so I think the change is worth it.

I also incidentally made some cosmetic changes to `Mesh::CalculateLoadBalance` (mostly variable renames) to improve my understanding of what it was doing. 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


example/advection/redefine_parthenon_defaults.cpp
```diff
@@ -97,8 +97,7 @@ void Mesh::UserWorkAfterLoop(ParameterInput *pin, SimTime &tm) {
   Real l1_err = 0.0;
   Real max_err = 0.0;
 
-  MeshBlock *pmb = pblock;
-  while (pmb != nullptr) {
+  for (auto pmb = pblock.begin(); pmb != pblock.end(); pmb++) {
```
JoshuaSBrown: This is slightly unrelated but still relevant in terms of cleanup. It doesn't really look like anything in this loop is being changed other than the l1_err and max_err variables, is there anyway we can add const to more of the variables in this loop, for instance I know the IndexRange structs could be made const?

JoshuaSBrown: Placing this loop in its own private method, or declaring a static function with local file scope would also improve the readability e.g. Calculatel1andMaxError. This is just a suggestion though.

example/calculate_pi/pi_driver.cpp
```diff
@@ -64,15 +64,13 @@ parthenon::DriverStatus PiDriver::Execute() {
   // All the blocks are done, now do a global reduce and spit out the answer
   // first sum over blocks on this rank
   Real area = 0.0;
-  MeshBlock *pmb = pmesh->pblock;
-  while (pmb != nullptr) {
-    Container<Real> &rc = pmb->real_containers.Get();
+  for (auto &block : pmesh->pblock) {
+    Container<Real> &rc = block.real_containers.Get();
```
JoshuaSBrown: It's a shame there is not a const get method for the container collection.

example/face_fields/face_fields_example.cpp
```diff
@@ -61,13 +61,12 @@ DriverStatus FaceFieldExample::Execute() {
...
-  while (pmb != nullptr) {
+  for (auto &block : pmesh->pblock) {
     parthenon::IndexDomain interior = parthenon::IndexDomain::interior;
-    parthenon::IndexRange ib = pmb->cellbounds.GetBoundsI(interior);
-    parthenon::IndexRange jb = pmb->cellbounds.GetBoundsJ(interior);
-    parthenon::IndexRange kb = pmb->cellbounds.GetBoundsK(interior);
-    Container<Real> &rc = pmb->real_containers.Get();
+    parthenon::IndexRange ib = block.cellbounds.GetBoundsI(interior);
+    parthenon::IndexRange jb = block.cellbounds.GetBoundsJ(interior);
+    parthenon::IndexRange kb = block.cellbounds.GetBoundsK(interior);
```
JoshuaSBrown: Make these const.

example/kokkos_pi/kokkos_pi.cpp
```diff
@@ -169,41 +167,24 @@ static MeshBlock *setupMesh(const int &n_block, const int &n_mesh, const double
...
         // Add variable for in_or_out
-        Container<Real> &base = pmb->real_containers.Get();
-        base.setBlock(pmb);
+        Container<Real> &base = pmb.real_containers.Get();
+        base.setBlock(&pmb);
         base.Add("in_or_out", myMetadata);
-        // repoint lastBlock for next iteration
-        lastBlock = pmb;
       }
     }
```
JoshuaSBrown, line 182: This is a lot cleaner, nice!

example/kokkos_pi/kokkos_pi.cpp
```diff
@@ -169,41 +167,24 @@ static MeshBlock *setupMesh(const int &n_block, const int &n_mesh, const double
...
-void deleteMesh(MeshBlock *firstBlock) {
-  // deletes the mesh chain starting with firstBlock
-  auto pmb = firstBlock;
-  while (pmb) {
-    auto next = pmb->next;
-    delete pmb;
-    pmb = next;
-  }
+  return block_list;
 }
```
JoshuaSBrown, line 189: Nice!

src/mesh/amr_loadbalance.cpp
```diff
@@ -70,69 +70,69 @@ void Mesh::LoadBalancingAndAdaptiveMeshRefinement(ParameterInput *pin) {
...
 //----------------------------------------------------------------------------------------
-// \!fn void Mesh::CalculateLoadBalance(double *clist, int *rlist, int *slist,
-//                                      int *nlist, int nb)
 // \brief Calculate distribution of MeshBlocks based on the cost list
-
-void Mesh::CalculateLoadBalance(double *clist, int *rlist, int *slist, int *nlist,
-                                int nb) {
+void Mesh::CalculateLoadBalance(int total_blocks, std::vector<double> const &costlist,
+                                std::vector<int> &ranklist, std::vector<int> &nslist,
+                                std::vector<int> &nblist) {
```
JoshuaSBrown: I like the improvement in variable names, it makes it a lot clearer what clist and rlist are.

src/mesh/amr_loadbalance.cpp
```diff
@@ -70,69 +70,69 @@ void Mesh::LoadBalancingAndAdaptiveMeshRefinement(ParameterInput *pin) {
...
   std::stringstream msg;
   double real_max = std::numeric_limits<double>::max();
   double totalcost = 0, maxcost = 0.0, mincost = (real_max);
 
-  for (int i = 0; i < nb; i++) {
-    totalcost += clist[i];
-    mincost = std::min(mincost, clist[i]);
-    maxcost = std::max(maxcost, clist[i]);
+  for (int block = 0; block < total_blocks; block++) {
```
JoshuaSBrown: Since you are now using the vectors for the cost list etc, is the variable total_blocks really needed, can you not just use the size of the vector? 

At least I think you could declare total_blocks within the function instead of having to pass it in.

JoshuaSBrown: or just replace with a range based for?

src/mesh/amr_loadbalance.cpp
```diff
@@ -70,69 +70,69 @@ void Mesh::LoadBalancingAndAdaptiveMeshRefinement(ParameterInput *pin) {
...
+      mycost += costlist[block];
+      ranklist[block] = rank;
+      if (mycost >= targetcost && rank > 0) {
+        rank--;
+        totalcost -= mycost;
+        mycost = 0.0;
+        targetcost = totalcost / (rank + 1);
+      }
     }
   }
```
JoshuaSBrown: I would move this into it's own static function. I also think you can get away without using the total_blocks and instead rely on the costlist size or ranklist size in this case as well.

src/mesh/amr_loadbalance.cpp
```diff
@@ -70,69 +70,69 @@ void Mesh::LoadBalancingAndAdaptiveMeshRefinement(ParameterInput *pin) {
...
+    nslist[0] = 0;
+    int rank = 0;
+    for (int block = 1; block < total_blocks; block++) {
+      if (ranklist[block] != ranklist[block - 1]) {
+        nblist[rank] = block - nslist[rank];
+        nslist[++rank] = block;
+      }
     }
+    nblist[rank] = total_blocks - nslist[rank];
   }
```
JoshuaSBrown: I would move to static function.

src/mesh/amr_loadbalance.cpp
JoshuaSBrown: Even better can we just replace this with
```
double mincost = *std::min_element(costlist.begin(), costlist.end());
double maxcost = *std::max_element(costlist.begin(), costlist.end());
double totalcost = std::accumulate(costlist.being(), costlist.end(), 0.0);
```

Though you would need to make sure costlist is not empty, I think that would be problematic anyway.

src/driver/driver.cpp
```diff
@@ -133,10 +133,8 @@ void EvolutionDriver::PostExecute(DriverStatus status) {
 
 void EvolutionDriver::InitializeBlockTimeSteps() {
   // calculate the first time step
-  MeshBlock *pmb = pmesh->pblock;
-  while (pmb != nullptr) {
-    pmb->SetBlockTimestep(Update::EstimateTimestep(pmb->real_containers.Get()));
-    pmb = pmb->next;
+  for (auto &block : pmesh->pblock) {
```
JoshuaSBrown: Here you are using block as a reference to the meshblock but below in amr_loadbalance.cpp you are using it as an id. I would advise consistency. Maybe here use `mb` for meshblock?

src/mesh/amr_loadbalance.cpp
```diff
@@ -416,15 +407,15 @@ void Mesh::RedistributeAndRefineMeshBlocks(ParameterInput *pin, int ntot) {
   int onbe = onbs + nblist[Globals::my_rank] - 1;
 #endif
   // Step 2. Calculate new load balance
-  CalculateLoadBalance(newcost, newrank, nslist, nblist, ntot);
+  CalculateLoadBalance(ntot, newcost, newrank, nslist, nblist);
```
JoshuaSBrown: ntot is not needed.

src/bvals/bvals_var.cpp
```diff
@@ -108,7 +108,7 @@ void BoundaryVariable::DestroyBoundaryData(BoundaryData<> &bd) {
 void BoundaryVariable::CopyVariableBufferSameProcess(NeighborBlock &nb, int ssize) {
   // Locate target buffer
   // 1) which MeshBlock?
-  MeshBlock *ptarget_block = pmy_mesh_->FindMeshBlock(nb.snb.gid);
+  auto ptarget_block = pmy_mesh_->FindMeshBlock(nb.snb.gid);
```
JoshuaSBrown: I have mixed feelings about using auto a lot. I know that it helps to reduce the verbosity but it also hides the type.

JoshuaSBrown: I think a good compromise if you are going to be using auto for these, is to alter the variable name so that it is clear that it is an iterator. We have been prepending `p` for pointer maybe we could prepent `i` for iterator, or come up with some other convention.

src/mesh/mesh.cpp
```diff
@@ -427,26 +426,25 @@ Mesh::Mesh(ParameterInput *pin, Properties_t &properties, Packages_t &packages,
...
-    brdisp = new int[Globals::nranks];
-    bddisp = new int[Globals::nranks];
+    nref = std::vector<int>(Globals::nranks);
+    nderef = std::vector<int>(Globals::nranks);
+    rdisp = std::vector<int>(Globals::nranks);
+    ddisp = std::vector<int>(Globals::nranks);
+    bnref = std::vector<int>(Globals::nranks);
+    bnderef = std::vector<int>(Globals::nranks);
+    brdisp = std::vector<int>(Globals::nranks);
+    bddisp = std::vector<int>(Globals::nranks);
```
JoshuaSBrown, line 454: This is a huge improvement.

src/mesh/mesh.cpp
```diff
@@ -427,26 +426,25 @@ Mesh::Mesh(ParameterInput *pin, Properties_t &properties, Packages_t &packages,
...
   }
 
   // initialize cost array with the simplest estimate; all the blocks are equal
-  for (int i = 0; i < nbtotal; i++)
-    costlist[i] = 1.0;
+  costlist = std::vector<double>(nbtotal, 1.0);
 
-  CalculateLoadBalance(costlist, ranklist, nslist, nblist, nbtotal);
+  CalculateLoadBalance(nbtotal, costlist, ranklist, nslist, nblist);
```
JoshuaSBrown: nbtotal is not needed.

src/mesh/mesh.cpp
```diff
@@ -463,19 +461,10 @@ Mesh::Mesh(ParameterInput *pin, Properties_t &properties, Packages_t &packages,
...
-    } else {
-      pblock->next = new MeshBlock(i, i - nbs, loclist[i], block_size, block_bcs, this,
-                                   pin, properties, packages, gflag);
-      pblock->next->prev = pblock;
-      pblock = pblock->next;
-    }
-    pblock->pbval->SearchAndSetNeighbors(tree, ranklist, nslist);
+    pblock.emplace_back(i, i - nbs, loclist[i], block_size, block_bcs, this, pin,
+                        properties, packages, gflag);
+    pblock.front().pbval->SearchAndSetNeighbors(tree, ranklist.data(), nslist.data());
```
JoshuaSBrown: Nice!

src/mesh/mesh.cpp
```diff
@@ -746,17 +734,17 @@ Mesh::Mesh(ParameterInput *pin, IOWrapper &resfile, Properties_t &properties,
...
+    rdisp = std::vector<int>(Globals::nranks);
+    ddisp = std::vector<int>(Globals::nranks);
+    bnref = std::vector<int>(Globals::nranks);
+    bnderef = std::vector<int>(Globals::nranks);
+    brdisp = std::vector<int>(Globals::nranks);
+    bddisp = std::vector<int>(Globals::nranks);
   }
 
-  CalculateLoadBalance(costlist, ranklist, nslist, nblist, nbtotal);
+  CalculateLoadBalance(nbtotal, costlist, ranklist, nslist, nblist);
```
JoshuaSBrown: nbtotal is not needed.

src/mesh/mesh.cpp
```diff
@@ -816,36 +794,7 @@ Mesh::Mesh(ParameterInput *pin, IOWrapper &resfile, Properties_t &properties,
...
-    delete[] bddisp;
-  }
-  // delete user Mesh data
-  if (nuser_history_output_ > 0) {
-    delete[] user_history_output_names_;
-    delete[] user_history_func_;
-    delete[] user_history_ops_;
-  }
-}
+Mesh::~Mesh() = default;
```
JoshuaSBrown, line 747: Nice!

src/mesh/mesh.hpp
```diff
@@ -361,27 +363,32 @@ class Mesh {
   int GetRootLevel() { return root_level; }
   int GetMaxLevel() { return max_level; }
   int GetCurrentLevel() { return current_level; }
-  std::vector<int> GetNbList() {
-    std::vector<int> nlist;
-    nlist.assign(nblist, nblist + Globals::nranks);
-    return nlist;
-  }
+  std::vector<int> GetNbList() { return nblist; }
```
JoshuaSBrown: Can you make these const noexcept?

src/mesh/mesh.hpp
```diff
@@ -361,27 +363,32 @@ class Mesh {
...
-  int *bnref, *bnderef;
-  int *brdisp, *bddisp;
+  /// Count of blocks to refine on each rank
+  std::vector<int> nref;
+  /// Count of blocks to de-refine on each rank
+  std::vector<int> nderef;
+  std::vector<int> rdisp, ddisp;
+  std::vector<int> bnref, bnderef;
+  std::vector<int> brdisp, bddisp;
   // the last 4x should be std::size_t, but are limited to int by MPI
```
JoshuaSBrown, line 422: This is a lot cleaner and more descriptive.

example/advection/redefine_parthenon_defaults.cpp
AndrewGaspar: This PR was already racking up a large change count, so I was trying to avoid touching too much additional code. Maybe this could be done in a different PR.

example/calculate_pi/pi_driver.cpp
AndrewGaspar: File an issue and this could be fixed later.

src/bvals/bvals_var.cpp
AndrewGaspar: Well, the goal was to avoid changing any of the code that references ptarget_block... I added a concrete type.

src/driver/driver.cpp
AndrewGaspar: Good idea on mb.

AndrewGaspar:
> Thanks for going through and doing this. I have one rather stylistic and pedantic concern. The naming convention in `Athena++` is that pointer names begin with `p`. Now with the mesh block list changed from a pointer to a `std::list`, I'm not sure the `p` makes sense anymore. Thoughts?

Yeah, I'll change `pblock` to `block_list`.

example/calculate_pi/pi_driver.cpp
AndrewGaspar: https://github.com/lanl/parthenon/issues/197

pgrete:
@AndrewGaspar looks like the CI fail for MPI is real. The run gets stuck in the first cycle (which eventually results in the timeout).

@jmstone  would you mind reviewing this from the AthenaPK side and I'll focus on reviewing the Kokkos related PR #177 and #196

Yurlungur:
@jmstone mentioned:

> Comment: I believe it was Jonah who found the neighbor find algorithm in Athena++ was taking a lot of time. Kengo fixed this in the latest (development) version. Now the MeshBlocks are stored as an array, rather than a linked list (which has implications for #193 I think). When the MeshBlockTree gets large (when there are lots of MeshBlocks within an MPI rank), then this improves performance and might be the reason why the public version of Athena++ does not perform as well as the plot shown in the method paper

So that's perhaps worth discussing  here.  Clearly we should update to that fix. That said,  I think this should still go through. In my opinion, C++  container semantics mean replacing `std::list` with `std::vector` will be easier than replacing a hand-implemented linked list with an array.

src/mesh/amr_loadbalance.cpp
```diff
@@ -70,73 +72,94 @@ void Mesh::LoadBalancingAndAdaptiveMeshRefinement(ParameterInput *pin) {
...
+// Private routines
+namespace {
+/**
+ * @brief This routine assigns blocks to ranks by attempting to place index-contiguous
+ * blocks of equal total cost on each rank.
+ *
+ * @param costlist (Input) A map of global block ID to a relative weight.
+ * @param ranklist (Output) A map of global block ID to ranks.
+ */
+void AssignBlocks(std::vector<double> const &costlist, std::vector<int> &ranklist) {
```
forrestglines, line 85: Since a lot of this pull request will affect AMR load balancing - do we have good tests in place for load balancing?

pgrete:
We briefly talked about the PR during the bi-weekly call and decided to add WIP again given that there still seems to be a bug in combination with the current limited availability of the main author.

Yurlungur:
@AndrewGaspar @pgrete I found the MPI bug. `meshblock-pbval->SearchAndSetNeighbors` needs to be called for every block on every rank when `Mesh` is created, but it was only being called for the first block on each rank. It should be fixed now.

Once CI passes, assuming @JoshuaSBrown's concerns are addressed, this is probably ready for merge.

One thing to note: the bug was a little hard to track down. It would be easier if we had unit tests for the components of the AMR machinery. I understand this probably requires a refactor, since everything is tangled. I'm in the process of writing a unit test for prolongation, which might serve as a model for at least some unit tests without disentangling everything. That should be a separate MR request though. Especially since it depends on the changes here.

Yurlungur:
Removing WIP status

example/kokkos_pi/kokkos_pi.cpp
```diff
@@ -169,41 +167,24 @@ static MeshBlock *setupMesh(const int &n_block, const int &n_mesh, const double
...
         // set coordinates of first cell center
         h_xyz(0, idx) = dxyzCell * (static_cast<Real>(i_mesh * n_block) + 0.5) - delta;
         h_xyz(1, idx) = dxyzCell * (static_cast<Real>(j_mesh * n_block) + 0.5) - delta;
         h_xyz(2, idx) = dxyzCell * (static_cast<Real>(k_mesh * n_block) + 0.5) - delta;
         // Add variable for in_or_out
-        auto &base = pmb->real_containers.Get();
-        base->setBlock(pmb);
+        auto &base = pmb.real_containers.Get();
+        base->setBlock(&pmb);
         base->Add("in_or_out", myMetadata);
```
JoshuaSBrown: This is now a mesh block not a pointer, I would change this to mb not pmb.

JoshuaSBrown:
@Yurlungur, I will look over this tomorrow. It's been a while so I need to make sure I understand what is going on.

JoshuaSBrown:
> @AndrewGaspar @pgrete I found the MPI bug. `meshblock-pbval->SearchAndSetNeighbors` needs to be called for every block on every rank when `Mesh` is created, but it was only being called for the first block on each rank. It should be fixed now.
> 
> Once CI passes, assuming @JoshuaSBrown's concerns are addressed, this is probably ready for merge.
> 
> One thing to note: the bug was a little hard to track down. It would be easier if we had unit tests for the components of the AMR machinery. I understand this probably requires a refactor, since everything is tangled. I'm in the process of writing a unit test for prolongation, which might serve as a model for at least some unit tests without disentangling everything. That should be a separate MR request though. Especially since it depends on the changes here.

I'm in agreement with you on the unit tests as well, it would be nice to get our code coverage above 70% :) so we can pinpoint problems faster.

example/kokkos_pi/kokkos_pi.cpp
```diff
@@ -123,9 +123,8 @@ static double sumArray(MeshBlock *firstBlock, const int &n_block) {
   double theSum = 0.0;
   // reduce the sum on the device
   // I'm pretty sure I can do this better, but not worried about performance for this
-  MeshBlock *pmb = firstBlock;
-  while (pmb) {
-    auto &base = pmb->real_containers.Get();
+  for (auto &mb : blocks) {
+    auto &base = mb.real_containers.Get();
     auto inOrOut = base->PackVariables({Metadata::Independent});
```
JoshuaSBrown, line 129: Optional: This is a bit of rabbit hole sense most of these containers do not have const access methods, but really all these should be const because you are not really changing anything internally.

src/driver/driver.hpp
```diff
@@ -85,10 +85,8 @@ template <typename T, class... Args>
 TaskListStatus ConstructAndExecuteBlockTasks(T *driver, Args... args) {
   int nmb = driver->pmesh->GetNumMeshBlocksThisRank(Globals::my_rank);
   std::vector<TaskList> task_lists;
-  MeshBlock *pmb = driver->pmesh->pblock;
-  while (pmb != nullptr) {
-    task_lists.push_back(driver->MakeTaskList(pmb, std::forward<Args>(args)...));
-    pmb = pmb->next;
+  for (auto &mb : driver->pmesh->block_list) {
+    task_lists.push_back(driver->MakeTaskList(&mb, std::forward<Args>(args)...));
```
JoshuaSBrown, line 97: This needs is's own issue, it looks like the  MakeTaskList method does not necessarily exist in the driver class. This should probably be made a pure virtual method which forces users writing the driver to override it. Unless, that was some reason why the MakeTaskList should be optional.

src/mesh/amr_loadbalance.cpp
```diff
@@ -70,73 +72,94 @@ void Mesh::LoadBalancingAndAdaptiveMeshRefinement(ParameterInput *pin) {
...
-    totalcost += clist[i];
-    mincost = std::min(mincost, clist[i]);
-    maxcost = std::max(maxcost, clist[i]);
-  }
-
-  int j = (Globals::nranks)-1;
-  double targetcost = totalcost / Globals::nranks;
-  double mycost = 0.0;
+// Private routines
+namespace {
```
JoshuaSBrown, line 77: This is clever, I did not know about unnamed namespaces.

src/mesh/amr_loadbalance.cpp
JoshuaSBrown, line 85: ranklist should be returned by AssignBlocks and not be brought in as a reference. The only place AssignBlocks is called is within the CalculateLoadBalance method. It should also probably be returned by that method instead of passed in as a reference as well. Every time rank list is passed into CalculateLoadBalance it has nothing stored in it. 


### Optional:

This might also be a different issue, I would have CalcualteLoadBalance return some struct that has the results that are initialized, I do not like passing in objects as references if you don't actually need their contents. In this case rank list is always empty when it is passed in to CalculateLoadBalance.

src/driver/driver.hpp
Yurlungur: I recall that when @jdolence designed the driver, there was some reason `MakeTaskList` wasn't required by the driver. One reason is that not all drivers have a `step` iteration that should have a task list... for example the `pi` problem. Maybe @jdolence can comment more.

JoshuaSBrown: I'm still wondering if it might be better practice to have a pure virtual method, that way you enforce the correct prototype, you can always have an empty method if it is not needed. I don't know the answer though... 

Ok, I see what is going on you have a lot of flexibility with this current implementation. Disregard my suggestions.

src/mesh/amr_loadbalance.cpp
```diff
@@ -70,73 +72,94 @@ void Mesh::LoadBalancingAndAdaptiveMeshRefinement(ParameterInput *pin) {
...
-  nlist[j] = nb - slist[j];
+}
 
-  if (Globals::my_rank == 0) {
-    for (int i = 0; i < Globals::nranks; i++) {
-      double rcost = 0.0;
-      for (int n = slist[i]; n < slist[i] + nlist[i]; n++)
-        rcost += clist[n];
+void UpdateBlockList(std::vector<int> const &ranklist, std::vector<int> &nslist,
+                     std::vector<int> &nblist) {
```
JoshuaSBrown, line 115: ### Optional

Is this function completely over writing nslist and nblist? I'm not sure, but if it is I would think it would be preferable to simply return ns and nb list in the return value. This might be more stylistic than necessary though.

example/kokkos_pi/kokkos_pi.cpp
Yurlungur: That's worth doing, but should probably be a separate issue.

Yurlungur: Issue created.

src/mesh/amr_loadbalance.cpp
Yurlungur: This now lives in its own function, outside of the `MeshBlock` class.

src/mesh/amr_loadbalance.cpp
Yurlungur: Done.

src/bvals/bvals_var.cpp
Yurlungur: @JoshuaSBrown is this resolved?

src/mesh/mesh.cpp
Yurlungur: Resolved.

src/mesh/mesh.hpp
Yurlungur: Done.

src/mesh/amr_loadbalance.cpp
Yurlungur: I do not think we have any tests of load balancing.

example/kokkos_pi/kokkos_pi.cpp
Yurlungur: Done.

src/mesh/amr_loadbalance.cpp
Yurlungur: Regarding `AssignBlocks`, I disagree. The reason to pass it in by reference is because it's a `std::vector`. It's best practice not to return vectors by value, since they're essentially arrays. And they could be substantial in size, `nblocks`, which could be O(100) or more.

Same with with `CalculateLoadBalance`, although I agree that the the arguments could be consolidated into a `struct`, I think that struct should still be passed by reference. And since it's only 3 or 4 args, I'm not sure it's worth it.

src/mesh/amr_loadbalance.cpp
Yurlungur: Let's pass on this for now. I really want to get this thing merged in, and it's been sitting for a long time.

Yurlungur:
I think I've resolved things to everyone's satisfaction, and I expect tests to pass (or I will make it so). Thanks to @nmsriram for help with getting restarts working and fixing the IO bug.

@AndrewGaspar do you have any objections to me merging this in as soon as tests pass?

AndrewGaspar:
Updating the branch so tests can run again... Thank you for getting this working.

Yurlungur:
Pulling the trigger.

PR #194, opened 2020-06-10
Add cpplint to build

AndrewGaspar:
## PR Summary

I always had to look up how to run cpplint, so now cpplint runs as part of CMake. The lint target is automatically added to the default build target, but can be disabled with `PARTHENON_LINT_DEFAULT`.

## PR Checklist

- [x] Code passes cpplint (Now it must!)
- [x] Code is formatted


example/main.cpp
```diff
@@ -1,131 +0,0 @@
-//========================================================================================
```
Yurlungur, line 1: I thought we deleted this file a while ago. Why is it showing up in the diff now?

src/CMakeLists.txt
```diff
@@ -12,9 +12,6 @@
 # the public, perform publicly and display publicly, and to permit others to do so.
 #=========================================================================================
 
-# Configure defs.hpp
-set(PROBLEM_GENERATOR "<not-implemented>") # TODO: Figure out what to put here
-
```
Yurlungur, line 17: I'm not sure I would delete this. I think this is still an outstanding problem we will eventually need to solve.

example/main.cpp
AndrewGaspar: https://github.com/lanl/parthenon/blob/master/example/main.cpp

Yurlungur: huh ok I could have sworn we deleted this a while ago. My bad.

PR #200, opened 2020-06-18
Adjusting gitlab yml for internal trigger of ci

JoshuaSBrown:
## PR Summary

Currently, testing triggering of internal ci for use on darwin. Preliminary testing is complete. Ci makes use of build scripts that are located in /scripts/darwin. The reason for using scripts was to help isolate the gitlab ci instance from the user environment. 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


JoshuaSBrown:
Current, output of the test can be seen on gitlab here: https://xcp-gitlab.lanl.gov/eap-oss/parthenon/-/jobs/75494

pgrete:
I saw that you mentioned the external repo for the scripts also in #224 
What's the advantage of keeping those separate rather than storing the scripts here (e.g., in the scripts directory). I see potential problem when scripts/environment/code get out of sync. In addition, I think that scripts/tests/anything that specifically related to "this" code is also best placed in "this" repo so that it's a one-stop-shop.

Few additional/minor comments
- We should agree on a naming convention for the yml files (`.internal-gitlab-ci.yml` versus `.gitlab-ci-ias.yml` in #224). I feel less strong about pre - or postfixed and am happy to update my PR, but I suggest to rename `internal` here to `lanl` or `darwin` or sth else more specific
- Similar to #224 a brief note in the docs explaining the environment and what's tested would be great (we could take care of it once #224 is merged) or add it later
- Is the mirroring/CI triggered manually or automatically and/or does the gitlab instance has direct outside access? I'm asking as we could then add a webhook to report the results along the current results
- Is it worth keeping the history of all the commits or can we squash the 50 commits on the single file to one?

JoshuaSBrown:
> I saw that you mentioned the external repo for the scripts also in #224
> What's the advantage of keeping those separate rather than storing the scripts here (e.g., in the scripts directory). I see potential problem when scripts/environment/code get out of sync. In addition, I think that scripts/tests/anything that specifically related to "this" code is also best placed in "this" repo so that it's a one-stop-shop.

The main advantage comes with automating the build of docker images through the parthenon-buildenv-ci, and any other containers that are needed by parthenon. This advantage is not currently seen in the parthenon-buildenv because, at the moment docker is not an option, but there are several projects at lanl that use a similar setup.

pgrete:
> > I saw that you mentioned the external repo for the scripts also in #224
> > What's the advantage of keeping those separate rather than storing the scripts here (e.g., in the scripts directory). I see potential problem when scripts/environment/code get out of sync. In addition, I think that scripts/tests/anything that specifically related to "this" code is also best placed in "this" repo so that it's a one-stop-shop.
> 
> The main advantage comes with automating the build of docker images through the parthenon-buildenv-ci, and any other containers that are needed by parthenon. This advantage is not currently seen in the parthenon-buildenv because, at the moment docker is not an option, but there are several projects at lanl that use a similar setup.

May I'm don't fully understand the setup (and sorry for potentially being nitpicking here), but why is an external repository required to make this happen?
And in case it's absolutely required, I'd still vote for having the primary source of the docker image stored here (so that it's always consistent with the source it actually tests) and that the other repo is a (passive) mirror that just contains/extracts the data from here. 

> > Is the mirroring/CI triggered manually or automatically and/or does the gitlab instance has direct outside access? I'm asking as we could then add a webhook to report the results along the current results.
> 
> Currently, it is triggered manually. This needs more discussion. The problem is that for security reasons github can not access anything on the gitlab server. Gitlab can push to outside sources, we are just a little limited on what can be pushed. It should be fine to push the performance statistics out.

Great. This is exactly what I had in mind (just pushing the results out).

JoshuaSBrown:
> May I'm don't fully understand the setup (and sorry for potentially being nitpicking here), but why is an external repository required to make this happen?
And in case it's absolutely required, I'd still vote for having the primary source of the docker image stored here (so that it's always consistent with the source it actually tests) and that the other repo is a (passive) mirror that just contains/extracts the data from here.

So you don't have to have the docker images in a separate repo, they could potentially be built in a separate branch of this repo. The problem comes when/if the containers are utilized by other projects, which may very well occur particularly given that parthenon is the infrastructure component of several potential physics applications. If those applications (which are not part of parthenon but depend on it) want to make changes to the containers they will need access to parthenon. The second advantage of putting the images in a separate repo is that you would then have access to a different pot of compute resources to actually build the containers which may or may not matter. At this point in time I don't think these are large factors so it is not a big deal, but it might make things simpler in the future.

Yurlungur:
Is there a way to keep the script that is used to generate the docker image here in this repo and perhaps use a CI pipeline to generate it and upload it to wherever it lives? That might be a good way to keep things consistent.

JoshuaSBrown:
> Is there a way to keep the script that is used to generate the docker image here in this repo and perhaps use a CI pipeline to generate it and upload it to wherever it lives? That might be a good way to keep things consistent.

I guess we could do something like that with github actions. Where the container images are only updated and deployed if the dockerfiles are touched.

JoshuaSBrown:
This pull request just demonstrates a working ci on darwin. I am linking to an [issue](https://github.com/lanl/parthenon/issues/240), so I can get an idea of what combination of compiler/architecture settings people are interested in testing on.

Here is the link to the internal run https://xcp-gitlab.lanl.gov/eap-oss/parthenon/-/jobs/75979.

Yurlungur:
> I guess we could do something like that with github actions. Where the container images are only updated and deployed if the dockerfiles are touched.

That would be cool. At the very least, can we keep the scripts used to generate the images here, just as documentation?

> Here is the link to the internal run https://xcp-gitlab.lanl.gov/eap-oss/parthenon/-/jobs/75979.

Why `xcp-gitlab` instead of `gitlab.lanl.gov`? Doesn't matter much. I'm just curious.

.gitlab-ci-darwin.yml
```diff
@@ -0,0 +1,73 @@
...
+    - echo ${Kokkos_ARCH_POWER9}     >> build_cmd.txt
+    - echo ${Kokkos_ARCH_VOLTA70}    >> build_cmd.txt
+    - echo ${Kokkos_ENABLE_CUDA}     >> build_cmd.txt
+    - echo ${Kokkos_ENABLE_CUDA_UVM} >> build_cmd.txt
+    - echo ${Kokkos_ENABLE_CXX11}    >> build_cmd.txt
+    - echo ${Kokkos_ENABLE_OPENMP}   >> build_cmd.txt
+    - echo ${NUM_MPI_PROC_TESTING}   >> build_cmd.txt
+    - echo ${OMP_NUM_THREADS}        >> build_cmd.txt
+    - echo ${PARTHENON_DISABLE_HDF5} >> build_cmd.txt
+    - >
```
Yurlungur: I like that you echo all these parameters! That's great for debugging.

.gitlab-ci-darwin.yml
```diff
@@ -0,0 +1,73 @@
...
+  - if [ `cat tst/regression/gold_standard/current_version` != ${GOLD_STANDARD_VER} ]; then
+    echo "Updating regression Gold standard to ${GOLD_STANDARD_VER}." &&
+    wget -qO- https://pgrete.de/dl/parthenon_regression_gold_r${GOLD_STANDARD_VER}.tgz | tar -xz -C tst/regression/gold_standard; fi
+
+parthenon-power9-gcc-mpi-performance-regression:
+  variables: 
+    BUILD_DIR: "build_power9_perf_regression_gcc_mpi"
+    CMAKE_CXX_COMPILER: $CI_PROJECT_DIR/external/Kokkos/bin/nvcc_wrapper
+  stage: performance-regression
+  when: manual
```
Yurlungur: instead of manual triggers, could we do a scheduled build, maybe nightly or something? I'd like an email from gitlab if we break the regression tests on power9s.

JoshuaSBrown:
> > I guess we could do something like that with github actions. Where the container images are only updated and deployed if the dockerfiles are touched.
> 
> That would be cool. At the very least, can we keep the scripts used to generate the images here, just as documentation?
> 

I'm not currently using docker images, the ci for darwin is running in the gitlab runner bash shell, but yes. There were some complications with using docker initially because we cannot build docker images directly on our infrastructure hence the current implementation.  

> > Here is the link to the internal run https://xcp-gitlab.lanl.gov/eap-oss/parthenon/-/jobs/75979.
> 
> Why `xcp-gitlab` instead of `gitlab.lanl.gov`? Doesn't matter much. I'm just curious.

Good question, you would have to ask Galen.

Yurlungur:
> 'm not currently using docker images, the ci for darwin is running in the gitlab runner bash shell, but yes. There were some complications with using docker initially because we cannot build docker images directly on our infrastructure hence the current implementation. All the scripts used by the ci have already been added in /scripts/darwin/

Just noticed that. I'm happy with this solution. :)

> > > Here is the link to the internal run https://xcp-gitlab.lanl.gov/eap-oss/parthenon/-/jobs/75979.
> > 
> > 
> > Why `xcp-gitlab` instead of `gitlab.lanl.gov`? Doesn't matter much. I'm just curious.
> 
> Good question, you would have to ask Galen.

@gshipman ?

.gitlab-ci-darwin.yml
JoshuaSBrown: I'd prefer to never break anything. The manual option should allow you to trigger any pull request before it is merged with master. It just requires that you have access to lanl gitlab to cause the ci to build. We could add a scheduled build but then you have already merged with master.

Yurlungur: Can we do both? I'm nervous about making it *only* manual. Because we're all only human. At some point, somebody is going to forget. And that might be the time it breaks.

JoshuaSBrown: Fair point.

JoshuaSBrown: > Can we do both? I'm nervous about making it _only_ manual. Because we're all only human. At some point, somebody is going to forget. And that might be the time it breaks.

The other option is to make the manual trigger blocking. As in you can't merge unless it is run.

Yurlungur: I'd be okay with that. I like that option less, though, because it's an extra barrier for external collaborators. And sometimes it won't make sense, like for documentation.

JoshuaSBrown: > I'd be okay with that. I like that option less, though, because it's an extra barrier for external collaborators. And sometimes it won't make sense, like for documentation.

Good point.

.gitlab-ci-darwin.yml
```diff
@@ -0,0 +1,83 @@
...
+  MODULE_CUDA: "cuda/10.1"
+  MODULE_MPI: "openmpi/p9/4.0.2-gcc_7.4.0"
+  NUM_MPI_PROC_TESTING: 2
+  OMP_NUM_THREADS: 40
+  CMAKE_BUILD_TYPE: "Release" 
+  Kokkos_ARCH_POWER9: "ON" 
+  Kokkos_ARCH_VOLTA70: "ON" 
+  Kokkos_ENABLE_CUDA: "ON" 
+  Kokkos_ENABLE_CUDA_UVM: "OFF"
+  Kokkos_ENABLE_CXX11: "ON"
```
AndrewGaspar: I don't think this is meant to be defined - I think Kokkos sets this internally. We already specify the C++ 14 standard to Kokkos via `Kokkos_CXX_STANDARD`. Do we need this?

.gitlab-ci-darwin.yml
```diff
@@ -0,0 +1,83 @@
...
+
+stages:
+  - performance-regression
+
+# Is performed before the scripts in the stages step
+before_script:
+  - env -i bash ./scripts/darwin/setup.sh
+  - env -i bash ./scripts/darwin/build_hdf5_parallel.sh
+  - export GOLD_STANDARD_VER=2
+  - if [ `cat tst/regression/gold_standard/current_version` != ${GOLD_STANDARD_VER} ]; then
```
AndrewGaspar: This bit is outdated with #236, right?

scripts/darwin/build.sh
```diff
@@ -0,0 +1,81 @@
...
+# Make sure home is pointing to current directory
+export HOME=$(pwd)
+# Calculate number of available cores
+export J=$(( $(nproc --all) )) && echo Using ${J} cores during build
+
+# Load system modules
+module purge
+module load cmake/3.11.1
+module load gcc/7.4.0
+module load clang/8.0.0
```
AndrewGaspar: Could you add `export NVCC_WRAPPER_DEFAULT_COMPILER=g++` somewhere before the cmake configure so that it's explicit we're using GCC? I think nvcc_wrapper defaults to a GCC compiler, so I don't think it's broken right now, but seeing clang loaded _after_ gcc is confusing at first glance

scripts/darwin/setup.sh
```diff
@@ -0,0 +1,30 @@
...
+# Initialize spack env
+. spack/share/spack/setup-env.sh
+
+# Create .spack folder
+if [ ! -d ".spack" ]; then
+  mkdir -p .spack
+fi
+
+# Setup spack package yaml
+echo "packages:" > .spack/packages.yaml
```
AndrewGaspar: Maybe this could just be a separate file in source control and you just copy it?

.gitlab-ci-darwin.yml
```diff
@@ -0,0 +1,83 @@
+variables:
+  SCHEDULER_PARAMETERS: '--nodes=1 --partition=power9 --export=NONE'
+  GIT_SUBMODULE_STRATEGY: recursive
+  MODULE_CMAKE: "cmake/3.11.1"
+  MODULE_CLANG: "clang/8.0.1"
```
AndrewGaspar, line 5: Is this used?

.gitlab-ci-darwin.yml
```diff
@@ -0,0 +1,83 @@
+variables:
+  SCHEDULER_PARAMETERS: '--nodes=1 --partition=power9 --export=NONE'
+  GIT_SUBMODULE_STRATEGY: recursive
+  MODULE_CMAKE: "cmake/3.11.1"
```
AndrewGaspar: Is this used?

.gitlab-ci-darwin.yml
```diff
@@ -0,0 +1,83 @@
...
+  MODULE_CLANG: "clang/8.0.1"
+  MODULE_COMPILER: "gcc/7.4.0"
```
AndrewGaspar, line 6: Is this used?

.gitlab-ci-darwin.yml
```diff
@@ -0,0 +1,83 @@
...
+  MODULE_CUDA: "cuda/10.1"
```
AndrewGaspar, line 7: ?

.gitlab-ci-darwin.yml
```diff
@@ -0,0 +1,83 @@
...
+  MODULE_MPI: "openmpi/p9/4.0.2-gcc_7.4.0"
```
AndrewGaspar, line 8: ?

.gitlab-ci-darwin.yml
```diff
@@ -0,0 +1,83 @@
...
+      ${Kokkos_ENABLE_CUDA_UVM}
+      ${Kokkos_ENABLE_CXX11}
+      ${Kokkos_ENABLE_OPENMP}
+      ${NUM_MPI_PROC_TESTING}
+      ${OMP_NUM_THREADS}
+      ${PARTHENON_DISABLE_HDF5}
+  artifacts:
+    expire_in: 3 days
+    paths:
+      - ${CI_PROJECT_DIR}/${BUILD_DIR}/tst/regression/advection_convergence/advection-errors.dat
```
pgrete: Did you check of the performance plot is also generated?
https://github.com/lanl/parthenon/blob/c18c2267f6b0889d9a51bd304cb95f85bf80d438/.gitlab-ci-ias.yml#L119
We should keep this and push it to the PR (I'll see how that works from gitlab over the next days and report back)

.gitlab-ci-darwin.yml
JoshuaSBrown: Yes

.gitlab-ci-darwin.yml
JoshuaSBrown: Yes, this needs to be updated, the standard varaible does need to be accessible though, I think for getting the xl compiler to work. I could be wrong but xl may not work with all features of c++14.

.gitlab-ci-darwin.yml
JoshuaSBrown: Yes it is generated, there was an error in the PATH though.

.gitlab-ci-darwin.yml
JoshuaSBrown: These comments have all been addressed.

scripts/darwin/setup.sh
JoshuaSBrown: I have made it dependent on whatever, packages are specified with the env variables in the ci. It is more clear why I took this approach now. But really it doesn't matter one could simply define every module that might be needed in a single file... I'm not sure how much it matters.

scripts/darwin/build.sh
JoshuaSBrown: This has been addressed.

JoshuaSBrown:
Latest merge https://xcp-gitlab.lanl.gov/eap-oss/parthenon/-/jobs/77169

JoshuaSBrown:
https://xcp-gitlab.lanl.gov/eap-oss/parthenon/-/jobs/77197

.gitlab-ci-darwin.yml
JoshuaSBrown: Any updates?

JoshuaSBrown:
## Serial
performance

## MPI
performance

JoshuaSBrown:
@pgrete @AndrewGaspar This has been open for 3 days, I would like to merge this soon if there are no objections?

.gitlab-ci-darwin.yml
pgrete: Is this potentially related to #256 ?
It somehow feels odd to set CXX11 here while our default in the root CMakeLists.txt is supposed to be 14.

.gitlab-ci-darwin.yml
```diff
@@ -0,0 +1,95 @@
...
+      ${PARTHENON_DISABLE_HDF5}
+      ${MODULE_CMAKE}
+      ${MODULE_CLANG}
+      ${MODULE_COMPILER}
+      ${MODULE_CUDA}
+      ${MODULE_MPI}
+  artifacts:
+    expire_in: 3 days
+    paths:
+      - ${CI_PROJECT_DIR}/${BUILD_DIR}/tst/regression/outputs/advection_convergence/advection-errors.dat
```
pgrete: Is this an old version of the file?
The performance regression results should be in ` .../tst/regression/outputs/advection_performance/performance.png` don't they?

scripts/darwin/build.sh
```diff
@@ -0,0 +1,115 @@
...
+MPI_MODULE=${17}
+
+export TMPDIR=${HOME}/tmp
+
+# Temp directory apparently needed by spack because of OSError: [Errno 18] Invalid cross-device link
+if [ -d ${TMPDIR} ] 
+then
+  echo "Removing ${TMPDIR}"
+  rm -rf ${TMPDIR}/*
+  rmdir ${TMPDIR}
```
pgrete:
```suggestion
  rm -rf ${TMPDIR}
```
Otherwise the second command my fail if there are hidden files and directories in the tmp folder.

scripts/darwin/build.sh
```diff
@@ -0,0 +1,115 @@
...
+ -DKokkos_ENABLE_CUDA_UVM=$7 \
+ -DKokkos_ENABLE_CXX11=$8 \
+ -DKokkos_ENABLE_OPENMP=$9 \
+ -DNUM_MPI_PROC_TESTING=${10} \
+ -DOMP_NUM_THREADS=${11} \
+ -DPARTHENON_DISABLE_HDF5=${12} ../
+
+make -j $J VERBOSE=1
+
+ctest --output-on-failure -j $J -LE 'coverage' -L 'performance|regression'
```
pgrete: Did you choose to run all the regression tests incl. the hdf5 test on purpose?
This seems to go beyond the "performance regression" intention, which would actually be a good thing to my mind.
I think it'd be a plus of the whole test suite is run on Power9(+GPU) as we don't test that yet.

CONTRIBUTING.md
```diff
@@ -147,6 +147,13 @@ follow the instructions [below](#integrating-the-regression-test-with-cmake) *an
 `perf-reg` label to the test (see bottom of the regression
 [CMakeLists.txt](tst/regression/CMakeLists.txt)).
 
+A third pipeline is run using LANL internal systems and is run manually when
+approved, and is scheduled to run on a dailly basis on the development branch.
+These machines use the newest IBM powerPC processors and the NVIDIA V100
+(Volta) GPUs. Tests run on these systems are primarily aimed at measuring the
+performance of this specific architecture. Compilation and testing details can be found by looking
+in the [.gitlab-ci-darwin.yml](.gitlab-ci-darwin.yml) file *and* the /scripts/darwin folder..
```
pgrete: I think it'd be great to be a little more explicit here.
For example, I'm currently not sure what we're actually testing in terms of architecture and tests.
Based on the current description I would expect 
- unit performance test using a single core and single host thread, or a single GPU,
- the performance regression test using CPUs  (serial and with some MPI processes) with one thread per process
- the performance regression test using a single process with a single GPUs, and multiple processes with multiple GPUs

I think (without knowing what's in the logs) this may not match what's currently tested,

.gitlab-ci-darwin.yml
JoshuaSBrown: Ya, I need to correct this.

scripts/darwin/build.sh
JoshuaSBrown: Ok, that can be done.

.gitlab-ci-darwin.yml
JoshuaSBrown: > Is this potentially related to #256 ?
> It somehow feels odd to set CXX11 here while our default in the root CMakeLists.txt is supposed to be 14.

Not to my knowledge. I will remove it.

scripts/darwin/build.sh
JoshuaSBrown: We can add a bunch of different test cases in a separate pull request, I just want to get something working in this one.

JoshuaSBrown:
> I have two additional follow up questions beyond the ones in the code:
> 
>     1. Where does that huge performance difference (looks like >3x even for a single large MeshBlock) between MPI and serial come from in the plots you report?

This bears looking into but is beyond the scope of this pull request. I will create an issue. 
> 
>     2. Does the MPI version compile correctly (unfortunately I cannot see the CI output)? I'm asking as I didn't see any logic to change the c++ compiler for openmpi to nvcc_wrapper (e.g., through the `OMPI_CXX` var)

I will quote @AndrewGaspar here:

> CMake doesn't invoke the compiler wrappers - find_package(MPI ...) extracts the compiler flags from mpic++ and applies them to the final compile

pgrete:
> > ```
> > 2. Does the MPI version compile correctly (unfortunately I cannot see the CI output)? I'm asking as I didn't see any logic to change the c++ compiler for openmpi to nvcc_wrapper (e.g., through the `OMPI_CXX` var)
> > ```
> 
> I will quote @AndrewGaspar here:
> 
> > CMake doesn't invoke the compiler wrappers - find_package(MPI ...) extracts the compiler flags from mpic++ and applies them to the final compile

Interesting. I wasn't ware of this. I was probably mislead my not finding any reference to MPI related options in `compile_commands.json` when looking at it on Summit.
I may need to update the Summit build doc then as it currently mentions to set `OMPI_CXX` which is unnecessary when using `nvcc_wrapper` for as cxx compiler even with MPI.

CONTRIBUTING.md
JoshuaSBrown: This is a fair point. This is related https://github.com/lanl/parthenon/issues/270. It looks like some work needs to be done on running the tests with more than one gpu. After a discussion with @jonahm-LANL the advection-performance test will need to be updated to do this, and possibly some of the regression testing infrastructure.

JoshuaSBrown:
This pull request needs to wait until https://github.com/lanl/parthenon/pull/281 is merged. It turns out that the ci was not using the resources correctly.

JoshuaSBrown:
https://xcp-gitlab.lanl.gov/eap-oss/parthenon/pipelines/9557

JoshuaSBrown:
This PR is ready for review. Just as a reminder this pr does not attempt to add and test every feature it is just to cover our bases e.g. testing on the power9 nodes. I will be working on improving this once it gets merged. If you want to see additional functionality I welcome it and created an issue a while ago here: https://github.com/lanl/parthenon/issues/240.

.gitlab-ci-darwin.yml
```diff
@@ -0,0 +1,95 @@
+variables:
+  SCHEDULER_PARAMETERS: '--nodes=1 --partition=power9 --export=NONE'
+  GIT_SUBMODULE_STRATEGY: recursive
+  MODULE_CMAKE: "cmake/3.12.4"
+  MODULE_CLANG: "clang/8.0.1"
+  MODULE_COMPILER: "gcc/7.4.0"
+  MODULE_CUDA: "cuda/10.1"
```
Yurlungur, line 7: why `cuda/10.1` and not `cuda/10.2`?

.gitlab-ci-darwin.yml
```diff
@@ -0,0 +1,95 @@
...
+  NUM_GPU_DEVICES_PER_NODE: 2
+  NUM_MPI_PROC_TESTING: 2
+  OMP_NUM_THREADS: 40
+  CMAKE_BUILD_TYPE: "Release" 
+  Kokkos_ARCH_POWER9: "ON" 
+  Kokkos_ARCH_VOLTA70: "ON" 
+  Kokkos_ENABLE_CUDA: "ON" 
+  Kokkos_ENABLE_CUDA_UVM: "OFF"
+  Kokkos_ENABLE_OPENMP: "ON" 
+  PARTHENON_DISABLE_HDF5: "OFF"
```
Yurlungur: Perhaps it would be better to put these in a machine file?

Yurlungur (approved):
Overall looks good! A minor structural nitpick, which I consider non-blocking.

I don't love that the configure settings are in this `build_darwin.sh` script. I feel like these should be in a machine file, since we have that machinery.

Yurlungur:
Also you need to remove `WIP`

.gitlab-ci-darwin.yml
JoshuaSBrown: @jlippuner had issues with cuda/10.2 https://github.com/lanl/parthenon/pull/114.

JoshuaSBrown: Ah, it was related to kokkos support for cuda.

.gitlab-ci-darwin.yml
JoshuaSBrown: I will make that a second pr. I was thinking of added a machine config file.

.gitlab-ci-darwin.yml
JoshuaSBrown: https://github.com/lanl/parthenon/issues/327

.gitlab-ci-darwin.yml
JoshuaSBrown: https://github.com/lanl/parthenon/issues/326

.gitlab-ci-darwin.yml
Yurlungur: Oh I see. Okay. We can update later then, when we update kokkos.

JoshuaSBrown:
> I don't love that the configure settings are in this `build_darwin.sh` script. I feel like these should be in a machine file, since we have that machinery.

Noted, I will fix this when I add the config file in a separate pr.

.gitlab-ci-darwin.yml
Yurlungur: sounds good.

Yurlungur:
> > I don't love that the configure settings are in this `build_darwin.sh` script. I feel like these should be in a machine file, since we have that machinery.
> 
> Noted, I will fix this when I add the config file in a separate pr.

Sounds good.

Yurlungur:
@JoshuaSBrown you should probably remove the `WIP` from the title. Also there are some merge conflicts.

JoshuaSBrown:
https://xcp-gitlab.lanl.gov/eap-oss/parthenon/-/jobs/88725

JoshuaSBrown:
Now that the regression tests are passing this is good to go. @pgrete 

When this is merged the commits should definitely be squashed.

JoshuaSBrown:
https://xcp-gitlab.lanl.gov/eap-oss/parthenon/-/jobs/88726

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Added (new features/APIs/variables/...)
 - [[PR 335]](https://github.com/lanl/parthenon/pull/335) Support for project-relative `MACHINE_CFG` with `@PAR_ROOT@`
+- [[PR 200]](https://github.com/lanl/parthenon/pull/200) Adds support for running ci on power9 nodes.
```
pgrete: This should probably be moved to the new Infrastructure heading.

.gitlab-ci-darwin.yml
```diff
@@ -0,0 +1,95 @@
...
+  SCHEDULER_PARAMETERS: '--nodes=1 --partition=power9 --export=NONE'
+  GIT_SUBMODULE_STRATEGY: recursive
+  MODULE_CMAKE: "cmake/3.12.4"
+  MODULE_CLANG: "clang/8.0.1"
+  MODULE_COMPILER: "gcc/7.4.0"
+  MODULE_CUDA: "cuda/10.1"
+  MODULE_MPI: "openmpi/p9/4.0.2-gcc_7.4.0"
+  NUM_GPU_DEVICES_PER_NODE: 2
+  NUM_MPI_PROC_TESTING: 2
+  OMP_NUM_THREADS: 40
```
pgrete:
```suggestion
  OMP_NUM_THREADS: 1
```

As long as we don't have proper threading support I strongly suggest to set num threads to 1. On the machine I using using more threads significantly slows down the code (in addition to technically not being properly supported).

pgrete (approved):
I have two small final comments and then this can go squashed into developed.

PR #202, opened 2020-06-19
Remove Cell variable from kernels

pgrete:
## PR Summary

Fix  HtoD call issues.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


pgrete:
Ups. The updated Kokkos module was supposed to be part of the pgrete/partmgr branch for the threads/streams where we need the current develop branch of Kokkos which contain a bugfix in the partitionmaster I found along the way.
Would you prefer to move that change away from there and add it to the other branch or keep it here?

AndrewGaspar:
> Would you prefer to move that change away from there and add it to the other branch or keep it here?

I'd prefer we keep the kokkos change in the PR that requires it - so remove it from this PR.

pgrete:
> > Would you prefer to move that change away from there and add it to the other branch or keep it here?
> 
> I'd prefer we keep the kokkos change in the PR that requires it - so remove it from this PR.

done

PR #204, opened 2020-06-23
Add Physical Constants

brryan:
## PR Summary

This PR adds a new class, `PhysicalConstants`, that is `constexpr` and provides a number of common physical constants.  This class is templated on a struct of unit system conversion factors, allowing for multiple consistent unit systems without a lot of duplicate code. Both verbose, e.g. `SpeedOfLight`, and terse, e.g. `c`, names are provided for each constant.

This would allow multiple, independently developed physics packages based on Parthenon to conveniently use the same physical constants. Even small disagreements in constants inside physics calculations can cause insidious bugs.

There is some extra code defining static members required to make C++14 happy; this could go away when compiling in C++17 or greater.

This is based on the `PhysicalConstexprs` class in LANL's open source [Draco](https://github.com/lanl/Draco) library.

Hat tip to @Yurlungur for some earlier discussions that ended up streamlining this class.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


src/utils/constants.hpp
```diff
@@ -0,0 +1,228 @@
...
+// itself and others acting on its behalf a nonexclusive, paid-up, irrevocable worldwide
+// license in this material to reproduce, prepare derivative works, distribute copies to
+// the public, perform publicly and display publicly, and to permit others to do so.
+//========================================================================================
+
+#ifndef UTILS_CONSTANTS_HPP_
+#define UTILS_CONSTANTS_HPP_
+
+#include <cmath>
+#include <string>
```
AndrewGaspar, line 18: It doesn't look like this is ever used in this file?

brryan:
> Is there somewhere in the code where you could replace a constant with one of these? If so, could you use it there so that we have an example in the code?

@AndrewGaspar I actually don't know of a place in parthenon itself where this could be used since everything is physics-agnostic; I checked the examples too but they all look dimensionless. Do you have anything in mind? Otherwise, usage would just be as shown in the unit test, e.g.
```c++
PhysicalConstants<SI> pc;
double c = pc.c; // speed of light in m / s
double me = pc.me; // Electron mass in kg
```
Would it be OK to write this up in the documentation instead (I should do this anyway)?

src/utils/constants.hpp
brryan: Good catch, thanks

AndrewGaspar:
@brryan Yeah, documentation is fine.

src/utils/constants.hpp
```diff
@@ -0,0 +1,228 @@
...
+
+namespace parthenon {
+namespace constants {
+
+// Conversion factors from the SI unit system
+struct SI {
```
forrestglines, line 24: The SI units are order unity here because the physical constants are in SI in the `PhysicalConstants` class, correct? This doesn't constrain us to use SI units as order unity for the code units, right?

src/utils/constants.hpp
```diff
@@ -0,0 +1,228 @@
...
+  static constexpr double Charge = 2.997924580e9;           // Statcoulomb
+  static constexpr double Capacitance = 8.9831483395497e11; // Statfarad
+  static constexpr double Angle = 1.;                       // Radian
+};
+
+// Defines and encapsulates physical and mathematical constants in a purely
+// constexpr way, with support for multiple unit systems.
+template <typename UNITSYSTEM>
+class PhysicalConstants {
+ private:
```
forrestglines, line 50: I think these should be `protected` so that `PhysicalConstants can be extended. For my application I would need to add some constants that are only applicable for astrophysics.

jmstone:
Thanks for flagging me.  To be honest I am not sure what problem this is addressing.  I always use dimensionless units for so many good reasons (such as those insidious bugs mentioned in the PR), especially for relativistic flows where G=c=1 is natural.  I thought Parthenon was supposed to physics-free, so I guess this is really for the application codes?  But then maybe it should be a part of them directly?  In the end I suppose it doesn't matter (I am guessing the extra bloat in the executable if this class is not used is completely negligible).  But I do think this will need careful documentation to explain how it is to be used, and what is provided.

As an aside: what happened to sticking to the Google/Athena++ style guide where only classes and functions have names beginning with capital letters? This implementation mixes all different naming conventions.  It will be a shame if Parthenon fails to enforce any consistent source code style.

Sorry Ben if this comes across as critical!  I don't mean to be, especially if it is something that you might find extremely useful.  I am just focused on my own use-cases right now and what we need to get AthenaPK working.

Yurlungur:
Oh I completely forgot about the style guide! My apologies. I should have caught that in my review.

Regarding whether or not Parthenon is the right place for physical units to go: I admit I’m torn on this.  On the one hand, if one uses code units, one often needs to transform between different systems of code units for different physics modules,  or access physical units for use in certain subroutines. And if Parthenon is the the glue code that holds these modules together, providing a unit system doesn’t seem unreasonable to me.

On the other hand, leaving this for application code gives those codes flexibility to handle units the way they want. And i agree it isn’t obvious that this should belong in Parthenon. And we could end up with the unfortunate circumstance where the host code defines some units and Parthenon defines some others and units get split across an application codebase.

Not arguing one way or the other here. I defer to the judgement of others. Just my two cents as to the nature of the dilemna.

brryan:
@jmstone Hi Jim! Thanks for taking a thoughtful look at this. I'm embarrassed I violated the style guidelines, I will get that fixed.

I agree that Parthenon should be physics-free, and I'm not trying to bring physics in with this PR. This is really for the consistency-between-codes concern I mentioned above, which based on my experiences at LANL can be surprisingly expensive in terms of person-hours to address when it isn't enforced right from the beginning. This concern is probably less relevant to single codes built on top of parthenon, or single sets of packages developed by one team, but from the LANL perspective it could prevent a lot of headaches.

I figured putting a constants class in parthenon itself would be the most effective way to mitigate this consistency issue, but if this is really not in the spirit of parthenon there are other solutions.

I agree with you that there are advantages to just being in dimensionless units, but getting all potential LANL partners to buy in to this approach could require the expense of a lot of political capital.

src/utils/constants.hpp
brryan: Yes that's correct, SI is just the unit system used by the CODATA site where I copied and pasted the values here from. To keep your dynamical variables order unity you could use whatever unit system you want.

src/utils/constants.hpp
brryan: That's a good point I hadn't appreciated, thanks for bringing it up -- I've made this change.

brryan:
@forrestglines That's right, this PR isn't meant to impose a preferred set of units inside parthenon, or on codes that use parthenon. This is just so everyone can share a common definition of physical constants, whether or not they are using the same unit systems.

The `constexpr` vs. runtime constants issue you bring up is one I've run into myself previously (and forgot about until just now), and may actually be more severe in parthenon-based applications... apart from any potential performance advantages, we originally used `constexpr` in Draco because it was convenient when working in cuda, but I am not aware that that issue appears in kokkos. That all being said, flexibility in the physical constants provider is probably more important than marginal performance improvements (if any) and I'm now thinking this *shouldn't* be `constexpr`, for the concern you raise. Happy to hear other thoughts on this.

jmstone:
Ben, Jonah: no worries and thanks for addressing the style issues.  I know they are easy to forget but my experience is they really help in the long run.  And I think I understand better the need for units as an internal LANL issue based on what you said.  Us astrophysicists can just not use it, and it can still serve a purpose for internal applications.  So I'll let you guys decide what is the best strategy for your own uses.

pgrete:
I think having a standardized way of accessing physical variable (that may or may not be used by derived applications) is a good idea.
As long as this doesn't enforce anything on the algorithms implemented in Parthenon itself (which is currently the case), I see this as a plus.

Thinking one step ahead, it may be worth adding the *option* to add units, for example, to the outputs so that  a user/derived application can opt in for that.
This, for example, would allow analysis tool like yt  to directly process (and convert, if required) data without the user needed to specify a set of units to be used for a dataset externally.

Yurlungur:
> Thinking one step ahead, it may be worth adding the option to add units, for example, to the outputs so that a user/derived application can opt in for that.

This is a great idea.

brryan:
@forrestglines So your point about runtime modification of unit systems is well taken, but I've spent a while playing around with this without much success. Basically, I'm having a hard time balancing the const nature of a unit system with exposing runtime modification to individual packages/drivers. Solutions that I think would work tend to have unfortunate syntax, like doing all the unit conversions in a monster initializer list, creating custom runtime unit conversion structs by passing the unit conversions all in a row to a custom struct (e.g. `(1., 1., 1., 1.4839459493e-20, 1., 1., 5.2038)` etc., which I think would be very error-prone. 

I think this is suggesting that how runtime-manipulatable one wants physical constants to be is a decision best left to individual applications, and that the class in this PR would be most valuable just providing actual values in standard unit systems to be converted into whatever data structure the application uses for runtime manipulation if that is desired. There may also be some performance advantages to maintaining the `constexpr`ness for applications that can leverage it.

Does that seem reasonable? I'm happy to discuss, and maybe I missed some easy win, but I'm thinking there's no perfect solution here.

@pgrete Interesting point about optional units for output.

forrestglines:
> @forrestglines So your point about runtime modification of unit systems is well taken, but I've spent a while playing around with this without much success. Basically, I'm having a hard time balancing the const nature of a unit system with exposing runtime modification to individual packages/drivers. Solutions that I think would work tend to have unfortunate syntax, like doing all the unit conversions in a monster initializer list, creating custom runtime unit conversion structs by passing the unit conversions all in a row to a custom struct (e.g. `(1., 1., 1., 1.4839459493e-20, 1., 1., 5.2038)` etc., which I think would be very error-prone.

Thanks for taking the time to try to get it to work. That sounds about right though, and having it modifiable at run time means that `constexpr` wouldn't work.

> I think this is suggesting that how runtime-manipulatable one wants physical constants to be is a decision best left to individual applications, and that the class in this PR would be most valuable just providing actual values in standard unit systems to be converted into whatever data structure the application uses for runtime manipulation if that is desired. There may also be some performance advantages to maintaining the `constexpr`ness for applications that can leverage it.
> 
> Does that seem reasonable? I'm happy to discuss, and maybe I missed some easy win, but I'm thinking there's no perfect solution here.


I think that sounds reasonable. For my applications at least, I'll never be using MKS or CGS for code units, but this class would still be useful for getting the units in MKS. I'll probably end up using this class to make a user changeable one inside of my problem generator.

I do lean a bit towards leaving these physical constants out of Parthenon and putting them in the application codes as needed, but including this class in Parthenon is acceptable.

brryan:
Thanks all for looking at this. I dealt with the style issues (@Yurlungur do you agree I got it right?), added documentation, and assuming the CI passes I think this is ready to merge.

docs/constants.md
```diff
@@ -0,0 +1,25 @@
+# Parthenon Built-in Physical Constants
+
+PhysicalConstants is a class that provides numerical values of physical
+constants in common unit systems (SI and CGS). This is provided in the hopes
+that parthenon-derived applications can conveniently share exactly matching
+values for these constants, to avoid bugs that can be associated with
+disagreement and the effort required to synchronize a consistent move to newer
+values of physical constants.
+
+Internal values are hardcoded in SI units, and specific unit systems are
```
Yurlungur, line 10: SI as the base unit system isn't unreasonable. I'm curious why you chose that over cgs though.

docs/constants.md
```diff
@@ -0,0 +1,25 @@
...
+
+Both verbose and terse names are provided for each constant as public (albeit
+constexpr) members.
+
+### Usage
+
+To construct, call this class with PhysicalConstants<UNITSYSTEM>(), where
+UNITSYSTEM is a struct of conversion factors from SI units; both SI and CGS unit
+systems are provided in the parthenon::constants namespace. Then, to access
+constants simply use the public data members.
```
Yurlungur, line 25: I think it would be good to see an example of this. e.g., a block of code like
```C++
PhysicalConstants<CGS> pc();
std::cout << "Planck's constant = " << pc.h << std::endl;
```
it might also be good to explain how to define a `UNITSYSTEM`. e.g.,
```C++
struct MyUnitSystem {
 // whatever is appropriate
}
```
What fields does the struct need to define? I think more of the API still needs to be clarified here.

Maybe you could also link to your unit tests file for example usage?

docs/constants.md
brryan: SI is just the units they're provided in at https://physics.nist.gov/cuu/Constants/

Yurlungur: Ah okay. Yeah that makes sense.

brryan:
@AndrewGaspar Are you happy with the new documentation? Looks like you still have Changes Requested

Yurlungur:
@AndrewGaspar is on leave. This might have to wait until he's back.

Yurlungur:
@brryan I think we're good to go as soon as you update to master

AndrewGaspar:
@github let me know when the CI passes, please, thanks

brryan:
Thanks @AndrewGaspar

AndrewGaspar:
GitHub never emailed me smdh.

PR #205, opened 2020-06-25
WIP: Poisson Solver

Yurlungur:
## PR Summary

I am experimenting with introducing a basic Jacobi solver to solve the Poisson equation. This will require both changes to the task list and possibly to things like communication. This is very much in-progress and not ready for review.

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted


Yurlungur:
A first draft now functions. But the code is not synced with master and there's much more work to be done.

As a first pass, I solve the Poisson equation using second-order finite differences and a Jacobi smoother. I just use periodic boundary conditions and solve an equation of the form 

lapplace(phi) = k rho

and I choose k = 1 and rho = 1 for r < r0, else 0.

I define a new driver, an `IterationDriver` to  iterate through the algorithm. I was pleasantly surprised to find the code worked out of the box with MPI and multiple blocks, although mesh refinement does not work. (Likely because of the second-order finite differences form of the equations. A form with a divergence of a "flux" will probably behave.)

Here's a plot of the solution after the solver converged to a tolerance of 5%:


Some things I would like to try:
1. Non-periodic boundary conditions
2. Integrating the solver with, e.g., advection, either in a new driver or with a new mechanism such as sub-tasks
3. Get AMR working

berselius:
Hey Jonah - I'm interested in using this for the kinetic plasma code I'm building on top of Parthenon. Is there also a 1D flavor of this? It could make my testing a lot easier.

Yurlungur:
@berselius it should already work in 1D. I at least did some preliminary tests and it seems to behave. I’m not sure it’s really ready for being used for implicit updates though. The tasking probably still needs some work for that. Sleep i uses the simplest, stupidest possible solver, weighted Jacobi, so convergence is slow.

Happy to talk at some point and see if we can push it in a direction that’s useful for you.

jlippuner:
Deprecated by #556

PR #210, opened 2020-06-30
Clean up error checking in params

Yurlungur:
## PR Summary

This is a minor pull request that just cleans up the key existence error checking machinery in `params.hpp` to use the `PARTHENON_REQUIRE` machinery introduced by @brryan. I also enable that error checking by default, after hitting a cryptic segfault with it disabled.

The performance implications of enabling these checks by default should be minimal. To check return type, the code must search an extra `std::map` of types. To check existence, no extra searches are required.

There is a cost to this rewritten check in that the code dies with an `exit` rather than throws an error, so one test that checks for an error had to be removed and one had to be rewritten.

I also slightly modified the type checking machinery to be a bit more concise. But the overall structure is the same.

## PR Checklist

- [x] Code passes cpplint
- [x] Adds a test for any bugs fixed. Adds tests for new features. (No new features.)
- [x] Code is formatted


tst/unit/test_unit_params.cpp
```diff
@@ -39,41 +39,33 @@ TEST_CASE("Add and Get is called", "[Add,Get]") {
       REQUIRE_THROWS_AS(params.Get<int>(key), std::invalid_argument);
     }
   }
-
-  GIVEN("An empty params structure") {
-    Params params;
-    WHEN(" attempting to get a key that does not exist ") {
-      std::string non_existent_key = "key";
-      REQUIRE_THROWS_AS(params.Get<double>(non_existent_key), std::invalid_argument);
```
JoshuaSBrown, line 47: Are you no longer checking what happens when accessing a key that no longer exists?

Yurlungur: I can't test for this, because `PARTHENON_REQUIRE` causes the code to die with a useful error message but without throwing an error. I could replace the `PARTHENON_REQUIRE` call with an error throw if that's preferred. Then this test could be added again.

JoshuaSBrown: If it's a performance issue I would leave it as it is. I tend to use exceptions, if it a class the end user is likely to have some kind of access too (I would expect the params class would fit into this category), and assertion type checks for internal checks. However, because of the performance issues with exceptions I keep getting corrected on my suggestions related to this issue :). I was more just curious.

Yurlungur: I don't think this is a performance-critical piece of code. So exceptions are probably fine here. I can change it back. The code will still be cleaner than it was.

JoshuaSBrown: I would see what @nmsriram says, before doing any more work, as it is, it is an improvement.

src/interface/params.hpp
```diff
@@ -94,31 +90,11 @@ class Params {
...
-      std::cout << "----------------------------------------------" << std::endl;
-      std::cout << std::endl;
-
-      std::string message = "Key " + key + " doesn't exist";
-      PARTHENON_DEBUG_REQUIRE(!die, message.c_str());
-    }
-#endif // NDEBUG
+    bool badtype = myTypes_[key].compare(std::string(typeid(T).name()));
+    std::string message = "WRONG TYPE FOR KEY '" + key + "'";
+    PARTHENON_REQUIRE(!badtype, message.c_str());
```
nmsriram, line 97: Since you've reduced this to a single line, does it make sense to eliminate the typeCheck() procedure and instead absorb it above where it is called (since that is the only instance?)

tst/unit/test_unit_params.cpp
nmsriram: I agree with @Yurlungur that this is not a performance critical piece of code, since it should only be done once per physics.  Testing things that kill the code are always problematic and I'm not sure I know of any testing frameworks that handle it better.  

Is it possible to capture this by using std::terminate functionality so that the Kokkos abort would eventually end up in a function we designate?

pgrete: Could we address this with https://cmake.org/cmake/help/v3.7/prop_test/WILL_FAIL.html ?
Overall, I'm generally in favor of adding test rather than removing ;)

Yurlungur: @pgrete @JoshuaSBrown @nmsriram I don't know what the performance implications of `std::terminate` are... and at a cursory pass, I don't see how to get `WILL_FAIL` to behave properly with `Catch2`. How about instead I incorporate error throwing into error checking? I.e., I add a new `PARTHENON_REQUIRE_THROWS` and `PARTHENON_FAIL_THROWS` which throws errors?

JoshuaSBrown: @Yurlungur I'm in favor of your suggestion.

src/interface/params.hpp
Yurlungur: Yes I'll fix that.

Yurlungur:
> I think those changes make sense (and the code cleaner).
> One additional item I noticed that the
> 
> ```
>       throw std::invalid_argument("Key value pair already exists, cannot add key.");
>     }
> ```
> 
> is still there. Wouldn't a `PARTHENON_REQUIRE` also make sense there?

This should be resolved now.

tst/unit/test_unit_params.cpp
Yurlungur: @JoshuaSBrown @nmsriram @pgrete can you take another look? I've added the ability to throw in `error_checking.hpp`.

JoshuaSBrown:
I feel much better about this now. @Yurlungur thanks for taking the time to refactor.

docs/README.md
```diff
@@ -53,15 +53,21 @@ There are several weakly linked member functions that applications can (and ofte
...
-that these macros take a C style string, not a C++ style string. This
-is a limitation of GPU compatibility. Examples of use can be found
-[here](../tst/unit/test_error_checking.cpp).
+* PARTHENON_DEBUG_THROW(message) throws a runtime error when in debug mode.
+
+All macros print the message, and filename and line number where the
+macro is called. PARTHENON_REQUIRE also prints the condition. The
+macros take a `std::string`, a `std::stringstream`, or a C-style
+string. As a rule of thumb:
+- Use the exception throwing versions in non-GPU, non-performance critical code.
```
pgrete, line 68:
```suggestion
- Use the exception throwing versions outside of kernels/parallel regions and non-performance critical code.
```

docs/README.md
```diff
@@ -53,15 +53,21 @@ There are several weakly linked member functions that applications can (and ofte
...
+- On GPUs and in performance-critical sections, use the non-throwing
```
pgrete, line 69:
```suggestion
- Within parallel regions and in performance-critical sections, use the non-throwing
```

Yurlungur:
Pulling the trigger.

PR #212, opened 2020-07-02
Collection of necessary updates/fixes

jdolence:
## PR Summary

The changes introduced in this PR were required to get LANL's application code working.  Most of the changes are related to `SparseVariable`s.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


src/interface/container.cpp
```diff
@@ -405,8 +404,32 @@ template <typename T>
...
+        auto &svec = (sit->second)->GetVector();
+        for (auto s = svec.rbegin(); s != svec.rend(); ++s) {
+          vars.push_front(*s);
+        }
+      }
+    }
+  }
+  for (auto &v : vars) {
+    expanded_names.push_back(v->label());
+  }
```
Yurlungur, line 432: Ah, too bad the subcontainer strategy didn't work out.

src/interface/variable_pack.hpp
```diff
@@ -38,7 +38,23 @@ using IndexPair = std::pair<int, int>;
...
+    }
+    return map_[key];
+  }
+  void insert(std::pair<std::string, vpack_types::IndexPair> keyval) {
+    map_[keyval.first] = keyval.second;
+  }
+
+ private:
+  std::map<std::string, vpack_types::IndexPair> map_;
+};
```
Yurlungur, line 57: If you use `map::operator[]` and element doesn't exist, `std::map` constructs one. I was surprised to discover there isn't a way to set the default element of a `std::map`. It always constructs the object with the default constructor if a key doesn't exist. This leads to two solutions:
1. Do what you did and wrap `std::map` with an operator that does the thing we want and uses the default we want
2. Create a new `IndexPair` object that does what we want. i.e., something like this:
```C++
namespace vpack_types {
class IndexPair {
 public:
  IndexPair() : first(0), second(-1) {}
  IndexPair(int a, int b) : first(a), second(b) {}
  int first, second;
};
} // vpack_types
```
then 
```C++
using PackIndexMap = std::map<std::string, vpack_types::IndexPair>;
```
would produce the right default.

I'm not sure which path is the better one. They each have trade-offs. Option 1 limits what the map can do. Option 2 limits what the pair can do, as every function needed must be hand-implemented. What you've done is fine, I think.

jdolence:
@nmsriram @AndrewGaspar can one of you take a look at this when you get a chance?

src/interface/variable_pack.hpp
```diff
@@ -142,17 +166,43 @@ VariableFluxPack<T> MakeFluxPack(const vpack_types::VarList<T> &vars,
...
+      }
+      if (sparse_name != sparse_trim) {
+        vmap->insert(std::pair<std::string, IndexPair>(
+            sparse_name, IndexPair(sparse_start, vindex - 1)));
+        sparse_name = sparse_trim;
+        sparse_start = vindex;
+      }
+
+    } else if (!(sparse_name == "")) {
+      vmap->insert(std::pair<std::string, IndexPair>(
```
JoshuaSBrown, line 197: How do you know that vmap is not still a nullptr here?

src/interface/variable_pack.hpp
```diff
@@ -162,9 +212,33 @@ VariableFluxPack<T> MakeFluxPack(const vpack_types::VarList<T> &vars,
...
+          sparse_name, IndexPair(sparse_start, vindex - 1)));
+      sparse_name = "";
+    }
```
JoshuaSBrown, line 241: The MakeFluxPack function is really large, I wish it could be broken up into subfunctions which would make it more readable.

src/interface/variable_pack.hpp
```diff
@@ -181,11 +255,16 @@ VariableFluxPack<T> MakeFluxPack(const vpack_types::VarList<T> &vars,
           std::pair<std::string, IndexPair>(v->label(), IndexPair(vstart, vindex - 1)));
     }
   }
+  if (!(sparse_name == "")) {
```
JoshuaSBrown, line 258: This is more preference but I would replace lines like these with. 
```
if ( not sparse_name.empty() ) 
```
I think it is a little bit more readable.

src/interface/variable_pack.hpp
```diff
@@ -200,17 +279,29 @@ VariablePack<T> MakePack(const vpack_types::VarList<T> &vars,
...
+      sparse_trim.erase(sparse_trim.find_last_of("_"));
       if (sparse_name == "") {
-        sparse_name = v->label();
-        sparse_name.erase(sparse_name.find_last_of("_"));
+        sparse_name = sparse_trim;
         sparse_start = vindex;
       }
+      if (sparse_name != sparse_trim) {
+        vmap->insert(std::pair<std::string, IndexPair>(
+            sparse_name, IndexPair(sparse_start, vindex - 1)));
```
JoshuaSBrown, line 300: Again I don't see where you are checking if vmap is a nullptr?

src/interface/sparse_variable.cpp
```diff
@@ -28,7 +28,9 @@ void SparseVariable<T>::Add(int varIndex) {
     }
     // create the variable and add to map
     std::string my_name = label_ + "_" + std::to_string(varIndex);
-    auto v = std::make_shared<CellVariable<T>>(my_name, dims_, metadata_);
+    Metadata metadata = metadata_;
+    metadata.SetSparseId(varIndex);
+    auto v = std::make_shared<CellVariable<T>>(my_name, dims, metadata);
```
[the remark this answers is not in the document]
jdolence: good catch.  removed.

src/interface/variable_pack.hpp
jdolence: we don't.  good catch.  fixed.

src/interface/variable_pack.hpp
jdolence: broke it up.

Issue #207, opened 2020-06-25
Install option for building as standalone library


brryan:
Right now my understanding is that to use parthenon in an application one must bring in the source code as e.g. a `git submodule` and then build parthenon and the application together. It could be nice to be able to add an install option to the cmake build system so one can just link in a standalone parthenon install. Three ways this could be convenient:
* A physics code built out of two parthenon-based packages doesn't have to worry about each of these packages having their own parthenon source via two `git submodule`s.
* Not having to worry about parthenon's build rules, as in #206 
* Allowing for `spack install parthenon` :)
This may have already been brought up before but I didn't see an open issue and #206 reminded me of this.

JoshuaSBrown:
Are you suggesting a precompiled shared library and or a spack implementation?

Yurlungur:
I think @brryan is suggesting that the former enables the latter.

brryan:
@Yurlungur @JoshuaSBrown Yes once Parthenon can be built as a shared library, I think adding spack support should just be a ~20 line PR to the spack repository.

AndrewGaspar:
Yeah, if you have a "standard" CMake package (at least as standard you can get in a turing complete build system), adding it to Spack is trivial.

Yurlungur:
Not only is this possible, we're considering disabling it.

PR #214, opened 2020-07-04
Convert weak linking to function pointers

brryan:
## PR Summary

Parthenon currently uses weak linking to allow driver applications to optionally replace the default internal functions:
```
ParthenonManager::ProcessProperties
ParthenonManager::ProcessPackages
ParthenonManager::SetFillDerivedFunctions
Mesh::InitUserMeshData
Mesh::UserWorkInLoop
Mesh::UserWorkAfterLoop
MeshBlock::InitApplicationMeshBlockData
MeshBlock::InitUserMeshBlockData
MeshBlock::ProblemGenerator
MeshBlock::UserWorkInLoop
MeshBlock::UserWorkBeforeOutput
```

This works fine for internal example applications. However, when building a separate application (or an application that uses multiple separate physics packages that each compile their own drivers) or when building parthenon as a shared library in the future (Issue #207) this can/probably would cause issues that are not always trivial to fix, especially for non-cmake experts.

An alternative would be to use `std::function` function pointers instead, that are defined with the usual default functions by Parthenon itself and can be redefined by user applications. I don't know that the extra pointer dereference would cause any performance issues when used with the above functions.

I've provided one example of this change with `SetFillDerivedFunctions` as used in the advection example. Otherwise leaving this as a WIP now to see what people think. In particular, maybe I am just mistaken about weak linking being as much of a deal as I make it out to be here. But I've been having trouble with this model and others might as well.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


example/advection/main.cpp
```diff
@@ -20,6 +20,9 @@ int main(int argc, char *argv[]) {
   using parthenon::ParthenonStatus;
   ParthenonManager pman;
 
+  // Redefine parthenon defaults
+  pman.SetFillDerivedFunctions = &(advection_example::SetFillDerivedFunctions);
```
Yurlungur: Ah, putting it in the manager is a clever trick.

brryan:
@berselius Tagging you since I believe youre developing a parthenon app and if this goes in it would probably break your code

berselius:
> @berselius Tagging you since I believe youre developing a parthenon app and if this goes in it would probably break your code

Thanks, Ben! I was just starting to wonder the same thing.

brryan:
> > @berselius Tagging you since I believe youre developing a parthenon app and if this goes in it would probably break your code
> 
> Thanks, Ben! I was just starting to wonder the same thing.

If it's any consolation this will break my app too -- should be a quick fix though if/when this goes in, just set the non-default function pointers in `ParthenonManager` in main before you call `ParthenonManager::ParthenonInit` like in the advection example here. You could also move your custom functions out of the `parthenon` namespace.

brryan:
This ended up being a more substantial changeset than I anticipated -- custom functions pointed to by function pointers can't in general be members of e.g. `Mesh`, so I added a `FunctionInput` class that is passed around mostly in parallel with `ParameterInput` and contains function pointers for all user-replaceable parthenon functions. These default to a `nullptr` value, and unless they're explicitly set Parthenon just sets its internal function pointers to default versions of these functions. This *does* mean that user-replaceable functions only have access to public data members of `ParthenonManager`, `Mesh`, and `MeshBlock`, but I think this is reasonable. Does this approach seem reasonable to everyone? @jdolence suggested a more general `AppDescriptor` that `FunctionInput` might want to metamorphose into at some point.

AndrewGaspar:
I was actually trying to build parthenon for Windows the other day and hit this, so thank you for trying to remove the weak link directives.

src/function_input.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+#include "defs.hpp"
+#include "interface/properties_interface.hpp"
+#include "interface/state_descriptor.hpp"
+#include "parameter_input.hpp"
+#include "parthenon_arrays.hpp"
+
+namespace parthenon {
+
+class FunctionInput {
```
Yurlungur: I'd make this a `struct` to clarify that it's entirely public scope.

src/function_input.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+ public:
+  FunctionInput() {}
```
Yurlungur: Should we include a non-trivial constructor? Probably not necessary.

example/advection/main.cpp
```diff
@@ -20,6 +20,12 @@ int main(int argc, char *argv[]) {
   using parthenon::ParthenonStatus;
   ParthenonManager pman;
 
+  // Redefine parthenon defaults
+  pman.finput->ProcessPackages = &(advection_example::ProcessPackages);
+  pman.finput->ProblemGenerator = &(advection_example::ProblemGenerator);
+  pman.finput->UserWorkAfterLoop = &(advection_example::UserWorkAfterLoop);
+  pman.finput->SetFillDerivedFunctions = &(advection_example::SetFillDerivedFunctions);
```
Yurlungur: I'm pretty sure `std::function` lets you set them without the pointer syntax. i.e.,
```C++
pman.finput.ProcessPackages = advection_example::ProcessPackages;
```
At least, that's what [this example](https://en.cppreference.com/w/cpp/utility/functional/function) seems to imply.

example/advection/redefine_parthenon_defaults.cpp
```diff
@@ -211,4 +207,16 @@ void Mesh::UserWorkAfterLoop(ParameterInput *pin, SimTime &tm) {
...
+  packages[pkg->label()] = pkg;
+  return packages;
+}
+
+void SetFillDerivedFunctions() {
+  parthenon::FillDerivedVariables::SetFillDerivedFunctions(advection_package::PreFill,
+                                                advection_package::PostFill);
+}
+
+} // namespace advection_example
```
Yurlungur, line 222: Does the filename `redefine_parthenon_defaults` make sense anymore? I don't know of a better name though

example/calculate_pi/pi_driver.cpp
```diff
@@ -39,7 +46,7 @@ int main(int argc, char *argv[]) {
     return 1;
   }
 
-  PiDriver driver(pman.pinput.get(), pman.pmesh.get());
+  PiDriver driver(pman.pinput.get(), pman.finput.get(), pman.pmesh.get());
```
Yurlungur: It suddenly strikes me as a little weird that we call these peices of the manager and pass them into the driver. Maybe we should just pass the whole manager in?

example/face_fields/face_fields_example.cpp
```diff
@@ -23,7 +23,7 @@
 
 namespace parthenon {
```
Yurlungur: should this be in the parthenon namespace now?

src/defs.hpp
```diff
@@ -137,6 +138,19 @@ using MGBoundaryFunc = void (*)(ParArrayND<Real> &dst, Real time, int nvar, int
                                 int ie, int js, int je, int ks, int ke, int ngh, Real x0,
                                 Real y0, Real z0, Real dx, Real dy, Real dz);
 
+//----------------------------------------------------------------------------------------
+// Opaque pointer to application data
+class MeshBlockApplicationData {
```
Yurlungur, line 143: This seems useful, but I don't quite see it at the moment. What's the intended use?

src/defs.hpp
```diff
@@ -137,6 +138,19 @@ using MGBoundaryFunc = void (*)(ParArrayND<Real> &dst, Real time, int nvar, int
...
+ public:
+  // make this pure virtual so that this class cannot be instantiated
+  // (only derived classes can be instantiated)
+  virtual ~MeshBlockApplicationData() = 0;
```
Yurlungur, line 147: Is making the destructor pure virtual the standard way to make an empty abstract base class? Something about this strikes me as weird. I'd rather make the default constructor virtual and pure.

src/parthenon_manager.cpp
```diff
@@ -77,6 +77,17 @@ ParthenonStatus ParthenonManager::ParthenonInit(int argc, char *argv[]) {
...
+  if (finput->SetFillDerivedFunctions != nullptr) {
+    SetFillDerivedFunctions = finput->SetFillDerivedFunctions;
+  }
+  if (finput->ProcessProperties != nullptr) {
+    ProcessProperties = finput->ProcessProperties;
+  }
+  if (finput->ProcessPackages != nullptr) {
+    ProcessPackages = finput->ProcessPackages;
+  }
+
```
Yurlungur: hmm. Could we just have the defaults be explicitly set in the fields of the `finput` struct when it's defined? Currently they're defaulted to `nullptr` but we could use the explicit defaults we want maybe?

src/function_input.hpp
brryan: My guess is a useful function signature for a nontrivial constructor would have way too many input values, so a typical application would have to pass a long, error-prone list of `nullptr`s.

I removed this default constructor but I can put it back if that's not OK

Yurlungur: No I agree. I was just musing. The way you have it now seems entirely reasonable to me.

example/advection/main.cpp
brryan: Yes that's right, my mistake, I'll get rid of the `&`

example/advection/redefine_parthenon_defaults.cpp
brryan: I think it's at least close to making sense... `internal_parthenon_functions`? I don't know, nothing really great is coming to my mind right now.

example/calculate_pi/pi_driver.cpp
brryan: I think that's a good idea, and would clean up the internal function signatures -- there's no case where you would want to call driver with the manager's pinput but a custom mesh, right?

example/advection/redefine_parthenon_defaults.cpp
Yurlungur: maybe something like `parthenon_app_settings`?

example/calculate_pi/pi_driver.cpp
Yurlungur: I can imagine wanting a custom mesh object, but if we had a custom mesh object, I imagine lots more would go wrong than this.

example/face_fields/face_fields_example.cpp
brryan: I moved the other facefields-specific stuff from `parthenon` to `FaceFields` namespaces too.

src/defs.hpp
brryan: I don't know what this is, I just saw that it was used by a function that could be weakly linked. I had to move it out of `mesh.hpp` to avoid a circular dependency.

src/defs.hpp
brryan: See above, I don't have any background on this class.

src/defs.hpp
Yurlungur: Oh my bad. I thought you wrote that.

src/parthenon_manager.cpp
brryan: As of now the defaults are member functions of their respective classes (`Mesh` etc.) and since at least `ParthenonManager` includes the `FunctionInput` class, I think you may get in trouble with circular dependencies trying switch to this. I think that would work but it would only save you so many lines (and classes like `Mesh` would spend at least some time with null function pointers whereas right now they are set in the class declaration to the defaults)

Yurlungur: hmm ok fair enough

example/advection/redefine_parthenon_defaults.cpp
brryan: Gone with `parthenon_app_inputs.cpp` for now

example/calculate_pi/pi_driver.cpp
brryan: Still think this is a good idea but potentially wider-reaching than this PR so I'm inclined to leave this for later. I'll make a github issue to keep a record of this.

Yurlungur: Works for me.

Yurlungur:
@brryan is this ready for full review? I see it's not WIP anymore.

brryan:
@Yurlungur I just need to add some documentation, then I think this is ready. Worth making sure enough people are happy with this, since it's a sizeable change to the interface apps will see, or that existing apps already expect.

berselius:
Now that I'm finally up to date with parthenon/master, I'll give this a spin with my app and see how much of a pain point it is development-wise for ignorant app developers...

brryan:
@berselius Thanks! The `main()` functions of the parthenon examples in this branch should exhibit the usage pretty well

src/application_input.hpp
```diff
@@ -0,0 +1,54 @@
...
+#include "interface/state_descriptor.hpp"
+#include "parameter_input.hpp"
+#include "parthenon_arrays.hpp"
+
+namespace parthenon {
+
+struct ApplicationInput {
+ public:
+  // ParthenonManager functions
+  std::function<void()> SetFillDerivedFunctions = nullptr;
```
AndrewGaspar, line 33: Ah, it's a little too late for this, I guess, but I think this is a good candidate for just being a virtual base class rather than using `std::function`s, unless there's some optimization that can be taken if you know the user doesn't provide a given function.

If you _were_ to change it, I would provide default implementations of all these that do nothing, and then users just override the routines that they need to.

berselius:
FWIW I'm getting errors ala

```
/home/haack/EAP/test_weak/zapp/src/main.cpp: In function ‘int main(int, char**)’:
/home/haack/EAP/test_weak/zapp/src/main.cpp:25:45: error: ‘ProcessPackages’ is not a member of ‘vlasov’
   pman.app_input->ProcessPackages = vlasov::ProcessPackages;
                                             ^~~~~~~~~~~~~~~
/home/haack/EAP/test_weak/zapp/src/main.cpp:26:46: error: ‘ProblemGenerator’ is not a member of ‘vlasov’
   pman.app_input->ProblemGenerator = vlasov::ProblemGenerator;
                                              ^~~~~~~~~~~~~~~~
/home/haack/EAP/test_weak/zapp/src/main.cpp:27:47: error: ‘UserWorkAfterLoop’ is not a member of ‘vlasov’
   pman.app_input->UserWorkAfterLoop = vlasov::UserWorkAfterLoop;
                                               ^~~~~~~~~~~~~~~~~
/home/haack/EAP/test_weak/zapp/src/main.cpp:28:53: error: ‘SetFillDerivedFunctions’ is not a member of ‘vlasov’
   pman.app_input->SetFillDerivedFunctions = vlasov::SetFillDerivedFunctions;
                                                     ^~~~~~~~~~~~~~~~~~~~~~~
/home/haack/EAP/test_weak/zapp/src/main.cpp:28:53: note: suggested alternatives:
In file included from /home/haack/EAP/test_weak/zapp/src/main.cpp:16:0:
/home/haack/EAP/test_weak/zapp/src/vlasov/vlasov.hpp:42:6: note:   ‘SetFillDerivedFunctions’
 void SetFillDerivedFunctions();
      ^~~~~~~~~~~~~~~~~~~~~~~
In file included from /home/haack/EAP/test_weak/zapp/parthenon/src/mesh/mesh.hpp:41:0,
                 from /home/haack/EAP/test_weak/zapp/parthenon/src/driver/driver.hpp:26,
                 from /home/haack/EAP/test_weak/zapp/parthenon/src/parthenon_manager.hpp:22,
                 from /home/haack/EAP/test_weak/zapp/src/main.cpp:14:
/home/haack/EAP/test_weak/zapp/parthenon/src/interface/update.hpp:40:6: note:   ‘parthenon::FillDerivedVariables::SetFillDerivedFunctions’
 void SetFillDerivedFunctions(FillDerivedFunc *pre, FillDerivedFunc *post);
      ^~~~~~~~~~~~~~~~~~~~~~~

```

When I try to build in a similar fashion as you did in the advection example. Entirely possible that this is just me being dumb about namespaces, but I don't see any show stoppers.

berselius:
Nevermind, I just forgot to add these to the driver header file. Rolling now!

brryan:
@berselius Glad it's working for you, sorry I didn't reply sooner. Thanks for testing this PR out!

brryan:
@Yurlungur documentation done, look good to you? One option would be to add specific descriptions of each app-redefinable function. 

@Yurlungur @AndrewGaspar I'm inclined to stick with `std::function`s over an inheritance model here, since the user only has to write a function rather than an inherited class and a function, and also because we already have signoff on/implementation of `std::function` from multiple stakeholders. Also as you say it's already done :)

If everyone is OK with all that, I think this is ready to go in assuming the CI passes (I just merged `master` back in)

AndrewGaspar:
@Yurlungur can you sign off?

JoshuaSBrown:
@brryan is there a reason you haven't merged this? Are you still working on it? It looks like everything has been approved barring the ci successfully completes once you update the branch.

brryan:
@JoshuaSBrown Thanks for the ping, this is good to go, ~~merging now~~ merging once I update and CI passes. Just stuck in the mindset of not merging my own PRs from another project I'm on.

PR #216, opened 2020-07-08
Move gold files into git lfs

AndrewGaspar:
## PR Summary

This change moves the gold files into [git lfs](https://git-lfs.github.com/). This provides a more standard method for managing the gold files, with automatic checkout of the gold files when changing branches. The nice thing about `git lfs` is that git (via hooks) manages the gold standard files.

Unfortunately, it's not completely automatic - you need to run `git lfs install` at least once before git-lfs will start managing the lfs added files. However, CI seems to automatically be setup to consume lfs files.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


JoshuaSBrown:
I have worked with git lfs before and am not a huge fan, unless they have improved it much since I last used it. I honestly had more success with simply storing files on google drive and downloading them with curl. What is your reason for wanting to use lfs?

AndrewGaspar:
I've never used git lfs before, so I'm not familiar with the issues with it. I like git lfs for two reasons:
1) the files are still hosted in GitHub, so anybody on the project can update them.
2) git-lfs more-or-less automatically checks out the correct files when changing branches

I'm open to better solutions, but rolling out own git hook solution doesn't sound that different from what git lfs is doing?

JoshuaSBrown:
These are fair points. I will say this though, there is a limit on how much space you can use with gitlfs which is not that large, before you have to pay for it. 

> If you use more than 1 GB of bandwidth per month without purchasing a data pack, Git LFS support is disabled on your account until the next month.

If you use something like google drive by default you start with 15gb, and I'm not aware of a limit on the bandwidth. This can be problematic if we are pulling from git lfs in the ci... you can quickly eat up the bandwidth.

AndrewGaspar:
Ah, seriously? That's pretty messed up. Hm... any pointers to ready-made solutions?

JoshuaSBrown:
I wrote my own script to pull from google drive you can take a look to get some ideas, it is not as streamlined as you could make it. But it might give you some ideas. 

https://github.com/JoshuaSBrown/QC_Tools/blob/master/scripts/test_suite_install.bash

I used gdown, to download from google drive. You can place your files in a folder on google drive and make them shared with the public (read access).

JoshuaSBrown:
If you do go this direction be aware that getting the right https link to use with gdown not simply the link you see in the browser window when you are in google drive. It is some special link that has the file id associated with it.

Yurlungur:
> These are fair points. I will say this though, there is a limit on how much space you can use with gitlfs which is not that large, before you have to pay for it.
> 
> > If you use more than 1 GB of bandwidth per month without purchasing a data pack, Git LFS support is disabled on your account until the next month.
> 
> If you use something like google drive by default you start with 15gb, and I'm not aware of a limit on the bandwidth. This can be problematic if we are pulling from git lfs in the ci... you can quickly eat up the bandwidth.

What is the status of LANL's github subscription? Perhaps we get this already as part of our plan?

JoshuaSBrown:
> What is the status of LANL's github subscription? Perhaps we get this already as part of our plan?

I don't see it listed anywhere but I might not be looking in the right spots. However, that might also pose a problem for external collaborators.

Yurlungur:
> > What is the status of LANL's github subscription? Perhaps we get this already as part of our plan?
> 
> I don't see it listed anywhere but I might not be looking in the right spots. However, that might also pose a problem for external collaborators.

Can external collaborators not use the hosting paid for by LANL on github?

AndrewGaspar:
@junghans Do we already have a data plan for the lanl repo? If not, what's the prospect for getting one? It looks like $5/month gets you 50 GB of storage and 50 GB of monthly bandwidth, which should be plenty for our purposes. See: https://docs.github.com/en/github/setting-up-and-managing-billing-and-payments-on-github/upgrading-git-large-file-storage

junghans:
@AndrewGaspar we don't have a data plan, you will need to do a purchase exception to get it.

junghans:
Btw. mantevo ran out of bandwidth, which had all kind of strange implications: https://github.com/Mantevo/mantevo.github.io/issues/2

pgrete:
While I like the idea of having an "integrated" storage solution for the binary reference files, the bandwidth limitation on GitHub is a showstopper from my point of view.
As far as I can tell GitLab has 10GB of space and unlimited bandwidth. Could it be an option to put the file on the CI mirror?
Overall, I like the idea of having the data at a place where everyone in the collab can update them when new tests are required.
In another collaboration we're using the "Release" feature (https://docs.github.com/en/github/managing-large-files/distributing-large-binaries) to provide the reference files, see https://github.com/enzo-project/enzo-dev/releases

JoshuaSBrown:
> In another collaboration we're using the "Release" feature (https://docs.github.com/en/github/managing-large-files/distributing-large-binaries) to provide the reference files, see https://github.com/enzo-project/enzo-dev/releases

Clarify: Are you suggesting having a separate repository just for this? I don't understand how this would work otherwise. If you need the files for testing in the ci how else can you get them?

pgrete:
> > In another collaboration we're using the "Release" feature (https://docs.github.com/en/github/managing-large-files/distributing-large-binaries) to provide the reference files, see https://github.com/enzo-project/enzo-dev/releases
> 
> Clarify: Are you suggesting having a separate repository just for this? I don't understand how this would work otherwise. If you need the files for testing in the ci how else can you get them?

The design would be very close to the current one in `master` where we'd just `wget` the reference files.
The difference would be to the download would come from this repo rather than somewhere else (my webspace, google drive, ...).
Within this repo the file(s) would be provided via a binary release file (so it's not part of the version controlled source), but still people within the collaboration could add new reference files when necessary as it's hosted here.

JoshuaSBrown:
> > > In another collaboration we're using the "Release" feature (https://docs.github.com/en/github/managing-large-files/distributing-large-binaries) to provide the reference files, see https://github.com/enzo-project/enzo-dev/releases
> > 
> > 
> > Clarify: Are you suggesting having a separate repository just for this? I don't understand how this would work otherwise. If you need the files for testing in the ci how else can you get them?
> 
> The design would be very close to the current one in `master` where we'd just `wget` the reference files.
> The difference would be to the download would come from this repo rather than somewhere else (my webspace, google drive, ...).
> Within this repo the file(s) would be provided via a binary release file (so it's not part of the version controlled source), but still people within the collaboration could add new reference files when necessary as it's hosted here.

Does this mean a new release would need to be added each time you wanted to update the files?

pgrete:
> > > > In another collaboration we're using the "Release" feature (https://docs.github.com/en/github/managing-large-files/distributing-large-binaries) to provide the reference files, see https://github.com/enzo-project/enzo-dev/releases
> > > 
> > > 
> > > Clarify: Are you suggesting having a separate repository just for this? I don't understand how this would work otherwise. If you need the files for testing in the ci how else can you get them?
> > 
> > 
> > The design would be very close to the current one in `master` where we'd just `wget` the reference files.
> > The difference would be to the download would come from this repo rather than somewhere else (my webspace, google drive, ...).
> > Within this repo the file(s) would be provided via a binary release file (so it's not part of the version controlled source), but still people within the collaboration could add new reference files when necessary as it's hosted here.
> 
> Does this mean a new release would need to be added each time you wanted to update the reference files?

Yes, same as updating the git-lfs references or the current gold standard variable.

pgrete:
Are there any objections in trying out the "release" based approach as discussed during last week's call?
If no objections are voiced by Wednesday I'll update this PR accordingly as Andrew is on leave.

JoshuaSBrown:
@pgrete I'm pretty sure you have the support of all parties to move forward, given our discussion last week.

pgrete:
Decided to not update this branch but start with a clean one (the changes were effectively orthogonal), see #236

Yurlungur:
Should we close this branch then?

Issue #209, opened 2020-06-29
Labels for different tests. 


JoshuaSBrown:
It would probably be a good idea to clarify labels for the different tests that can be run and when they should apply. I would like to propose and open to discussion what cmake/Catch2 labels should be used with a given test. 

[unit] - when testing the correctness of a standalone unit of code, should not take more than 5s to run for any single test.
[performance/benchmarking] - for testing the performance of any part of the code base. 
[integration] - Testing the correctness of two or more parts of the code. 
[regression] - It seems that we are currently using this label when testing the correctness/performance of the code base as a whole.
[mpi-yes] - Running with mpi
[mpi-no] - Running without mpi

JoshuaSBrown:
[This is partly related](https://github.com/lanl/parthenon/issues/158)

Yurlungur:
I prefer `performance` over `benchmarking`. But I agree. This is a good idea.

Yurlungur:
Resolved

PR #224, opened 2020-07-15
Add performance regression test

pgrete:
## PR Summary

The second CI environment at IAS is finally working and this PR adds the basis of performance regression tests automatically creating plots like https://gitlab.com/theias/hpc/jmstone/athena-parthenon/parthenon-ci-mirror/-/jobs/639901312/artifacts/file/build-cuda/tst/regression/outputs/performance/performance.png
image

I expect this to expand in the future (in other PRs such as #190 ) so this is just mean to add the infrastructure to the codebase.
For now I left the performance tests as part of the "regression" test as they reuse the python based testing infrastructure.
We could update/refactor is based on #187 as well as while addressing #209 (but I see it as a separate effort from this).


tst/regression/utils/test_case.py
```diff
@@ -207,20 +208,21 @@ def Run(self):
         print(run_command)
         sys.stdout.flush()
         try:
-            subprocess.check_call(run_command)
+            proc = subprocess.run(run_command, check=True, stdout=PIPE, stderr=PIPE)
+            return proc.stdout
```
JoshuaSBrown, line 212: Instead of returning, can you simply add a stdouts attribute to the parameters class at the top of this file. You can then simply append to self.parameters.stdouts, and then there is no reason to have to pass it in to analse function.

tst/regression/utils/test_case.py
```diff
@@ -207,20 +208,21 @@ def Run(self):
...
         except subprocess.CalledProcessError as err:
             raise TestManagerError('\nReturn code {0} from command \'{1}\''
                               .format(err.returncode, ' '.join(err.cmd)))
 
-    def Analyse(self):
+    def Analyse(self, stdouts):
```
JoshuaSBrown, line 217: Follow instructions above and then remove stdouts argument.

tst/regression/utils/test_case.py
```diff
@@ -207,20 +208,21 @@ def Run(self):
...
 
         test_pass = False
 
         print("*****************************************************************")
         print("Analysing Driver Output")
         print("*****************************************************************")
         sys.stdout.flush()
-        test_pass = self.test_case.Analyse(self.parameters)
+        test_pass = self.test_case.Analyse(self.parameters, stdouts)
```
JoshuaSBrown, line 225: Follow instructions above and remove stdouts argument, it should be available in self.parameters.

tst/regression/CMakeLists.txt
```diff
@@ -110,3 +117,9 @@ foreach( index RANGE ${end_index})
   setup_test(${TEST_DIR} ${TEST_ARG})
   setup_test_mpi(${TEST_PROC} ${TEST_DIR} ${TEST_ARG})
 endforeach()
+
+# manually add label to performance regression test (to selectively enable/disable)
+set_tests_properties(regression_test:performance PROPERTIES LABELS "regression;mpi-no;perf-reg" )
+if( MPI_FOUND )
+  set_tests_properties(regression_mpi_test:performance PROPERTIES LABELS "regression;mpi-yes;perf-reg" )
+endif()
```
JoshuaSBrown, line 125: This is fine as it is, but it points to an underlying problem with how I implemented the cmake script in this case. The above foreach loop and functions would ideally be flexible enough to allow for extra labels and other customization without having to introduce if branches like this, which makes the script less readable. I will create an issue to come up with a better design.

tst/regression/utils/test_case.py
pgrete: Great point! That make this much cleaner.

pgrete:
Thanks for the review. I addressed the changes (parsing of `stdout`, directory naming, added doc) and updated to `master`.
Let me know what you think.

pgrete:
ping @jdolence (or someone else from the LANL physics team) 
Any input on this PR?

Yurlungur:
I'll take a look this afternoon.

tst/regression/CMakeLists.txt
```diff
@@ -53,6 +53,13 @@ list(APPEND TEST_ARGS "--driver ${CMAKE_BINARY_DIR}/example/advection/advection-
 --driver_input ${CMAKE_CURRENT_SOURCE_DIR}/test_suites/advection_convergence/parthinput.advection \
 --num_steps 25")
 
+# Performance regression test
+list(APPEND TEST_DIRS advection_performance)
+list(APPEND TEST_PROCS "1")
+list(APPEND TEST_ARGS "--driver ${CMAKE_BINARY_DIR}/example/advection/advection-example \
+--driver_input ${CMAKE_CURRENT_SOURCE_DIR}/test_suites/advection_performance/parthinput.advection_performance \
+--num_steps 5")
```
Yurlungur, line 61: This does not seem consistent with the documentation. It looks like we have to manually add tests to `cmake`?

pgrete: Yes. Good point. I'll update the doc.

pgrete:
Updated the docs, test pass, got two approval, pulling the trigger.
 @JoshuaSBrown  feel free to put the docker file to the other repo, too, pending the result of the discussion in #200

Issue #231, opened 2020-07-21
Allow non-static member functions to be used as tasks


jdolence:
@jmstone pointed out that developers may well want to use such functions as tasks, but the current `AddTask` functionality does not support it.

AndrewGaspar:
Looks like this should be closed now.

PR #232, opened 2020-07-21
Allow non-static class member functions to be tasks

jdolence:
## PR Summary

This PR adds an overload of the AddTask function that enables app developers to add non-static class member functions to task lists.  The interface is necessarily slightly different because non-static member functions need to be invoked by an object of the class, which in turn needs to be captured by the lambda we use to represent tasks.

Addresses #231 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/docs/tasks.md
+++ b/docs/tasks.md
@@ -8,6 +8,8 @@ The `TaskList` class implements methods to build and execute a set of tasks with
 
 When adding functions that are non-static class member functions, a slightly different interface is required.  The first argument should be the class-name-scoped name of the function.  For example, for a function named `DoSomething` in class `SomeClass`, the first argument would be `&SomeClass::DoSomething`.  The second argument should be a pointer to the object that should invoke this member function.  Finally, the dependencies and function arguments should be provided as described above.
 
+Examples of both `AddTask` calls can be found in the advection example [here](../examples/advection/advection_driver.cpp).
+
 ### DoAvailable
 `DoAvailable` loops over the task list once, executing all tasks whose dependencies are satisfied.  The function returns either `TaskListStatus::complete` if all tasks have been executed (and the task list is therefore empty) or `TaskListStatus::running` if tasks remain to be completed.
 
--- a/docs/tasks.md
+++ b/docs/tasks.md
@@ -6,6 +6,8 @@ The `TaskList` class implements methods to build and execute a set of tasks with
 ### AddTask
 `AddTask` is a templated variadic function that takes the task function to be executed, the task dependencies (see `TaskID` below), and the arguments to the task function as it's arguments.  All arguments are captured by value in a lambda for later execution.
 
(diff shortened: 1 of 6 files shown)


jdolence:
@jmstone, here's one possible solution to the member function as task issue.  The interface to AddTask is a little different, but necessary I think.  You'd use it like this:
```c++
auto taskid = tl.AddTask(&ClassName::member_func, pointer_to_object, dependencies, arg1, arg2, ...);
```
so if you wanted to add a task that calls `CalculateSomething` which is a non-static member function of the MeshBlock class, the line might look something like
```c++
auto taskid = tl.AddTask(&MeshBlock::CalculateSomething, pmb, deps, arg);
```
for example.

docs/tasks.md
```diff
@@ -6,6 +6,8 @@ The `TaskList` class implements methods to build and execute a set of tasks with
 ### AddTask
 `AddTask` is a templated variadic function that takes the task function to be executed, the task dependencies (see `TaskID` below), and the arguments to the task function as it's arguments.  All arguments are captured by value in a lambda for later execution.
 
+When adding functions that are non-static class member functions, a slightly different interface is required.  The first argument should be the class-name-scoped name of the function.  For example, for a function named `DoSomething` in class `SomeClass`, the first argument would be `&SomeClass::DoSomething`.  The second argument should be a pointer to the object that should invoke this member function.  Finally, the dependencies and function arguments should be provided as described above.
+
```
Yurlungur, line 10: Some example calls might be useful.

jdolence: advection example now uses this with refactored Container tasks.  doc points to the example.  good?

jdolence:
@JoshuaSBrown I'd appreciate a quick look at this when you get a chance.  I think it's ready to go.

docs/tasks.md
Yurlungur: Yep. Thanks for adding that!

jdolence:
@JoshuaSBrown any chance you can clarify why the codecov/project test failed?  I don't understand what it's telling me and therefore how to fix it.  Thanks!

JoshuaSBrown:
> @JoshuaSBrown any chance you can clarify why the codecov/project test failed? I don't understand what it's telling me and therefore how to fix it. Thanks!

I think it is because Codecov is a fickle child. This should be addressed with https://github.com/lanl/parthenon/pull/235.

src/tasks/task_list.hpp
```diff
@@ -93,6 +93,17 @@ class TaskList {
     return id;
   }
 
+  // overload to add member functions of class T to task list
+  // NOTE: we must capture the object pointer
+  template <class F, class T, class... Args>
+  TaskID AddTask(F func, T *obj, TaskID &dep, Args &&... args) {
```
JoshuaSBrown, line 99: This is a bit nitpicking on my part, but I would recommend moving the argument T *obj over, so that the function prototype is more consistent with the previous AddTask call.

jdolence: this signature was motivated by consistency with things like `std::async` calls with member functions.

jdolence:
given @JoshuaSBrown's comment on the codecov/project test failure, I'm pulling the trigger.

PR #234, opened 2020-07-22
Correct CMAKE variable, make local to project

JoshuaSBrown:
## PR Summary

Fixes an incorrect cmake variable that uses the wrong path when parthenon is included as a submodule. https://github.com/lanl/parthenon/issues/233

## PR Checklist

- [x] Code passes cpplint

### DIFF
--- a/cmake/CodeCov.cmake
+++ b/cmake/CodeCov.cmake
@@ -40,15 +40,15 @@ if(CODE_COVERAGE)
     COMMAND mkdir -p ${COVERAGE_PATH}/${COVERAGE_NAME}
     COMMAND ${PATH_LCOV} --version
     # Clean
-    COMMAND ${PATH_LCOV} --gcov-tool ${PATH_GCOV} --directory ${CMAKE_BINARY_DIR} -b ${CMAKE_SOURCE_DIR} --zerocounters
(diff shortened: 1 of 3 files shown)


pgrete:
For consistency should we also update the paths for the coverage?

JoshuaSBrown:
> For consistency should we also update the paths for the coverage?

What do you mean update the paths? I don't quite follow. 

I just submitted a pull request so the coverage test is not so sensitive to small changes like this, does that work? It think coverage is important as a tool but for stuff like this... I think some leeway is preferable.

pgrete:
> > For consistency should we also update the paths for the coverage?
> 
> What do you mean update the paths? I don't quite follow.

In retrospect I could/should have been more specific. What I meant was the usage of the `CMAKE_BINARY_DIR` variable also in https://github.com/lanl/parthenon/blob/a4b09fe7ef8c134572135b581d12349b9e8d134d/cmake/CodeCov.cmake#L33 (and other places in that file).

JoshuaSBrown:
> n retrospect I could/should have been more specific. What I meant was the usage of the `CMAKE_BINARY_DIR` variable also in

Good point.

PR #235, opened 2020-07-22
Adjust coverage settings so coverage fail is not so sensitive

JoshuaSBrown:
## PR Summary

Adjust coverage threshold, so test does not fail unless a drop in coverage exceeds 1 %. This will help prevent false positives. For instance as seen here: https://github.com/lanl/parthenon/pull/234

## PR Checklist

- [x] Code passes cpplint


JoshuaSBrown:
> Looks like this PR now adds a `codecov.yaml` alongside the existing `codecov.yml`.
> I assume that the original one is a leftover and can go.
> Otherwise looks good!

Oh, good catch.

JoshuaSBrown:
ping @pgrete, I have fixed the file issues.

PR #236, opened 2020-07-23
Move gold standard to GitHub and add CMake target

pgrete:
## PR Summary

Following up on the discussion in #216 this move the gold standard binary files to GitHub (by adding a release, see https://github.com/lanl/parthenon/releases) and add a cmake target to download them.

## PR Checklist


pgrete:
ping @junghans @AndrewGaspar @gshipman who were involved in the discussion before

CMakeLists.txt
```diff
@@ -192,6 +196,30 @@ endif()
...
+      ExternalProject_Add(
+        gold_standard
+        URL https://github.com/lanl/parthenon/releases/download/regression-gold-v${REGRESSION_GOLD_STANDARD}/parthenon_regression_gold_v${REGRESSION_GOLD_STANDARD}.tgz
+        URL_HASH SHA512=45f57d16b76a3a44940e40ea642c1d5a2b3aea681a3064e4c06b031bf264d627f8009e967b92a2cb0fbc08e5cd0ffda381038483f390fa00f47c42a551ca3646
+        SOURCE_DIR "${PROJECT_SOURCE_DIR}/tst/regression/gold_standard"
+        DOWNLOAD_DIR "${PROJECT_SOURCE_DIR}/tst/regression/gold_standard"
+        # just need the extracted data so no configure/build/install
+        CONFIGURE_COMMAND ""
+        BUILD_COMMAND ""
+        INSTALL_COMMAND ""
```
JoshuaSBrown, line 216: This is neat, I didn't know about the ExternalProject cmake module.

CMakeLists.txt
```diff
@@ -53,6 +54,9 @@ include(cmake/Lint.cmake)
 # internal variable for number of ghost cells
 set(NUMBER_GHOST_CELLS ${PARTHENON_NGHOST}) 
 
+set(REGRESSION_GOLD_STANDARD 2 CACHE STRING "Version of gold standard to download and use")
```
Yurlungur, line 57: Should this also be an option so that we can set it at configure time? It might be useful when debugging to backtest these things with older versions of the gold standard.

CMakeLists.txt
```diff
@@ -192,6 +196,30 @@ endif()
...
+    if (EXISTS ${PROJECT_SOURCE_DIR}/tst/regression/gold_standard/current_version)
+      file(READ ${PROJECT_SOURCE_DIR}/tst/regression/gold_standard/current_version REGRESSION_GOLD_STANDARD_ON_DISK)
+    else()
+      set(REGRESSION_GOLD_STANDARD_ON_DISK 0)
+    endif()
+    if (NOT (${REGRESSION_GOLD_STANDARD_ON_DISK} EQUAL ${REGRESSION_GOLD_STANDARD}))
+      ExternalProject_Add(
+        gold_standard
+        URL https://github.com/lanl/parthenon/releases/download/regression-gold-v${REGRESSION_GOLD_STANDARD}/parthenon_regression_gold_v${REGRESSION_GOLD_STANDARD}.tgz
+        URL_HASH SHA512=45f57d16b76a3a44940e40ea642c1d5a2b3aea681a3064e4c06b031bf264d627f8009e967b92a2cb0fbc08e5cd0ffda381038483f390fa00f47c42a551ca3646
```
Yurlungur, line 210: Can we make these variable set further up in the script? Or even better, can we auto-compute the hash? When the version changes, presumably the hash will too... I can see this being annoying to maintain.

docs/README.md
```diff
@@ -26,16 +26,21 @@ The tests currently cover
...
+To execute the tests run, e.g.,
 ```bash
 # from within the build directory (add -V fore more detailed output)
 ctest -R regression
 ```
+The gold standard files (reference solutions) used in the regression tests should automatically be downloaded during the `make` phase.
+Alternatively, you can download them by directly make the appropriate target
+```bash
+make gold_standard
+```
```
Yurlungur, line 38: this is obviously a huge improvement over the `wget` command. 👍

docs/README.md
```diff
@@ -26,16 +26,21 @@ The tests currently cover
...
+or download them as a release from [GitHub](https://github.com/lanl/parthenon/releases/).
+Make sure to get the correct version matching your source
+(stored in the `REGRESSION_GOLD_STANDARD` CMake variable).
+Note: If you results are (slightly) different, that may stem from using different
+compiler/optimization options.
```
Yurlungur, line 43: I'm not sure it's worth expounding on this, but there's more to expound on. We shouldn't expect bitwise compatibility even if the compiler/optimization is identical if we're running on different machines. There *will* be slight differences, and this has to be something we expect and account for.

CMakeLists.txt
junghans: You make a good point, if the version is a cached variable, the hash should be as well and maybe even the url, so that one can test against a local file (url support paths) before uploading.

docs/README.md
junghans: Wouldn't `cmake --build . --target gold_standard` be the more cmakish thing to do?

CMakeLists.txt
junghans: In some way, if you don't run configure/build/install, `FetchContent_Declare()` or even `file(DOWNLOAD..)` might be the more natural thing to do.

CMakeLists.txt
pgrete: @Yurlungur this already is an option, it's just not using the "option" keyword as it's not a bool variable.
@junghans this is a good point (hash and url). However, from my point of view, this may be more convoluted than just using "local" files (as checked by the "current_version" file content), i.e., instead of letting the user set the version variable, the new url, and the hash, the user just puts arbitrary files in the gold_standard folder and sets a matching string in the "current_version" file.
This way we'd keep the logic to do testing/getting the gold files in a clean from a downstream user point of view.
I updated the docs below to reflect this. Please let me know what you think.

CMakeLists.txt
pgrete: Good point. I'm now directly using the `file` features.
At some point it may be worth thinking about using this for Kokkos and catch2 (instead of cloning the entire repos as submodules).

CMakeLists.txt
pgrete: The hash is computed manually for the file downloaded and then compared against this reference value. This way it's ensured that the file downloaded is not corrupted or altered.
Given that the hash and the version change at the same time (and likely very infrequently) I don't see this as too annoying to maintain. However, if this is an issue we could remove potentially remove the hash.
Anyway, now moved the HASH variable further up next to the version as those two would be updated simultaneously.

docs/README.md
pgrete: Given that I now use `file` there's no target anymore and the file are always downloaded (if out of sync). How do people feel about this?

docs/README.md
pgrete: We implicitly already account for this. The test data has been generated in double prec but stored single prec in the hdf5 files.
So far all different machine and compilers I tested successfully passed this test.
Did anyone come across a machine/configuration where our tests currently fail for those reasons?

pgrete:
I made couple of changes that slightly change the logic (no target anymore) and expanded the documentation on how to handle/update gold standard files.
Let me know what you think.

CMakeLists.txt
Yurlungur: I think it's very likely that in debugging, the file will be updated frequently. Since a user may want to run tests against an older version or against their own file. So my preference would be to remove the `HASH`. But it's not a big deal.

docs/README.md
Yurlungur: No. I wasn't expecting this to cause a test failure. I just meant the `README` might be misleading if you don't know about architecture-level differences.

CMakeLists.txt
Yurlungur: oh I see. The hash only matters for downloading the file, not locally. Never mind. This is fine with me.

CMakeLists.txt
Yurlungur: The improved docs and this strategy work for me.

docs/README.md
Yurlungur: This seems fine. It is a little weird that there's no way to explicitly ask the build system to "update my files" but since it happens automatically every time that seems fine.

PR #243, opened 2020-07-29
Added cmake module which checks for required python modules, updated …

JoshuaSBrown:
## PR Summary

Fixes python version issue as well as detecting potential problems with python modules. Solves problems raised here https://github.com/lanl/parthenon/issues/242.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


cmake/PythonModuleCheck.cmake
```diff
@@ -0,0 +1,28 @@
...
+# Function will check that the specified modules are available to the python interpreter
+# If they are not cmake will throw an error indicating which module not available
+function(required_python_modules_found module_list)
+  if(Python_Interpreter_FOUND)
+    foreach(module IN LISTS module_list )
+      execute_process(COMMAND ${Python3_EXECUTABLE} -c "import ${module}"
+        RESULT_VARIABLE IMPORT_MODULE ERROR_QUIET)
+    
+      if(${IMPORT_MODULE} EQUAL 1)
+        message(FATAL_ERROR "Required python module ${module} was not found.")
```
AndrewGaspar: Could you instead aggregate a list of all the missing modules and print them out as a single error message? That way you find out what you're missing all at once.

tst/regression/CMakeLists.txt
```diff
@@ -61,22 +58,27 @@ list(APPEND TEST_ARGS "--driver ${PROJECT_BINARY_DIR}/example/advection/advectio
...
+  list(APPEND TEST_ARGS "--driver ${PROJECT_BINARY_DIR}/example/advection/advection-example \
+    --driver_input ${CMAKE_CURRENT_SOURCE_DIR}/test_suites/output_hdf5/parthinput.advection \
+    --num_steps 4")
+  # h5py is only neeed if hdf5 is enabled
+  list(APPEND REQUIRED_PYTHON_MODULES h5py) 
 endif()
 
+# Any modules that are required by python can be added to REQUIRED_PYTHON_MODULES list variable,
+# before including TestSetup.cmake. 
+list(APPEND REQUIRED_PYTHON_MODULES argparse errno math matplotlib numpy os shutil)
```
AndrewGaspar: I suppose it doesn't hurt anything, but I don't think it's necessary to list builtin modules here - we're requiring 3.6 already, and so those builtin modules should be there.

tst/regression/CMakeLists.txt
```diff
@@ -61,22 +58,27 @@ list(APPEND TEST_ARGS "--driver ${PROJECT_BINARY_DIR}/example/advection/advectio
...
+    --driver_input ${CMAKE_CURRENT_SOURCE_DIR}/test_suites/output_hdf5/parthinput.advection \
+    --num_steps 4")
+  # h5py is only neeed if hdf5 is enabled
+  list(APPEND REQUIRED_PYTHON_MODULES h5py) 
 endif()
 
+# Any modules that are required by python can be added to REQUIRED_PYTHON_MODULES list variable,
+# before including TestSetup.cmake. 
+list(APPEND REQUIRED_PYTHON_MODULES argparse errno math matplotlib numpy os shutil) 
+list(APPEND REQUIRED_PYTHON_MODULES subprocess sys)
```
AndrewGaspar: You can just split the arguments to a function call in CMake across multiple lines.

cmake/TestSetup.cmake
```diff
@@ -11,6 +11,14 @@
 # the public, perform publicly and display publicly, and to permit others to do so.
 #=========================================================================================
 
+# Seaches for the python interpreter
+message(STATUS "Searching for python 3.6, required to run regression tests.")
+include(FindPython)
```
AndrewGaspar: I don't think this is necessary. I'm actually not even sure what this would do. `find_package(Foo)` looks for a module called `FindFoo` and invokes that, so the `find_package(Python3)` on the next line is sufficient.

AndrewGaspar:
Oh, sorry, I suppose I shouldn't have "reviewed" this yet.

cmake/PythonModuleCheck.cmake
JoshuaSBrown: Sure.

cmake/TestSetup.cmake
JoshuaSBrown: They don't do the same thing. include(FindPython) exposes several variables that are otherwise hidden. It is not an old feature and is currently why the ci is failing because the module is only available after cmake version 12 something.

nmsriram: @JoshuaSBrown Is this ready to merge into trunk?

tst/regression/CMakeLists.txt
JoshuaSBrown: Ok, that would just leave matplotlib and numpy.

cmake/TestSetup.cmake
JoshuaSBrown: No, I'm working on it right now, your recent merge, does not appear to correctly use the PythonModuleCheck.cmake, because you removed the inlcude(FindPython), I'm adding a conditional statement to deal with this if cmake is less than 3.12.

AndrewGaspar: Seeing @nmsriram's PR, I think it's because `find_package(Python3 ...)` doesn't set the the `Python_` variables, only the `Python3_` variables, but I think that all the result variables you need are there: https://cmake.org/cmake/help/latest/module/FindPython3.html?highlight=findpython3#result-variables

cmake/TestSetup.cmake
```diff
@@ -11,6 +11,43 @@
...
 #=========================================================================================
 
+
+
+# Search for the python interpreter
+message(STATUS "Searching for python 3.6 or higher, required to run regression tests.")
+find_package(Python3 3.6 QUIET COMPONENTS Interpreter)
+
+# search executable path
+if ( NOT Python3_FOUND)
```
AndrewGaspar: Which system required the following code to manually implement the find for Python?

JoshuaSBrown: Give me a few minutes.

cmake/TestSetup.cmake
JoshuaSBrown: Oh, is that what's going on, @AndrewGaspar. Ok, then some of the variables in the python module in the /cmake/ folder need to be updated.

cmake/TestSetup.cmake
JoshuaSBrown: ping @nmsriram you added this section.

cmake/TestSetup.cmake
JoshuaSBrown: This variable does not appear to be defined using just find_package(Python3 ...) blah blah blah. At least not for cmake 3.11. The problem appears to be the cmake versions. find_package(Python3 ... ) only defines the needed Python3 variables after cmake 3.13, whereas for the Python variables they are defined for cmake 3.12. We either need to update the cmake version or specifically define these variables, or a workaround.

AndrewGaspar: Have we considered simply throwing ourselves into the sea?

cmake/TestSetup.cmake
AndrewGaspar: We discussed on the phone - sounds like Cmake 3.11 and earlier did not have `FindPython`. However, it did have `FindPythonInterp`, so we should use that instead for older versions.

nmsriram: @agaspar The CI fails without this section because the default image used for CPU and GPU tests doesn't come with a FindPython.cmake file.  It might be fixed by going to a newer CMake, but for the current CI image, this is a necessity.

nmsriram: @AndrewGaspar @JoshuaSBrown CI is now failing with:
````
61CMake Error at CMakeLists.txt:14 (cmake_minimum_required):
62  CMake 3.12 or higher is required.  You are running version 3.10.3
````

Comment (AndrewGaspar) [reply cmake/TestSetup.cmake]: Yeah, I'm proposing something like:
```cmake
if (CMAKE_VERSION VERSION_LESS "3.12")
    find_package(PythonInterp 3.6 REQUIRED)
    set(Python_Interpreter_FOUND ON)
    set(Python_EXECUTABLE ${PYTHON_EXECUTABLE})
else()
    find_package(Python 3.6 REQUIRED)
endif()
```

JoshuaSBrown: I'm working on this will update within the next 20 minutes.

JoshuaSBrown: Ok, that took far longer than I thought. This works as expected.

JoshuaSBrown:
Here is a question, would it be idea to also be able to choose the specific version of Python to use in the case that you have multiple versions installed, or specify the path to python version you want to use explicitly?

The other option is to use all the python versions above the minimum required and use the first one that has access to all the required submodules. There is no way currently with the build in cmake find_packages function to do this, I would need to add some code that does this. 

The third option is to implement both options.

AndrewGaspar:
> Here is a question, would it be idea to also be able to choose the specific version of Python to use in the case that you have multiple versions installed, or specify the path to python version you want to use explicitly?

This is supported out of the box in CMake - set the `Python_ROOT` variable. See: https://cmake.org/cmake/help/latest/command/find_package.html#search-procedure

AndrewGaspar:
Oh, but I suppose if you have all your python versions installed in the same directory (e.g. `python36`, `python37`, `python38`, etc.), that wouldn't work...

I'd say don't worry about it now and somebody can add that later if they need it.

AndrewGaspar:
Actually, you can explicitly specify `Python_EXECUTABLE`, which should be sufficient: https://cmake.org/cmake/help/latest/module/FindPython.html#artifacts-specification

cmake/TestSetup.cmake
```diff
@@ -11,6 +11,27 @@
 # the public, perform publicly and display publicly, and to permit others to do so.
 #=========================================================================================
 
+
+
+# Search for the python interpreter
+if ( CMAKE_VERSION VERSION_LESS "3.12.4")
```
AndrewGaspar: Make this 3.12 or 3.12.0 - you can only browse the documentation for the latest patch release of each minor version, but I don't think documentation-worthy changes are made in patch releases. So `FindPython` should be in 3.12.0, too.

JoshuaSBrown: Well other than this, is it good to go?

tst/regression/test_suites/output_hdf5/output_hdf5.py
```diff
@@ -81,7 +81,17 @@ def Analyse(self,parameters):
         analyze_status = True
         print(os.getcwd())
 
-        sys.path.insert(1, '../../../../../scripts/python')
+        # Determine path to parthenon installation
+        # Fallback to relative path on failure
+        try:
+            parthenonPath = os.path.realpath(__file__)
+            idx = parthenonPath.rindex('/parthenon/')
+            parthenonPath = parthenonPath[:idx]+'/parthenon'
```
Yurlungur: I'm not sure... does adding a path object to a string result in a path object? Or does it result in a string? Just to be safe, I'd do
```suggestion
            parthenonPath = os.path.join(parthenonPath[:idx],'parthenon')
```

cmake/TestSetup.cmake
```diff
@@ -11,6 +11,27 @@
...
 
+
+
+# Search for the python interpreter
+if ( CMAKE_VERSION VERSION_LESS "3.12")
+  # Version number has been intentionally excluded from find_package call, so that latest version 
+  # will be grabbed. Including the version number would prioritise the version provided over more 
+  # up to date versions
+  find_package(PythonInterp REQUIRED)
+  if( ${PYTHON_VERSION_STRING} VERSION_LESS "3.6")
```
Yurlungur: Is 3.6 the version we want to pin on? I don't object---Python development moves slowly and minor versions don't change very often. Just wanted to make sure there's no users or environments we're missing out on by picking this.

JoshuaSBrown: I think it is a good choice. Has most modern features.

Yurlungur: Yeah but are we actually using any of those modern features? We should certainly pin ourselves on version >3, since 2 is deprecated. But I don't want to pick a python version that's missing on, say, an HPC cluster we need. 

That said python 3.6 was first released in 2016, so we're probably fine. I just wanted to make sure. @pgrete, @forrestglines , @jmstone : are there any HPC systems you use that are stuck on an older version of Python? 3.2 might be the safest choice, as that was released in 2011.

JoshuaSBrown: @pgrete @forrestglines @jmstone

pgrete: I think Python 3.6 is a good choice. It's been some time since I came across an old python 3.x version (all the systems I checked have python >= 3.6 available either by default or through modules)

pgrete:
I really like this. Definitely an improvement!
In terms of functionality, did you verify that the module check works?
I tried and it doesn't fail for me if I add a bogus module name to the list, e.g. 
```diff
 # Any external modules that are required by python can be added to REQUIRED_PYTHON_MODULES 
 # list variable, before including TestSetup.cmake. 
-list(APPEND REQUIRED_PYTHON_MODULES matplotlib numpy) 
+list(APPEND REQUIRED_PYTHON_MODULES matplotlib numpy blaa) 
 
 # Include test setup functions, and check for python interpreter and modules
```
works without problem (cmake version 3.16.1 and python 3.7)

cmake/TestSetup.cmake
Yurlungur: Fine by me then.

JoshuaSBrown:
> I really like this. Definitely an improvement!
> In terms of functionality, did you verify that the module check works?
> I tried and it doesn't fail for me if I add a bogus module name to the list, e.g.
> 
> ```diff
>  # Any external modules that are required by python can be added to REQUIRED_PYTHON_MODULES 
>  # list variable, before including TestSetup.cmake. 
> -list(APPEND REQUIRED_PYTHON_MODULES matplotlib numpy) 
> +list(APPEND REQUIRED_PYTHON_MODULES matplotlib numpy blaa) 
>  
>  # Include test setup functions, and check for python interpreter and modules
> ```
> 
> works without problem (cmake version 3.16.1 and python 3.7)

I did check, but let me see if I can reproduce the problems you encountered.

JoshuaSBrown:
> > I really like this. Definitely an improvement!
> > In terms of functionality, did you verify that the module check works?
> > I tried and it doesn't fail for me if I add a bogus module name to the list, e.g.
> > ```diff
> >  # Any external modules that are required by python can be added to REQUIRED_PYTHON_MODULES 
> >  # list variable, before including TestSetup.cmake. 
> > -list(APPEND REQUIRED_PYTHON_MODULES matplotlib numpy) 
> > +list(APPEND REQUIRED_PYTHON_MODULES matplotlib numpy blaa) 
> >  
> >  # Include test setup functions, and check for python interpreter and modules
> > ```
> > 
> > 
> > works without problem (cmake version 3.16.1 and python 3.7)
> 
> I did check, but let me see if I can reproduce the problems you encountered.

I found the problem I missed a file when committing.

pgrete:
Great. I can confirm that it fails now properly -> it works :smile: 

Given that decided to bump cmake to 3.12 in #200 do you want to remove the logic for version <3.12 already here?

JoshuaSBrown:
> Great. I can confirm that it fails now properly -> it works 😄
> 
> Given that decided to bump cmake to 3.12 in #200 do you want to remove the logic for version <3.12 already here?

Yes 👍, sounds like a plan.

AndrewGaspar:
Could you add a note to the README: https://github.com/lanl/parthenon/issues/265

AndrewGaspar:
Looks like the tests are failing, probably because of the newer CMake requirement. Is that being addressed?

PR #245, opened 2020-07-30
used absolute path for parthenon scripts instead of relative paths

nmsriram:
## PR Summary
Added absolute path for parthenon rather than relative one in output_hdf5.py

## PR Checklist

- [ ] N/A: Code passes cpplint
- [ ] N/A: New features are documented.
- [ ] N/A: Adds a test for any bugs fixed. Adds tests for new features.
- [x] Code is formatted


tst/regression/test_suites/output_hdf5/output_hdf5.py
```diff
@@ -81,7 +81,10 @@ def Analyse(self,parameters):
         analyze_status = True
         print(os.getcwd())
 
-        sys.path.insert(1, '../../../../../scripts/python')
+        parthenonPath = os.path.realpath(__file__)
+        idx = parthenonPath.find('/parthenon/')
```
JoshuaSBrown, line 85: Can you use rfind instead:

`
parthenonPath.rfind('/parthenon/')
`

I can see cases where there might be a folder inside of a folder.

JoshuaSBrown: find will return the first instance.

nmsriram: That's a good suggestion: that will also eliminate the '/parthenon/' from the rest of the strings.

.gitlab-ci.yml
```diff
@@ -2,6 +2,7 @@ image: cuda10.0-mpi-hdf5-mpl
 
 # Is performed before the scripts in the stages step
 before_script:
+    - apt-get install python3.7
```
JoshuaSBrown, line 5: Why is this needed?

nmsriram: CI was failing because default has python2.7.
Maybe a red herring.

nmsriram: @JoshuaSBrown I'm still debugging the CI.  Will update when done.

nmsriram: This has been removed.  CI issues were resolved without this additional install.

cmake/TestSetup.cmake
```diff
@@ -11,11 +11,40 @@
 # the public, perform publicly and display publicly, and to permit others to do so.
 #=========================================================================================
 
-# Seaches for the python interpreter
-message(STATUS "Searching for python 3.6, required to run regression tests.")
-include(FindPython)
```
JoshuaSBrown, line 16: Thanks @nmsriram for all your work, just a heads up though by removing this module `include(FindPython)` some other problems have been introduced. There are several variables that are defined in FindPython that are populated after calling find_package, by removing it you have to be careful because these variables are now no longer available. E.g. Python_Interpreter_FOUND is no longer defined even though Python3 3.6 is found.

AndrewGaspar: Shouldn't it be [`Python3_Interpreter_FOUND`](https://cmake.org/cmake/help/latest/module/FindPython3.html?highlight=findpython3#result-variables)?

JoshuaSBrown: Yes good catch.

PR #247, opened 2020-08-03
WIP: Add -gpu flag for Spectrum MPI and run serial tests under mpiexec

AndrewGaspar:
## PR Summary

This PR fixes some small issues when running Parthenon's tests with Spectrum MPI. Mainly, it applies the `-gpu` flag to `mpiexec`. It also runs the serial tests under `mpiexec`, as Spectrum MPI requires that.

## PR Checklist

- [x] Code passes cpplint (No C++ changes)
- [x] New features are documented. (No new features)
- [x] Adds a test for any bugs fixed. Adds tests for new features. (Fixes some tests)
- [x] Code is formatted (No C++ changes)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -111,6 +111,16 @@ if (NOT PARTHENON_DISABLE_MPI)
   set(ENABLE_MPI ON)
 endif()
 
+execute_process(
+  COMMAND ${MPIEXEC_EXECUTABLE} --version
+  OUTPUT_VARIABLE MPIEXEC_VERSION_RAW
+)
+
+if (MPIEXEC_VERSION_RAW MATCHES "mpiexec \\(IBM Spectrum MPI\\)")
+  message(STATUS "Using IBM Spectrum MPI")
+  set(MPI_IMPLEMENTATION "Spectrum")
+endif()
+
 set(ENABLE_OPENMP OFF)
 if (NOT PARTHENON_DISABLE_OPENMP)
   find_package(OpenMP COMPONENTS CXX)
@@ -192,6 +202,11 @@ elseif(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/external/Kokkos/CMakeLists.txt)
   add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/external/Kokkos Kokkos)
 else()
   find_package(Kokkos 3 REQUIRED)
+endif()
+
+if (Kokkos_ENABLE_CUDA AND MPI_IMPLEMENTATION STREQUAL "Spectrum")
+  # Cuda awareness must be explicitly enabled for Spectrum builds builds
+  set(MPIEXEC_PREFLAGS "${MPIEXEC_PREFLAGS} -gpu")
 endif()
 
 # Build Tests and download Catch2
--- a/cmake/TestSetup.cmake
+++ b/cmake/TestSetup.cmake
@@ -30,8 +30,16 @@ endfunction()
 # test output will be sent to /tst/regression/outputs/dir
 # test property labels: regression, mpi-no
 function(setup_test dir arg)
-  separate_arguments(arg) 
-  add_test( NAME regression_test:${dir} COMMAND python "${CMAKE_CURRENT_SOURCE_DIR}/run_test.py" 
+  separate_arguments(arg)
+  if (ENABLE_MPI AND MPI_IMPLEMENTATION STREQUAL "Spectrum")
+    # Spectrum MPI requires all MPI programs to be run under mpiexec
+    set(ADDITIONAL_ARGS
+      --mpirun ${MPIEXEC_EXECUTABLE}
+      --mpirun_opts=${MPIEXEC_NUMPROC_FLAG} --mpirun_opts=1
+      --mpirun_opts=${MPIEXEC_PREFLAGS})
+  endif()
+  add_test( NAME regression_test:${dir} COMMAND python "${CMAKE_CURRENT_SOURCE_DIR}/run_test.py"
+    ${ADDITIONAL_ARGS}
     ${arg} --test_dir "${CMAKE_CURRENT_SOURCE_DIR}/test_suites/${dir}"
     --output_dir "${PROJECT_BINARY_DIR}/tst/regression/outputs/${dir}")
   set_tests_properties(regression_test:${dir} PROPERTIES LABELS "regression;mpi-no" )


pgrete:
I see where the problem comes from but would like to discuss a different solution approach.
The background is that I'm wondering how much logic we want to introduce on our side and whether it's maintainable, e.g.,
Spectrum requires `-gpu` whereas MVAPICH requires `MV2_USE_CUDA=1` etc.
So I'm leaning towards using environment variables where possible (e.g., on TACC's Longhorn I can set `export MY_SPECTRUM_OPTIONS="--gpu"`) in combination with a build documentation for different systems in favor of adding (potentially increasingly complex) logical statements to Cmake trying to adapt to changing environments.

Regarding running problem compiled with MPI support using mpiexec it seems that this hints at a bug/inconsistency in our regression tests.
I think we should be more strict in terms of what's MPI and what's not and make sure the resulting binary is executed the correct way.

AndrewGaspar:
> So I'm leaning towards using environment variables where possible (e.g., on TACC's Longhorn I can set export MY_SPECTRUM_OPTIONS="--gpu") in combination with a build documentation for different systems in favor of adding (potentially increasingly complex) logical statements to Cmake trying to adapt to changing environments.

I don't agree with that. People will waste a lot of time (like I did) if you expect them to read a bunch of documentation before they can even run `ctest`, which is one of the first things any new developer will presumably do when checking out the repository. My opinion is that, at least on typically used machines, the following script should simply succeed on the systems people typically build on with minimal advanced configuration:
```
cmake -S. -Bbuild
cmake --build build
cd build && ctest
```

I think if you do anything else, it ends up creating a lot of duplicated work. If you'd like, I can compartmentalize this code a bit more?

AndrewGaspar:
> Regarding running problem compiled with MPI support using mpiexec it seems that this hints at a bug/inconsistency in our regression tests.

So just always run the serial tests under mpiexec if MPI is enabled? I can do that.

pgrete:
> > Regarding running problem compiled with MPI support using mpiexec it seems that this hints at a bug/inconsistency in our regression tests.
> 
> So just always run the serial tests under mpiexec if MPI is enabled? I can do that.

From my point of view this'd be more consistent. If we compile with MPI we run with MPI even if it's serial code (eventually the ifdefs in the code checking for MPI don't matter at runtime ).

pgrete:
> > So I'm leaning towards using environment variables where possible (e.g., on TACC's Longhorn I can set export MY_SPECTRUM_OPTIONS="--gpu") in combination with a build documentation for different systems in favor of adding (potentially increasingly complex) logical statements to Cmake trying to adapt to changing environments.
> 
> I don't agree with that. People will waste a lot of time (like I did) if you expect them to read a bunch of documentation before they can even run `ctest`, which is one of the first things any new developer will presumably do when checking out the repository. My opinion is that, at least on typically used machines, the following script should simply succeed on the systems people typically build on with minimal advanced configuration:

My experience with trying to make things work at the "automated" level results in hardly to maintain and causes additional problem when similar environments (here identified as Spectrum MPI) require slightly different configuration/parameters as the vendor fixed sth on a different system.
 
> ```
> cmake -S. -Bbuild
> cmake --build build
> cd build && ctest
> ```
> 
> I think if you do anything else, it ends up creating a lot of duplicated work. If you'd like, I can compartmentalize this code a bit more?

How about we add machine config files for all our favorite machines, or, in other words, move the parameters required to a separate entity and don't introduce logic that tries to determine the required parameters automatically.
With respect to a "minimal" workflow I think it'd be fine to ask new people to add a single additional parameter
```bash
$ cmake -S. -Bbuild -DMACHINE=Darwin-GPU
-- The C compiler identification is GNU 10.1.0
...
-- This is the Darwin GPU default configuration. It expects that
     - the modules ABC and XYZ are loaded
     - environment variables MNO are set
     - MPI runs are executed with jsrun -n N ./parthenon --kokkos-ndevices=4
...
-- Generating done
-- Build files have been written to: /my/build/dir
$ cmake --build build
$ cd build && ctest
```

JoshuaSBrown:
I would think using defaults based on what is available, that work on most of our machines would be the way to go. I don't understand the spectrum issue, if spectrum is used aren't you always going to want the "--gpu" option, if a gpu is available? 

It seems to me, to be more maintainable and generally applicable to determine the appropriate flags from what applications are detected by cmake than to explicitly define machine specific setups which are susceptible to becoming stale over time and will not scale well. I think also that if you want parthenon to be of use to the wider community having some level of automation is preferable. There are a number of open source projects I could never use because I could not get an appropriate build configuration setup.

pgrete:
> I would think using defaults based on what is available, that work on most of our machines would be the way to go. I don't understand the spectrum issue, if spectrum is used aren't you always going to want the "--gpu" option, if a gpu is available?

In principle yes, but in practice at lot of things changes (not necessarily in a transparent or documented way) from system to system,  e.g, who guarantees that the next spectrum release (or the implementation on a different system) still uses this flag (while the MPI version is still identified as "Spectrum").
I had a lot of trouble when originally trying to get K-Athena run on different systems as every system had some subtleties.
Based on this experience I'm afraid that trying to implement some logic that "generally" works hides the complexity and makes it harder to debug.
For this reason, I'm advocating a more explicit approach (i.e., properly documenting the environment for different machine and potentially be very verbose during configure as illustrated above).

> It seems to me, to be more maintainable and generally applicable to determine the appropriate flags from what applications are detected by cmake than to explicitly define machine specific setups which are susceptible to becoming stale over time and will not scale well. I think also that if you want parthenon to be of use to the wider community having some level of automation is preferable. There are a number of open source projects I could never use because I could not get an appropriate build configuration setup.

Unfortunately, the current generation of heterogeneous systems are a lot more complex than the mostly homogeneous ones from 10 years ago where a simple `mpirun` would do the trick in 95%.
Thus, I think it'd actually be of advantage to inform the user of this complexity and make a point of failure easily identifiable.
I'd probably even go a step further and look for people who're willing to maintain a machine config file (e.g., by checking that the current version in the repo works with the current environment available on a machine once every 3 months or so).
Based on current allocations I'd be happy to maintain configurations for NASA's Pleiades/Electra, Summit, Stampede2, Frontera, and Longhorn.

Yurlungur:
I lean towards @pgrete's point of view on this. I like the suggestion of `cmake`-specific machine files. @AndrewGaspar didn't you implement this before? I thought it even made it into the build system? 

One reason I prefer this is that as the code ages, the software stacks we need to support are going to change, and so are the configuration options. And if we want to maintain backwards compatibility, the `cmake` logic for the "generic" case is going to become very dense. 

I understand this isn't the `cmake` way... `cmake` and `autoconf`, etc., are supposed to make building as automatic as possible. But I worry that this means that eventually a user who doesn't know `cmake` may try to build `parthenon` on some exotic system and simply discover they can't figure it out.

Perhaps a compromise would be to try to do our best to automate things, but when that fails, allow for and document how to override `cmake` and tell it what to do. i.e., we could support both this MR and @pgrete's suggestion.

Yurlungur:
I.e., my preferred compromise is @JoshuaSBrown's suggestion of a hierarchy of defaults.

pgrete:
After some further digging and testing (trying to make things work on Summit) I came across a lot of related problems, see #254 #255 #257 that all make me wonder if larger overhaul of the testing infrastructure is the right approach to the issue here.
I think the individual changes are all small (in number of lines of code) but would in the long run simplify testing and getting things work on different machines.

AndrewGaspar:
At some point this will get pulled into a Darwin machine configuration file.

PR #249, opened 2020-08-04
Officially Create Version 0.1.0

AndrewGaspar:
## PR Summary

This change primarily sets the version number in CMakeList.txt and puts some boiler plate about the initial release in CHANGELOG.md.

Also changes the rules a little about tagged releases - prefixes the releases with "v", and also calls for branches that track the latest minor release for each major version.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,5 +10,6 @@
 - dolor (#911)
 
 ## Release 0.1.0
+Date: 8/4/2020
 
 Initial release of Parthenon AMR infrastructure.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,25 +3,12 @@
 ## Current develop
 
 ### Added (new features/APIs/variables/...)
-- Introduced CHANGELOG.md
-
-### Changed (changing behavior/API/variables/...)
-
-### Fixed (not changing behavior/API/variables/...)
-
-### Removed
-
-
-## Release 0.MAJOR.MINOR
-Note:
-- will only contain the subsections of develop that were filled
-- will (ideally) include the PR that introduced the change
-
-### Added (new features/APIs/variables/...)
 - lorem (#11)
 - ipsum (#14)
 
 ### Fixed (not changing behavior/API/variables/...)
-
 - dolor (#911)
 
+## Release 0.1.0
+
+Initial release of Parthenon AMR infrastructure.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,11 +3,12 @@
 ## Current develop
 
 ### Added (new features/APIs/variables/...)
-- lorem (#11)
-- ipsum (#14)
+
+### Changed (changing behavior/API/variables/...)
 
 ### Fixed (not changing behavior/API/variables/...)
-- dolor (#911)
+
+### Removed
 
 ## Release 0.1.0
 Date: 8/4/2020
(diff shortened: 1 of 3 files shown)


JoshuaSBrown:
I thought we were going to be including a date in the version... or am I confused by something else?

AndrewGaspar:
That was originally discussed, but we decided to use semantic versioning instead, modified to treat the minor version as the major version for now until we reach 1.0.

AndrewGaspar:
Semantic versioning: https://semver.org/

JoshuaSBrown (approved):

CHANGELOG.md
```diff
@@ -2,26 +2,14 @@
 
 ## Current develop
 
-### Added (new features/APIs/variables/...)
-- Introduced CHANGELOG.md
-
-### Changed (changing behavior/API/variables/...)
```
pgrete, line 8: Was the 4 way skeleton removed on purpose or by accident?
I'm in favor of keeping it 4 way.

CONTRIBUTING.md
```diff
@@ -28,7 +28,8 @@ We aim at creating a new release everyone 4 to 6 weeks.
 The decision on creating a new release is made during the bi-weekly calls.
 A release consists of of merging `develop` into `stable` and create a new tag for that version
 using a modified [semantic versioning](https://semver.org/) scheme.
-Releases will be tagged `0.MAJOR.MINOR` given the current rapid development.
+Releases will be tagged `v0.MAJOR.MINOR` given the current rapid development.
+Release branches named `v0.MAJOR` will track the latest minor release.
```
pgrete: Would this result in having a persistent release branch for every major release?
I'm wondering what the use case is compared to directly using the tags, e.g., if I want to update the Parthenon submodule in a downstream code I'd have to pull the changes and commit the updated version to the downstream repo anyway.

CHANGELOG.md
```diff
@@ -2,26 +2,14 @@
...
-
-
-## Release 0.MAJOR.MINOR
-Note:
-- will only contain the subsections of develop that were filled
-- will (ideally) include the PR that introduced the change
-
 ### Added (new features/APIs/variables/...)
 - lorem (#11)
 - ipsum (#14)
```
Yurlungur: I get that these are examples... but should they stay?

Yurlungur (approved):

CHANGELOG.md
AndrewGaspar: It was an accident. I'll fix.

CONTRIBUTING.md
AndrewGaspar: Yeah, I suppose. I think it just makes it easier to get the latest ABI compatible version if you want to update from an older 0.1 to a newer 0.1, without having to know what the latest minor release is.

pgrete: Given the current rapid change I suggest that we hold off the separate branch until the first official 1.x release.
For now we could add a hint to the docs saying that with a simple `git tag --list 'v0.MAJOR.*'` one can find the latest minor release.

AndrewGaspar:
@pgrete Should be good to go now.

pgrete (approved):

pgrete:
Shall we add info from #214 (and #259) to the changelog here? I think this would be quite important as it actually changes the API.

AndrewGaspar:
I'll add info from #214. #259 doesn't seem to have any user-facing issues - just additional documentation.

PR #250, opened 2020-08-04
Add restart feature to parthenon

nmsriram:
## PR Summary

This PR provides a restart capability to parthenon and all codes built on top of it.  In theory no changes are required in physics codes to enable this feature.

Restart dumps are controlled by a parthenon/output block that has file type set to 'rst'.  We have chosen HDF5 as the restart file format.  All variables that have either the Metadata::Independent or Metadata::Restart flag are added to the restart file.  The restart file also contains a full input dump from which physics parameters are initialized.

## TODO
- [x] Document changes
- [x] Fix strange deviation from original run of the order of 1e-7
- [x] Add a test

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


src/basic_types.hpp
```diff
@@ -38,10 +38,9 @@ enum class AmrTag : int { derefine = -1, same = 0, refine = 1 };
 struct SimTime {
   SimTime() = default;
   SimTime(const Real tstart, const Real tstop, const int nmax, const int ncurr,
-          const int nout)
-      : start_time(tstart), time(tstart), tlim(tstop),
-        dt(std::numeric_limits<Real>::max()), nlim(nmax), ncycle(ncurr),
-        ncycle_out(nout) {}
+          const int nout, const Real dt_in = std::numeric_limits<Real>::max())
```
Yurlungur, line 41: Why is `dt_in` set to maximum double by default? This seems a little strange to me... should it not be set to either  `inf` or 0? Or signal NaN?

src/driver/driver.hpp
```diff
@@ -57,12 +58,15 @@ class Driver {
 class EvolutionDriver : public Driver {
  public:
   EvolutionDriver(ParameterInput *pin, Mesh *pm) : Driver(pin, pm) {
-    Real start_time = pinput->GetOrAddReal("parthenon/time", "start_time", 0.0);
-    Real tstop = pinput->GetReal("parthenon/time", "tlim");
+    Real start_time = pinput->GetOrAddPrecise("parthenon/time", "start_time", 0.0);
+    Real tstop = pinput->GetOrAddPrecise("parthenon/time", "tlim",
+                                         std::numeric_limits<Real>::max());
```
Yurlungur: I don't like this default floating around in multiple places. Maybe we could put it in `defs` as a `constexpr Real default default_time` or something? Also it really looks to me like what you want here is a sginal NaN, not just the max Real.

src/outputs/restart.hpp
```diff
@@ -0,0 +1,145 @@
...
+#ifdef HDF5OUTPUT
+    try {
+      herr_t status;
+
+      // compute block size, probaby could cache this.
+      hid_t theHdfType = getHdfType(data);
+
+      hid_t dataset = H5Dopen2(fh_, name, H5P_DEFAULT);
+      if (dataset < 0) {
+	return -1;
```
Yurlungur: I know this isn't your fault, but all these magic numbers are driving me batty. Can we define them at the top of the file, maybe?

src/basic_types.hpp
nmsriram: Which would you prefer?

src/driver/driver.hpp
nmsriram: I disagree. I think it belongs here.  If a time limit is not specified, the simulation is supposed to run 'forever'.  In terms of double precision, that 'forever' is represented by max().

src/outputs/restart.hpp
nmsriram: It should actually throw a parthenon::warn with an appropriate warning message.  In the calling code I wanted to just capture whether it is less than 0 or not. However for debugging the different numbers help narrow down where the failure occurred.  Probably makes sense to create an enum (class?) for the different failure types, but then I'd need special code for the callers rather than a simple check less than 0.  Will look at this, but is lower priority than fixing conflicts with develop.

src/basic_types.hpp
Yurlungur: I think the safest choice is probably a signal `NaN` and that's where I lean. Though maybe @jdolence @jmstone or @pgrete has thoughts on this.

src/driver/driver.hpp
Yurlungur: Hmm okay. I buy that.

Is there some reason it shouldn't be `std::numeric_limits<Real>::infinity()` then? Infinity would really be forever, as it's not just the largest double, it always compares as bigger than any double, including `std::numeric_limits<Real>::max()`?

src/outputs/restart.hpp
Yurlungur: Makes sense.

Sounds like a raw enum might be best here actually, since then it automatically casts, and you don't need special code.

src/driver/driver.hpp
nmsriram: I'll have to look up the result of comparisons with infinity. Do you happen to know off-hand?

src/outputs/restart.hpp
nmsriram: Actually now that the code is working, I'm just going to return a boolean.

nmsriram: Actually @Yurlungur it **IS** my fault.

src/driver/driver.hpp
Yurlungur: I'm pretty sure it works like you intuitively expect. Infinity > any non-inf, non-NaN > -infinity

src/outputs/restart.hpp
nmsriram: I've changed this to return either number of blocks read, or -1 on error.

src/basic_types.hpp
nmsriram: Setting it to 0 doesn't work because for whatever reason dt is not initialized in the advection example, so it takes min(dt,whatever it computes), which turns out to be 0 and the simulation never progresses (i.e. t stays at t_start).

Setting it to signaling_nan causes the simulation to stop because all comparisons return as true.  

So, I've left it as max().

tst/regression/CMakeLists.txt
```diff
@@ -46,6 +46,16 @@ list(APPEND TEST_ARGS "--driver ${PROJECT_BINARY_DIR}/example/advection/advectio
...
 
+  # Restart
+  list(APPEND TEST_DIRS restart)
+  list(APPEND TEST_PROCS ${NUM_MPI_PROC_TESTING})
+  list(APPEND TEST_ARGS "--driver ${PROJECT_BINARY_DIR}/example/advection/advection-example \
+    --driver_input ${CMAKE_CURRENT_SOURCE_DIR}/test_suites/restart/parthinput.restart \
+    --num_steps 2")
+  
+  # h5py is only neeed if hdf5 is enabled
+  list(APPEND REQUIRED_PYTHON_MODULES h5py)
```
JoshuaSBrown: This is redundant. It is added to REQUIRED_PYTHON_MODULE on line 78.

src/mesh/mesh.cpp
```diff
@@ -663,6 +657,8 @@ Mesh::Mesh(ParameterInput *pin, IOWrapper &resfile, Properties_t &properties,
 
   InitUserMeshData(pin);
 
+#if 0
```
Yurlungur: should the stuff inside this `#if` block be removed now? Or is there reason to keep it?

tst/regression/test_suites/restart/restart.py
```diff
@@ -0,0 +1,57 @@
...
+            
+        if step == 1:
+            parameters.driver_cmd_line_args = ['parthenon/job/problem_id=gold']
+        else:
+            parameters.driver_cmd_line_args = [
+                '-r',
+                'gold.out0.00001.rhdf',
+                'parthenon/job/problem_id=silver'
+            ]
+        return parameters
```
Yurlungur: clever!

src/outputs/restart.cpp
```diff
@@ -17,30 +17,582 @@
...
+
+  // set starting poing in hyperslab for our blocks and
+  // number of blocks on our PE
+
+  // open blocks tab
+  hid_t gBlocks = H5Gcreate(file, "/Blocks", H5P_DEFAULT, H5P_DEFAULT, H5P_DEFAULT);
+
+  // write Xmin[ndim] for blocks
+  {
+    Real *tmpData = new Real[num_blocks_local * 3];
```
Yurlungur: I believe this is spelled `std::vector<Real> tmpData(num_blocks_local*3);`

src/outputs/restart.cpp
```diff
@@ -17,30 +17,582 @@
...
+  }
+
+  // write Block ID
+  {
+    // LOC.lx1,2,3
+    hsize_t n;
+    int i;
+
+    n = 3;
+    auto *tmpLoc = new int64_t[num_blocks_local * n];
```
Yurlungur: `std::vector<int64_t>`...

src/outputs/restart.cpp
```diff
@@ -17,30 +17,582 @@
...
+      tmpLoc[i++] = pmb->loc.lx3;
+      pmb = pmb->next;
+    }
+    WRITEH5SLABI64("loc.lx123", tmpLoc, gBlocks, local_start, local_count, global_count,
+                   property_list);
+    delete[] tmpLoc;
+
+    // (LOC.)level, GID, LID, cnghost, gflag
+    n = 5;
+    auto *tmpID = new int[num_blocks_local * n];
```
Yurlungur: `std::vector`...

src/outputs/restart.cpp
```diff
@@ -17,30 +17,582 @@
...
+  hid_t local_DSpace = H5Screate_simple(5, local_count, NULL);
+  hid_t global_DSpace = H5Screate_simple(5, global_count, NULL);
+
+  // while we could do this as n variables and load all variables for
+  // a block at one time, this is memory-expensive.  I think it is
+  // well worth the multiple iterations through the blocks to load up
+  // one variable at a time.  Besides most of the time will be spent
+  // writing the HDF5 file to disk anyway...
+  // If I'm wrong about this, we can always rewrite this later.
+  // Sriram
```
Yurlungur, line 413: When we're running at scale, this will probably deserve some profiling. But I agree, we can rethink it later.

src/outputs/restart.cpp
```diff
@@ -17,30 +17,582 @@
...
+      vGlobalSpace = H5Screate_simple(5, global_count, NULL);
+    }
+
+    // load up data
+    hsize_t index = 0;
+    pmb = pm->pblock;
+    while (pmb != nullptr) { // for every block
+      auto ci = ContainerIterator<Real>(
+          pmb->real_containers.Get(),
+          {parthenon::Metadata::Independent, parthenon::Metadata::Restart}, true);
```
Yurlungur, line 443: Don't want to impose this on you now. But I think my preferred solution to this would be to have a step when variables are created which adds the restart flag if they're independent. Then this line can just be `{Parthenon::Metadata::Restart}`.

Food for thought for later. Shouldn't block this MR.

src/outputs/restart.cpp
```diff
@@ -17,30 +17,582 @@
...
+  auto ciX = ContainerIterator<Real>(
+      pm->pblock->real_containers.Get(),
+      {parthenon::Metadata::Independent, parthenon::Metadata::Restart}, true);
+  for (auto &vwrite : ciX.vars) { // for each variable we write
+    const std::string vWriteName = vwrite->label();
+    hid_t vLocalSpace, vGlobalSpace;
+    pmb = pm->pblock;
+    const hsize_t vlen = vwrite->GetDim(4);
+    local_count[4] = global_count[4] = vlen;
+    Real *tmpData = new Real[varSize * vlen * num_blocks_local];
```
Yurlungur: `std::vector`...

src/outputs/restart.hpp
```diff
@@ -0,0 +1,165 @@
...
+#ifdef HDF5OUTPUT
+    try {
+      herr_t status;
+
+      // compute block size, probaby could cache this.
+      hid_t theHdfType = getHdfType(data);
+
+      hid_t dataset = H5Dopen2(fh_, name, H5P_DEFAULT);
+      if (dataset < 0) {
+        return -1;
```
Yurlungur, line 66: Again, non-blocking. But I want these to be`enum`s.

src/mesh/meshblock.cpp
```diff
@@ -164,11 +164,119 @@ MeshBlock::MeshBlock(int igid, int ilid, LogicalLocation iloc, RegionSize input_
...
+  for (auto const &pkg : packages) {
+    for (auto const &q : pkg.second->AllFields()) {
+      real_container->Add(q.first, q.second);
+    }
+    for (auto const &q : pkg.second->AllSparseFields()) {
+      for (auto const &m : q.second) {
+        real_container->Add(q.first, m);
+      }
+    }
+  }
```
Yurlungur, line 253: My feeling is we repeat a lot of this code in a lot of the meshblock constructors. Maybe we can abstract it away into a helper function? This is non-blocking. Just food for thought.

Yurlungur (approved):
Overall looks good! It'll be great to have this in. Some nitpicky comments follow below.

One thing worth thinking about. The current logic loops through all meshblocks multiple times. It might be worth thinking if we can avoid that at some time in the future.

src/outputs/restart.hpp
nmsriram: You're like a bulldog, once you sink your teeth in there, you refuse to let go...

src/mesh/meshblock.cpp
nmsriram: Agreed, but this is one I'm not going to fix in this MR.
It deserves its own MR.

src/outputs/restart.hpp
Yurlungur: **woof** *No magic numbers!* **woof**

src/mesh/meshblock.cpp
Yurlungur: Works for me.

tst/regression/CMakeLists.txt
nmsriram: Done, I deleted the line 78 instance.

src/outputs/restart.cpp
nmsriram: Didn't want to use std::vector because this is passed to hdf interface as (void *).  Unnecessary casts back and forth.

src/outputs/restart.cpp
nmsriram: I'm not sure I agree with the philosophy.  But am not going to address it here.

src/outputs/restart.cpp
nmsriram: And you would be wrong.  That would only be for single-lingual programmers.  For bilingual programmers using the C interface of HDF5, this is actually spelled `void *tmpData;`

src/outputs/restart.cpp
nmsriram: ditto `void*`

src/outputs/restart.cpp
Yurlungur: Just pass it in to hdf5 as `tmpData.data()` if it's a vector. No cast necessary. Or if you're me and evil, `&(tmpData[0])`

I strongly prefer using a vector because I do not like `new` and `delete` floating around. It's asking for a memory leak.

tst/regression/CMakeLists.txt
nmsriram: fixed

src/outputs/restart.cpp
Yurlungur: Yeah that's fine. Like I said, don't want to block this MR.

src/mesh/mesh.cpp
nmsriram: Removed.  Kept because I wasn't sure it was needed.

src/outputs/restart.cpp
Yurlungur: As I mentioned in another comment, I argue for pulling out the pointer from a vector and then passing it in to HDF5. I don't like the new and deletes floating around.

src/outputs/restart.hpp
nmsriram: I'm going to defer this one.  
However I do agree with you.

nmsriram:
> Overall looks good! It'll be great to have this in. Some nitpicky comments follow below.
> 
> One thing worth thinking about. The current logic loops through all meshblocks multiple times. It might be worth thinking if we can avoid that at some time in the future.

Perhaps we could write say 10 variables at a time instead of one. That would be faster and would reduce the memory pressure.

src/outputs/restart.cpp
nmsriram: This is serious enough that I should fix it now.

Yurlungur:
> > Overall looks good! It'll be great to have this in. Some nitpicky comments follow below.
> > One thing worth thinking about. The current logic loops through all meshblocks multiple times. It might be worth thinking if we can avoid that at some time in the future.
> 
> Perhaps we could write say 10 variables at a time instead of one. That would be faster and would reduce the memory pressure.

Yeah that's something worth exploring I think. Not right now though. Step 1 is get something working and in. Step 2 is return to it if it's ever a problem.

Just thought I'd mention it so we had it in our minds when the time comes.

src/outputs/restart.hpp
Yurlungur: Yeah that's fine.

src/outputs/restart.cpp
nmsriram: You beat me into submission.

src/outputs/restart.cpp
nmsriram: Done.

src/outputs/restart.cpp
nmsriram: 👍

tst/regression/test_suites/restart/restart.py
```diff
@@ -0,0 +1,62 @@
...
+
+class TestCase(utils.test_case.TestCaseAbs):
+    def Prepare(self, parameters, step):
+            
+        if step == 1:
+            parameters.driver_cmd_line_args = ['parthenon/job/problem_id=gold']
+        else:
+            parameters.driver_cmd_line_args = [
+                '-r',
+                'gold.out0.00001.rhdf',
```
JoshuaSBrown, line 40: It might help if you use parameters.output_path to specify the location of the restart file. I would recommend using abs paths.

JoshuaSBrown: Well nevermind, looks like the tests are passing except for the coverage test. You might try rerunning it, I think the code coverage server might have had a glitch, I had an issue a few days ago with the coverage failing and I only altered the documentation.

tst/regression/test_suites/restart/restart.py
```diff
@@ -28,6 +28,10 @@
 class TestCase(utils.test_case.TestCaseAbs):
     def Prepare(self, parameters, step):
             
+        # enable coverage testing on pass where restart
+        # files are both read and written
+        parameters.coverage_status = "both"
+
```
JoshuaSBrown, line 34: Ooh that was the problem, makes sense now.

nmsriram: Yeah, I was hoping the old files would stick around, but I guess coverage is a clean checkout.

Issue #239, opened 2020-07-24
Establish Distinction between public and private headers


JoshuaSBrown:
A clear distinction in missing between what in parthenon should be publicly exposed. This makes it easy to break down stream code that depends on parthenon as a library as internal changes are likely to break down stream packages.

Yurlungur:
This is clearly important and something we need to eventually clarify and stick to. Thanks for bringing it up.

That said, I'm not sure we're ready to solidify this... It's true that we're beginning to build things downstream with `parthenon`, but we haven't written enough, I think, to know what should and should not be public. And also `parthenon` itself is still changing pretty rapidly. I think trying to keep up an API while parthenon is still in flux will be a challenge.

AndrewGaspar:
I agree with @JoshuaSBrown. My concern with waiting is that it ossifies the exported APIs.

AndrewGaspar:
My proposal for this is:
- Create separate `include` and `internal` directories for public and private headers, respectively.
- Move every currently existing header file to `include`.
- Going forward, names should be deliberately placed in either the `include` or `internal` directory
- For names that must be published but are ultimately implementation details, they should go into the `parthenon::internal` namespace

Yurlungur:
That seems reasonable to me. As discussed with @nmsriram yesterday, I also think we should try to deliberately make better use of forward declarations if we can, to reduce the amount of private headers that end up "published" via a tree of includes.

JoshuaSBrown:
Well, I will start this process by moving all headers to a public includes folder. Will probably wait until after this has been merged though. https://github.com/lanl/parthenon/pull/238

PR #252, opened 2020-08-05
Move default loop policy to meshblock. Closes #68

### LINKED ISSUE #68: Where to put default choices for `Kokkos` loop patterns?
Currently the default loop pattern used by the `kokkos_encapsulation` code is a compile-time parameter. It would be better to make it a run-time parameter. So the question, then is where to put the default, to keep the dependency graph clean. This is a continuation of my discussion with @pgrete in MR #34.

@pgrete suggests

> How about moving the "default" wrapper from the header file also to the MeshBlock so that the wrapper only provides the most flexible interface (in the sense that everything needs to be specified) whereas the MeshBlock offer the most specialized interface (which can be configured through the input file at instantiation).

which seems reasonable to me. @all, any opinions?

pgrete:
## PR Summary

This PR cleans up the default wrapper choices effectively removing one level in the nested hierarchy as the default pattern is now set at the `MeshBlock` level.
Moreover, the default pattern is now set by CMake rather than `ifdefs` in the header.

That the intermediate (now removed) wrappers were not actively used can also be seen by effectively only unit tests being affected by this change.
Already previously, the usage of `pmb->par_for` was encouraged as this is ensures that we use the correct execution space.

## PR Checklist

- [X] Code passes cpplint
- [X] New features are documented. (does not apply)
- [X] Adds a test for any bugs fixed. Adds tests for new features. (nothing fixed, no interface change)
- [X] Code is formatted


src/CMakeLists.txt
```diff
@@ -72,6 +72,30 @@ message(STATUS "PAR_LOOP_LAYOUT='${PAR_LOOP_LAYOUT}' (default par_for wrapper la
...
+endif()
+
+if (${PAR_LOOP_INNER_LAYOUT} STREQUAL "TVR_INNER_LOOP")
+  set(PAR_LOOP_INNER_LAYOUT_TAG inner_loop_pattern_tvr_tag)
+elseif (${PAR_LOOP_INNER_LAYOUT} STREQUAL "SIMDFOR_INNER_LOOP")
+  set(PAR_LOOP_INNER_LAYOUT_TAG inner_loop_pattern_simdfor_tag)
+else()
+  set(PAR_LOOP_INNER_LAYOUT_TAG loop_pattern_undefined_tag)
+endif()
+
```
Yurlungur, line 101: I'm actually not sure I like moving this into `cmake`. The preprocessor is ugly, sure, but at least there this information is all available in one place, the single kokkos abstraction file. Not sure I like having to jump around files to find this. Maybe we could put a comment in the `kokkos_abstraction.hpp` file explaining?

src/mesh/mesh.hpp
```diff
@@ -187,47 +188,55 @@ class MeshBlock {
   template <typename Function>
   inline void par_for(const std::string &name, const int &il, const int &iu,
                       const Function &function) {
-    parthenon::par_for(name, exec_space, il, iu, function);
+    // using loop_pattern_mdrange_tag instead of DEFAULT_LOOP_PATTERN for now
+    // as the other wrappers are not implemented yet for 1D loops
+    parthenon::par_for(loop_pattern_mdrange_tag, name, exec_space, il, iu, function);
```
Yurlungur: Could we hack this using 3d loops? i.e.,
```C++
parthenon::par_for(DEFAULT_LOOP_PATTERN, name, exec_space, 0,0,0,0,il,iu,
                               KOKKOS_LAMBDA(int k, int j, int i) { function (i); });
```
Also there should probably be a TODO in the code to implement the other loops. Or a comment in the docs. I can see this being a nasty gotcha when someone goes to profile performance of the different loop patterns on some future architecture.

src/kokkos_abstraction.hpp
```diff
@@ -98,122 +98,6 @@ static struct InnerLoopPatternTVR {
...
-
-#define DEFAULT_OUTER_LOOP_PATTERN outer_loop_pattern_teams_tag
-
-#ifdef TVR_INNER_LOOP
-#define DEFAULT_INNER_LOOP_PATTERN inner_loop_pattern_tvr_tag
-#elif defined SIMDFOR_INNER_LOOP
-#define DEFAULT_INNER_LOOP_PATTERN inner_loop_pattern_simdfor_tag
-#else
-#define DEFAULT_INNER_LOOP_PATTERN loop_pattern_undefined_tag
-#endif
```
Yurlungur, line 133: Maybe we could have a comment in this file listing the available loop pattern types? I think that would be useful for a new user/developer trying to understand this file.

src/mesh/mesh.hpp
```diff
@@ -187,47 +188,55 @@ class MeshBlock {
...
 
   // 2D default loop pattern
   template <typename Function>
   inline void par_for(const std::string &name, const int &jl, const int &ju,
                       const int &il, const int &iu, const Function &function) {
-    parthenon::par_for(name, exec_space, jl, ju, il, iu, function);
+    // using loop_pattern_mdrange_tag instead of DEFAULT_LOOP_PATTERN for now
+    // as the other wrappers are not implemented yet for 1D loops
+    parthenon::par_for(loop_pattern_mdrange_tag, name, exec_space, jl, ju, il, iu,
+                       function);
```
Yurlungur, line 202: Here too.

src/reconstruct/plm_inline.hpp
```diff
@@ -47,12 +47,13 @@ void PiecewiseLinearX1(parthenon::team_mbr_t const &member, const int k, const i
 
   // compute L/R slopes for each variable
   for (int n = 0; n <= nu; ++n) {
-    parthenon::par_for_inner(member, il, iu, [&](const int i) {
-      // renamed dw* -> dq* from plm.cpp
-      dql(n, i) = (q(n, k, j, i) - q(n, k, j, i - 1));
-      dqr(n, i) = (q(n, k, j, i + 1) - q(n, k, j, i));
-      qc(n, i) = q(n, k, j, i);
-    });
+    parthenon::par_for_inner(DEFAULT_INNER_LOOP_PATTERN, member, il, iu,
```
Yurlungur: Why not pmb->par_for_inner? Maybe we should pass the meshblock in to the reconstruction routines?

src/reconstruct/dc_inline.hpp
```diff
@@ -39,7 +39,8 @@ void DonorCellX1(parthenon::team_mbr_t const &member, const int k, const int j,
   // compute L/R states for each variable
   for (int n = 0; n <= nu; ++n) {
     parthenon::par_for_inner(
-        member, il, iu, [&](const int i) { ql(n, i + 1) = qr(n, i) = q(n, k, j, i); });
+        DEFAULT_INNER_LOOP_PATTERN, member, il, iu,
```
Yurlungur: Here too

Yurlungur (changes requested):
Overall, this looks good. There's a few places the abstraction leaks, which I'd like to close up.

AndrewGaspar (approved):

src/mesh/mesh.hpp
pgrete: In principle this could be hacked into 3D loops (and even more so this would apply to other pieces, too, e.g., we could even go as far as expressing everything in terms of 4D loops).
On the other hand, I'm not sure that it'd clean up the code too much and we should measure overhead first.
How strongly do you feel about this? Should we address this now, open an issue to keep track of it, or is the TODO/warning fine for now?

Yurlungur: It's not a huge deal. The TODO/warning is fine for now. 

Actually maybe let's just add a comment about this in the docs too... just so somebody doesn't report incorrect numbers without looking at the code.

src/CMakeLists.txt
pgrete: I looked at it from the "configure" point of view. It's a compile time variable that is set while executing cmake. Thus, I thought it's most appropriate to have the logic in cmake.
I will definitely add a comment to point users/devs to the right place.

Yurlungur: Works for me.

src/reconstruct/plm_inline.hpp
pgrete: There were two reason
- from a practical point of view I didn't want to change the interface
- from a logical point of view there's no reason to use the `pmb->par_for_inner` as it's a 1:1 match to the `parthenon::par_for_inner(DEFAULT_INNER_LOOP_PATTERN,...`. This is different from the `pmb->par_for` where we want/need to use the `pmb` as the ExecutionSpace is tied to the MeshBlock. At the "inner" level there's no ExecutionSpace as the corresponding information is passed by the `member` variable.

Yurlungur: > At the "inner" level there's no ExecutionSpace as the corresponding information is passed by the member variable.

Ah, I don't think I appreciated this. In that case, perhaps `parthenon::par_for_inner(member, il, iu...` should still be supported? I.e., without the default loop pattern needing to be specified?

pgrete: Yes, that's a good point. I'll add this.

src/mesh/mesh.hpp
pgrete: Good point. I added a note on that.

src/kokkos_abstraction.hpp
pgrete: The loop patterns/tags are still defined in this file (see above this removed block of code). I added some more info on what the tags stand for.

src/reconstruct/plm_inline.hpp
pgrete: I now added this to `mesh.hpp` which I'm not super happy about.
The main reason I put it there is that `kokkos_abstraction` is currently independent of `config.hpp`, which indicates that the abstraction is properly separated, and because mesh.hpp contains the other wrappers (for the MeshBlock).
Nevertheless, this still feels like the overall interface may benefit from an overhaul/cleaner design, which is likely also linked to #239.

pgrete:
@Yurlungur I tried to address all your comments. This should be ready for re-review now.

src/reconstruct/plm_inline.hpp
Yurlungur: Yeah I don't love it there, but I agree with your reasoning. It think that's where it should go for now.

Yurlungur (approved):
Thanks for addressing all my nitpicks. Looks good now.

PR #263, opened 2020-08-11
Variable packs over the entire mesh

Yurlungur:
## PR Summary

This pull request is motivated by the performance discussion on Monday. Since we are now looking at coalescing kernels over meshblocks, I decided to see if I could throw together a prototype for a piece of what we need: a data structure we can capture and pass into a Kokkos kernel.

The path I decided to go down was to generalize the `VariablePack` mechanism @jdolence implemented. I implement here a `MeshPack` object and `PackMesh` method, which takes essentially the same arguments as `PackVariables` or `PackVariablesAndFluxes`. But it also takes a container and a mesh pointer. The resulting object is a view of VariablePacks. The way to interact with it is as a 5D view, where the indices are: `MeshBlock`, `Variable`, `k`, `j`, `i`.

I decided to go down this path mainly because it seemed straightforward. We don't have to rewrite the underlying data layout or make high-level objects like `MeshBlock` CUDA-friendly. 

To test it out, I implemented an alternate version of the `pi_example` that computes the area step with a global Kokkos kernel rather than a task.

### A few caveats

- I have done **zero** performance testing. If this is a path we want to go down, we need to test this thing out more rigorously.
- It needs to be a little more general to support buffer packs (which aren't needed in the pi example). The boundary buffers need to be addable to this pack object. I don't think this is an impediment though. It should be straightforward.
-  I implemented but didn't test this packing for `FluxPack`s. That's for the future, I think.
- At the moment, I restricted this packing to pack all blocks on an MPI rank. I just wanted to see if I could do it. It should be straightforward to select a set of block IDs or partition into N block objects, each owning some (roughly) equal number of blocks.
- A major missing piece of this machinery is a way of integrating tasks with these superblocks or bulk-synchronous operations. Ideally, we could seamleslsy weave together tasks on individual blocks and on these superblocks. 

### Concluding thoughts

To be clear, I'm not proposing this as the final implementation we should merge in. Rather, I wanted to just share it with you all to start a discussion and see what you think and whether we should pursue something like this. I was pleasantly surprised how easy it was to implement this thing. The infrastructure we've written so far is flexible enough to support it with very little suffering.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/docs/mesh/packing.md
+++ b/docs/mesh/packing.md
@@ -39,9 +39,9 @@ For example:
 // pack all variables on the base container accross the whole mesh
 auto meshpack = PackVariablesAndFluxes(pmesh, "base");
 auto variablepack = meshpack(m); // Indexes into the m'th meshblock
-auto auto var = mesh(m,l); // Indexes into the k'th variable on the m'th MB
+auto var = meshpack(m,l); // Indexes into the k'th variable on the m'th MB
 // The l'th variable in the i,j,k'th cell of the m'th meshblock
-Real r = mesh(m,l,k,j,i); 
+Real r = meshpack(m,l,k,j,i); 
 ```
 
 For convenience, `MeshPack` also includes the following methods and fields:
--- a/docs/mesh/packing.md
+++ b/docs/mesh/packing.md
@@ -1,0 +1,65 @@
+# Packing Variables Accross the Entire Mesh
+
+`Kokkos` kernel launches have a 6 microsecond overhead. For small
+kernels that perform little work, this can be a perforamnce bottleneck
+when each kernel is launched per meshblock. Parthenon therefore
+provides the capability to combine variables into a single data
+structure that spans all meshblocks, the `MeshPack`. 
+
+## Creating a Mesh Pack
+
+There are two methods for creating mesh packs, which are analogous to the `VariablePack` and `VariableFluxPack` available in [containers](../interface/containers.md).
+```C++
+template <typename... Args>
+auto PackVariablesOnMesh(Mesh *pmesh, const std::string &container_name,
+                         Args &&... args)
+```
+and
+```C++
+auto PackVariablesAndFluxesOnMesh(Mesh *pmesh, const std::string &container_name,
+                                  Args &&... args)
+```
+The former packs only the variables, the latter packs in the variables and associated flux vectors.
+
+The variatic arguments take exactly the same arguments as
+`container.PackVariables` and `container.PackVariablesAndFluxes`. You
+can pass in a metadata vector, a vector of names, and optionally IDs
+and a map from names to indices. See
+[here](../interface/containers.md) for more details.
+
+## Data Layout
+
+The Mesh Pack is indexable as a five-dimensional `Kokkos` view. The
+slowest moving index indexes into a 4-dimensional `VariablePack`. The
+next slowest indexes into a variable. The fastest three index into the
+cells on a meshblock.
+
+For example:
+```C++
+// pack all variables on the base container accross the whole mesh
+auto meshpack = PackVariablesAndFluxes(pmesh, "base");
+auto variablepack = meshpack(m); // Indexes into the m'th meshblock
+auto auto var = mesh(m,l); // Indexes into the k'th variable on the m'th MB
+// The l'th variable in the i,j,k'th cell of the m'th meshblock
+Real r = mesh(m,l,k,j,i); 
+```
+
+For convenience, `MeshPack` also includes the following methods and fields:
+```C++
+// the IndexShape object describing the bounds of all blocks in the pack
+IndexShape bounds = meshpack.cellbounds; 
+
+// a 1D array of coords objects
+auto coords = meshpack.coords(m); // gets the Coordinates_t object on the m'th MB
+
+// The dimensionality of the simulation. Will be 1, 2, or 3.
+// This is needed because components of the flux vector
+// are only allocated for dimensions in use.
+int ndim = meshpack.GetNdim();
+
+// Get the sparse index of the n'th sparse variable in the pack.
+int sparse = meshpack.GetSparse(n);
+
+// The size of the n'th dimension of the pack
+int dim = meshpack.GetDim(n);
+```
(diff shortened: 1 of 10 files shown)


pgrete:
FYI I started to test/benchmark this implementation, see https://github.com/pgrete/parthenon/blob/jmm/pack-mesh/example/pack_perf/pack_perf.cpp#L104 (and the entire pack_perf example).
It currently contains just a simple test for the FluxDivergence kernel. No other tuning etc. yet.
Results are as follows
image
and expected to change once we start playing with loop patterns etc.
Overall, I'd say this looks like promising way forward.

Yurlungur:
> FYI I started to test/benchmark this implementation 

This looks really promising! And also, I think, matches our expectations that the preferred regime will depend on meshblock size, and perhaps hardware. That makes a compelling case that, once we start really building on this, we should make it flexible enough to make variable-sized packs.

pgrete:
image

Update using flat ranges for MeshPack. Looks really promising (if I didn't miss anything)

src/mesh/mesh_pack.hpp
```diff
@@ -101,13 +98,13 @@ auto PackMesh(const Mesh *pmesh, F &packing_function) {
   Kokkos::deep_copy(packs, packs_host);
   Kokkos::deep_copy(coords, coords_host);
 
-  return MeshPack<T>(packs, pmesh->pblock->cellbounds, coords, dims);
+  return MeshPack<T>(packs, pmesh->block_list.front().cellbounds, coords, dims);
```
AndrewGaspar, line 102: Not your fault, but I really hate this `block_list.front().blah` stuff. It's like, hm... should this attribute really hang off the `MeshBlock`?

Yurlungur: Yeah I was thinking about that actually. Was considering fixing that in another PR.

Yurlungur:
@nmsriram @pgrete @AndrewGaspar this is ready for merge. @jdolence and anyone else interested if you wanna give it a cursory review too please do. :)

docs/mesh/packing.md
```diff
@@ -0,0 +1,65 @@
...
+slowest moving index indexes into a 4-dimensional `VariablePack`. The
+next slowest indexes into a variable. The fastest three index into the
+cells on a meshblock.
+
+For example:
+```C++
+// pack all variables on the base container accross the whole mesh
+auto meshpack = PackVariablesAndFluxes(pmesh, "base");
+auto variablepack = meshpack(m); // Indexes into the m'th meshblock
+auto auto var = mesh(m,l); // Indexes into the k'th variable on the m'th MB
```
AndrewGaspar: The elusive `auto auto`.

AndrewGaspar: Is `mesh` supposed to be `meshpack` here?

docs/mesh/packing.md
```diff
@@ -0,0 +1,65 @@
...
+// The l'th variable in the i,j,k'th cell of the m'th meshblock
+Real r = mesh(m,l,k,j,i);
```
AndrewGaspar: Again, is this supposed to be `meshpack`?

src/mesh/mesh_pack.hpp
```diff
@@ -0,0 +1,129 @@
...
+template <typename T>
+using ViewOfPacks = ParArray1D<VariablePack<T>>;
+template <typename T>
+using ViewOfFluxPacks = ParArray1D<VariableFluxPack<T>>;
+template <typename T>
+using MeshVariablePack = MeshPack<VariablePack<T>>;
+template <typename T>
+using MeshVariableFluxPack = MeshPack<VariableFluxPack<T>>;
+
+// TODO(JMM): Should this be cached?
```
AndrewGaspar, line 76: I think the issue with caching with the mesh pack in its current form is that if you do the pack over the whole mesh, you would basically invalidate the cache every cycle. However, you could, I suppose, over allocate the pack sizes and grow the caches automatically. We'll probably? need caching, but it's an open question how  we do it.

AndrewGaspar (approved):

src/mesh/mesh_pack.hpp
jdolence: cache would only be invalidated when the meshblocks on a node change, so _possibly_ when AMR and/or load balancing are done.  In a block AMR code, that's typically not every cycle.

docs/mesh/packing.md
Yurlungur: good catches thanks

docs/mesh/packing.md
Yurlungur: yep. Thanks for the catch.

src/mesh/mesh_pack.hpp
Yurlungur: Another factor: caching is already *kind* of implemented because `MeshPack` leverages the `VariablePack`, which is cached. I think we need to measure to see how much it hurts us.

pgrete:
One post merge comment I mentioned on Matrix:
Conceptually, I'd have gone with "MeshBlockPack" to indicate that it's a pack of MeshBlocks (especially once we want to have the capability to pack subsets of MeshBlocks) similarly to a VariablePack being a pack of Variables.
A MeshPack may indicate a pack of Meshes (though we only have one).

Yurlungur:
> One post merge comment I mentioned on Matrix:
> Conceptually, I'd have gone with "MeshBlockPack" to indicate that it's a pack of MeshBlocks (especially once we want to have the capability to pack subsets of MeshBlocks) similarly to a VariablePack being a pack of Variables.
> A MeshPack may indicate a pack of Meshes (though we only have one).

I prefer the name `MeshPack` over `MeshBlockPack`. It's shorter and easier to say, which I think which matters. It's just a name, so I'm willing to be overridden. But that's my two cents. Does anyone else have thoughts?

Yurlungur:
> > One post merge comment I mentioned on Matrix:
> > Conceptually, I'd have gone with "MeshBlockPack" to indicate that it's a pack of MeshBlocks (especially once we want to have the capability to pack subsets of MeshBlocks) similarly to a VariablePack being a pack of Variables.
> > A MeshPack may indicate a pack of Meshes (though we only have one).
> 
> I prefer the name `MeshPack` over `MeshBlockPack`. It's shorter and easier to say, which I think which matters. It's just a name, so I'm willing to be overridden. But that's my two cents. Does anyone else have thoughts?

Assuming I am overruled, I will submit a PR to fix the naming after @jdolence 's branch and the integration branch are in. That way I'll only have to do a `sed` once instead of three times.

PR #266, opened 2020-08-17
Enable openmp in kokkos if openmp is found 

JoshuaSBrown:
## PR Summary

It does not make sense that parthenon has openmp enabled by default but that it is not used by Kokkos unless Kokkos_ENABLE_OPENMP is enabled, this is redundant, this pull request attempts to address this by using parthenon's defaults. 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,3 +17,4 @@ Initial release of Parthenon AMR infrastructure.
 
 ### Changed
 [[PR 214]](https://github.com/lanl/parthenon/pull/214): The weak linked routines for user-specified parthenon behavior have been removed in favor of a more portable approach. See [the documentation](docs/README.md#user-specified-internal-functions).
+[[PR 266]](https://github.com/lanl/parthenon/pull/266): It is no longer necessary to specify Kokkos_ENABLE_OPENMP this is by default enabled, to turn off one can specify PARTHENON_DISABLE_OPENMP.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added (new features/APIs/variables/...)
 
 ### Changed (changing behavior/API/variables/...)
+[[PR 266]](https://github.com/lanl/parthenon/pull/266): It is no longer necessary to specify Kokkos_ENABLE_OPENMP this is by default enabled, to turn off one can specify PARTHENON_DISABLE_OPENMP.
 
 ### Fixed (not changing behavior/API/variables/...)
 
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -120,6 +120,7 @@ if (NOT PARTHENON_DISABLE_OPENMP)
     "If you want to build Parthenon without OpenMP, please rerun CMake with -DPARTHENON_DISABLE_OPENMP=ON")
   endif()
   set(ENABLE_OPENMP ON)
+  set(Kokkos_ENABLE_OPENMP ON)
```
AndrewGaspar: Could you make this a cache variable instead so that it doesn't override the user setting if the user explicitly disables it?

CHANGELOG.md
```diff
@@ -17,3 +17,4 @@ Initial release of Parthenon AMR infrastructure.
 
 ### Changed
 [[PR 214]](https://github.com/lanl/parthenon/pull/214): The weak linked routines for user-specified parthenon behavior have been removed in favor of a more portable approach. See [the documentation](docs/README.md#user-specified-internal-functions).
+[[PR 266]](https://github.com/lanl/parthenon/pull/266): It is no longer necessary to specify Kokkos_ENABLE_OPENMP this is by default enabled, to turn off one can specify PARTHENON_DISABLE_OPENMP.
```
AndrewGaspar: This should be moved up to the `Changed` section under develop - v0.1.0 has already been released.

AndrewGaspar (changes requested):

CMakeLists.txt
JoshuaSBrown: Isn't that redundant with ```PARTHENON_DISABLE_OPENMP```, I could always warn the user to use ```PARTHENON_DISABLE_OPENMP``` instead if they are trying to alter `Kokkos_ENABLE_OPENMP` and tell them to use the parthenon equivalent.

JoshuaSBrown: ping @AndrewGaspar

Yurlungur (approved):
No objections from me. I defer to you and @AndrewGaspar as far as the `cmake` fu.

CMakeLists.txt
AndrewGaspar: I want it to be redundant - you might be in a situation where Kokkos + OpenMP is on the fritz for some reason, but still want to enable Parthenon's raw OpenMP. Making it a cache variable gives you an out so that you can disable Kokkos' OpenMP without disabling Parthenon's OpenMP.

AndrewGaspar (approved):

pgrete (approved):

Issue #221, opened 2020-07-13
Buffer Packing Kernel


forrestglines:
After performance discussions in #189, we're fairly confident that the launch overhead from launching one kernel for each buffer of communicated ghostzones is hampering performance when using small ( less than 256^3) meshblocks. One solution would be to combine the packing and unpacking of each buffer into a single kernel for packing all buffers and a single kernel for unpacking all buffers. We can discuss and share solutions here.

We can also start with a kernel for just uniform grids, packing the 6 faces, 12 edges, and 8 corners of a box into 26 buffers. With AMR there are 57 buffers. Note that for the face-centered variables, the box will not necessarily be a cube; the box might be one longer along an axis.

forrestglines:
@pgrete @jmstone Here's the issue to collaborate on the buffer packing kernels. 

My unfinished kernel to do the buffer packing in CUDA is here:

https://github.com/forrestglines/buffer_packing/blob/master/tests/cuda_buffer_packing/cuda_buffer_packing.cu

I started with a version in CUDA since I wanted to be exactly sure of how the kernel was being parallelized. After getting that working, I was planning on moving it to Kokkos.

pgrete:
Ping @cielling @gshipman 
This is where input by Nvidia folks would be great. Efficient ghost zone/guard cell/buffer packing functions are likely a solved problem.

pgrete:
What's the status here? IIRC we discussed on Wednesday to make a minor update here and then ask for help?

forrestglines:
@pgrete Sorry, I got caught up in other projects. My first attempt doesn't work - it was running out of shared memory at larger sizes and I can't figure out some bugs at smaller sizes. I'm taking a simpler approach that I hope to have ready today.

forrestglines:
I have an almost entirely working packing function running with Kokkos at 
https://github.com/forrestglines/buffer_packing/blob/master/tests/kokkos_buffer_packing/kokkos_buffer_packing.cpp

The packing kernel works fine and matches a cpp version for all grid sizes on CPUs and grid sizes up to 64. For larger grids, Kokkos seems to be using a larger thread block size than is supported by the device - I don't know why. The parthenon wrappers might avoid the issue.


Timing looks like this:

![image](https://user-images.githubusercontent.com/5799640/88029575-2dfdb500-cb08-11ea-960d-2b39830cbe9f.png)

It looks like it's dominated by overhead at small grid sizes. Time should scale as nx^2 at larger sizes.

[...]

PR #267, opened 2020-08-17
Generalize Tasking

jdolence:
## PR Summary

This PR introduces a little more generality into the tasking to accomplish two primary goals:
1. Allow for tasks that operate on more than one `MeshBlock`, which was difficult/impossible to capture with only a `TaskList` per `MeshBlock` model
2. Introduce enough generality to support more complicated notions of a "cycle" that might include different time-stepping strategies, global solves, etc.

Supports #221 when used in conjunction with #263.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/docs/tasks.md
+++ b/docs/tasks.md
@@ -15,3 +15,36 @@ Examples of both `AddTask` calls can be found in the advection example [here](..
 
 ## TaskID
 The `TaskID` class implements methods that allow Parthenon to keep track of tasks, their dependencies, and what remains to be completed.  The main way application code will interact with this object is as a returned object from `TaskList::AddTask` and as an argument to subsequent calls to `TaskList::AddTask` as a dependency for other tasks.  When used as a dependency, `TaskID` objects can be combined with the bitwise or operator (`|`) to specify multiple dependencies.
+
+ ## TaskRegion
+ `TaskRegion` is defined via a simple using statement as a `std::vector<TaskList>`.  During task execution (described below), all task lists in a `TaskRegion` can be operated on concurrently.  For example, a `TaskRegion` can be used to construct independent task lists for each `MeshBlock`.
+
+ ## TaskCollection
+ A `TaskCollection` contains a `std::vector<TaskRegion>`, i.e. an ordered list of `TaskRegion`s.  Importantly, each `TaskRegion` will be executed to completion before subsequent `TaskRegion`s, introducing a notion of sequential execution and enabling flexibility in task granularity.  For example, the following code fragment uses the `TaskCollection` and `TaskRegion` abstractions to express work that can be done asynchronously across blocks, followed by a bulk synchronous task involving all blocks, and finally another round of asynchronous work.
+ ```c++
+ TaskCollection tc;
+ TaskRegion &tr1 = tc.AddRegion(nmb);
+ for (int i = 0; i < nmb; i++) {
+   auto task_id = tr1[i].AddTask(foo, dep, args, blocks[i]);
+ }
+
+ {
+   TaskRegion &tr2 = tc.AddRegion(1);
+   auto sync_task = tr2[0].AddTask(bar, dep, args, blocks);
+ }
+
+ TaskRegion &tr3 = tc.AddRegion(nmb);
+ for (int i = 0; i < nmb; i++) {
+     auto task_id = tr3[i].AddTask(foo, dep, args, blocks[i]);
+ }
+ ```
+
+ `TaskCollection` provides two member functions, `AddRegion` and `Execute`.
+
+ ### AddRegion
+
+ `AddRegion` simply adds a new `TaskRegion` to the back of the collection and returns it as a reference.  The integer argument determines how many task lists make up the region.
+
+ ### Execute
+
+ Calling the `Execute` method on the `TaskCollection` executes all the tasks that have been added to the collection, processing each `TaskRegion` in the order they were added, and allowing tasks in different `TaskList`s but the same `TaskRegion` to be executed concurrently.
--- a/example/calculate_pi/calculate_pi.cpp
+++ b/example/calculate_pi/calculate_pi.cpp
@@ -23,6 +23,8 @@
 
 // Parthenon Includes
 #include <coordinates/coordinates.hpp>
+#include <kokkos_abstraction.hpp>
+#include <mesh/mesh_pack.hpp>
 #include <parthenon/package.hpp>
 
 using namespace parthenon::package::prelude;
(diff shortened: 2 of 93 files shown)


example/calculate_pi/calculate_pi.cpp
```diff
@@ -77,28 +79,27 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
...
-        larea += v(k, j, i) * coords.Area(parthenon::X3DIR, k, j, i);
-      },
-      area);
-  Kokkos::deep_copy(pmb->exec_space, v.Get(0, 0, 0, 0, 0, 0), area);
+TaskStatus ComputeAreas(std::vector<MeshBlock *> &blocks) {
+  IndexRange ib = blocks[0]->cellbounds.GetBoundsI(IndexDomain::interior);
+  IndexRange jb = blocks[0]->cellbounds.GetBoundsJ(IndexDomain::interior);
+  IndexRange kb = blocks[0]->cellbounds.GetBoundsK(IndexDomain::interior);
 
+  for (auto pmb : blocks) {
```
Yurlungur, line 113: I notice that the infrastructure now uses `std::vector<MeshBlock*>` as an underlying data structure for the tasking. We may want to rethink this at some point. It should probably be consistent with @AndrewGaspar 's MR #193 . Of course, we may want to change `std::list` there to `std::vector` too.

jdolence: yeah that's not a bad idea.  actually it may be trivial to support both.

Yurlungur: For now I wouldn't worry about it. But we should keep it in mind. I think they may come into alignment at some point, because the performance bottleneck for `FindMeshBlock` I think benefits from `blocks` being a `std::vector` instead of a `std::list`.

Yurlungur: At least that's what I remember @jmstone saying?

Yurlungur (approved):
Changelist needed. Otherwise looks good.

example/calculate_pi/calculate_pi.cpp
jmstone: Yes, performance of FindMeshBlock is improved using a std::vector compared to std::list when there are very many MBs on a single rank.  Also beware of subtle bugs related to the way Athena++ uses pointers to parent objects (e.g. pmy_mblock) stored in classes when STL containers are used to store the MB list instead of the original hand-rolled linked list.  It seems the address of a pointer to an object added to an STL container is not reliable until the container is fully constructed.  This caused a bug it took me several days to find in my new test code.  Perhaps the container class used in Parthenon avoids this.

Yurlungur: > Also beware of subtle bugs related to the way Athena++ uses pointers to parent objects (e.g. pmy_mblock) stored in classes when STL containers are used to store the MB list instead of the original hand-rolled linked list. It seems the address of a pointer to an object added to an STL container is not reliable until the container is fully constructed. 

Oh man. That's a nasty gotcha. Thanks for the warning! That's something we should keep in mind.

docs/tasks.md
```diff
@@ -14,4 +14,37 @@ Examples of both `AddTask` calls can be found in the advection example [here](..
...
+ }
+
+ TaskRegion &tr3 = tc.AddRegion(nmb);
+ for (int i = 0; i < nmb; i++) {
+     auto task_id = tr3[i].AddTask(foo, dep, args, blocks[i]);
+ }
+ ```
+
+ `TaskCollection` provides two member functions, `AddRegion` and `Execute`.
+
```
JoshuaSBrown, line 43: This description could really benefit from a diagram, illustrating the different components. Though, it would be more appropriate to include something like that in the wiki. Including something like that might be wishful thinking on my part though. I'm not quite sure we have a good way of including images related to documentation at this point.

example/calculate_pi/calculate_pi.cpp
```diff
@@ -77,28 +79,27 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
...
-                                             {1, 1, ib.e + 1 - ib.s}),
-      KOKKOS_LAMBDA(int k, int j, int i, Real &larea) {
-        larea += v(k, j, i) * coords.Area(parthenon::X3DIR, k, j, i);
-      },
-      area);
-  Kokkos::deep_copy(pmb->exec_space, v.Get(0, 0, 0, 0, 0, 0), area);
+TaskStatus ComputeAreas(std::vector<MeshBlock *> &blocks) {
+  IndexRange ib = blocks[0]->cellbounds.GetBoundsI(IndexDomain::interior);
+  IndexRange jb = blocks[0]->cellbounds.GetBoundsJ(IndexDomain::interior);
+  IndexRange kb = blocks[0]->cellbounds.GetBoundsK(IndexDomain::interior);
```
JoshuaSBrown: These can be made const.

example/calculate_pi/calculate_pi.cpp
```diff
@@ -77,28 +79,27 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
...
 
+  for (auto pmb : blocks) {
+    auto &rc = pmb->real_containers.Get();
+    auto &coords = pmb->coords;
```
JoshuaSBrown: It looks like coords can also be made const.

src/tasks/task_list.hpp
```diff
@@ -110,10 +102,39 @@ class TaskList {
...
+using TaskRegion = std::vector<TaskList>;
+
+struct TaskCollection {
+  TaskCollection() = default;
+  TaskRegion &AddRegion(const int num_lists) {
+    regions.push_back(TaskRegion(num_lists));
+    return regions.back();
+  }
+  TaskListStatus Execute() {
+    for (auto region : regions) {
```
JoshuaSBrown: Is there a reason this cannot be a reference?

JoshuaSBrown (approved):
Looks good to me, just a question or two to clarify behavior.

example/calculate_pi/calculate_pi.cpp
JoshuaSBrown: > For now I wouldn't worry about it. But we should keep it in mind. I think they may come into alignment at some point, because the performance bottleneck for `FindMeshBlock` I think benefits from `blocks` being a `std::vector` instead of a `std::list`.

I've run into the same problem before, some of the stl containers are safer than others, particulalry the node based containers, pointers and iterators to items in a list, map, set, multimap and multiset are guaranteed to remain valid as long as the element is not removed from the container. Like @jmstone alluded too, if you add an element to a vector and it has to reallocate memory then all the items in the vector may be copied to a new location, the same can happen for unordered maps and the other stl containers. If you never change the size of your vector you can get around this limitation. The boost library even has a stable vector which addresses this problem.

docs/tasks.md
Yurlungur: What kind of diagram are you thinking, @JoshuaSBrown ? It might be possible to draw one in ascii

pgrete:
@Yurlungur given that the MeshPacks are now in develop, I think it'd make sense to close #282 and add the integration to this PR.
If I see this correctly, this is required anyway in order to properly address the current conflict, isn't it?

Yurlungur:
@pgrete yes good point. Alternatively, we could close this one and merge in #282, which might be a smoother process. Should be equivalent either way, so long as review comments are all addressed.

Thoughts, @jdolence ?

example/calculate_pi/calculate_pi.cpp
Yurlungur: Yes, it makes sense. so `std::vector<T>::push_back` might copy elements. 

I guess this means replacing elements in a vector with `std::shared_ptr` would also fix the problem, as the pointers are what would get copied.

src/tasks/task_list.hpp
jdolence: don't think so...fixed.

jdolence:
@pgrete @Yurlungur I think I got everything merged appropriately.

pgrete (approved):
Pending open comments and integration I think this looks good to go (and getting some field testing).

docs/tasks.md
jdolence: I guess I'm not sure what picture I would draw even if it was easy to include a picture.

JoshuaSBrown: I don't know if this is accurate...

TaskDiagram

Yurlungur: @JoshuaSBrown whoa nice work throwing that together.

Am I reading your picture correctly that the green blocks are tasks? And if they're one after the other, it's synchronous and if they're next to each other it's asynchronous? That's not inaccurate, I think. It doesn't really capture that tasks can be over one meshblock or many, and that if they're over one meshblock in a synchronous region, each meshblock likely individually owns the task. Not sure how to get that accross in a picture. Maybe we don't have to.

Regarding embedding a picture in the docs... I don't know if we want to do this, but we could just stick the picture in the `docs` directory and link to it with markdown `name`, which will make it appear embedded in github at least.

Could also try throwing it through a png->ascii converger. I think your graph is simple enough it should be representable as ascii

jdolence: @JoshuaSBrown can we iterate on this some and add it in a subsequent PR?

JoshuaSBrown: That's fine with me.

jdolence:
ok I think this is ready.  pulling the trigger.

example/calculate_pi/calculate_pi.cpp
AndrewGaspar: `std::list` is fine in terms of iterator invalidation. By design, adding or removing elements to a `std::list` doesn't invalidate iterators (or pointers to stored objects) unless you remove the object the iterator is pointing to.

To get a `std::vector` without pointer invalidation, just do `std::vector<std::unique_ptr<T>>`

Issue #264, opened 2020-08-16
Case sensitive input parsing


jdolence:
Right now, the input file parsing is case sensitive.  There are reasonable arguments in favor of this, but it can be a little error prone.  For example, with `<Parthenon/mesh>` instead of `<parthenon/mesh>`, the code aborts because required mesh parameters aren't found.  I just wasted more time than I'd like to admit debugging a problem that turned out to be an incarnation of this basic issue.

I propose we just transform all block and parameter names (but not values on the RHS of the =) to all lowercase internal to Parthenon.  This prevents the sort of issue above since `<Parthenon/mesh>`, `<Parthenon/Mesh>`, and `<PaRtHeNoN/MeSh>` are all equivalent, but it potentially introduces a new issue if someone relies on the fact that, for example, `important_parameter` and `Important_Parameter` actually set two different fields.  I'd argue nobody should be doing that anyway, so I'd suggest we just transform everything to lowercase and document it.

Anyone have an opinion on this?  @jmstone does this break compatibility with existing Athena input files?

Yurlungur:
I can see a reason for parameter names being case sensitive. I can imagine someone wanting to name a parameter in CamelCase, for example.

However, even if we leave parameter names case sensitive, I think block names should not be.

jdolence:
@Yurlungur there wouldn't be anything stopping someone from using CamelCase.  The trouble would come in if they used CamelCase and camelCase to indicate different parameters because both of these, internal to Parthenon, would become camelcase.  But, maybe I haven't understood your comment fully.

Yurlungur:
@jdolence hmm that's true. Ignore that objection then.

The other case I can see this mattering (although maybe you shouldn't be doing this) is with parameters or physical constants. G vs. g for Newton's constant and the acceleration due to gravity (or the metric) for example.

jdolence:
Oh man...G and g.  Yeah that's not inconceivable, though I agree you shouldn't do that.  Perhaps the thing to do there is to make that a runtime error.  If parthenon caught and reported this as an error with a useful message would that cover this case adequately?

Yurlungur:
Yeah, I like that. Don't quietly accept it, loudly complain.

jmstone:
There would be no compatibility issues with Athena++ input files.  Currently input block names are always all-lowercase (although sometimes users add parameter names that use capitals).  Still I don't think will be a serious problem.

[...]

PR #268, opened 2020-08-18
WIP: Case sensitivity in input parsing

jdolence:
## PR Summary

This PR makes input parsing case insensitive by transforming all block and parameter names to lowercase both when they are initially read in from file and also when accessed via the `Get` methods.

At least partly addresses #264 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


src/parameter_input.cpp
```diff
@@ -109,6 +109,11 @@ InputBlock::~InputBlock() {
   }
 }
 
+void toLower(std::string &name) {
+  std::transform(name.begin(), name.end(), name.begin(),
+                 [](unsigned char c) { return std::tolower(c); });
+}
```
Yurlungur, line 107: I can't believe there's no built-in `std::tolower` for an entire string. But there does not seem to be. That said, I don't think you need the lambda, `std::tolower` already has the right signature.
```suggestion
void toLower(std::string &name) {
  std::transform(name.begin(), name.end(), name.begin(), std::tolower);
}
```

src/parameter_input.cpp
```diff
@@ -199,6 +205,15 @@ void ParameterInput::LoadFromStream(std::istream &is) {
 
     if (!continuing) {
       if (param_name != "") {
+        if (DoesParameterExist(block_name, param_name)) {
+          msg << "### FATAL ERROR in function [ParameterInput::LoadFromStream]"
+              << std::endl
+              << "Block/parameter " << block_name << "/" << param_name
+              << " appears more than once in input" << std::endl
+              << "NOTE: input block and parameter names are NOT case sensitive"
+              << std::endl;
```
Yurlungur, line 206: `PARTHENON_THROW` takes care of the function location and final line break. (It spits out a file and line number.) I think the redundant information can't hurt, so this is fine as is. Just a heads up.

jlippuner:
@jdolence Is this ready to be reviewed and merged?

jdolence:
> @jdolence Is this ready to be reviewed and merged?

well this is a year old now.  I still feel like this is a good idea (basically making input parsing case insensitive).  I think this is basically done, just need to update the docs/changelog.  I'll do that and remove WIP.

Yurlungur:
@jdolence what's the status of this PR?

PR #271, opened 2020-08-19
Fix setting default CXX standard

### LINKED ISSUE #256: Incorrect default C++ standard on first configure
Our current method to pick the C++ 14 standard does (at least with CMake 3.14.2) not work as expected, e.g.,
```bash
$ cmake -DKokkos_ARCH_POWER9=ON -DKokkos_ARCH_VOLTA70=True -DKokkos_ENABLE_CUDA=True -DKokkos_ENABLE_OPENMP=True -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_COMPILER=mpic++ -DMPIEXEC_EXECUTABLE=jsrun ${PARTHENON_ROOT}
[...]
-- Setting default Kokkos CXX standard to 11
-- Setting policy CMP0074 to use <Package>_ROOT variables
-- The project name is: Kokkos
-- Using -std=c++11 for C++11 standard as feature
[...]
```
only if the command is executed again it's set to 14
```bash
[...]
-- The project name is: Kokkos
-- Using -std=c++14 for C++14 standard as feature
[...]
```

pgrete:
## PR Summary

The Kokkos prefixed variable seem to be overwritten/ignored when Kokkos is included.
If we use the CMake prefixed var instead then this is processed properly by the Kokkos CMakeLists.txt

Fix #256 

Before:

On first cmake call:
```
-- Using -std=c++11 for C++11 standard as feature
```
On second (identical) cmake call:
```
-- Using -std=c++14 for C++14 standard as feature
```

After:
On first cmake call:
```
-- Using -std=c++14 for C++14 standard as feature
```

## PR Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Changed (changing behavior/API/variables/...)
 
 ### Fixed (not changing behavior/API/variables/...)
+- [[#256]](https://github.com/lanl/parthenon/issues/256): Fix setting default CXX standard.
 
 ### Removed
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -186,7 +186,7 @@ endif()
 
 # We want Kokkos to be built with C++14, since that's what we're using in
 # Parthenon.
-set(Kokkos_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD 14)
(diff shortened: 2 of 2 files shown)


Yurlungur (approved):
Good catch.

AndrewGaspar (approved):

PR #279, opened 2020-08-22
add metadata documentation for fillghosts

### LINKED ISSUE #272: Fluxes only allocated if `Metadata::FillGhost` is set
@pgrete found an interesting feature. At some point the allocation of fluxes moved behind an `if (Metadata::FillGhosts)` statement. This is potentially incorrect. Fluxes are required for independent variables, even if said variables don't need ghost zones. These two should be decoupled.

Yurlungur:
## PR Summary

Recently @pgrete discovered an interesting, undocumented bit of behaviour. The fluxes for an independent variable are only allocated if `Metadata::FillGhosts` is set. While this isn't necessarily obvious from the documentation, it does make some sense, since independent variables probably need their ghost zones filled and since the flux is used for flux correction between mesh blocks.

While this may merit a redesign in the future, I think that for now the solution is document it and move on. For the reasons outlined above, the design at the moment makes sense for our needs, and we can revisit if that changes.

As such, this PR can be considered to resolve issue #272 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/docs/interface/Metadata.md
+++ b/docs/interface/Metadata.md
@@ -5,6 +5,7 @@
 Variables can be tagged with a variety of `MetadataFlag` values. These flags
 primarily allow an application to tell Parthenon to apply certain behaviors to
 each field.
+
 ### Variable Topology
 
 Topology essentially specifies on which place on the finite volume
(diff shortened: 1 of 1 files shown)


Yurlungur:
@pgrete @AndrewGaspar anyone want to take a quick look? It's a 1 line change to just documentation.

pgrete (approved):
Thanks for clarifying this.
And I agree, this may be updated in the future, but for now this works.

JoshuaSBrown (approved):
Thanks for this.

PR #281, opened 2020-08-24
Joshua s brown/detect nvidia gpus

### LINKED ISSUE #270: Improve regresion testing infrastructure and compatibility with Kokkos devices
It would be idea to be able to copy and paste output from the python tests to rerun them, currently the regression tests output all the arguments passed in but they are all encapsulated in [] brackets. Secondly, there is no way to pass in --kokkos-ndevices= flag which is needed to run the tests with more than a single kokkos device.

JoshuaSBrown:
## PR Summary

This branch attempts to fix a number of annoying issues. Many of these changes have been motivated in order to get the ci running on more than a single cuda device.

1. Auto detection of the number of cuda capable devices using nvidia-smi
2. Warning and error messages if mpi - and cuda capable devices are not lined up
3. Improvement of python test messages
4. Allowing weak scaling of all mpi regression tests.

## PR Checklist

- [x] Code passes cpplint
- [Not really applicable] Adds a test for any bugs fixed. Adds tests for new features.
- [x] Code is formatted


cmake/DetectGPUs.cmake
```diff
@@ -0,0 +1,26 @@
...
+if( NVIDIA_SMI )
+  execute_process( COMMAND ${NVIDIA_SMI} -L OUTPUT_VARIABLE SMI_OUTPUT)
+  set(GPU_ID "0")
+  set(GPU_DETECT "GPU ${GPU_ID}: ")
+  while("${SMI_OUTPUT}" MATCHES "${GPU_DETECT}")
+    math(EXPR NUM_GPUS_DETECTED "${NUM_GPUS_DETECTED} + 1")
+    math(EXPR GPU_ID "${GPU_ID} + 1")
+    set(GPU_DETECT "GPU ${GPU_ID}: ")
+  endwhile()
+endif()
```
Yurlungur: Nice! This is clever.

Is there any danger this parsing could be hijacked maliciously by a fake `nvidia-smi`? And should we care?

cmake/TestSetup.cmake
```diff
@@ -44,44 +44,69 @@ endfunction()
 # Adds test that will run in serial
 # test output will be sent to /tst/regression/outputs/dir
 # test property labels: regression, mpi-no
-function(setup_test dir arg)
+function(setup_test dir arg extra_labels)
```
Yurlungur, line 47: The extra labels functionality is a nice improvement.

cmake/TestSetup.cmake
```diff
@@ -44,44 +44,69 @@ endfunction()
...
+          "\nEach mpi processor must have access to a single gpu. Consider setting:\n"
+          "NUM_GPU_DEVICES_TESTING=${NUM_GPUS_DETECTED}\n"
+          "NUM_MPI_PROC_TESTING=${NUM_GPUS_DETECTED}\n"
+          "\nOr consider building without CUDA")
+      else()
+        set(PARTHENON_KOKKOS_TEST_ARGS "--kokkos-num-devices=${NUM_GPU_DEVICES_TESTING}")
+        list(APPEND labels "cuda")
+      endif()
+    endif()
     add_test( NAME regression_mpi_test:${dir} COMMAND ${Python3_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/run_test.py
```
Yurlungur: :+1:

tst/regression/CMakeLists.txt
```diff
@@ -39,10 +39,11 @@
 
 # Performance regression test
 list(APPEND TEST_DIRS advection_performance)
-list(APPEND TEST_PROCS "1")
+list(APPEND TEST_PROCS ${NUM_MPI_PROC_TESTING})
```
Yurlungur, line 45: Wait... is `NUM_MPI_PROC_TESTING` now automatically set to 1 for the non-mpi tests?

tst/regression/test_suites/advection_convergence/advection_convergence.py
```diff
@@ -66,7 +66,7 @@ def Prepare(self,parameters, step):
             if lin_res[step - 1] == 32:
                 parameters.coverage_status = "both"
             parameters.driver_cmd_line_args = [
-                'parthenon/mesh/nx1=%d' % lin_res[step % n_res -1],
+                'parthenon/mesh/nx1=%d' % (lin_res[step % n_res -1] * parameters.num_ranks),
```
Yurlungur, line 69: This is cool. But do we actually want to do a scaling test and a convergence test in one? This could get ludicrously expensive, depending on what we're doing?

Or am I misunderstanding what's going on here?

tst/regression/test_suites/restart/restart.py
```diff
@@ -38,7 +38,8 @@ def Prepare(self, parameters, step):
             parameters.driver_cmd_line_args = [
                 '-r',
                 'gold.out0.00001.rhdf',
-                'parthenon/job/problem_id=silver'
+                'parthenon/job/problem_id=silver',
+                'parthenon/mesh/nx1=%d' % (64 * parameters.num_ranks)
```
Yurlungur, line 42: One issue I see with setting the number of cells equal to the number of ranks is that for tests that compare to a gold file, the test will fail when run with a different number of cores as the gold file.

We should document somewhere how many cores the gold files were run with so we can reproduce on non-CI machines.

CMakeLists.txt
```diff
@@ -220,6 +231,16 @@ if (${PARTHENON_ENABLE_UNIT_TESTS} OR ${PARTHENON_ENABLE_INTEGRATION_TESTS} OR $
...
+
+
+    # Throw a warning if the number of processors available is less than 4
+    include(ProcessorCount)
+    ProcessorCount(N)
+    if( "${NUM_MPI_PROC_TESTING}" GREATER "${N}") 
+      message(WARNING "Consider changing the number of MPI processors used in testing, currently "
+        "NUM_MPI_PROC_TESTING is set to ${NUM_MPI_PROC_TESTING} but cmake has only detected a"
+        " total of ${N} available processors")
+    endif()
```
Yurlungur, line 285: Nice!

tst/regression/utils/test_case.py
```diff
@@ -104,6 +107,19 @@ def __init__(self,run_test_path,**kwargs):
...
+        try:
+          index = self.parameters.mpi_opts.index("-np")
+          self.parameters.num_ranks = int(self.parameters.mpi_opts[index+1])
+        except ValueError:
+          pass
+        try:
+          index = self.parameters.mpi_opts.index("-n")
+          self.parameters.num_ranks = int(self.parameters.mpi_opts[index+1])
+        except:
+          pass
```
Yurlungur: Using `try`-`except` logic feels a little unsafe to me here. I don't like that the `index` is behind the try along with the typecast. Also I don't love the same code being repeated twice. 

Also this implicitly sets `-n` to override `-np`. We should complain or warn or something if that's happening.

What about the following?
```suggestion
        argstrings = ['-np','-n']
        if len(set(argstrings) & set(self.parameters.mpi_opts)) > 1:
          print('Warning! You have set both "-n" and "-np" in your MPI options.")
        for s in argstrings:
          index = self.parameters.mpi_opts.index(s)
          if index < len(self.parameters.mpi_opts) - 1:
            try:
              self.parameters.num_ranks = int(self.parameters.mpi_opts[index+1])
            except ValueError:
              pass
```
Or am I missing something for why this wouldn't work?

Yurlungur (commented):
This looks great so far! A much needed improvement. A few stream-of-thought comments follow below.

One overarching question I have: is weak scaling something we want to enable within the CI? It's clearly something we should be doing... but we'll quickly overwhelm the CI's resources won't we? And if we *do* want to do weak scaling... what about strong scaling?

tst/regression/CMakeLists.txt
JoshuaSBrown: NUM_MPI_PROC_TESTING is not used for the serial tests.

tst/regression/test_suites/advection_convergence/advection_convergence.py
JoshuaSBrown: Maybe weak scaling is the wrong word choice. The point is to be able to run the tests using more than a single mpi processor. The number of mpi procs for the tests is set at compile time... (for the tests) and we are only running them once. 

Weak scaling would imply that I'm running each tests using several different numbers of mpi procs this is not true... so my choice of weak scaling is ... misleading. In theory you could do this but like you said we would be using a lot of resources. The point is we now have the **flexibility** to run the tests with different numbers of gpus and mpi procs. 

Before this we were limited essentially to running the tests with only a single mpi proc and a single gpu. When I update the documentation I would appreciate your feedback to be sure this is clear.

tst/regression/test_suites/restart/restart.py
JoshuaSBrown: Hmm... this is interesting I have been running the regression tests and have not noticed them failing, when running on darwin with more than one proc/gpu.

cmake/DetectGPUs.cmake
JoshuaSBrown: I can't see it being a problem even if we did detect a different number of gpus. The only thing that would happen is that cmake would complain that the number of GPUs detected is fewer than what has been specified in the tests (If the GPUS_DETECTED is fewer. If the number of GPU's detected is more than what is specified in the TEST variable than it will compile but probably just run slower. However, the number of procs used in the tests is set in the CI and is chosen based on the machine that it runs on, so someone would have to change those variables to actually oversubscribe the machine... which would be very obvious.

tst/regression/CMakeLists.txt
```diff
@@ -90,18 +95,21 @@ include(${PROJECT_SOURCE_DIR}/cmake/TestSetup.cmake)
...
   list(GET TEST_PROCS ${index} TEST_PROC)
   list(GET TEST_ARGS ${index} TEST_ARG)
-  setup_test(${TEST_DIR} ${TEST_ARG})
-  setup_test_coverage(${TEST_DIR} ${TEST_ARG})
-  setup_test_mpi(${TEST_PROC} ${TEST_DIR} ${TEST_ARG})
-  setup_test_mpi_coverage(${TEST_PROC} ${TEST_DIR} ${TEST_ARG})
+  list(GET EXTRA_TEST_LABELS ${index} TEST_LABELS )
+  message(STATUS "Label ${TEST_LABELS}")
+  setup_test(${TEST_DIR} ${TEST_ARG} "${TEST_LABELS}")
+  setup_test_coverage(${TEST_DIR} ${TEST_ARG} "${TEST_LABELS}")
```
JoshuaSBrown, line 107: @Yurlungur See here `setup_test` is setting up the serial tests, TEST_PROC is not passed to this function. Maybe I should relabel to `setup_serial_test`.

JoshuaSBrown:
@pgrete how many gpu's are available on the IAS machine?

cmake/DetectGPUs.cmake
Yurlungur: Cool. Yeah I wouldn't expect it to be an issue. Just thinking "out loud."

tst/regression/test_suites/advection_convergence/advection_convergence.py
Yurlungur: Ah, ok. I understand. I agree, this flexibility is desirable. Yeah weak scaling means something specific in my brain, so I got confused.

> When I update the documentation I would appreciate your feedback to be sure this is clear.

Sounds good.

tst/regression/test_suites/restart/restart.py
Yurlungur: Huh. Maybe I'm misunderstanding how the gold comparisons work... but I have trouble understanding how that can pass. If the gold file has a different number of cells than the test output file, you *should* get a failure.

Yurlungur: If we're not getting a failure in that case, I'd argue the regression testing infrastructure has a bug.

tst/regression/CMakeLists.txt
Yurlungur: Nah, I think it's fine. I just read the diffs, not the whole files, so I got confused. You don't need to change anything here.

tst/regression/test_suites/restart/restart.py
JoshuaSBrown: Well let me check before we investigate, I might not have run them all the way through.

pgrete:
There's are two GPUs on that node at IAS, if I remember correctly.

A more general comment I have is that I'm not sure how I feel about some parts of this PR in line with the points I raised in #247 and hope we have time to discuss tomorrow during the call.

While I agree that having more flexibility in terms of "passing" arguments to the test, I still think that "machine configuration" files are preferable over Cmake automation (e.g., for detecting the GPUs).
I see this (similar to the specific `-gpu` flag in #247) as adding more logic that will be hard to maintain (and hard to adapt within CMake if a machine varies only a tiny, but significant bit).

Yurlungur:
> There's are two GPUs on that node at IAS, if I remember correctly.
> 
> A more general comment I have is that I'm not sure how I feel about some parts of this PR in line with the points I raised in #247 and hope we have time to discuss tomorrow during the call.
> 
> While I agree that having more flexibility in terms of "passing" arguments to the test, I still think that "machine configuration" files are preferable over Cmake automation (e.g., for detecting the GPUs).
> I see this (similar to the specific `-gpu` flag in #247) as adding more logic that will be hard to maintain (and hard to adapt within CMake if a machine varies only a tiny, but significant bit).

This is an excellent point. As far as the automatic detection goes, perhaps the thing to do is just throw warnings? I.e., the build system should let the user know "you might be doing something wrong but I will trust you."

JoshuaSBrown:
Yes, we should probably discuss this.

> While I agree that having more flexibility in terms of "passing" arguments to the test, I still think that "machine configuration" files are preferable over Cmake automation (e.g., for detecting the GPUs).

I think to clarify for this pull request, there is auto detection of the number of GPUs available, that detection does not however automatically change any variables. An error is simply thrown if either of the below criteria are not met. 

1. if the mpi processors is not equal to the number of gpus specified for the test
2. if the number of devices specified for the test is more than what is actually detected

Auto detection in this case does not mean the number of gpus used in the tests are automatically changed. It is still up to the user to make sure they are setting the variables to reasonable numbers.

The second criteria should probably just be a warning and not throw an error because you could be building for instance on a login node and then run it on a compute node with more devices. I think for the first criteria it is fine to throw an error because running the tests with a different number of gpu devices from the number of mpi processors is incorrect, as kokkos does not yet support this behavior (at least that is my understanding).

It is also safe to call nvidia-smi to determine the number of devices because if you are building with cuda that binary will be available.

tst/regression/utils/test_case.py
JoshuaSBrown: Yes, this would be an improvement.

JoshuaSBrown:
@pgrete I was not quite sure after our discussion, what the consensus was from our meeting related to this pull request. Do I need to throw out the gpu auto detection part?

Yurlungur:
> @pgrete I was not quite sure after our discussion, what the consensus was from our meeting related to this pull request. Do I need to throw out the gpu auto detection part?

Not @pgrete who should also chime in. But I would say make it a warning not an error.

pgrete:
My thinking here is even a warning is potentially misleading/may point towards a unintended use-case.
The latter specifically applies to my expectation that the vast majority of application cases will be executed on compute clusters where the nodes used to build Parthenon (or downstream apps) are different from the nodes that run Parthenon (plus scheduler nodes in between).
For example, I expect to see this warning on literally every GPU machine I have access to (and use for development) and only not see this warning when I do development on my laptop (which does not have a Cuda GPU).

From my point of view the intention of this PR is to make sure that it is straightforward for user/developer to run all test with the intended resources.
I think we can achieve this is a more transparent way using the machine config files (I'll get working on them momentarily).
Thus, I suggest to wait for the machine config PR and everyone gives it a chance to see whether it works for them.
If not, or, if the intention of this PR is different, or, if people have strong feelings about his, I'm happy to go ahead with adding a warning.

Yurlungur:
@JoshuaSBrown  in light of the discussion going on any machine configuration, maybe we could split this PR into two: one focusing on the more flexible python infrastructure for enabling different sized resources to run the tests, and one on the GPU stuff? As these seem to me to be separate issues.

JoshuaSBrown:
> @JoshuaSBrown in light of the discussion going on any machine configuration, maybe we could split this PR into two: one focusing on the more flexible python infrastructure for enabling different sized resources to run the tests, and one on the GPU stuff? As these seem to me to be separate issues.

That sounds like a good idea.

JoshuaSBrown:
> From my point of view the intention of this PR is to make sure that it is straightforward for user/developer to run all test with the intended resources.

What if we moved the logic of checking the number of gpus to the python test code? Or when the tests are actually run using CTEST_CUSTOM_PRE_TEST? This avoids the problem of throwing warnings on a system you don't intend to run the tests on.

pgrete:
> > From my point of view the intention of this PR is to make sure that it is straightforward for user/developer to run all test with the intended resources.
> 
> What if we moved the logic of checking the number of gpus to the python test code? Or when the tests are actually run using CTEST_CUSTOM_PRE_TEST? This avoids the problem of throwing warnings on a system you don't intend to run the tests on.

If there's a good way to solve this I think this'd be nice to have.
However, I'm actually not sure there a universal way to address this. For example, if you run MPI tests (both on a single node with multiple GPUs or across multiple nodes [with potentially multiple GPUs each] ) the results of a Cuda device query depends on the GPUs visible to a process (which is often controlled by by mpirun/srun/hpcind/...).
I'm not trying to be discouraging here, but I was trying sth similar (checking for Cuda aware MPI in K-Athena at runtime for a more meaningful error message), but ended up removing the code I originally added because some MPI implementation didn't implement the required function even though they stated "compatibility" with a version/standard that included it.

JoshuaSBrown:
> If there's a good way to solve this I think this'd be nice to have.
> However, I'm actually not sure there a universal way to address this. For example, if you run MPI tests (both on a single node with multiple GPUs or across multiple nodes [with potentially multiple GPUs each] ) the results of a Cuda device query depends on the GPUs visible to a process (which is often controlled by by mpirun/srun/hpcind/...).

I think you can get around this though if you do it from the cmake side, because all those settings will be available to cmake. You would just check that they are correct when you run the tests when calling`ctest` instead of during the configuration part when you call `cmake`. 

> I'm not trying to be discouraging here, but I was trying sth similar (checking for Cuda aware MPI in K-Athena at runtime for a more meaningful error message), but ended up removing the code I originally added because some MPI implementation didn't implement the required function even though they stated "compatibility" with a version/standard that included it.

Can you send me the code so I can take a look at what you already tried?

JoshuaSBrown:
@pgrete So I really like this solution now. It will check/detect the number of gpu's only when ctest is called right before running the tests. This avoid the problems of compiling on different node and running tests on another.

docs/building.md
```diff
@@ -6,31 +6,32 @@ If you come across a disfunctional setup, please report it by open an issue or p
...
+   | REGRESSION\_GOLD\_STANDARD\_HASH | SHA512=...             | String | Hash value of gold standard file to be downloaded. Used to ensure that the download is not corrupted. |
+   | REGRESSION\_GOLD\_STANDARD\_SYNC | ON                     | Option | Create `gold_standard` target to download gold standard files |
+   |          ENABLE\_UNIT\_TESTS | ${BUILD\_TESTING}          | Option | Enable unit tests |
+   |               CODE\_COVERAGE | OFF                        | Option | Builds with code coverage flags |
+   |       CMAKE\_INSTALL\_PREFIX | machine specific           | String | Optional path for library installation |
+   |                 Kokkos\_ROOT | unset                      | String | Path to a Kokkos source directory (containing CMakeLists.txt) |
+   |  PARTHENON\_IMPORT\_KOKKOS | ON/OFF                       | Option | If ON, attempt to link to an external Kokkos library. If OFF, build Kokkos from source and package with Parthenon |
+   |          BUILD\_SHARED\_LIBS | OFF                        | Option | If installing Parthenon, whether to build as shared rather than static |
```
Yurlungur: cleaning up the markdown is a much needed improvement. Thanks for taking the time.

tst/regression/test_suites/restart/restart.py
JoshuaSBrown: Yes, this was failing. I guess we will need a different gold standard file if we want to test against more than one gpu.

JoshuaSBrown:
@pgrete This is ready for review. @AndrewGaspar I still need a CS person to look over this.

CHANGELOG.md
```diff
@@ -13,9 +13,8 @@
 - [[PR 293]](https://github.com/lanl/parthenon/pull/293) Changed `VariablePack` and related objects to use `ParArray1D` objects instead of `ParArrayND` objects under the hood to reduce the size of the captured objects.
 - [[PR 310]](https://github.com/lanl/parthenon/pull/310) Fix Cuda 11 builds.
 - [[PR 313]](https://github.com/lanl/parthenon/pull/313) Add include guards for Kokkos in cmake.
-
-### Removed (removing behavior/API/varaibles/...)
-
+- [[PR 281]](https://github.com/lanl/parthenon/pull/281) Allows one to run regression tests with more than one cuda device, Also improves readability of regression tests output.
+>>>>>>> develop
```
AndrewGaspar: Remove this.
```suggestion
```

docs/building.md
```diff
@@ -6,31 +6,32 @@ If you come across a disfunctional setup, please report it by open an issue or p
...
+   |        TEST\_ERROR\_CHECKING | OFF                        | Option | Enables the error checking unit test. This test will FAIL |
+   |    TEST\_INTEL\_OPTIMIZATION | OFF                        | Option | Test intel optimization and vectorization |
+   |    CHECK\_REGISTRY\_PRESSURE | OFF                        | Option | Check the registry pressure for Kokkos CUDA kernels |
+   |               BUILD\_TESTING | ON                         | Option | Multi-testing enablement |
+   | PARTHENON\_DISABLE\_EXAMPLES | OFF                        | Option | Toggle building of examples, if regression tests are on, drivers needed by the tests will still be built |
+   |   ENABLE\_INTEGRATION\_TESTS | ${BUILD\_TESTING}          | Option | Enable integration tests |
+   |    ENABLE\_REGRESSION\_TESTS | ${BUILD\_TESTING}          | Option | Enable regression tests |
+   |      NUM\_MPI\_PROC\_TESTING | 4                          | String | Number of MPI ranks used for MPI-enabled regression tests |
+   |   NUM\_GPU\_DEVICES\_TESTING | ${NUM\_MPI\_PROC\_TESTING} | String | Number of GPUs to use if buildt with kokkos_ENABLE_CUDA |
```
AndrewGaspar:
```suggestion
   |   NUM\_GPU\_DEVICES\_TESTING | ${NUM\_MPI\_PROC\_TESTING} | String | Number of GPUs to use if built with Kokkos_ENABLE_CUDA |
```

CONTRIBUTING.md
```diff
@@ -247,20 +247,29 @@ Essentially, the command for running `run_test.py` must be called from within cm
...
+list(APPEND TEST_PROCS ${NUM_MPI_PROC_TESTING})
 list(APPEND TEST_ARGS "--driver ${CMAKE_BINARY_DIR}/example/calculate_pi/parthenon-example --driver_input ${CMAKE_CURRENT_SOURCE_DIR}/test_suites/calculate_pi/parthinput.regression")
+list(APPEND EXTRA_TEST_LABELS "perf")
 ```
 
-NOTE: By default all regression tests added to these lists will be run in serial and in parallel
-with mpi. The number of mpi processors used is by default set to 4. This default can be adjusted
-by changing the cmake variable NUM\_MPI\_PROC\_TESTING. The number of OpenMP threads is by default set
-to 1 but can be adjusted in the driver input file deck. 
+NOTE: The TEST_PROCS list indicates how many processors to use when running
```
AndrewGaspar:
```suggestion
NOTE: The TEST\_PROCS list indicates how many processors to use when running
```

CONTRIBUTING.md
```diff
@@ -247,20 +247,29 @@ Essentially, the command for running `run_test.py` must be called from within cm
...
+mpi. The cmake variable NUM_MPI_PROC_TESTING can be used if you do not want to
```
AndrewGaspar:
```suggestion
mpi. The cmake variable NUM\_MPI\_PROC\_TESTING can be used if you do not want to
```

.gitlab-ci.yml
```diff
@@ -41,6 +41,7 @@ parthenon-cuda-unit:
     - cd build-cuda-debug
     - cmake -DCMAKE_BUILD_TYPE=Debug
       -DMACHINE_VARIANT=cuda-mpi
+      -DNUM_MPI_PROC_TESTING=1
```
AndrewGaspar: Are these needed given the changes to the CI machine file?

scripts/device_check.sh
```diff
@@ -0,0 +1,62 @@
...
+# This script is designed to detect the number of gpu devices that are available before the
+# ctest suite is run. It does this by using the nvidia-smi. The script is only called if
+# parthenon is built with kokkos_ENABLE_CUDA.
+#
+# The script takes a single argument the number of GPUs that are meant to be used with the 
+# tests. The script then checks that the number of GPUs that are actually avaliable on the
+# system are enough to satisfy this requirement. 
+
+if [ "$#" -ne 1 ]; then
+  printf "You must enter exactly 1 command line arguments, which should indicate
```
AndrewGaspar:
```suggestion
  printf "You must enter exactly 1 command line argument, which should indicate
```

scripts/device_check.sh
```diff
@@ -0,0 +1,62 @@
...
+Number of CUDA devices available: ${GPUS_DETECTED}\n
+The number of gpus detected is less than then the number of devices requested,
+consider changing:\n
+NUM_GPU_DEVICES_TESTING=${GPUS_DETECTED}
+NUM_MPI_PROC_TESTING=${GPUS_DETECTED}\n
+Or consider building without CUDA.\n"
+  exit 1
+elif [ "$NUM_GPUS_ASSIGNED_FOR_TEST" -lt "$GPUS_DETECTED" ]
+then
+  printf "\nNumber of GPUs detected: $GPUS_DETECTED
```
AndrewGaspar: Hm... this warning seems kind of annoying and unnecessary - I would expect running tests with 1 MPI rank on a Sierra or Summit-like node to be common. Either get rid of this only output this warning if `NUM_GPUS_ASSIGNED_FOR_TEST` does not equal `NUM_MPI_PROC_TESTING`. If this has already been hashed out, feel free to ignore.

AndrewGaspar (changes requested):

.gitlab-ci.yml
JoshuaSBrown: Hmm, probably a better idea to but this as a set command in the machine cfg file, is that what you were thinking? 

Ok, I see what happened, seems like something was not getting overwritten correctly.

scripts/device_check.sh
JoshuaSBrown: This is not a warning, and does not appear during the config step. It just helps the user to know what values are actually being used when the tests are being run, this is not triggered by cmake, the script is called from within the ctest framework. I think it is worth keeping as cmake and ctest are not guaranteed to have the same values.

JoshuaSBrown: Disregard first comment, I only saw the top line that was highlighted.

JoshuaSBrown: Ok, fair point.

cmake/machinecfg/CI.cmake
```diff
@@ -20,7 +20,7 @@ message(STATUS "Loading machine configuration for default CI machine. "
 
 # common options
 set(Kokkos_ARCH_WSM ON CACHE BOOL "CPU architecture")
-set(NUM_MPI_PROC_TESTING ON CACHE STRING "1")
+set(NUM_MPI_PROC_TESTING "1")
```
AndrewGaspar: Make this change and then you shouldn't need to do all the checking above about checking the cache. You just had the arguments to set messed up before - sorry I didn't notice it.

```suggestion
set(NUM_MPI_PROC_TESTING "1" CACHE STRING "CI runs tests with 1 MPI rank")
```

CMakeLists.txt
```diff
@@ -134,7 +134,14 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
...
 set(ENABLE_MPI OFF)
-set(NUM_MPI_PROC_TESTING "4" CACHE STRING "Number of mpi processors to use when running tests with MPI")
+if( NOT NUM_MPI_PROC_TESTING )
+  set(NUM_MPI_PROC_TESTING "4" CACHE STRING "Number of mpi processors to use when running tests with MPI")
+else()
+  get_property(EXISTS_IN_CACHE CACHE NUM_MPI_PROC_TESTING PROPERTY TYPE)
+  if( NOT EXISTS_IN_CACHE )
+    set(NUM_MPI_PROC_TESTING "${NUM_MPI_PROC_TESTING}" CACHE STRING "Number of mpi processors to use when running tests with MPI")
+  endif()
+endif()
```
AndrewGaspar: Revert this, I think - you had just previously made a mistake in setting the cache variable in CI.cmake.

```suggestion
set(NUM_MPI_PROC_TESTING "4" CACHE STRING "Number of mpi processors to use when running tests with MPI")
```

AndrewGaspar (changes requested):

pgrete:
I think testing the capabilities at the time of testing is definitely preferable over at compile time.
Regarding the current (hard) failure modes, I have two questions (:
- Is running multiple MPI processes on a single GPU still supported (using MPS)?
- Is running MPI test across hosts still supported (i.e., there are less GPUs available as reported by `nvidia-smi`  than in total for all MPI processes)

Finally, it'd be great if there's a configure switch that disables all these tests.
I'm still concerned that this approach is not maintainable (with Intel and AMD GPUs showing up on multiple different machines next year), which is why I expect to not use this feature and just go with machine files.
(Latter is my personal expectation/approach and should not prevent this feature being merged in if other people think they'll use and maintain it)

JoshuaSBrown:
>     * Is running multiple MPI processes on a single GPU still supported (using MPS)?

And that is why we have code review. I will need to fix this.

JoshuaSBrown:
@pgrete I think I have addressed your concerns, I have made the checks simply output info to the screen as opposed to failing. You can also turn the script off. I have also addressed the test cases you mentioned with running with multiple nodes and with having more mpi procs than 1 per gpu device.

JoshuaSBrown:
Now when ctest is called the following header is output so one can know what resources are actually being utilized. 

```

****************************************** Beginning Precheck ******************************************

Number of GPUs detected per node: 2
Number of GPUs per node, requested in tests: 2

Number of nodes detected: 1
Number of MPI procs, requested in tests: 8
Number of MPI procs per node: 8

******************************************  Ending Precheck   ******************************************


```

This is output from a power9 node.

JoshuaSBrown:
ping @pgrete @AndrewGaspar @Yurlungur, this has been open for more than 24 hours.

AndrewGaspar (approved):

CMakeLists.txt
```diff
@@ -62,6 +62,7 @@ built some, drivers needed by the regression tests will still be built."
   OFF
 )
 
+option(PARTHENON_ENABLE_GPU_MPI_CHECKS "Checks if possible that the mpi num of procs and the number of gpu devices detected are appropriate." ${BUILD_TESTING})
```
Yurlungur: Can we put a line break here? The comment is long enough it's going to word wrap in text editors.

CMakeLists.txt
```diff
@@ -196,10 +197,15 @@ set(CMAKE_CXX_EXTENSIONS OFF)
 SET (Kokkos_ENABLE_AGGRESSIVE_VECTORIZATION ON CACHE BOOL
      "Kokkos aggressive vectorization")
 
-# Tell Kokkos we need lambdas in Cuda.
+# Check that gpu devices are actually detected
+set(NUM_GPU_DEVICES_PER_NODE "1" CACHE STRING "Number of gpu devices to use when testing if built with Kokkos_ENABLE_CUDA")
```
Yurlungur, line 202: Might want to declare this publicly as an option, as we may end up wanting to run tests on multiple GPUs.

docs/building.md
```diff
@@ -6,31 +6,33 @@ If you come across a disfunctional setup, please report it by open an issue or p
...
+   |    REGRESSION\_GOLD\_STANDARD\_HASH | SHA512=...        | String | Hash value of gold standard file to be downloaded. Used to ensure that the download is not corrupted. |
+   |    REGRESSION\_GOLD\_STANDARD\_SYNC | ON                | Option | Create `gold_standard` target to download gold standard files |
+   |                 ENABLE\_UNIT\_TESTS | ${BUILD\_TESTING} | Option | Enable unit tests |
+   |                      CODE\_COVERAGE | OFF               | Option | Builds with code coverage flags |
+   |              CMAKE\_INSTALL\_PREFIX | machine specific  | String | Optional path for library installation |
+   |                        Kokkos\_ROOT | unset             | String | Path to a Kokkos source directory (containing CMakeLists.txt) |
+   |           PARTHENON\_IMPORT\_KOKKOS | ON/OFF            | Option | If ON, attempt to link to an external Kokkos library. If OFF, build Kokkos from source and package with Parthenon |
+   |                 BUILD\_SHARED\_LIBS | OFF               | Option | If installing Parthenon, whether to build as shared rather than static |
```
Yurlungur, line 35: cleaning up this table is a welcome change!

Yurlungur (approved):
This is a nice solution. I have a few nitpicks, but I consider them non-blocking.

CMakeLists.txt
JoshuaSBrown: So the option command is only really appropriate for bool values. Is this what you are suggesting. By setting a value with CACHE it is publicly accessible. And that variable is documented in the cmake config table for parthenon. Is this what you meant?

PR #283, opened 2020-08-25
Remove performance and regression from commit-wise tests

### LINKED ISSUE #278: An issue with the CI?
I've noticed the CI seems to be failing kind of regularly and then passes upon restart. I think this might be coming from the `overlapping space instances` test from `test_kokkos_abstraction`. Do we know what's going on there? Is something going on with the CI server again?

@pgrete @JoshuaSBrown @forrestglines

Yurlungur:
## PR Summary

Test MR for changing performance and regression tests to only run on nightly builds. Hopefully this will significantly speed up CI, at not too much cost to stability, since the full test suite is run nightly. Partially addresses #278 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -73,6 +73,8 @@ parthenon-cpu-unit:
       - build-debug/CMakeFiles/CMakeOutput.log
 
 parthenon-cuda-performance_and_regression:
+  only:
+    - schedules
   tags:
     - cuda
   stage: performance_and_regression
@@ -96,6 +98,8 @@ parthenon-cuda-performance_and_regression:
       - build-cuda-perf/tst/regression/outputs/advection_convergence/advection-errors.png
 
 parthenon-cuda-performance_and_regression-mpi:
+  only:
+    - schedules
   tags:
     - cuda
   stage: performance_and_regression
@@ -122,6 +126,8 @@ parthenon-cuda-performance_and_regression-mpi:
 
 # run performance and regression on CPUs without MPI
 parthenon-cpu-performance_and_regression:
+  only:
+    - schedules
   tags:
     - cpu
   stage: performance_and_regression
@@ -144,6 +150,8 @@ parthenon-cpu-performance_and_regression:
 
 # run performance and regression on CPUs with MPI
 parthenon-cpu-performance_and_regression-mpi:
+  only:
+    - schedules
   tags:
     - cpu
   stage: performance_and_regression
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -19,7 +19,7 @@ variables:
   GIT_SUBMODULE_STRATEGY: recursive
 
 stages:
-  - unit
+  - short
   - performance_and_regression
   - coverage
   - style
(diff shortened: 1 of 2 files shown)


Yurlungur:
This is now ready for review. I would appreciate people's feedback on this, as philosophically, it's a fairly substantial change, although it's only a 4 line diff.

This is motivated by the fact that the CI currently takes an extremely long time. This PR doesn't completely fix the issue, but it helps. As things currently stand, each test takes It reduces test times from **72 minutes**. The coverage and regression stages each take **approximately half an hour**. That's per commit. This is why the CI is overscribed and why each CI run is often in the queue for **half an hour or more**. This is making development increasingly onerous, especially if I have several PR's in the pipeline at once.

I reduce this to about 10 minutes by moving the regression and coverage tests to nightly instead of per-commit. My recommendation is that for big PR's we trigger the pipeline manually. That's kind of onerous, but at least the CI will be usable.

Pinging @pgrete @AndrewGaspar @jdolence @nmsriram @forrestglines @jmstone @JoshuaSBrown and all other interested parties.

We might need to change the github workflow requirements for merges with the coverage test removed.

pgrete:
I think this goes in the right direction.
One thing I'd update is to add a brief test (say <1 min total) that covers a good chunk of the code, e.g., `32^3` AMR with one level for 100 cycles with writing/restarting and comparing to hdf5.
I happy to add that test to this PR if there are no objections.

Yurlungur:
> I happy to add that test to this PR if there are no objections.

Please do, @pgrete ! So long as it's fast, I think that would be an improvement.

JoshuaSBrown:
I don't see a problem with this, like you said in it's current state it simply takes too long. Even with more resources your hitting a bottleneck because of the advection convergence tests. I think @nmsriram made some comments about using analytical solutions opposed to numerical convergence, that would probably help with this issue. Though I think there is value in having both numerical and analytical implementations any test that takes over 1 minute to run is probably adding to much overhead to run on a commit basis. 

I'm wondering if there is a way to run these longer tests even if it is on a pull request, if the pull request is left over night. That way if there is an error it gets caught before it gets merged but it won't slow down merging there are limited resources during the day.

Yurlungur:
As discussed on the call, merges into develop should now also trigger the full tests.

Yurlungur:
I think the last thing that this needs then, before it's ready for review is the promised fast integration test by @pgrete

pgrete:
How does this look like? I consider this a temporary solution as I'd like to merge the "output hdf5" test with the "restart test" (as the latter also requires hdf5) and create a test that runs shorter and covers
- start sim (with MPI and nranks > 1) and write a restart file after N1 cycles
- restart sim (potentially with a different number of ranks), continue for N2 cycles, write hdf5 analysis output
- compare hdf5 analysis output versus a gold standard file
That would cover quite a bit of code that we have and should allow to quickly identify problems.

This temporary solution on a volta takes about 2 min. Let's see what it looks like on the CI machine.

pgrete:
Looks like the total runtime is now 15 min. Do you think this is okay or still too long?

Yurlungur:
> Looks like the total runtime is now 15 min. Do you think this is okay or still too long?

I think that's fine. It's a factor of 5 improvement over the current times.

> How does this look like?

Sounds good! Agreed that the tests should be merged at some point.

Yurlungur:
I think this is ready for review then

pgrete (approved):

AndrewGaspar (approved):

PR #284, opened 2020-08-26
[WIP] Joshua s brown/add separate public headers

### LINKED ISSUE #239: Establish Distinction between public and private headers

### LINKED ISSUE #273: Test correct build/export of parthenon as a library
Following @junghans comments, the examples can be used as a test. This is in reference to [238](https://github.com/lanl/parthenon/pull/238). This is to ensure that parthenon can be used as a third party library and that it's dependencies are correctly configured. 

Add an `examples` folder with the pi example and a `CMakeLists.txt` like:
```cmake
cmake_minimum_required(VERSION 3.11)
project(parthenon_linking_example)
if(NOT TARGET Parthenon::parthenon)
  find_package(parthenon REQUIRED)
endif()
add_executable(
  pi-example
  pi_driver.cpp
  pi_driver.hpp
  calculate_pi.cpp
  calculate_pi.hpp
  )
target_link_libraries(pi-example PRIVATE Parthenon::parthenon)
```
This `CMakeLists.txt` will work for internal as well as external builds, so you can build that example as part of parthenon, but also standalone. The latter you can add to the CI, so that the exports are always tested.

JoshuaSBrown:
## PR Summary

Separates the public headers from the source code, this is the first step to defining a public api. All the headers have been made point as a starting point. This pull request will also add a test to insure a library install works. 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


CMakeLists.txt
```diff
@@ -297,12 +297,16 @@ if (TEST_INTEL_OPTIMIZATION)
 endif()
 
 add_subdirectory(src)
-add_subdirectory(example)
+add_subdirectory(include/parthenon)
```
AndrewGaspar: Why the additional subdirectory - why not just put the files directly under include?

include/parthenon/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
...
+# for the U.S. Department of Energy/National Nuclear Security Administration. All rights
+# in the program are reserved by Triad National Security, LLC, and the U.S. Department
+# of Energy/National Nuclear Security Administration. The Government is granted for
+# itself and others acting on its behalf a nonexclusive, paid-up, irrevocable worldwide
+# license in this material to reproduce, prepare derivative works, distribute copies to
+# the public, perform publicly and display publicly, and to permit others to do so.
+#=========================================================================================
+
+# The header files are added to parthenon for linting
+list(APPEND PARTHENON_PUBLIC_HEADERS
```
AndrewGaspar: Hm... So what I typically like to do in a cmake project is:

```
CMakeLists.txt
src/
    *.cpp
include/
    *.hpp
```

And then the `add_library` goes into the CMakeLists.txt that sits next to `src`. It would resolve having to do this thing you're doing here with collecting the header files in a separate CMakeLists.txt.

So, should we perhaps just eliminate `src/CMakeLists.txt` and move its contents into the parent directory?

AndrewGaspar: Or maybe add a `parthenon` subdirectory and then move both src and include into that directory if you want to keep it cleanly separated from the other files in the root directory.

JoshuaSBrown: I think it would be a good idea to keep the current directory structure because it follows convention. However, I'm up for suggestions. I think the only reason a list of the header files are needed is because of the way the linter is set, that might be the best spot to look into changing things.

CMakeLists.txt
JoshuaSBrown: This follows my understanding of convention, but I can change it if needed. https://cliutils.gitlab.io/modern-cmake/chapters/basics/structure.html, https://hsf-training.github.io/hsf-training-cmake-webpage/06-projectstructure/index.html.

Yurlungur:
Since the lead author is on leave, and there are merge conflicts, I've set this to WIP for now. 

I don't love the `src/` vs. `include/` separation. But I'm willing to live with it if it's necessary.

AndrewGaspar:
> I don't love the src/ vs. include/ separation. But I'm willing to live with it if it's necessary.

It's not necessary per-se, but it is intended to make the public API more deliberate. Right now we just kind of export everything, which makes it more likely that you'll break downstream apps, or that you'll need to support APIs that you didn't intend to.

Yurlungur:
> It's not necessary per-se, but it is intended to make the public API more deliberate.

Oh wait. I might have misunderstood. Do private headers still live in `src`? Or do all headers move to `include`?

JoshuaSBrown:
> > It's not necessary per-se, but it is intended to make the public API more deliberate.
> 
> Oh wait. I might have misunderstood. Do private headers still live in `src`? Or do all headers move to `include`?

In this pull request everything has been moved to include, so everything has been made public. In the future, private headers should be left in the src dir.

JoshuaSBrown:
This is ready for review.

CHANGELOG.md
```diff
@@ -3,11 +3,17 @@
 ## Current develop
 
 ### Added (new features/APIs/variables/...)
+- [[PR 250]](https://github.com/lanl/parthenon/pull/250) Feature::Restart. If output file format 'rst' is specified restart files are written using independent variables and those marked with Restart metadata flag.  Simulations can be restarted with a '-r \<restartFile\>' argument to the code.
+- [[PR 263]](https://github.com/lanl/parthenon/pull/263) Added MeshPack, a mechanism for looping over the whole mesh at once within a `Kokkos` kernel. See [documentation](docs/mesh/packing.md)
+- [[PR 267]](https://github.com/lanl/parthenon/pull/267) Introduced TaskRegions and TaskCollections to allow for task launches on multiple blocks.
+- [[PR 287]]((https://github.com/lanl/parthenon/pull/287) Added machine configuration file for compile options, see [documentation](https://github.com/lanl/parthenon/blob/develop/docs/building.md#default-machine-configurations)
```
AndrewGaspar: You accidentally duplicated these changes.

```suggestion
```

CMakeLists.txt
AndrewGaspar: Sorry, I didn't see this response - you're right, that all of the headers *should* go under a `parthenon` directory. In fact, that's the path they should be included from in the app. That's why when I added the explicitly public headers (e.g. `parthenon/prelude.hpp`, etc.), I put them in a `parthenon` directory.

Unfortunately, most of our includes don't currently work like that - they are expected to live at the "root" directory. Hopefully, we can over time move all of our header files into the `parthenon` directory over time.

My opinion is that the `include` directory should match the structure that the user is expected to include them, even if that structure is less than desirable.

AndrewGaspar (changes requested):
Looks good, but I'm going to ask you to move up the header files a directory.

CHANGELOG.md
```diff
@@ -3,11 +3,7 @@
 ## Current develop
 
 ### Added (new features/APIs/variables/...)
-- [[PR 250]](https://github.com/lanl/parthenon/pull/250) Feature::Restart. If output file format 'rst' is specified restart files are written using independent variables and those marked with Restart metadata flag.  Simulations can be restarted with a '-r \<restartFile\>' argument to the code.
-- [[PR 263]](https://github.com/lanl/parthenon/pull/263) Added MeshPack, a mechanism for looping over the whole mesh at once within a `Kokkos` kernel. See [documentation](docs/mesh/packing.md)
-- [[PR 267]](https://github.com/lanl/parthenon/pull/267) Introduced TaskRegions and TaskCollections to allow for task launches on multiple blocks.
-- [[PR 287]]((https://github.com/lanl/parthenon/pull/287) Added machine configuration file for compile options, see [documentation](https://github.com/lanl/parthenon/blob/develop/docs/building.md#default-machine-configurations)
-- [[PR 285]](https://github.com/lanl/parthenon/pull/285) Parthenon can now be linked in CMake as `Parthenon::parthenon` when used as a subdirectory, matching install.
```
AndrewGaspar: Keep this:
```
- [[PR 285]](https://github.com/lanl/parthenon/pull/285) Parthenon can now be linked in CMake as `Parthenon::parthenon` when used as a subdirectory, matching install.
```

CMakeLists.txt
JoshuaSBrown: Ok, I see, that makes this a bit awkward.

JoshuaSBrown:
> My opinion is that the include directory should match the structure that the user is expected to include them, even if that structure is less than desirable.

This is going to make a mess of the include folder where they are installed.

src/CMakeLists.txt
```diff
@@ -252,31 +204,12 @@ endif()
...
 target_link_libraries(parthenon PUBLIC Kokkos::kokkos)
 
-lint_target(parthenon)
-
 target_include_directories(parthenon PUBLIC
-  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
-  $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/generated>
+  $<BUILD_INTERFACE:${parthenon_SOURCE_DIR}/include>
+  $<BUILD_INTERFACE:${parthenon_BINARY_DIR}/include/parthenon/generated>
   $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}/parthenon>
```
AndrewGaspar: I think something needs to be fixed here - either the install still needs to put the headers into a `parthenon` directory, or this just needs to be `${CMAKE_INSTALL_INCLUDEDIR}`.

AndrewGaspar:
Yeah, as far as the directory the files are installed to, I'm not partial... on the one hand, putting them into a subdirectory breaks things like `CPATH` or implicit system includes. On the other hand, they should be using CMake to import the library anyway, so maybe that's OK.

src/CMakeLists.txt
JoshuaSBrown: Good catch.

JoshuaSBrown:
@jmstone @pgrete @AndrewGaspar @Yurlungur @jdolence I still need approvals and feedback on this. Any takers?

AndrewGaspar (approved):

JoshuaSBrown:
Ping @Yurlungur @pgrete

docs/building.md
```diff
@@ -34,11 +34,28 @@ If you come across a disfunctional setup, please report it by open an issue or p
...
+
+And then you can use parthenon in your CMake project by adding it as a
+subdirectory:
+```cmake
+add_subdirectory(path/to/parthenon)
+
+add_executable(myapp ...)
+target_link_libraries(myapp PRIVATE Parthenon::parthenon)
+```
+
```
Yurlungur, line 53: Nice to have this additional documentation

.gitlab-ci.yml
```diff
@@ -157,6 +157,32 @@ parthenon-cuda-performance_and_regression-mpi:
       - build-cuda-perf-mpi/tst/regression/outputs/advection_convergence_mpi/advection-errors.dat
       - build-cuda-perf-mpi/tst/regression/outputs/advection_convergence_mpi/advection-errors.png
 
+parthenon-cuda-mpi-examples-library:
```
Yurlungur, line 160: Nice addition

.gitlab-ci.yml
```diff
@@ -157,6 +157,32 @@ parthenon-cuda-performance_and_regression-mpi:
...
+  only:
+    - schedules
+    - develop
```
Yurlungur, line 163: What's the develop run condition do?

Yurlungur (commented):
The install test and additional documentation are welcome additions.

On the side of moving the headers into the include directory... I'm honestly still having a lot of trouble mustering enthusiasm for this change. Four a couple reasons:
- What happens if I tell `cmake` to install this into 
```bash
~/local
```
that also contains other stuff. Will that local directory be readable? what if another library also installs a header with the same name? like `utils/packing_utils.hpp` or something?

- I understand that a solidified public API is desirable, but until we have one, it's going to be pretty inconvenient navigating the directory structure in the include directory.

That said. I'm certainly willing to be overruled here. It sounds like most people are on board with this. So I'll approve. But I'd like to hear from someone on the Athena physics team as well before it gets merged in.

.gitlab-ci.yml
JoshuaSBrown: the `only` keyword takes the following names as logical or. So this ci workflow will run when the `develop` branch is triggered or when a `scheduled` test run is triggered etc.

JoshuaSBrown:
I am going to recapitulate the internal discussion about this pr. As I am in favor of this pr my arguments will reflect this. 

# Arguments in Favor:

1. It follows convention, public headers are traditionally placed in a separate **include** dir and source files and private headers are left in a **src** dir. 
2. It will make the cmake configuration easier as all the public headers will be in one spot. 
3. It makes it possible to make headers private, this was not really possible before. Since https://github.com/lanl/parthenon/pull/238 all the headers have been public. 
4. The build configuration will now fail if you think you have a private internal file but it is actually included in one of the public headers, this will help with better design. 

# Arguments Against 

1. This was raised by @Yurlungur, this pr in it's current state will muddy up the directory where the public api is installed. Possibly leading to name clashes.

This can be addressed by appropriate scoping, placing the includes in a sub folder @AndrewGaspar ideas for the best way to address @Yurlungur's concern would be appreciated. 

2. It is too early to establish a public api, the code is still changing to rapidly. 

Down stream developers are already using parthenon, parthenon is already a public api whether this pr is merged or not since pr 238 it is made that way when installed as a shared library. Secondly, even without pr 238 it is being used as a submodule where the internal headers in the src directory are being used anyway, this pr does not change that. Leaving the code as is, only provides a false sense of flexibility. Changes to the code now will still cause downstream code to break. This pr just makes it clearer going forward where changes will likely break down stream code. 

3. This pr will require getting used to because the header files are being rearranged. 

Separation of public and private header files is going to happen sooner or later. It will be easier if we address it sooner than later. By delaying we are just introducing technical debt that we will be paying interest on later. 

4. This pr will slow progress. 

This is similar to 2., it depends, it can be a trade off. If you want to make a header file private and internal you now have the option but yes it will require more thought if you want the file to be part of the api you will be sacrificing speed for stability and better design. 

If a piece of code is likely to change rapidly but it needs to be accessible it can be placed in public without much thought, yes you will likely break something downstream but at least you will know you are breaking something. You can also get around people using experimental pieces by using the [[deprecated]] keyword with a message to warn users. 

I don't remember the other concerns but feel free to add them. I still need input from the Athena++ guys before moving forward.

brryan:
I'm sorry to weigh in negatively so late in the game since clearly a lot of work has already been done here but I didn't fully appreciate what this PR was doing until now. Putting all headers in one folder, especially separating class header/implementation files, would be disruptive and confusing for me, and I think other developers, in both short and long terms. The Arguments Against listed above are significant in my mind.

If I remember right, @Yurlungur suggested implementing a set of public headers through the build system -- I like this solution since it wouldn't overhaul the source directory structure. I think this would satisfy 2) and 4) of the Arguments in Favor. Since the parthenon code will probably always have a large fraction of its header files public, I'm not convinced that the convention in 1) makes good sense for parthenon, and I think losing the private-unless-explicitly-public check in 4) is a small concession to retaining the existing file structure. 

My hope is this changeset will not be merged, as it would decrease my productivity. I'm creating a blank change request here to reflect that I think there is broad consensus among the LANL physics people right now that we are not in favor of this at least until we discuss further, though @jdolence and @Yurlungur may disagree.

brryan (dismissed):
I think it's fair to say that there's a lot of uncertainty among LANL Physics members (@brryan @jdolence @Yurlungur) about what these changes would mean for us going forward.

JoshuaSBrown:
> Putting all headers in one folder, especially separating class header/implementation files, would be disruptive and confusing for me, and I think other developers, in both short and long terms. The Arguments Against listed above are significant in my mind.

What about this proves challenging? and let me clarify: 1. the file structure of the headers follows the same format as what is in the src directory, it is not flat, 2. there should be no difference in how you are including the files, you won't have to do anything special with relative includes, this can be handled by cmake such that you are using the same pragma statements.

AndrewGaspar:
Maybe we need to take a step back and figure out if there are some shared values here, and then we can figure out what the best approach to addressing those values is.

In my opinion, a software library distribution should:
1) Have a clearly defined public API
2) Have a deliberately designed public API
3) Follow established conventions around package installation.

I think this PR, while not completely addressing these first two values, is a step in that direction.

The third value is fully addressed - we can certainly discuss if this isn't desirable, but the downstream installer already has tools for addressing the concerns with it being a mess of an install directory.

So, I would ask if we have any disagreement on these three values - if so, what are they? If not, what are better alternatives for addressing them? I personally think the status quo is in direct conflict with the first two values.

brryan:
@JoshuaSBrown Thanks for clarifying about the non-flat header directory structure and include pragma syntax. This still means that I can't just see the header files associated with source files when I am working in the `src` tree, though.

@AndrewGaspar Thanks for formalizing this. I personally agree that your 1, 2, and 3 constitute a well-formed library distribution, and that currently `develop` is in conflict with all three. An API is probably especially desirable for parthenon because one wouldn't want to let packages mess with parthenon internals in unauthorized ways that could lead to surprising problems with other packages down the line when they are combined. At the same time I echo the concerns of others that API-level features (how to pack blocks in a per-rank way, for example) are sufficiently in flux that concretizing an API now could force expensive refactors later. There will always be a tradeoff between the benefit of having an established API and the cost of having to modify it going forward, I imagine, but my guess is the potential costs relative to benefits will decrease significantly with time in this case. But I'm not the best person to make those estimates for parthenon. 

My understanding is by using cmake to designate header files as either public or private (which should be relatively natural since we don't have `GLOB`s except in some install syntax) we can retain the existing internal directory structure while also satisfying 1 and 2. I'm not sure I'm fully aware of established conventions for 3 but if that means maintaining parthenon's internal directory structure in installed header files we should also be able to do this with the build system, no?

JoshuaSBrown:
> are sufficiently in flux that concretizing an API now could force expensive refactors later.

To clarify the current pr does not concertize the API anymore than the current status quo. The current status quo already treats everything as a public API. The status quo only gives you a false sense of flexibility. If you are using parthenon now you are having to include header files in /src/ there is nothing from preventing you from doing this. You will be paying the cost regardless of this pr. This pr simply lets you choose to make things private which will give you flexibility and confidence that you are not breaking things downstream. 

Just a head's up I will not be merging this pr until there is more consensus. As in, I will not force this, but I will definitively argue for it.

AndrewGaspar:
> My understanding is by using cmake to designate header files as either public or private (which should be relatively natural since we don't have GLOBs except in some install syntax) we can retain the existing internal directory structure while also satisfying 1 and 2.

I don't like this approach because it feels like a no-win compromise - I'd rather just continue operating under the assumption that all header files are public. By making a separate list of headers you have to cross-reference to determine which are public vs. private, parthenon developers will almost certainly make mistakes, and parthenon users who aren't aware of this can't easily determine which headers are public vs. private. They'd have to know we have this list somewhere, which isn't really something that's normally done.

> I'm not sure I'm fully aware of established conventions for 3

Typically libs are installed directly into the prefix/lib folder and headers are installed directly into prefix/include. By putting them anywhere else (e.g. in include/parthenon and lib/parthenon), it means you get none of the benefits of installing to a standard location, including automatic resolution of header files and libs - you would have to specify `-L` and `-I` paths to get the parthenon headers and libs. So you're sticking your headers and libs into a system install path, but still making the user know about our different directory structure to use the library. If the user is worried about conflicts, the right solution is to just install parthenon into its own prefix.

Now, ideally everybody uses CMake and this situation isn't really an issue. But my understand is we'd like to support non-cmake consumers.

pgrete:
I'm somewhat optimistic that we could have a `1.0.0`  release this year, which includes fully offloaded, sufficiently performance AMR on any devices with MeshBlock sizes of say 8^3 or 16^3 (basically our first big milestone).
How would people feel about trying to make a "`1.0.0`  - `0.1.0`" release with the features above an then focus the `0.1.0` increment on a clean interface separation that does not make everything public?

Yurlungur:
> I'm somewhat optimistic that we could have a `1.0.0` release this year, which includes fully offloaded, sufficiently performance AMR on any devices with MeshBlock sizes of say 8^3 or 16^3 (basically our first big milestone).
> How would people feel about trying to make a "`1.0.0` - `0.1.0`" release with the features above an then focus the `0.1.0` increment on a clean interface separation that does not make everything public?

This essentially kicks the can down the road. But I would be supportive of doing so, since I think the API will be more concrete after we've reached our performance goals.

JoshuaSBrown:
> I'm somewhat optimistic that we could have a `1.0.0` release this year, which includes fully offloaded, sufficiently performance AMR on any devices with MeshBlock sizes of say 8^3 or 16^3 (basically our first big milestone).
> How would people feel about trying to make a "`1.0.0` - `0.1.0`" release with the features above an then focus the `0.1.0` increment on a clean interface separation that does not make everything public?

I think this is a reasonable compromise if we can set a soft and hard date.

brryan:
> To clarify the current pr does not concertize the API anymore than the current status quo. The current status quo already treats everything as a public API. The status quo only gives you a false sense of flexibility. If you are using parthenon now you are having to include header files in /src/ there is nothing from preventing you from doing this. You will be paying the cost regardless of this pr. This pr simply lets you choose to make things private which will give you flexibility and confidence that you are not breaking things downstream.

I agree that this PR doesn't privatize any headers -- I was thinking about an API in a broader context in that paragraph. Once headers become private, if downstream application development reveals that a private header should actually be public, one is a merge to parthenon/develop away from progressing on one's own work. That's the loss of flexibility I meant. 

> Just a head's up I will not be merging this pr until there is more consensus. As in, I will not force this, but I will definitively argue for it.

Thanks, I thought this might be on track to be merged in right away with an athena++ approval and panicked. I'll close out my requested changes now but I think it would be good to have a LANL physics approver before this gets merged in. My concerns with moving all the headers and having a lot of them probably remain moved remain, but I am hardly the head honcho over at LANL physics.

> I don't like this approach because it feels like a no-win compromise - I'd rather just continue operating under the assumption that all header files are public. By making a separate list of headers you have to cross-reference to determine which are public vs. private, parthenon developers will almost certainly make mistakes, and parthenon users who aren't aware of this can't easily determine which headers are public vs. private. They'd have to know we have this list somewhere, which isn't really something that's normally done.

Your point about cross-referencing is well-taken but would the worst case scenario be getting a compiler error for a not-found header when compiling an application? One should get feedback pretty rapidly in that case -- even if one builds parthenon in-source, the cmake parthenon target would still not provide the private headers, right?. In my understanding this PR is vulnerable to a similar category of error -- one sees a header in the parthenon directory they're reading source code in and just `#include`s it. The more fearsome errors in my mind would be being able to include a private header when building in source, and then having that fail much later when importing parthenon as a library, but I don't think either this PR or my (Jonah's) suggestion suffer from that. Also, at least for me, seeing `include` and `src` directories in the parthenon source tree might not automatically tell me to avoid including headers in `src` since I'm only used to thinking about that breakdown in library installs. 

If the prelude-related `parthenon/driver.hpp` and `parthenon/package.hpp` headers (which would only include public headers I imagine) are promoted to inexperienced users, they should be able to just include those and get every public thing that they know about from the documentation I think. Those are all I use for my application-side work. 

> Typically libs are installed directly into the prefix/lib folder and headers are installed directly into prefix/include. By putting them anywhere else (e.g. in include/parthenon and lib/parthenon), it means you get none of the benefits of installing to a standard location, including automatic resolution of header files and libs - you would have to specify -L and -I paths to get the parthenon headers and libs. So you're sticking your headers and libs into a system install path, but still making the user know about our different directory structure to use the library. If the user is worried about conflicts, the right solution is to just install parthenon into its own prefix.

Understood, thanks -- I've just always been a custom prefix/-L/-I person (although now I'm a `find_package(parthenon)` person and it works great!)

jlippuner:
Will revisit if needed

PR #287, opened 2020-08-28
Machine configuration files

pgrete:
## PR Summary

This implement the machine configuration files we discussed during the talk.
All variables are optional, so this could be seen as reducing the number of arguments when calling cmake from the terminal.
Moreover, I hope the current design is flexible enough to include different configuration by allowing user to simply create their own machine files (hopefully not too convoluted) "variant" handling.

I only added Summit and the CI machine to illustrate how this could be used as a starting point for discussion.

## PR Checklist

- [X] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,30 @@
 #=========================================================================================
 
 cmake_minimum_required(VERSION 3.12)
+
+# Load machine specific defaults such as architecture and mpi launch command.
+# Command line argument takes precedence over environment variable.
+# Loading this before project definition to allow setting the compiler.
+if (MACHINE_CFG)
+  if(EXISTS "${MACHINE_CFG}")
+    include(${MACHINE_CFG})
+  else()
+    message(FATAL_ERROR "Given machine configuration at "
+      "${MACHINE_CFG} not found.")
+  endif()
+elseif (DEFINED ENV{MACHINE_CFG})
+  if(EXISTS "$ENV{MACHINE_CFG}")
+    include($ENV{MACHINE_CFG})
+  else()
+    message(FATAL_ERROR "Given machine configuration from environment variable "
+      "MACHINE_CFG at $ENV{MACHINE_CFG} not found.")
+  endif()
+else()
+  message(WARNING "Not using any machine configuration. Consider creating a configuration "
+    "file following the examples in ${PROJECT_SOURCE_DIR}/cmake/machine_cfgs/ and then "
+    "point the MACHINE_CFG variable to your custom file.")
+endif()
+
 project(parthenon VERSION 0.1.0 LANGUAGES C CXX)
 
 include(CTest)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,7 +33,8 @@ elseif (DEFINED ENV{MACHINE_CFG})
 else()
   message(WARNING "Not using any machine configuration. Consider creating a configuration "
     "file following the examples in ${PROJECT_SOURCE_DIR}/cmake/machine_cfgs/ and then "
-    "point the MACHINE_CFG variable to your custom file.")
+    "point the MACHINE_CFG variable to your custom file."
+    "Note, that the machine file can be placed in any directory (also outside the repo).")
 endif()
 
 project(parthenon VERSION 0.1.0 LANGUAGES C CXX)
--- a/cmake/TestSetup.cmake
+++ b/cmake/TestSetup.cmake
@@ -69,16 +69,35 @@ function(setup_test_coverage dir arg)
   endif()
 endfunction()
 
+function(process_mpi_args nproc)
+  list(APPEND TMPARGS "--mpirun")
+  # use custom mpiexec
+  if (TEST_MPIEXEC)
+    list(APPEND TMPARGS "${TEST_MPIEXEC}")
+  # use CMake determined mpiexec
+  else()
+    list(APPEND TMPARGS "${MPIEXEC_EXECUTABLE}")
+  endif()
+  list(APPEND TMPARGS "--mpirun_opts=${MPIEXEC_NUMPROC_FLAG}")
+  list(APPEND TMPARGS "--mpirun_opts=${nproc}")
+  # set additional options from machine configuration
+  foreach(MPIARG ${TEST_MPIOPTS})
+    list(APPEND TMPARGS "--mpirun_opts=${MPIARG}")
+  endforeach()
+
+  # make the result accessible in the calling function
+  set(MPIARGS ${TMPARGS} PARENT_SCOPE)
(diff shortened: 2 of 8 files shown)


CMakeLists.txt
```diff
@@ -12,6 +12,30 @@
...
+
+# Load machine specific defaults such as architecture and mpi launch command.
+# Command line argument takes precedence over environment variable.
+# Loading this before project definition to allow setting the compiler.
+if (MACHINE_CFG)
+  if(EXISTS "${MACHINE_CFG}")
+    include(${MACHINE_CFG})
+  else()
+    message(FATAL_ERROR "Given machine configuration at "
+      "${MACHINE_CFG} not found.")
```
Yurlungur, line 24: I like that it dies if you provide a missing file. I think this will be useful for debugging.

CMakeLists.txt
```diff
@@ -12,6 +12,30 @@
...
+  if(EXISTS "$ENV{MACHINE_CFG}")
+    include($ENV{MACHINE_CFG})
+  else()
+    message(FATAL_ERROR "Given machine configuration from environment variable "
+      "MACHINE_CFG at $ENV{MACHINE_CFG} not found.")
+  endif()
+else()
+  message(WARNING "Not using any machine configuration. Consider creating a configuration "
+    "file following the examples in ${PROJECT_SOURCE_DIR}/cmake/machine_cfgs/ and then "
+    "point the MACHINE_CFG variable to your custom file.")
```
Yurlungur: I also like this warning message.

cmake/TestSetup.cmake
```diff
@@ -69,16 +69,41 @@ function(setup_test_coverage dir arg)
   endif()
 endfunction()
 
+function(process_mpi_args nproc)
```
Yurlungur, line 72: This might be just a cmake misunderstanding on my part: are we using `nproc` in this function? Does it need to be passed in?

cmake/machinecfg/CI.cmake
```diff
@@ -0,0 +1,36 @@
+#========================================================================================
```
Yurlungur, line 1: Should we bother making a generic "mac" or "linux" machine file? Or perhaps the CI file is similar enough?

It would be good if one of the people using power9s on Darwin, say @AndrewGaspar , @jdolence , or @jlippuner added a file for that one.

cmake/machinecfg/Summit.cmake
```diff
@@ -0,0 +1,37 @@
...
+# of Energy/National Nuclear Security Administration. The Government is granted for
+# itself and others acting on its behalf a nonexclusive, paid-up, irrevocable worldwide
+# license in this material to reproduce, prepare derivative works, distribute copies to
+# the public, perform publicly and display publicly, and to permit others to do so.
+#========================================================================================
+
+message(STATUS "Loading machine configuration for OLCF's Summit.\n"
+  "Supported MACHINE_VARIANT includes 'cuda', 'mpi', and 'cuda-mpi'\n"
+  "This configuration has been tested using the following modules: "
+  "module load cuda gcc cmake/3.14.2 python hdf5\n")
```
Yurlungur, line 21: :+1:

tst/regression/utils/test_case.py
```diff
@@ -156,34 +148,14 @@ def __checkMPIExecutable(self,mpi_executable):
...
-        error_msg += mpi_exec 
-        error_msg += "\nExpected choices are:"
-        for choice in choices:
-            error_msg += choice
-            error_msg += "\n"
-        raise TestManagerError(error_msg)
+        if which(mpi_exec) is None:
+            error_msg = "mpi executable path provided, but no file found: "
+            error_msg += mpi_exec
+            raise TestManagerError(error_msg)
```
Yurlungur, line 154: Vast improvement here thanks to moving this logic into the machine specific files and leaving it up to the user.

docs/building.md
```diff
@@ -134,6 +134,26 @@ export PARTHENON_ROOT=$(pwd)/parthenon
...
+
+To make the default configuration on widely used systems easier, Parthenon provides machine configuration files that contain default options.
+Defaults options include, but are not limited to setting
+- the compiler (e.g., `nvcc_wrapper` for Cuda builds), or
+- paths to non default package locations (e.g., for a custom HDF5 install), or 
+- custom MPI related commands used in the Parthenon test suite (e.g., the launch command).
+
+The machine configurations shipped with Parthenon are located in [`PARTHENON_ROOT/cmake/machinecfg`](../cmake/machinecfg) and are named by the machine name.
+In order to use them either
+- set the `MACHINE_CFG` environment variable to the appropriate file, or
```
Yurlungur, line 147: If I do `cmake .. -DMACHINE_CFG=path/to/my/CMakeLists.txt -DDISABLE_HDF5=True` or something, will the `-DDISABLE_HDF5` override whatever's in my custom machine file? It should, right?

Yurlungur (approved):
I am 100% in favor. I think this is a vast improvement. Thanks for going through with it. A couple questions/comments below. But I think we should get this moved in regardless.

CMakeLists.txt
AndrewGaspar: I'd make a point, I suppose, that these machine files can live outside the repo for user-configurations.

docs/building.md
AndrewGaspar: Yes, it should, as long as you're doing `set(DISABLE_HDF5 OFF CACHE BOOL "Blah")`. `CACHE` variables do nothing if the variable is already set in cache.

AndrewGaspar (approved):

cmake/TestSetup.cmake
pgrete: It's used further down in line 88.

cmake/machinecfg/CI.cmake
pgrete: I think that a generic Mac file would be great (based on what I've heard about building programs on MacOS, e.g., with GCC or OpenMP).

Also, in addition to power9 on Darwin I got the impression that RZAnsel (?) is also used quite a bit by many people involved in development here.
May be worth adding a file for that machine, too.

CMakeLists.txt
pgrete:
```suggestion
    "point the MACHINE_CFG variable to your custom file."
    "Note, that the machine file can be placed in any directory (also outside the repo).)
```

pgrete: I added a note on that.

pgrete:
From my point of view this can go in and we can add additional machine config as needed later (and update the logic, if required).
If there are no objection by Monday evening, I'll merge then.

JoshuaSBrown (approved):
This is great thanks for doing the work @pgrete!

PR #291, opened 2020-09-01
Add Task diagram illustrating relationship between classes

### LINKED ISSUE #288: Diagram For Task Documentation
I could use feedback on how accurate this is. I would like to add a diagram to help illustrate how the different abstraction layers work. 

TaskDiagram

JoshuaSBrown:
## PR Summary

This pull request simply attempts to improve the documentation on the task infrastructure with a diagram. I could use some feedback as to whether the diagram is accurate.

## PR Checklist


Yurlungur (changes requested):
Thanks for pulling this together, @JoshuaSBrown ! It's a great start!

So I am having trouble connecting the picture to what's actually going on on an intuitive level. And I'm struggling to put into words why. Please bear with me as I try to explain. Hopefully my explanation will be understandable.

So each task region basically is initialized to allow some set number of task lists. Internally it's a list, but really ti's a graph of dependencies, and then as the dependencies are satisfied the tasks are launched. And each task list is "independent" of the other lists. i.e., there are no cross dependencies. Your picture captures this by showing lists of tasks in each region.

But one thing it doesn't capture, I think, is how this will typically be used. You're showing a very general scenario, where per meshblock tasks are mixed with aggregated tasks. I think a more realistic scenario is something like this:

1. You start with a fully asynchronous region which allows for `nmeshblocks` task lists. Here each task owns a meshblock and there are "nmeshblocks" lists. Again, these are really DAGs, not lists.

2. Then you move to a single task list region where you have only one task list and each task is over the whole mesh.

3. Then you move to a task region with 2 task lists, where each task is over 1/2 the mesh.

Probably impossible (and undesirable) to convey that each list is a DAG in the diagram. (Maybe an inset or something but that's probably too fancy.) But I think it would be good try and go for a workflow like above, instead of the more general case you're showing. What do you think?

I'd be interested in other people's thoughts too.

JoshuaSBrown:
Thanks for the feedback @Yurlungur. Can you give me an idea of what a simple DAG might consists of? Would each task list probably have the same DAG topology?

Yurlungur:
> Can you give me an idea of what a simple DAG might consists of?

Well for example, it might look like something like this? Where here the edges are all directed to the right

```
Dependency1----|                                              |-----ChildTask1
               |---TaskWith2Depencies---TaskWtih1Dependency---|
Dependency2----|                                              |-----ChildTask2
```

I wouldn't worry about this too much. DAGs can be essentially arbitrarily complicated. Just no cycles. ;)

> Would each task list probably have the same DAG topology?

Within a task region? Yes, I think so. You could have them be different, but I think that would be a pretty weird use-case.

Yurlungur:
> Can you give me an idea of what a simple DAG might consists of? 

Maybe another good example would be this one

```
Dependency1---------------------|
                                |
Dependency2--|                  |
             |--Child1--Child2--|--Child3
Dependency3--|
```

Where here Child1 depends on Dependency2 and Dependency3. Child3 depends on Child2 and Dependency1.

Yurlungur:
Setting as WIP until @JoshuaSBrown  returns.

JoshuaSBrown:
TaskDiagram

@Yurlungur How about this?

Yurlungur:
Hey, I like this, @JoshuaSBrown! What do you think, @jdolence ?

brryan:
Im not part of this convo but just saw this nice diagram pop up in my email and had a quick thought -- what if you label the tasks by which blocks are processing each one, like below (I didn't fill in all the labels), then one might not need the color-coded tasks and legend:
image

JoshuaSBrown:
TaskDiagram

jdolence:
I think this is getting really close.  Couple comments:

1. It's kind of ambiguous where the `TaskCollection` line is pointing
2. I think we need a caption, but I found the current one kind of confusing.  I'll take a stab at it below.
3. The horizontal direction is an implied time axis.  Is there any obvious way to indicate that explicitly?

Attempt at caption (just threw this together, so feel free to edit or not use this):

Tasks are organized into regions which are in turn organized into collections.  Task regions within a collection are executed sequentially and each task region can have a different granularity.  In the diagram, a task collection is composed of three task regions, controlling the execution of tasks on four mesh blocks, indicated by the numbers.  The first region launches (potentially) concurrent tasks on each mesh block, where the dependencies of a given task can only be other tasks that operate on the same mesh block.  Once all tasks in the region are complete, execution moves to the next region where three tasks are launched that each operate on all four mesh blocks simultaneously.  Finally, once these are complete, execution moves to the final region which defines tasks that operate of subsets of mesh blocks.  In this way, task granularity (with respect to how much data a task operates on) is controlled at the task region level, and overall execution is controlled at the collection level.

JoshuaSBrown:
Picture1

JoshuaSBrown:
@Yurlungur can you approve this, I need approvals from two other people, any takers, @jmstone, @forrestglines @AndrewGaspar @nmsriram ?

AndrewGaspar (approved):
Looks good.

Yurlungur (approved):
Yep looks good now. My bad, should have approved earlier.

PR #293, opened 2020-09-01
Remove HtoD copies

jdolence:
## PR Summary

Addresses the issue of explicit async HtoD MemCpy calls that came about because the size of the objects captured in the FluxDivergence kernel exceeded a critical size.  The fix was to move from `ParArrayND` objects to `ParArray1D` objects in VariablePack objects.  This change should be transparent to downstream codes.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - [[PR 303]](https://github.com/lanl/parthenon/pull/303) Changed `Mesh::BlockList` from a `std::list<MeshBlock>` to a `std::vector<std::shared_ptr<MeshBlock>>`, making `FindMeshBlock` run in constant, rather than linear, time. Loops over `block_list` in application drivers must be cahnged accordingly.
 
 ### Fixed (not changing behavior/API/variables/...)
+- [[PR 293]](https://github.com/lanl/parthenon/pull/293) Changed `VariablePack` and related objects to use `ParArray1D` objects instead of `ParArrayND` objects under the hood to reduce the size of the captured objects.
 
 ### Removed (removing behavior/API/varaibles/...)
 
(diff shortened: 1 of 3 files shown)


jdolence:
@pgrete @Yurlungur this might fix the HtoD copies reported in the chat, but I don't have time to get it profiled to confirm.  If one of you gets to it, please let me know if it fixes it.  Otherwise, I'll profile it as soon as I can.

pgrete:
Screenshot from 2020-09-01 17-35-06

Doesn't look like it's fully solved yet.

Yurlungur (approved):

AndrewGaspar (approved):
Thinking out loud: what if we added a `static_assert` to `par_for` that constrains lambdas from passing a certain size? Would help us catch this issue earlier.

Yurlungur:
> Thinking out loud: what if we added a `static_assert` to `par_for` that constrains lambdas from passing a certain size? Would help us catch this issue earlier.

We might end up needing to pass lambdas of that size at some point for some reason though. Is there a `static_warn`?

AndrewGaspar:
> We might end up needing to pass lambdas of that size at some point for some reason though. Is there a static_warn?

We could potentially add an optional tag that suppresses the warning. So it would go to a different overload and skip the static_assert - it would require you to think if that's what you really want.

e.g.
```c++
struct SuppressSizeWarningTag {};
constexpr SuppressSizeWarningTag SuppressSizeWarning{};

pmb->par_for(parthenon::SuppressSizeWarning, ...);
```

jdolence:
Could we have something in `par_for` that warns us, but only if enabled at compile time (and is off by default)?  It seems like yes is the answer.  Still not sure we want this.  Now that we appreciate this issue, I'd vote for just trying to be careful going forward.  If it pops up again in the profile, we'll know immediately what's going on.

jdolence:
@pgrete, ok if I go ahead and merge this in or did you want to review it?  Trying to give you a pass because I know you're busy.

pgrete (approved):
Screenshot from 2020-09-24 12-03-28

Happy to confirm that the HtoD copies are gone. Attaching profiling screenshot for reference.

PR #294, opened 2020-09-01
Fix `IndexShape::GetTotal(IndexDomain)`

AndrewGaspar:
## PR Summary
The conditional in `IndexDomain::GetTotal` was backwards. This PR adds a test and fixes it.

## PR Checklist

- [x] Code passes cpplint
- [x] New features are documented. (N/A)
- [x] Code is formatted

### DIFF
--- a/tst/unit/test_unit_domain.cpp
+++ b/tst/unit/test_unit_domain.cpp
@@ -36,6 +36,7 @@ TEST_CASE("Checking IndexShape indices", "[is,ie,js,je,ks,ke]") {
     REQUIRE(shape.je(interior) == 0);
     REQUIRE(shape.ks(interior) == 0);
     REQUIRE(shape.ke(interior) == 0);
+    REQUIRE(shape.GetTotal(interior) == 6);
 
     REQUIRE(shape.is(entire) == 0);
     REQUIRE(shape.ie(entire) == 2 * num_ghost + nx1 - 1);
@@ -43,6 +44,7 @@ TEST_CASE("Checking IndexShape indices", "[is,ie,js,je,ks,ke]") {
     REQUIRE(shape.je(entire) == 0);
     REQUIRE(shape.ks(entire) == 0);
     REQUIRE(shape.ke(entire) == 0);
+    REQUIRE(shape.GetTotal(entire) == 8);
   }
 
   GIVEN("A 2D Index Shape") {
@@ -57,6 +59,7 @@ TEST_CASE("Checking IndexShape indices", "[is,ie,js,je,ks,ke]") {
     REQUIRE(shape.je(interior) == num_ghost + nx2 - 1);
     REQUIRE(shape.ks(interior) == 0);
     REQUIRE(shape.ke(interior) == 0);
+    REQUIRE(shape.GetTotal(interior) == 6);
 
     REQUIRE(shape.is(entire) == 0);
     REQUIRE(shape.ie(entire) == 2 * num_ghost + nx1 - 1);
@@ -64,6 +67,7 @@ TEST_CASE("Checking IndexShape indices", "[is,ie,js,je,ks,ke]") {
     REQUIRE(shape.je(entire) == 2 * num_ghost + nx2 - 1);
     REQUIRE(shape.ks(entire) == 0);
     REQUIRE(shape.ke(entire) == 0);
+    REQUIRE(shape.GetTotal(entire) == 24);
   }
 
   GIVEN("A 3D Index Shape") {
@@ -79,6 +83,7 @@ TEST_CASE("Checking IndexShape indices", "[is,ie,js,je,ks,ke]") {
     REQUIRE(shape.je(interior) == num_ghost + nx2 - 1);
     REQUIRE(shape.ks(interior) == num_ghost);
     REQUIRE(shape.ke(interior) == num_ghost + nx3 - 1);
+    REQUIRE(shape.GetTotal(interior) == 24);
 
     REQUIRE(shape.is(entire) == 0);
     REQUIRE(shape.ie(entire) == 2 * num_ghost + nx1 - 1);
@@ -86,6 +91,7 @@ TEST_CASE("Checking IndexShape indices", "[is,ie,js,je,ks,ke]") {
     REQUIRE(shape.je(entire) == 2 * num_ghost + nx2 - 1);
     REQUIRE(shape.ks(entire) == 0);
     REQUIRE(shape.ke(entire) == 2 * num_ghost + nx3 - 1);
+    REQUIRE(shape.GetTotal(entire) == 144);
   }
 
   GIVEN("A 3D Index Shape initialize with vector") {
@@ -102,6 +108,7 @@ TEST_CASE("Checking IndexShape indices", "[is,ie,js,je,ks,ke]") {
     REQUIRE(shape.je(interior) == num_ghost + nx2 - 1);
     REQUIRE(shape.ks(interior) == num_ghost);
     REQUIRE(shape.ke(interior) == num_ghost + nx3 - 1);
+    REQUIRE(shape.GetTotal(interior) == 24);
 
     REQUIRE(shape.is(entire) == 0);
     REQUIRE(shape.ie(entire) == 2 * num_ghost + nx1 - 1);
@@ -109,6 +116,7 @@ TEST_CASE("Checking IndexShape indices", "[is,ie,js,je,ks,ke]") {
     REQUIRE(shape.je(entire) == 2 * num_ghost + nx2 - 1);
     REQUIRE(shape.ks(entire) == 0);
     REQUIRE(shape.ke(entire) == 2 * num_ghost + nx3 - 1);
+    REQUIRE(shape.GetTotal(entire) == 144);
   }
 
   GIVEN("A 3D Index Shape 0 dim in nx3") {
@@ -125,6 +133,7 @@ TEST_CASE("Checking IndexShape indices", "[is,ie,js,je,ks,ke]") {
     REQUIRE(shape.je(interior) == num_ghost + nx2 - 1);
     REQUIRE(shape.ks(interior) == 0);
     REQUIRE(shape.ke(interior) == 0);
+    REQUIRE(shape.GetTotal(interior) == 6);
 
     REQUIRE(shape.is(entire) == 0);
     REQUIRE(shape.ie(entire) == 2 * num_ghost + nx1 - 1);
@@ -132,6 +141,7 @@ TEST_CASE("Checking IndexShape indices", "[is,ie,js,je,ks,ke]") {
     REQUIRE(shape.je(entire) == 2 * num_ghost + nx2 - 1);
     REQUIRE(shape.ks(entire) == 0);
     REQUIRE(shape.ke(entire) == 0);
+    REQUIRE(shape.GetTotal(entire) == 24);
   }
 }
 
--- a/tst/unit/test_unit_domain.cpp
+++ b/tst/unit/test_unit_domain.cpp
@@ -30,8 +30,8 @@ TEST_CASE("Checking IndexShape indices", "[is,ie,js,je,ks,ke]") {
     int num_ghost = 1;
 
     parthenon::IndexShape shape(nx1, num_ghost);
-    REQUIRE(shape.is(interior) == num_ghost);
-    REQUIRE(shape.ie(interior) == num_ghost + nx1 - 1);
+    REQUIRE(shape.is(interior) == 1);
+    REQUIRE(shape.ie(interior) == 6);
     REQUIRE(shape.js(interior) == 0);
     REQUIRE(shape.je(interior) == 0);
     REQUIRE(shape.ks(interior) == 0);
@@ -39,7 +39,7 @@ TEST_CASE("Checking IndexShape indices", "[is,ie,js,je,ks,ke]") {
     REQUIRE(shape.GetTotal(interior) == 6);
 
     REQUIRE(shape.is(entire) == 0);
-    REQUIRE(shape.ie(entire) == 2 * num_ghost + nx1 - 1);
+    REQUIRE(shape.ie(entire) == 7);
     REQUIRE(shape.js(entire) == 0);
     REQUIRE(shape.je(entire) == 0);
     REQUIRE(shape.ks(entire) == 0);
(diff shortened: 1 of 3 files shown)


Yurlungur (approved):

tst/unit/test_unit_domain.cpp
```diff
@@ -36,13 +36,15 @@ TEST_CASE("Checking IndexShape indices", "[is,ie,js,je,ks,ke]") {
...
     REQUIRE(shape.ke(interior) == 0);
+    REQUIRE(shape.GetTotal(interior) == 6);
 
     REQUIRE(shape.is(entire) == 0);
     REQUIRE(shape.ie(entire) == 2 * num_ghost + nx1 - 1);
     REQUIRE(shape.js(entire) == 0);
     REQUIRE(shape.je(entire) == 0);
     REQUIRE(shape.ks(entire) == 0);
     REQUIRE(shape.ke(entire) == 0);
+    REQUIRE(shape.GetTotal(entire) == 8);
```
pgrete, line 47:
```suggestion
    REQUIRE(shape.GetTotal(entire) == 2 * num_ghost + nx1);
```

tst/unit/test_unit_domain.cpp
```diff
@@ -36,13 +36,15 @@ TEST_CASE("Checking IndexShape indices", "[is,ie,js,je,ks,ke]") {
     REQUIRE(shape.je(interior) == 0);
     REQUIRE(shape.ks(interior) == 0);
     REQUIRE(shape.ke(interior) == 0);
+    REQUIRE(shape.GetTotal(interior) == 6);
```
pgrete, line 39:
```suggestion
    REQUIRE(shape.GetTotal(interior) == nx1);
```

tst/unit/test_unit_domain.cpp
```diff
@@ -57,13 +59,15 @@ TEST_CASE("Checking IndexShape indices", "[is,ie,js,je,ks,ke]") {
     REQUIRE(shape.je(interior) == num_ghost + nx2 - 1);
     REQUIRE(shape.ks(interior) == 0);
     REQUIRE(shape.ke(interior) == 0);
+    REQUIRE(shape.GetTotal(interior) == 6);
```
pgrete, line 62:
```suggestion
    REQUIRE(shape.GetTotal(interior) == nx1 * nx2);
```

tst/unit/test_unit_domain.cpp
```diff
@@ -57,13 +59,15 @@ TEST_CASE("Checking IndexShape indices", "[is,ie,js,je,ks,ke]") {
...
 
     REQUIRE(shape.is(entire) == 0);
     REQUIRE(shape.ie(entire) == 2 * num_ghost + nx1 - 1);
     REQUIRE(shape.js(entire) == 0);
     REQUIRE(shape.je(entire) == 2 * num_ghost + nx2 - 1);
     REQUIRE(shape.ks(entire) == 0);
     REQUIRE(shape.ke(entire) == 0);
+    REQUIRE(shape.GetTotal(entire) == 24);
```
pgrete, line 70:
```suggestion
    REQUIRE(shape.GetTotal(entire) == (2 * num_ghost + nx1) * (2 * num_ghost + nx2));
```

tst/unit/test_unit_domain.cpp
```diff
@@ -79,13 +83,15 @@ TEST_CASE("Checking IndexShape indices", "[is,ie,js,je,ks,ke]") {
     REQUIRE(shape.je(interior) == num_ghost + nx2 - 1);
     REQUIRE(shape.ks(interior) == num_ghost);
     REQUIRE(shape.ke(interior) == num_ghost + nx3 - 1);
+    REQUIRE(shape.GetTotal(interior) == 24);
```
pgrete, line 86:
```suggestion
    REQUIRE(shape.GetTotal(interior) == nx1 * nx2 * nx3);
```

tst/unit/test_unit_domain.cpp
```diff
@@ -79,13 +83,15 @@ TEST_CASE("Checking IndexShape indices", "[is,ie,js,je,ks,ke]") {
...
 
     REQUIRE(shape.is(entire) == 0);
     REQUIRE(shape.ie(entire) == 2 * num_ghost + nx1 - 1);
     REQUIRE(shape.js(entire) == 0);
     REQUIRE(shape.je(entire) == 2 * num_ghost + nx2 - 1);
     REQUIRE(shape.ks(entire) == 0);
     REQUIRE(shape.ke(entire) == 2 * num_ghost + nx3 - 1);
+    REQUIRE(shape.GetTotal(entire) == 144);
```
pgrete, line 94:
```suggestion
    REQUIRE(shape.GetTotal(entire) == (2 * num_ghost + nx1) * (2 * num_ghost + nx2) * (2 * num_ghost + nx3));
```

tst/unit/test_unit_domain.cpp
```diff
@@ -102,13 +108,15 @@ TEST_CASE("Checking IndexShape indices", "[is,ie,js,je,ks,ke]") {
     REQUIRE(shape.je(interior) == num_ghost + nx2 - 1);
     REQUIRE(shape.ks(interior) == num_ghost);
     REQUIRE(shape.ke(interior) == num_ghost + nx3 - 1);
+    REQUIRE(shape.GetTotal(interior) == 24);
```
pgrete, line 111:
```suggestion
    REQUIRE(shape.GetTotal(interior) == nx1 * nx2 * nx3);
```

tst/unit/test_unit_domain.cpp
```diff
@@ -102,13 +108,15 @@ TEST_CASE("Checking IndexShape indices", "[is,ie,js,je,ks,ke]") {
...
 
     REQUIRE(shape.is(entire) == 0);
     REQUIRE(shape.ie(entire) == 2 * num_ghost + nx1 - 1);
     REQUIRE(shape.js(entire) == 0);
     REQUIRE(shape.je(entire) == 2 * num_ghost + nx2 - 1);
     REQUIRE(shape.ks(entire) == 0);
     REQUIRE(shape.ke(entire) == 2 * num_ghost + nx3 - 1);
+    REQUIRE(shape.GetTotal(entire) == 144);
```
pgrete, line 119:
```suggestion
    REQUIRE(shape.GetTotal(entire) == (2 * num_ghost + nx1) * (2 * num_ghost + nx2) * (2 * num_ghost + nx3));
```

tst/unit/test_unit_domain.cpp
```diff
@@ -125,13 +133,15 @@ TEST_CASE("Checking IndexShape indices", "[is,ie,js,je,ks,ke]") {
...
     REQUIRE(shape.ke(interior) == 0);
+    REQUIRE(shape.GetTotal(interior) == 6);
 
     REQUIRE(shape.is(entire) == 0);
     REQUIRE(shape.ie(entire) == 2 * num_ghost + nx1 - 1);
     REQUIRE(shape.js(entire) == 0);
     REQUIRE(shape.je(entire) == 2 * num_ghost + nx2 - 1);
     REQUIRE(shape.ks(entire) == 0);
     REQUIRE(shape.ke(entire) == 0);
+    REQUIRE(shape.GetTotal(entire) == 24);
```
pgrete, line 144:
```suggestion
    REQUIRE(shape.GetTotal(entire) == (2 * num_ghost + nx1) * (2 * num_ghost + nx2));
```

tst/unit/test_unit_domain.cpp
```diff
@@ -125,13 +133,15 @@ TEST_CASE("Checking IndexShape indices", "[is,ie,js,je,ks,ke]") {
     REQUIRE(shape.je(interior) == num_ghost + nx2 - 1);
     REQUIRE(shape.ks(interior) == 0);
     REQUIRE(shape.ke(interior) == 0);
+    REQUIRE(shape.GetTotal(interior) == 6);
```
pgrete, line 136:
```suggestion
    REQUIRE(shape.GetTotal(interior) == nx1 * nx2);
```

pgrete (approved):
Good catch. The suggested changes are for consistency with the use of `num_ghost` and `nx#` in the other REQUIRE statements.

pgrete:
While looking at this, I was wondering if 
```c++
  GIVEN("A 3D Index Shape 0 dim in nx3") {
    int nx1 = 6;
    int nx2 = 1;
    int nx3 = 0;
    std::vector<int> nxs = {nx3, nx2, nx1};
    int num_ghost = 1;

    parthenon::IndexShape shape(nxs, num_ghost);
    REQUIRE(shape.is(interior) == num_ghost);
    REQUIRE(shape.ie(interior) == num_ghost + nx1 - 1);
    REQUIRE(shape.js(interior) == num_ghost);
    REQUIRE(shape.je(interior) == num_ghost + nx2 - 1);
    REQUIRE(shape.ks(interior) == 0);
    REQUIRE(shape.ke(interior) == 0);
    REQUIRE(shape.GetTotal(interior) == 6);

    REQUIRE(shape.is(entire) == 0);
    REQUIRE(shape.ie(entire) == 2 * num_ghost + nx1 - 1);
    REQUIRE(shape.js(entire) == 0);
    REQUIRE(shape.je(entire) == 2 * num_ghost + nx2 - 1);
    REQUIRE(shape.ks(entire) == 0);
    REQUIRE(shape.ke(entire) == 0);
    REQUIRE(shape.GetTotal(entire) == 24);
  }
````

is actually desired behavior?
As far as I can tell the convention/assumption is that for <3D sims the the nx3, and/or nx3 and nx2 variables are 1 and never 0 (which may have consequences with respect to indexing.

**Comment by Yurlungur** [comment]
> While looking at this, I was wondering if
> 
> ```c++
>   GIVEN("A 3D Index Shape 0 dim in nx3") {
>     int nx1 = 6;
>     int nx2 = 1;
>     int nx3 = 0;
>     std::vector<int> nxs = {nx3, nx2, nx1};
>     int num_ghost = 1;
> 
>     parthenon::IndexShape shape(nxs, num_ghost);
>     REQUIRE(shape.is(interior) == num_ghost);
>     REQUIRE(shape.ie(interior) == num_ghost + nx1 - 1);
>     REQUIRE(shape.js(interior) == num_ghost);
>     REQUIRE(shape.je(interior) == num_ghost + nx2 - 1);
>     REQUIRE(shape.ks(interior) == 0);
>     REQUIRE(shape.ke(interior) == 0);
>     REQUIRE(shape.GetTotal(interior) == 6);
> 
>     REQUIRE(shape.is(entire) == 0);
>     REQUIRE(shape.ie(entire) == 2 * num_ghost + nx1 - 1);
>     REQUIRE(shape.js(entire) == 0);
>     REQUIRE(shape.je(entire) == 2 * num_ghost + nx2 - 1);
>     REQUIRE(shape.ks(entire) == 0);
>     REQUIRE(shape.ke(entire) == 0);
>     REQUIRE(shape.GetTotal(entire) == 24);
>   }
> ```
> 
> is actually desired behavior?
> As far as I can tell the convention/assumption is that for <3D sims the the nx3, and/or nx3 and nx2 variables are 1 and never 0 (which may have consequences with respect to indexing.

Good catch. `nx3` should be 1 here. That said, I think `ks=ke=0` is correct, because bounds are inclusive. And the totals are also correct. So only change is we should have `nx3=1`.

**Comment by AndrewGaspar** [comment]
@pgrete Well, I noticed that all of the other tests derived the expected value, and I didn't follow that on purpose.

The reason I don't like that pattern in unit tests is it makes it incredibly easy to just make the same mistake both in the actual code and in the unit test. Whereas hard-coding the expected result means you (hopefully) actually analytically determined the correct result.

Thoughts?

**Comment by AndrewGaspar** [comment]
I was also surprised at first by the behavior with `0` dimensions. I'm not sure I should change it in this PR, though... It was intentionally designed that way. @JoshuaSBrown, thoughts?

> nx3 should be 1 here.

This test is intentionally testing a 0 dimension.

**Comment by Yurlungur** [comment]
> This test is intentionally testing a 0 dimension.

I don't think it's a problem that a 0 dimension works like this, but we should test the behavior for 1 dimension because that's how it's actually initialized in the code.

**Comment by AndrewGaspar** [comment]
Ah, I think I understand - Yeah, actually, the result of `GetTotal(entire)` is wrong for 1-length, right?

**Comment by AndrewGaspar** [comment]
Like, a "6x1x1" mesh with nghost=1 is supposed to be a 1D mesh, right? So that should have 8x1x1 cells, not 8x3x3, right?

**Comment by Yurlungur** [comment]
> Like, a "6x1x1" mesh with nghost=1 is supposed to be a 1D mesh, right? So that should have 8x1x1 cells, not 8x3x3, right?

Right. That's how I think it should be working.

But actually, looking at the code, the plot thickens a little bit. Maybe we should indeed put this off for a later PR and get @JoshuaSBrown 's feedback. I see in `meshblock.cpp` that we have
```C++
  if (pmy_mesh->ndim >= 3) {
    InitializeIndexShapes(block_size.nx1, block_size.nx2, block_size.nx3);
  } else if (pmy_mesh->ndim >= 2) {
    InitializeIndexShapes(block_size.nx1, block_size.nx2, 0);
  } else {
    InitializeIndexShapes(block_size.nx1, 0, 0);
  }
```
meaning `IndexShape` is actually initialized with `nx3=0`, not `nx3=1`. And in `container.cpp` we have
```C++
template <typename T>
void Container<T>::calcArrDims_(std::array<int, 6> &arrDims, const std::vector<int> &dims,
                                const Metadata &metadata) {
  const int N = dims.size();

  if (metadata.Where() == Metadata::Cell || metadata.Where() == Metadata::Face ||
      metadata.Where() == Metadata::Edge || metadata.Where() == Metadata::Node) {
    // Let the FaceVariable, EdgeVariable, and NodeVariable
    // classes add the +1's where needed.  They all expect
    // these dimensions to be the number of cells in each
    // direction, NOT the size of the arrays
    assert(N >= 0 && N <= 3);
    const IndexDomain entire = IndexDomain::entire;
    arrDims[0] = pmy_block->cellbounds.ncellsi(entire);
    arrDims[1] = pmy_block->cellbounds.ncellsj(entire);
    arrDims[2] = pmy_block->cellbounds.ncellsk(entire);
    for (int i = 0; i < N; i++)
      arrDims[i + 3] = dims[i];
    for (int i = N; i < 3; i++)
      arrDims[i + 3] = 1;
  } else {
    // This variable is not necessarily tied to any specific
    // mesh element, so dims will be used as the actual array
    // size in each dimension
    assert(N >= 1 && N <= 6);
    for (int i = 0; i < N; i++)
      arrDims[i] = dims[i];
    for (int i = N; i < 6; i++)
      arrDims[i] = 1;
  }
}
```
where for an unused dimension, `ncells(entire)` gives `2*ng`. I think that's a bug. It should give `1`.

I'm going to file an issue. But I don't think it should block this MR.

Yurlungur:
As discussed on Matrix/Element: I missed an if-case in my sleuthing. Current behaviour is correct, but a little unintuitive for those of us expecting `nx3=1` to indicate a 2D mesh. I have opened issue #295 to remind myself to address this in a separate PR.

AndrewGaspar:
We discussed on the Matrix channel - we've decided that it's kind of confusing that `IndexShape` uses `0` for representing dimensions without ghost zones rather than `1`, since that's what the rest of the code uses. I appreciate, though, the reason it was chosen - within `IndexDomain`'s scope, there's no reason you should treat `1` differently from `2`. However, we never really allow dimensions with ghost zones at size 1 anyway, so it may be for the best to just change `IndexShape` to treat size `1` dimensions the same as size `0`.

AndrewGaspar:
@pgrete I made it consistent.

AndrewGaspar:
I think @JoshuaSBrown is, indeed, out now - I'll merge this when the CI passes.

PR #298, opened 2020-09-03
Infrastructure Required to Support MeshPacks of arbitrary size.

Yurlungur:
## PR Summary

As discussed in the call today, we are looking to move to always packing meshblocks and looping over meshblocks. This PR implements the required infrastructure to enable this on the mesh pack and tasking side. I apply it to the `calculate_pi` example.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/docs/mesh/packing.md
+++ b/docs/mesh/packing.md
@@ -27,6 +27,35 @@ can pass in a metadata vector, a vector of names, and optionally IDs
 and a map from names to indices. See
 [here](../interface/containers.md) for more details.
 
+## Packing over a piece of the mesh
+
+Instead of packing over the whole mesh, you can pack over only a piece
+of it by using the `Partition` machinery found in
+`utils/partition_stl_containers.hpp`. For example, to break the mesh
+into four evenly spaced meshpacks, do
+```C++
+using parthenon::MeshBlock;
+using parthenon::Partition::Partition_t;
+Partition_t<MeshBlock> partitions;
+parthenon::Partition::ToNPartitions(mesh->block, 4, partitions);
+MeshPack<VariablePack<Real>> packs[4];
+for (int i = 0; i < partitions.size() {
+  packs[i] = PackVariablesOnMesh(partitions[i], "base");
+}
+```
+
+There are two partitioning functions:
+```C++
+// Splits container into N equally sized partitions
+template <typename Container_t, typename T>
+void ToNPartitions(Container_t &container, const int N, Partition_t<T> &partitions);
+
+// Splits container into partitions of size N
+template <typename Container_t, typename T>
+void ToSizeN(Container_t &container, const int N, Partition_t<T> &partitions);
+```
+Both functions live within the namespace `parthenon::Partition`.
+
 ## Data Layout
 
 The Mesh Pack is indexable as a five-dimensional `Kokkos` view. The
--- a/docs/mesh/packing.md
+++ b/docs/mesh/packing.md
@@ -95,3 +95,16 @@ int sparse = meshpack.GetSparse(n);
 // The size of the n'th dimension of the pack
 int dim = meshpack.GetDim(n);
 ```
+
+## Type
+
+The types for packs are:
+```C++
+MeshBlockVarPack<T>
+```
+and
+```C++
+MeshBlockVarFluxPack<T>
+```
+which correspond to packs over meshblocks that contain just variables
+or contain variables and fluxes.
--- a/docs/mesh/packing.md
+++ b/docs/mesh/packing.md
@@ -35,7 +35,7 @@ and a map from names to indices. See
 Instead of packing over the whole mesh, you can pack over only a piece
 of it by using the `Partition` machinery found in
 `utils/partition_stl_containers.hpp`. For example, to break the mesh
-into four evenly spaced meshpacks, do
+into four evenly sized meshpacks, do
 ```C++
 using parthenon::MeshBlock;
 using parthenon::Partition::Partition_t;
--- a/docs/mesh/packing.md
+++ b/docs/mesh/packing.md
@@ -46,6 +46,12 @@ for (int i = 0; i < partitions.size() {
   packs[i] = PackVariablesOnMesh(partitions[i], "base");
 }
 ```
+To pack only the variables "var1" and "var2" in the container named "mycontainer", do:
+```C++
+for (int i = 0; i < partitions.size() {
+  packs[i] = PackVariablesOnMesh(partitions[i], "myContainer", {"var1","var2"});
+}
+```
 
 There are two partitioning functions:
 ```C++
--- a/docs/mesh/packing.md
+++ b/docs/mesh/packing.md
@@ -4,22 +4,25 @@
 kernels that perform little work, this can be a perforamnce bottleneck
 when each kernel is launched per meshblock. Parthenon therefore
 provides the capability to combine variables into a single data
-structure that spans all meshblocks, the `MeshPack`. 
+structure that spans some number of meshblocks, the `MeshBlockPack`.
 
-## Creating a Mesh Pack
+## Creating a MeshBlockPack
 
 There are two methods for creating mesh packs, which are analogous to the `VariablePack` and `VariableFluxPack` available in [containers](../interface/containers.md).
 ```C++
-template <typename... Args>
-auto PackVariablesOnMesh(Mesh *pmesh, const std::string &container_name,
+template <typename T, typename... Args>
+auto PackVariablesOnMesh(T &blocks, const std::string &container_name,
                          Args &&... args)
 ```
 and
 ```C++
-auto PackVariablesAndFluxesOnMesh(Mesh *pmesh, const std::string &container_name,
+template <typename T, typename... Args>
+auto PackVariablesAndFluxesOnMesh(T &blocks, const std::string &container_name,
                                   Args &&... args)
 ```
 The former packs only the variables, the latter packs in the variables and associated flux vectors.
+
+Here `T` can be the mesh or any standard template container that contains meshblocks.
 
 The variatic arguments take exactly the same arguments as
 `container.PackVariables` and `container.PackVariablesAndFluxes`. You
@@ -38,7 +41,7 @@ using parthenon::MeshBlock;
 using parthenon::Partition::Partition_t;
 Partition_t<MeshBlock> partitions;
 parthenon::Partition::ToNPartitions(mesh->block, 4, partitions);
-MeshPack<VariablePack<Real>> packs[4];
+MeshBlockPack<VariablePack<Real>> packs[4];
 for (int i = 0; i < partitions.size() {
   packs[i] = PackVariablesOnMesh(partitions[i], "base");
 }
@@ -73,7 +76,7 @@ auto var = meshpack(m,l); // Indexes into the k'th variable on the m'th MB
 Real r = meshpack(m,l,k,j,i); 
 ```
 
-For convenience, `MeshPack` also includes the following methods and fields:
+For convenience, `MeshBlockPack` also includes the following methods and fields:
 ```C++
 // the IndexShape object describing the bounds of all blocks in the pack
 IndexShape bounds = meshpack.cellbounds; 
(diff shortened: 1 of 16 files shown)


Yurlungur:
I think this still requires some iterations on the API. However, it works and shows how to pass through mesh packs of variable size through the `pi` example. This is somewhat related to @pgrete 's MR #299 and maybe bears some more discussion.

I'm interested to hear the thoughts of interested parties.

Yurlungur:
Example output, run on my laptop VM. (Still needs further testing on GPUs.)

This one uses the whole mesh in the pack.
```bash
jonahm@ubuntu:$ mpirun -np 2 ./example/calculate_pi/pi-example -i ../example/calculate_pi/parthinput.example Pi/pack_size=-1
### Warning in Mesh::Initialize
The number of MeshBlocks increased more than twice during initialization.
More computing power than you expected may be required.
### Warning in Mesh::Initialize
The number of MeshBlocks increased more than twice during initialization.
More computing power than you expected may be required.
### Warning in Mesh::Initialize
The number of MeshBlocks increased more than twice during initialization.
More computing power than you expected may be required.

PI = 3.14323    rel error = 0.000520918


walltime used = 1.20218
zone-cycles/wallsecond = 19378.2
```

This one uses packs, each of size 4 meshblocks
```bash
jonahm@ubuntu:$ mpirun -np 2 ./example/calculate_pi/pi-example -i ../example/calculate_pi/parthinput.example Pi/pack_size=4
### Warning in Mesh::Initialize
The number of MeshBlocks increased more than twice during initialization.
More computing power than you expected may be required.
### Warning in Mesh::Initialize
The number of MeshBlocks increased more than twice during initialization.
More computing power than you expected may be required.
### Warning in Mesh::Initialize
The number of MeshBlocks increased more than twice during initialization.
More computing power than you expected may be required.

PI = 3.14323    rel error = 0.000520918


walltime used = 1.12169
zone-cycles/wallsecond = 20768.7
```
And 1 meshblock
```bash
jonahm@ubuntu:$ mpirun -np 2 ./example/calculate_pi/pi-example -i ../example/calculate_pi/parthinput.example Pi/pack_size=1
### Warning in Mesh::Initialize
The number of MeshBlocks increased more than twice during initialization.
More computing power than you expected may be required.
### Warning in Mesh::Initialize
The number of MeshBlocks increased more than twice during initialization.
More computing power than you expected may be required.
### Warning in Mesh::Initialize
The number of MeshBlocks increased more than twice during initialization.
More computing power than you expected may be required.

PI = 3.14323    rel error = 0.000520918


walltime used = 1.08579
zone-cycles/wallsecond = 21455.4
```

Don't take the performance numbers too seriously. The load on my machine is not even.

Yurlungur:
As discussed on the call, MeshPack has been renamed MeshBlockPack.

Yurlungur:
Okay, I think this is ready for review. 

There's one thing I'd like feedback on, which could be changed, or we could keep it as is. Currently, I demand that users *first* partition `pmesh->block_list` and *then* pull out `MeshBlockPacks` e.g., the code loop looks something like this:
```C++
  Partition::Partition_t<MeshBlock> partitions;
  Partition::ToSizeN(blocks, pack_size, partitions);
  ParArrayHost<Real> areas("areas", partitions.size());

  TaskRegion &async_region = tc.AddRegion(partitions.size());
  {
    // asynchronous region where area is computed per mesh pack
    for (int i = 0; i < partitions.size(); i++) {
      TaskID none(0);
      auto pack = PackVariablesOnMesh(partitions[i], "base",
                                      std::vector<std::string>{"in_or_out"});
      auto get_area = async_region[i].AddTask(ComputeArea, none, pack, areas, i);
    }
  }
```
This allows the user to re-use the `partitions` object, which I expect will be common in multi-stage task-lists, where different packs are constructed over the same set of meshblocks.

However I *could* introduce some syntactic sugar so that the partitioning can be handled automatically when the packs are constructed. I'd be interested to hear what people would prefer.

Pinging @pgrete @jdolence @nmsriram @forrestglines @jmstone @AndrewGaspar and any other interested parties.

docs/mesh/packing.md
```diff
@@ -4,29 +4,61 @@
...
 can pass in a metadata vector, a vector of names, and optionally IDs
 and a map from names to indices. See
 [here](../interface/containers.md) for more details.
 
+## Packing over a piece of the mesh
+
+Instead of packing over the whole mesh, you can pack over only a piece
+of it by using the `Partition` machinery found in
+`utils/partition_stl_containers.hpp`. For example, to break the mesh
+into four evenly spaced meshpacks, do
```
AndrewGaspar: evenly sized?

```suggestion
into four evenly sized meshpacks, do
```

src/utils/partition_stl_containers.hpp
```diff
@@ -0,0 +1,83 @@
...
+#ifndef UTILS_PARTITION_STL_CONTAINERS_HPP_
+#define UTILS_PARTITION_STL_CONTAINERS_HPP_
+
+#include <string>
+#include <vector>
+
+#include "error_checking.hpp"
+
+namespace parthenon {
+namespace Partition {
```
AndrewGaspar: lower-case to match root namespace
```suggestion
namespace partition {
```

docs/mesh/packing.md
```diff
@@ -4,29 +4,61 @@
...
+```C++
+using parthenon::MeshBlock;
+using parthenon::Partition::Partition_t;
+Partition_t<MeshBlock> partitions;
+parthenon::Partition::ToNPartitions(mesh->block, 4, partitions);
```
AndrewGaspar: This is supposed to be `block_list`, right?
```suggestion
parthenon::Partition::ToNPartitions(mesh->block_list, 4, partitions);
```

docs/mesh/packing.md
```diff
@@ -4,29 +4,61 @@
...
+MeshBlockPack<VariablePack<Real>> packs[4];
+for (int i = 0; i < partitions.size() {
+  packs[i] = PackVariablesOnMesh(partitions[i], "base");
```
AndrewGaspar, line 46: Probably should demonstrate passing a list of variable names, too?

src/utils/partition_stl_containers.hpp
```diff
@@ -0,0 +1,83 @@
...
+// TODO(JMM): The templated type safety with T* and T may need to be
+// changed if we move to sufficiently general container objects.
+template <typename T>
+using Partition_t = std::vector<std::vector<T *>>;
+
+// x/y rounded up
+// See discussion here for limitations and alternatives
+// https://stackoverflow.com/questions/2745074/fast-ceiling-of-an-integer-division-in-c-c
+KOKKOS_INLINE_FUNCTION
+int IntCeil(int x, int y) {
```
AndrewGaspar, line 33: Put this in an `internal` or `details` namespace.

src/utils/partition_stl_containers.hpp
```diff
@@ -0,0 +1,83 @@
...
+void ToSizeN(Container_t &container, const int N, Partition_t<T> &partitions) {
+  using std::to_string;
+
+  PARTHENON_REQUIRE_THROWS(N > 0, "You must have at least 1 partition");
+  int nelements = container.size();
+  PARTHENON_REQUIRE_THROWS(nelements > 0,
+                           "You must have at least 1 element to partition");
+  std::string msg = ("Cannot partition " + to_string(nelements) +
+                     " elements into partitions of size " + to_string(N) + ".");
+  PARTHENON_REQUIRE_THROWS(nelements >= N, msg);
```
AndrewGaspar: Hm... I don't think I like any of these checks. I think the algorithm otherwise produces perfectly reasonable results for size-0 containers and for containers that are smaller than the partition size.

I think we already have a check that each rank gets at least 1 block, and that's going to give the user a more useful error message. So don't complicate the algorithm and tests for this - just allow passing a container of size 0.

I don't think we should require the first partition to be filled - if it's determined that a specific package benefits from a precise partition size, it shouldn't fail out just because in some instance a partition didn't get filled up. I think what you end up if you do that is everybody guarding their calls to `ToSizeN` with `N=std::min(container.size(), N)`. Consider these two scenarios for a chosen partition size of 32: 1) you have a container size of 31, and so it fails, and 2) you have a container size of 33, which gives you two partitions of size 32 and 1. Arguably 2) is worse because you probably would have been better off with a single partition of 33, but we allow 2) and disallow 1).

example/calculate_pi/calculate_pi.hpp
```diff
@@ -18,25 +18,27 @@
...
-// Task over whole mesh, no packs
-parthenon::TaskStatus RetrieveAreas(std::vector<parthenon::MeshBlock *> &blocks,
-                                    parthenon::Packages_t &packages);
-
-// Run task on the entire mesh at once
-parthenon::TaskStatus ComputeAreaOnMesh(std::vector<parthenon::MeshBlock *> &blocks,
-                                        parthenon::Packages_t &packages);
+// Note pass by value here. Required for capture.
+// All objects here have reference semantics, so capture by value is ok.
+// TODO(JMM) A std::shared_ptr might be better.
```
AndrewGaspar, line 37: Why would `shared_ptr` be better? Aren't the heavy bits of `Pack_t` and `ParArrayHost` kokkos views, which are essentially the same as `shared_ptr`?

AndrewGaspar (changes requested):

src/utils/partition_stl_containers.hpp
AndrewGaspar: (Oh, sorry, the N > 0 check is just fine, it's the rest of them that I'd remove)

src/utils/partition_stl_containers.hpp
Yurlungur: So I was actually unsure about this one. I know we've been using lower-case namespaces... but the reason I created a namespace at all is because it felt like a nice shorthand `Partition::RestOfTheSentence`. And that felt like it should be capitalized because that's how we've been doing it for functions.

That said, I can see how it's a style violation, so I'll change it.

example/calculate_pi/calculate_pi.hpp
Yurlungur: Yeah, that's why I didn't implement a `shared_ptr`. I was just concerned about metadata. If you carry enough metadata around with your pointer to your big heavy array, you'll eventually start to see a cost.

My thinking was don't add a `shared_ptr` now but keep it in mind in case it becomes an issue.

src/utils/partition_stl_containers.hpp
Yurlungur: Ah good point. I'll remove those checks.

Yurlungur:
@AndrewGaspar I think all your requested changes have been implemented.

AndrewGaspar:
Something we'll need to be careful with for regions is that `TaskCollection::Execute` doesn't explicitly barrier, so if there is asynchronous kernel launches still executing at the end of a region, the subsequent region will need to be careful to synchronize on those same streams before, say, calling MPI.

AndrewGaspar (approved):

Yurlungur:
> Something we'll need to be careful with for regions is that `TaskCollection::Execute` doesn't explicitly barrier, so if there is asynchronous kernel launches still executing at the end of a region, the subsequent region will need to be careful to synchronize on those same streams before, say, calling MPI.

We should maybe create a "synchronize streams" task for that. It's probably common enough.

docs/mesh/packing.md
```diff
@@ -4,29 +4,67 @@
...
+parthenon::Partition::ToNPartitions(mesh->block_list, 4, partitions);
+MeshBlockPack<VariablePack<Real>> packs[4];
+for (int i = 0; i < partitions.size() {
+  packs[i] = PackVariablesOnMesh(partitions[i], "base");
+}
+```
+To pack only the variables "var1" and "var2" in the container named "mycontainer", do:
+```C++
+for (int i = 0; i < partitions.size() {
+  packs[i] = PackVariablesOnMesh(partitions[i], "myContainer", {"var1","var2"});
```
pgrete: Does this actually work?
IIRC the compiler I used failed in the template deduction on the strongs, so I had to add `  std::vector<std::string>` (which is also what happens below in the code)

pgrete (approved):
Looks good!
And regarding your question on the syntactic sugar vs the current manual implementation, I think we should just keep it as it is right now and potentially update/add base on our downstream code implementation experience.

docs/mesh/packing.md
Yurlungur: Good catch. I'll fix that.

PR #300, opened 2020-09-08
Change AddTask signature slightly

AndrewGaspar:
## PR Summary

- Takes dependency task by const ref
- Makes the member function pointer signature more explicit
- Adds a test for task_list
- Implements member function AddTask using the generic AddTask
- Tries to forward arguments better to reduce copies

The original purpose of this PR was to make it possible to use move-only objects in a Task function, but the `std::function` at the root of the `Task` object throws a wrench into that. Nonetheless, did some clean-up of some things I found along the way.

The reason this came up is that I needed to implement clean-up of the xRage-side task objects.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/src/tasks/task_list.hpp
+++ b/src/tasks/task_list.hpp
@@ -73,7 +73,7 @@ class TaskList {
   TaskID AddTask(F &&func, TaskID const &dep, Args &&... args) {
     TaskID id(tasks_added_ + 1);
     task_list_.push_back(
-        Task(id, dep, [=, func{std::forward<F>(func)}]() mutable -> TaskStatus {
+        Task(id, dep, [=, func = std::forward<F>(func)]() mutable -> TaskStatus {
           return func(args...);
         }));
     tasks_added_++;
--- a/src/tasks/task_list.hpp
+++ b/src/tasks/task_list.hpp
@@ -70,23 +70,23 @@ class TaskList {
   }
 
   template <class F, class... Args>
-  TaskID AddTask(F func, TaskID &dep, Args &&... args) {
+  TaskID AddTask(F &&func, TaskID const &dep, Args &&... args) {
     TaskID id(tasks_added_ + 1);
     task_list_.push_back(
-        Task(id, dep, [=]() mutable -> TaskStatus { return func(args...); }));
+        Task(id, dep, [=, func{std::forward<F>(func)}]() mutable -> TaskStatus {
+          return func(args...);
+        }));
     tasks_added_++;
     return id;
   }
 
   // overload to add member functions of class T to task list
   // NOTE: we must capture the object pointer
-  template <class F, class T, class... Args>
-  TaskID AddTask(F func, T *obj, TaskID &dep, Args &&... args) {
-    TaskID id(tasks_added_ + 1);
-    task_list_.push_back(
-        Task(id, dep, [=]() mutable -> TaskStatus { return (obj->*func)(args...); }));
-    tasks_added_++;
-    return id;
+  template <class T, class... Args>
+  TaskID AddTask(TaskStatus (T::*func)(Args...), T *obj, TaskID const &dep,
+                 Args &&... args) {
+    return this->AddTask([=]() mutable -> TaskStatus { return (obj->*func)(args...); },
+                         dep);
   }
 
   void Print() {
--- a/src/tasks/task_list.hpp
+++ b/src/tasks/task_list.hpp
@@ -70,7 +70,7 @@ class TaskList {
   }
 
   template <class F, class... Args>
-  TaskID AddTask(F &&func, TaskID const &dep, Args &&... args) {
+  TaskID AddTask(TaskID const &dep, F &&func, Args &&... args) {
     TaskID id(tasks_added_ + 1);
     task_list_.push_back(
         Task(id, dep, [=, func = std::forward<F>(func)]() mutable -> TaskStatus {
@@ -83,10 +83,10 @@ class TaskList {
   // overload to add member functions of class T to task list
   // NOTE: we must capture the object pointer
   template <class T, class... Args>
-  TaskID AddTask(TaskStatus (T::*func)(Args...), T *obj, TaskID const &dep,
+  TaskID AddTask(TaskID const &dep, TaskStatus (T::*func)(Args...), T *obj,
                  Args &&... args) {
-    return this->AddTask([=]() mutable -> TaskStatus { return (obj->*func)(args...); },
-                         dep);
+    return this->AddTask(dep,
+                         [=]() mutable -> TaskStatus { return (obj->*func)(args...); });
   }
 
   void Print() {
--- a/tst/unit/test_tasklist.cpp
+++ b/tst/unit/test_tasklist.cpp
@@ -11,34 +11,40 @@
 // the public, perform publicly and display publicly, and to permit others to do so.
 //========================================================================================
 
-#ifndef TASKS_TASK_TYPES_HPP_
-#define TASKS_TASK_TYPES_HPP_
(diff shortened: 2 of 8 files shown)


src/tasks/task_list.hpp
```diff
@@ -70,23 +70,23 @@ class TaskList {
   }
 
   template <class F, class... Args>
-  TaskID AddTask(F func, TaskID &dep, Args &&... args) {
+  TaskID AddTask(F &&func, TaskID const &dep, Args &&... args) {
     TaskID id(tasks_added_ + 1);
     task_list_.push_back(
-        Task(id, dep, [=]() mutable -> TaskStatus { return func(args...); }));
+        Task(id, dep, [=, func{std::forward<F>(func)}]() mutable -> TaskStatus {
```
Yurlungur: What is this line doing? I've never seen a `std::forward` applied to a closure like this.

src/tasks/task_list.hpp
```diff
@@ -70,23 +70,23 @@ class TaskList {
...
   // NOTE: we must capture the object pointer
-  template <class F, class T, class... Args>
-  TaskID AddTask(F func, T *obj, TaskID &dep, Args &&... args) {
-    TaskID id(tasks_added_ + 1);
-    task_list_.push_back(
-        Task(id, dep, [=]() mutable -> TaskStatus { return (obj->*func)(args...); }));
-    tasks_added_++;
-    return id;
+  template <class T, class... Args>
+  TaskID AddTask(TaskStatus (T::*func)(Args...), T *obj, TaskID const &dep,
```
Yurlungur: I assume this more explicit typing is safer?

src/tasks/task_list.hpp
```diff
@@ -70,23 +70,23 @@ class TaskList {
...
+                 Args &&... args) {
+    return this->AddTask([=]() mutable -> TaskStatus { return (obj->*func)(args...); },
+                         dep);
```
Yurlungur: Good improvement here, calling `AddTask` instead of explicitly re-implementing it.

tst/unit/test_tasklist.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+// Internal Includes
+#include "basic_types.hpp"
+#include "tasks/task_list.hpp"
+
+using parthenon::TaskID;
+using parthenon::TaskList;
+using parthenon::TaskStatus;
+
+TEST_CASE("Task Object Lifecycle", "[TaskList][AddTask]") {
```
Yurlungur, line 28: So this test checks that tasks and objects within them are garbage collected appropriately. That's good. But it strikes me as an odd choice of a thing to test. I don't object, but can you explain your reasoning?

Yurlungur (approved):
No objections to these changes. Just a few questions.

tst/unit/test_tasklist.cpp
AndrewGaspar: Oh, the only reason I added this test was because of the original purpose of my changes before I realized it was going to be more work than I expected. I can delete it if it's not useful.

src/tasks/task_list.hpp
AndrewGaspar: Yeah, but more-so will produce fewer template instantiations. Hopefully produces better error messages, too, if you don't return the correct type.

src/tasks/task_list.hpp
AndrewGaspar: Basically, if you pass a lambda directly to `AddTask` (i.e. rvalue-ref), it will move it instead of copy it. It will still copy those lambdas that are passed by lvalue ref. Again, originally this was to support adding tasks with move only objects in it, but then I ran into the `std::function` wall. Still, this is strictly "better" without removing anything that was previously supported.

tst/unit/test_tasklist.cpp
Yurlungur: Nah, leave it. It's better than nothing. And if this is something you're going to have to tackle eventually, might as well get the task in now.

src/tasks/task_list.hpp
Yurlungur: Huh. I see. That's subtle.

tst/unit/test_tasklist.cpp
AndrewGaspar: Oh, I'll probably stick a `std::shared_ptr` in the middle and call it good so I can still copy my move-only state...

AndrewGaspar:
@pgrete This is ready to go if you could review.

pgrete (approved):
Looks good.
Two minor comments:
- The changelog entry should be moved up now as this will be part of the next release
- It'd be great if we could reorder to arguments when updating the signature anyway. To me, it's kind of unintuitive to have the dependencies for one type of function as the second argument and for the other type of function as last argument. What do you think?

AndrewGaspar:
> To me, it's kind of unintuitive to have the dependencies for one type of function as the second argument and for the other type of function as last argument.

Well, technically it's the last non-optional argument to both `AddTask` overloads, but I'd think it would read better if it was the first argument.

In my opinion, if we're going to go down the route of changing the signature more dramatically, we don't really need the variadic args list. If you want to capture some additional context, you should just pass a lambda. And once you've done that, you can get rid of the member function overload, too.

@jdolence is there a reason to have the argument capture rather than just using a lambda at the call site (which can then capture the arguments)? Is it because it makes adding a task more terse?

jdolence:
@AndrewGaspar @pgrete I would not be in favor of using a lambda at the call site.  There's obviously nothing wrong with it technically, but I think it would make the code far less readable.

As far as moving the dependencies arg, I guess I'm OK with that, but I still prefer how it is now.  Seems most natural to me to have the first arg to `AddTask` be the actual task.  The additional argument required when the task is a non-static member function is a necessary evil.  As @AndrewGaspar points out, though, the dependencies arg really just follows the complete specification of the function, which requires one arg if it's a free floating or statically declared function in a class, or two args if it's a non-static member function.  So, at least conceptually, it's always the second thing that needs to be specified.

On the other hand, I guess an argument in favor of moving the dependency arg to the first slot is that it would mean the remaining args specify the function and the function's args, rather than strangely separating them with the dependency arg between them.  That does seem nice.

I'm OK either way, so do what you think is best.

pgrete:
I'm in favor of moving the dependencies to the front and also not using lambdas at the call site as the user facing interface (personally) looks cleaner without a lambda.

My main motivation to have a "fixed" position of the dependencies is nicely illustrated in two lines of the advection example:
```

  auto start_recv = tl.AddTask(&Container<Real>::StartReceiving, sc1.get(), none,
                               BoundaryCommSubset::all);

  auto advect_flux = tl.AddTask(advection_package::CalculateFluxes, none, sc0);
```

Just looking at how I should use AddTask from this code (not looking in the code of AddTask or our documention) it's not entirely clear why in one case it "looks" like the dependency is following the "container"  obtained by a a ".get()` and in the other case the dependency precedes the "container".

Yurlungur:
I tend to prefer using lambdas at the call site. But I think allowing users to not have to do that is worthwhile.

AndrewGaspar:
> On the other hand, I guess an argument in favor of moving the dependency arg to the first slot is that it would mean the remaining args specify the function and the function's args, rather than strangely separating them with the dependency arg between them. That does seem nice.

Yeah, that's the main reason I think the dependency should go in front - otherwise it is kind of odd that it just falls in the middle of the function signature.

AndrewGaspar:
Sounds like we're ok moving the dependency ID to the front, so I think I'll do that.

pgrete:
Thanks for the update. Shall we pull the trigger?

PR #303, opened 2020-09-12
make block_list a std::vector instead of a std::list

### LINKED ISSUE #211: Improve FindMeshBlock performance
See: https://github.com/PrincetonUniversity/athena-public-version/issues/32

Yurlungur:
## PR Summary

As we have known for some time (see issue #211), `FindMeshBlock` is slow when there are a large number of meshblocks per core. As @pgrete discovered today, it appears this may be a significant blocking factor when meshblocks are small (i.e., around 16^3 or less). This is an attempt to fix the issue. 

I replace the
```C++
std::list<MeshBlock> block_list;
```
in `Mesh` with
```C++
using BlockList_t = std::vector<std::shared_ptr<MeshBlock>>;
BlockList_t block_list;
```
I choose to use a `std::vector` of shared pointers, rather than a `vector` of `MeshBlock`s, for a couple reasons:
1. This avoids the issue @jmstone raised about pointers of objects within a vector not being reliable, because when the vector is appended to, it is re-allocated and the data copied.
2. This avoids the gotcha of attempting to copy meshblocks. The naive thing to do is to copy the shared pointer, which will work fine.
3. This simplifies the logic and typing in `partition_stl_containers.hpp` and `meshblock_pack.hpp`, and saves some ampersands floating around.

I'm not married to this choice of course. If people feel differently, please speak up.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/example/calculate_pi/pi_driver.cpp
+++ b/example/calculate_pi/pi_driver.cpp
@@ -130,7 +130,8 @@ TaskCollection PiDriver::MakeTasks(T &blocks) {
   int pack_size = pinput->GetOrAddInteger("Pi", "pack_size", 1);
   if (pack_size < 1) pack_size = blocks.size();
 
-  partition::Partition_t<MeshBlock> partitions;
+  // partition::Partition_t<MeshBlock> partitions;
+  std::vector<BlockList_t> partitions;
   partition::ToSizeN(blocks, pack_size, partitions);
   ParArrayHost<Real> areas("areas", partitions.size());
 
--- a/example/calculate_pi/pi_driver.cpp
+++ b/example/calculate_pi/pi_driver.cpp
@@ -130,7 +130,6 @@ TaskCollection PiDriver::MakeTasks(T &blocks) {
   int pack_size = pinput->GetOrAddInteger("Pi", "pack_size", 1);
   if (pack_size < 1) pack_size = blocks.size();
 
-  // partition::Partition_t<MeshBlock> partitions;
   std::vector<BlockList_t> partitions;
   partition::ToSizeN(blocks, pack_size, partitions);
   ParArrayHost<Real> areas("areas", partitions.size());
--- a/src/mesh/mesh.cpp
+++ b/src/mesh/mesh.cpp
@@ -1122,8 +1122,8 @@ void Mesh::Initialize(int res_flag, ParameterInput *pin, ApplicationInput *app_i
 /// the search at.
 std::shared_ptr<MeshBlock> Mesh::FindMeshBlock(int tgid) {
   // Attempt to simply index into the block list.
-  int nbs = block_list[0]->gid;
-  int i = tgid - nbs;
+  const int nbs = block_list[0]->gid;
+  const int i = tgid - nbs;
   PARTHENON_DEBUG_REQUIRE(0 <= i && i < block_list.size(),
                           "MeshBlock local index out of bounds.");
   PARTHENON_DEBUG_REQUIRE(block_list[i]->gid == tgid, "MeshBlock not found!");
--- a/src/mesh/mesh.cpp
+++ b/src/mesh/mesh.cpp
@@ -1008,7 +1008,7 @@ void Mesh::Initialize(int res_flag, ParameterInput *pin, ApplicationInput *app_i
 #pragma omp parallel for num_threads(nthreads)
       for (int i = 0; i < nmb; ++i) {
         auto &pmb = block_list[i];
-        pmb->ProblemGenerator(pmb, pin);
+        pmb->ProblemGenerator(pmb.get(), pin);
       }
     }
 
--- a/src/mesh/mesh.cpp
+++ b/src/mesh/mesh.cpp
@@ -471,13 +471,16 @@ Mesh::Mesh(ParameterInput *pin, ApplicationInput *app_in, Properties_t &properti
   int nbs = nslist[Globals::my_rank];
   int nbe = nbs + nblist[Globals::my_rank] - 1;
   // create MeshBlock list for this process
+  block_list.clear();
+  block_list.resize(nbe - nbs + 1);
   for (int i = nbs; i <= nbe; i++) {
     SetBlockSizeAndBoundaries(loclist[i], block_size, block_bcs);
     // create a block and add into the link list
-    block_list.push_back(std::make_shared<MeshBlock>(i, i - nbs, loclist[i], block_size,
-                                                     block_bcs, this, pin, app_in,
-                                                     properties, packages, gflag));
-    block_list.back()->pbval->SearchAndSetNeighbors(tree, ranklist.data(), nslist.data());
+    block_list[i - nbs] =
+        std::make_shared<MeshBlock>(i, i - nbs, loclist[i], block_size, block_bcs, this,
+                                    pin, app_in, properties, packages, gflag);
+    block_list[i - nbs]->pbval->SearchAndSetNeighbors(tree, ranklist.data(),
+                                                      nslist.data());
   }
 
   ResetLoadBalanceVariables();
@@ -727,6 +730,8 @@ Mesh::Mesh(ParameterInput *pin, ApplicationInput *app_in, RestartReader &rr,
   auto xmin = rr.ReadDataset<double>("/Blocks/xmin");
 
   // Create MeshBlocks (parallel)
+  block_list.clear();
+  block_list.resize(nbe - nbs + 1);
   for (int i = nbs; i <= nbe; i++) {
     for (auto &v : block_bcs) {
       v = parthenon::BoundaryFlag::undef;
@@ -734,10 +739,11 @@ Mesh::Mesh(ParameterInput *pin, ApplicationInput *app_in, RestartReader &rr,
     SetBlockSizeAndBoundaries(loclist[i], block_size, block_bcs);
 
     // create a block and add into the link list
-    block_list.push_back(std::make_shared<MeshBlock>(
+    block_list[i - nbs] = std::make_shared<MeshBlock>(
         i, i - nbs, this, pin, app_in, properties, packages, loclist[i], block_size,
-        block_bcs, costlist[i], gflag));
-    block_list.back()->pbval->SearchAndSetNeighbors(tree, ranklist.data(), nslist.data());
+        block_bcs, costlist[i], gflag);
+    block_list[i - nbs]->pbval->SearchAndSetNeighbors(tree, ranklist.data(),
+                                                      nslist.data());
   }
 
   ResetLoadBalanceVariables();
@@ -1002,8 +1008,9 @@ void Mesh::Initialize(int res_flag, ParameterInput *pin, ApplicationInput *app_i
 #ifdef OPENMP_PARALLEL
   int nthreads = GetNumMeshThreads();
 #endif
-  int nmb = GetNumMeshBlocksThisRank(Globals::my_rank);
   do {
+    int nmb = GetNumMeshBlocksThisRank(Globals::my_rank);
+
     if (res_flag == 0) {
 #pragma omp parallel for num_threads(nthreads)
       for (int i = 0; i < nmb; ++i) {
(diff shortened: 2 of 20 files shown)


Yurlungur:
I think this PR is ready for review, assuming tests pass. A couple of things:
1. @jmstone I'd appreciate your input on this. And whether you agree that this approach is reasonable. Now having looked at the Athena version, I think I basically copied your approach, but with smart pointers instead of raw ones and a std::vector instead of an athena array.
2. @pgrete, @gshipman performance of FindMeshBlock should be a lot better now. I haven't had a chance to do a comparison yet, but if you get to a performance test before I do, I'd be interested to see how it looks.
3. This shouldn't go in until you merge in @AndrewGaspar 's PR #304 as I base my changelog on the version 0.2.0 release.

Yurlungur:
Some performance number updates. On a simple test with the advection problem, with AMR active, on GPU, I'm seeing about a 50% speedup for 16^3 meshblocks and a 3x speedup for 8^3 meshblocks, with a 128^3 grid.
image
Cycle =1 includes a remeshing step. Cycle=3 does not.

jmstone (approved):
Looks good to me.  As you say, it replicates the approach taken in the current master branch of Athena++.  Since the Z-ordering used in the binary tree in Athena++ guarantees that MeshBlock IDs within a given MPI rank are in sequential order, then it makes finding a MeshBlock based on the vector (or AthenaArray) index very simple and essentially free.

AndrewGaspar (changes requested):
I don't like that we now have to dereference mesh block all over the place. I think I can come up with a fix. Give me a hot second and I'll propose a fix.

src/mesh/mesh.cpp
```diff
@@ -1121,10 +1120,14 @@ void Mesh::Initialize(int res_flag, ParameterInput *pin, ApplicationInput *app_i
...
 /// Finds location of a block with ID `tgid`. Can provide an optional "hint" to start
 /// the search at.
-std::list<MeshBlock>::iterator Mesh::FindMeshBlock(int tgid) {
-  // search the rest of the list
-  return std::find_if(block_list.begin(), block_list.end(),
-                      [tgid](MeshBlock const &bl) { return bl.gid == tgid; });
+std::shared_ptr<MeshBlock> Mesh::FindMeshBlock(int tgid) {
+  // Attempt to simply index into the block list.
+  int nbs = block_list[0]->gid;
+  int i = tgid - nbs;
```
pgrete:
```suggestion
  const int nbs = block_list[0]->gid;
  const int i = tgid - nbs;
```

src/mesh/mesh.hpp
```diff
@@ -23,7 +23,7 @@
 
 #include <cstdint>
 #include <functional>
-#include <list>
+// #include <list>
```
pgrete:
```suggestion
```

src/mesh/mesh.hpp
```diff
@@ -341,10 +342,15 @@ class Mesh {
...
   int GetNumMeshThreads() const { return num_mesh_threads_; }
   std::int64_t GetTotalCells() {
-    auto &mb = block_list.front();
-    return static_cast<std::int64_t>(nbtotal) * mb.block_size.nx1 * mb.block_size.nx2 *
-           mb.block_size.nx3;
+    auto &pmb = block_list.front();
+    return static_cast<std::int64_t>(nbtotal) * pmb->block_size.nx1 *
+           pmb->block_size.nx2 * pmb->block_size.nx3;
   }
+  // TODO(JMM): Move block_size into mesh.
```
pgrete, line 348: Is that a todo for later then I suggest to open an issue to keep track or a todo for this PR?

src/utils/partition_stl_containers.hpp
```diff
@@ -20,10 +20,11 @@
 
 namespace parthenon {
 namespace partition {
-// TODO(JMM): The templated type safety with T* and T may need to be
-// changed if we move to sufficiently general container objects.
+// Note the interface here has objects in partition are now
+// COPIED, not pointed at.
+// Change is due to using std::shared_ptr<MeshBlock>
```
pgrete, line 25: I now read the comment as: All elements (e.g., variables of a `Container`) are not copied into the Partition.
Is this correct?

example/calculate_pi/pi_driver.cpp
```diff
@@ -130,7 +130,8 @@ TaskCollection PiDriver::MakeTasks(T &blocks) {
   int pack_size = pinput->GetOrAddInteger("Pi", "pack_size", 1);
   if (pack_size < 1) pack_size = blocks.size();
 
-  partition::Partition_t<MeshBlock> partitions;
+  // partition::Partition_t<MeshBlock> partitions;
```
pgrete:
```suggestion
```

Yurlungur:
> I don't like that we now have to dereference mesh block all over the place. I think I can come up with a fix. Give me a hot second and I'll propose a fix.

What's wrong with dereferencing the mesh block? It is a pointer after all.

src/mesh/mesh.hpp
Yurlungur: Didn't want to handle it in this PR I think it's slightly bigger surgery. I'll open an issue.

src/utils/partition_stl_containers.hpp
Yurlungur: The partitioner now copies the element of the stl container into the partition. But the element is a pointer. Before, I had to access the pointer to the objects to avoid unnecessary copies.

pgrete: Gotcha!

src/mesh/mesh.hpp
Yurlungur: Done. #305

AndrewGaspar:
> What's wrong with dereferencing the mesh block? It is a pointer after all.

The additional syntax + the additional null state (you have to now think about whether the block can be null, even though it, of course, cannot be). All we really want here is to add the pointer-stability property and the O(1) access property - no reason to give up the certainty that references give you. Should be pretty easy to fix - and undo most of your changes to switch mb's to pmb's.

Yurlungur:
> The additional syntax + the additional null state (you have to now think about whether the block can be null, even though it, of course, cannot be). All we really want here is to add the pointer-stability property and the O(1) access property - no reason to give up the certainty that references give you. Should be pretty easy to fix - and undo most of your changes to switch mb's to pmb's.

You're gonna have to help me here. I still don't understand. The syntax seems to be the difference between `.` and `->`. Those seem equivalent to me.

And a reference can be null too. If you let a reference go out of scope, awful things happen. That's one reason why I like the smart pointer solution. It's clear what we're doing.

AndrewGaspar:
I've decided not to pursue the alternative - I always forget how much code it takes to implement a standard collection. I was essentially going to add a `std::vector`-like container that internally stored everything as `std::vector<std::unique_ptr<T>>`. The container would make it impossible to insert a null object, and you get stable references. I wish something like this was included in the STL, but alas.

-------

> And a reference can be null too.

As far as the standard is concerned, there is no such thing as a null reference. In order to get a "null" reference as you may imagine it would require you to "dereference" a null pointer, which is illegal. Of course, it's trivial to create a reference to an "object" located at "nullptr", but it's technically illegal from the moment you "dereference" the pointer to get the reference.

> That's one reason I like the smart pointer solution. It's clear what we're doing.

On that... I don't think `shared_ptr` is the right smart pointer here. `shared_ptr` is for when you have an object of indeterminate ownership. We have no such issue - only a `Mesh` should own `MeshBlock`s, therefore we should be using `unique_ptr`. Granted `shared_ptr` _can_ have better asserty-ness around lifetime validity, but to get that you need to go and replace all the long-lived raw pointers to MeshBlock with `weak_ptr`. See: `MeshRefinement`, `Container`, `BoundaryValues`, `BoundaryVariable`, `Reconstruction` - all of them have back pointers to the MeshBlock. Otherwise, the `shared_ptr` doesn't do anything `unique_ptr` doesn't since we don't have any other long-lived pointers to the `MeshBlock`, as far as I know.

You'll also want to be careful with `shared_ptr` in general in our code... because we love our back-pointers, it's easy to imagine somebody adding a shared_ptr loop and start leaking memory (though, honestly, we really aught to create a test for this).

AndrewGaspar (approved):
I won't make you do it, but I would really prefer you either change `shared_ptr<MeshBlock>` to `unique_ptr<MeshBlock>` to reflect that MeshBlocks are unique owned by the Mesh, or make the long-lived pointers to MeshBlock `weak_ptr<MeshBlock>` in `MeshRefinement`, `Container`, `BoundaryValues`, `BoundaryVariable`, and `Reconstruction` to at least get the purported benefit of `shared_ptr`.

Yurlungur:
> I've decided not to pursue the alternative - I always forget how much code it takes to implement a standard collection. I was essentially going to add a `std::vector`-like container that internally stored everything as `std::vector<std::unique_ptr<T>>`. The container would make it impossible to insert a null object, and you get stable references. I wish something like this was included in the STL, but alas.

Ah I see. That's a cool idea.
 
> On that... I don't think `shared_ptr` is the right smart pointer here. `shared_ptr` is for when you have an object of indeterminate ownership. We have no such issue - only a `Mesh` should own `MeshBlock`s, therefore we should be using `unique_ptr`. Granted `shared_ptr` _can_ have better asserty-ness around lifetime validity, but to get that you need to go and replace all the long-lived raw pointers to MeshBlock with `weak_ptr`. See: `MeshRefinement`, `Container`, `BoundaryValues`, `BoundaryVariable`, `Reconstruction` - all of them have back pointers to the MeshBlock. Otherwise, the `shared_ptr` doesn't do anything `unique_ptr` doesn't since we don't have any other long-lived pointers to the `MeshBlock`, as far as I know.

Is that entirely true? There's a spot in `amr_loadbalance.cpp` where I take advantage of the pointer being shared. When I construct the new block list, it's important that the old block list remain unchanged. And so for a moment, the block is pointed to by two separate pointers:
```C++
// Step 7. construct a new MeshBlock list (moving the data within the MPI rank)
  {

    BlockList_t new_block_list(nbe - nbs + 1);
    for (int n = nbs; n <= nbe; n++) {
      int on = newtoold[n];
      if ((ranklist[on] == Globals::my_rank) && (loclist[on].level == newloc[n].level)) {
        // on the same MPI rank and same level -> just move it
        new_block_list[n - nbs] = FindMeshBlock(on);
      } else {
        // on a different refinement level or MPI rank - create a new block
        BoundaryFlag block_bcs[6];
        SetBlockSizeAndBoundaries(newloc[n], block_size, block_bcs);
        // append new block to list of MeshBlocks
        new_block_list[n - nbs] = std::make_shared<MeshBlock>(
            n, n - nbs, newloc[n], block_size, block_bcs, this, pin, app_in, properties,
            packages, gflag, true);
        // fill the conservative variables
        if ((loclist[on].level > newloc[n].level)) { // fine to coarse (f2c)
          for (int ll = 0; ll < nleaf; ll++) {
            if (ranklist[on + ll] != Globals::my_rank) continue;
            // fine to coarse on the same MPI rank (different AMR level) - restriction
            auto pob = FindMeshBlock(on + ll);
            FillSameRankFineToCoarseAMR(pob.get(), new_block_list[n - nbs].get(),
                                        loclist[on + ll]);
          }
        } else if ((loclist[on].level < newloc[n].level) && // coarse to fine (c2f)
                   (ranklist[on] == Globals::my_rank)) {
          // coarse to fine on the same MPI rank (different AMR level) - prolongation
          auto pob = FindMeshBlock(on);
          FillSameRankCoarseToFineAMR(pob.get(), new_block_list[n - nbs].get(),
                                      newloc[n]);
        }
        ApplyBoundaryConditions(new_block_list[n - nbs]->real_containers.Get());
        FillDerivedVariables::FillDerived(new_block_list[n - nbs]->real_containers.Get());
      }
    }

    // Replace the MeshBlock list
    block_list = std::move(new_block_list);

    // Ensure local and global ids are correct
    for (int n = nbs; n <= nbe; n++) {
      block_list[n - nbs]->gid = n;
      block_list[n - nbs]->lid = n - nbs;
    }
  }
```
This was by far the trickiest thing to get right. The shared pointer solution avoids a lot of additional book-keeping.

> You'll also want to be careful with `shared_ptr` in general in our code... because we love our back-pointers, it's easy to imagine somebody adding a shared_ptr loop and start leaking memory (though, honestly, we really aught to create a test for this).

This is a good point.

> or make the long-lived pointers to MeshBlock weak_ptr<MeshBlock> in MeshRefinement, Container, BoundaryValues, BoundaryVariable, and Reconstruction to at least get the purported benefit of shared_ptr.

This is worth doing. Mind if I do it in a separate PR? I'll make an issue and pursue it next.

Yurlungur:
I will pursue the changes suggested by @pgrete and @AndrewGaspar in separate PRs, marked by issues #305 and #306 respectively. I'm going to merge this in for now and get started on those. Thanks for the feedback, all!

AndrewGaspar:
> Is that entirely true? There's a spot in amr_loadbalance.cpp where I take advantage of the pointer being shared. When I construct the new block list, it's important that the old block list remain unchanged. And so for a moment, the block is pointed to by two separate pointers:

I don't think you'd need to do it that way - you could just move the object from the old list to the new list since it doesn't need to be in the old list anymore.

Yurlungur:
> I don't think you'd need to do it that way - you could just move the object from the old list to the new list since it doesn't need to be in the old list anymore.

That was the first thing I tried, but it messed up the stability of `FindMeshBlock`. You could do it, but you'd need more book-keeping.

Yurlungur:
> > I don't think you'd need to do it that way - you could just move the object from the old list to the new list since it doesn't need to be in the old list anymore.
> 
> That was the first thing I tried, but it messed up the stability of `FindMeshBlock`. You could do it, but you'd need more book-keeping.

The book-keeping approach is what's done in Athena++. So obviously that's an option. But I think my solution is cleaner.

AndrewGaspar:
Hm... I mean literally `std::move` - so you'd leave the "slot" in the old list in-place, but with a moved-from pointer. I think that should be essentially the same as what we were doing with `splice` in the std::list before...

Yurlungur:
> Hm... I mean literally `std::move` - so you'd leave the "slot" in the old list in-place, but with a moved-from pointer. I think that should be essentially the same as what we were doing with `splice` in the std::list before...

Yeah that's the first thing I tried and it won't work as designed, because `FindMeshBlock` uses the local id of the first meshblock in the list. If the pointer is moved, that reference is invalid. You also can't use the variable `nbs` because it points to the local id for the first meshblock in the *new* list, not the old one. 

So we'd have to introduce another variable for the mesh which references the local id of the first meshblock in the list and update it accordingly. That could be done, of course. But like I said, it's a little additional book-keeping.

AndrewGaspar:
> Yeah that's the first thing I tried and it won't work as designed, because FindMeshBlock uses the local id of the first meshblock in the list.


Aaaah. Well, that's pretty easy to fix, but yeah, doesn't matter here since we're using `shared_ptr`.

PR #307, opened 2020-09-18
MeshBlock Cleanup and Smart Backpointers. Resolves #306.

### LINKED ISSUE #306: Back pointers to meshblocks should be shared pointers.
Since MR #303 makes the meshblock list own shared pointers, we can now take advantage of this with our back pointers, as @AndrewGaspar points out.

Yurlungur:
## PR Summary

As discussed in #306 and #303, smart pointers to meshblocks means we can use them elsewhere in the code, for example, with weak back pointers. This Implements that change.

In addition, I substantially clean up the MeshBlock class. I noticed that all three constructors had the same functionality copy-pasted. I therefore removed all but one of them. I also changed the constructors to a factory method. This was necessary because of a subtlety with smart pointers. You cannot create a weak pointer to an object in its own constructor.

## PR Checklist

- [x] Code passes cpplint
- [X] Adds a test for any bugs fixed. Adds tests for new features.
- [X] Code is formatted

### DIFF
--- a/example/advection/advection_driver.cpp
+++ b/example/advection/advection_driver.cpp
@@ -124,7 +124,7 @@ TaskList AdvectionDriver::MakeTaskList(MeshBlock *pmb, int stage) {
   if (stage == integrator->nstages) {
     auto new_dt = tl.AddTask(
         [](std::shared_ptr<Container<Real>> &rc) {
-          MeshBlock *pmb = rc->pmy_block;
+          auto pmb = rc->pmy_block.lock();
           pmb->SetBlockTimestep(parthenon::Update::EstimateTimestep(rc));
           return TaskStatus::complete;
         },
--- a/example/advection/advection_driver.cpp
+++ b/example/advection/advection_driver.cpp
@@ -124,7 +124,7 @@ TaskList AdvectionDriver::MakeTaskList(MeshBlock *pmb, int stage) {
   if (stage == integrator->nstages) {
     auto new_dt = tl.AddTask(
         [](std::shared_ptr<Container<Real>> &rc) {
-          auto pmb = rc->pmy_block.lock();
+          auto pmb = rc->GetBlockPointer();
           pmb->SetBlockTimestep(parthenon::Update::EstimateTimestep(rc));
           return TaskStatus::complete;
         },
--- a/src/interface/container.hpp
+++ b/src/interface/container.hpp
@@ -45,7 +45,7 @@ class Container {
   //-----------------
   // Public Variables
   //-----------------
-  MeshBlock *pmy_block = nullptr; // ptr to MeshBlock
+  std::weak_ptr<MeshBlock> pmy_block;
 
   //-----------------
   // Public Methods
@@ -71,7 +71,7 @@ class Container {
 
   ///
   /// Set the pointer to the mesh block for this container
-  void setBlock(MeshBlock *pmb) { pmy_block = pmb; }
+  void setBlock(std::weak_ptr<MeshBlock> pmb) { pmy_block = pmb; }
 
   ///
   /// Allocate and add a variable<T> to the container
--- a/src/interface/container.hpp
+++ b/src/interface/container.hpp
@@ -22,6 +22,7 @@
 #include "interface/sparse_variable.hpp"
 #include "interface/variable.hpp"
 #include "interface/variable_pack.hpp"
+#include "utils/error_checking.hpp"
 
 namespace parthenon {
 
@@ -43,11 +44,6 @@ template <typename T>
 class Container {
  public:
   //-----------------
-  // Public Variables
-  //-----------------
-  std::weak_ptr<MeshBlock> pmy_block;
-
-  //-----------------
   // Public Methods
   //-----------------
   /// Constructor
@@ -59,6 +55,14 @@ class Container {
   Container<T>(const Container<T> &src, const std::vector<std::string> &names,
                const std::vector<int> sparse_ids = {});
   Container<T>(const Container<T> &src, const std::vector<MetadataFlag> &flags);
+
+  /// Returns shared pointer to a block
+  std::shared_ptr<MeshBlock> GetBlockPointer() {
+    if (pmy_block.expired()) {
+      PARTHENON_THROW("Invalid pointer to MeshBlock!");
+    }
+    return pmy_block.lock();
+  }
 
   /// We can initialize a container with slices from a different
   /// container.  For variables that have the sparse tag, this will
@@ -72,6 +76,9 @@ class Container {
   ///
   /// Set the pointer to the mesh block for this container
   void setBlock(std::weak_ptr<MeshBlock> pmb) { pmy_block = pmb; }
+  void setBlock(const std::shared_ptr<Container<T>> &other) {
+    pmy_block = other->GetBlockPointer();
+  }
 
   ///
   /// Allocate and add a variable<T> to the container
@@ -304,6 +311,7 @@ class Container {
 
  private:
   int debug = 0;
+  std::weak_ptr<MeshBlock> pmy_block;
 
   CellVariableVector<T> varVector_; ///< the saved variable array
   FaceVector<T> faceVector_;        ///< the saved face arrays
--- a/src/interface/container.hpp
+++ b/src/interface/container.hpp
@@ -75,8 +75,8 @@ class Container {
 
   ///
   /// Set the pointer to the mesh block for this container
-  void setBlock(std::weak_ptr<MeshBlock> pmb) { pmy_block = pmb; }
-  void setBlock(const std::shared_ptr<Container<T>> &other) {
+  void SetBlockPointer(std::weak_ptr<MeshBlock> pmb) { pmy_block = pmb; }
+  void SetBlockPointer(const std::shared_ptr<Container<T>> &other) {
     pmy_block = other->GetBlockPointer();
   }
 
--- a/src/mesh/amr_loadbalance.cpp
+++ b/src/mesh/amr_loadbalance.cpp
@@ -603,9 +603,9 @@ void Mesh::RedistributeAndRefineMeshBlocks(ParameterInput *pin, ApplicationInput
         BoundaryFlag block_bcs[6];
         SetBlockSizeAndBoundaries(newloc[n], block_size, block_bcs);
         // append new block to list of MeshBlocks
-        new_block_list[n - nbs] = std::make_shared<MeshBlock>(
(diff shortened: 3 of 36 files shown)


src/mesh/amr_loadbalance.cpp
```diff
@@ -753,8 +754,8 @@ void Mesh::PrepareSendSameLevel(MeshBlock *pmb, ParArray1D<Real> &sendbuf) {
 
 // step 6, branch 2 (c2f: just pack+send)
 
-void Mesh::PrepareSendCoarseToFineAMR(MeshBlock *pb, ParArray1D<Real> &sendbuf,
-                                      LogicalLocation &lloc) {
+void Mesh::PrepareSendCoarseToFineAMR(std::shared_ptr<MeshBlock> pb,
```
AndrewGaspar: Applies here, but also many other places. If a routine is not going to actually create a new reference to a `shared_ptr` that persists past the scope of the function, then it doesn't really make sense to pass as a `shared_ptr`. Instead, you should really just leave these as `MeshBlock *`. It better expresses intent at the call site (`pmb.get()` to get a `MeshBlock *`) - that you're handing out a temporary borrow rather than creating new shared ownership. The ownership of the object is still asserted at the call site since you have a shared_ptr.

Additionally, it's a bad pattern for performance (doesn't apply in this specific case because it's a big function) - passing a `shared_ptr` by value results in an atomic reference AND de-reference, which can overwhelm the cost of small functions

example/advection/advection_driver.cpp
```diff
@@ -124,7 +124,7 @@ TaskList AdvectionDriver::MakeTaskList(MeshBlock *pmb, int stage) {
   if (stage == integrator->nstages) {
     auto new_dt = tl.AddTask(
         [](std::shared_ptr<Container<Real>> &rc) {
-          MeshBlock *pmb = rc->pmy_block;
+          auto pmb = rc->pmy_block.lock();
```
AndrewGaspar: Just a note: `lock` will definitely help in case somebody introduces a bug that makes the MeshBlock get destroyed before the containers it owns are destroyed, as it will definitely result in a segmentation fault, as opposed to trying to reference a heap address that's potentially still mapped in user memory space. Keep in mind that it will still be a segmentation fault, not something more informative.

AndrewGaspar: If you're going to change all the places `pmy_block` is used anyway, I'd prefer you actually add accessor functions like `std::shared_ptr<MeshBlock> Block() const { return pmy_block.lock(); }` and make `pmy_block` private.

AndrewGaspar: You might even put a null check in there and make a loud error like "Parthenon internal error: Block object destroyed too early!"

src/mesh/mesh.hpp
```diff
@@ -73,25 +73,44 @@ KOKKOS_INLINE_FUNCTION void par_for_inner(const team_mbr_t &team_member, const i
...
-            int igflag, bool ref_flag = false);
-  MeshBlock(int igid, int ilid, Mesh *pm, ParameterInput *pin, ApplicationInput *app_in,
-            Properties_t &properties, Packages_t &packages, LogicalLocation iloc,
-            RegionSize input_block, BoundaryFlag *input_bcs, double icost, int igflag);
   ~MeshBlock();
 
+  // Initializer to set up a meshblock called with the default constructor
+  // This is necessary because the back pointers can't be set up until
+  // the block is allocated.
+  void Initialize(int igid, int ilid, LogicalLocation iloc, RegionSize input_block,
```
AndrewGaspar: Can this be private?

src/mesh/mesh.hpp
```diff
@@ -73,25 +73,44 @@ KOKKOS_INLINE_FUNCTION void par_for_inner(const team_mbr_t &team_member, const i
...
+  static std::shared_ptr<MeshBlock>
+  Make(int igid, int ilid, LogicalLocation iloc, RegionSize input_block,
+       BoundaryFlag *input_bcs, Mesh *pm, ParameterInput *pin, ApplicationInput *app_in,
+       Properties_t &properties, Packages_t &packages, int igflag, double icost = 1.0) {
+    auto pmb = std::make_shared<MeshBlock>();
+    pmb->Initialize(igid, ilid, iloc, input_block, input_bcs, pm, pin, app_in, properties,
+                    packages, igflag, icost);
+    return pmb;
+  }
+  static std::shared_ptr<MeshBlock> MakeAndSetNeighbors(
```
AndrewGaspar: Hm... I'm not sure this constructor is necessary. Doesn't really reduce LOC at all - seems to increase it, actually. If there was some optimization we could take because we're setting the neighbors at the same time as initialization, but that doesn't seem to be the case.

AndrewGaspar (changes requested):

src/mesh/mesh.hpp
Yurlungur: If we always use the factory functions, yes. But I figured there might be cases where a non-factory call will be made.

src/mesh/mesh.hpp
Yurlungur: It's just syntactic sugar. I'm not married to it.

src/mesh/amr_loadbalance.cpp
Yurlungur: Yeah ok. This makes sense. I'll undo those changes.

src/mesh/mesh.hpp
AndrewGaspar: In general for API design, I prefer to limit new APIs by default, and only introduce them when need has arisen. The issue with adding APIs because you anticipate their usage is that they tend not to be built-to-fit.

src/mesh/mesh.hpp
AndrewGaspar: Yeah, I'd say axe this and call `SetNeighbors` at the call site.

example/advection/advection_driver.cpp
Yurlungur: Resolved.

src/mesh/amr_loadbalance.cpp
Yurlungur: Resolved.

src/mesh/mesh.hpp
Yurlungur: Yeah ok. I suspect something like this may need to be public eventually... or more factory functions may need to be created. But for now I'll make it private.

src/mesh/mesh.hpp
Yurlungur: Resolved.

Yurlungur:
@AndrewGaspar I believe all your requested changes have been implemented.

src/mesh/mesh.hpp
```diff
@@ -276,7 +258,9 @@ class MeshBlock : public std::enable_shared_from_this<MeshBlock> {
   int GetNumberOfMeshBlockCells() const {
     return block_size.nx1 * block_size.nx2 * block_size.nx3;
   }
-  void SearchAndSetNeighbors(MeshBlockTree &tree, int *ranklist, int *nslist);
+  void SearchAndSetNeighbors(MeshBlockTree &tree, int *ranklist, int *nslist) {
```
AndrewGaspar, line 261: Funny this wasn't caught until now.

AndrewGaspar (approved):
`GetBlockPointer` is a mouthful, but I'll accept it.

Yurlungur:
@pgrete @forrestglines can I get one more approval from the Athena team?

AndrewGaspar:
Yo @github just go ahead and merge this when the CI passes - could you do that? No?

Yurlungur:
> `GetBlockPointer` is a mouthful, but I'll accept it.

Yeah. I know. `Block` felt too vague, though.

src/interface/container.hpp
```diff
@@ -71,7 +75,10 @@ class Container {
 
   ///
   /// Set the pointer to the mesh block for this container
-  void setBlock(MeshBlock *pmb) { pmy_block = pmb; }
+  void setBlock(std::weak_ptr<MeshBlock> pmb) { pmy_block = pmb; }
+  void setBlock(const std::shared_ptr<Container<T>> &other) {
```
pgrete: [nitpicking]
Should be potentially call this SetBlockPointer (so that it matches GetBlockPointer).
As far as I see this is only called like 5 times in the codebase so the extra characters wouldn't clutter too much
[/nitpicking]

src/utils/buffer_utils.hpp
```diff
@@ -19,6 +19,8 @@
 //! \file buffer_utils.hpp
 //  \brief prototypes of utility functions to pack/unpack buffers
 
+#include <memory>
+
```
pgrete:
```suggestion
```

src/utils/buffer_utils.cpp
```diff
@@ -17,6 +17,8 @@
 //! \file buffer_utils.cpp
 //  \brief namespace containing buffer utilities.
 
+#include <memory>
+
```
pgrete:
```suggestion
```

src/mesh/meshblock.cpp
```diff
@@ -109,118 +117,13 @@ MeshBlock::MeshBlock(int igid, int ilid, LogicalLocation iloc, RegionSize input_
...
-
-  if (pm->multilevel) {
-    pmr = std::make_unique<MeshRefinement>(this, pin);
-    // This is very redundant, I think, but necessary for now
-    for (int n = 0; n < nindependent; n++) {
-      pmr->AddToRefinement(ci.vars[n]->data, ci.vars[n]->coarse_s);
-    }
-  }
-
-  // Create user mesh data
```
pgrete, line 159: Maybe I'm missing something, but where does this happen now?

Yurlungur: It happens [here](https://github.com/lanl/parthenon/blob/a004fa7ac8f5f7964a62b0f68d6bebb0344105d6/src/mesh/meshblock.cpp#L126). There were three `MeshBlock` constructors, which all (as far as I can tell) were identical but with slightly different signatures and duplicated code. So I combined them.

src/interface/container.hpp
Yurlungur: Done.

src/mesh/meshblock.cpp
pgrete: Silly me... I just looked at the diff and didn't look at the whole file. Thanks for clarifying.

pgrete (approved):
Looks good and the cleanup of the constructors was well overdue.
Given the extent of the changes I'll trigger the full test now.
Once that passes this can get merged from my point of view.

Yurlungur:
Pulling the trigger.

PR #308, opened 2020-09-19
Add Cache for MeshBlockPacks

Yurlungur:
## PR Summary

With PR #263 , we have `MeshBlockPack`s, which pack a variable over the entire `Mesh`. As we discussed on the call and in that PR, those packs were only partially cached. The `VariablePack` is cached within a `container`, so when a `MeshBlockPack` is generated, the `VariablePack`s are not regenerated. But the `MeshBlockPack` itself may be regenerated every cycle. Here I resolve that inefficiency and also provide some quality of life improvements when dealing with `MeshBlockPack`s.

I implement a mechanism where `MeshBlockPack`s can be registered with the `Mesh` at initialization time inside a package `Initialize` function. The packs live within a package namespace and can be named, which may avoid tedious and verbose usage of `Metadata` flags. A complication is that `MeshBlockPack`s must be regenerated every time the `Mesh` changes, e.g., for load balancing or AMR. To get around this complication, the user doesn't register the pack itself, the user registers a *function* that performs the packing. 

The infrastructure is also completely capable of handling `MeshBlockPack`s of various sizes, from 1 block to the whole mesh. I implemented an example usage in `calculate_pi`. The registration looks like:
```C++
  package->AddMeshBlockPack("in_or_out", [](Mesh *pmesh) {
    int pack_size = pmesh->DefaultPackSize();
    auto partitions = partition::ToSizeN(pmesh->block_list, pack_size);
    std::vector<MeshBlockVarPack<Real>> packs(partitions.size());
    for (int i = 0; i < partitions.size(); i++) {
      packs[i] = PackVariablesOnMesh(partitions[i], "base",
                                     std::vector<std::string>{"in_or_out"});
    }
    return packs;
```
and the pack is accessed in the task list as
```C++
  auto &packs = pmesh->real_varpacks["calculate_pi"]["in_or_out"];
  ParArrayHost<Real> areas("areas", packs.size());
  TaskRegion &async_region = tc.AddRegion(packs.size());
  {
    // asynchronous region where area is computed per mesh pack
    for (int i = 0; i < packs.size(); i++) {
      TaskID none(0);
      auto get_area = async_region[i].AddTask(ComputeArea, none, packs[i], areas, i);
    }
  }
```

### Default packs

I also implement a default `MeshBlockPack` under the `"default"` namespace. In particular, if there are any variables with `Metadata::FillGhosts`, a pack over them is available as 
```C++
pmesh->real_varpacks["default"]["fill_ghosts"];
```
which is a `std::vector` of `MeshBlockPacks` of size `pmesh->DefaultPackSize()`. This size can be set as a user parameter via the input file. By default it's
```
<parthenon/mesh>
pack_size = -1
```
where any value < 1 means use the whole grid.

More default packs can be added in `Mesh::RegisterAllMeshBlockPackers`, and we should add them as we discover a need.

## PR Checklist

- [x] Code passes cpplint
- [x] Adds a test for any bugs fixed. Adds tests for new features. (Features are in Mesh. Tests added in example/regression.)
- [x] Code is formatted


Yurlungur:
There may be a merge conflict with #307 depending on which one gets merged in first.

Yurlungur:
@pgrete this implements the features we discussed as far as default `MeshBlockPack`s and caching. I'd be interested in seeing if you can integrate it into the pack in one branch you're working on.

Yurlungur:
Ah one more thing worth mentioning. Putting `MeshBlockPack`s into the `Mesh` required disentangling some of the headers. In particular, I split the `mesh.hpp` header into two headers: `mesh.hpp` and `meshblock.hpp`. I also removed several includes across the code and replaced them with forward declarations.

example/calculate_pi/calculate_pi.cpp
```diff
@@ -71,12 +71,29 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
   Metadata m({Metadata::Cell, Metadata::Derived});
   package->AddField(field_name, m, DerivedOwnership::unique);
 
+  // Add a named MeshPack by registering a function that packs it
+  package->AddMeshBlockPack("in_or_out", [](Mesh *pmesh) {
+    // 1 means pack is 1 meshblock, <1 means use entire mesh
+    int pack_size = pmesh->DefaultPackSize();
```
jdolence, line 76: This seems unnecessarily complicated.  Why not just have `pmesh->DefaultPackSize()` return `block_list.size()` instead of something < 1?

example/calculate_pi/calculate_pi.cpp
```diff
@@ -71,12 +71,29 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
...
+    if (pack_size < 1) pack_size = pmesh->block_list.size();
+    std::vector<BlockList_t> partitions;
+    std::vector<MeshBlockVarPack<Real>> packs;
+    partition::ToSizeN(pmesh->block_list, pack_size, partitions);
+    packs.resize(partitions.size());
```
jdolence, line 79: Suggestion: `partition::ToSizeN` could just take the `pmesh->block_list` and `pack_size` args and return the `std::vector<BlockList_t> partitions`.  Then you'd do something like
```c++
std::vector<MeshBlockVarPack<Real>> packs(partitions.size());
```

Yurlungur: The reason I didn't go that route is I didn't want to return a `std::vector` for performance reasons. But if we wrap it in a `std::move` it should be fine.

example/calculate_pi/calculate_pi.cpp
Yurlungur: Because I didn't think of it. Yes, I can make that change. In `parthinput`, `<1` needs to be an option since I don't want the user to have to know what sizes are available. But I agree here.

example/calculate_pi/calculate_pi.cpp
jdolence: I don't think the `std::move` is actually required in this case.

Yurlungur: Well you'd need a line like
```C++
std::vector<MeshBlockVarPack<Real>> packs(partition::ToSizeN(pmesh->block_list,pack_size));
```
right? And this *might* automatically do a `std::move` but you could enforce it with 
```C++
std::vector<MeshBlockVarPack<Real>> packs = std::move(partition::ToSizeN(pmesh->block_list,pack_size));
```
Or am I missing something?

jdolence: Not sure I follow.  I was suggesting something like
```c++
auto partitions = partition::ToSizeN(pmesh->block_list, pack_size);
std::vector<MeshBlockVarPack<Real>> packs(partitions.size());
```
I believe if you make `partition::ToSizeN` just return the local-to-that-function `std::vector<BlockList_t>` the compiler will take care of the rest.  In fact, I believe this is the preferred way to write such a capability.  No explicit `std::move`s should be used.  I could be wrong, of course.

Yurlungur: I might be stuck in the past of C++. I remember the advice not to return things explicitly like that. Either way, easy to check. Just switch the design and see what happens.

Also if it's only called once per remeshing, the difference is almost certainly negligible.

jdolence: Yeah I'm not really worried about performance.  I just think it's cleaner this way.

Yurlungur: Agreed. I'll make that change.

example/calculate_pi/calculate_pi.cpp
Yurlungur: Good point. Resolved.

example/calculate_pi/calculate_pi.cpp
Yurlungur: Getting type inference to work with this change required a little template foo, but I got it.

Yurlungur:
@jdolence I think your concerns have been addressed now.

Yurlungur:
> There may be a merge conflict with #307 depending on which one gets merged in first.

Minor merge conflict addressed. This is ready.

pgrete:
Collecting notes here for items that may need discussion:
- [ ] partition namespace is ambiguous
```c++
driver.hpp(67, 11): candidates are: ‘namespace parthenon::driver::prelude::partition { }’
package.hpp(67, 11): ‘namespace parthenon::package::prelude::partition { }’
```
- [ ] How to (transparently) handle caches for multi stage integrator
- [ ] Is there/how does a clean separation of concerns between `Containers` and `MeshBlockPacks` look like?
- [ ] Similarly, what's our "primary" interface? `Containers` or `Packs`?
- [ ] `GetNDim` host access

Yurlungur:
As discussed, I'm marking this non-Wip again. We will merge it in, and then move to a more desirable caching infrastructure.

pgrete (approved):
As discussed during the call, I approve (tested the changes downstream and they work as expected, and we'll update the interfaces along the way)

docs/mesh/packing.md
```diff
@@ -30,17 +30,15 @@ can pass in a metadata vector, a vector of names, and optionally IDs
...
 Instead of packing over the whole mesh, you can pack over only a piece
 of it by using the `Partition` machinery found in
 `utils/partition_stl_containers.hpp`. For example, to break the mesh
 into four evenly sized meshpacks, do
 ```C++
 using parthenon::MeshBlock;
-using parthenon::Partition::Partition_t;
-Partition_t<MeshBlock> partitions;
-parthenon::Partition::ToNPartitions(mesh->block_list, 4, partitions);
+auto partitions = parthenon::Partition::ToNPartitions(mesh->block_list, 4);
```
AndrewGaspar:
```suggestion
auto partitions = parthenon::partition::ToNPartitions(mesh->block_list, 4);
```

docs/mesh/packing.md
```diff
@@ -57,16 +55,21 @@ for (int i = 0; i < partitions.size() {
...
+template <typename T, typename Container_t>
+Partition_t<T> ToNPartitions(Container_t<T> &container, const int N);
 
 // Splits container into partitions of size N
-template <typename Container_t, typename T>
-void ToSizeN(Container_t &container, const int N, Partition_t<T> &partitions);
+template <typename T, typename Container_t>
+std::vector<std::vector<T>> ToSizeN(Container_t<T> &container, const int N);
+```
+Both functions live within the namespace `parthenon::Partition` and `Partition_t`
```
AndrewGaspar:
```suggestion
Both functions live within the namespace `parthenon::partition` and `Partition_t` 
```

docs/mesh/packing.md
```diff
@@ -57,16 +55,21 @@ for (int i = 0; i < partitions.size() {
...
 // Splits container into partitions of size N
-template <typename Container_t, typename T>
-void ToSizeN(Container_t &container, const int N, Partition_t<T> &partitions);
+template <typename T, typename Container_t>
+std::vector<std::vector<T>> ToSizeN(Container_t<T> &container, const int N);
+```
+Both functions live within the namespace `parthenon::Partition` and `Partition_t` 
+is defined as:
+```C++
+templat<typename T>
```
AndrewGaspar:
```suggestion
template<typename T>
```

docs/interface/state.md
```diff
@@ -13,6 +13,7 @@ The ```StateDescriptor``` class is intended to be used to inform Parthenon about
 Provides the means to add new variables to a Parthenon-based application with associated ```Metadata```.  This function does not allocate any storage or create any of the objects below, it simply adds the name and ```Metadata``` to a list so that those objects can be populated at the appropriate time.
 * ```void AddParam<T>(const std::string& key, T& value)``` adds a parameter (e.g. a timestep control coefficient, refinement tolerance, etc.) with name ```key``` and value ```value```.
 * ```const T& Param(const std::string& key)``` provides the getter to access parameters previously added by ```AddParam```.
+* ```void AddMeshBlockPack(const std::string &pack_name, const Function &packer)``` adds a function that generates `MeshBlockPack`s on the `Mesh` as described [here](../mesh/packing.md)
```
AndrewGaspar:
```suggestion
* ```void AddMeshBlockPack(const std::string &pack_name, const Function &packer)``` registers a callback that generates `MeshBlockPack`s on the `Mesh` as described [here](../mesh/packing.md)
```

src/mesh/mesh.cpp
```diff
@@ -900,6 +911,79 @@ void Mesh::OutputMeshStructure(int ndim) {
...
+void Mesh::BuildMeshBlockPacks() {
+  // JMM: I want C++17 structured bindings...
+  for (auto &outer : real_varpackers_) {
+    auto &package = outer.first;
+    auto &packers = outer.second;
+    for (auto &pair : packers) {
+      auto &name = pair.first;
+      auto &func = pair.second;
+      // avoid unnecessary copies
+      real_varpacks[package][name] = std::move(func(this));
```
AndrewGaspar: `std::move` doesn't do anything here - the result of `func(this)` is already an rvalue.

src/mesh/mesh.cpp
```diff
@@ -900,6 +911,79 @@ void Mesh::OutputMeshStructure(int ndim) {
...
+    }
+  }
+  for (auto &outer : real_fluxpackers_) {
+    auto &package = outer.first;
+    auto &packers = outer.second;
+    for (auto &pair : packers) {
+      auto &name = pair.first;
+      auto &func = pair.second;
+      real_fluxpacks[package][name] = std::move(func(this));
```
AndrewGaspar: same comment as previous on `std::move`

src/mesh/mesh.cpp
AndrewGaspar:
```suggestion
      real_varpacks[package][name] = func(this);
```

src/mesh/mesh.cpp
AndrewGaspar:
```suggestion
      real_fluxpacks[package][name] = func(this);
```

docs/mesh/packing.md
```diff
@@ -115,3 +118,124 @@ MeshBlockVarFluxPack<T>
...
+std::vector<MeshBlockPack<VariableFluxPack<Real>>>`, respectively.
+
+An example registration might look like this:
+```C++
+int pack_size = 4; // The number of meshblocks in a pack
+pmesh->RegisterMeshBlockPack("default", "fill_ghost", [pack_size, metadata](Mesh *pmesh) {
+    // Add containers if not already present
+    for (auto &pmb : pmesh->block_list) {
+        auto &base = pmb->real_containers.Get();
+        pmb->real_containers.Add("new",base);
```
AndrewGaspar, line 156: Isn't this what AddField is for? When would you chose to allocate "per-block" variables vs "mesh-wide" variables?

docs/mesh/packing.md
```diff
@@ -115,3 +118,124 @@ MeshBlockVarFluxPack<T>
 ```
 which correspond to packs over meshblocks that contain just variables
 or contain variables and fluxes.
+
+## Registering MeshBlockPacks
```
AndrewGaspar, line 122: Don't change anything here, because I guess we've already decided this is going in. I just want to make sure my comments on the design are recorded here.

I'm unsure the user needs that level of flexibility this design provides. My main concern with it is the coarseness at which the caching is done. As far as I'm concerned, the `RequestMeshBlockPack` callback shouldn't be concerned with partitioning at all. That's what the `TaskCollection` is for. This seems to make it very error prone - you better make sure your cache allocation partitions the mesh _the exact same way_ as the block partitioner for TaskRegion does. It also removes flexibility from the infrastructure - we can't make optimizations in Parthenon to, say, preserve block packs between cycles because the application has now encoded the precise method that Parthenon currently partitions blocks.

Since we're doing explicit caching here rather than implicit caching (though I'm not sure why we decided to go this route), why can't the app simply request a cache for ["x", "y", "z"] variables. Something like

```c++
pmesh->RegisterMeshBlockPack("default", "fill_ghost", [](Mesh *pmesh) {
    return std::vector<std::string> {"new"};
});
```

Then, the app must request the `("default", "fill_ghost")` cache for a region instead, something like `auto pack = pmesh->GetPack(region, "default", "fill_ghost");`.

docs/mesh/packing.md
```diff
@@ -115,3 +118,124 @@ MeshBlockVarFluxPack<T>
...
+    return packs;
+});
+```
+for a `Mesh* pmesh`.
+
+Note that the packing function expects you to prepare the state of the
+mesh for the pack. This may mean you need to create new containers if
+they are not available in your container collection. It may also mean
+you must partition the your meshblocks if you want to pack over a
+piece of the mesh, as in the example above.
```
AndrewGaspar: I feel like there's either a missing or extra word in this sentence.

docs/mesh/packing.md
```diff
@@ -115,3 +118,124 @@ MeshBlockVarFluxPack<T>
...
+Mesh::real_fluxpacks;
+```
+
+The former contains `MeshBlockPack<VariablePack<Real>>`s and the
+latter contains `MeshBlockPack<VariableFluxPack<Real>>`s. The actual
+pack must be accessed with two strings, and an integer, a package
+namespace, a pack name, and an index. For example, to access one of
+the packs registed above, you would call
+```C++
+auto &pack = pmesh->real_varpacks["default"]["fill_ghost"][0];
```
AndrewGaspar, line 207: Turn this into a function that returns the pack - this exposes implementation details.

docs/mesh/packing.md
```diff
@@ -115,3 +118,124 @@ MeshBlockVarFluxPack<T>
...
+should be contained within a single pack. This can be accessed within
+the program via
+```C++
+pmesh->DefaultPackSize();
+```
+The default packs available are:
+
+| Namespace | Name        | Contains Fluxes | Metadata Condition     |
+| --------- | ----------- | --------------- | ---------------------- |
+| default   | fill_ghosts | No              | `Metadata::FillGhosts` |
```
AndrewGaspar:
```suggestion
| default   | fill_ghosts | No              | `Metadata::FillGhost` |
```

src/interface/state_descriptor.hpp
```diff
@@ -125,6 +150,28 @@ class StateDescriptor {
...
 
+  // Get all MeshBlockPacker functions
+  const std::map<std::string, VarPackingFunc<Real>> &AllMeshBlockVarPackers() {
+    return realVarPackerMap_;
+  }
+  const std::map<std::string, FluxPackingFunc<Real>> &AllMeshBlockFluxPackers() {
+    return realFluxPackerMap_;
+  }
+
+  bool FlagsPresent(std::vector<MetadataFlag> &flags, bool matchAny = false) {
```
AndrewGaspar:
```suggestion
  bool FlagsPresent(std::vector<MetadataFlag> const &flags, bool matchAny = false) {
```

docs/mesh/packing.md
AndrewGaspar: With this callback, btw, you only need to call this once after all packages are registered. Then the infrastructure has maximum flexibility to decided when and how new caches are allocated.

AndrewGaspar: It's important to note that this change will require a wider change to the code. Currently the tasking model is only kind-of coupled to the mesh, but not enough. We'll need to do a tighter coupling - we describe our task lists in a very granular way, and that would probably need to change, too.

AndrewGaspar (changes requested):
Only address the small change requests, but I have some design comments, too, which I understand won't be addressed here.

src/mesh/mesh.cpp
Yurlungur: I always struggle with that. Thanks for the catch.

docs/mesh/packing.md
Yurlungur: This is a design issue we'll have to handle later in the `MeshData` vs. `MeshBlockData`. But this isn't adding fields, it's adding stages in `ContainerCollection`. It's typically done at the driver level. But if you want a `MeshBlockPack` for a particular stage, you need to create that before the task list is run. That's why it's here.

I wouldn't worry about this. It's needed for this design but will go away in the refactor.

docs/mesh/packing.md
Yurlungur: > As far as I'm concerned, the RequestMeshBlockPack callback shouldn't be concerned with partitioning at all. 

Hmm... I'm not sure how *not* to include that. I can imagine applications sometimes wanting to pack over `N` meshblocks and sometimes over `M`. I think that needs to be available somehow. However, I agree that it should be automatically synchronized with `TaskCollection` if possible. And that's probably doable. 

> Since we're doing explicit caching here rather than implicit caching

That's going to change. Once we move to implicit caching, things should be more workable.

> It's important to note that this change will require a wider change to the code. Currently the tasking model is only kind-of coupled to the mesh, but not enough. We'll need to do a tighter coupling - we describe our task lists in a very granular way, and that would probably need to change, too.

This probably requires a *lot* more thought. But I think you're right that's a good path forward.

docs/mesh/packing.md
Yurlungur: I don't quite see where. How would you write this sentence?

docs/mesh/packing.md
Yurlungur: Is this worth changing? This is going to go away in the new implementation.

docs/mesh/packing.md
AndrewGaspar: "you must partition the your meshblocks" - something weird going on with "the your"

Yurlungur: oh man I completely missed that when reading over it. Yes, I'll fix that.

docs/mesh/packing.md
AndrewGaspar: OK

docs/mesh/packing.md
AndrewGaspar: > Hmm... I'm not sure how not to include that.

It's not that the pack shouldn't be per-partition, it's just that the user shouldn't be manually doing the partitioning.

Yurlungur: Right, but the user should be doing the partitioning somewhere. Either at the pack level, or at the tasking level.

pgrete: I think this is very much in line with what we discussed during the call.
With the implicit caching  (and invalidating the cache, e.g., after a load balancing call) the TaskCollection would be the only* place left where task packing logic is required and self-consistently handled for full cycle.

\* plus the calls in the infrastructure (e.g., derived fields or EstimateTimestep), but I'm in favor of changing those interfaces in a way that the information required (about the packs or directly about the `MeshBlocks` if the application decides to define a Task at the MeshBlock level) is provided from within `TaskCollection` rather than extracted from logic in a container

Yurlungur: Yes, I think that's true.

Yurlungur:
@AndrewGaspar I believe all issues have been addressed. Is this ready for merge?

AndrewGaspar (approved):

AndrewGaspar:
Yeah, this can be merged when the CI passes.

PR #314, opened 2020-10-02
Abstractions for parallel reductions

### LINKED ISSUE #179: Parthenon Reductions
I think MR #136 implies we need parthenon wrappers for `Kokkos::parallel_reduce`. This isn't urgent but this is something we should look at eventually.

jdolence:
## PR Summary

This PR generalizes our loop abstractions to provide access to `Kokkos::parallel_reduce` with a consistent interface.  I used somewhat advanced C++ features to do this to avoid lots of code duplication, so I'm especially interested to hear how strenuously people object.

Addresses #179.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


jdolence:
I'd be especially interested in feedback from @pgrete @Yurlungur @nmsriram and @AndrewGaspar.

example/advection/advection_package.cpp
```diff
@@ -173,11 +173,8 @@ AmrTag CheckRefinement(std::shared_ptr<Container<Real>> &rc) {
   IndexRange kb = pmb->cellbounds.GetBoundsK(IndexDomain::entire);
 
   typename Kokkos::MinMax<Real>::value_type minmax;
-  Kokkos::parallel_reduce(
-      "advection check refinement",
-      Kokkos::MDRangePolicy<Kokkos::Rank<3>>(pmb->exec_space, {kb.s, jb.s, ib.s},
-                                             {kb.e + 1, jb.e + 1, ib.e + 1},
-                                             {1, 1, ib.e + 1 - ib.s}),
+  pmb->par_for(
+      "advection check refinement", kb.s, kb.e, jb.s, jb.e, ib.s, ib.e,
```
Yurlungur: I think I mentioned that I don't love that `par_for` is now both a for loop and a reduction. I'd prefer a new name that's a little more generic. But I can live with it.

example/advection/advection_package.cpp
```diff
@@ -270,22 +267,21 @@ Real EstimateTimestep(std::shared_ptr<Container<Real>> &rc) {
...
 
-  Real min_dt = std::numeric_limits<Real>::max();
   auto &coords = pmb->coords;
 
   // this is obviously overkill for this constant velocity problem
-  for (int k = kb.s; k <= kb.e; k++) {
-    for (int j = jb.s; j <= jb.e; j++) {
-      for (int i = ib.s; i <= ib.e; i++) {
+  Real min_dt;
+  pmb->par_for(
```
Yurlungur: we may want to make explicitly clear somewhere that this doesn't work for the simd for loop pattern and explicitly pass in (say) the mdrange tag.

src/refinement/refinement.hpp
```diff
@@ -18,21 +18,26 @@
 #include <string>
 
 #include "defs.hpp"
-#include "interface/container.hpp"
-#include "interface/state_descriptor.hpp"
-#include "interface/variable.hpp"
+#include "parthenon_arrays.hpp"
+//#include "interface/container.hpp"
+//#include "interface/state_descriptor.hpp"
+//#include "interface/variable.hpp"
```
Yurlungur: remove these since we don't seem to need them.

src/refinement/refinement.cpp
```diff
@@ -88,7 +88,7 @@ AmrTag CheckAllRefinement(std::shared_ptr<Container<Real>> &rc) {
   return delta_level;
 }
 
-AmrTag FirstDerivative(DevExecSpace exec_space, const ParArrayND<Real> &q,
```
Yurlungur, line 91: Should the reduce be capable of operating over a specific stream? That's probably an enhancement we can skip for now.

src/kokkos_abstraction.hpp
```diff
@@ -118,65 +120,93 @@ static struct InnerLoopPatternTVR {
...
+
+template <class... Args>
+inline void kokkos_dispatch(ParallelForDispatch, Args &&... args) {
+  Kokkos::parallel_for(std::forward<Args>(args)...);
+}
+template <class... Args>
+inline void kokkos_dispatch(ParallelReduceDispatch, Args &&... args) {
+  Kokkos::parallel_reduce(std::forward<Args>(args)...);
+}
+
```
Yurlungur, line 143: I'm torn about whether or not these dispatch types should be hidden inside an internal namespace. On the one hand, they're currently "implementation details." On the other hand, they could be convenient to expose.

pgrete:
I generally like how those few lines of code add the reduction functionality to all wrappers.
I'm still unsure how I feel about "hiding" the reduction by just adding an additional argument to the wrapper (versus being more verbose like Kokkos with a `par_reduce` or similar).
Let's see what other people think.

Yurlungur (commented):
This is a much needed improvement and I think the solution you've implemented is elegant. Not too much repeated code, and indeed the original code remains much the same.

Obviously some of this code will need to move/change when we finish the transition to meshblockpacks.

Regarding @pgrete 's concern about the naming---I kind of agree. I don't like that reduction has the same name as par for. But I don't dislike it enough to object. I think the benefits of little code, augmented by documentation on this behaviour, override the cost.

One suggestion. If the user sets the default loop pattern to simdfor, then the reductions won't work. Maybe we should explicitly set them to something or add a default reduce loop pattern, which is by default MDRange or something.

example/advection/advection_package.cpp
jdolence: yes, I agree, just don't want to have to update a whole bunch of code or break things downstream.  and I also don't find it _too_ offensive.

example/advection/advection_package.cpp
jdolence: if you were to try the simd loop pattern, it would fail at compile time.  In the `MeshBlock` wrappers, this is already handled transparently.  In fact, I've been testing this in an environment where the DEFAULT_LOOP_POLICY is the simd for loop, and the code automatically switches to the MDRange loop if you're doing a reduction.

src/refinement/refinement.hpp
jdolence: done.

src/refinement/refinement.cpp
jdolence: maybe I'm not sure what you're asking.  I believe it already is tied to a specific stream in the same way the `pmb->par_for` loops always are.

example/advection/advection_package.cpp
Yurlungur: Oh nice. Okay I missed that somewhere. My above comment about making a default reduce pattern still holds though, I think. As the most efficient pattern for reduction may not be the same as the most efficient pattern for a stencil.

src/kokkos_abstraction.hpp
jdolence: I introduced a namespace `dispatch_impl`.  what do you think @Yurlungur?

src/refinement/refinement.cpp
Yurlungur: I may have just missed that when reading through it. Ignore this.

example/advection/advection_package.cpp
jdolence: I started going down that path, but abandoned it (probably mostly out of laziness).  This solution works for now, so I figured we could return to this if we find reductions showing up performance wise.

example/advection/advection_package.cpp
Yurlungur: Yeah fair enough.

src/kokkos_abstraction.hpp
Yurlungur: Yep works for me.

example/advection/advection_package.cpp
Yurlungur: Yeah fair enough.

jdolence:
Well I think this is ready to go.  I'm still willing to be convinced we should _not_ pursue this path (I think it might introduce the code's first real template metaprogramming, which one might imagine we could sensibly forbid).

Yurlungur (approved):
Documentation looks good. I'm happy to get this in. :+1:

Yurlungur:
@nmsriram @JoshuaSBrown @AndrewGaspar can we get one LANL CS person to take a look at this?

@pgrete have your thoughts on this changed?

AndrewGaspar:
I really want the naming to change - a reduction should be `par_reduce`.

jdolence:
> I really want the naming to change - a reduction should be `par_reduce`.

Is there a way to do this without a bunch of code duplication?

JoshuaSBrown:
> @nmsriram @JoshuaSBrown @AndrewGaspar can we get one LANL CS person to take a look at this?
> 
> @pgrete have your thoughts on this changed?

Ya, I've been looking over it. Had to refresh myself on some things.

Yurlungur:
> > I really want the naming to change - a reduction should be `par_reduce`.
> 
> Is there a way to do this without a bunch of code duplication?

What if we renamed the variable length dispatch methods @jdolence wrote `par_dispatch` and then wrapped `par_dispatch with inline functions named `par_for` and `par_reduce`? i.e.,

```C++
inline void par_for(stuff) {
    par_dispatch(stuff);
}
inline void par_reduce(stuff) {
  par_dispatch(stuff);
}
```

AndrewGaspar:
That would be fine.

src/kokkos_abstraction.hpp
```diff
@@ -118,65 +120,100 @@ static struct InnerLoopPatternTVR {
...
+namespace dispatch_impl {
+static struct ParallelForDispatch {
+} parallel_for_dispatch_tag;
+static struct ParallelReduceDispatch {
+} parallel_reduce_dispatch_tag;
+
+template <class... Args>
+struct DispatchType {
+  typedef typename std::conditional<sizeof...(Args) == 0, ParallelForDispatch,
+                                    ParallelReduceDispatch>::type type;
```
JoshuaSBrown, line 132: This is clever.

JoshuaSBrown (approved):
I have no qualms with this other than what @Yurlungur and @AndrewGaspar have raised about naming, and we already have unit tests for kokkos_abstraction :).

AndrewGaspar (changes requested):
Add separate `par_reduce` function, otherwise looks good.

jdolence:
> Add separate `par_reduce` function, otherwise looks good.

done.  there is now a `par_for` and a `par_reduce`, though it doesn't _actually_ matter which one you call because they both just forward on to generic functions that handle both.  happy with this?

src/refinement/refinement.cpp
```diff
@@ -106,12 +106,9 @@ AmrTag FirstDerivative(DevExecSpace exec_space, const ParArrayND<Real> &q,
     il = 1;
     iu = dim1 - 2;
   }
-
   Real maxd = 0.0;
-  Kokkos::parallel_reduce(
-      "refinement first derivative",
-      Kokkos::MDRangePolicy<Kokkos::Rank<3>>(
-          exec_space, {kl, jl, il}, {ku + 1, ju + 1, iu + 1}, {1, 1, iu + 1 - il}),
+  pmb->par_for(
```
AndrewGaspar: I don't love that this still compiles, but anyway, this should be `par_reduce`

```suggestion
  pmb->par_reduce(
```

AndrewGaspar (changes requested):

AndrewGaspar (approved):

jdolence:
@pgrete based on your comment and the changes that have gone in since, I'm planning to merge once the CI passes.  Last chance to object.

PR #317, opened 2020-10-07
Add particle framework but for CI

brryan:
See #126, I just made this copycat PR to get the CI to run since the other PR's compare branch is from my fork of the parthenon repo.


brryan:
@Yurlungur @pgrete The unit test for `Swarm` fails in this PR on the CUDA CI build, but works fine for me on a Power9/Nvidia V100 node. The error is:
```
  what():  cudaMemcpy(dst, src, n, cudaMemcpyDefault) error( cudaErrorIllegalAddress): an illegal memory access was encountered /builds/pgrete/parthenon/external/Kokkos/core/src/Cuda/Kokkos_CudaSpace.cpp:93
```
On my end I'm compiling in `Debug` mode with the appropriate Kokkos backend for Power9/V100s and calling `ctest -LE 'performance|regression'` or just running the `unit_tests` executable directly. Is there anything more I can do to get this error to occur locally for me? The CI error log is not verbose so I can't track this issue down easily.

Yurlungur:
Ouch. I have no idea. That's pretty confusing. locally we have some x86 V100's, so we could try that.

I can suggest some debugging you could add to the CI, but that's not a very satisfying answer.

brryan:
Just littered `test_swarm.cpp` with `printf("%s:%i\n", __FILE__, __LINE__);`s so we'll see what comes out of the CI this time.

Yurlungur:
Try the kokkos tools kernel logger

brryan:
OK I'm zeroing in on the problem already I think I can take it from here

pgrete:
Using the kernel logger is often very useful in these kind of cases.
Given the async API calls it's sometimes not obvious where the error actually came from.
Alternatively, setting `CUDA_LAUNCH_BLOCKING=1` can also help in debugging if you don't want to use the Kokkos tools.

Yurlungur (approved):

PR #321, opened 2020-10-12
make inner loop pattern tags constexpr. resolves #319

### LINKED ISSUE #319: Inner loop pattern tags not portable
I recently ran into an issue with `par_for_inner` default loop pattern. The issue is that the `inner_*_tag` variables are global and thus don't have host-device scope so dispatch can fail on lower optimization levels. Essentially, generically this fails:
```cuda
struct S {} s;

__host__ __device__ void bar(S) {
}

__host__ __device__ void foo(){
  bar(s);
}
```
but this succeeds:
```cuda
struct S {};
__device__ S s;

__host__ __device__ void bar(S) {
}

__host__ __device__ void foo(){
  bar(s);
}
```
This also succeeds
```cuda
struct S {};

__host__ __device__ void bar(S) {
}

__host__ __device__ void foo(){
  bar(S());
}
```
as does this
```C++
struct S {};
constexpr S s;

__host__ __device__ void bar(S) {
}

__host__ __device__ void foo(){
  bar(s);
}
```
We currently do the former and we need to change to one of the latter. See the discussion in the Kokkos chat [here](https://kokkosteam.slack.com/archives/C5BGU5NDQ/p1602271717063700)

Yurlungur:
## PR Summary

As discussed in issue #319, the inner loop pattern tag dispatch can fail on lower optimization levels, because the tags do not have `device` scope. One way to fix this is to add `__device__` to each tag, but this is not performance portable. Another way is to make the tags `constexpr` so that they can pull `device` from local scope when they are used. This implements the latter solution.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 - [[PR 293]](https://github.com/lanl/parthenon/pull/293) Changed `VariablePack` and related objects to use `ParArray1D` objects instead of `ParArrayND` objects under the hood to reduce the size of the captured objects.
 - [[PR 310]](https://github.com/lanl/parthenon/pull/310) Fix Cuda 11 builds.
 - [[PR 313]](https://github.com/lanl/parthenon/pull/313) Add include guards for Kokkos in cmake.
+- [[PR 321]](https://github.com/lanl/parthenon/pull/321) Make inner loop pattern tags constexpr
 
 ### Removed (removing behavior/API/varaibles/...)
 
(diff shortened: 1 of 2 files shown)


forrestglines (approved):
Looks good, compiles on my end.

JoshuaSBrown (approved):
Nice fix!

Yurlungur:
Cool. Pulling the trigger. Thanks.

PR #325, opened 2020-10-16
Fix multi rank convergence tests

### LINKED ISSUE #323: MPI results incorrect?
It looks like results (in the 3D regression test) differ for runs with different number of MPI procs.
```
36: -------------------------------------------
36:     Filename=advection_3d.out0.00001.phdf
36: -------------------------------------------
36:                 Time=1.0000
36:               NCycle=854
36:              NumDims=3
36:             MaxLevel=1
36:            NumBlocks=529
36:        MeshBlockSize=[8 8 8]
36:        CellsPerBlock=512
36:           TotalCells=270848
36:       TotalCellsReal=270848
36:                NumPE=4
36:          BlocksPerPE=[132 132 132 133]
36:               NGhost=2
36:        IncludesGhost=0
36:          Coordinates=b'UniformCartesian'
36: --------------------------------------------
36:            Variables=['Info', 'Locations', 'advected', 'one_minus_advected', 'one_minus_advected_sq', 'one_minus_sqrt_one_minus_advected_sq_12', 'one_minus_sqrt_one_minus_advected_sq_37']
36: --------------------------------------------
36: 
36: 
36: -------------------------------------------
36:     Filename=/home/pgrete/src/parthenon-master/tst/regression/test_suites/output_hdf5/../../../../tst/regression/gold_standard/advection_3d.out0.00001.phdf
36: -------------------------------------------
36:                 Time=1.0000
36:               NCycle=214
36:              NumDims=3
36:             MaxLevel=1
36:            NumBlocks=204
36:        MeshBlockSize=[8 8 8]
36:        CellsPerBlock=512
36:           TotalCells=104448
36:       TotalCellsReal=104448
36:                NumPE=1
36:          BlocksPerPE=[204]
36:               NGhost=2
36:        IncludesGhost=0
36:          Coordinates=b'UniformCartesian'
36: --------------------------------------------
36:            Variables=['Info', 'Locations', 'advected', 'one_minus_advected', 'one_minus_advected_sq', 'one_minus_sqrt_one_minus_advected_sq_12', 'one_minus_sqrt_one_minus_advected_sq_37']
36: --------------------------------------------
```
2D seems to be fine.

pgrete:
## PR Summary

#281 introduced a dependency of the domain size with number of MPI ranks used for the regression tests.
This resulted in convergence tests failing when using different number of MPI ranks.

This PR fixes this while keeping the original intent, i.e., when more MPI ranks are used the blocks are getting smaller but the domain size remains constant.
This does not apply to AMR tests, as all the current tests already include enough blocks so that they can be distributed among all ranks.
In addition, the MPI tests in the CI suite now run with 2 processes (versus previously just one) so that this is automatically tested from now on.

Closes #323 

## PR Checklist

- [X] Code passes cpplint
- [does not apply] New features are documented.
- [X] Code is formatted

### DIFF
--- a/tst/regression/test_suites/advection_convergence/advection_convergence.py
+++ b/tst/regression/test_suites/advection_convergence/advection_convergence.py
@@ -28,6 +28,7 @@ import utils.test_case
 """ To prevent littering up imported folders with .pyc files or __pycache_ folder"""
 sys.dont_write_bytecode = True
 
+# if this is updated make sure to update the assert statements for the number of MPI ranks, too
 lin_res = [32, 64, 128, 256, 512] # resolution for linear convergence
 
 class TestCase(utils.test_case.TestCaseAbs):
@@ -58,6 +59,10 @@ class TestCase(utils.test_case.TestCaseAbs):
 
 
         n_res = len(lin_res)
+        # make sure we can evenly distribute the MeshBlock sizes
+        assert parameters.num_ranks % 2 == 0, "Num ranks must be multiples of 2 for convergence test."
+        # ensure a minimum block size of 4
+        assert lin_res[0] / parameters.num_ranks <= 8, "Use <= 8 ranks for convergence test."
 
         # TEST: Advection only in x-direction 
         # using nx2 = nx3 = 4 > 1 for identical errors between dimensions
@@ -66,8 +71,8 @@ class TestCase(utils.test_case.TestCaseAbs):
             if lin_res[step - 1] == 32:
                 parameters.coverage_status = "both"
             parameters.driver_cmd_line_args = [
-                'parthenon/mesh/nx1=%d' % (lin_res[step % n_res -1] * parameters.num_ranks),
-                'parthenon/meshblock/nx1=%d' % lin_res[step % n_res -1],
+                'parthenon/mesh/nx1=%d' % lin_res[step % n_res -1],
+                'parthenon/meshblock/nx1=%d' % (lin_res[step % n_res -1] // parameters.num_ranks),
                 'parthenon/mesh/nx2=1',
                 'parthenon/meshblock/nx2=1',
                 'parthenon/mesh/nx3=1',
@@ -83,8 +88,8 @@ class TestCase(utils.test_case.TestCaseAbs):
             parameters.driver_cmd_line_args = [
                 'parthenon/mesh/nx1=4',
                 'parthenon/meshblock/nx1=4',
-                'parthenon/mesh/nx2=%d' % (lin_res[step % n_res -1] * parameters.num_ranks),
-                'parthenon/meshblock/nx2=%d' % lin_res[step % n_res -1],
+                'parthenon/mesh/nx2=%d' % lin_res[step % n_res -1],
+                'parthenon/meshblock/nx2=%d' % (lin_res[step % n_res -1] // parameters.num_ranks),
                 'parthenon/mesh/nx3=1',
                 'parthenon/meshblock/nx3=1',
                 'Advection/vx=0.0',
@@ -100,8 +105,8 @@ class TestCase(utils.test_case.TestCaseAbs):
                 'parthenon/meshblock/nx1=4',
                 'parthenon/mesh/nx2=4',
                 'parthenon/meshblock/nx2=4',
-                'parthenon/mesh/nx3=%d' % (lin_res[step % n_res -1] * parameters.num_ranks),
-                'parthenon/meshblock/nx3=%d' % lin_res[step % n_res -1],
+                'parthenon/mesh/nx3=%d' % lin_res[step % n_res -1],
+                'parthenon/meshblock/nx3=%d' % (lin_res[step % n_res -1] // parameters.num_ranks),
                 'Advection/vx=0.0',
                 'Advection/vy=0.0',
                 'Advection/ang_2_vert=true',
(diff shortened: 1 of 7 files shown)


pgrete:
I also just manually triggered the full pipeline. Let's see how it goes.

tst/regression/test_suites/advection_performance/advection_performance.py
```diff
@@ -33,12 +33,35 @@
...
+
+            # fill to 3 dims
+            while len(factors) < 3:
+                factors.append(1)
+            # split to 3 dims
+            while len(factors) > 3:
+                factors = sorted(factors[:-2] + [factors[-2]*factors[-1]], reverse=True)
+            return sorted(factors, reverse=True)
+
+        num_proc_x, num_proc_y, num_proc_z = get_split(parameters.num_ranks)
```
Yurlungur, line 57: It feels like there should be a built-in for this. But on a quick google I couldn't find one. 🤷

Yurlungur (approved):
Nice fix. I think this nicely keeps the original intent of @JoshuaSBrown 's MR while fixing discovered issues.

tst/regression/test_suites/restart/restart.py
```diff
@@ -38,8 +38,7 @@ def Prepare(self, parameters, step):
             parameters.driver_cmd_line_args = [
                 '-r',
                 'gold.out0.00001.rhdf',
-                'parthenon/job/problem_id=silver',
-                'parthenon/mesh/nx1=%d' % (64 * parameters.num_ranks)
+                'parthenon/job/problem_id=silver'
```
JoshuaSBrown, line 41: Does this mean the restart regression test can only be used with a singe gpu/proc?

JoshuaSBrown: Nevermind, I just answered my own question.

tst/regression/test_suites/advection_convergence/advection_convergence.py
```diff
@@ -58,6 +59,10 @@ def Prepare(self,parameters, step):
 
 
         n_res = len(lin_res)
+        # make sure we can evenly distribute the MeshBlock sizes
+        assert parameters.num_ranks % 2 == 0, "Num ranks must be multiples of 2 for convergence test."
+        # ensure a minimum block size of 4
+        assert lin_res[0] / parameters.num_ranks <= 8, "Use <= 8 ranks for convergence test."
```
JoshuaSBrown: Do we need similar checks like this in the other regression tests?

JoshuaSBrown:
It's not immediately obvious to me what the problem was with the original code, @pgrete could you explain what assumptions I made that was causing this to fail? Is the mesh/nx values tied to the physical size of the problem?

JoshuaSBrown (approved):

tst/regression/test_suites/advection_convergence/advection_convergence.py
pgrete: I don't think so because the other tests are AMR tests that have many MeshBlocks anyway (so unless you try to run the test on, say, more than 100 ranks this should not come up)

pgrete:
> It's not immediately obvious to me what the problem was with the original code, @pgrete could you explain what assumptions I made that was causing this to fail? Is the mesh/nx values tied to the physical size of the problem?

The failing of the tests were actually caused by two separate problems.

1. For static grids the convergence tests (and the way we measure the error) depends on many different factors including the effective  `dx_1`,`dx_2`,and `dx_3` (`dx_#`). This is actually what happens for the convergence test. The problem itself is identical only the resolution (here grid spacing) changes and the error should go down. This is independent of how many MPI ranks are used to run the simulation as the answer should be independent of the number of ranks.
So when `mesh/nx#` are changed while keeping the physical/code lengths constant (`mesh/x#min`/`mesh/x#max`) this changes the effective `dx_#` and thus the answers change.
Thus, now the overall grid dimensions and lengths  are constant (-> problem is identical) and when you use more and more ranks the `MeshBlocks` are getting smaller and smaller so that there are enough `MeshBlocks` to distribute among all ranks.

2. For AMR the issues is even more subtle as the refinement criteria often depend on `dx_#`. Thus, when  `mesh/nx#` and therefore `dx_#` are changed then refinement may be triggered in a different way leading to slightly different solutions.
Those are not more right or wrong than the original problem, but it's a different setup (which is caught by the tests failing given the way we set those up).

I hope this helps. Does this answer your question?

pgrete:
Extended test suite now passed, too. Merging.

JoshuaSBrown:
> > It's not immediately obvious to me what the problem was with the original code, @pgrete could you explain what assumptions I made that was causing this to fail? Is the mesh/nx values tied to the physical size of the problem?
> 
> The failing of the tests were actually caused by two separate problems.
> 
>     1. For static grids the convergence tests (and the way we measure the error) depends on many different factors including the effective  `dx_1`,`dx_2`,and `dx_3` (`dx_#`). This is actually what happens for the convergence test. The problem itself is identical only the resolution (here grid spacing) changes and the error should go down. This is independent of how many MPI ranks are used to run the simulation as the answer should be independent of the number of ranks.
>        So when `mesh/nx#` are changed while keeping the physical/code lengths constant (`mesh/x#min`/`mesh/x#max`) this changes the effective `dx_#` and thus the answers change.
>        Thus, now the overall grid dimensions and lengths  are constant (-> problem is identical) and when you use more and more ranks the `MeshBlocks` are getting smaller and smaller so that there are enough `MeshBlocks` to distribute among all ranks.
> 
>     2. For AMR the issues is even more subtle as the refinement criteria often depend on `dx_#`. Thus, when  `mesh/nx#` and therefore `dx_#` are changed then refinement may be triggered in a different way leading to slightly different solutions.
>        Those are not more right or wrong than the original problem, but it's a different setup (which is caught by the tests failing given the way we set those up).
>
> 
> I hope this helps. Does this answer your question?

Thanks for the explanation, that was enlightening. Sorry for causing extra work, but thanks!

PR #328, opened 2020-10-21
Unify Containers and MeshBlockPacks

jdolence:
## PR Summary

Here's a stab at what was discussed for generalizing the idea of `Container`s (now `MeshBlockData`) to the `Mesh` level with the `MeshData` class.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


jdolence:
@pgrete @Yurlungur you might be interested to take a look.  The diff is probably hard to parse because it includes @pgrete's commits in the branch I based this off of.  Should I target that branch for merging?

Yurlungur:
> @pgrete @Yurlungur you might be interested to take a look.  The diff is probably hard to parse because it includes @pgrete's commits in the branch I based this off of.  Should I target that branch for merging?

I'll take a look tomorrow. Not sure which diff will be easier to parse, since I haven't looked at @pgrete 's branch yet either

pgrete:
Good morning/evening :smile: 
I think it'll be easier to jointly work on this branch (potentially neglecting the raw buffer pack-in-one code), as all changes will come one way or the other, and try to get these changes into develop quickly.
Also, here's the diff for the raw changes to the other branch: https://github.com/lanl/parthenon/compare/c48455eaae1...804b852

pgrete:
FYI the changes I introduced earlier are incorrect for some reason.
The code still runs but the results are not deterministic any more (I discovered this while doing scaling tests with this branch).
I'm not sure whether I added buggy code or exposed some unintended behavior.

jdolence:
@pgrete are you sure the problem is your changes?  I touched _lots_ of code.  Are you referring to your changes before or after my commits?

pgrete:
I'm fairly sure that at least sth I added (likely around the flexible number of variables in the advection vector) did something.
Your original commit of yesterday evening seems to result in the same outputs as current develop.

jdolence:
and the nondeterminism is observed even when `num_vars=1`?  It doesn't look like you initialize any more than one in the ProblemGenerator (unless I just missed this in the diff), but you use all the vars to do AMR tagging, for example.

pgrete:
Alright, it looks like it's not the `num_vars`, but  the `refine_interval` is causing the issue.
I'm just not quite sure why.

docs/interface/Metadata.md
```diff
@@ -100,7 +100,7 @@ classes may be allocated. The behaviours are the following:
   MPI communication buffers are allocated, and buffers for a coarse
   grid are allocated. These buffers are *one-copy*, meaning they are
   shared between all instances of a variable in all `Containers` in a
-  `ContainerCollection`.
+  `DataCollection`.
```
Yurlungur, line 103: Should this be `MeshDataCollection`?

docs/interface/state.md
```diff
@@ -54,10 +54,10 @@ The ```SparseVariable``` class is designed to support multi-component state wher
...
 
-# ContainerCollection
+# DataCollection
 
-The ```ContainerCollection``` class is the highest level abstraction in Parthenon's state management.  Each ```MeshBlock``` in a simulation owns a ```ContainerCollection``` that through the classes just described, manages all of the simulation data.  Every ```ContainerCollection``` is initialized with a ```Container``` named ```"base"```.  The ```Get``` function, when invoked without arguments, returns a reference to this base ```Container``` which is intended to contain all of the simulation data that persists between timesteps (if applicable).
+The ```DataCollection``` class is the highest level abstraction in Parthenon's state management.  Each ```MeshBlock``` in a simulation owns a ```DataCollection``` that through the classes just described, manages all of the simulation data.  Every ```DataCollection``` is initialized with a ```Container``` named ```"base"```.  The ```Get``` function, when invoked without arguments, returns a reference to this base ```Container``` which is intended to contain all of the simulation data that persists between timesteps (if applicable).
```
Yurlungur: presumably container should now be `MeshBlockData` here. Also are these `MeshBlockDataCollection`s? Or `DataCollections`? What about `MeshDataCollection`?

pgrete:
Something about the caching doesn't work as expected yet.
One place were extra memcopies are introduced is here:
```
#0  0x0000000000568440 in cudaMemcpy ()
#1  0x00000000005209c7 in Kokkos::Impl::DeepCopy<Kokkos::CudaSpace, Kokkos::HostSpace, Kokkos::Cuda>::DeepCopy (
    this=<optimized out>, dst=<optimized out>, src=<optimized out>, n=<optimized out>)
    at /mnt/home/gretephi/src/parthenon-master/external/Kokkos/core/src/Cuda/Kokkos_CudaSpace.cpp:98
#2  0x00000000005222be in Kokkos::Impl::SharedAllocationRecord<Kokkos::CudaSpace, void>::SharedAllocationRecord (
    this=0x1ee1ac0, arg_space=..., arg_label=..., arg_alloc_size=4608, arg_dealloc=<optimized out>)
    at /mnt/home/gretephi/src/parthenon-master/external/Kokkos/core/src/Cuda/Kokkos_CudaSpace.cpp:521
#3  0x0000000000492bf4 in Kokkos::Impl::SharedAllocationRecord<Kokkos::CudaSpace, Kokkos::Impl::ViewValueFunctor<Kokkos::Cuda, parthenon::VariablePack<double>, false> >::SharedAllocationRecord (arg_alloc=4608, arg_label=..., arg_space=..., 
    this=0x1ee1ac0) at /mnt/home/gretephi/src/parthenon-master/external/Kokkos/core/src/impl/Kokkos_SharedAlloc.hpp:224
#4  Kokkos::Impl::SharedAllocationRecord<Kokkos::CudaSpace, Kokkos::Impl::ViewValueFunctor<Kokkos::Cuda, parthenon::VariablePack<double>, false> >::allocate (arg_alloc=<optimized out>, arg_label=..., arg_space=...)
    at /mnt/home/gretephi/src/parthenon-master/external/Kokkos/core/src/impl/Kokkos_SharedAlloc.hpp:246
#5  Kokkos::Impl::ViewMapping<Kokkos::ViewTraits<parthenon::VariablePack<double>*, Kokkos::LayoutRight, Kokkos::CudaSpace>, void>::allocate_shared<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::integral_constant<unsigned int, 0u>, Kokkos::CudaSpace, Kokkos::Cuda> (arg_layout=..., arg_prop=..., this=0x7fffffff3468)
    at /mnt/home/gretephi/src/parthenon-master/external/Kokkos/core/src/impl/Kokkos_ViewMapping.hpp:3171
#6  Kokkos::View<parthenon::VariablePack<double>*, Kokkos::LayoutRight, Kokkos::CudaSpace>::View<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >(Kokkos::Impl::ViewCtorProp<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > const&, std::enable_if<!Kokkos::Impl::ViewCtorProp<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >::has_pointer, Kokkos::LayoutRight>::type const&) (arg_layout=..., arg_prop=..., 
    this=0x7fffffff3460) at /mnt/home/gretephi/src/parthenon-master/external/Kokkos/core/src/Kokkos_View.hpp:1976
#7  Kokkos::View<parthenon::VariablePack<double>*, Kokkos::LayoutRight, Kokkos::CudaSpace>::View<char [31]> (
    arg_N7=18446744073709551615, arg_N6=18446744073709551615, arg_N5=18446744073709551615, arg_N4=18446744073709551615, 
    arg_N3=18446744073709551615, arg_N2=18446744073709551615, arg_N1=18446744073709551615, arg_N0=64, arg_label=..., 
    this=0x7fffffff3460) at /mnt/home/gretephi/src/parthenon-master/external/Kokkos/core/src/Kokkos_View.hpp:2098
#8  parthenon::MeshData<double>::PackVariables<std::vector<parthenon::MetadataFlag, std::allocator<parthenon::MetadataFlag> >&> (this=0x1cd7890) at /mnt/home/gretephi/src/parthenon-master/src/interface/mesh_data.hpp:82
#9  0x00000000004877cd in parthenon::Update::FluxDivergenceMesh (in_obj=..., dudt_obj=...)
    at /opt/software/GCCcore/9.3.0/include/c++/9.3.0/bits/shared_ptr_base.h:1020
```
i.e., the pack that should exist in that function (potentially from the previous cycle) is not found and newly created.

jdolence:
> Something about the caching doesn't work as expected yet.
> One place were extra memcopies are introduced is here:
> 
> ```
> #0  0x0000000000568440 in cudaMemcpy ()
> #1  0x00000000005209c7 in Kokkos::Impl::DeepCopy<Kokkos::CudaSpace, Kokkos::HostSpace, Kokkos::Cuda>::DeepCopy (
>     this=<optimized out>, dst=<optimized out>, src=<optimized out>, n=<optimized out>)
>     at /mnt/home/gretephi/src/parthenon-master/external/Kokkos/core/src/Cuda/Kokkos_CudaSpace.cpp:98
> #2  0x00000000005222be in Kokkos::Impl::SharedAllocationRecord<Kokkos::CudaSpace, void>::SharedAllocationRecord (
>     this=0x1ee1ac0, arg_space=..., arg_label=..., arg_alloc_size=4608, arg_dealloc=<optimized out>)
>     at /mnt/home/gretephi/src/parthenon-master/external/Kokkos/core/src/Cuda/Kokkos_CudaSpace.cpp:521
> #3  0x0000000000492bf4 in Kokkos::Impl::SharedAllocationRecord<Kokkos::CudaSpace, Kokkos::Impl::ViewValueFunctor<Kokkos::Cuda, parthenon::VariablePack<double>, false> >::SharedAllocationRecord (arg_alloc=4608, arg_label=..., arg_space=..., 
>     this=0x1ee1ac0) at /mnt/home/gretephi/src/parthenon-master/external/Kokkos/core/src/impl/Kokkos_SharedAlloc.hpp:224
> #4  Kokkos::Impl::SharedAllocationRecord<Kokkos::CudaSpace, Kokkos::Impl::ViewValueFunctor<Kokkos::Cuda, parthenon::VariablePack<double>, false> >::allocate (arg_alloc=<optimized out>, arg_label=..., arg_space=...)
>     at /mnt/home/gretephi/src/parthenon-master/external/Kokkos/core/src/impl/Kokkos_SharedAlloc.hpp:246
> #5  Kokkos::Impl::ViewMapping<Kokkos::ViewTraits<parthenon::VariablePack<double>*, Kokkos::LayoutRight, Kokkos::CudaSpace>, void>::allocate_shared<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::integral_constant<unsigned int, 0u>, Kokkos::CudaSpace, Kokkos::Cuda> (arg_layout=..., arg_prop=..., this=0x7fffffff3468)
>     at /mnt/home/gretephi/src/parthenon-master/external/Kokkos/core/src/impl/Kokkos_ViewMapping.hpp:3171
> #6  Kokkos::View<parthenon::VariablePack<double>*, Kokkos::LayoutRight, Kokkos::CudaSpace>::View<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >(Kokkos::Impl::ViewCtorProp<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > const&, std::enable_if<!Kokkos::Impl::ViewCtorProp<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >::has_pointer, Kokkos::LayoutRight>::type const&) (arg_layout=..., arg_prop=..., 
>     this=0x7fffffff3460) at /mnt/home/gretephi/src/parthenon-master/external/Kokkos/core/src/Kokkos_View.hpp:1976
> #7  Kokkos::View<parthenon::VariablePack<double>*, Kokkos::LayoutRight, Kokkos::CudaSpace>::View<char [31]> (
>     arg_N7=18446744073709551615, arg_N6=18446744073709551615, arg_N5=18446744073709551615, arg_N4=18446744073709551615, 
>     arg_N3=18446744073709551615, arg_N2=18446744073709551615, arg_N1=18446744073709551615, arg_N0=64, arg_label=..., 
>     this=0x7fffffff3460) at /mnt/home/gretephi/src/parthenon-master/external/Kokkos/core/src/Kokkos_View.hpp:2098
> #8  parthenon::MeshData<double>::PackVariables<std::vector<parthenon::MetadataFlag, std::allocator<parthenon::MetadataFlag> >&> (this=0x1cd7890) at /mnt/home/gretephi/src/parthenon-master/src/interface/mesh_data.hpp:82
> #9  0x00000000004877cd in parthenon::Update::FluxDivergenceMesh (in_obj=..., dudt_obj=...)
>     at /opt/software/GCCcore/9.3.0/include/c++/9.3.0/bits/shared_ptr_base.h:1020
> ```
> 
> i.e., the pack that should exist in that function (potentially from the previous cycle) is not found and newly created.

I'm looking...

jdolence:
@pgrete I think I fixed the caching.  Seems to work for me now.

pgrete:
Great @jdolence !
I'm happy to confirm that this fixed it:
```
Generating CUDA Memory Operation Statistics...
CUDA Memory Operation Statistics (nanoseconds)

Time(%)      Total Time  Operations         Average         Minimum         Maximum  Name                                                                            
-------  --------------  ----------  --------------  --------------  --------------  --------------------------------------------------------------------------------
  100.0       133291237      125972          1058.1             960           13696  [CUDA memcpy DtoD] 
```

With #336 we should find these things automatically in the future.

pgrete:
@brryan I was about to merge current develop in here and but turns out that the merge conflicts are significant because of the swarm containers.
Given that haven't looked at the particle implementation in more detail, I was wondering if you could have a look at the current state of this PR and let us know the implications for the particle implementation.
I'm specifically wondering if a "SwarmContainer" is still required or if the new implementation of the abstract MeshBlockData class may serve as replacement.

docs/interface/containers.md
```diff
@@ -57,12 +57,12 @@ Similar methods are available for metadata and sparse IDs:
 container.PackVariables(metadata, ids, map)
 container.PackVariables(metadata, map)
```
Yurlungur: `data.*`?

docs/interface/state.md
```diff
@@ -54,10 +54,10 @@ The ```SparseVariable``` class is designed to support multi-component state wher
...
+The ```Add(const std::string& label, MeshBlockData<T>& src)``` member function creates a new ```Container``` with the provided label.  This new ```Container``` is populated with all of the variables in ```src```.  When a variable has the ```Metadata::OneCopy``` flag set, the variables in the new ```Container``` are are just shallow copies from ```src```, i.e. no new storage for data is allocated, the ```std::shared_ptr``` to the variable is just copied.  For variables that do not have ```Metadata::OneCopy``` set, new storage is allocated.  Once created, these new containers are accesible by calling ```Get``` with the name of the desired ```Container``` as an argument.  NOTE: The ```Add``` function checks if a ```Container``` by the name ```label``` already exists in the collection, immediately returning if one is found (or throwing a ```std::runtime_error``` if the new and pre-existing containers are not equivalent).  Therefore, adding a ```Container``` to the collection multiple times results in a single new container, with the remainder of the calls no-ops.
```
Yurlungur: `Container` still referenced here.

example/advection/advection_driver.cpp
```diff
@@ -44,106 +47,155 @@ AdvectionDriver::AdvectionDriver(ParameterInput *pin, ApplicationInput *app_in,
...
+      auto md = pmesh->mesh_data.Add(name);
+      md->Set(partition, mb_name);
+    };
+    for (int i = 0; i < partitions.size(); i++) {
+      setup_mesh_data("base" + std::to_string(i), partitions[i], "base");
+      setup_mesh_data("dUdt" + std::to_string(i), partitions[i], "dUdt");
+      for (int j = 1; j < integrator->nstages; j++) {
+        setup_mesh_data(stage_name[j] + std::to_string(i), partitions[i], stage_name[j]);
+      }
+    }
```
Yurlungur: This is a little painful. I wonder if there's a way to automate it, or at least add a bit more syntactic sugar.

example/advection/advection_driver.cpp
```diff
@@ -44,106 +47,155 @@ AdvectionDriver::AdvectionDriver(ParameterInput *pin, ApplicationInput *app_in,
...
   if (stage == 1) {
-    auto &base = pmb->real_containers.Get();
-    pmb->real_containers.Add("dUdt", base);
-    for (int i = 1; i < integrator->nstages; i++)
-      pmb->real_containers.Add(stage_name[i], base);
+    auto setup_mesh_data = [=](const std::string &name, BlockList_t partition,
+                               const std::string &mb_name) {
+      auto md = pmesh->mesh_data.Add(name);
+      md->Set(partition, mb_name);
+    };
```
Yurlungur: Maybe `MeshDataCollection` should just have an overloaded `Add` function that does this?

example/advection/advection_driver.cpp
```diff
@@ -44,106 +47,155 @@ AdvectionDriver::AdvectionDriver(ParameterInput *pin, ApplicationInput *app_in,
...
+    auto &mc1 = pmesh->mesh_data.Get(stage_name[stage] + std::to_string(i));
+    auto &mdudt = pmesh->mesh_data.Get("dUdt" + std::to_string(i));
+
+    // compute the divergence of fluxes of conserved variables
+    auto flux_div = tl.AddTask(none, parthenon::Update::FluxDivergenceMesh, mc0, mdudt);
+
+    // apply du/dt to all independent fields in the container
+    auto update =
+        tl.AddTask(flux_div, UpdateMeshData, stage, integrator, mc0, mbase, mdudt, mc1);
+  }
```
Yurlungur: I like how this came out. It ended up looking fairly nice.

example/advection/advection_package.cpp
```diff
@@ -130,30 +130,36 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
...
   std::string field_name = "advected";
-  Metadata m({Metadata::Cell, Metadata::Independent, Metadata::FillGhost});
+  Metadata m({Metadata::Cell, Metadata::Independent, Metadata::FillGhost},
+             std::vector<int>({num_vars}));
   pkg->AddField(field_name, m);
 
   field_name = "one_minus_advected";
-  m = Metadata({Metadata::Cell, Metadata::Derived, Metadata::OneCopy});
+  m = Metadata({Metadata::Cell, Metadata::Derived, Metadata::OneCopy},
+               std::vector<int>({num_vars}));
```
Yurlungur, line 143: good improvement here

example/advection/advection_package.cpp
```diff
@@ -163,7 +169,7 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
   return pkg;
 }
 
-AmrTag CheckRefinement(std::shared_ptr<Container<Real>> &rc) {
+AmrTag CheckRefinement(std::shared_ptr<MeshBlockData<Real>> &rc) {
```
Yurlungur, line 172: Should these support `MeshData` too?

example/advection/advection_package.cpp
```diff
@@ -192,70 +201,76 @@ AmrTag CheckRefinement(std::shared_ptr<Container<Real>> &rc) {
 }
 
 // demonstrate usage of a "pre" fill derived routine
-void PreFill(std::shared_ptr<Container<Real>> &rc) {
+void PreFill(std::shared_ptr<MeshBlockData<Real>> &rc) {
```
Yurlungur, line 204: Same question... should these support `MeshData` *and* `MeshBlockData`?

example/advection/parthenon_app_inputs.cpp
```diff
@@ -32,7 +32,7 @@ using namespace parthenon;
 namespace advection_example {
 
 void ProblemGenerator(MeshBlock *pmb, ParameterInput *pin) {
-  auto &rc = pmb->real_containers.Get();
+  auto &rc = pmb->meshblock_data.Get();
```
Yurlungur, line 35: problem generators are an interesting case... these probably do need to be per meshblock.

example/calculate_pi/pi_driver.cpp
```diff
@@ -121,19 +121,26 @@ void PiDriver::PostExecute(Real pi_val) {
...
-  TaskRegion &async_region = tc.AddRegion(packs.size());
+  const int pack_size = pmesh->DefaultPackSize();
+  auto partitions = partition::ToSizeN(blocks, pack_size);
+  for (int i = 0; i < partitions.size(); i++) {
+    auto md = pmesh->mesh_data.Add(std::to_string(i));
+    md->Set(partitions[i], "base");
+  }
+
+  ParArrayHost<Real> areas("areas", partitions.size());
+  TaskRegion &async_region = tc.AddRegion(partitions.size());
```
Yurlungur, line 137: Overall I like how simple the pi driver came out too. Although again, I think the `set` loop thing should just be overloaded in `mesh_data`.

src/bvals/bvals_interfaces.hpp
```diff
@@ -266,6 +266,7 @@ class BoundaryVariable : public BoundaryCommunication, public BoundaryBuffer {
   bool ReceiveBoundaryBuffers() override;
   void ReceiveAndSetBoundariesWithWait() override;
   void SetBoundaries() override;
+  auto GetBdVar() { return &bd_var_; }
```
Yurlungur: should this be `GetPBdVar` to indicate you get a pointer?

src/interface/data_collection.hpp
```diff
@@ -10,29 +10,36 @@
...
 #include <string>
 
-#include "interface/container.hpp"
-#include "interface/metadata.hpp"
-
 namespace parthenon {
 
 template <typename T>
-class ContainerCollection {
+class DataCollection {
```
Yurlungur, line 31: I like that `DataCollection` is templated on `MeshData` and `MeshBlockData`. That's a nice simplification. Should we have aliases for each? i.e., somewhere at the bottom of this file,
```C++
using MeshDataCollection = DataCollection<MeshBlockData<Real>>;
using MeshBlockDataCollection = DataCollection<MeshBlockData<Real>>;
```

src/interface/mesh_data.hpp
```diff
@@ -0,0 +1,167 @@
...
+      Kokkos::deep_copy(packs, packs_host);
+      Kokkos::deep_copy(coords, coords_host);
+
+      auto cellbounds = block_data_[0]->GetBlockPointer()->cellbounds;
+      auto mbp = MeshBlockVarFluxPack<T>(packs, cellbounds, coords, dims);
+      varFluxPackMap_[key] = mbp;
+      return mbp;
+    }
+    return kvpair->second;
+  }
```
Yurlungur: There's a fair amount of duplicated code between these two methods, which you can get around if you re-introduce my `PackingFunction` template abstraction. Or does the caching ruin that little trick?

src/interface/meshblock_data.cpp
```diff
@@ -122,7 +124,7 @@ Container<T>::Container(const Container<T> &src, const std::vector<std::string>
     auto sv = sparse_map.find(name);
     if (sv != sparse_map.end()) {
       if (found) {
-        std::cerr << "Container: " << name << " found more than once!" << std::endl;
+        std::cerr << "MeshBlockData: " << name << " found more than once!" << std::endl;
         std::exit(1);
```
Yurlungur: Since we're here...
```suggestion
        std::stringstream msg;
        msg << "MeshBlockData: " << name << " found more than once!" << std::endl;
        PARTHENON_TRHOW(msg);
```

src/interface/meshblock_data.cpp
```diff
@@ -137,20 +139,21 @@ Container<T>::Container(const Container<T> &src, const std::vector<std::string>
     auto fv = face_map.find(name);
     if (fv != face_map.end()) {
       if (found) {
-        std::cerr << "Container: " << name << " found more than once!" << std::endl;
+        std::cerr << "MeshBlockData: " << name << " found more than once!" << std::endl;
         std::exit(1);
```
Yurlungur:
```suggestion
        std::stringstream msg;
        msg << "MeshBlockData: " << name << " found more than once!" << std::endl;
        PARTHENON_TRHOW(msg);
```

src/interface/meshblock_data.cpp
```diff
@@ -137,20 +139,21 @@ Container<T>::Container(const Container<T> &src, const std::vector<std::string>
...
       }
       found = true;
       Add(fv->second);
     }
     if (!found) {
-      std::cerr << "Container: " << name << " not found!" << std::endl;
+      std::cerr << "MeshBlockData: " << name << " not found!" << std::endl;
       std::exit(1);
```
Yurlungur:
```suggestion
      std::stringstream msg;
      msg << "MeshBlockData: " << name << " found more than once!" << std::endl;
      PARTHENON_TRHOW(msg);
```

src/interface/meshblock_data.cpp
```diff
@@ -226,11 +229,11 @@ std::shared_ptr<Container<T>> Container<T>::SparseSlice(int id) {
...
 ///        variable.
 template <typename T>
-VariableFluxPack<T> Container<T>::PackVariablesAndFluxesHelper_(
+VariableFluxPack<T> MeshBlockData<T>::PackVariablesAndFluxesHelper_(
     const std::vector<std::string> &var_names, const std::vector<std::string> &flx_names,
     const vpack_types::VarList<T> &vars, const vpack_types::VarList<T> &fvars,
-    PackIndexMap &vmap) {
-  auto key = std::make_pair(var_names, flx_names);
+    PackIndexMap &vmap, vpack_types::StringPair &key) {
+  key = std::make_pair(var_names, flx_names);
```
Yurlungur, line 240: Passing in the key optionally is a clever idea. That simplifies the caching machinery.

src/interface/meshblock_data.cpp
```diff
@@ -463,12 +574,12 @@ Container<T>::MakeList_(const std::vector<MetadataFlag> &flags,
 // TODO(JMM): this could be cleaned up, I think.
 // Maybe do only one loop, or do the cleanup at the end.
 template <typename T>
-void Container<T>::Remove(const std::string label) {
-  throw std::runtime_error("Container<T>::Remove not yet implemented");
+void MeshBlockData<T>::Remove(const std::string label) {
+  throw std::runtime_error("MeshBlockData<T>::Remove not yet implemented");
```
Yurlungur, line 582: Since this isn't in use, maybe just remove this entirely? Do we ever plan to add it at this point?

src/interface/update.cpp
```diff
@@ -64,45 +68,98 @@ TaskStatus FluxDivergence(std::shared_ptr<Container<Real>> &in,
   return TaskStatus::complete;
 }
 
-void UpdateContainer(std::shared_ptr<Container<Real>> &in,
-                     std::shared_ptr<Container<Real>> &dudt_cont, const Real dt,
-                     std::shared_ptr<Container<Real>> &out) {
+TaskStatus FluxDivergenceMesh(std::shared_ptr<MeshData<Real>> &in_obj,
```
Yurlungur, line 71: I really dislike having there be a separate `FluxDivergeneceMesh` and `FluxDivergenceBlock` is there not some way these can be the same function, just templated?

src/interface/update.cpp
```diff
@@ -64,45 +68,98 @@ TaskStatus FluxDivergence(std::shared_ptr<Container<Real>> &in,
...
+  std::vector<MetadataFlag> flags({Metadata::Independent});
+  auto vin = in_obj->PackVariablesAndFluxes(flags);
+  auto dudt = dudt_obj->PackVariables(flags);
+  const IndexRange ib = vin.cellbounds.GetBoundsI(interior);
+  const IndexRange jb = vin.cellbounds.GetBoundsJ(interior);
+  const IndexRange kb = vin.cellbounds.GetBoundsK(interior);
+
+  const int ndim = vin.GetNdim();
+  parthenon::par_for(
+      DEFAULT_LOOP_PATTERN, "FluxDivergenceMesh", DevExecSpace(), 0, vin.GetDim(5) - 1, 0,
```
Yurlungur, line 84: Why do we need to specify `DEFAULT_LOOP_PATTERN` here? Shouldn't it automatically know?

src/mesh/mesh.cpp
```diff
@@ -911,79 +907,6 @@ void Mesh::OutputMeshStructure(int ndim) {
...
-      RegisterMeshBlockPack(package_name, packer_name, packer_func);
-    }
-    for (auto &pair : fluxpackers) {
-      auto &packer_name = pair.first;
-      auto &packer_func = pair.second;
-      RegisterMeshBlockPack(package_name, packer_name, packer_func);
-    }
-  }
-}
-
```
Yurlungur, line 983: And good riddance

src/mesh/mesh.cpp
```diff
@@ -1326,4 +1248,15 @@ int Mesh::ReserveTagPhysIDs(int num_phys) {
 
 void Mesh::ReserveMeshBlockPhysIDs() { return; }
 
+std::int64_t Mesh::GetTotalCells() {
+  auto &pmb = block_list.front();
+  return static_cast<std::int64_t>(nbtotal) * pmb->block_size.nx1 * pmb->block_size.nx2 *
+         pmb->block_size.nx3;
+}
+// TODO(JMM): Move block_size into mesh.
```
Yurlungur, line 1257: TODO can go away now presumably

Yurlungur (commented):
Overall, I really like where this is going. I appreciate that the `MeshData` builds on `MeshBlockData` and that `DataCollection` is abstracted so there's minimal duplicated code.

I do not like that a bunch of functions for the `MeshData` and the `MeshBlockData` are separate. I'd much rather those be combined if possible. Some minor comments below.

docs/interface/Metadata.md
Yurlungur: Nevermind, I see how this works now.

src/bvals/bvals_interfaces.hpp
pgrete: yes, good point.
(FY this functionality is technically not required here but I'm using it for the buffer-pack-in-one)

example/advection/advection_driver.cpp
pgrete: Yes, I agree.
How do people feel about having an option to enable/disable automated partitioning?
I'd probably put that on the list of "things to add in a smaller PR later" (see other comments)

example/advection/advection_package.cpp
pgrete: Yes, I'd say let's handle them the same as the Fill functions, i.e., let the user choose to enroll whatever they like.

example/advection/advection_package.cpp
pgrete: Yes, as discussed in the Matrix room, we should support both. I'll add that.

src/interface/meshblock_data.cpp
pgrete: @AndrewGaspar do we potentially need this with "dense on block"? Otherwise, I also vote for letting it go.

src/interface/update.cpp
pgrete: I generally agree. 
As @jdolence mentioned this is likely more involved than it looks at first sight.
I could imagine a cleaner solution with updated wrappers (so that the wrappers now whether is Block or Mesh data and handling the accesses transparently) or trying to leverage the variable indexing of the `ParArrayNDs` to make this happen (though I'm not sure/didn't check if all the `ParArrayNDs` features survive the packing in general).
Here, I also vote for putting this on the list of things to fix in a separate PR later.

example/advection/advection_driver.cpp
pgrete: Yes, let me add that.

Yurlungur:
One more todo for this: delete my documentation for `MeshBlockPack` packing

src/interface/update.cpp
Yurlungur: I'm willing to fix it in a later PR, sure. So long as we make an issue to remind us to do it.

example/advection/advection_driver.cpp
Yurlungur: A later PR to add some automatic partitioning seems like a nice feature, yes.

example/advection/advection_driver.cpp
pgrete: While going further through the code, I'm now wondering if it would make sense to change the interface to Get/Set "Data" should include an integer (particularly the the blocks/partitions).
For the latter the use case I see will almost entirely work through indices so we could skip the step of creating label based on indices.
What do people think?

Yurlungur: I'm not quite sure I understand. What would the new calls look like?

pgrete: Currently strings are used in the example code as primary means to identify variable packs and encode lots of information and I'm wondering how robust this suggested use case (as we provide it in the example).
For example, the `stage_name`s (except for the `base`) are also `int` as are the partitions.
So what we have definitely works, but I see ways manually encoding information a string could break (or lead to unintended behavior).
I think what would be nice is sth. like a simple `pmesh->mesh_data.Get(stage_str, partition_int)` where the partitions are setup transparently without registering them (however, that may require more refactoring as these regions may run in parallel).

pgrete: At the expense of having a Mesh pointer in the DataCollection this now kind of works, see https://github.com/lanl/parthenon/compare/71b46aa...51e2b8e including updated doc
I'm happy to hear what people think about it. My main goal was to hide the complexity of creating/duplicating the information that is effectively already provided by creating `MeshBlockData`

example/advection/advection_driver.cpp
pgrete: The automated partitioning now works (in combination with automated MeshData setup), see other comment.

brryan:
@pgrete Sorry about the particles stuff. I think the interaction between particles and meshblockpacks is going to be pretty strong (one might want only one swarm per meshblockpack, for example).

The whole notion of a swarm_container is probably going away but that's a nontrivial change... do you want me to try my hand at merging develop in to this PR and just seeing whether I can hack the particles implementation to still work for now?

pgrete:
> @pgrete Sorry about the particles stuff. I think the interaction between particles and meshblockpacks is going to be pretty strong (one might want only one swarm per meshblockpack, for example).

Yes, I think so too, which is why I brought this up.
A "blind" (in terms of not know how you work with this downstream) merge and conflict fix from me may have caused additional work on your side.

> The whole notion of a swarm_container is probably going away but that's a nontrivial change... do you want me to try my hand at merging develop in to this PR and just seeing whether I can hack the particles implementation to still work for now?

It'd be great if you could work on the integration of develop in here. In terms of easier review (e.g., to see how the particle/swarm infrastructure is affected) it'd be great if you could work on a separate branch, i.e., derive from this one, merge develop into the new one, and then open a PR to merge the new one against this one.
I'll be able to provide feedback on that PR on short notice so that we can get that quickly in here and then merge this into develop.
This will also make it easier to add additional updates here (which I'm currently working on and that are unrelated to the swarm/particles).

docs/interface/containers.md
```diff
@@ -1,16 +1,16 @@
-# Containers, Container Iterators, and Variable Packs
+# MeshBlockDatas, MeshBlockData Iterators, and Variable Packs
 
-## Containers
+## MeshBlockDatas
```
AndrewGaspar:
```suggestion
## MeshBlockData
```

docs/interface/containers.md
```diff
@@ -1,16 +1,16 @@
-# Containers, Container Iterators, and Variable Packs
+# MeshBlockDatas, MeshBlockData Iterators, and Variable Packs
```
AndrewGaspar:
```suggestion
# MeshBlockData, MeshBlockData Iterators, and Variable Packs
```

docs/interface/containers.md
```diff
@@ -1,16 +1,16 @@
...
 
-## Containers
+## MeshBlockDatas
 
-A *container* owns *variables.* Each variable is named, and the
-container knows about various types of variables, such as whether the
+A *meshblock data container* owns *variables.* Each variable is named, and the
```
AndrewGaspar: Hm... lowercase meshblock, hard choice, but I think I have to go with:

```suggestion
A *mesh-block data container* owns *variables.* Each variable is named, and the
```

docs/interface/containers.md
```diff
@@ -1,16 +1,16 @@
...
+meshblock data container knows about various types of variables, such as whether the
```
AndrewGaspar: The formal style guide outlined in my previous ~~tweet~~ review comment requires mesh-block now

```suggestion
mesh-block data container knows about various types of variables, such as whether the
```

docs/interface/containers.md
```diff
@@ -1,16 +1,16 @@
...
 variable is for cell-centered data, face-centered data, sparse data,
 or dense data. (For more details on anonymous variables, see
-[here](Metadata.md).) Variables in a container can be different
+[here](Metadata.md).) Variables in a meshblock data container can be different
```
AndrewGaspar:
```suggestion
[here](Metadata.md).) Variables in a mesh-block data container can be different
```

docs/interface/containers.md
```diff
@@ -1,16 +1,16 @@
...
 shapes, e.g., scalar, tensor, etc.A variable can be added to a
-container as:
+meshblock data container as:
```
AndrewGaspar:
```suggestion
mesh-block data container as:
```

docs/interface/containers.md
```diff
@@ -34,7 +34,7 @@ lives, then shape is the shape of the full array. I.e., a
 shape = std::vector<int>({11,12,13,14});
 ```
 
-It is often desirable to extract from a container a specific set of
+It is often desirable to extract from a meshblock data container a specific set of
```
AndrewGaspar:
```suggestion
It is often desirable to extract from a mesh-block data container a specific set of
```

docs/interface/containers.md
```diff
@@ -57,12 +57,179 @@ Similar methods are available for metadata and sparse IDs:
 container.PackVariables(metadata, ids, map)
 container.PackVariables(metadata, map)
 ```
-If you would like all variables in a container, 
+If you would like all variables in a meshblock data container,
```
AndrewGaspar:
```suggestion
If you would like all variables in a mesh-block data container, 
```

docs/interface/containers.md
```diff
@@ -57,12 +57,179 @@ Similar methods are available for metadata and sparse IDs:
...
+needs to be registered as `MeshBlockData` first before it can be accessed through
+`MeshData` and `MeshBlockPacks`.
+
+## Registering `MeshData`
+
+`MeshData` is a lightweight object that aggregates multiple `MeshBlock`s.
+Therefore, it needs to be setup/registered with some number of `MeshBlock`s
+(at least one and at most all), which is referred to as partitioning.
+
+The `Partition` machinery found in [`utils/partition_stl_containers.hpp`](../../src/utils/partition_stl_containers.hpp).
```
AndrewGaspar: Sentence fragment.

docs/interface/containers.md
```diff
@@ -57,12 +57,179 @@ Similar methods are available for metadata and sparse IDs:
...
+
+Registration and partitioning can be controlled manually or automatically
+(recommended in multi-stage drivers).
+
+### Manual registration
+
+The following steps (used in the `calculate_pi` example [here](../../example/calculate_pi/pi_driver.cpp) ) are need to manually register and fill a `MeshData` object.
```
AndrewGaspar:
```suggestion
The following steps (used in the `calculate_pi` example [here](../../example/calculate_pi/pi_driver.cpp) ) are needed to manually register and fill a `MeshData` object.
```

docs/interface/containers.md
```diff
@@ -57,12 +57,179 @@ Similar methods are available for metadata and sparse IDs:
...
+```c++
+// Number of MeshBlocks per Partition
+const int pack_size = pmesh->DefaultPackSize();
+// Partition all blocks of the Mesh into separate partitions containing pack_size MeshBlocks
+auto partitions = partition::ToSizeN(pmesh->block_list, pack_size);
+// Register a MeshData object for each partition (collection of blocks) using the partition
+// number as label and containing references to the data stored in the "base" MeshBlockPack
+for (int i = 0; i < partitions.size(); i++) {
+  const std::string label = std::to_string(i);
+  auto mest_data = pmesh->mesh_data.Add(label);
```
AndrewGaspar:
```suggestion
  auto mesh_data = pmesh->mesh_data.Add(label);
```

docs/interface/containers.md
```diff
@@ -57,12 +57,179 @@ Similar methods are available for metadata and sparse IDs:
...
+
+// Splits container into partitions of size N
+template <typename T, typename Container_t>
+std::vector<std::vector<T>> ToSizeN(Container_t<T> &container, const int N);
+```
+Both functions live within the namespace `parthenon::partition` and `Partition_t`
+is defined as:
+```C++
+template<typename T>
+using Parition_t = std::vector<std::vector<T>>
```
AndrewGaspar:
```suggestion
using Partition_t = std::vector<std::vector<T>>
```

example/advection/parthenon_app_inputs.cpp
AndrewGaspar: It's just initialization, so hopefully it shouldn't matter from a performance perspective...

example/calculate_pi/pi_driver.cpp
AndrewGaspar: I didn't like that task lists were instantiated at the granularity of mesh block before, and I really don't like that we're instantiating them at the granularity of a partition/region now. I wish our method for describing task lists was more declarative, leaving more of the partitioning and scheduling decisions up to parthenon.


But obviously this is a discussion for another day, and making a refactor that big doesn't really solve any of our immediate issues.

src/driver/multistage.hpp
```diff
@@ -50,7 +51,7 @@ class MultiStageBlockTaskDriver : public MultiStageDriver {
   // An application driver that derives from this class must define this
   // function, which defines the application specific list of tasks and
   // there dependencies that must be executed.
-  virtual TaskList MakeTaskList(MeshBlock *pmb, int stage) = 0;
+  virtual auto MakeTaskCollection(BlockList_t &blocks, int stage) -> TaskCollection = 0;
```
AndrewGaspar: Why put the type on the right? I think right-hand return types should be reserved for situations where they can't be put on the left-hand side.

src/interface/mesh_data.hpp
```diff
@@ -0,0 +1,167 @@
...
+class Mesh;
+template <typename T>
+class MeshBlockData;
+
+template <typename T>
+class MeshData {
+ public:
+  MeshData() = default;
+  // MeshData(const Mesh *pmesh, const std::string &name);
+  // explicit MeshData(const Mesh *pmesh) : MeshData(pmesh, "base") { }
```
AndrewGaspar:
```suggestion
```

src/interface/mesh_data.hpp
```diff
@@ -0,0 +1,167 @@
...
+#include "utils/error_checking.hpp"
+
+namespace parthenon {
+
+class Mesh;
+template <typename T>
+class MeshBlockData;
+
+template <typename T>
+class MeshData {
```
AndrewGaspar, line 74: Could you add a documentation comment before each of the new types describing, at least in a brief way, the purpose of each type?

src/interface/data_collection.hpp
```diff
@@ -10,36 +10,49 @@
...
-    containers_["base"] = std::make_shared<Container<T>>(); // always add "base" container
+  DataCollection() {
+    containers_["base"] = std::make_shared<T>(); // always add "base" container
+    pmy_mesh_ = nullptr;
   }
 
-  void Add(const std::string &label, const std::shared_ptr<Container<T>> &src);
+  void SetMeshPointer(Mesh *pmesh) { pmy_mesh_ = pmesh; }
+
+  std::shared_ptr<T> Add(const std::string &label, const std::shared_ptr<T> &src);
```
AndrewGaspar, line 40: Doesn't seem like this is implemented anywhere?

src/interface/meshblock_data.hpp
```diff
@@ -64,19 +64,49 @@ class Container {
     return pmy_block.lock();
   }
 
+  void Copy(const std::shared_ptr<MeshBlockData<T>> src) {
```
AndrewGaspar:
```suggestion
  void Copy(const std::shared_ptr<MeshBlockData<T>> &src) {
```

src/interface/meshblock_data.hpp
```diff
@@ -247,6 +277,18 @@ class Container {
...
+                                             PackIndexMap &vmap,
+                                             vpack_types::StringPair &key);
+  VariableFluxPack<T> PackVariablesAndFluxes(const std::vector<std::string> &var_names,
+                                             const std::vector<std::string> &flx_names,
+                                             vpack_types::StringPair &key);
+  VariableFluxPack<T> PackVariablesAndFluxes(const std::vector<MetadataFlag> &flags,
+                                             PackIndexMap &vmap,
+                                             vpack_types::StringPair &key);
+  VariableFluxPack<T> PackVariablesAndFluxes(const std::vector<MetadataFlag> &flags,
+                                             vpack_types::StringPair &key);
```
AndrewGaspar, line 292: I think once you start getting this many overloads, it becomes very confusing. Can you change the names of the ones you added to better reflect the difference between each version of the function? Same with `PackVariables` below.

src/mesh/mesh.cpp
AndrewGaspar: Should it? Seems like `block_size` is still on `MeshBlock`.

src/mesh/meshblock.hpp
```diff
@@ -30,9 +30,9 @@
 #include "coordinates/coordinates.hpp"
 #include "defs.hpp"
 #include "domain.hpp"
-#include "interface/container.hpp"
-#include "interface/container_collection.hpp"
-#include "interface/update.hpp"
+#include "interface/data_collection.hpp"
+#include "interface/meshblock_data.hpp"
+//#include "interface/update.hpp"
```
AndrewGaspar:
```suggestion
```

AndrewGaspar (changes requested):

docs/interface/containers.md
Yurlungur: I don't know if I agree with this. I think `meshblock` has been the de facto standard both here and in the original Athena code.

docs/interface/containers.md
Yurlungur: Oops. This is my fault. It might actually be `Partition_t` in the code. Worth double-checking.

example/advection/advection_driver.cpp
Yurlungur: hmm interesting. Yeah, I think I like the idea of splitting it like that.

example/calculate_pi/pi_driver.cpp
Yurlungur: I have mixed feelings about this. Sometimes, I think these decisions are best left to the application developer. But that's perhaps not always true.

In any case, I think it's a big lift to disentangle if we want to go that route.

src/interface/meshblock_data.cpp
Yurlungur: And/or @jdolence do we need this for dense on block?

src/mesh/mesh.cpp
Yurlungur: oh I thought it moved. Nevermind.

src/interface/meshblock_data.hpp
Yurlungur: I don't think that's necessary... All of these functions essentially do the same thing. The difference is in what inputs vs. outputs they provide. I think renaming them would be more confusing.

example/calculate_pi/pi_driver.cpp
pgrete: Fortunately, if we ever decide this gets priority there's effectively only one place where downstream codes would need to adjust and given that the current implementation gets the job done I'm also in favor of putting this on the list of future improvements.

src/interface/update.cpp
pgrete: Keeping track in #349

docs/interface/containers.md
pgrete: I'm also in favor of keeping it to meshblock (or preferable even MeshBlock).

src/driver/multistage.hpp
pgrete: I followed what my IDE suggested ;) https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-trailing-return-type.html
I don't feel strongly about this so if you think this is causing problems or reduces readability I'm happy to change that.

src/interface/data_collection.hpp
pgrete: It's in the source file https://github.com/lanl/parthenon/blob/51e2b8e593243e0bb1b3307a8e36699983586765/src/interface/data_collection.cpp#L25

src/interface/meshblock_data.hpp
pgrete: @AndrewGaspar are you thinking of splitting this into `PackVariablesAndFluxesFromMetadata`, `PackVariablesAndFluxesFromStringLabels`, etc?
In that case, I'm in favor of keeping the interface as is given that the intention and result of all function is identical.

docs/interface/containers.md
pgrete: The code is correct. It was just the doc.

src/interface/meshblock_data.hpp
AndrewGaspar: It's fine, leave it.

docs/interface/containers.md
AndrewGaspar: If you're going to spell it out as one concatenated word, then it should be `MeshBlock`, preferably with the code-styling ticks. I don't think "meshblock" is a common enough word to compound it without a hyphen.

src/driver/multistage.hpp
AndrewGaspar: It's weird that this change was only applied to this one function. If we're making a style change, then we can discuss that, but if that's the preferred style, I don't understand why it was only made for this one function signature.

I don't think I really agree with this specific lint. The style is still pretty atypical and it's more verbose.

AndrewGaspar:
```suggestion
  virtual TaskCollection MakeTaskCollection(BlockList_t &blocks, int stage) = 0;
```

pgrete: Agreed. I'll revert.

src/interface/data_collection.hpp
pgrete: I added the typedefs to `mesh_data.hpp` and `meshblock_data.hpp` to prevent a circular dependency.

src/interface/update.cpp
pgrete: Because this is a `parthenon::par_for` and not a `pmb->par_for`.
Currently the abstraction only provides the "flexible" version and we set all the default in the MeshBlock member functions.

docs/interface/containers.md
pgrete: Fixed.

docs/interface/containers.md
pgrete: I now went with using "`MeshBlockData`" everywhere so that it's consistent with the implementation.

docs/interface/containers.md
pgrete: Fixed.

docs/interface/state.md
pgrete: Fixed. Now talking about "`MeshBlockData` containers" given the familiarity with the "container" concept but new naming.

docs/interface/state.md
pgrete: I updated this (probably only a temporary fix as I noticed that the overall docs are kind of split and/or duplicated between `state.md` and `container.md`).
Are you happy with the updated state for now and that we put a general overhaul of the docs on the todo list?

src/interface/mesh_data.hpp
pgrete: I think this https://github.com/lanl/parthenon/pull/328/commits/25fd028bb2d207df7d2445ef3bcf6efc149ad07a did the trick.

docs/interface/containers.md
Yurlungur: My thinking was `meshblock` or `mb` are instances and `MeshBlock` is the class.

pgrete: I don't have a strong opinion about the chosen style and am fine with the person with the strongest opinion just updating it.

src/interface/mesh_data.hpp
pgrete: Done.

pgrete:
@jdolence @jdolence @AndrewGaspar  after having stared at this for some more time I now suggest to make a final review of the changes that were introduced to address the comments and leave the update of the `CheckRefinement` and `FillDerived` functions for separate PRs.
My main reason is that those changes are likely more involved than we hoped for, e.g.,
- `CheckRefinement` will require either a reduction or an `Kokkos::View` for the `MeshBlockPacks` that stored the results for each individual MeshBlock
- `FillDerived` handles the `PreFill` `FillDerived` and `PostFill`  functions in different ways (in the same function call to support multiple packages) that is not immediate obvious to me any more why we decided on that approach. So I'd like to discuss this in a separate issue first before making changes and then review them in a PR that (hopefully) has less than 70 files touched and 1000+ lines of code changed.

What do you think?

AndrewGaspar (approved):

pgrete (approved):

AndrewGaspar:
> FillDerived handles the PreFill FillDerived and PostFill functions in different ways (in the same function call to support multiple packages) that is not immediate obvious to me any more why we decided on that approach.

I actually don't fully appreciate why we need `PreFill` and `PostFill`. Is it so packages can have dependencies on the `FillDerived` of other packages?

AndrewGaspar:
Actually, it can't be that, because PreFill and PostFIll aren't per-package, as pointed out.

Yurlungur:
I can take another pass through if you like. But from my point of view, let's get this in and fix incremental things still missing later.

docs/interface/containers.md
```diff
@@ -34,35 +34,202 @@ lives, then shape is the shape of the full array. I.e., a
...
+const int num_partitions = pmesh->DefaultNumPartitions();
+TaskRegion &single_tasklist_per_partition = tc.AddRegion(num_partitions);
+for (int i = 0; i < num_partitions; i++) {
+  auto &tl = single_tasklist_per_partition[i];
+  // "base" MeshBlockData of blocks in partition i
+  auto &mbase = pmesh->mesh_data.GetOrAdd("base", i);
+  // MeshBlockData of the previous stage of blocks in partition i
+  auto &mc0 = pmesh->mesh_data.GetOrAdd(stage_name[stage - 1], i);
+  // MeshBlockData of the current stage of blocks in partition i
+  auto &mc1 = pmesh->mesh_data.GetOrAdd(stage_name[stage], i);
```
Yurlungur, line 169: I like this.

example/advection/advection_driver.cpp
```diff
@@ -44,106 +47,139 @@ AdvectionDriver::AdvectionDriver(ParameterInput *pin, ApplicationInput *app_in,
...
+  const int num_partitions = pmesh->DefaultNumPartitions();
+  // note that task within this region that contains one tasklist per pack
+  // could still be executed in parallel
+  TaskRegion &single_tasklist_per_pack_region = tc.AddRegion(num_partitions);
+  for (int i = 0; i < num_partitions; i++) {
+    auto &tl = single_tasklist_per_pack_region[i];
+    auto &mbase = pmesh->mesh_data.GetOrAdd("base", i);
+    auto &mc0 = pmesh->mesh_data.GetOrAdd(stage_name[stage - 1], i);
+    auto &mc1 = pmesh->mesh_data.GetOrAdd(stage_name[stage], i);
+    auto &mdudt = pmesh->mesh_data.GetOrAdd("dUdt", i);
```
Yurlungur, line 117: much improved compared to the more manual method before.

example/advection/advection_driver.cpp
```diff
@@ -44,106 +47,139 @@ AdvectionDriver::AdvectionDriver(ParameterInput *pin, ApplicationInput *app_in,
...
+      // Update refinement
+      if (pmesh->adaptive) {
+        auto tag_refine = tl.AddTask(
+            fill_derived,
+            [](std::shared_ptr<MeshBlock> pmb) {
+              pmb->pmr->CheckRefinementCondition();
+              return TaskStatus::complete;
+            },
+            pmb);
+      }
```
Yurlungur, line 179: In a future MR, we should try to make these work over the whole `MeshBlockPack` as there is probably some advantage to doing so.

I think that's a topic for a future MR, however.

example/calculate_pi/pi_driver.cpp
Yurlungur: Since this is a discussion for the future. Resolving this conversation.

src/bvals/bvals_interfaces.hpp
```diff
@@ -266,6 +266,7 @@ class BoundaryVariable : public BoundaryCommunication, public BoundaryBuffer {
   bool ReceiveBoundaryBuffers() override;
   void ReceiveAndSetBoundariesWithWait() override;
   void SetBoundaries() override;
+  auto GetPBdVar() { return &bd_var_; }
```
Yurlungur, line 269: :+1:

src/interface/data_collection.cpp
```diff
@@ -0,0 +1,66 @@
...
+}
+
+template <>
+std::shared_ptr<MeshData<Real>> &
+DataCollection<MeshData<Real>>::GetOrAdd(const std::string &mbd_label,
+                                         const int &partition_id) {
+  const std::string label = mbd_label + "_part-" + std::to_string(partition_id);
+  auto it = containers_.find(label);
+  if (it == containers_.end()) {
+    // TODO(someone) add caching of partitions to Mesh at some point
```
Yurlungur, line 51: Do we know if this is a performance bottleneck? Partitioning should be very fast, as it just moves some shared pointers around. Also if packs and `MeshData<Real>>` are already cached, then will the partition caching ever be relevant?

src/interface/data_collection.hpp
```diff
@@ -0,0 +1,80 @@
...
+  std::shared_ptr<T> &Get() { return containers_["base"]; }
+  std::shared_ptr<T> &Get(const std::string &label) {
+    auto it = containers_.find(label);
+    if (it == containers_.end()) {
+      throw std::runtime_error("Container " + label + " does not exist in collection.");
+    }
+    return it->second;
+  }
+
+  std::shared_ptr<T> &GetOrAdd(const std::string &mbd_label, const int &partition_id);
```
Yurlungur, line 60: I feel like `GetOrAdd` could be unified with `Get/Add` a bit better. For example, perhaps there could be a version with signature
```C++
  std::shared_ptr<T> &GetOrAdd(const std::string &mbd_label, const std::string &base);
```
to replace separate `Get` and `Add` calls elsewhere in the code.

Also perhaps the signature for the one with an int partition id should have a default for the case of only 1 partition. e.g.,
```suggestion
  std::shared_ptr<T> &GetOrAdd(const std::string &mbd_label, const int partition_id = 0);
```
This can all be saved for another time (or never) though.

src/interface/mesh_data.hpp
```diff
@@ -0,0 +1,166 @@
...
+  template <typename... Args>
+  auto PackVariablesAndFluxes(Args &&... args) {
+    auto pack_function = [&](std::shared_ptr<MeshBlockData<T>> meshblock_data) {
+      return meshblock_data->PackVariablesAndFluxes(std::forward<Args>(args)...);
+    };
+    vpack_types::StringPair key;
+    auto vpack = block_data_[0]->PackVariablesAndFluxes(std::forward<Args>(args)..., key);
+    return pack_on_mesh_impl::PackOnMesh<VariableFluxPack<T>>(key, varFluxPackMap_,
+                                                              block_data_, pack_function);
+  }
```
Yurlungur, line 134: :+1:

src/interface/meshblock_data.cpp
Yurlungur: since no answer from @jdolence or @AndrewGaspar  lets leave it for now.

Yurlungur (approved):
I'm happy with this. As soon as the long regression tests pass, I'll pull the trigger.

example/advection/advection_driver.cpp
pgrete: Yes, that's another item on the list of things to be addressed in a separate MR (sooner than later) as it'll make reviewing easier.

src/interface/data_collection.cpp
pgrete: As far as I've looked at the profiling data (without looking at this part specifically) this is now likely less of an issue.
I only added the comment as in the first iteration in this PR without the `GetOrAdd` function, the partitioning would happen in every (sub)cycle just to extract the right (cached) container.
With the current automated caching in place this should be negligible compared to other piece that are triggered when the caches are invalid (such as after load balancing).

src/interface/data_collection.hpp
pgrete: The main reason why I didn't go a more "unified" version as the `GetOrAdd` is specific to the `DataCollection<MeshData>` and not implemented for the `DataCollection<MeshBlockData>` (both implementation would be distinct whereas the high level `GetOrAdd` work in their templated form.

Yurlungur:
Tests pass. Pulling the trigger.

PR #331, opened 2020-10-22
WIP: attempt at enabling user data in a generic, flexible way

Yurlungur:
## PR Summary

At some point, we identified a need for registering data at the mesh and meshblock level that users can create and modify. This PR is a first pass at that, although it's still a little half baked. The idea is:
1. Abstract `Params` into a new class `DictAnyType`, which is identical to `Params`, but you can modify values.
2. `Params` now wraps `DictAnyType` but returns const references instead of references.
3. `Mesh` and `MeshBlock` each own a `DictAnyType` object for user data.
4. The user can register `InitApplicationData` and `InitUserMeshData` to meshblock and mesh to set the values in these objects.

I'd be interested in feedback from folks.

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted


jlippuner:
Will revisit if needed

PR #332, opened 2020-10-22
Boundary conditions

Yurlungur:
## PR Summary

Our internal code utilizes non-periodic boundary conditions. However, these currently do not work on GPU. This is an attempt at rewriting them to:
1. Make them run on GPU
2. Generalize them a bit to support user-submitted boundaries
3. Maybe fix inconsistencies if there are any

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


Yurlungur:
PR sill very much in progress. Two pieces now (quasi) implemented in support:
1. Variable type (vector, tensor, etc) and vector component is now available in `VariablePack`. This may need to change if it results in the performance problems we were seeing before with lightweight kokkos kernels.
2. `IndexShape` can now return bounds for ghost zone regions. Interested in your opinion on this @JoshuaSBrown

src/mesh/domain.hpp
```diff
@@ -114,68 +126,117 @@ class IndexShape {
...
-    int total = 1;
-    if (domain == IndexDomain::entire) {
-      for (int i = 0; i < NDIM; ++i)
-        total *= entire_ncells_[i];
-    } else {
-      for (int i = 0; i < NDIM; ++i)
-        total *= x_[i].e - x_[i].s + 1;
-    }
-    return total;
+    return ncellsi(domain) * ncellsj(domain) * ncellsk(domain);
```
JoshuaSBrown, line 259: This is cute. :)

src/mesh/domain.hpp
```diff
@@ -45,8 +45,20 @@ struct IndexRange {
...
+// meshblock
+enum class IndexDomain {
+  entire,
+  interior,
+  inner_x1,
+  outer_x1,
+  inner_x2,
+  outer_x2,
+  inner_x3,
+  outer_x3
```
JoshuaSBrown, line 77: I think, `inner` and `outer` might be a bit misleading, inner - makes me think it is a domain within the interior while outer makes me think it is located on the outer edge. I think bottom and top would work better, or front/back, anterior/posterior, fore/aft, any of these combinations would be a bit more consistent with the domain enum. If you had a separate enum for edge then I can see inner and outer working.

Yurlungur: The reason I went with `inner` and `outer` is that 's the naming convention elsewhere throughout the code. I.e., in the `parthinput` files, that's what those are called. I'm open to changing it, but I would rather that be a code-wide change that happens in a later PR after some discussion. Would that be ok?

Yurlungur:
@JoshuaSBrown if you have a chance, I'd appreciate another look at the domain stuff. I've got all the tests passing for it now.

src/defs.hpp
```diff
@@ -105,7 +105,7 @@ struct RegionSize { // aggregate and POD type; do NOT reorder member declaration
 // X1DIR x, r, etc...
 // X2DIR y, theta, etc...
 // X3DIR z, phi, etc...
-enum CoordinateDirection { X0DIR = 0, X1DIR = 1, X2DIR = 2, X3DIR = 3 };
+enum CoordinateDirection { NODIR = -1, X0DIR = 0, X1DIR = 1, X2DIR = 2, X3DIR = 3 };
```
JoshuaSBrown: Why are we using c style enums instead of an enum class?

tst/unit/test_unit_domain.cpp
```diff
@@ -22,9 +22,15 @@
...
-TEST_CASE("Checking IndexShape indices", "[is,ie,js,je,ks,ke]") {
+TEST_CASE("Checking IndexShape indices", "[IndexShape]") {
   const parthenon::IndexDomain interior = parthenon::IndexDomain::interior;
   const parthenon::IndexDomain entire = parthenon::IndexDomain::entire;
+  const parthenon::IndexDomain inner_x1 = parthenon::IndexDomain::inner_x1;
+  const parthenon::IndexDomain outer_x1 = parthenon::IndexDomain::outer_x1;
+  const parthenon::IndexDomain inner_x2 = parthenon::IndexDomain::inner_x2;
+  const parthenon::IndexDomain outer_x2 = parthenon::IndexDomain::outer_x2;
+  const parthenon::IndexDomain inner_x3 = parthenon::IndexDomain::inner_x3;
+  const parthenon::IndexDomain outer_x3 = parthenon::IndexDomain::outer_x3;
```
JoshuaSBrown: Using `const auto` would be appropriate here, but it doesn't really matter.

src/mesh/domain.hpp
JoshuaSBrown: https://github.com/lanl/parthenon/issues/354

src/mesh/domain.hpp
```diff
@@ -114,68 +126,124 @@ class IndexShape {
...
+      return x_[1].s;
+    case IndexDomain::outer_x2:
+      return entire_ncells_[1] == 1 ? 0 : x_[1].e + 1;
+    default:
+      return 0;
+    }
   }
 
+  // note ks and ke are different than the others for boundaries.
+  // this is because eextending i and j over extra ghost zones
```
JoshuaSBrown: typo

src/mesh/domain.hpp
JoshuaSBrown: It would be nice if there were a figure to help illustrate these different zones.

tst/unit/test_unit_domain.cpp
```diff
@@ -145,9 +295,15 @@ TEST_CASE("Checking IndexShape indices", "[is,ie,js,je,ks,ke]") {
...
-TEST_CASE("Checking IndexShape cell counts", "[ncellsi,ncellsj,ncellsk]") {
+TEST_CASE("Checking IndexShape cell counts", "[IndexShape]") {
   const parthenon::IndexDomain interior = parthenon::IndexDomain::interior;
   const parthenon::IndexDomain entire = parthenon::IndexDomain::entire;
+  const parthenon::IndexDomain inner_x1 = parthenon::IndexDomain::inner_x1;
+  const parthenon::IndexDomain outer_x1 = parthenon::IndexDomain::outer_x1;
+  const parthenon::IndexDomain inner_x2 = parthenon::IndexDomain::inner_x2;
+  const parthenon::IndexDomain outer_x2 = parthenon::IndexDomain::outer_x2;
+  const parthenon::IndexDomain inner_x3 = parthenon::IndexDomain::inner_x3;
+  const parthenon::IndexDomain outer_x3 = parthenon::IndexDomain::outer_x3;
```
JoshuaSBrown: const auto

JoshuaSBrown (commented):
So I think this is a great implementation, and the testing is very thorough which I appreciate, I just have a problem with the semantics:
1. as mentioned previously I don't think interior and exterior are appropriate, but I have created an issue for that.  
2. I do not particularly like how the ghost zones are being handled in terms of the lack of symmetry between the IndexDomain and the portion of the ghost zone boundary that it represents. I think that the names should reflect that somehow e.g. how the k indices are treated differently. This is confusing to me that semantically the IndexDomain is the basically the same but I will get different behavior. 

E.g. I would change the names

inner_x1_trunc_x3
outer_x1_trunc_x3
inner_x2_trunc_x3
outer_x2_trunc_x3
inner_x3
outer_x3

Or something of the sort.

src/mesh/domain.hpp
Yurlungur: Yeah I had one in there at some point, but I think it was difficult to parse. I'll try my hand at it again.

Yurlungur:
> I do not particularly like how the ghost zones are being handled in terms of the lack of symmetry between the IndexDomain and the portion of the ghost zone boundary that it represents. I think that the names should reflect that somehow e.g. how the k indices are treated differently. This is confusing to me that semantically the IndexDomain is the basically the same but I will get different behavior.
> 
> E.g. I would change the names
> 
> inner_x1_trunc_x3
> outer_x1_trunc_x3
> inner_x2_trunc_x3
> outer_x2_trunc_x3
> inner_x3
> outer_x3
> 
> Or something of the sort.

Yeah this is a good point @JoshuaSBrown . Let me push back a little though. The reason I chose these names is because these are the *precise* choices required to exactly fill in the boundary. I.e., if you did one 3D loop each over `inner_x1`, `inner_x2`, `inner_x3`, `outer_x1`, `outer_x2`, `outer_x3`, you will have looped over the entire ghost region exactly once. I'm using that feature in my boundary conditions code. And I want to expose it to application developers.

src/defs.hpp
Yurlungur: `X1DIR` and friends are used for indexing into arrays in various places. Making them a class, and thus type-safe, would break that behaviour and/or make it much more verbose.

tst/unit/test_unit_domain.cpp
Yurlungur: done.

src/mesh/domain.hpp
Yurlungur: fixed

tst/unit/test_unit_domain.cpp
Yurlungur: fixed.

Yurlungur:
Also thanks for taking a look, @JoshuaSBrown !

JoshuaSBrown:
> Yeah this is a good point @JoshuaSBrown . Let me push back a little though. The reason I chose these names is because these are the _precise_ choices required to exactly fill in the boundary. I.e., if you did one 3D loop each over `inner_x1`, `inner_x2`, `inner_x3`, `outer_x1`, `outer_x2`, `outer_x3`, you will have looped over the entire ghost region exactly once. I'm using that feature in my boundary conditions code. And I want to expose it to application developers.

I still don't understand why it would be any more difficult to have different names for the third dimension, can you post an example of your loops so I can better understand?

Yurlungur:
> > Yeah this is a good point @JoshuaSBrown . Let me push back a little though. The reason I chose these names is because these are the _precise_ choices required to exactly fill in the boundary. I.e., if you did one 3D loop each over `inner_x1`, `inner_x2`, `inner_x3`, `outer_x1`, `outer_x2`, `outer_x3`, you will have looped over the entire ghost region exactly once. I'm using that feature in my boundary conditions code. And I want to expose it to application developers.
> 
> I still don't understand why it would be any more difficult to have different names for the third dimension, can you post an example of your loops so I can better understand?

Yeah, so here's a call that uses `inner_x1`:

```C++
void ReflectInnerX1(std::shared_ptr<MeshBlockData<Real>> &rc, bool coarse) {
  int ref = pmb->cellbounds.GetBoundsI(IndexDomain::interior).s;
  auto q = rc->PackVariables(std::vector<MetadataFlag>{Metadata::Independent}, coarse);
  pmb->par_for_4D(
      "ReflectInnerX1", 0, q.GetDim(4), IndexDomain::inner_x1,
      KOKKOS_LAMBDA(const int &l, const int &k, const int &j, const int &i) {
        Real reflect = q.VectorComponent(l) == X1DIR ? -1.0 : 1.0;
        q(l, k, j, i) = reflect * q(l, k, j, 2 * ref - i - 1);
      });
}
```
And here's one that uses `inner_x3`:
```C++
void ReflectInnerX3(std::shared_ptr<MeshBlockData<Real>> &rc, bool coarse) {
  int ref = pmb->cellbounds.GetBoundsK(IndexDomain::interior).s;
  auto q = rc->PackVariables(std::vector<MetadataFlag>{Metadata::Independent}, coarse);
  pmb->par_for_4D(
      "ReflectInnerX3", 0, q.GetDim(4), IndexDomain::inner_x3,
      KOKKOS_LAMBDA(const int &l, const int &k, const int &j, const int &i) {
        Real reflect = q.VectorComponent(l) == X3DIR ? -1.0 : 1.0;
        q(l, k, j, i) = reflect * q(l, 2 * ref - k - 1, j, i);
      });
}
```
where note that I've introduced `Par_for_4D`, which takes bounds for the 4th dimension and `IndexDomain` object for the inner 3. Notice how the syntax is almost identical. That's the advantage.

JoshuaSBrown:
I'm still not a fan, for the reasons I gave but this is probably more of a preference thing, thanks for clarifying.

Yurlungur:
Not all the work needed is finished. Tests don't pass yet, for example. But I think this is now in a place where I can begin soliciting feedback. @jdolence @pgrete @nmsriram @jmstone  I'd be especially interested in your feedback. Thanks @JoshuaSBrown for the feedback you've already given. The overall plan as implemented is:

1. Re-enable the function pointer machinery from Athena++ for boundary conditions. Each boundary face has a function pointer for the boundary condition, which is applied.
2. I re-write the boundary interface map a little better to the `VariablePack` and `MeshBlockPack` machinery. The boundaries are applied over all variables at once and over the entire face of a meshblock, rather than per block-neighbor interface.
3. To expose this to the user, I implement `IndexDomain`s for each set of ghost cells in a meshblock. I also implement a `par_for_bndry` method which loops over the coarse or fine bounds of a given ghost cell region.
4. The user thus can register a boundary function for each face by setting it in the `ApplicationInput` struct and then choosing `user` as the boundary condition for that face in the input file. Unfortunately, the user is responsible for making sure the function calls the right loop bounds.
5. Enabling this machinery at positions in the mesh where there's a coarse-fine mesh boundary means the new boundary functions need to be callable on either coarse buffers or fine buffers. To enable this, I set up `VariablePack` to optionally pack the coarse buffer in a variable, rather than the fine buffer.
6. I also needed to enable the boundary functions to be able to know about which component of a variable pack is a vector. I enable this in the variable pack.

src/bvals/boundary_conditions.cpp
```diff
@@ -12,9 +12,9 @@
 //========================================================================================
```
Yurlungur, line 12: This file has essentially been completely re-written, so the diff may not be as informative as just looking at the file.

Yurlungur:
I believe this is now ready for review. A few things to note:
- I just commented out the `applyBounds` stuff in `mesh_refinement.cpp`. @nmsriram and I discussed. There's still some confusion as to what caused the restart issue and it's not fully resolved in our downstream code.
- I would like to add an `outflow` regression test using the advection example. However, I'm not sure how to do it. I'd like to engage with either @JoshuaSBrown or @pgrete to do this. And this can happen in a separate MR or in this one.

src/mesh/domain.hpp
Yurlungur: Figure added.

example/advection/advection_driver.cpp
```diff
@@ -141,16 +141,16 @@ TaskCollection AdvectionDriver::MakeTaskCollection(BlockList_t &blocks, const in
...
+    if (pmesh->multilevel) {
+      prolongBound = tl.AddTask(prev_task, parthenon::ProlongateBoundaries, sc1);
+    }
 
     // set physical boundaries
-    auto set_bc = tl.AddTask(prolongBound, parthenon::ApplyBoundaryConditions, sc1);
+    // note "false" in call to ApplyBoundaryConditions.
+    // AddTask is not smart enough to resolve function overloads or default arguments.
+    auto set_bc =
+        tl.AddTask(prolongBound, parthenon::ApplyBoundaryConditions, sc1, false);
```
jdolence: It's entirely possible I don't understand this all well enough yet, but this false feels a bit like some details of the implementation leaking through to the interface.  Is it always false when called from the driver like this?

src/bvals/bvals.hpp
```diff
@@ -125,7 +125,8 @@ class BoundaryValues : public BoundaryBase, // public BoundaryPhysics,
   // non-inhertied / unique functions (do not exist in BoundaryVariable objects):
   // (these typically involve a coupled interaction of boundary variable/quantities)
   // ------
-  void ProlongateBoundaries(const Real time, const Real dt);
```
jdolence, line 128: I know we weren't really supporting them, but does this change mean we've "lost" the ability to do time dependent boundaries?

src/interface/meshblock_data.cpp
```diff
@@ -815,6 +816,50 @@ TaskStatus MeshBlockData<T>::ClearBoundary(BoundaryCommSubset phase) {
   return TaskStatus::complete;
 }
 
+template <typename T>
+void MeshBlockData<T>::RestrictBoundaries() {
+  auto pmb = GetBlockPointer();
+  pmb->pbval->RestrictBoundaries();
+  // TODO(JMM): once we have variable-level pbvals we can do this
+  /*
```
jdolence: is this dead code or did you intend to keep it around?

src/interface/meshblock_data.cpp
```diff
@@ -815,6 +816,50 @@ TaskStatus MeshBlockData<T>::ClearBoundary(BoundaryCommSubset phase) {
...
+  }
+  */
+}
+
+template <typename T>
+void MeshBlockData<T>::ProlongateBoundaries() {
+  auto pmb = GetBlockPointer();
+  pmb->pbval->ProlongateBoundaries();
+  // TODO(JMM): once we have variable-level pbvals we can do this
+  /*
```
jdolence: same as the last one.

src/interface/meshblock_data.hpp
```diff
@@ -291,33 +291,40 @@ class MeshBlockData {
...
+                                bool coarse = false);
   VariablePack<T> PackVariables(const std::vector<MetadataFlag> &flags,
-                                PackIndexMap &vmap);
-  VariablePack<T> PackVariables(const std::vector<MetadataFlag> &flags);
-  VariablePack<T> PackVariables(PackIndexMap &vmap);
+                                PackIndexMap &vmap, bool coarse = false);
+  VariablePack<T> PackVariables(const std::vector<MetadataFlag> &flags,
+                                bool coarse = false);
+  VariablePack<T> PackVariables(PackIndexMap &vmap, bool coarse = false);
+  // no coarse flag because you need to be able to specify non one-copy variables
```
jdolence, line 312: you could probably solve this with some gross template metaprogramming, but let's not...

src/interface/variable.cpp
```diff
@@ -113,6 +113,12 @@ void CellVariable<T>::allocateComms(std::weak_ptr<MeshBlock> wpmb) {
 
   // enroll CellCenteredBoundaryVariable object
   vbvar->bvar_index = pmb->pbval->bvars.size();
+  // TODO(JMM): This means RestrictBoundaries()
```
jdolence, line 116: Is the implication that we'll currently be launching  something like 2x more kernels than necessary for dealing with boundaries?

src/mesh/mesh.cpp
```diff
@@ -75,12 +75,12 @@ Mesh::Mesh(ParameterInput *pin, ApplicationInput *app_in, Properties_t &properti
                 pin->GetInteger("parthenon/mesh", "nx2"),
                 pin->GetInteger("parthenon/mesh", "nx3")},
       mesh_bcs{
-          GetBoundaryFlag(pin->GetOrAddString("parthenon/mesh", "ix1_bc", "reflecting")),
```
jdolence: I'd rather not change these defaults to periodic.

src/mesh/mesh_refinement.cpp
```diff
@@ -41,6 +41,7 @@ namespace parthenon {
 
 // TODO(Yurlunger) Identify an idiomatic way of applying reflective
 // boundary conditions in the boundary condition refactor code
+/*
```
jdolence: We should figure this out and either take this out or enable it if necessary.

src/mesh/meshblock.hpp
```diff
@@ -293,6 +315,16 @@ class MeshBlock : public std::enable_shared_from_this<MeshBlock> {
                             std::forward<Args>(args)...);
   }
 
+  template <typename Function, class... Args>
+  inline typename std::enable_if<sizeof...(Args) <= 1, void>::type
+  par_dispatch_(const std::string &name, const IndexRange &kb, const IndexRange &jb,
```
jdolence, line 314: 👍

jdolence (commented):
Looks good.  I'd like to better understand the `coarse` bool that seems to need to be passed around everywhere.

example/advection/advection_driver.cpp
Yurlungur: Yes. This is a leaky implementation detail coming from the fact that `AddTask` isn't smart enough to handle overloads or default arguments. I could hide it by renaming the function with the bool option.

src/bvals/bvals.hpp
Yurlungur: At the moment, yes. I can re-enable it by passing time through the boundary machinery easily enough. Should we do that now?

src/interface/meshblock_data.cpp
Yurlungur: This code never worked. Its essentially an aspirational comment for how I'd like things to look if we refactor the `BoundaryValues` class.

src/interface/meshblock_data.hpp
Yurlungur: @nmsriram and I actually looked at this for a while. There are other solutions, but this is by far the simplest one.

src/interface/variable.cpp
Yurlungur: we're launching `nstages x` more kernels than necessary for dealing with boundaries. I didn't fix that here, because it may be larger surgery. But it should be fixed. I can create an issue.

src/mesh/mesh_refinement.cpp
Yurlungur: I vote for taking it out for now. Whether or not it's necessary requires understanding the restart issue. It does not appear necessary at runtime. Which implies to me that it probably doesn't belong here.

Yurlungur:
> Looks good. I'd like to better understand the `coarse` bool that seems to need to be passed around everywhere.

Thanks, @jdolence ! Yeah, this `coarse` bool is used for applying boundaries on the coarse buffer present on meshblocks for multilevel runs. It's needed because if you have a coarse/fine mesh boundary next to a physical boundary, then the ghost zones used to prolongate onto the coarse meshblock must be filled.

To enable this, I threaded this boolean through the `meshblockpack` so you can pack together all coarse buffers for all variables on a block.

example/advection/advection_driver.cpp
jdolence: I think you could hide it in a lambda as well, something like:
```c++
auto set_bc = tl.AddTask(prolongBound, [](std::shared_ptr<MeshBlockData<Real>> &sc) {
           return parthenon::ApplyBoundaryConditions(sc);
}, sc1);
```
Or just use a `static_cast` to select the right overload.

Yurlungur: resolved by introducing
```C++
TaskStatus ApplyBoundaryConditionsOnCoarseOrFine(std::shared_ptr<MeshBlockData<Real>> &rc,
                                                 bool coarse);

inline TaskStatus ApplyBoundaryConditions(std::shared_ptr<MeshBlockData<Real>> &rc) {
  return ApplyBoundaryConditionsOnCoarseOrFine(rc,false);
}
```

src/bvals/bvals.hpp
Yurlungur: #366

src/interface/variable.cpp
Yurlungur: #366

src/bvals/bvals.hpp
Yurlungur: #367

src/interface/meshblock_data.cpp
Yurlungur: I'm going to delete this code as it never worked. But I will leave the TODO in.

src/interface/meshblock_data.cpp
Yurlungur: deleting this code but leaving a `TODO`

src/mesh/mesh.cpp
Yurlungur: I'll change it back. But I would be curious as to why.

Yurlungur: changed.

src/mesh/mesh_refinement.cpp
Yurlungur: Removed. Tests in our downstream code appear to show the downstream restart problem is no longer present.

Yurlungur:
@jdolence I think I've addressed all your comments. Please take another look.

@JoshuaSBrown I think I've also addressed all your comments. Would you mind taking another look?

I have also added the regression test and published a new test suite as a release. @pgrete @forrestglines can someone take a look from the AthenaPK team?

Yurlungur:
I'm going to run the full test suite as well as a check.

Yurlungur:
Full test suite caught an issue with `VariablePack`, where the packs would segfault if the required pack should be empty. Now resolved. Test suite now passes.

docs/mesh/domain.md
```diff
@@ -4,10 +4,23 @@
 
 The index shape class provides access to the index bounds of a mesh block. The
 bounds are split into two domains which can be accessed using a class enum
```
JoshuaSBrown: There are more than two domains now, **several**.

src/bvals/boundary_conditions.cpp
```diff
@@ -23,248 +23,237 @@
...
+  // Step 1. Apply necessary variable restrictions when ghost-ghost zone is on same lvl
+  rc->RestrictBoundaries(); // Step 1: restrict physical boundaries
+
+  // Step 2. Re-apply physical boundaries on the coarse boundary,
+  ApplyBoundaryConditionsOnCoarseOrFine(rc, true);
+
+  // Step 3. Finally, the ghost-ghost zones are ready for prolongation:
+  rc->ProlongateBoundaries();
+
+  return TaskStatus::complete;
```
JoshuaSBrown, line 54: This reads well. 👍🏻

src/bvals/boundary_conditions.cpp
```diff
@@ -23,248 +23,237 @@
...
-      break;
-
-    default:
-      break;
+TaskStatus ApplyBoundaryConditionsOnCoarseOrFine(std::shared_ptr<MeshBlockData<Real>> &rc,
+                                                 bool coarse) {
+  using namespace boundary_cond_impl;
+  std::shared_ptr<MeshBlock> pmb = rc->GetBlockPointer();
+  Mesh *pmesh = pmb->pmy_mesh;
+  int ndim = pmesh->ndim;
```
JoshuaSBrown: make const.

src/bvals/boundary_conditions.cpp
```diff
@@ -23,248 +23,237 @@
...
+  int ref = bounds.GetBoundsK(IndexDomain::interior).e;
+  auto q = rc->PackVariables(std::vector<MetadataFlag>{Metadata::FillGhost}, coarse);
+  auto nb = IndexRange{0, q.GetDim(4) - 1};
+  pmb->par_for_bndry(
+      "ReflectOuterX3", nb, IndexDomain::outer_x3, coarse,
+      KOKKOS_LAMBDA(const int &l, const int &k, const int &j, const int &i) {
+        Real reflect = q.VectorComponent(l) == X3DIR ? -1.0 : 1.0;
+        q(l, k, j, i) = reflect * q(l, 2 * ref - k + 1, j, i);
+      });
+}
```
JoshuaSBrown, line 239: There is a lot of repetitive logic here, is there anyway some of it could be packaged into a local subfunction?

src/bvals/boundary_conditions.cpp
```diff
@@ -23,248 +23,237 @@
...
+
+} // namespace BoundaryFunction
+
+namespace boundary_cond_impl {
+bool DoPhysicalBoundary_(BoundaryFlag flag, BoundaryFace face, int ndim) {
```
JoshuaSBrown: These parameters could be made const.

src/bvals/boundary_conditions.hpp
```diff
@@ -14,15 +14,45 @@
...
+void OutflowInnerX2(std::shared_ptr<MeshBlockData<Real>> &rc, bool coarse);
+void OutflowOuterX2(std::shared_ptr<MeshBlockData<Real>> &rc, bool coarse);
+void OutflowInnerX3(std::shared_ptr<MeshBlockData<Real>> &rc, bool coarse);
+void OutflowOuterX3(std::shared_ptr<MeshBlockData<Real>> &rc, bool coarse);
+void ReflectInnerX1(std::shared_ptr<MeshBlockData<Real>> &rc, bool coarse);
+void ReflectOuterX1(std::shared_ptr<MeshBlockData<Real>> &rc, bool coarse);
+void ReflectInnerX2(std::shared_ptr<MeshBlockData<Real>> &rc, bool coarse);
+void ReflectOuterX2(std::shared_ptr<MeshBlockData<Real>> &rc, bool coarse);
+void ReflectInnerX3(std::shared_ptr<MeshBlockData<Real>> &rc, bool coarse);
+void ReflectOuterX3(std::shared_ptr<MeshBlockData<Real>> &rc, bool coarse);
```
JoshuaSBrown, line 53: Can you make coarse const?

src/bvals/bvals_refine.cpp
```diff
@@ -399,4 +305,88 @@ void BoundaryValues::ProlongateGhostCells(const NeighborBlock &nb, int si, int e
...
+    nje = std::min(nb.ni.ox2 + 1, 1);
+  }
+
+  if (pmb->block_size.nx3 == 1) {
+    nks = 0;
+    nke = 0;
+  } else {
+    nks = std::max(nb.ni.ox3 - 1, -1);
+    nke = std::min(nb.ni.ox3 + 1, 1);
+  }
```
JoshuaSBrown: Repetitive logic here, I would place in local sub function.

src/bvals/bvals_refine.cpp
```diff
@@ -399,4 +305,88 @@ void BoundaryValues::ProlongateGhostCells(const NeighborBlock &nb, int si, int e
...
+        sk -= cn;
+      }
+    }
+  } else if (nb.ni.ox3 > 0) {
+    sk = pmb->c_cellbounds.ke(interior) + 1;
+    ek = pmb->c_cellbounds.ke(interior) + cn;
+  } else {
+    sk = pmb->c_cellbounds.ks(interior) - cn;
+    ek = pmb->c_cellbounds.ks(interior) - 1;
+  }
```
JoshuaSBrown: Repetitive logic.

src/interface/meshblock_data.cpp
```diff
@@ -340,82 +340,84 @@ template <typename T>
...
+                                std::vector<std::string> &key, bool coarse) {
   vpack_types::VarList<T> vars = MakeList_(names, key, sparse_ids);
-  return PackVariablesHelper_(key, vars, vmap);
+  return PackVariablesHelper_(key, vars, vmap, coarse);
 }
 template <typename T>
 VariablePack<T> MeshBlockData<T>::PackVariables(const std::vector<std::string> &names,
                                                 const std::vector<int> &sparse_ids,
-                                                PackIndexMap &vmap) {
+                                                PackIndexMap &vmap, bool coarse) {
```
JoshuaSBrown: Can coarse be made const?

src/interface/variable_pack.hpp
```diff
@@ -70,9 +71,9 @@ class VariablePack {
  public:
   VariablePack() = default;
   VariablePack(const ViewOfParArrays<T> view, const ParArray1D<int> sparse_ids,
-               const std::array<int, 4> dims)
-      : v_(view), sparse_ids_(sparse_ids), dims_(dims),
-        ndim_((dims[2] > 1 ? 3 : (dims[1] > 1 ? 2 : 1))) {}
+               const ParArray1D<int> vector_component, const std::array<int, 4> dims)
```
JoshuaSBrown: I'm pretty sure all these arguments can be made references so you aren't copying them twice.

src/interface/variable_pack.hpp
```diff
@@ -105,8 +117,9 @@ class VariableFluxPack : public VariablePack<T> {
   VariableFluxPack() = default;
   VariableFluxPack(const ViewOfParArrays<T> view, const ViewOfParArrays<T> f0,
                    const ViewOfParArrays<T> f1, const ViewOfParArrays<T> f2,
-                   const ParArray1D<int> sparse_ids, const std::array<int, 4> dims)
-      : VariablePack<T>(view, sparse_ids, dims), f_({f0, f1, f2}) {}
+                   const ParArray1D<int> sparse_ids,
+                   const ParArray1D<int> vector_component, const std::array<int, 4> dims)
```
JoshuaSBrown: The same I think is true here, it looks like the arguments are being copied when they are passed in and also when they are initialized.

src/interface/variable_pack.hpp
```diff
@@ -184,7 +199,14 @@ void FillVarView(const vpack_types::VarList<T> &vars, PackIndexMap *vmap,
       for (int j = 0; j < v->GetDim(5); j++) {
         for (int i = 0; i < v->GetDim(4); i++) {
           host_sp(vindex) = sparse_id;
-          host_view(vindex++) = v->data.Get(k, j, i);
+          // TODO(JMM): Is this safe for vector-valued sparse variables?
+          // returns 1 for X1DIR, 2 for X2DIR, 3 for X3DIR
+          // for tensors, returns flattened index.
+          // for scalar-objects, returns NODIR.
+          bool is_vec = v->IsSet(Metadata::Vector) || v->IsSet(Metadata::Tensor);
```
JoshuaSBrown: const

src/mesh/meshblock.hpp
```diff
@@ -182,14 +182,15 @@ class MeshBlock : public std::enable_shared_from_this<MeshBlock> {
...
-                      const Function &function) {
-    parthenon::par_for(DEFAULT_LOOP_PATTERN, name, exec_space, bl, bu, nl, nu, kl, ku, jl,
-                       ju, il, iu, function);
+  inline void par_for_bndry(const std::string &name, const IndexRange &nb,
+                            const IndexDomain &domain, const bool coarse,
+                            const Function &function) {
+    auto bounds = coarse ? c_cellbounds : cellbounds;
+    auto ib = bounds.GetBoundsI(domain);
+    auto jb = bounds.GetBoundsJ(domain);
+    auto kb = bounds.GetBoundsK(domain);
```
JoshuaSBrown, line 192: const reference?

src/bvals/bvals.hpp
```diff
@@ -151,6 +152,10 @@ class BoundaryValues : public BoundaryBase, // public BoundaryPhysics,
                                             int sk, int ek);
   void ProlongateGhostCells(const NeighborBlock &nb, int si, int ei, int sj, int ej,
                             int sk, int ek);
+  void ComputeRestrictionBounds_(const NeighborBlock &nb, int &nis, int &nie, int &njs,
+                                 int &nje, int &nks, int &nke);
+  void ComputeProlongationBounds_(const NeighborBlock &nb, int &si, int &ei, int &sj,
+                                  int &ej, int &sk, int &ek);
```
JoshuaSBrown: I would recommend using IndexRange here instead of specifying every index bound individually, it will help prevent future users from putting them in, in the wrong order.

src/bvals/boundary_conditions.cpp
Yurlungur: Yeah I don't love this the way it currently is. But no easy solution came to mind. The problem is that the KOKKOS_LAMBDA is slightly different each time in a way that makes it difficult to generalize more.

Also there may be some utility in leaving these as-is as reference implementations for downstream codes.

Suggestions welcome.

JoshuaSBrown: Oh, I see what you mean, ya, I should have paid more attention.

src/mesh/domain.hpp
```diff
@@ -114,68 +143,124 @@ class IndexShape {
...
 
   KOKKOS_INLINE_FUNCTION int ncellsk(const IndexDomain &domain) const noexcept {
-    return (domain == IndexDomain::entire) ? entire_ncells_[2] : x_[2].e - x_[2].s + 1;
+    if (entire_ncells_[2] == 1 &&
+        (domain == IndexDomain::inner_x3 || domain == IndexDomain::outer_x3)) {
+      return 0; // if x3 is zero-dimensional, there are no ghost zones
+    }
+    auto bnds = GetBoundsK(domain);
+    return bnds.e - bnds.s + 1;
   }
```
JoshuaSBrown: Can you make bnds const auto &? I think this might be the same problem with returning a temporary you pointed out above, but there should be a way to return the length of the domain directly instead of having to call bnds.e -bnds.s.

src/mesh/mesh.cpp
```diff
@@ -909,18 +912,39 @@ void Mesh::OutputMeshStructure(int ndim) {
...
+      } else {
+        std::stringstream msg;
+        msg << "A user boundary condition for face " << f
+            << " was requested. but no condition was enrolled." << std::endl;
+        PARTHENON_THROW(msg);
+      }
+      break;
+    default: // periodic/block BCs handled elsewhere.
+      break;
+    }
```
JoshuaSBrown, line 945: It would be nice if they were all handled in the same place.

src/bvals/boundary_conditions.cpp
```diff
@@ -23,248 +23,237 @@
...
+  using namespace boundary_cond_impl;
+  std::shared_ptr<MeshBlock> pmb = rc->GetBlockPointer();
+  Mesh *pmesh = pmb->pmy_mesh;
+  int ndim = pmesh->ndim;
+
+  for (int i = 0; i < BOUNDARY_NFACES; i++) {
+    if (DoPhysicalBoundary_(pmb->boundary_flag[i], static_cast<BoundaryFace>(i), ndim)) {
+      PARTHENON_DEBUG_REQUIRE(pmesh->MeshBndryFnctn[i] != nullptr,
+                              "boundary function must not be null");
+      pmesh->MeshBndryFnctn[i](rc, coarse);
```
JoshuaSBrown: So what happens if all the boundary conditions are undef? Is there some check for that, is that an error?

src/mesh/mesh.cpp
```diff
@@ -1066,24 +1081,9 @@ void Mesh::Initialize(int res_flag, ParameterInput *pin, ApplicationInput *app_i
...
-        //          if (pbval->nblevel[1][2][1] != -1) ju += NGHOST;
-        //        }
-        //        if (pmb->block_size.nx3 > 1) {
-        //          if (pbval->nblevel[0][1][1] != -1) kl -= NGHOST;
-        //          if (pbval->nblevel[2][1][1] != -1) ku += NGHOST;
-        //        }
-
+        if (multilevel) {
+          ProlongateBoundaries(pmb->meshblock_data.Get());
+        }
```
JoshuaSBrown, line 1085: I love that you put curly braces, safer I think!

JoshuaSBrown (approved):
I just have a few small comments, however nothing that I think should block merging. Thanks @Yurlungur this is great. There might be some other things that may be relevant that Codacy can find related to this pr here, though not everything will be related or worth addressing. https://app.codacy.com/gh/lanl/parthenon/issues?categoryType=Performance&bid=20713268

docs/mesh/domain.md
Yurlungur: good catch thanks.

src/bvals/boundary_conditions.hpp
Yurlungur: In principal, yes. Since these functions are often expected to be supplied by downstream codes via `ApplicationInput`, though... I think I'd rather let `coarse` be non-const though, just to simplify the function signature. I'm willing to change it if you feel strongly.

src/mesh/meshblock.hpp
Yurlungur: Wait will that actually work? These functions return a new, initialized object. I think the move operator is used.
```C++
  KOKKOS_INLINE_FUNCTION const IndexRange GetBoundsI(const IndexDomain &domain) const
      noexcept {
    return (domain == IndexDomain::interior) ? x_[0] : IndexRange{is(domain), ie(domain)};
  }

  KOKKOS_INLINE_FUNCTION const IndexRange GetBoundsJ(const IndexDomain &domain) const
      noexcept {
    return (domain == IndexDomain::interior) ? x_[1] : IndexRange{js(domain), je(domain)};
  }

  KOKKOS_INLINE_FUNCTION const IndexRange GetBoundsK(const IndexDomain &domain) const
      noexcept {
    return (domain == IndexDomain::interior) ? x_[2] : IndexRange{ks(domain), ke(domain)};
  }
```

src/mesh/domain.hpp
Yurlungur: See my above comment. `GetBounds*` returns a whole new `IndexRange` object. I can make it `const` but not `const auto &`. We could maybe pre-compute and store the `Bounds` objects... but that seems like it would require some surgery.

src/mesh/meshblock.hpp
JoshuaSBrown: Ya your right, that's my bad.

src/mesh/mesh.cpp
Yurlungur: Unfortunately, they can't be because periodic/block boundaries utilize the inter-block communication routines. In some sense, periodic boundaries are not physical boundaries at all. They're MPI topology.

src/mesh/domain.hpp
JoshuaSBrown: If not a method should probably be added that does this.

JoshuaSBrown: Ah, ya you should just use the `ncells` methods.

src/bvals/bvals.hpp
Yurlungur: This is a good idea.

src/mesh/mesh.cpp
Yurlungur: I have been burned so many times by leaving the braces out, trying to add code to the loop later, and getting unexpected behaviour because I forgot to add braces to the now multi-line loop.

src/bvals/boundary_conditions.cpp
Yurlungur: Do you mean `boundary_flag[i]==BoundaryFlag::undef`? It's checked in `DoPhysicalBoundary` as:
```C++
bool DoPhysicalBoundary_(const BoundaryFlag flag, const BoundaryFace face, const int ndim) {
  if (flag == BoundaryFlag::block) return false;
  if (flag == BoundaryFlag::undef) return false;
  if (flag == BoundaryFlag::periodic) return false;

  if (ndim < 3 && (face == BoundaryFace::inner_x3 || face == BoundaryFace::outer_x3)) {
    return false;
  }
  if (ndim < 2 && (face == BoundaryFace::inner_x2 || face == BoundaryFace::outer_x2)) {
    return false;
  } // ndim always at least 1

  return true; // reflect, outflow, user, dims correct
}
```
so the boundary condition is simply not applied. I could also stick a debug require there. Not sure what's preferable.

JoshuaSBrown: I just wanted to be sure that it wasn't going to be a problem if none of the boundaries were defined.

Yurlungur:
> There might be some other things that may be relevant that Codacy can find related to this pr here, though not everything will be related or worth addressing. https://app.codacy.com/gh/lanl/parthenon/issues?categoryType=Performance&bid=20713268

 Looks like most isn't related. I'll address the things in files I touched though.

src/bvals/boundary_conditions.hpp
JoshuaSBrown: It's just a suggestion, best practice says make it a const if you can.

src/bvals/boundary_conditions.cpp
Yurlungur: The code should run and be "technically" correct. However, I don't know under circumstances it would be desirable.

src/bvals/boundary_conditions.hpp
Yurlungur: Cool. I'll make the others constant but pass on this one.

src/mesh/domain.hpp
Yurlungur: Oh I see. Yes, that's better.

Yurlungur:
Thanks for all the suggestions, @JoshuaSBrown I think they're all addressed now.

docs/mesh/domain.md
```diff
@@ -3,11 +3,24 @@
...
+full meshblock including ghost zones
+- **interior,** which allows access to the interior of the meshblock
+excluding ghost cells
+- **inner_x1,** which accesses the ghost zones for "inner" face of
+  a meshblock in the `x1` direction
+- **outer_x1,** which accesses the ghost zones for "outer" face of
+  a meshblock in the `x1` direction
+- **inner_x2,** which accesses the ghost zones for "inner" face of
+  a meshblock in the `x2` direction
+- **outer_x2,** which accesses the ghost zones for "outer"` face of
```
forrestglines: Are the back quotes (starting on this line) intentional?

src/bvals/bvals_interfaces.hpp
```diff
@@ -82,21 +82,12 @@ enum {
...
-  outer_x2 = 3,
-  inner_x3 = 4,
-  outer_x3 = 5
-};
 // TODO(felker): BoundaryFace must be unscoped enum, for now. Its enumerators are used as
 // int to index raw arrays (not ParArrayNDs)--> enumerator vals are explicitly specified
 
 // identifiers for boundary conditions
-enum class BoundaryFlag { block = -1, undef, reflect, outflow, periodic };
+// TODO(JMM): Can BoundaryFlag be removed?
```
forrestglines: Why can `BoundaryFlag` be removed? Is there another enum to replace it?

src/bvals/bvals_refine.cpp
```diff
@@ -76,47 +62,26 @@ namespace parthenon {
...
-    }
+    // fill the required ghost-ghost zone
+    IndexRange bni, bnj, bnk;
+    ComputeRestrictionBounds_(nb, bni, bnj, bnk);
 
-    // Step 1. Apply necessary variable restrictions when ghost-ghost zone is on same lvl
-    for (int nk = nks; nk <= nke; nk++) {
-      for (int nj = njs; nj <= nje; nj++) {
-        for (int ni = nis; ni <= nie; ni++) {
+    // TODO(JMM): this loop should probably be a kokkos loop
```
forrestglines, line 81: I'm not sure - there looks to be kernels being called within this loop, although those might be made into inner loops when this is Kokkosized. This code might be difficult to optimize for Kokkos.

src/interface/variable_pack.hpp
```diff
@@ -86,15 +87,26 @@ class VariablePack {
...
+    assert(0 <= n && n < dims_[3]);
+    return vector_component_(n);
+  }
+  KOKKOS_FORCEINLINE_FUNCTION
   int GetNdim() const { return ndim_; }
 
  protected:
   ViewOfParArrays<T> v_;
   ParArray1D<int> sparse_ids_;
+  ParArray1D<int> vector_component_;
```
forrestglines, line 109: Adding this member is going to take up data in every kernel with a VariablePack. This might interfere with future optimizations. Is there a way around adding array?

jdolence (approved):

forrestglines (approved):
Looks good. I see that it will probably be a large task to kokkosize efficiently some of these loops for AMR prolongation and restriction - but that's for another PR.

I have some concern about adding another data member (`velocity_component_`) to the VariablePack. Too many members like that and we might find it hard to get decent occupancy/performance, like the optimization issues that Jim was seeing.


Also, there seemed to be a typo in the documentation. I made a comment pointing it out.

Yurlungur:
Thanks, @forrestglines ! 

> I have some concern about adding another data member (velocity_component_) to the VariablePack. Too many members like that and we might find it hard to get decent occupancy/performance, like the optimization issues that Jim was seeing.

Yes, I share this concern. However, I didn't see a way around that here... this piece of information is needed to handle the boundary conditions correctly in a generic way. Hopefully one `ParArray1D` only is ok.

src/bvals/bvals_interfaces.hpp
Yurlungur: I ended up moving it to `defs.hpp` to avoid a circular dependency in the headers. That said, that `TODO` is clearly wrong and I will remove it.

docs/mesh/domain.md
Yurlungur: Nope. Good catch. Thanks.

src/bvals/bvals_refine.cpp
Yurlungur: Yes. Agreed. This is going to be nasty to unify. I'm waiting for @pgrete 's pack in one MR for guidance to fix that.

src/interface/variable_pack.hpp
Yurlungur: Yes, I agree this is an issue. I couldn't think of a better solution, though. And I'm open to suggestions. I've created an issue #375 to track this.

src/bvals/bvals_interfaces.hpp
Yurlungur: Oh wait. I was thinking of `BoundaryFace`. This TODO was more of... since there's the function pointer machinery, maybe the flags are extraneous. That said, I don't think they are. I will remove the TODO.

src/bvals/bvals_refine.cpp
Yurlungur: I'll add an issue for this too. #376

Yurlungur:
Pulling the trigger.

PR #333, opened 2020-10-23
WIP: Add regression test for particles

brryan:
## PR Summary

A new particles capability just got merged in to `develop`, but only with unit testing. There's also an example particles implementation that would be good for a regression test. This PR adds that regression test.

One thing to point out now -- this test will expose potential reproducibility issues with the Kokkos RNG implementation. This appears to be reproducible at least on CPUs locally for me. 

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted

### DIFF
--- a/tst/regression/CMakeLists.txt
+++ b/tst/regression/CMakeLists.txt
@@ -76,6 +76,15 @@ if (ENABLE_HDF5)
   --num_steps 25")
   list(APPEND EXTRA_TEST_LABELS "")
 
+  # Particles example
+  list(APPEND TEST_DIRS particles)
+  list(APPEND TEST_PROCS 1) # Temporarily force scalar testing
+  list(APPEND TEST_ARGS "--driver ${PROJECT_BINARY_DIR}/example/particles/particles-example \
+    --driver_input ${CMAKE_CURRENT_SOURCE_DIR}/test_suites/particles/parthinput.particles \
+    --num_steps 1")
+  list(APPEND EXTRA_TEST_LABELS "")
+
+  # HDF5 output
   list(APPEND TEST_DIRS output_hdf5)
   list(APPEND TEST_PROCS ${NUM_MPI_PROC_TESTING})
   list(APPEND TEST_ARGS "--driver ${PROJECT_BINARY_DIR}/example/advection/advection-example \
--- a/tst/regression/test_suites/particles/particles.py
+++ b/tst/regression/test_suites/particles/particles.py
@@ -1,0 +1,65 @@
+#========================================================================================
+# Athena++ astrophysical MHD code
+# Copyright(C) 2014 James M. Stone <jmstone@princeton.edu> and other code contributors
+# Licensed under the 3-clause BSD License, see LICENSE file for details
+#========================================================================================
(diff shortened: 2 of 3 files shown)


brryan:
@JoshuaSBrown @pgrete I've made a first pass at this but I'm not sure how to ~~handle~~ commit the gold file that I have locally.

pgrete:
What RNG problems are you referring to?
IIRC at least on the host the numbers should be deterministic.
And for the purpose of testing the runs on GPUs could create the numbers on the host and transfer to the GPU so that they're also deterministic there.

Regarding the test itself, it'd great if you could add some more comments/documentation to the source files.
Based on the current info I'm not sure what this test is currently testing.

Finally, regarding the "gold" file. Once this PR is ready for a merge we will upload the new gold standard as a "release" on GitHub  (https://github.com/lanl/parthenon/releases) and then make a final update to this PR increasing the version of the gold standard to reflect the latest version.

JoshuaSBrown (approved):
LGTM, ditto what Phil mentioned. Thanks for doing this, this will ensure that your work isn't broken in the future without warning.

tst/regression/CMakeLists.txt
```diff
@@ -76,6 +76,15 @@ if (ENABLE_HDF5)
   --num_steps 25")
   list(APPEND EXTRA_TEST_LABELS "")
 
+  # Particles example
+  list(APPEND TEST_DIRS particles)
+  list(APPEND TEST_PROCS 1) # Temporarily force scalar testing
```
JoshuaSBrown, line 81: How temporary is this, otherwise we will essentially be running two tests that are doing the same thing? Also how long does this test actually take to run?

brryan: My next task with this particles implementation is to add MPI communication, so this should be temporary only until the next particles PR (though that may take a little while to put together). This test runs in less than a second on my laptop, so I don't think this will cause any issues? Though I might increase the particle number a bit for good measure. I didn't see another simple way to enroll a regression test as scalar-only.

pgrete: There are `mpi-yes` and `mpi-no` so for a "scalar" test the `mpi-no` label should be sufficient that the test is only run when not compiled for MPI.

tst/regression/test_suites/particles/particles.py
```diff
@@ -0,0 +1,65 @@
...
+
+        # Determine path to parthenon installation
+        # Fallback to relative path on failure
+        try:
+            parthenonPath = os.path.realpath(__file__)
+            idx = parthenonPath.rindex('/parthenon/')
+            parthenonPath = os.path.join(parthenonPath[:idx],'parthenon')
+        except ValueError:
+            baseDir = os.path.dirname(__file__)
+            parthenonPath = baseDir + '/../../../..'
```
JoshuaSBrown, line 49: @brryan I just merged in a pr that so you can get the parthenon path directly from the parameters struct, so you don't have to put all this logic here. https://github.com/lanl/parthenon/pull/340

brryan: @JoshuaSBrown Thanks for letting me know, I'll make that change

jlippuner:
Will restart this under new particle interface

PR #335, opened 2020-10-23
Add Darwin machine config with support for CPU builds on x86 and power9

### LINKED ISSUE #326: Add Machine Config file for Darwin
Add config file that reproduces the ci on power9 nodes.

AndrewGaspar:
## PR Summary

This change adds a machine config file for LANL's Darwin cluster and documents building on that cluster. Currently only CPU builds for x86 and power9 are supported, but before this is merged I want to support CUDA builds on power9, too.

Additionally, it revamps the `MACHINE_CFG` code a little bit. Most importantly, it adds a capability to specify a `MACHINE_CFG` environment variable that specifies a path relative to the parthenon source directory, and it also caches the full machine config path so that changing the current working directory doesn't cause that config path to be re-interpreted. (e.g. if you configured from the source directory, then changed your directory to the build directory and a re-configure was triggered)

LANL folks, if you want to see how the dependencies are being installed, check out this URL: https://xcp-gitlab.lanl.gov/eap-oss/parthenon-project

I intend to add more documentation next week.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


docs/building.md
```diff
@@ -264,7 +318,7 @@ $ jsrun -p 2 -g 1 -c 20 -M "-gpu" ./example/advection/advection-example -i ../ex
 ```bash
 # configure and build
 $ mkdir build-cuda && cd build-cuda
-$ cmake -DCMAKE_BUILD_TYPE=Release -DMACHINE_CFG=${PARTHENON_ROOT}/cmake/machinecfg/Summit.cma
```
pgrete, line 267: I think there's an extra linebreak here (the missing `ke` is on the next line)

pgrete (approved):
:+1: Looks good!

CMakeLists.txt
```diff
@@ -17,23 +17,37 @@ cmake_minimum_required(VERSION 3.12)
...
-  endif()
+  set(USE_MACHINE_CFG ${MACHINE_CFG})
 elseif (DEFINED ENV{MACHINE_CFG})
-  if(EXISTS "$ENV{MACHINE_CFG}")
-    include($ENV{MACHINE_CFG})
-  else()
-    message(FATAL_ERROR "Given machine configuration from environment variable "
-      "MACHINE_CFG at $ENV{MACHINE_CFG} not found.")
+  set(USE_MACHINE_CFG $ENV{MACHINE_CFG})
+endif()
```
JoshuaSBrown: I think this logic should probably be used on a lot of the variables we set throughout our cmake config.

cmake/machinecfg/Darwin.cmake
```diff
@@ -0,0 +1,157 @@
...
+        CACHE STRING "MPI Location")
+endif()
+
+# Kokkos settings
+if (DARWIN_ARCH STREQUAL "x86_64")
+    set(Kokkos_ARCH_HSW ON CACHE BOOL "Target Haswell")
+elseif(DARWIN_ARCH STREQUAL "ppc64le")
+    # set(Kokkos_ENABLE_CUDA ON CACHE BOOL "Cuda")
+    set(Kokkos_ARCH_POWER9 ON CACHE BOOL "Target Power9")
+    # set(Kokkos_ARCH_VOLTA70 ON CACHE BOOL "Target V100s")
```
JoshuaSBrown: Are these commented out lines needed?

cmake/machinecfg/Darwin.cmake
```diff
@@ -0,0 +1,157 @@
...
+#           `x86_64` (assumes at least haswell)
+#           `ppc64le` (assumes power9 + volta gpus)
+# - `DARWIN_VIEW_DATE` - The date the dependencies were installed.
+#       Default:
+#           unset, which results in using the view associated with your
+#           current commit. See `DARWIN_VIEW_DATE_LATEST`
+# - `DARWIN_COMPILER` - Compiler family to use
+#       Default: "GCC"
+#       Possible Values: "GCC", "GCC9"
+
```
JoshuaSBrown: It might be nice to add the ${MACHINE_VARIANT} variable that @pgrete makes use of in the config files he has. To indicate what resources are actually being used, unless you think that is redundant with DARWIN_ARCH. I'm just thinking in terms of standardizing some of these variables.

cmake/machinecfg/Darwin.cmake
JoshuaSBrown: Also if by default we are going to be running with 2 mpi procs, it might be worth setting. 
```cmake
set(NUM_GPU_DEVICES_PER_NODE "2" CACHE STRING "Number of gpu devices to use when testing if built with Kokkos_ENABLE_CUDA")
```

The default is a single gpu.

AndrewGaspar: Yeah, I'm working on re-enabling these right now.

Good suggestion.

docs/building.md
AndrewGaspar: oh, doh, yep.

AndrewGaspar:
Darwin Power9 builds for GPU now... A couple more things I'd like to before merge:
* Add clang-format builds
* Run tests with 1 OpenMP thread

cielling (approved):

cielling (approved):

cmake/TestSetup.cmake
```diff
@@ -109,10 +113,13 @@ function(setup_test_mpi nproc dir arg extra_labels)
...
     list(APPEND labels "${extra_labels}")
 
-    if( "${Kokkos_ENABLE_CUDA}" )
+    if(Kokkos_ENABLE_CUDA)
       set(PARTHENON_KOKKOS_TEST_ARGS "--kokkos-num-devices=${NUM_GPU_DEVICES_PER_NODE}")
       list(APPEND labels "cuda")
     endif()
+    if (Kokkos_ENABLE_OPENMP)
+      set(PARTHENON_KOKKOS_TEST_ARGS "${PARTHENON_KOKKOS_TEST_ARGS} --kokkos-threads=${NUM_OMP_THREADS_PER_RANK}")
+    endif()
```
pgrete, line 122: If Cuda and OpenMP are enabled, aren't the options overridden?

CMakeLists.txt
```diff
@@ -17,23 +17,36 @@ cmake_minimum_required(VERSION 3.12)
...
+
+if (USE_MACHINE_CFG)
+  # The `:PAR_ROOT:` string allows the devleper to specify a relative path to
```
Yurlungur: Nice improvement!

Why use `:` to delineate`:PAR_ROOT:` instead of something else? Just curious. The more common syntax I've seen is `@`

cmake/PythonModuleCheck.cmake
```diff
@@ -22,13 +22,13 @@ function(required_python_modules_found module_list)
       execute_process(COMMAND ${Python3_EXECUTABLE} -c "import ${module}"
         RESULT_VARIABLE IMPORT_MODULE ERROR_QUIET)
     
-      if(${IMPORT_MODULE} EQUAL 1)
+      if(NOT ${IMPORT_MODULE} EQUAL 0)
```
Yurlungur, line 25: why this change?

cmake/machinecfg/Darwin.cmake
```diff
@@ -0,0 +1,258 @@
...
+execute_process(
+    COMMAND nvidia-smi -L
+    OUTPUT_VARIABLE FOUND_GPUS)
+
+if (FOUND_GPUS MATCHES "GPU [0-9]")
+    set(DARWIN_CUDA_DEFAULT ON)
+else()
+    set(DARWIN_CUDA_DEFAULT OFF)
+endif()
```
Yurlungur: Nice.

docs/building.md
```diff
@@ -215,6 +215,60 @@ Currently Loaded Modules:
...
+$ salloc -p power9
+```
+
+#### Set-Up Environment (Optional, but Still Recommended, for Non-CUDA Builds)
+
+You can import all tools you need to start building with by sourcing the
+project `.bashrc`:
+
+```bash
+$ source /projects/parthenon-int/parthenon-project/.bashrc
```
Yurlungur, line 239: it might be nice to store this somewhere on the parthenon repo (or similar), maybe in the wiki? I'd like the environment code and the cmake tied together somehow.

Yurlungur (approved):
Thanks for doing this!

cmake/PythonModuleCheck.cmake
AndrewGaspar: Oh, Geoff hit an error with a python module, but it wasn't caught because the exit code was something other than 1. So this change just covers a bigger umbrella of non-success codes.

CMakeLists.txt
AndrewGaspar: Yeah, I can do that instead - it was just the first thing that came to mind.

docs/building.md
AndrewGaspar: It's stored in a separate, internal git repo that manages the project space layout.

So, the script is intentionally very brief and it doesn't actually load all that much into the user's environment - just cmake, ninja, and a recent git build, as well as setting `MACHINE_CFG`. Because it's sparse like this, the goal is you should always be able to use the latest .bashrc with any version of Parthenon. Instead, coupling between versions of Parthenon and dependencies is achieved by encoding the view "date" in the CMake file here.

One reason I wanted to put it in a known directory on darwin is that users can stick that exact string into their profiles and not have to think about it.

Thoughts?

See: https://xcp-gitlab.lanl.gov/eap-oss/parthenon-project

Currently working on a README for it.

cmake/TestSetup.cmake
AndrewGaspar: Do you mean `--kokkos-threads` would override `--kokkos-num-devices`? The second set command appends --kokkos-threads to the previously contents of `PARTHENON_KOKKOS_TEST_ARGS`. See:
```cmake
set(PARTHENON_KOKKOS_TEST_ARGS "${PARTHENON_KOKKOS_TEST_ARGS}
```

cmake/machinecfg/Darwin.cmake
AndrewGaspar: Oh, I need to fix this for the x86 volta nodes - they only have 1 GPU.

cmake/PythonModuleCheck.cmake
AndrewGaspar: cc @JoshuaSBrown

CMakeLists.txt
Yurlungur: I think I would prefer `@`, but not a big deal either way.

cmake/PythonModuleCheck.cmake
Yurlungur: Got it.

cmake/TestSetup.cmake
```diff
@@ -48,9 +48,13 @@ function(setup_test dir arg extra_labels)
   separate_arguments(arg) 
   list(APPEND labels "regression;mpi-no")
   list(APPEND labels "${extra_labels}")
+  if (Kokkos_ENABLE_OPENMP)
+    set(PARTHENON_KOKKOS_TEST_ARGS "${PARTHENON_KOKKOS_TEST_ARGS} --kokkos-threads=${NUM_OMP_THREADS_PER_RANK}")
+  endif()
```
JoshuaSBrown, line 53: Nice.

CMakeLists.txt
```diff
@@ -17,23 +17,36 @@ cmake_minimum_required(VERSION 3.12)
...
+
+if (MACHINE_CFG_FULL)
+  include(${MACHINE_CFG_FULL})
 else()
   message(WARNING "Not using any machine configuration. Consider creating a configuration "
     "file following the examples in ${PROJECT_SOURCE_DIR}/cmake/machine_cfgs/ and then "
-    "point the MACHINE_CFG variable to your custom file."
+    "point the MACHINE_CFG cmake or environment variable to your custom file."
     "Note, that the machine file can be placed in any directory (also outside the repo).")
 endif()
```
JoshuaSBrown: Can we move this logic into a cmake module file? That's my final suggestion for this pr, let me test it real quick than I'm golden.

cmake/PythonModuleCheck.cmake
JoshuaSBrown: Yes, that is correct.

JoshuaSBrown (approved):
I had no problems on power9 or x86 on darwin unless my bashrc file preloaded modules.

docs/building.md
Yurlungur: > One reason I wanted to put it in a known directory on darwin is that users can stick that exact string into their profiles and not have to think about it.

I agree with this strongly.

> So, the script is intentionally very brief and it doesn't actually load all that much into the user's environment

Oh yeah, I see. It's very simple. I guess since there's basically nothing in the script, it's probably fine not to have them together in the same repo. Maybe if the user doesn't want to use the bashrc, we could provide documentation somewhere for what it's doing?

Also, why Ninja and not make?

JoshuaSBrown: I'm pretty sure Ninja is faster.

Yurlungur: But I hate change. :P

AndrewGaspar: Yeah, Ninja completes a clean CUDA build ~70% faster on power9, in my experience.

AndrewGaspar: It's essentially CMake's fault, unfortunately - its Make generator won't overlap compilation of source files from dependent targets. So it builds all of Kokkos, links it, then builds all of parthenon, links it, and then builds the examples and tests. The Ninja generator doesn't suffer from this.

CMakeLists.txt
AndrewGaspar: Fixed.

cmake/machinecfg/Darwin.cmake
AndrewGaspar: Ah, yeah, currently it always builds MPI. hmmm... this will take a little while, but I'll do this.

CMakeLists.txt
AndrewGaspar: Talked offline - this can be resolved later.

cmake/machinecfg/Darwin.cmake
AndrewGaspar: We decided offline to add this later when we need it.

PR #340, opened 2020-10-27
Add parthenon root dir to parameters

### LINKED ISSUE #337: Add parthenon_path to python TestCase parameters structure
The root parthenon path is used by several of the regression tests, it would be good to provide this automatically so that everyone is not being required to redefine it.

JoshuaSBrown:
## PR Summary

This is a small pr that moves some logic from some of the restart files to the python test_case class. Namely, the need to locate the parthenon root dir. This was used in the output_hdf5.py file as well as in the currently open pr related to the particles regression test. 

This change is to simply provide additional functionality when writing regression tests. 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/tst/regression/test_suites/output_hdf5/output_hdf5.py
+++ b/tst/regression/test_suites/output_hdf5/output_hdf5.py
@@ -81,16 +81,7 @@ class TestCase(utils.test_case.TestCaseAbs):
         analyze_status = True
         print(os.getcwd())
 
-        # Determine path to parthenon installation
-        # Fallback to relative path on failure
-        try:
-            parthenonPath = os.path.realpath(__file__)
-            idx = parthenonPath.rindex('/parthenon/')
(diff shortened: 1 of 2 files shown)


pgrete:
I think this change makes sense and now I'm wondering if there's a clean way we can inform the python files from CMake about the Parthenon root directory.
I think I remember that there was recently a discussion about the Parthenon root dir variable that may be related/helpful, wasn't it, ping @AndrewGaspar ?

JoshuaSBrown:
We could pass the path directly in when the tests are created it would be simple to do, I don't really see very many cases where this implementation would really ever cause problems. Unless someone decided to have another folder named parthenon within the test folder, or containing the test folder.

Yurlungur (approved):

pgrete (approved):

PR #341, opened 2020-10-27
[WIP] Joshua s brown/refactor plm functions

### LINKED ISSUE #182: Optimization of PLM functions
As pointed out by @JoshuaSBrown  in #136 the PLM functions contain repetitive (and currently unused) code that could be optimized/clean up.

JoshuaSBrown:
## PR Summary

This pr attempts to reduce the amount of redundant code in the plm source file. There was a lot of boiler plate that was refactored into smaller hopefully clearer functions. 

## PR Checklist

- [x] Code passes cpplint
- [NA] New features are documented.
- [NA] Adds a test for any bugs fixed. Adds tests for new features.
- [x] Code is formatted

### DIFF
--- a/src/reconstruct/plm.cpp
+++ b/src/reconstruct/plm.cpp
@@ -26,247 +26,238 @@
 #include "mesh/mesh.hpp"
 #include "reconstruct/reconstruction.hpp"
 
+namespace parthenon {
+
 namespace {
 
-  class enum Dimension {
-    X1,
-    X2,
-    X3
-  };
-    
-  struct DataInternal {
+enum class Dimension { X1, X2, X3 };
 
-    const int k, j, il, iu, nu;
-    ParArrayND<Real> &q;
-    ParArrayND<Real> &qc;
-    ParArrayND<Real> &ql;
-    ParArrayND<Real> &qr;
+struct DataInternal {
+  const int k, j, il, iu, nu;
+  const ParArrayND<Real> &q;
+  ParArrayND<Real> &qc;
+  ParArrayND<Real> &ql;
+  ParArrayND<Real> &qr;
 
-    ParArrayND<Real> &dql;
-    ParArrayND<Real> &dqr;
-    ParArrayND<Real> &dqm;
-    const Coordinates_t & coords;
-    DataInternal(
-        const int k, const int j, const int il, const int iu, const int nu,
-        ParArrayND<Real> & q, ParArrayND<Real> & qc, ParArrayND<Real> & ql, ParArrayND<Real> & qr,
-        ParArrayND<Real> & dql, ParArrayND<Real> & dqr, ParArrayND<Real> & dqm, const Coordinates_t & coords) :
-        k(k), j(j), il(il), iu(iu), nu(nu),
-        q(q), qc(qc), ql(ql), qr(qr),
-        dql(dql), dqr(dqr), dqm(dqm), coords(coords) {};
-  };
+  ParArrayND<Real> &dql;
+  ParArrayND<Real> &dqr;
+  ParArrayND<Real> &dqm;
+  const Coordinates_t &coords;
+  DataInternal(const int k, const int j, const int il, const int iu, const int nu,
+      const ParArrayND<Real> &q, ParArrayND<Real> &qc, ParArrayND<Real> &ql,
+               ParArrayND<Real> &qr, ParArrayND<Real> &dql, ParArrayND<Real> &dqr,
+               ParArrayND<Real> &dqm, const Coordinates_t &coords)
+    : k(k), j(j), il(il), iu(iu), nu(nu),
+    q(q), qc(qc), ql(ql), qr(qr), dql(dql),
+    dqr(dqr), dqm(dqm), coords(coords){}
+};
 
-  void init_dql_and_dqr_and_qc_(DataInternal & data,const Dimension dimension) {
 
-    int delta_i = 0;
-    int delta_j = 0;
-    int delta_k = 0;
-    if ( dimension == Dimension::X1 ){
-      delta_i = 1;
-    }else if (dimension == Dimension::X2 ) {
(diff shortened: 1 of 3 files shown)


JoshuaSBrown:
@par-hermes format

JoshuaSBrown:
Full pipeline https://gitlab.com/pgrete/parthenon/-/pipelines/208418086

pgrete:
I'm referring to @jmstone here for review as he also refactored the PLM functions.

Independently, given that this is a performance critical part of Parthenon, it'd be great to see some profiling/benchmarking results of those function on both CPUs and GPUs comparing to the original implementation.

JoshuaSBrown:
> I'm referring to @jmstone here for review as he also refactored the PLM functions.
> 
> Independently, given that this is a performance critical part of Parthenon, it'd be great to see some profiling/benchmarking results of those function on both CPUs and GPUs comparing to the original implementation.

This is a fair point.

jmstone:
This has changed a lot from the original implementation in Athena++ so I don't have much to add.  In the Kokkos version I wrote for my test code, I restricted the reconstruction to a uniform Cartesian grid and use only one function for all dimensions, so I am afraid there isn't much comparison that can be made between the two.  I agree that simplifying/refactoring these functions is a useful exercise!

Yurlungur:
Thanks for taking a shot at this. This is certainly worth rethinking. I'm looking through now.

I agree with @pgrete that some performance tests are probably very much in order.

src/reconstruct/plm.cpp
```diff
@@ -28,226 +28,234 @@
...
 
-void Reconstruction::PiecewiseLinearX1(const int k, const int j, const int il,
-                                       const int iu, const ParArrayND<Real> &q,
-                                       ParArrayND<Real> &ql, ParArrayND<Real> &qr) {
-  auto pmb = GetBlockPointer();
-  auto &coords = pmb->coords;
-  // set work arrays to shallow copies of scratch arrays
-  ParArrayND<Real> &qc = scr1_ni_, &dql = scr2_ni_, &dqr = scr3_ni_, &dqm = scr4_ni_;
-  const int nu = q.GetDim(4) - 1;
+enum class Dimension { X1, X2, X3 };
```
Yurlungur: This already exists as `X1DIR`, `X2DIR`, `X3DIR` in `defs.hpp`. I think we should use that one for this kind of thing whenever possible.

I know it's an enum there, not an enum class, but I don't like introducing new local notation. I think we should use the same notation code-wide.

src/reconstruct/plm.cpp
```diff
@@ -28,226 +28,234 @@
...
+
+void init_dql_and_dqr_and_qc_(DataInternal &data, const Dimension dimension) {
+  int delta_i = 0;
+  int delta_j = 0;
+  int delta_k = 0;
+  if (dimension == Dimension::X1) {
+    delta_i = 1;
+  } else if (dimension == Dimension::X2) {
+    delta_j = 1;
+  } else {
```
Yurlungur, line 62: I like having a comment here, something like
```suggestion
  } else { // if (dimmension == Dimension::X3)
```
that said, I'd rather replace `Dimension::*` with `X*DIR` for consistency.

src/reconstruct/plm.cpp
```diff
@@ -28,226 +28,234 @@
...
+    for (int i = data.il; i <= data.iu; ++i) {
       // renamed dw* -> dq* from plm.cpp
-      dql(n, i) = (q(n, k, j, i) - q(n, k, j, i - 1));
-      dqr(n, i) = (q(n, k, j, i + 1) - q(n, k, j, i));
-      qc(n, i) = q(n, k, j, i);
+      data.dql(n, i) =
+          (data.q(n, k, j, i) - data.q(n, k - delta_k, j - delta_j, i - delta_i));
+      data.dqr(n, i) =
+          (data.q(n, k + delta_k, j + delta_j, i + delta_i) - data.q(n, k, j, i));
+      data.qc(n, i) = data.q(n, k, j, i);
```
Yurlungur, line 77: This is a host-side loop, but `data.q` lives on device. If compiled for GPU, this will die at runtime.

src/reconstruct/plm.cpp
```diff
@@ -28,226 +28,234 @@
...
+#pragma omp simd simdlen(SIMD_WIDTH)
+      for (int i = data.il; i <= data.iu; ++i) {
+        const Real dqF = data.dqr(n, i) * dxF;
+        const Real dqB = data.dql(n, i) * dxB;
+        const Real dq2 = dqF * dqB;
+        // (modified) VL limiter (Mignone eq 37)
+        data.dqm(n, i) =
+            (dq2 * (cf * dqB + cb * dqF) / (SQR(dqB) + SQR(dqF) + dq2 * (cf + cb - 2.0)));
+        if (dq2 <= 0.0)
+          data.dqm(n, i) = 0.0; // ---> no concern for divide-by-0 in above line
```
Yurlungur: As above, this loop is going to fail on GPU

Yurlungur (changes requested):
Overall, I like the idea of cleaning this up to be less dimension-specific. However, I'm pretty sure we're not using this code, so I don't think it's being tested. And it will for sure die on GPU. I'm pretty sure we use the header-only version `plm_header_only.hpp` or something.

I also have a few nitpicks below.

JoshuaSBrown:
This pr is being put on hold. It looks like there is some discussion on how to treat the coordinate system which should probably be resolved before any optimizations are done here. 

E.g. whether to use a fixed grid and do transforms on top, or don't use a fixed grid etc.

@Yurlungur correct me if I'm wrong.

Yurlungur:
That seems reasonable to me. I think we could refactor `plm_inline.hpp` now, but it could be much simpler if certain conditions on the coordinate systems hold.

PR #352, opened 2020-11-03
Upgrade Kokkos to 3.2.00 and clean up all warnings with nvcc_wrapper

### LINKED ISSUE #327: Update Kokkos submodule version
The kokkos submodule version needs to be updated to major version that has better cuda support.

AndrewGaspar:
## PR Summary

This change updates Kokkos to the latest release and fixes all warnings encountered on Darwin with cuda 11.0.

To review, it's probably easiest to just review one commit at a time.

Most common issues:
- unused variables
- implicit "calls" to host destructors (optimized out in practice, I think?)

Fixes #327

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


AndrewGaspar:
@pgrete is it OK to update Kokkos? I know we were on some unmerged branch of Kokkos at one point, but I think that's defunct now since we're not targeting streams anymore.

src/bvals/bvals_refine.cpp
```diff
@@ -357,44 +357,6 @@ void BoundaryValues::ProlongateGhostCells(const NeighborBlock &nb, int si, int e
...
-  // KGF: COUPLING OF QUANTITIES (must be manually specified)
-  // Field prolongation completed, calculate cell centered fields
-  // TODO(KGF): passing nullptrs (pf) if no MHD (coarse_* no longer in MeshRefinement)
-  // (may be fine to unconditionally directly set to pmb->pfield now. see above comment)
-
-  // KGF: COUPLING OF QUANTITIES (must be manually specified)
-  // calculate conservative variables
-  // pmb->peos->PrimitiveToConserved(ph->w, pf->bcc, ph->u, pmb->pcoord,
-  //                                fsi, fei, fsj, fej, fsk, fek);
-  return;
```
Yurlungur, line 399: Rather than remove this, I would comment it out. We may want to keep this as referencefor the MHD machinery.

src/mesh/meshblock.hpp
```diff
@@ -65,9 +66,22 @@ KOKKOS_INLINE_FUNCTION void par_for_inner(const team_mbr_t &team_member, const i
...
+class MeshBlockDevice {
+ public:
+  // Inner loop default pattern
+  template <typename Function>
+  KOKKOS_INLINE_FUNCTION void par_for_inner(const team_mbr_t &team_member, const int &il,
+                                            const int &iu, const Function &function) const
+      noexcept {
+    parthenon::par_for_inner(DEFAULT_INNER_LOOP_PATTERN, team_member, il, iu, function);
+  }
+};
```
Yurlungur: This seems weird to me. Why not just expose a meshblock-independent `par_for_inner`? As this `MeshBlockDevice` has no internal state?

src/outputs/parthenon_hdf5.cpp
```diff
@@ -245,28 +245,25 @@ void PHDF5Output::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm) {
...
      pass some information onto the underlying MPI_File_open call */
   MPI_Info FILE_INFO_TEMPLATE;
-  int ierr;
-  MPI_Status stat;
-  ierr = MPI_Info_create(&FILE_INFO_TEMPLATE);
-  ierr = H5Pset_sieve_buf_size(acc_file, 262144);
-  ierr = H5Pset_alignment(acc_file, 524288, 262144);
+  MPI_Info_create(&FILE_INFO_TEMPLATE);
+  H5Pset_sieve_buf_size(acc_file, 262144);
+  H5Pset_alignment(acc_file, 524288, 262144);
```
Yurlungur: I don't like removing `ierr` here... I understand why you did it, to remove the warning. But instead, I think we should check `ierr` and throw if it's not `H5_SUCCESS`

(And same elsewhere where it's been removed, obviously.)

src/outputs/parthenon_hdf5.cpp
```diff
@@ -276,40 +273,40 @@ void PHDF5Output::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm) {
 
   int max_level = pm->GetCurrentLevel() - pm->GetRootLevel();
   if (tm != nullptr) {
-    status = writeH5AI32("NCycle", &(tm->ncycle), file, localDSpace, myDSet);
-    status = writeH5AF64("Time", &(tm->time), file, localDSpace, myDSet);
+    writeH5AI32("NCycle", &(tm->ncycle), file, localDSpace, myDSet);
+    writeH5AF64("Time", &(tm->time), file, localDSpace, myDSet);
   }
```
Yurlungur: Same with `status` as `ierr`. We should check that it's a success and die if it's not, rather than just not looking.

src/reconstruct/reconstruction.cpp
```diff
@@ -33,15 +33,6 @@
 
 namespace parthenon {
 
-namespace {
-
-// TODO(felker): replace these hand-rolled linear algebra routines with a real library
-constexpr Real lu_tol = 3e-16;
-int DoolittleLUPDecompose(Real **a, int n, int *pivot);
-void DoolittleLUPSolve(Real **lu, int *pivot, Real *b, int n, Real *x);
```
Yurlungur, line 42: I think it's fine to remove this. But just to check, @pgrete @jmstone is there a reason to keep these around?

Yurlungur (changes requested):
Thanks for doing this! It'll be great to have fewer warnings. Some nitpicks below.

src/mesh/meshblock.hpp
AndrewGaspar: I'm happy to. It just seemed deliberate that it was made a member function, so I wasn't sure if there was some longer term plan.

@pgrete ?

pgrete:
> @pgrete is it OK to update Kokkos? I know we were on some unmerged branch of Kokkos at one point, but I think that's defunct now since we're not targeting streams anymore.

Happy to update. Some features we actually need (really soon), e.g., with respect to the (implicit) fencing behavior.

src/bvals/bvals_refine.cpp
pgrete: Kind of related: Is the code above actually used anywhere?

Otherwise, I think it's fine to remove the code as (from what I can tell) those indices are set to be used in a physics related piece of code and, thus, should be implemented downstream.
Otherwise, if they're more general then we should fix this implementation here (not in this PR).

src/interface/swarm.cpp
```diff
@@ -243,8 +243,6 @@ ParArrayND<bool> Swarm::AddEmptyParticles(const int num_to_add) {
 // No particles removed: nmax_active_index unchanged
 // Particles removed: nmax_active_index is new max active index
 void Swarm::RemoveMarkedParticles() {
-  int new_max_active_index = -1; // TODO(BRR) this is a magic number, needed for Defrag()
```
pgrete, line 246: @brryan is this "TODO" important, or, in other words, are you going to miss something for a future update if the note is gone?

src/outputs/parthenon_hdf5.cpp
pgrete: I also think it's worth keeping this.
Could we may just to a `ierr &= ...` and do a final check or similar?

src/reconstruct/reconstruction.cpp
pgrete: I think  this can go (and being reintroduced later if required).

src/mesh/meshblock.hpp
pgrete: The main reason it was a member function is that it uses `DEFAULT_INNER_LOOP_PATTERN` which is something we set at compile time and (some time ago) we decided to only provide abstractions in the `kokkos_abstraction` layer but nothing that changes behavior.

It's not entirely clear what we gain from using the MeshBlockDevice and think it's potentially confusing.
The inner loop never uses a "device" as that information is already carried over by the `team_member`.
So I'm in favor of keeping the original `pmb->par_for_inner` (as it is consistent with the use of a `pmb->par_for_outer` from my point of view).

pgrete (changes requested):
Thanks for taking care of this (I also had this on my long term todo list)!
This (automatically) also removes all warnings in the CI.
With the latter in place how about we introduce/enable the "warnings-are-errors" compile option (only for the short unit tests so that the short regression tests for correctness still compile and run)?

src/outputs/io_wrapper.cpp
```diff
@@ -81,7 +81,7 @@ int IOWrapper::Open(const char *fname, FileMode rw) {
 //! \fn int IOWrapper::Read(void *buf, IOWrapperSizeT size, IOWrapperSizeT count)
 //  \brief wrapper for {MPI_File_read} versus {std::fread}
 
-std::size_t IOWrapper::Read(void *buf, IOWrapperSizeT size, IOWrapperSizeT count) {
+std::ptrdiff_t IOWrapper::Read(void *buf, IOWrapperSizeT size, IOWrapperSizeT count) {
 #ifdef MPI_PARALLEL
```
JoshuaSBrown: Nice

src/mesh/domain.hpp
```diff
@@ -58,12 +58,11 @@ class IndexShape {
...
   std::array<int, NDIM> entire_ncells_;
 
-  KOKKOS_INLINE_FUNCTION bool DimensionProvided_(const std::vector<int> &interior_dims,
-                                                 int dim) {
+  bool DimensionProvided_(const std::vector<int> &interior_dims, int dim) {
     return dim <= interior_dims.size();
   }
 
-  KOKKOS_INLINE_FUNCTION void MakeZeroDimensional_(int index) {
+  void MakeZeroDimensional_(int index) {
```
JoshuaSBrown: make index const.

src/outputs/io_wrapper.cpp
```diff
@@ -97,7 +97,7 @@ std::size_t IOWrapper::Read(void *buf, IOWrapperSizeT size, IOWrapperSizeT count
 //! \fn int IOWrapper::Read_all(void *buf, IOWrapperSizeT size, IOWrapperSizeT count)
 //  \brief wrapper for {MPI_File_read_all} versus {std::fread}
 
-std::size_t IOWrapper::Read_all(void *buf, IOWrapperSizeT size, IOWrapperSizeT count) {
+std::ptrdiff_t IOWrapper::Read_all(void *buf, IOWrapperSizeT size, IOWrapperSizeT count) {
```
JoshuaSBrown: Can we make count and size const, here and for the below methods.

JoshuaSBrown (approved):
I just have a few minor suggestions, thanks this is an improvement.

AndrewGaspar:
> hanks for taking care of this (I also had this on my long term todo list)!

Yeah, compiler output with nvcc was basically unreadable.

AndrewGaspar:
I'll add new parthenon error checking macros for MPI and HDF5.

AndrewGaspar:
@pgrete - take a look at this thread. https://github.com/lanl/parthenon/pull/352#discussion_r517420460

I wanted to understand the purpose of attaching `par_for_inner` to `MeshBlock`.

AndrewGaspar:
(GitHub renders review-attached thread replies in the dumbest way possible)

src/mesh/meshblock.hpp
AndrewGaspar: > So I'm in favor of keeping the original pmb->par_for_inner (as it is consistent with the use of a pmb->par_for_outer from my point of view).

That was deliberately removed - let me show you the warning we get when you use `pmb->par_for_inner` in `advection_package.cpp`:

```
../example/advection/advection_package.cpp(323): warning: calling a __host__ function from a __host__ __device__ function is not allowed

../example/advection/advection_package.cpp(326): warning: calling a __host__ function from a __host__ __device__ function is not allowed

../example/advection/advection_package.cpp(355): warning: calling a __host__ function from a __host__ __device__ function is not allowed

../example/advection/advection_package.cpp(359): warning: calling a __host__ function from a __host__ __device__ function is not allowed

../example/advection/advection_package.cpp(390): warning: calling a __host__ function from a __host__ __device__ function is not allowed

../example/advection/advection_package.cpp(394): warning: calling a __host__ function from a __host__ __device__ function is not allowed

/usr/local/cuda-11.0/bin/../targets/ppc64le-linux/include/vector_types.h(421): warning: calling a __host__ function("std::__shared_count<( ::__gnu_cxx::_Lock_policy)2> ::~__shared_count") from a __host__ __device__ function("") is not allowed

/usr/local/cuda-11.0/bin/../targets/ppc64le-linux/include/vector_types.h(421): warning: calling a __host__ function("std::__shared_count<( ::__gnu_cxx::_Lock_policy)2> ::~__shared_count") from a __host__ __device__ function("") is not allowed

/usr/local/cuda-11.0/bin/../targets/ppc64le-linux/include/vector_types.h(421): warning: calling a __host__ function("std::__shared_count<( ::__gnu_cxx::_Lock_policy)2> ::~__shared_count") from a __host__ __device__ function("") is not allowed

../example/advection/advection_package.cpp(330): warning: calling a __host__ function("std::__shared_count<( ::__gnu_cxx::_Lock_policy)2> ::__shared_count") from a __host__ __device__ function("") is not allowed

../example/advection/advection_package.cpp(330): warning: calling a __host__ function("std::__shared_count<( ::__gnu_cxx::_Lock_policy)2> ::~__shared_count") from a __host__ __device__ function("") is not allowed

../example/advection/advection_package.cpp(364): warning: calling a __host__ function("std::__shared_count<( ::__gnu_cxx::_Lock_policy)2> ::__shared_count") from a __host__ __device__ function("") is not allowed

../example/advection/advection_package.cpp(364): warning: calling a __host__ function("std::__shared_count<( ::__gnu_cxx::_Lock_policy)2> ::~__shared_count") from a __host__ __device__ function("") is not allowed

../example/advection/advection_package.cpp(399): warning: calling a __host__ function("std::__shared_count<( ::__gnu_cxx::_Lock_policy)2> ::__shared_count") from a __host__ __device__ function("") is not allowed

../example/advection/advection_package.cpp(399): warning: calling a __host__ function("std::__shared_count<( ::__gnu_cxx::_Lock_policy)2> ::~__shared_count") from a __host__ __device__ function("") is not allowed

../external/Kokkos/core/src/Kokkos_View.hpp(979): warning: calling a __host__ function("std::__shared_count<( ::__gnu_cxx::_Lock_policy)2> ::~__shared_count") from a __host__ __device__ function("") is not allowed

../external/Kokkos/core/src/Kokkos_View.hpp(979): warning: calling a __host__ function("std::__shared_count<( ::__gnu_cxx::_Lock_policy)2> ::~__shared_count") from a __host__ __device__ function("") is not allowed

../external/Kokkos/core/src/Kokkos_View.hpp(979): warning: calling a __host__ function("std::__shared_count<( ::__gnu_cxx::_Lock_policy)2> ::~__shared_count") from a __host__ __device__ function("") is not allowed
```

It's not 100% clear to me _why_ this is resulting in a "call" to the `__shared_count` destructor (I suspect it must be optimized out in practice, even if it technically exists). But I will say there's something eye-brow raising about accessing a `std::shared_ptr<MeshBlock>` in a device kernel, since 1) `std::shared_ptr` is not really GPU compatible and 2) `MeshBlock` is largely filled with member variables that aren't accessible from within the device kernel anyway. So that was the thought behind `MeshBlockData` - it was like, well, it seems like people want to be able to refer to the `MeshBlock` inside the device kernel, so this gives people a way to do it. But I'm open to just getting rid of the member function altogether.

AndrewGaspar: Yeah, I tried to scour the ptx, and, as far as I can tell, no call to `std::__shared_count<( ::__gnu_cxx::_Lock_policy)2> ::__shared_count` or `std::__shared_count<( ::__gnu_cxx::_Lock_policy)2> ::~__shared_count` are ever made.

AndrewGaspar:
Add -Werror to cuda unit tests in CI.

src/bvals/bvals_refine.cpp
Yurlungur: This *is* physics related, but I'm leery of removing it. I think this is for advection of field loops.

src/mesh/meshblock.hpp
Yurlungur: I think my preferred implmentation here would be to put `par_for_inner` in the `meshblock.hpp` header, but remove it from the meshblock class.

@pgrete , @AndrewGaspar would that satisfy constraints?

src/bvals/bvals_refine.cpp
Yurlungur: That said, I'm willing to lose it. We have version control, we can always get it back if we need to. That's the power of git, after all.

src/outputs/parthenon_hdf5.cpp
Yurlungur: I think it should be `+=`, but yes, I like checking it at the end.

AndrewGaspar: I'm adding macros to `error_checking.hpp` and checking each function call... I don't like the accumulation idea because it means you continue to execute additional library functions after you've entered a presumably erroneous state.

.gitignore
```diff
@@ -34,3 +34,7 @@ compile_commands.jsonbin
 *.phdf
 *.phdf.xdmf
 *.dat
+
+# Files created by bad CMake hygiene
+a.out
+cmake_hdf5_test.o
```
Yurlungur: add newline at the end of line 40

src/mesh/domain.hpp
AndrewGaspar:
```suggestion
  void MakeZeroDimensional_(int const index) {
```

src/utils/error_checking.cpp
```diff
@@ -0,0 +1,46 @@
...
+#ifdef HDF5OUTPUT
+[[noreturn]] void parthenon::ErrorChecking::fail_throws_hdf5(herr_t err,
+                                                             char const *const expr,
+                                                             char const *const filename,
+                                                             int const linenumber) {
+  std::stringstream ss;
+  ss << "HDF5 failure: `" << expr << "`, Code: 0x" << std::hex << err;
+  fail_throws(ss, filename, linenumber);
+}
+#endif
```
Yurlungur, line 46: Really nice idea here.

src/outputs/parthenon_hdf5.hpp
```diff
@@ -50,8 +51,64 @@ using parthenon::Real;
...
+
+using H5A = H5Handle<&H5Aclose>;
+using H5D = H5Handle<&H5Dclose>;
+using H5F = H5Handle<&H5Fclose>;
+using H5G = H5Handle<&H5Gclose>;
+using H5P = H5Handle<&H5Pclose>;
+using H5T = H5Handle<&H5Tclose>;
+using H5S = H5Handle<&H5Sclose>;
+} // namespace parthenon
+
```
Yurlungur: Oh this is a neat idea.

src/outputs/parthenon_hdf5.hpp
```diff
@@ -62,11 +119,11 @@ using parthenon::Real;
         }                                                                                \
       }                                                                                  \
     }                                                                                    \
-  }
+  } while (false)
```
Yurlungur, line 127: wait, aren't these in use? Why are you disabling the loop?

Yurlungur (approved):
Looks good to me. I'd feel more comfortable if @nmsriram took a look at the restart changes you made, since he originally wrote it.

src/outputs/parthenon_hdf5.hpp
AndrewGaspar: This doesn't disable the loop - it's just good macro hygiene. See: https://stackoverflow.com/a/4674580

Yurlungur: Ah, I see. Didn't know about this. Thanks.

src/mesh/meshblock.hpp
AndrewGaspar: sounds good to me.

src/interface/swarm.cpp
AndrewGaspar: @brryan Want to comment on this?

brryan: Sorry I missed this @pgrete, thanks for the ping @AndrewGaspar. I don't think this will break anything unless we store particles in negative array indices in the future. It's just a little messy so I added the TODO to go back at some point and think through the logic again. If the `TODO(BRR)` is registering on some list of issues, I'm fine with just making the comment `// This is a magic number, needed for Defrag()` if that's OK.

src/outputs/io_wrapper.cpp
AndrewGaspar: Sure. I actually didn't realize you could mark parameters const in the definition, but not in the declaration, but that actually turns out to be a pretty reasonable thing to do. Google even recommends it! https://abseil.io/tips/109

src/interface/swarm.cpp
AndrewGaspar: The variable isn't being used.

brryan: Hm ok this may just be vestigial code then. Let's say it's fine to go ahead and delete this now, but I'll take a closer look at this in the near future to ensure the logic is sound (the unit test gives me some confidence).

src/utils/error_checking.cpp
```diff
@@ -0,0 +1,46 @@
+//========================================================================================
+// Athena++ astrophysical MHD code
+// Copyright(C) 2014 James M. Stone <jmstone@princeton.edu> and other code contributors
+// Licensed under the 3-clause BSD License, see LICENSE file for details
```
pgrete: This code was not taken from Athena++, wasn't it? So 
```suggestion
// Parthenon performance portable AMR framework
// Copyright(C) 2020 The Parthenon collaboration
// Licensed under the 3-clause BSD License, see LICENSE file for details
```
may be more appropriate, wouldn't it?

src/outputs/restart.hpp
```diff
@@ -207,7 +200,7 @@ class RestartReader {
     // Allocate array of correct size
     int rank = H5Sget_simple_extent_ndims(dataspace);
     std::vector<hsize_t> dims(rank);
-    status = H5Sget_simple_extent_dims(dataspace, dims.data(), NULL);
+    H5Sget_simple_extent_dims(dataspace, dims.data(), NULL);
```
pgrete:
```suggestion
    PARTHENON_HDF5_CHECK(H5Sget_simple_extent_dims(dataspace, dims.data(), NULL));
```

src/outputs/restart.hpp
```diff
@@ -128,7 +123,7 @@ class RestartReader {
     hid_t filespace = H5Dget_space(dataset);
     int rank = H5Sget_simple_extent_ndims(filespace);
     std::vector<hsize_t> dims(rank);
-    status = H5Sget_simple_extent_dims(filespace, dims.data(), NULL);
+    H5Sget_simple_extent_dims(filespace, dims.data(), NULL);
```
pgrete: Should this (and below) also get a check?
```suggestion
    PARTHENON_HDF5_CHECK((H5Sget_simple_extent_dims(filespace, dims.data(), NULL));
```

src/outputs/restart.hpp
```diff
@@ -84,12 +83,10 @@ class RestartReader {
...
       /** Define memory dataspace **/
       hid_t memspace = H5Screate_simple(5, count, NULL);
-      hsize_t offsetMem[5] = {0, 0, 0, 0, 0};
-      status = H5Sselect_hyperslab(dataspace, H5S_SELECT_SET, offset, NULL, count, NULL);
+      H5Sselect_hyperslab(dataspace, H5S_SELECT_SET, offset, NULL, count, NULL);
 
       // Read data from file
-      status =
-          H5Dread(dataset, H5T_NATIVE_DOUBLE, memspace, dataspace, H5P_DEFAULT, data);
+      H5Dread(dataset, H5T_NATIVE_DOUBLE, memspace, dataspace, H5P_DEFAULT, data);
```
pgrete:
```suggestion
      PARTHENON_HDF5_CHECK(H5Dread(dataset, H5T_NATIVE_DOUBLE, memspace, dataspace, H5P_DEFAULT, data));
```
Error check?

src/outputs/restart.hpp
```diff
@@ -84,12 +83,10 @@ class RestartReader {
 
       /** Define memory dataspace **/
       hid_t memspace = H5Screate_simple(5, count, NULL);
-      hsize_t offsetMem[5] = {0, 0, 0, 0, 0};
-      status = H5Sselect_hyperslab(dataspace, H5S_SELECT_SET, offset, NULL, count, NULL);
+      H5Sselect_hyperslab(dataspace, H5S_SELECT_SET, offset, NULL, count, NULL);
```
pgrete:
```suggestion
      PARTHENON_HDF5_CHECK(H5Sselect_hyperslab(dataspace, H5S_SELECT_SET, offset, NULL, count, NULL));
```

Error check?

src/outputs/restart.cpp
```diff
@@ -148,330 +145,304 @@ void RestartOutput::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm)
...
-  H5Sclose(global_DSpace);
-
-#ifdef MPI_PARALLEL
-  /* release the file access template */
-  ierr = H5Pclose(acc_file);
-  ierr = MPI_Info_free(&FILE_INFO_TEMPLATE);
-#endif
-
-  H5Pclose(property_list);
-  H5Fclose(file);
```
pgrete, line 479: Are these now automatically closed due to the different scoping?

src/outputs/restart.hpp
```diff
@@ -142,8 +137,8 @@ class RestartReader {
     hid_t memspace = H5Screate_simple(rank, dims.data(), NULL);
 
     // Read data from file
-    status = H5Dread(dataset, theHdfType, memspace, dataspace, H5P_DEFAULT,
-                     static_cast<void *>(data.data()));
+    H5Dread(dataset, theHdfType, memspace, dataspace, H5P_DEFAULT,
+            static_cast<void *>(data.data()));
```
pgrete:
```suggestion
    PARTHENON_HDF5_CHECK((H5Dread(dataset, theHdfType, memspace, dataspace, H5P_DEFAULT,
            static_cast<void *>(data.data())));
```

src/outputs/restart.hpp
```diff
@@ -219,7 +212,7 @@ class RestartReader {
     std::vector<T> data(isize);
 
     // Read data from file
-    status = H5Aread(attr, theHdfType, static_cast<void *>(data.data()));
+    H5Aread(attr, theHdfType, static_cast<void *>(data.data()));
```
pgrete:
```suggestion
    PARTHENON_HDF5_CHECK((H5Aread(attr, theHdfType, static_cast<void *>(data.data())));
```

src/outputs/parthenon_hdf5.cpp
```diff
@@ -225,232 +227,213 @@ void PHDF5Output::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm) {
...
-  H5Sclose(global_DSpace);
-
-#ifdef MPI_PARALLEL
-  /* release the file access template */
-  ierr = H5Pclose(acc_file);
-  ierr = MPI_Info_free(&FILE_INFO_TEMPLATE);
-#endif
-
-  H5Pclose(property_list);
-  H5Fclose(file);
```
pgrete, line 455: Same as below. No need to close here any more?

src/mesh/meshblock.hpp
```diff
@@ -191,17 +191,23 @@ class MeshBlock : public std::enable_shared_from_this<MeshBlock> {
   inline void par_for_outer(const std::string &name, const size_t &scratch_size_in_bytes,
                             const int &scratch_level, const int &kl, const int &ku,
                             const Function &function) {
-    parthenon::par_for_outer(DEFAULT_OUTER_LOOP_PATTERN, name, exec_space,
-                             scratch_size_in_bytes, scratch_level, kl, ku, function);
+    parthenon::par_for_outer(
+        DEFAULT_OUTER_LOOP_PATTERN, name, exec_space, scratch_size_in_bytes,
+        scratch_level, kl, ku, KOKKOS_LAMBDA(parthenon::team_mbr_t const &member, int k) {
+          return function(member, k);
+        });
```
pgrete: What's behind this change? I'm not sure I follow the need for the extra lambda.

src/mesh/meshblock.hpp
```diff
@@ -191,17 +191,23 @@ class MeshBlock : public std::enable_shared_from_this<MeshBlock> {
...
                             const int &jl, const int &ju, const Function &function) {
-    parthenon::par_for_outer(DEFAULT_OUTER_LOOP_PATTERN, name, exec_space,
-                             scratch_size_in_bytes, scratch_level, kl, ku, jl, ju,
-                             function);
+    parthenon::par_for_outer(
+        DEFAULT_OUTER_LOOP_PATTERN, name, exec_space, scratch_size_in_bytes,
+        scratch_level, kl, ku, jl, ju,
+        KOKKOS_LAMBDA(parthenon::team_mbr_t const &member, int k, int j) {
+          return function(member, k, j);
+        });
```
pgrete: As above: What's behind this change? I'm not sure I follow the need for the extra lambda.

pgrete (commented):
Great work! Looks like a small fix turned into a major effort, but I think it's definitely worth the work given the increased code quality.
I still have a couple of (minor) questions, but otherwise this looks good to go.

pgrete:
One more thing: Looks like `-Werror` hasn't made it yet to the `.gitlab-ci.yml` (and `/builds/pgrete/parthenon/src/outputs/parthenon_hdf5.cpp(329): warning: variable "gLocations" was declared but never referenced` showed in the latest CI).

src/outputs/restart.hpp
AndrewGaspar: Weird, missed this file.

pgrete:
FYI the non-mpi runs are currently failing: https://gitlab.com/pgrete/parthenon/-/jobs/857955521

src/outputs/restart.hpp
AndrewGaspar: Yeah, I missed these files somehow.

src/outputs/restart.cpp
AndrewGaspar: That's right - all HDF5 handles are now closed automatically.

AndrewGaspar: There's a one-off type that frees the MPI_Info, too.

src/outputs/parthenon_hdf5.cpp
AndrewGaspar: Yeah, see other thread.

src/mesh/meshblock.hpp
AndrewGaspar: ah, when I removed the `MeshBlockDevice` bit, I forgot to revert this change.

src/utils/error_checking.cpp
AndrewGaspar: Yes, though I don't think error_checking.hpp was either...

AndrewGaspar:
OMG this merge.

AndrewGaspar:
nvm, not so bad.

JoshuaSBrown:
As a final comment can you run this through Codacy and see if anything pops up.

JoshuaSBrown:
Retry DeepCode

JoshuaSBrown:
"retry deepcode"

pgrete (approved):

PR #358, opened 2020-11-06
More Steps Towards using MeshData and MeshBlockData Interchangeably

### LINKED ISSUE #349: Unify common function for blocks and mesh
Currently, there's functionally duplicated code, e.g., calculating the FluxDivergence, depending on whether the kernel loops only over Variable(Packs) or over MeshBlockPacks.
Ideally, there'd only be a single interface and a single kernel implementation.

jdolence:
## PR Summary

Taking some steps to provide more flexibility for downstream codes by providing Parthenon functions that operate seamlessly on `MeshData` or `MeshBlockData` objects.

Addresses #349 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


example/advection/parthenon_app_inputs.cpp
```diff
@@ -209,12 +209,12 @@ Packages_t ProcessPackages(std::unique_ptr<ParameterInput> &pin) {
...
   packages[pkg->label()] = pkg;
-  return packages;
-}
 
-void SetFillDerivedFunctions() {
-  parthenon::FillDerivedVariables::SetFillDerivedFunctions(advection_package::PreFill,
-                                                           advection_package::PostFill);
+  auto &app = packages["Parthenon::AppInput"];
+  app->AddParam("PreFillDerivedBlock", advection_package::PreFill);
+  app->AddParam("PostFillDerivedBlock", advection_package::PostFill);
```
Yurlungur: Hmm. I *think* I like moving application input functions into the `Parthenon::AppInput`. It feels very flexible. But I would worry that we need to be somewhat careful here. It's less safe in the sense that it's harder to type check, and it's harder for the user/application developer to see what things are available to set.

We should try to carefully document what's available.

example/particles/particles.cpp
```diff
@@ -73,26 +73,20 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
   Metadata m({Metadata::Cell, Metadata::Independent});
   pkg->AddField(field_name, m);
 
-  pkg->EstimateTimestep = EstimateTimestep;
+  pkg->AddParam("EstimateTimestepBlock", EstimateTimestepBlock);
```
Yurlungur: Same thing here. I like this more than I like it for `ApplicationInput` since it fits nicely into the package.

src/interface/meshblock_data.hpp
```diff
@@ -64,6 +65,17 @@ class MeshBlockData {
...
+
+  IndexRange GetBoundsI(const IndexDomain &domain) {
+    return GetBlockPointer()->cellbounds.GetBoundsI(domain);
+  }
+  IndexRange GetBoundsJ(const IndexDomain &domain) {
+    return GetBlockPointer()->cellbounds.GetBoundsJ(domain);
+  }
+  IndexRange GetBoundsK(const IndexDomain &domain) {
+    return GetBlockPointer()->cellbounds.GetBoundsK(domain);
+  }
```
Yurlungur, line 80: This can be for later... but for my boundary machinery, it would be nice to be able to return coarse or fine bounds.

src/interface/state_descriptor.hpp
```diff
@@ -226,11 +211,23 @@ class StateDescriptor {
...
+ public:
+  Packages_t() {
+    std::string name("Parthenon::AppInput");
+    pkgs_[name] = std::make_shared<StateDescriptor>(name);
+  }
+  std::shared_ptr<StateDescriptor> &operator[](const std::string &label) {
+    return pkgs_[label];
+  }
+  auto begin() { return pkgs_.begin(); }
+  auto end() { return pkgs_.end(); }
```
Yurlungur: Is this enough to enable `for (auto & pkg : packages)`?

src/interface/state_descriptor.hpp
```diff
@@ -226,11 +211,23 @@ class StateDescriptor {
...
-  std::map<std::string, VarPackingFunc<Real>> realVarPackerMap_;
-  std::map<std::string, FluxPackingFunc<Real>> realFluxPackerMap_;
 };
 
-using Packages_t = std::map<std::string, std::shared_ptr<StateDescriptor>>;
+class Packages_t {
+ public:
+  Packages_t() {
+    std::string name("Parthenon::AppInput");
+    pkgs_[name] = std::make_shared<StateDescriptor>(name);
```
Yurlungur: :+1:

src/interface/update.cpp
```diff
@@ -48,36 +63,23 @@ TaskStatus FluxDivergenceBlock(std::shared_ptr<MeshBlockData<Real>> &in,
...
-              (coords.Area(X2DIR, k, j + 1, i) * vin.flux(X2DIR, l, k, j + 1, i) -
-               coords.Area(X2DIR, k, j, i) * vin.flux(X2DIR, l, k, j, i));
-        }
-        if (ndim == 3) {
-          dudt(l, k, j, i) +=
-              (coords.Area(X3DIR, k + 1, j, i) * vin.flux(X3DIR, l, k + 1, j, i) -
-               coords.Area(X3DIR, k, j, i) * vin.flux(X3DIR, l, k, j, i));
-        }
-        dudt(l, k, j, i) /= -coords.Volume(k, j, i);
+        dudt(l, k, j, i) = FluxDiv_(l, k, j, i, ndim, coords, vin);
```
Yurlungur, line 66: frankly, this is cleaner even if it wasn't necessary for flexibility :+1:

src/interface/update.hpp
```diff
@@ -13,44 +13,114 @@
...
+  auto out_pack = out->PackVariables(flags);
+  auto dudt_pack = dudt->PackVariables(flags);
+  parthenon::par_for(
+      DEFAULT_LOOP_PATTERN, "UpdateMeshData", DevExecSpace(), 0, in_pack.GetDim(5) - 1, 0,
+      in_pack.GetDim(4) - 1, 0, in_pack.GetDim(3) - 1, 0, in_pack.GetDim(2) - 1, 0,
+      in_pack.GetDim(1) - 1,
+      KOKKOS_LAMBDA(const int b, const int l, const int k, const int j, const int i) {
+        out_pack(b, l, k, j, i) = in_pack(b, l, k, j, i) + dt * dudt_pack(b, l, k, j, i);
+      });
+}
```
Yurlungur: You could put the implementations of these functions in the `update.cpp` file if you explicitly instantiated them for `MeshData` and `MeshBlockData` at the bottom of the file.

src/parthenon_manager.cpp
```diff
@@ -84,9 +84,6 @@ ParthenonStatus ParthenonManager::ParthenonInit(int argc, char *argv[]) {
   }
 
   // Allow for user overrides to default Parthenon functions
-  if (app_input->SetFillDerivedFunctions != nullptr) {
-    SetFillDerivedFunctions = app_input->SetFillDerivedFunctions;
-  }
   if (app_input->ProcessProperties != nullptr) {
     ProcessProperties = app_input->ProcessProperties;
   }
```
Yurlungur, line 89: should `app_input` entirely be replaced with `params`? Or is it worth keeeping? I think that may be a longer discussion.

example/advection/parthenon_app_inputs.cpp
jdolence: The type checking part, at least, is not so bad.  The `Params` class records the type of the thing being stashed and then checks it against the explicitly provided templated type of the `Get` call, throwing an error if they don't agree exactly.  I had this problem early on when using a `std::function` for the template parameter, which `Params` wouldn't allow.

As far as the other comments, I agree with them.  This is very flexible, and I like the fact that we're reusing the `Params` object in the `StateDescriptor` rather than having to include explicitly typed function pointers as member variables.  I agree good documentation is certainly a must.  The real disadvantage here in that regard is that you would never know there were these magic strings by reading the `StateDescriptor`/`Params` source.  Actually, perhaps that suggests an improvement -- there could be string literals defined in one of these headers that could be used everywhere of relevance, rather than just typing stuff out like I've done now.  Then there is at least _some_ measure of self-documentation in the source.  Thoughts?

src/interface/state_descriptor.hpp
jdolence: yes

src/interface/update.cpp
jdolence: Need to see if there's any performance implications of this.  I wonder if this `FluxDiv_` function shouldn't be `FORCEINLINE`?  I worry about vectorization, but maybe needlessly.

example/advection/parthenon_app_inputs.cpp
jdolence: Replaced all the special but undocumented strings with constexpr char arrays.  I think I like this, including that the source is now more self documenting.

src/interface/state_descriptor.hpp
```diff
@@ -26,6 +26,38 @@
 
 namespace parthenon {
 
+namespace magic {
```
Yurlungur: I don't love this namespace name. Maybe `SpecialParams` or `Reserved`?

example/advection/advection_driver.cpp
```diff
@@ -56,8 +58,8 @@ TaskStatus UpdateMeshData(const int stage, Integrator *integrator,
                           std::shared_ptr<parthenon::MeshData<Real>> &out) {
   const Real beta = integrator->beta[stage - 1];
   const Real dt = integrator->dt;
-  parthenon::Update::AverageMeshData(in, base, beta);
-  parthenon::Update::UpdateMeshData(in, dudt, beta * dt, out);
+  parthenon::Update::AverageIndependentData(in, base, beta);
+  parthenon::Update::UpdateIndependentData(in, dudt, beta * dt, out);
```
pgrete, line 62: How about we keep the `AverageData` and `UpdateData`  and pass in the flags used in the packing as arguments?
That'd make the function more flexible.

src/bvals/bvals_refine.cpp
```diff
@@ -207,6 +207,7 @@ void BoundaryValues::RestrictGhostCellsOnSameLevel(const NeighborBlock &nb, int
   std::shared_ptr<MeshBlock> pmb = GetBlockPointer();
   MeshRefinement *pmr = pmb->pmr.get();
 
+  int ndim = pmb->pmy_mesh->ndim;
```
pgrete: Is this used somewhere?
```suggestion
```

src/bvals/bvals_refine.cpp
```diff
@@ -299,6 +300,9 @@ void BoundaryValues::ProlongateGhostCells(const NeighborBlock &nb, int si, int e
   std::shared_ptr<MeshBlock> pmb = GetBlockPointer();
   auto &pmr = pmb->pmr;
 
+  int ndim = pmb->pmy_mesh->ndim;
```
pgrete: Same as above. Is this used?
```suggestion
```

src/bvals/bvals_refine.cpp
```diff
@@ -364,7 +368,7 @@ void BoundaryValues::ProlongateGhostCells(const NeighborBlock &nb, int si, int e
   // calculate the loop limits for the finer grid
   int fsi, fei, fsj, fej, fsk, fek;
 
-  const IndexDomain interior = IndexDomain::interior;
+  // const IndexDomain interior = IndexDomain::interior;
```
pgrete:
```suggestion
```

src/interface/mesh_data.hpp
```diff
@@ -76,12 +77,23 @@ class MeshData {
   MeshData() = default;
 
   Mesh *GetMeshPointer() const { return pmy_mesh_; }
+  auto GetGridPointer() { return GetMeshPointer(); }
```
pgrete: How about calling this `GetDataPointer()` (as it effectively either points to `MeshData` or `MeshBlockData`).
We never use "Grid" anywhere so far so this may be confusing/inconsistent.

Also adding a comment in code here why redefine do what looks like redefining a function back to back would be great.

src/interface/state_descriptor.hpp
pgrete: As much as I love code magic I agree that something more descriptive would be great ;)

src/interface/state_descriptor.hpp
```diff
@@ -26,6 +26,38 @@
...
+// some "magic" strings
+constexpr char app_input[] = "A Parthenon package, provided by default in a Packages_t "
```
pgrete: I'm not sure how much I like strings as identifiers and string comparison further down to identify which function to call.
Maybe a tag dispatch idiom together with `enable_if`s would be cleaner.
I also wonder what @AndrewGaspar thinks about this.

src/interface/update.cpp
pgrete: Old: `|-> 1.20e-01 sec 11.8% 0.0% 0.0% 100.0% 0.00e+00 20 FluxDivergenceMesh [region]` (using #365 )
New: `|-> 2.21e+00 sec 71.5% 0.0% 0.0% 100.0% 0.00e+00 20 FluxDivergenceMesh [region]`
Suggested: `|-> 1.24e-01 sec 12.4% 0.0% 0.0% 100.0% 0.00e+00 20 FluxDivergenceMesh [region]`
on a single AVX512 core for a single 128^3 MeshBlock using Intel compilers.
`FORCEINLINE` also did not help but passing by reference helps ;)

src/interface/update.cpp
```diff
@@ -31,14 +29,31 @@ namespace parthenon {
 
 namespace Update {
 
-TaskStatus FluxDivergenceBlock(std::shared_ptr<MeshBlockData<Real>> &in,
-                               std::shared_ptr<MeshBlockData<Real>> &dudt_cont) {
+KOKKOS_INLINE_FUNCTION
+Real FluxDiv_(const int l, const int k, const int j, const int i, const int ndim,
+              const Coordinates_t &coords, VariableFluxPack<Real> v) {
```
pgrete:
```suggestion
Real FluxDiv_(const int &l, const int &k, const int &j, const int &i, const int &ndim,
              const Coordinates_t &coords, const VariableFluxPack<Real> &v) {
```

See below. Passing `v` by reference is crucial for performance.

src/mesh/mesh.hpp
```diff
@@ -149,9 +149,12 @@ class Mesh {
   int GetMaxLevel() const noexcept { return max_level; }
   int GetCurrentLevel() const noexcept { return current_level; }
   std::vector<int> GetNbList() const noexcept { return nblist; }
+  void SetAllowedDt(const Real dt) { max_allowed_dt_ = dt; }
```
pgrete: This need to be thread safe, doesn't it (because it could be set by multiple partitions at the same time)?

src/interface/variable_pack.hpp
```diff
@@ -80,9 +80,13 @@ class VariablePack {
     return v_(n)(k, j, i);
   }
   KOKKOS_FORCEINLINE_FUNCTION
+  T &operator()(const int m, const int n, const int k, const int j, const int i) const {
+    return v_(n)(k, j, i);
```
pgrete, line 85: That's elegant!

NB: Given the success we have in using Views of Views I'm wondering whether we could remove the `ParArrayND`s and just go with `ParArray#D`s (and replicate the flexible indexing where required using Views of Views.
I do *not* think the performance would increase dramatically (due to fewer index calculation), but it could remove extra code that does not need to be maintained.
Having said that this may be sth for a longer discussion and definitely not in the scope of this PR.

src/interface/variable_pack.hpp
```diff
@@ -80,9 +80,13 @@ class VariablePack {
     return v_(n)(k, j, i);
   }
   KOKKOS_FORCEINLINE_FUNCTION
+  T &operator()(const int m, const int n, const int k, const int j, const int i) const {
```
pgrete, line 84:
```suggestion
  T &operator()(const int &m, const int &n, const int &k, const int &j, const int &i) const {
```

src/interface/update.hpp
pgrete: Also, in all these functions and interface here we could add more const references (even for the packs as they're handled somewhere else anyway).

src/parthenon_manager.cpp
```diff
@@ -152,7 +149,7 @@ ParthenonStatus ParthenonManager::ParthenonInit(int argc, char *argv[]) {
     }
   }
 
-  SetFillDerivedFunctions();
+  // SetFillDerivedFunctions();
```
pgrete:
```suggestion
```

src/parthenon_manager.cpp
```diff
@@ -171,7 +168,7 @@ ParthenonStatus ParthenonManager::ParthenonFinalize() {
 }
 
 void ParthenonManager::SetFillDerivedFunctionsDefault() {
-  FillDerivedVariables::SetFillDerivedFunctions(nullptr, nullptr);
+  // FillDerivedVariables::SetFillDerivedFunctions(nullptr, nullptr);
```
pgrete:
```suggestion
```

src/refinement/refinement.cpp
```diff
@@ -131,5 +136,20 @@ AmrTag FirstDerivative(MeshBlock *pmb, const ParArrayND<Real> &q,
...
 }
 
+TaskStatus Tag(std::shared_ptr<MeshBlockData<Real>> &rc) {
+  auto pmb = rc->GetBlockPointer();
+  pmb->pmr->SetRefinement(CheckAllRefinement(rc));
+  return TaskStatus::complete;
+}
+
+TaskStatus Tag(std::shared_ptr<MeshData<Real>> &rc) {
+  auto pm = rc->GetMeshPointer();
```
pgrete:
```suggestion
```
Doesn't seem to be used here.

pgrete (changes requested):
Thanks for giving this a shot!
Overall, we definitely need this sooner than later.
Given the more general changes (e.g., how to enlist and handle app functions) I suggest that more people have a look at this (not a detailed review but a conceptual one) and we discuss it on Wednesday in the call.

example/advection/parthenon_app_inputs.cpp
Yurlungur: I like the idea that the strings passed into `params` are themselves the documentation. That's a good idea.

src/interface/update.cpp
Yurlungur: So looks like if we pass by reference we're good here?

src/interface/variable_pack.hpp
Yurlungur: That may introduce its own new code that needs to be maintained.

pgrete: Yes, of course. But if the new code is only 10% of the old code and logically simpler and/or more flexible then this would be a change worth pursing.

Now that I think about it, what's the status down stream adoption/testing of MeshBlockPacks for real ParArrayND data?
I'm asking as my current impression is that the design of the MeshBlockPacks silently assumed that `(b[block], v[ariable] , k, j, i)` indexing is used.

Yurlungur: @jdolence has been working on an adapting to meshblock packs in a branch but it's not in our master branch yet. Yes, I think we assume that block,variable,k,j,i indexing everywhere.

Yurlungur: > Yes, of course. But if the new code is only 10% of the old code and logically simpler and/or more flexible then this would be a change worth pursing.

I'm not convinced that would be the case... We use `ParArrayND`s fairly extensively in our downstream code for, e.g., vector-valued variables.

src/interface/update.hpp
```diff
@@ -66,61 +66,31 @@ void AverageIndependentData(T &c1, T &c2, const Real wgt1) {
...
-
 template <typename T>
 TaskStatus EstimateTimestep(std::shared_ptr<T> &rc) {
-  using Tstep_t = TimeStepFuncType<T>;
   using Desc_t = std::shared_ptr<StateDescriptor>;
-  auto pm = rc->GetGridPointer();
   Real dt_min = std::numeric_limits<Real>::max();
-  std::string fname = (IsBlock(rc) ? magic::estimate_dt_block : magic::estimate_dt_mesh);
+  auto pm = rc->GetGridPointer();
   for (const std::pair<std::string, Desc_t> &pkg : pm->packages) {
```
Yurlungur: could this be `for(const auto & pkg : pm->packages)`? You might be able to remove the `Desc_t` using statement then too

src/interface/update.hpp
```diff
@@ -66,61 +66,31 @@ void AverageIndependentData(T &c1, T &c2, const Real wgt1) {
...
-  std::string pre_name =
-      (is_block ? magic::pre_fill_derived_block : magic::pre_fill_derived_mesh);
-  std::string post_name =
-      (is_block ? magic::post_fill_derived_block : magic::post_fill_derived_mesh);
-  std::string name = (is_block ? magic::fill_derived_block : magic::fill_derived_mesh);
-  auto &params = app_pkg->AllParams();
-  if (params.hasKey(pre_name)) {
-    params.Get<DeriveFunc_t *>(pre_name)(rc);
+  auto pm = rc->GetGridPointer();
+  for (const std::pair<std::string, Desc_t> &pkg : pm->packages) {
```
Yurlungur: Same as above. Can this be `auto`-ified?

src/interface/state_descriptor.hpp
```diff
@@ -223,10 +200,51 @@ class StateDescriptor {
...
+  }
+  Real EstimateTimestep(std::shared_ptr<MeshData<Real>> &rc) {
+    if (EstimateTimestepMesh != NULL) return EstimateTimestepMesh(rc);
+    return std::numeric_limits<Real>::max();
+  }
+
+  AmrTag CheckRefinement(std::shared_ptr<MeshBlockData<Real>> &rc) {
+    if (CheckRefinementBlock != NULL) return CheckRefinementBlock(rc);
+    return AmrTag::derefine;
+  }
```
Yurlungur: This is an elegant solution to the problem you were having before, and I like it a lot.

jdolence:
I'm sure there's junk I still need to clean up, but I'd like to get this merged quickly if possible.  I think I've addressed the main point (i.e. the use of Params to stash function pointers) we discussed on the call.  Anyone care to approve?  Or point out things I still need to mess with?

example/advection/advection_driver.cpp
jdolence: check out what I did to see if that fits the bill

src/interface/mesh_data.hpp
jdolence: it _does not_ point to the data.  it points to either `Mesh` or `MeshBlock`.  That's why I went with `Grid` because we hadn't gone there yet but has the closest meaning intuitively I think.

src/interface/variable_pack.hpp
jdolence: does this actually matter?  I mean I'm actually really curious.

src/interface/state_descriptor.hpp
```diff
@@ -214,10 +200,49 @@ class StateDescriptor {
...
```
JoshuaSBrown: Change NULL to nullptr.

jdolence: oops. got it.

src/mesh/mesh.hpp
```diff
@@ -149,9 +149,12 @@ class Mesh {
...
+  Real NewDt() { return max_allowed_dt_; }
```
JoshuaSBrown: Make method const noexcept.

docs/README.md
```diff
@@ -57,9 +57,8 @@ During a simulation, Parthenon calls a number of default internal functions whos
...
-* `SetFillDerivedFunctions` Each package can register a function pointer in the Packages_t object that provides a callback mechanism for derived quantities (e.g. velocity, from momentum and mass) to be filled.  Additionally, this function provides a mechanism to register functions to fill derived quantities before and/or after all the individual package calls are made.  This is particularly useful for derived quantities that are shared by multiple packages.
 * `ProcessProperties` Constructs and returns a Properties_t object that is often filled with runtime specified (i. e. determined from the input file) settings and parameters.  For example, this might hold an equation of state.
-* `ProcessPackages` Constructs and returns a Packages_t object that contains a listing of all the variables and their metadata associated with each package.
+* `ProcessPackages` Constructs and returns a `Packages_t` object that contains a listing of all the variables and their metadata associated with each package.  Every `Packages_t` object has a special package accessible via the name `parthenon::app::inputs` that allows applications to specify application-level parameters and function pointers.
```
Yurlungur: revert this documentation, yeah?

src/interface/state_descriptor.hpp
```diff
@@ -26,23 +26,18 @@
 
 namespace parthenon {
 
+namespace app {
+// some "magic" strings
+constexpr char inputs[] = "A Parthenon package, provided by default in a Packages_t "
+                          "object, in which special functions can be registered to "
+                          "allow downstream applications to specialize functionality";
+} // namespace app
```
Yurlungur: Should this still be here?

src/interface/state_descriptor.hpp
```diff
@@ -140,13 +141,6 @@ class StateDescriptor {
     return true;
   }
 
-  void AddMeshBlockPack(const std::string &pack_name, const VarPackingFunc<Real> &func) {
-    realVarPackerMap_[pack_name] = func;
-  }
-  void AddMeshBlockPack(const std::string &pack_name, const FluxPackingFunc<Real> &func) {
-    realFluxPackerMap_[pack_name] = func;
-  }
-
```
Yurlungur, line 149: is this branch based on Phil Grette's branch or something? Should we merge that in first? To make sure things are consistent? I thought we already removed this stuff.

src/interface/state_descriptor.hpp
```diff
@@ -226,11 +251,23 @@ class StateDescriptor {
...
+  }
+  std::shared_ptr<StateDescriptor> &operator[](const std::string &label) {
+    return pkgs_[label];
+  }
+  auto begin() { return pkgs_.begin(); }
+  auto end() { return pkgs_.end(); }
+
+ private:
+  std::map<std::string, std::shared_ptr<StateDescriptor>> pkgs_;
+};
```
Yurlungur: should this be reverted?

src/interface/update.hpp
```diff
@@ -13,44 +13,94 @@
...
-                         std::shared_ptr<MeshBlockData<Real>> &dudt_cont, const Real dt,
-                         std::shared_ptr<MeshBlockData<Real>> &out);
-void UpdateMeshData(std::shared_ptr<MeshData<Real>> &in,
-                    std::shared_ptr<MeshData<Real>> &dudt, const Real dt,
-                    std::shared_ptr<MeshData<Real>> &out);
-void AverageMeshData(std::shared_ptr<MeshData<Real>> &c1_pack,
-                     std::shared_ptr<MeshData<Real>> &c2_pack, const Real wgt1);
-Real EstimateTimestep(std::shared_ptr<MeshBlockData<Real>> &rc);
+template <typename T>
+TaskStatus FluxDivergence(std::shared_ptr<T> &in, std::shared_ptr<T> &dudt_obj);
```
Yurlungur, line 37: As mentioned before, I'd rather these go back into the implementation file with the specializations explicitly declared in the `cpp` file. But it's not a huge deal either way.

src/refinement/refinement.cpp
```diff
@@ -131,5 +130,19 @@ AmrTag FirstDerivative(MeshBlock *pmb, const ParArrayND<Real> &q,
...
+}
+
+TaskStatus Tag(std::shared_ptr<MeshData<Real>> &rc) {
+  for (int i = 0; i < rc->NumBlocks(); i++) {
+    auto &pbd = rc->GetBlockData(i);
+    auto status = Tag(pbd);
+  }
+  return TaskStatus::complete;
+}
+
```
Yurlungur, line 146: This is a nice solution. I'm not sure either is more performant, since we're still doing tagging per MeshBlock, but I don't know if there's a way around that.

Yurlungur (changes requested):
Make sure to merge `develop` in and run the full test suite. I notice the `Kokkos` submodule doesn't agree with develop, which is odd.

src/refinement/refinement.cpp
jdolence: yeah agreed, it won't be more performant.  The `MeshData` interface is just for convenience for applications that are working mostly/exclusively with those objects.

src/interface/update.hpp
jdolence: For these sorts of short templated functions, I actually prefer they stay where they are.  Putting them in the cpp file will just lead to more SLOC and a slightly less flexible implementation (e.g. what if we change `MeshBlockData` to `SomeNewShinyName`?  This would be just one more place we'd have to touch.).  I'm going to leave them where they are unless there's more push back.

src/interface/state_descriptor.hpp
jdolence: I just don't think we completely scrubbed it.

src/mesh/mesh.hpp
jdolence: I went ahead and changed these to be consistent with the functions above them.  In general, though, I'm not a huge fan of needlessly marking functions `noexcept`.  `const` is a little better since it's a bit more than semantics.

src/parthenon_manager.cpp
Yurlungur: I think this discussion has been had and the answer is no.

src/interface/update.hpp
Yurlungur: The reason I prefer splitting prototypes and implementations, even for short templated functions, is I think it's easier to read what functions are available and what they do.

That said, I don't care that much.

Yurlungur (approved):

src/mesh/mesh.hpp
jdolence: hmmm, actually I think you're right.

src/mesh/mesh.hpp
jdolence: ended up removing these for other reasons.

src/mesh/mesh.hpp
jdolence: OK, I think I've addressed this now.  Good catch @pgrete!  That would have been annoying to debug at some point.

jdolence:
Retry DeepCode

src/interface/variable_pack.hpp
pgrete: Looks like it doesn't matter that much (or is potentially is even hurtful, though given this small inline it probably again doesn't matter), see https://stackoverflow.com/questions/3009543/passing-integers-as-constant-references-versus-copying

src/bvals/bvals_refine.cpp
pgrete: Why did you resolve the conversation?
As far as I can see `ndim` is not used.
What are they needed for?
I'm removing them now. Feel free to reintroduce if I missed something.

src/interface/mesh_data.hpp
pgrete: I see. How about `GetParentPointer` or `GetOwningObjectPointer`?
I'm arguing for this because "Grid" has a very specific meaning in some communities so the double meaning here could become confusing.

src/interface/state_descriptor.hpp
```diff
@@ -214,10 +194,49 @@ class StateDescriptor {
...
-  AmrTag (*CheckRefinement)(std::shared_ptr<MeshBlockData<Real>> &rc);
+  void (*PreFillDerivedBlock)(std::shared_ptr<MeshBlockData<Real>> &rc);
+  void (*PreFillDerivedMesh)(std::shared_ptr<MeshData<Real>> &rc);
+  void (*PostFillDerivedBlock)(std::shared_ptr<MeshBlockData<Real>> &rc);
+  void (*PostFillDerivedMesh)(std::shared_ptr<MeshData<Real>> &rc);
+  void (*FillDerivedBlock)(std::shared_ptr<MeshBlockData<Real>> &rc);
+  void (*FillDerivedMesh)(std::shared_ptr<MeshData<Real>> &rc);
+  Real (*EstimateTimestepBlock)(std::shared_ptr<MeshBlockData<Real>> &rc);
+  Real (*EstimateTimestepMesh)(std::shared_ptr<MeshData<Real>> &rc);
+  AmrTag (*CheckRefinementBlock)(std::shared_ptr<MeshBlockData<Real>> &rc);
```
pgrete, line 239: Is there no `CheckRefinementMesh` on purpose or just because it's not implemented yet.
I'm happy to not take care of it in this PR but just want to know (not that there's a hidden roadblock or sth).

jdolence:
@pgrete good catch on the Kokkos version.  oops.

Just FYI in case you didn't notice, I ran the full suite on this last night and everything passed.

example/advection/advection_driver.cpp
pgrete: Yes, that looks good! :+1:

pgrete:
Yes, that's good news.
I'm currently finalizing the review (tracking another[/the same] performance regression on CPU in the FluxDivergence again...

jdolence:
> Yes, that's good news.
> I'm currently finalizing the review (tracking another[/the same] performance regression on CPU in the FluxDivergence again...

seriously???  i thought I had that covered.

I wonder if force inlining of the FluxDiv_ func might help?

pgrete:
Yes, it does. So my original statement was not 100% correct (though I was sure I tested that, too).
Give me two more secs.

pgrete:
Running performance test now.

src/interface/update.hpp
```diff
@@ -13,44 +13,94 @@
...
+  }
+  for (const auto &pkg : pm->packages) {
+    pkg.second->FillDerived(rc);
+  }
+  for (const auto &pkg : pm->packages) {
+    pkg.second->PostFillDerived(rc);
+  }
+  return TaskStatus::complete;
+}
+
```
pgrete: From a conceptual point of view, is my understanding correct that the current implementation means that a downstream app either enrolls either `MeshData` or `MeshBlockData` functions and determine which one to use by forwarding the appropriate container in the driver?
It'd be great if we'd have one more paragraph in the doc highlighting this "unify" abstraction and what this mean for enrolled functions.

pgrete (commented):
Looks good!
I left some final comments and would like to find an agreement on the Get*Grid*Pointer issue but otherwise this looks like ready to go (pending a review by the CS team?).
I can also confirm that the latest version with the inline function results in expected performance.

src/interface/state_descriptor.hpp
jdolence: The issue is that `CheckRefinement` really has to tag at the block level.  We could, in principle, have a `CheckRefinement` that takes in a `MeshData` object and does a reduction on an array of `AmrTag`s, but I just didn't write it.  The task that actually calls `CheckRefinement`, i.e. `Refinement::Tag`, comes in both forms, but it's just for convenience right now -- the `MeshData` form just has a loop in it over `MeshBlockData` objects, calling the per block form.

src/interface/update.hpp
jdolence: In fact, a downstream app can make use of _both_ forms, so long as the appropriate tasks are called.

I'll add a bit to the docs.

src/interface/mesh_data.hpp
jdolence: Hmm.  OK.  I'll go with `GetParentPointer`, but I wish the name gave some sense of what sort of "thing" the parent is.  Can't think of anything though.

src/bvals/bvals_refine.cpp
jdolence: Oh sorry.  I actually thought this got taken care of in the recent boundaries PR.

jdolence:
OK @pgrete, I think I've addressed your comments.  Mind taking a quick peek and marking approved if appropriate?  Thanks again for the review.

pgrete (approved):
Thanks for addressing the last bits. I'm happy to approve!

Issue #301, opened 2020-09-09
Scalability analysis 


gshipman:
A place to discuss scalability

gshipman:
Running on master with @pgrete input deck. We are seeing the following performance on RZAnsel: 

<img width="361" alt="image" src="https://user-images.githubusercontent.com/665019/92631077-1338f880-f28e-11ea-915c-dd25980c697c.png">

Run script: 

```
N=1
NX1=256
NX2=256
NX3=256
echo "N=${N}, NX1=${NX1}, NX2=${NX2}, NX3=${NX3}"
lrun -n ${N} -g 1 -c 10 -M "-gpu" ./example/advection/advection-example -i ./parthinput.advection.weak parthenon/mesh/nx1=${NX1} parthenon/mesh/nx2=${NX2} parthenon/mesh/nx3=${NX3} | tee ${N}.out

N=2
NX1=`expr $NX1 \\* 2 `
echo "N=${N}, NX1=${NX1}, NX2=${NX2}, NX3=${NX3}"
lrun -n ${N} -g 1 -c 10 -M "-gpu" ./example/advection/advection-example -i ./parthinput.advection.weak parthenon/mesh/nx1=${NX1} parthenon/mesh/nx2=${NX2} parthenon/mesh/nx3=${NX3} | tee ${N}.out

 N=4
 NX2=`expr $NX2 \\* 2 `
 echo "N=${N}, NX1=${NX1}, NX2=${NX2}, NX3=${NX3}"
lrun -n ${N} -g 1 -c 10 -M "-gpu" ./example/advection/advection-example -i ./parthinput.advection.weak parthenon/mesh/nx1=${NX1} parthenon/mesh/nx2=${NX2} parthenon/mesh/nx3=${NX3} | tee ${N}.out
```

Input deck: 
```
<parthenon/job>
problem_id = advection

<parthenon/mesh>
refinement = adaptive
numlevel = 2

nx1 = 256
x1min = -0.5
x1max = 0.5
ix1_bc = periodic
ox1_bc = periodic

nx2 = 256
x2min = -0.5
x2max = 0.5
ix2_bc = periodic
ox2_bc = periodic

nx3 = 256
x3min = -0.5
x3max = 0.5
ix3_bc = periodic
ox3_bc = periodic

<parthenon/meshblock>
nx1 = 32
nx2 = 32
nx3 = 32

<parthenon/time>
tlim = 1.0
integrator = rk1
nlim = 10
[...]

PR #359, opened 2020-11-11
Buffer-pack-in-one

pgrete:
## PR Summary

WIP to have a central place to keep track of performance numbers.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

# Open questions
- [ ] Are we still using `vbvar->var_cc` extensively or can than functionality go as it it effectively absorbed by `VariablePack`s`?


pgrete:
This is ready for a conceptual review.
The key idea is that all the information of what goes from where to where including indices is calculated on the host and then copied over to the device.
Only active buffers are used i.e., buffers allocated in the framework but not part of the simulation (e.g. coarse to fine buffer in a uniform grid calculation) are not handled.
The kernels themselves are not optimized yet (but there's room for -- potentially MeshBlock size dependent -- improvements).
In addition, the caching of the boundary information struct works in theory.
In practice, it works for the SetBoundaries function but doesn't for the SendBoundaryBuffer function (I cannot seem to find what's wrong.)
Another set of eyes would be very helpful. I'll add a pointer to the specific line in the following comment.

src/bvals/cc/bvals_cc_in_one.cpp
```diff
@@ -0,0 +1,551 @@
...
+      auto *bd_var_ = rc->GetCellVariableVector()[0]->vbvar->GetPBdVar();
+      if (bd_var_->sflag[nb.bufid] == parthenon::BoundaryStatus::completed) continue;
+      buffers_used += 1;
+    }
+  }
+  Kokkos::Profiling::popRegion(); // Reset boundaries
+
+  auto boundary_info = md->GetSendBuffers();
+
+  if (!boundary_info.is_allocated()) {
```
pgrete: Here, I test if the boundary info already exists from a previous call. The "cache" gets purged together with the MeshBlockPacks in the `ClearCache` call of the `MeshData`.
Changing this line to
`if (!boundary_info.is_allocated() || true) {`
i.e., always recreate the info results in a correct run.

pgrete:
A quick note on performance:
The current version increases the total performance on a static 256^3 grid  with 16^3 MeshBlocks on a single V100 by ~10x (i.e., the speedup for those two kernels is >10x).
The total overhead (in the ideal case) of going from one 256^3 MeshBlock to 16^3 MeshBlocks for a 256^3 Mesh is about 8.3x.
Also at 16^3 MeshBlocks the performance of a single V100 is now about that of 40 Skylake cores.

For AMR (which now works with this version here), I used the input deck of #301 but with 5 variables.

For 32^3 MeshBlocks (does not trigger loadbalancing; Number of MeshBlocks = 960; 448  created, 0 destroyed during this simulation.)
- V100 old: 4e7 zone-cycles/s
- V100 new:  1e8 zone-cycles/s
- 40 Skylake cores: 1e8 zone-cycles/s

For 16^3 MeshBlocks (Number of MeshBlocks = 6532; 2436  created, 0 destroyed during this simulation.)
- V100 old:  6.3e6 zone-cycles/s (for cycles without loadbalancing) and 3.1e6 (for cycles with load balancing)
- V100 new: 2.6e7 zone-cycles/s (for cycles without loadbalancing) and 4.5e6 (for cycles with load balancing)
- 40 Skylake cores: 6.5e7 zone-cycles/s (for cycles without loadbalancing) and 3.8e7 (for cycles with load balancing)

So there's still (obviously) room for improvement :D 

First indication on what to address next can be derived from here.
This is a profiling output from the AMR test above but with 128^3 root Mesh  with 32^3 MeshBlocks:

On a GPU:
```
|-> 4.46e+01 sec 67.4% 3.5% 0.0% 0.0% 2.29e+03 1 Driver_Main [region]
|   |-> 4.09e+01 sec 61.7% 1.1% 0.0% 0.0% 9.45e+02 20 LoadBalancingAndAdaptiveMeshRefinement [region]
|   |   |-> 4.09e+01 sec 61.7% 1.1% 0.0% 0.0% 9.45e+02 1 RedistributeAndRefineMeshBlocks [region]
|   |   |   |-> 4.04e+01 sec 61.0% 0.6% 0.0% 99.4% 4.28e+02 1 Step 7: Construct new MeshBlockList [region]
|   |   |   |-> 4.60e-01 sec 0.7% 46.0% 0.0% 0.7% 4.63e+04 1 Mesh::Initialize [region]
|   |   |   |   |-> 2.79e-01 sec 0.4% 38.0% 0.0% 62.0% 3.61e+04 204 Task_SendBoundaryBuffers_MeshBlockData [region]
|   |   |   |   |   |-> 5.59e-02 sec 0.1% 100.0% 0.0% ------ 4728 PackData 4D [for]
|   |   |   |   |   |-> 7.75e-03 sec 0.0% 100.0% 0.0% ------ 596 RestrictCellCenteredValues3d [for]
|   |   |   |   |-> 8.13e-02 sec 0.1% 35.6% 0.0% 0.8% 4.29e+04 205 Task_FillDerived [region]
|   |   |   |   |   |-> 2.73e-02 sec 0.0% 35.7% 0.0% 64.3% 4.26e+04 205 PostFillDerived [region]
|   |   |   |   |   |-> 2.71e-02 sec 0.0% 35.8% 0.0% 64.2% 4.30e+04 205 PreFillDerived [region]
|   |   |   |   |   |-> 2.63e-02 sec 0.0% 36.3% 0.0% 63.7% 4.43e+04 205 FillDerived [region]
|   |   |   |   |-> 5.57e-02 sec 0.1% 80.7% 0.0% 19.3% 8.48e+04 204 Task_ReceiveAndSetBoundariesWithWait [region]
|   |   |   |   |   |-> 4.50e-02 sec 0.1% 100.0% 0.0% ------ 4728 UnpackData 4D [for]
|   |   |   |   |-> 4.01e-02 sec 0.1% 78.8% 0.0% 1.8% 7.54e+04 204 Task_ProlongateBoundaries [region]
|   |   |   |   |   |-> 2.98e-02 sec 0.0% 80.2% 0.0% 19.8% 8.16e+04 204 RestrictBoundaries [region]
|   |   |   |   |   |   |-> 2.39e-02 sec 0.0% 100.0% 0.0% ------ 2430 RestrictCellCenteredValues3d [for]
|   |   |   |   |   |-> 9.41e-03 sec 0.0% 81.8% 0.0% 18.2% 6.33e+04 204 ProlongateBoundaries [region]
|   |   |   |   |   |   |-> 7.70e-03 sec 0.0% 100.0% 0.0% ------ 596 ProlongateCellCenteredValues3d [for]
|   |-> 3.78e+00 sec 5.7% 29.0% 0.0% 48.3% 1.68e+04 20 MultiStage_Step [region]
|   |   |-> 6.38e-01 sec 1.0% 5.9% 0.0% 94.1% 1.96e+03 40 Task_WeightedSumData [region]
|   |   |   |-> 2.78e-02 sec 0.0% 100.0% 0.0% ------ 40 WeightedSumData [for]
|   |   |-> 3.39e-01 sec 0.5% 77.5% 0.0% 2.5% 7.97e+04 2400 Task_ProlongateBoundaries [region]
|   |   |   |-> 2.43e-01 sec 0.4% 79.5% 0.0% 20.5% 8.68e+04 2400 RestrictBoundaries [region]
|   |   |   |   |-> 1.94e-01 sec 0.3% 100.0% 0.0% ------ 21120 RestrictCellCenteredValues3d [for]
|   |   |   |-> 8.52e-02 sec 0.1% 81.2% 0.0% 18.8% 6.95e+04 2400 ProlongateBoundaries [region]
|   |   |   |   |-> 6.92e-02 sec 0.1% 100.0% 0.0% ------ 5920 ProlongateCellCenteredValues3d [for]
|   |   |-> 2.39e-01 sec 0.4% 96.3% 0.0% 3.7% 1.00e+04 2400 Task_Tag_Block [region]
|   |   |   |-> 2.30e-01 sec 0.3% 100.0% 0.0% ------ 2400 advection check refinement [reduce]
|   |   |-> 1.96e-01 sec 0.3% 90.7% 0.0% 9.3% 3.67e+04 2400 Task_Advection_CalculateFluxes [region]
|   |   |   |-> 6.21e-02 sec 0.1% 100.0% 0.0% ------ 2400 x2 flux [for]
|   |   |   |-> 6.05e-02 sec 0.1% 100.0% 0.0% ------ 2400 x3 flux [for]
|   |   |   |-> 5.54e-02 sec 0.1% 100.0% 0.0% ------ 2400 x1 flux [for]
|   |   |-> 1.55e-01 sec 0.2% 70.5% 0.0% 5.1% 4.65e+04 2400 Task_FillDerived [region]
|   |   |   |-> 5.20e-02 sec 0.1% 72.5% 0.0% 27.5% 4.61e+04 2400 PostFillDerived [region]
|   |   |   |   |-> 3.77e-02 sec 0.1% 100.0% 0.0% ------ 2400 advection_package::PostFill [for]
|   |   |   |-> 4.77e-02 sec 0.1% 75.4% 0.0% 24.6% 5.03e+04 2400 PreFillDerived [region]
|   |   |   |   |-> 3.60e-02 sec 0.1% 100.0% 0.0% ------ 2400 advection_package::PreFill [for]
|   |   |   |-> 4.72e-02 sec 0.1% 75.2% 0.0% 24.8% 5.09e+04 2400 FillDerived [region]
|   |   |       |-> 3.55e-02 sec 0.1% 100.0% 0.0% ------ 2400 advection_package::SquareIt [for]
|   |   |-> 1.12e-01 sec 0.2% 76.5% 0.0% 0.2% 5.37e+04 20 Task_SendBoundaryBuffers_MeshData [region]
|   |   |   |-> 8.20e-02 sec 0.1% 70.5% 0.0% 29.5% 7.32e+04 20 Create bndinfo array [region]
|   |   |   |   |-> 5.52e-02 sec 0.1% 100.0% 0.0% ------ 5920 RestrictCellCenteredValues3d [for]
|   |   |   |-> 2.64e-02 sec 0.0% 100.0% 0.0% ------ 20 SendBoundaryBuffers [for]
|   |   |-> 8.33e-02 sec 0.1% 35.1% 0.0% 64.9% 4.61e+04 2400 Task_SendFluxCorrection [region]
|   |   |   |-> 6.89e-03 sec 0.0% 100.0% 0.0% ------ 640 SendFluxCorrection3D_x1 [for]
|   |   |-> 5.82e-02 sec 0.1% 85.2% 0.0% 14.8% 4.12e+04 2400 Task_EstimateTimestep [region]
|   |   |   |-> 4.96e-02 sec 0.1% 100.0% 0.0% ------ 2400 advection_package::EstimateTimestep [reduce]
|   |   |-> 4.84e-02 sec 0.1% 95.7% 0.0% 0.4% 4.96e+02 20 Task_SetBoundaries_MeshData [region]
|   |   |   |-> 4.61e-02 sec 0.1% 100.0% 0.0% ------ 20 SetBoundaries [for]
|   |   |-> 2.32e-02 sec 0.0% 74.4% 0.0% 25.6% 8.29e+04 2640 Task_ReceiveFluxCorrection [region]
|   |   |   |-> 1.16e-02 sec 0.0% 100.0% 0.0% ------ 1280 ReceiveFluxCorrection_x1 [for]
|   |   |-> 1.86e-02 sec 0.0% 100.0% 0.0% ------ 20 FluxDivergenceMesh [for]
```

On a CPU
```
|-> 9.65e-01 sec 83.9% 29.6% 0.0% 0.4% 9.66e+02 1 Driver_Main [region]
|   |-> 7.85e-01 sec 68.3% 27.5% 1.0% 1.2% 6.46e+02 20 MultiStage_Step [region]
|   |   |-> 2.17e-01 sec 18.8% 0.0% 3.2% 0.1% 0.00e+00 60 Task_FillDerived [region]
|   |   |   |-> 9.43e-02 sec 8.2% 0.0% 4.2% 1.5% 0.00e+00 60 PostFillDerived [region]
|   |   |   |   |-> 9.29e-02 sec 8.1% 0.0% 4.2% 100.0% 0.00e+00 60 advection_package::PostFill [region]
|   |   |   |-> 6.21e-02 sec 5.4% 0.0% 8.4% 1.6% 0.00e+00 60 FillDerived [region]
|   |   |   |   |-> 6.12e-02 sec 5.3% 0.0% 8.1% 100.0% 0.00e+00 60 advection_package::SquareIt [region]
|   |   |   |-> 6.02e-02 sec 5.2% 0.0% 9.2% 1.8% 0.00e+00 60 PreFillDerived [region]
|   |   |       |-> 5.91e-02 sec 5.1% 0.0% 9.1% 100.0% 0.00e+00 60 advection_package::PreFill [region]
|   |   |-> 1.65e-01 sec 14.3% 99.3% 3.7% 0.7% 1.09e+03 60 Task_Advection_CalculateFluxes [region]
|   |   |   |-> 5.56e-02 sec 4.8% 100.0% 5.7% ------ 60 x1 flux [for]
|   |   |   |-> 5.55e-02 sec 4.8% 100.0% 6.2% ------ 60 x2 flux [for]
|   |   |   |-> 5.27e-02 sec 4.6% 100.0% 4.4% ------ 60 x3 flux [for]
|   |   |-> 1.06e-01 sec 9.2% 0.0% 5.9% 0.8% 2.07e+02 40 Task_WeightedSumData [region]
|   |   |   |-> 1.06e-01 sec 9.2% 0.0% 5.9% 100.0% 0.00e+00 40 WeightedSumData [region]
|   |   |-> 9.95e-02 sec 8.6% 0.0% 5.6% 100.0% 0.00e+00 20 FluxDivergenceMesh [region]
|   |   |-> 5.54e-02 sec 4.8% 2.1% 8.8% 38.3% 2.17e+03 60 Task_SendBoundaryBuffers_MeshBlockData [region]
|   |   |   |-> 3.31e-02 sec 2.9% 0.0% 22.4% 100.0% 0.00e+00 1600 PackData 4D [region]
|   |   |   |-> 4.02e-04 sec 0.0% 100.0% 796.6% ------ 20 "recv buf 0"="send buf 4" [copy]
|   |   |   |-> 3.12e-04 sec 0.0% 100.0% 146.7% ------ 20 "recv buf 4"="send buf 0" [copy]
|   |   |   |-> 1.55e-04 sec 0.0% 100.0% 199.0% ------ 20 "recv buf 8"="send buf 12" [copy]
|   |   |   |-> 1.52e-04 sec 0.0% 100.0% 181.3% ------ 20 "recv buf 12"="send buf 8" [copy]
|   |   |-> 3.22e-02 sec 2.8% 0.0% 21.2% 3.8% 0.00e+00 60 Task_SetBoundaries_MeshBlockData [region]
|   |   |   |-> 3.10e-02 sec 2.7% 0.0% 21.2% 100.0% 0.00e+00 1600 UnpackData 4D [region]
|   |   |-> 2.98e-02 sec 2.6% 98.6% 20.3% 1.4% 2.01e+03 60 Task_Tag_Block [region]
|   |   |   |-> 2.94e-02 sec 2.6% 100.0% 19.9% ------ 60 advection check refinement [reduce]
|   |   |-> 2.78e-02 sec 2.4% 0.0% 43.4% 100.0% 0.00e+00 9934 Task_ReceiveBoundaryBuffers_MeshBlockData [region]
|   |   |-> 1.82e-02 sec 1.6% 96.8% 15.2% 3.2% 3.30e+03 60 Task_EstimateTimestep [region]
|   |   |   |-> 1.76e-02 sec 1.5% 100.0% 16.6% ------ 60 advection_package::EstimateTimestep [reduce]
|   |   |-> 1.11e-02 sec 1.0% 0.0% 136.4% 1.4% 0.00e+00 60 Task_ProlongateBoundaries [region]
|   |   |   |-> 8.92e-03 sec 0.8% 0.0% 140.6% 100.0% 0.00e+00 60 ProlongateBoundaries [region]
|   |   |   |-> 2.03e-03 sec 0.2% 0.0% 147.1% 100.0% 0.00e+00 60 RestrictBoundaries [region]
|   |   |-> 4.02e-03 sec 0.3% 100.0% 6.5% ------ 3 Kokkos::View::initialization [advected] [for]
|   |   |-> 3.95e-03 sec 0.3% 0.6% 133.3% 100.0% 0.00e+00 60 Task_SendFluxCorrection [region]
|   |   |-> 3.70e-03 sec 0.3% 0.0% 270.6% 100.0% 0.00e+00 60 Task_ReceiveFluxCorrection [region]
|   |   |-> 1.08e-03 sec 0.1% 0.0% 57.4% 100.0% 0.00e+00 60 Task_StartReceiving [region]
|   |   |-> 6.88e-04 sec 0.1% 0.0% 24.3% 100.0% 0.00e+00 60 Task_ClearBoundary [region]
|   |-> 1.74e-01 sec 15.2% 38.8% 9.5% 0.0% 2.41e+03 20 LoadBalancingAndAdaptiveMeshRefinement [region]
|   |   |-> 1.52e-01 sec 13.2% 44.5% 1.8% 0.2% 2.77e+03 1 RedistributeAndRefineMeshBlocks [region]
|   |   |   |-> 1.04e-01 sec 9.0% 60.7% 15.1% 39.3% 3.42e+03 1 Step 7: Construct new MeshBlockList [region]
|   |   |   |   |-> 6.26e-03 sec 0.5% 100.0% 21.1% ------ 2 Kokkos::View::initialization [one_minus_sqrt_one_minus_advected_sq_37] [for]
|   |   |   |   |-> 6.20e-03 sec 0.5% 100.0% 22.3% ------ 2 Kokkos::View::initialization [one_minus_sqrt_one_minus_advected_sq_12] [for]
|   |   |   |   |-> 6.16e-03 sec 0.5% 100.0% 23.3% ------ 2 Kokkos::View::initialization [advected] [for]
|   |   |   |   |-> 6.14e-03 sec 0.5% 100.0% 24.9% ------ 2 Kokkos::View::initialization [one_minus_advected_sq] [for]
|   |   |   |   |-> 6.13e-03 sec 0.5% 100.0% 27.7% ------ 2 Kokkos::View::initialization [advected.fluxX1] [for]
|   |   |   |   |-> 6.12e-03 sec 0.5% 100.0% 28.8% ------ 2 Kokkos::View::initialization [advected.fluxX2] [for]
|   |   |   |   |-> 6.10e-03 sec 0.5% 100.0% 30.3% ------ 2 Kokkos::View::initialization [advected.fluxX3] [for]
|   |   |   |   |-> 6.09e-03 sec 0.5% 100.0% 26.9% ------ 2 Kokkos::View::initialization [one_minus_advected] [for]
|   |   |   |   |-> 9.88e-04 sec 0.1% 100.0% 40.5% ------ 2 Kokkos::View::initialization [advected.coarse] [for]
|   |   |   |   |-> 2.62e-04 sec 0.0% 100.0% 38.6% ------ 4 Kokkos::View::initialization [send buf 11] [for]
|   |   |   |   |-> 2.61e-04 sec 0.0% 100.0% 33.1% ------ 4 Kokkos::View::initialization [send buf 8] [for]
|   |   |   |   |-> 2.60e-04 sec 0.0% 100.0% 40.2% ------ 4 Kokkos::View::initialization [recv buf 19] [for]
|   |   |   |   |-> 2.60e-04 sec 0.0% 100.0% 36.7% ------ 4 Kokkos::View::initialization [recv buf 13] [for]
|   |   |   |   |-> 2.60e-04 sec 0.0% 100.0% 38.1% ------ 4 Kokkos::View::initialization [recv buf 23] [for]
|   |   |   |   |-> 2.60e-04 sec 0.0% 100.0% 33.4% ------ 4 Kokkos::View::initialization [recv buf 7] [for]
|   |   |   |   |-> 2.60e-04 sec 0.0% 100.0% 43.6% ------ 4 Kokkos::View::initialization [send buf 19] [for]
|   |   |   |   |-> 2.60e-04 sec 0.0% 100.0% 33.4% ------ 4 Kokkos::View::initialization [recv buf 10] [for]
|   |   |   |   |-> 2.60e-04 sec 0.0% 100.0% 39.0% ------ 4 Kokkos::View::initialization [recv buf 12] [for]
|   |   |   |   |-> 2.60e-04 sec 0.0% 100.0% 31.8% ------ 4 Kokkos::View::initialization [send buf 9] [for]
|   |   |   |   |-> 2.60e-04 sec 0.0% 100.0% 34.0% ------ 4 Kokkos::View::initialization [send buf 5] [for]
|   |   |   |   |-> 2.60e-04 sec 0.0% 100.0% 36.0% ------ 4 Kokkos::View::initialization [send buf 21] [for]
|   |   |   |   |-> 2.59e-04 sec 0.0% 100.0% 36.8% ------ 4 Kokkos::View::initialization [send buf 22] [for]
|   |   |   |   |-> 2.59e-04 sec 0.0% 100.0% 38.0% ------ 4 Kokkos::View::initialization [send buf 16] [for]
|   |   |   |   |-> 2.59e-04 sec 0.0% 100.0% 39.5% ------ 4 Kokkos::View::initialization [recv buf 9] [for]
|   |   |   |   |-> 2.59e-04 sec 0.0% 100.0% 39.0% ------ 4 Kokkos::View::initialization [recv buf 16] [for]
|   |   |   |   |-> 2.59e-04 sec 0.0% 100.0% 32.9% ------ 4 Kokkos::View::initialization [send buf 15] [for]
|   |   |   |   |-> 2.59e-04 sec 0.0% 100.0% 43.5% ------ 4 Kokkos::View::initialization [send buf 18] [for]
|   |   |   |   |-> 2.59e-04 sec 0.0% 100.0% 38.2% ------ 4 Kokkos::View::initialization [send buf 4] [for]
|   |   |   |   |-> 2.58e-04 sec 0.0% 100.0% 54.6% ------ 4 Kokkos::View::initialization [recv buf 11] [for]
|   |   |   |   |-> 2.58e-04 sec 0.0% 100.0% 37.9% ------ 4 Kokkos::View::initialization [recv buf 17] [for]
|   |   |   |   |-> 2.58e-04 sec 0.0% 100.0% 35.5% ------ 4 Kokkos::View::initialization [send buf 12] [for]
|   |   |   |   |-> 2.58e-04 sec 0.0% 100.0% 41.9% ------ 4 Kokkos::View::initialization [recv buf 22] [for]
|   |   |   |   |-> 2.58e-04 sec 0.0% 100.0% 41.4% ------ 4 Kokkos::View::initialization [recv buf 2] [for]
|   |   |   |   |-> 2.58e-04 sec 0.0% 100.0% 38.2% ------ 4 Kokkos::View::initialization [recv buf 20] [for]
|   |   |   |   |-> 2.58e-04 sec 0.0% 100.0% 36.2% ------ 4 Kokkos::View::initialization [send buf 14] [for]
|   |   |   |   |-> 2.58e-04 sec 0.0% 100.0% 38.9% ------ 4 Kokkos::View::initialization [recv buf 0] [for]
|   |   |   |   |-> 2.58e-04 sec 0.0% 100.0% 43.7% ------ 4 Kokkos::View::initialization [recv buf 21] [for]
|   |   |   |   |-> 2.58e-04 sec 0.0% 100.0% 40.4% ------ 4 Kokkos::View::initialization [send buf 2] [for]
|   |   |   |   |-> 2.58e-04 sec 0.0% 100.0% 63.5% ------ 4 Kokkos::View::initialization [send buf 7] [for]
|   |   |   |   |-> 2.58e-04 sec 0.0% 100.0% 17.5% ------ 4 Kokkos::View::initialization [send buf 6] [for]
|   |   |   |   |-> 2.58e-04 sec 0.0% 100.0% 36.8% ------ 4 Kokkos::View::initialization [send buf 1] [for]
|   |   |   |   |-> 2.58e-04 sec 0.0% 100.0% 16.9% ------ 4 Kokkos::View::initialization [recv buf 14] [for]
|   |   |   |   |-> 2.57e-04 sec 0.0% 100.0% 40.7% ------ 4 Kokkos::View::initialization [recv buf 6] [for]
|   |   |   |   |-> 2.57e-04 sec 0.0% 100.0% 37.4% ------ 4 Kokkos::View::initialization [recv buf 15] [for]
|   |   |   |   |-> 2.57e-04 sec 0.0% 100.0% 33.0% ------ 4 Kokkos::View::initialization [recv buf 3] [for]
|   |   |   |   |-> 2.57e-04 sec 0.0% 100.0% 39.6% ------ 4 Kokkos::View::initialization [recv buf 5] [for]
|   |   |   |   |-> 2.56e-04 sec 0.0% 100.0% 41.3% ------ 4 Kokkos::View::initialization [recv buf 8] [for]
|   |   |   |   |-> 2.56e-04 sec 0.0% 100.0% 36.6% ------ 4 Kokkos::View::initialization [recv buf 18] [for]
|   |   |   |   |-> 2.56e-04 sec 0.0% 100.0% 16.8% ------ 4 Kokkos::View::initialization [send buf 13] [for]
|   |   |   |   |-> 2.56e-04 sec 0.0% 100.0% 18.1% ------ 4 Kokkos::View::initialization [send buf 10] [for]
|   |   |   |   |-> 2.56e-04 sec 0.0% 100.0% 24.3% ------ 4 Kokkos::View::initialization [recv buf 1] [for]
|   |   |   |   |-> 2.56e-04 sec 0.0% 100.0% 22.0% ------ 4 Kokkos::View::initialization [recv buf 4] [for]
|   |   |   |   |-> 2.56e-04 sec 0.0% 100.0% 19.7% ------ 4 Kokkos::View::initialization [send buf 23] [for]
|   |   |   |   |-> 2.55e-04 sec 0.0% 100.0% 19.6% ------ 4 Kokkos::View::initialization [send buf 20] [for]
|   |   |   |   |-> 2.54e-04 sec 0.0% 100.0% 36.1% ------ 4 Kokkos::View::initialization [send buf 17] [for]
|   |   |   |   |-> 2.53e-04 sec 0.0% 100.0% 19.3% ------ 4 Kokkos::View::initialization [send buf 3] [for]
|   |   |   |   |-> 2.50e-04 sec 0.0% 100.0% 19.3% ------ 4 Kokkos::View::initialization [send buf 0] [for]
|   |   |   |-> 3.56e-02 sec 3.1% 0.4% 172.2% 0.6% 1.57e+03 1 Mesh::Initialize [region]
|   |   |   |   |-> 1.83e-02 sec 1.6% 0.1% 8.0% 0.1% 1.97e+03 6 Task_FillDerived [region]
|   |   |   |   |   |-> 7.82e-03 sec 0.7% 0.1% 8.4% 4.2% 1.54e+03 6 PostFillDerived [region]
|   |   |   |   |   |   |-> 7.48e-03 sec 0.7% 0.0% 10.1% 100.0% 0.00e+00 5 advection_package::PostFill [region]
|   |   |   |   |   |-> 5.28e-03 sec 0.5% 0.2% 8.5% 5.3% 2.27e+03 6 PreFillDerived [region]
|   |   |   |   |   |   |-> 4.99e-03 sec 0.4% 0.0% 10.7% 100.0% 0.00e+00 5 advection_package::PreFill [region]
|   |   |   |   |   |-> 5.20e-03 sec 0.5% 0.2% 7.1% 4.9% 2.31e+03 6 FillDerived [region]
|   |   |   |   |       |-> 4.93e-03 sec 0.4% 0.0% 9.3% 100.0% 0.00e+00 5 advection_package::SquareIt [region]
|   |   |   |   |-> 1.28e-02 sec 1.1% 0.0% 458.5% 80.9% 0.00e+00 5 Task_ReceiveAndSetBoundariesWithWait [region]
|   |   |   |   |   |-> 2.44e-03 sec 0.2% 0.0% 47.2% 100.0% 0.00e+00 149 UnpackData 4D [region]
|   |   |   |   |-> 2.88e-03 sec 0.3% 4.5% 35.0% 41.4% 6.95e+03 5 Task_SendBoundaryBuffers_MeshBlockData [region]
|   |   |   |   |   |-> 1.58e-03 sec 0.1% 0.0% 36.8% 100.0% 0.00e+00 149 PackData 4D [region]
|   |   |   |   |-> 1.39e-03 sec 0.1% 0.0% 71.5% 1.8% 0.00e+00 5 Task_ProlongateBoundaries [region]
|   |   |   |   |   |-> 1.06e-03 sec 0.1% 0.0% 78.4% 100.0% 0.00e+00 5 ProlongateBoundaries [region]
|   |   |   |   |   |-> 3.06e-04 sec 0.0% 0.0% 66.2% 100.0% 0.00e+00 5 RestrictBoundaries [region]
|   |   |   |-> 6.16e-03 sec 0.5% 38.6% 65.4% 100.0% 0.00e+00 1 Step 6: Alloc, pack, and send buffers [region]
|   |   |   |-> 3.38e-03 sec 0.3% 0.0% 1200.4% 75.2% 5.92e+02 1 Step 8: Recv data and unpack [region]
|   |   |   |   |-> 8.36e-04 sec 0.1% 0.0% 127.2% 100.0% 0.00e+00 2 UnpackData 4D [region]
|   |   |   |-> 2.71e-03 sec 0.2% 83.2% 75.3% 68.5% 7.39e+02 1 Step 5: Alloc buffer and start recv [region]
|   |   |   |   |-> 4.29e-04 sec 0.0% 100.0% 103.6% ------ 1 Kokkos::View::initialization [recvbuf1] [for]
|   |   |   |   |-> 4.24e-04 sec 0.0% 100.0% 81.8% ------ 1 Kokkos::View::initialization [recvbuf0] [for]
|   |   |-> 2.26e-02 sec 2.0% 0.0% 72.3% 100.0% 0.00e+00 20 UpdateMeshBlockTree [region]
|   |-> 1.49e-03 sec 0.1% 98.9% 17.1% 1.1% 3.35e+03 5 Task_EstimateTimestep [region]
|   |   |-> 1.47e-03 sec 0.1% 100.0% 17.7% ------ 5 advection_package::EstimateTimestep [reduce]

```

 So `|   |   |   |-> 4.04e+01 sec 61.0% 0.6% 0.0% 99.4% 4.28e+02 1 Step 7: Construct new MeshBlockList [region]` is excessively expensive on a GPU.

Similarly, `Task_ProlongateBoundaries` uses about 9% of the time for a `Step()` on GPUs whereas it's only 1.5% on CPUs, which we already knew would remain expensive #376 ;)

example/advection/advection_driver.cpp
```diff
@@ -118,27 +119,68 @@ TaskCollection AdvectionDriver::MakeTaskCollection(BlockList_t &blocks, const in
...
+      blocks[0]->packages["advection_package"]->Param<bool>("buffer_send_pack");
+  if (buffer_send_pack) {
+    TaskRegion &tr = tc.AddRegion(num_partitions);
+    for (int i = 0; i < num_partitions; i++) {
+      auto &mc1 = pmesh->mesh_data.GetOrAdd(stage_name[stage], i);
+      tr[i].AddTask(none, parthenon::cell_centered_bvars::SendBoundaryBuffers, mc1);
+    }
+  } else {
+    TaskRegion &tr = tc.AddRegion(num_task_lists_executed_independently);
+    for (int i = 0; i < blocks.size(); i++) {
```
Yurlungur, line 134: I think you should probably go from `i = 0; i < num_task_lists_executed_independently` for consistency here, though of course it's the same as `blocks.size()`.

example/advection/advection_package.cpp
```diff
@@ -57,6 +57,13 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
   }
   pkg->AddParam<>("profile", profile_str);
 
+  auto buffer_send_pack = pin->GetOrAddBoolean("Advection", "buffer_send_pack", false);
+  auto buffer_recv_pack = pin->GetOrAddBoolean("Advection", "buffer_recv_pack", false);
+  auto buffer_set_pack = pin->GetOrAddBoolean("Advection", "buffer_set_pack", false);
+  pkg->AddParam<>("buffer_send_pack", buffer_send_pack);
+  pkg->AddParam<>("buffer_recv_pack", buffer_recv_pack);
+  pkg->AddParam<>("buffer_set_pack", buffer_set_pack);
```
Yurlungur, line 65: Presumably this option will go away? When we know we want pack in one? Or do we plan to keep both versions indefinitely?

src/bvals/cc/bvals_cc.cpp
```diff
@@ -132,12 +133,13 @@ int CellCenteredBoundaryVariable::LoadBoundaryBufferSameLevel(ParArray1D<Real> &
...
-  sj = (nb.ni.ox2 > 0) ? (cellbounds.je(interior) - NGHOST + 1) : cellbounds.js(interior);
-  ej = (nb.ni.ox2 < 0) ? (cellbounds.js(interior) + NGHOST - 1) : cellbounds.je(interior);
-  sk = (nb.ni.ox3 > 0) ? (cellbounds.ke(interior) - NGHOST + 1) : cellbounds.ks(interior);
-  ek = (nb.ni.ox3 < 0) ? (cellbounds.ks(interior) + NGHOST - 1) : cellbounds.ke(interior);
+
+  cell_centered_bvars::CalcIndicesLoadSame(nb.ni.ox1, si, ei,
+                                           cellbounds.GetBoundsI(interior));
+  cell_centered_bvars::CalcIndicesLoadSame(nb.ni.ox2, sj, ej,
+                                           cellbounds.GetBoundsJ(interior));
+  cell_centered_bvars::CalcIndicesLoadSame(nb.ni.ox3, sk, ek,
```
Yurlungur, line 141: Can you expand a bit on the naming convention here? Load vs. set, etc.?

src/bvals/cc/bvals_cc_in_one.cpp
```diff
@@ -0,0 +1,553 @@
...
+namespace parthenon {
+
+namespace cell_centered_bvars {
+
+//----------------------------------------------------------------------------------------
+//! \fn void cell_centered_bvars::CalcIndicesSetSame(int ox, int &s, int &e,
+//                                                   const IndexRange &bounds)
+//  \brief Calculate indices for SetBoundary routines for buffers on the same level
+
+void CalcIndicesSetSame(int ox, int &s, int &e, const IndexRange &bounds) {
```
Yurlungur, line 40: In MR #332 , @JoshuaSBrown suggested changing  `int &s` and `int &e` to `IndexRange &indices` and I kind of like this. It simplifies the API for all these functions in various places.

src/bvals/cc/bvals_cc_in_one.cpp
```diff
@@ -0,0 +1,553 @@
...
+    for (auto &v : rc->GetCellVariableVector()) {
+      if (v->IsSet(parthenon::Metadata::FillGhost)) {
+        v->resetBoundary();
+      }
+    }
+
+    for (int n = 0; n < pmb->pbval->nneighbor; n++) {
+      parthenon::NeighborBlock &nb = pmb->pbval->neighbor[n];
+      // TODO(?) currently this only works for a single "Variable" per container.
+      // Need to update the buffer sizes so that it matches the packed Variables.
```
Yurlungur: Is this not a serious issue? In the LANL code we have a very large state vector.

src/bvals/cc/bvals_cc_in_one.cpp
```diff
@@ -0,0 +1,553 @@
...
+      auto *bd_var_ = rc->GetCellVariableVector()[0]->vbvar->GetPBdVar();
+      if (bd_var_->sflag[nb.bufid] == parthenon::BoundaryStatus::completed) continue;
+      buffers_used += 1;
+    }
+  }
+  Kokkos::Profiling::popRegion(); // Reset boundaries
+
+  auto boundary_info = md->GetSendBuffers();
+
+  if (!boundary_info.is_allocated() || true) {
```
Yurlungur: Is the intent this always true? That seems odd.

src/bvals/cc/bvals_cc_in_one.cpp
```diff
@@ -0,0 +1,553 @@
...
+    for (auto block = 0; block < md->NumBlocks(); block++) {
+      auto &rc = md->GetBlockData(block);
+      auto pmb = rc->GetBlockPointer();
+
+      int mylevel = pmb->loc.level;
+      for (int n = 0; n < pmb->pbval->nneighbor; n++) {
+        parthenon::NeighborBlock &nb = pmb->pbval->neighbor[n];
+        // TODO(?) currently this only works for a single "Variable" per container.
+        // Need to update the buffer sizes so that it matches the packed Variables.
+        assert(rc->GetCellVariableVector().size() == 1);
```
Yurlungur: maybe replace these asserts with `PARTHENON_REQUIRE` so we can hunt them down if we hit them at runtime.

src/bvals/cc/bvals_cc_in_one.cpp
```diff
@@ -0,0 +1,553 @@
...
+
+//----------------------------------------------------------------------------------------
+//! \fn void cell_centered_bvars::CalcIndicesLoadToFiner(int &si, int &ei, int &sj,
+//                                                       int &ej, int &sk, int &ek,
+//                                                       const NeighborBlock &nb,
+//                                                       MeshBlock *pmb)
+//  \brief Calculate indices for LoadBoundary routines for buffers to finer levels
+
+void CalcIndicesLoadToFiner(int &si, int &ei, int &sj, int &ej, int &sk, int &ek,
+                            const NeighborBlock &nb, MeshBlock *pmb) {
```
Yurlungur, line 185: These are pretty much always the same fore each combination of coarse/fine/relative orientation. Maybe we should pre-compute them once and cache them.

src/bvals/cc/bvals_cc_in_one.cpp
```diff
@@ -0,0 +1,553 @@
...
+          const IndexShape &c_cellbounds = pmb->c_cellbounds;
+          // "Same" logic is the same for loading to a coarse buffer, just using
+          // c_cellbounds
+          CalcIndicesLoadSame(nb.ni.ox1, si, ei, c_cellbounds.GetBoundsI(interior));
+          CalcIndicesLoadSame(nb.ni.ox2, sj, ej, c_cellbounds.GetBoundsJ(interior));
+          CalcIndicesLoadSame(nb.ni.ox3, sk, ek, c_cellbounds.GetBoundsK(interior));
+
+          auto &coarse_buf = rc->GetCellVariableVector()[0]->vbvar->coarse_buf;
+          pmb->pmr->RestrictCellCenteredValues(var_cc, coarse_buf, 0, Nv - 1, si, ei, sj,
+                                               ej, sk, ek);
```
Yurlungur: If I recall, this is done variable-by-variable so this function needs to be re-written to use a pack.

src/bvals/cc/bvals_cc_in_one.cpp
```diff
@@ -0,0 +1,553 @@
...
+        sk += pmb->block_size.nx3 / 2 - pmb->cnghost;
+      else
+        ek -= pmb->block_size.nx3 / 2 - pmb->cnghost;
+    }
+  }
+}
+
+// send boundary buffers with MeshBlockPack support
+// TODO(pgrete) should probaly be moved to the bvals or interface folders
+TaskStatus SendBoundaryBuffers(std::shared_ptr<MeshData<Real>> &md) {
```
Yurlungur: This function is long and kind of hard to follow. It might be nice to break some pieces into sub-functions. For example the kokkos kernels have a lot of boiler plate.

src/bvals/cc/bvals_cc_in_one.cpp
```diff
@@ -0,0 +1,553 @@
...
+  for (int block = 0; block < md->NumBlocks(); block++) {
+    auto &rc = md->GetBlockData(block);
+    auto pmb = rc->GetBlockPointer();
+    for (int n = 0; n < pmb->pbval->nneighbor; n++) {
+      buffers_used += 1;
+    }
+  }
+
+  auto boundary_info = md->GetSetBuffers();
+  if (!boundary_info.is_allocated()) {
```
Yurlungur: The "set boundary info" parts of this functionality, for both this function and the one above, appear to me to be different functionality from the rest of this task. Could they be split out? That would make this much easier to follow I think.

src/bvals/cc/bvals_cc_in_one.hpp
```diff
@@ -0,0 +1,61 @@
...
+  int ei = 0;
+  int sj = 0;
+  int ej = 0;
+  int sk = 0;
+  int ek = 0;
+  parthenon::ParArray1D<Real> buf; // comm buffer
+  parthenon::ParArray4D<Real> var; // data variable (could also be coarse array)
+};
+
+using BufferCache_t = ParArray1D<BndInfo>;
```
Yurlungur: Overall I like this design here. hOwever, the com and var buffers probably need to be changed. Perhaps they should be the shape of a `VariablePack` or something similar?

Yurlungur (commented):
Overall this definitely seems like a step in the right direction. I confess I struggled to read through this just by looking at the diff or to find bugs by staring at it. So I'm going to have to try running it and step through it with a debugger  or something. Below are my first broad comments, which are mostly stylistic. I'll have more comments once I dig in more.

example/advection/advection_driver.cpp
pgrete: I think that loop needs to go over all blocks, as otherwise some blocks may not get linked to a task.
In principle, the number of task lists and the number of block are completely independent from each other.
In our current example, we use `blocks.size()` independent task lists, but we could also have fewer.
In the case that there are fewer task lists, we'd still need to assign either all blocks  (to fewer task lists) or only assign a specific subset of blocks to to a task.
Here, all blocks should execute that task, so we need to loop over all blocks.

example/advection/advection_package.cpp
pgrete: Ideally, yes. In practice, that may first happen at a later point as we still use some packing tasks in the loadbalancing step. So right now, we cannot get rid of both `MeshBlockData` and `MeshData` based functionality.

src/bvals/cc/bvals_cc.cpp
pgrete: Are you referring to adding more documentation to the code or in the comment here?
For the former, the functions already contain brief docstrings. Is more info desired?
For the latter, I followed the existing naming convection, i.e., load = from var to buffer, and set = from buffer to var.
It also took me some time to remember this, so I'm open for a broader renaming of that capability if we have a quorum for that (it will touch quite a bit of machinery).

src/bvals/cc/bvals_cc_in_one.cpp
pgrete: Good point, I'll update that.

src/bvals/cc/bvals_cc_in_one.cpp
pgrete: True. I think that's a nice to have for a future PR. Given that the computed indices are already cached and that this operation in generally is not that expensive, I don't think it's a performance critical part right now.

src/bvals/cc/bvals_cc_in_one.cpp
pgrete: Yes, this could be split into three dependent tasks.
I kept it in one (similar to the load balance function) as the subtasks are highly dependent (i.e., there'd be a linear dependency between those where no other task should interrupt), but I'm also happy to split it.

src/bvals/cc/bvals_cc_in_one.cpp
pgrete: That depends ;)
Yes, it's serious because it breaks the the functionality we offer to have >1 variable.
No, because as far as I understood the LANL code currently also only uses a single large vector to hold all variables (there's not limit on the number of variables that state vector holds - it's just that currently there can only be one state vector).

Anyway, I think this is an easy fix and I only didn't address it yet as part of the overall POC design (and because I don't have a test case yet).
Do you have an internal test case ready that you could immediately test when I make the necessary changes?

src/bvals/cc/bvals_cc_in_one.cpp
pgrete: That's the "bug" I was referring to and where I'd like additional eyes on.
With the `true` statement that caching of the `boundary_info` is effectively disabled (because the `boundary_info` array is recalculated for every function call).
Remove the `true` make use of the caching, but results in errors in the regression tests that I was not able to track down yet.
I use the same caching for the `boundary_info` in the set routine and there it works as expected.

src/bvals/cc/bvals_cc_in_one.cpp
pgrete: I plan to have this functionality implemented before we merge into develop, so this will go away eventually.

src/bvals/cc/bvals_cc_in_one.cpp
pgrete: Correct, though I think we should first do some profiling to determine how much of a performance impact this is.
Given that these are way fewer calls (only a subset of the meshblocks will call this on a subset of the blocks) we may get away with relying on executing the small pack kernels in parallel on the device with a negligible impact of the launch overhead.

src/bvals/cc/bvals_cc_in_one.cpp
pgrete: They are kind of related, but could be separated, see my longer reply above.

src/bvals/cc/bvals_cc_in_one.hpp
pgrete: Hmm, why is that?
Our current design guarantees (/requires) that the comm buffers are 1D and the Variable buffers are 4D.
If we'd change the design to go from 56 independent buffer to one large buffer (with subviews or manual indexing), then this may change.

pgrete:
> Overall this definitely seems like a step in the right direction. I confess I struggled to read through this just by looking at the diff or to find bugs by staring at it. So I'm going to have to try running it and step through it with a debugger or something. Below are my first broad comments, which are mostly stylistic. I'll have more comments once I dig in more.

Thanks for the first comments.
Unfortunately, there isn't really a "diff" to look at as all the implementation is in the `bvals_cc_in_one.*pp` files, which contains merged and updated pieces from various places of the existing code base.
I was trying to reuse some code, e.g., in the calc indices helper functions.

example/advection/advection_driver.cpp
Yurlungur: > Here, all blocks should execute that task, so we need to loop over all blocks.

Right that's why I said it's the same in this case.

> I think that loop needs to go over all blocks, as otherwise some blocks may not get linked to a task.

Since you're calling `AddTask` the loop should go over all tasks, which is tied to all partitions/meshdata right? Or am I missing something?

example/advection/advection_package.cpp
Yurlungur: Right, but presumably it would be done on a per-task basis? Not a per run basis?

src/bvals/cc/bvals_cc.cpp
Yurlungur: Ah, I actually meant here just so I understood. ;)

These names make sense and aren't so important. I think it's fine to keep them. But recording that in the developers guide might be useful.

src/bvals/cc/bvals_cc_in_one.cpp
Yurlungur: > No, because as far as I understood the LANL code currently also only uses a single large vector to hold all variables (there's not limit on the number of variables that state vector holds - it's just that currently there can only be one state vector).

Maybe I'm misunderstanding... The LANL code registers a new variable for each quantity in the state vector. e.g., density and momentum are two separate variables. Is this still supported?

> Anyway, I think this is an easy fix and I only didn't address it yet as part of the overall POC design (and because I don't have a test case yet).

Yeah I see the code is written with this in mind.

> Do you have an internal test case ready that you could immediately test when I make the necessary changes?

Yes. Riot is pretty close to up-to-date with Parthenon now and we have several test problems running on GPUs including classics like Rayleigh-Taylor and Kelvin-Helmholtz. We could also easily add an extra independent variable to the advection example to test this functionality.

src/bvals/cc/bvals_cc_in_one.cpp
Yurlungur: got it. I'll poke at it.

src/bvals/cc/bvals_cc_in_one.cpp
Yurlungur: Sure, true enough. Precomputing them once might be less bug prone than checking for allocation, however.

src/bvals/cc/bvals_cc_in_one.cpp
Yurlungur: we could also keep it a single task, but split out "subroutines" in the file just to give an easier picture of flow of the function.

src/bvals/cc/bvals_cc_in_one.hpp
Yurlungur: When you go from 1 variable to multiple, don't you need one buffer per variable? Or do they get packed into the single contiguous buffer?

Yurlungur:
> > Overall this definitely seems like a step in the right direction. I confess I struggled to read through this just by looking at the diff or to find bugs by staring at it. So I'm going to have to try running it and step through it with a debugger or something. Below are my first broad comments, which are mostly stylistic. I'll have more comments once I dig in more.
> 
> Thanks for the first comments.
> Unfortunately, there isn't really a "diff" to look at as all the implementation is in the `bvals_cc_in_one.*pp` files, which contains merged and updated pieces from various places of the existing code base.
> I was trying to reuse some code, e.g., in the calc indices helper functions.

Yeah I don't think it's your fault. It's just hard to go bug hunting with the eyeball norm. :)

example/advection/advection_driver.cpp
pgrete: This all depends on the actual use case.
In the current use case "a collection of tasks should be executed on all blocks" I think it's important to keep the loop as is (to emphasize that this happens for all blocks).
Here are two (theoretical) example where I could imagine this would be different
a) we only want N task lists because we have N host threads that execute the task loop but the task is still executed for all blocks
then the loop may look like 
```c++
TaskRegion &tr = tc.AddRegion(N);
for (int i = 0; i < blocks.size(); i++) {
  tr[i%N].AddTask(none, &MeshBlockData<Real>::SendBoundaryBuffers, sc1.get());
```

or b) we only want to execute task on a subset of blocks that contain certain properties (but all tasks are independent and could be executed concurrently
then the loop may look like
```c++
const int M = getNumberOfBlocksWith(PropertyX);
TaskRegion &tr = tc.AddRegion(M);
for (int i = 0; i < blocks.size(); i++) {
  if (block.has(PropertyX) {
    tr[i].AddTask(none, &MeshBlockData<Real>::SendBoundaryBuffers, sc1.get());
  }
```

In both cases would loop over all blocks (or in the second case just have a loop over the blocks with that property.
Does this make sense? Maybe I'm also missing sth.

example/advection/advection_package.cpp
pgrete: Yes, I was referring to the existence of the implementation not those variables (which are up to the downstream app and not tied to Parthenon itself)

src/bvals/cc/bvals_cc_in_one.cpp
pgrete: > > No, because as far as I understood the LANL code currently also only uses a single large vector to hold all variables (there's not limit on the number of variables that state vector holds - it's just that currently there can only be one state vector).
> 
> Maybe I'm misunderstanding... The LANL code registers a new variable for each quantity in the state vector. e.g., density and momentum are two separate variables. Is this still supported?

Oh, then I misremembered. I though this was the original design and since then has been replaced by a large state vector from which the vars are extracted as needed.

> > Anyway, I think this is an easy fix and I only didn't address it yet as part of the overall POC design (and because I don't have a test case yet).
> 
> Yeah I see the code is written with this in mind.
> 
> > Do you have an internal test case ready that you could immediately test when I make the necessary changes?
> 
> Yes. Riot is pretty close to up-to-date with Parthenon now and we have several test problems running on GPUs including classics like Rayleigh-Taylor and Kelvin-Helmholtz. We could also easily add an extra independent variable to the advection example to test this functionality.

Great. I'll get that working next then.

src/bvals/cc/bvals_cc_in_one.cpp
pgrete: There a subtle difference here.
The "checking for allocation" has to be done independent of  the caching here, because after each refinement, derefinement, or loadbalancing step the cached `boundary_info` is not valid any more[*] whereas these indices are fixed for a given combination neighbor configuration.
For example, right face has a coarse neighbor and there are 2 ghost cells would always yield identical indices for this "configuration", but whether the right face has a coarse neighbor may change between steps.

[*] Actually, this may be where the current caching fails. Let me double check if the buffers are only invalidated for load balancing or also for refinement/derefinement (without load balancing).
In the latter case, we'd need to clear the cache and then this should work.

src/bvals/cc/bvals_cc_in_one.hpp
pgrete: Correct, our current implementation uses one buffer per variable (vector), so there's room for improvement.
This is also related to introducing support for >1 variables. Ideally, we'd pack all variables in a single buffer and handle with a single call, but that may turn out to warrant a major restructuring (as the "packing" would now need to be implemented natively in the boundary functions, so we could not handle them independently anymore).

src/bvals/cc/bvals_cc_in_one.cpp
pgrete: > [*] Actually, this may be where the current caching fails. Let me double check if the buffers are only invalidated for load balancing or also for refinement/derefinement (without load balancing).
> In the latter case, we'd need to clear the cache and then this should work.

As far as I can tell that's not it. Clearing the caches is always called.

example/advection/advection_driver.cpp
Yurlungur: Hmm.... I hadn't thought of case (b). In case (a), I would argue the better way to design that would be to partition the grid into N `MeshData` objects, and add the the tasks per-meshdata.

In both cases, what you prescribe works if you don't have any dependencies. If have you have dependencies, then you would need to keep track of dependencies *per meshblock* as opposed to *per call to `AddTask`*, which would get cumbersome quickly. For that reason, even case (b), you might want to restructure the tasks over `MeshData` collections instead of over `MeshBlockData`s.

example/advection/advection_package.cpp
Yurlungur: Right. But do we want to keep these variables in `advection` forever?

src/bvals/cc/bvals_cc_in_one.cpp
Yurlungur: > For example, right face has a coarse neighbor and there are 2 ghost cells would always yield identical indices for this "configuration", but whether the right face has a coarse neighbor may change between steps.

Ah, I see. Yes, this is true.

example/advection/advection_driver.cpp
pgrete: > Hmm.... I hadn't thought of case (b). In case (a), I would argue the better way to design that would be to partition the grid into N `MeshData` objects, and add the the tasks per-meshdata.

I agree that there are better approaches, but this an example on how the task infrastructure could be used.
For this reason (demonstration purpose), there's a more verbose comment in the code further up
```c++

  // Number of task lists that can be executed indepenently and thus *may*
  // be executed in parallel and asynchronous.
  // Being extra verbose here in this example to highlight that this is not
  // required to be 1 or blocks.size() but could also only apply to a subset of blocks.
  auto num_task_lists_executed_independently = blocks.size();
  TaskRegion &async_region1 = tc.AddRegion(num_task_lists_executed_independently);

  for (int i = 0; i < blocks.size(); i++) {
```

> In both cases, what you prescribe works if you don't have any dependencies. If have you have dependencies, then you would need to keep track of dependencies _per meshblock_ as opposed to _per call to `AddTask`_, which would get cumbersome quickly. For that reason, even case (b), you might want to restructure the tasks over `MeshData` collections instead of over `MeshBlockData`s.

These would end up in different `TaskRegions` that we guarantee to be executed sequentially. This is also way the current example implemented here with the two types of interfaces (MeshData and MeshBlockData) in different `TaskRegions` works (as it implicitly guarantees that dependency between regions).

example/advection/advection_package.cpp
pgrete: No, I think they can go before merging to develop.

example/advection/advection_driver.cpp
Yurlungur: > These would end up in different TaskRegions that we guarantee to be executed sequentially. 

No not necessarily. Not if you're doing things like
```C++
tl[i%N].AddTask()
```

That's why I'm emphasizing you should always do the loop over tasks, not over blocks. In the case of example (b) this could be fixed as
```C++
const int M = getNumberOfBlocksWith(PropertyX);
TaskRegion &tr = tc.AddRegion(M);
int j = 0;
for (int i = 0; i < blocks.size(); i++) {
  if (block.has(PropertyX) {
    tr[j++].AddTask(none, &MeshBlockData<Real>::SendBoundaryBuffers, sc1.get());
  }
```

pgrete: Now I'm confused and am not sure we're talking about the same thing.
I understood your initial comment as "that `for` loop should not go over all blocks". That was the baseline for all my replies (-> "but we need to handle all blocks so the loop should [generally] include all blocks").

Could you maybe rephrase your initial comment and/or provide additional context/explanation? Or we can talk about it on Wednesday during the Profiling call.

Yurlungur: I guess my comment is that the for loop should (in general) loop over task lists, not blocks. In the case of this particular loop, and indeed all loops where you're doing all tasks on all blocks, they're the same.

But if you have a different number of task lists than you do `MeshBlockData`/`MeshData` objects, you're going to run into trouble in weird and subtle ways.

It's not super important. We don't need to discuss it over the call.

src/bvals/cc/bvals_cc_in_one.cpp
```diff
@@ -0,0 +1,553 @@
...
+      // Need to update the buffer sizes so that it matches the packed Variables.
+      assert(rc->GetCellVariableVector().size() == 1);
+      auto *bd_var_ = rc->GetCellVariableVector()[0]->vbvar->GetPBdVar();
+      if (bd_var_->sflag[nb.bufid] == parthenon::BoundaryStatus::completed) continue;
+      buffers_used += 1;
+    }
+  }
+  Kokkos::Profiling::popRegion(); // Reset boundaries
+
+  auto boundary_info = md->GetSendBuffers();
```
Yurlungur: I know kokkos views already have reference semantics, but should we add a `&` to this for good measure?

src/interface/mesh_data.hpp
```diff
@@ -92,6 +93,18 @@ class MeshData {
     }
   }
 
+  void SetSendBuffers(const cell_centered_bvars::BufferCache_t &send_buffers) {
+    send_buffers_ = send_buffers;
+  }
+
+  auto GetSendBuffers() const { return send_buffers_; }
```
Yurlungur:
```suggestion
  auto &GetSendBuffers() const { return send_buffers_; }
```

Yurlungur:
@pgrete what's the error you're getting on the regression tests? And under what conditions? I'm trying to poke at this now.

pgrete:
Here's the output of a failed run (one without the `|| true`): https://gitlab.com/pgrete/parthenon/-/jobs/886138389

So for 2d the even the total number of MeshBlocks is off whereas for 3D the contents differ (above our threshold `Files advection_3d.out0.00001.phdf and /builds/pgrete/parthenon/tst/regression/gold_standard/advection_3d.out0.00001.phdf are different`)
These are our standard short regression tests (run e.g. by `ctest -V -R regression_mpi_test:output_hdf5`) and failing identically on both CPUs and GPUs.

Yurlungur:
> Here's the output of a failed run (one without the `|| true`): https://gitlab.com/pgrete/parthenon/-/jobs/886138389
> 
> So for 2d the even the total number of MeshBlocks is off whereas for 3D the contents differ (above our threshold `Files advection_3d.out0.00001.phdf and /builds/pgrete/parthenon/tst/regression/gold_standard/advection_3d.out0.00001.phdf are different`)
> These are our standard short regression tests (run e.g. by `ctest -V -R regression_mpi_test:output_hdf5`) and failing identically on both CPUs and GPUs.

Got it thanks.

src/bvals/cc/bvals_cc_in_one.cpp
```diff
@@ -0,0 +1,553 @@
...
+      if (v->IsSet(parthenon::Metadata::FillGhost)) {
+        v->resetBoundary();
+      }
+    }
+
+    for (int n = 0; n < pmb->pbval->nneighbor; n++) {
+      parthenon::NeighborBlock &nb = pmb->pbval->neighbor[n];
+      // TODO(?) currently this only works for a single "Variable" per container.
+      // Need to update the buffer sizes so that it matches the packed Variables.
+      assert(rc->GetCellVariableVector().size() == 1);
```
Yurlungur: I just noticed that these asserts *always* fail, even for `advection` because there are three vectors. You actually want this to check for the number of variables with `Metadata::FillGhost` set.

src/bvals/cc/bvals_cc_in_one.cpp
```diff
@@ -0,0 +1,647 @@
...
+    auto &rc = md->GetBlockData(block);
+    auto pmb = rc->GetBlockPointer();
+
+    int mylevel = pmb->loc.level;
+    for (auto &v : rc->GetCellVariableVector()) {
+      if (v->IsSet(parthenon::Metadata::FillGhost)) {
+        v->resetBoundary();
+        for (int n = 0; n < pmb->pbval->nneighbor; n++) {
+          parthenon::NeighborBlock &nb = pmb->pbval->neighbor[n];
+          auto *bd_var_ = v->vbvar->GetPBdVar();
```
Yurlungur: The naming here is a little weird. If it's a pointer, shouldn't it be `pbd_var`?

src/bvals/cc/bvals_cc_in_one.cpp
```diff
@@ -0,0 +1,647 @@
...
+          if (bd_var_->sflag[nb.bufid] == parthenon::BoundaryStatus::completed) continue;
+          buffers_used += 1;
+
+          // Need to restrict here only if cached boundary_info is reused
+          // Otherwise restriction happens when the new boundary_info is created
```
Yurlungur, line 259: this is a little weird. Can restriction not be done in only one place?

src/bvals/cc/bvals_cc_in_one.cpp
```diff
@@ -0,0 +1,647 @@
...
+//  \return Complete when buffer filling is done and MPI communication started.
+//          Guarantees that buffers for MeshBlocks on the same rank are done, but MPI
+//          communication between ranks may still be in process.
+
+// TODO(pgrete) should probaly be moved to the bvals or interface folders
+TaskStatus SendBoundaryBuffers(std::shared_ptr<MeshData<Real>> &md) {
+  Kokkos::Profiling::pushRegion("Task_SendBoundaryBuffers_MeshData");
+
+  auto boundary_info = md->GetSendBuffers();
+  bool cache_is_valid = boundary_info.is_allocated();
```
Yurlungur, line 420: :+1:

src/parthenon_manager.cpp
```diff
@@ -75,6 +75,20 @@ ParthenonStatus ParthenonManager::ParthenonInit(int argc, char *argv[]) {
 
   Kokkos::initialize(argc, argv);
 
+  // pgrete: This is a hack to disable allocation tracking until the Kokkos
+  // tools provide a more fine grained control out of the box.
+  auto *env_track_alloc = std::getenv("KOKKOS_TRACK_ALLOC_OFF");
```
Yurlungur, line 80: is an environment variable how we want to keep track of this? We could also put it into `pin` or something?

Yurlungur (commented):
Our local resources are down for maintenance at the moment. But when we're back up, we'll try this with riot.

src/parthenon_manager.cpp
pgrete: Yes, I think that the most consistent way of action (given that the Kokkos profiling itself is also controlled via setting the `KOKKOS_PROFILE_LIBRARY` variable).
Also, this change comes from a different PR #388

src/bvals/cc/bvals_cc_in_one.cpp
pgrete: Unfortunately not with the current caching. More refactoring would be required to make that work.
The key design problem is that the caching only caches the buffer filling itself, which allows to keep the buffer filling kernel itself very simple.
One alternative approach (I already though about this) is to add more information to the cache, i.e, the `boundary_info`, so that the restriction could happen within the buffer filing kernel inline.
Another alternative is to completely separate the restriction into an own task and call that beforehand. This, in turn, would mean to introduce another cache where we store the indices and views to be restricted.

So bottom line: There's room for improvement but it would require some (potentially more extensive) refactoring.

src/bvals/cc/bvals_cc_in_one.cpp
pgrete: Yes, I'll update that.

src/bvals/cc/bvals_cc_in_one.cpp
Yurlungur: Understood. Good enough for now then. I think doing restriction as a separate task (in the long term) may be of value... if it's possible to disentangle. It may not be.

src/reconstruct/dc_inline.hpp
```diff
@@ -30,50 +30,47 @@ namespace parthenon {
 //! \fn Reconstruction::DonorCellX1()
 //  \brief reconstruct L/R surfaces of the i-th cells
 
-KOKKOS_FORCEINLINE_FUNCTION
-void DonorCellX1(parthenon::team_mbr_t const &member, const int k, const int j,
-                 const int il, const int iu, const ParArrayND<Real> &q,
-                 ScratchPad2D<Real> &ql, ScratchPad2D<Real> &qr) {
+template <typename T>
+KOKKOS_FORCEINLINE_FUNCTION void
+DonorCellX1(parthenon::team_mbr_t const &member, const int k, const int j, const int il,
```
AndrewGaspar, line 35: It's not clear to me why this was made a template. If it's for an external code, that's fine, but I don't see any motivation in this PR myself. I think the only place this is called is in `advection_package.cpp`, and it doesn't look like the types used there changed at all.

If it's for an external code, just put a comment in the PR explaining why this was changed.

src/reconstruct/plm_inline.hpp
```diff
@@ -36,13 +36,13 @@ namespace parthenon {
...
-KOKKOS_INLINE_FUNCTION
-void PiecewiseLinearX1(parthenon::team_mbr_t const &member, const int k, const int j,
-                       const int il, const int iu, const Coordinates_t &coords,
-                       const ParArrayND<Real> &q, ScratchPad2D<Real> &ql,
-                       ScratchPad2D<Real> &qr, ScratchPad2D<Real> &qc,
-                       ScratchPad2D<Real> &dql, ScratchPad2D<Real> &dqr,
-                       ScratchPad2D<Real> &dqm) {
+template <typename T>
+KOKKOS_INLINE_FUNCTION void
+PiecewiseLinearX1(parthenon::team_mbr_t const &member, const int k, const int j,
```
AndrewGaspar, line 41: Same comment as on DonorCell

src/parthenon_manager.cpp
```diff
@@ -75,6 +75,20 @@ ParthenonStatus ParthenonManager::ParthenonInit(int argc, char *argv[]) {
...
+  if (env_track_alloc != nullptr) {
+    std::string env_str(env_track_alloc); // deep-copies string
+    for (char &c : env_str) {
+      c = toupper(c);
+    }
+    if ((env_str == "TRUE") || (env_str == "ON") || (env_str == "1")) {
```
AndrewGaspar, line 86: Haha we have this exact same code all over xRage too.

AndrewGaspar:
@pgrete I'm in the middle of reviewing this. will publish soon.

CHANGELOG.md
```diff
@@ -3,10 +3,12 @@
 ## Current develop
 
 ### Added (new features/APIs/variables/...)
+- [[PR 412]](https://github.com/lanl/parthenon/pull/412) Add capability to use host (pinned) memory for communication buffers (via PARTHENON_ENABLE_HOST_COMM_BUFFERS - default OFF)
 - [[PR 400]](https://github.com/lanl/parthenon/pull/400) Extend `StateDescriptor` for customizable output via user-customizable function pointers `PreStepDiagnosticsMesh` and `PostStepDiagnosticsMesh`
 - [[PR 391]](https://github.com/lanl/parthenon/pull/391) Add `VariablePack<T>::GetSparseId` and `VariablePack<T>::GetSparseIndex` to return global sparse ids and pack-local sparse index, repsectively.
 - [[PR 381]](https://github.com/lanl/parthenon/pull/381) Overload `DataCollection::Add` to build `MeshData` and `MeshBlockData` objects with a subset of variables.
 - [[PR 378]](https://github.com/lanl/parthenon/pull/378) Add Kokkos profiling regions throughout the code to allow the collection characteristic application profiles
+- [[PR 359]](https://github.com/lanl/parthenon/pull/359) MeshBlockPack support for buffer un-/pack of CellCentered Variables
```
AndrewGaspar: "un-/pack" - took me a second to understand what this was trying to say... probably better as
```suggestion
- [[PR 359]](https://github.com/lanl/parthenon/pull/359) MeshBlockPack support for buffer pack and unpack of CellCentered Variables
```

AndrewGaspar (approved):

src/parthenon_manager.cpp
pgrete: I pulled that from Kokkos to keep the logic identical ;)

src/reconstruct/dc_inline.hpp
pgrete: This was actually pulled in from #365  (this branch here has been active for so long I fixed a couple of things along the go to make things work downstream).
Would you prefer that I update #365 to just include those reconstruction changes, get that merged in, and then update this branch
or
add another line in the changelog saying this happenend here, merge this branch and close #365 ?

pgrete:
With respect to getting this PR merged, would you @AndrewGaspar @Yurlungur  like to keep the option to switch between MeshBlock and MeshBlockPack in the advection example or should I remove it?
I'm in favor of keeping it for now and update once the proxy app is ready so that we have one example using MeshBlock based version and another (separate) example using the MeshBlockPack based versions.

Also, @Yurlungur did you get the chance to test this downstream for more than one Variable in riot or shall we merge in good faith for now and fix if there's an issue later? I verified that it works in principle by just adding another independent variable, so  MPI signaling seems to work, but I didn't test for correctness (but I fairly optimistic that it is) .
I'm in favor of merging now so that we can continue to work (on all the other open issues) on "clean" branches.

src/bvals/cc/bvals_cc_in_one.hpp
pgrete: Keeping track of this in #427

Yurlungur:
> With respect to getting this PR merged, would you @AndrewGaspar @Yurlungur like to keep the option to switch between MeshBlock and MeshBlockPack in the advection example or should I remove it?
> I'm in favor of keeping it for now and update once the proxy app is ready so that we have one example using MeshBlock based version and another (separate) example using the MeshBlockPack based versions.

Agreed, let's keep the switch for now and remove it once @jlippuner has the proxy app working.

> Also, @Yurlungur did you get the chance to test this downstream for more than one Variable in riot or shall we merge in good faith for now and fix if there's an issue later? I verified that it works in principle by just adding another independent variable, so MPI signaling seems to work, but I didn't test for correctness (but I fairly optimistic that it is) .
> I'm in favor of merging now so that we can continue to work (on all the other open issues) on "clean" branches.

Apologies, @pgrete I was sick last week, so didn't get a chance to do this yet. Let's go ahead and merge and I will push on the riot version and engage with you if things don't work.

Yurlungur (approved):

src/bvals/cc/bvals_cc.cpp
```diff
@@ -177,59 +179,16 @@ int CellCenteredBoundaryVariable::LoadBoundaryBufferToCoarser(ParArray1D<Real> &
...
-      else
-        ek -= pmb->block_size.nx3 / 2 - pmb->cnghost;
-    } else {
-      if (nb.ni.fi2 == 1)
-        sk += pmb->block_size.nx3 / 2 - pmb->cnghost;
-      else
-        ek -= pmb->block_size.nx3 / 2 - pmb->cnghost;
-    }
-  }
+  cell_centered_bvars::CalcIndicesLoadToFiner(si, ei, sj, ej, sk, ek, nb, pmb.get());
```
Yurlungur, line 191: :+1:

src/bvals/cc/bvals_cc.cpp
```diff
@@ -238,34 +197,24 @@ int CellCenteredBoundaryVariable::LoadBoundaryBufferToFiner(ParArray1D<Real> &bu
...
   IndexDomain interior = IndexDomain::interior;
-  CalcIndices(nb.ni.ox1, si, ei, cellbounds.GetBoundsI(interior));
-  CalcIndices(nb.ni.ox2, sj, ej, cellbounds.GetBoundsJ(interior));
-  CalcIndices(nb.ni.ox3, sk, ek, cellbounds.GetBoundsK(interior));
+  cell_centered_bvars::CalcIndicesSetSame(nb.ni.ox1, si, ei,
+                                          cellbounds.GetBoundsI(interior));
+  cell_centered_bvars::CalcIndicesSetSame(nb.ni.ox2, sj, ej,
+                                          cellbounds.GetBoundsJ(interior));
+  cell_centered_bvars::CalcIndicesSetSame(nb.ni.ox3, sk, ek,
+                                          cellbounds.GetBoundsK(interior));
```
Yurlungur, line 217: :+1:

src/bvals/cc/bvals_cc_in_one.cpp
```diff
@@ -0,0 +1,645 @@
...
+          }
+          boundary_info_h(b).buf = pbd_var_->recv[nb.bufid];
+          // safe to set completed here as the kernel updating all buffers is
+          // called immediately afterwards
+          pbd_var_->flag[nb.bufid] = parthenon::BoundaryStatus::completed;
+          b++;
+        }
+      }
+    }
+  }
```
Yurlungur, line 578: I'm fairly confident this should work for multiple variables.

src/reconstruct/dc_inline.hpp
pgrete: I now updated the changelog here to also reflect this change (as well as the "un-/pack" -> "pack and unpack" change you requested).
@AndrewGaspar  If you're happy feel free to merge (or schedule for merge once the pipeline passes).

PR #362, opened 2020-11-12
Remove old files before running tests

### LINKED ISSUE #361: Rerun of Advection Convergence test fails
The output directory of advection convergence tests needs to be cleaned before it is rerun, otherwise the test will fail. 

"advection_convergence test seems to not clean up advection-errors.dat file.  As a result running the test twice in a row causes it to fail."

JoshuaSBrown:
## PR Summary

Makes sure that tests are run in a clean folder. @nmsriram pointed out that the advection_convergence tests was failing if the folder was clean on a rerun. 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


Yurlungur (approved):

tst/regression/utils/test_case.py
```diff
@@ -182,6 +183,9 @@ def MakeOutputFolder(self):
         if not os.path.isdir(self.parameters.output_path):
             os.makedirs(self.parameters.output_path)
 
+        files = glob.glob(self.parameters.output_path + "/*")
+        for f in files:
+            os.remove(f)
```
pgrete: Could we potentially only clean the required files for specific tests, e.g., the `advection_errors.dat` for the advection test?
I always get goosebumps if automated pattern matching with a `*` is used to remove files in scripts/programs.

JoshuaSBrown: Where would you put the logic? In test_case.py?

pgrete: How about directly in the Prepare function of each test case like
```Python
if step == 1:
  magic_cleanup()
```

JoshuaSBrown: @pgrete I can see having a clean output folder being a good default. How about I simply remove the named output folder, that will avoid the wildcard. We could always have a setting later that turns this default feature off.

JoshuaSBrown: @pgrete

pgrete: Sounds good. My concern was mostly about the wildcard itself ;)

AndrewGaspar (approved):

JoshuaSBrown:
@pgrete can you approve this?

PR #377, opened 2020-11-23
Add machine config file for RZAnsel

JoshuaSBrown:
## PR Summary

This adds a Machine Config file for RZAnsel. This should make it easier for users to get up and running with a working build. Using the ideas that @pgrete had in the Summit config file three variants are allowed to be built, "cuda", "mpi", "cuda+mpi".

Note that I have not removed the old documentation for building on RZAnasel because this was erroneously altered from the Summit documentation. As per my conversation with @pgrete he was going to fix it, so I did not want to touch it.  

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


JoshuaSBrown:
Waiting for this https://re-git.lanl.gov/eap-oss/parthenon-project/-/merge_requests/2 to be merged first.

JoshuaSBrown:
This is now ready for review with the exception of these lines:
```
#set(RZANSEL_PROJECT_PREFIX /usr/gapps/parthenon_shared/parthenon-project
#    CACHE STRING "Path to parthenon-project checkout")

set(RZANSEL_PROJECT_PREFIX /g/g15/brown338/Software/parthenon-project
    CACHE STRING "Path to parthenon-project checkout")
```

I will fix this once the parthenon-project repo is updated and it is updated in RZAnsel.

JoshuaSBrown:
Ok, this is good to go.

CMakeLists.txt
```diff
@@ -122,7 +124,7 @@ endif()
 
 set(ENABLE_MPI OFF)
 set(NUM_MPI_PROC_TESTING "4" CACHE STRING "Number of mpi processors to use when running tests with MPI")
-if (NOT PARTHENON_DISABLE_MPI)
+if (NOT ${PARTHENON_DISABLE_MPI})
```
AndrewGaspar: This shouldn't be necessary.

CMakeLists.txt
```diff
@@ -29,7 +29,9 @@ set(PARTHENON_NGHOST 2 CACHE STRING "Number of ghost cells")
 
 # Compile Options
 option(PARTHENON_SINGLE_PRECISION "Run in single precision" OFF)
+#if(NOT PARTHENON_DISABLE_MPI)
```
AndrewGaspar: I'm guessing you intended to remove this line?

cmake/machinecfg/RZAnsel.cmake
```diff
@@ -34,34 +34,65 @@ set(RZANSEL_ARCH "ppc64le")
...
 set(RZANSEL_COMPILER_PREFERRED "GCC")
 set(RZANSEL_GCC8_VERSION "8.3.1")
-
+set(RZANSEL_VARIANT_PREFIX "_no-mpi")
 set(RZANSEL_MPI_DATE "2020.08.19")
 set(MPIEXEC_EXECUTABLE "/usr/tcetmp/bin/jsrun" CACHE STRING "Mpi executable to use on RZansel system")
 set(RZANSEL_CUDA_VERSION "10.1.243")
 
-set(RZANSEL_CUDA_DEFAULT ON)
+if (NOT MACHINE_VARIANT)
```
AndrewGaspar: Why not just make this an option instead of having to do this string parsing. `RZANSEL_MPI` in addition to `RZANSEL_CUDA`?

cmake/machinecfg/RZAnsel.cmake
```diff
@@ -34,34 +34,65 @@ set(RZANSEL_ARCH "ppc64le")
...
-set(RZANSEL_CUDA_DEFAULT ON)
+if (NOT MACHINE_VARIANT) 
+  set(MACHINE_VARIANT "cuda-mpi")
+endif()
 
-set(GPU_COUNT "4")
-execute_process(
+if (${MACHINE_VARIANT} MATCHES "cuda")
+  set(GPU_COUNT "4")
+  execute_process(
```
AndrewGaspar: You don't need to do any of this detection on RZAnsel at all - you can just assume it has 4 GPUs per-node.

src/CMakeLists.txt
```diff
@@ -14,8 +14,8 @@
 
 # Configure config.hpp
 set(PROBLEM_GENERATOR "<not-implemented>") # TODO: Figure out what to put here
-
-if (ENABLE_MPI)
+if ("${ENABLE_MPI}")
```
AndrewGaspar: This should be functionally equivalent to what was already there.

AndrewGaspar:
Did you intentionally update Kokkos?

cmake/machinecfg/RZAnsel.cmake
```diff
@@ -0,0 +1,258 @@
...
+# OPTIONS:
+# - `RZANSEL_VIEW_DATE` - The date the dependencies were installed.
+#       Default:
+#           unset, which results in using the view associated with your
+#           current commit. See `RZANSEL_VIEW_DATE_LATEST`
+# - `RZANSEL_COMPILER` - Compiler family to use
+#       Default: "GCC"
+#       Possible Values: "GCC", "GCC8"
+# - `RZANSEL_CUDA` - Build for CUDA
+#       Default: ON if nvidia-smi finds at least one GPU, OFF otherwise
```
AndrewGaspar: Just make it `Default: ON`. We don't need to do detection on RZAnsel, we know it has GPUs.

docs/building.md
```diff
@@ -270,6 +270,69 @@ Once you've configured your build directory, you can build with
...
+If you followed the "Set-Up Environment" section, configuration requires 0
+additional arguments:
+```bash
+$ cmake -S. -Bbuild
+```
+By default cmake will build parthenon with cuda and mpi support. Other machine
+variants exist and can be specified by using the `MACHINE_VARIANT` flag. The
+supported machine variants include:
+
+* cuda+mpi
```
AndrewGaspar: Up above you said this was `cuda-mpi`. But just get rid of `MACHINE_VARIANT` and add a separate option for MPI.

cmake/machinecfg/RZAnsel.cmake
JoshuaSBrown: Phil already set a precedent I wanted to be consistent with what had already been done for Summit. Moving forward at some point I would prefer to standardize the options on different systems so it's less confusing for users.

AndrewGaspar (changes requested):

AndrewGaspar:
> Did you intentionally update Kokkos?

Oops, ignore this - I missed that this was a reversion.

src/CMakeLists.txt
JoshuaSBrown: I think it is better practice to wrap variables in ${}, if ENABLE_MPI="Off" then 
```
if(ENABLE_MPI)
```
Will evaluate to true. This has caused me headaches in the past.

cmake/machinecfg/RZAnsel.cmake
AndrewGaspar: I feel like CMake has a perfectly good mechanism for specifying boolean options, so it's weird to replace that with a hyphen-separated list system.

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@
 - [[PR 369]](https://github.com/lanl/parthenon/pull/369) Initializes submodules when running on darwin ci.
 - [[PR 382]](https://github.com/lanl/parthenon/pull/382) Adds output on fail for fast ci implementation on Darwin.
 - [[PR 362]](https://github.com/lanl/parthenon/pull/362) Small fix to clean regression tests output folder on reruns
+- [[PR 377]](https://github.com/lanl/parthenon/pull/377) New machine configuration file for LNLL's RZAnasel cluster
```
AndrewGaspar:
```suggestion
- [[PR 377]](https://github.com/lanl/parthenon/pull/377) New machine configuration file for LNLL's RZAnsel cluster
```

cmake/machinecfg/RZAnsel.cmake
JoshuaSBrown: Ok, I don't really care, I just want it to be consistent. @pgrete what do you think?

CMakeLists.txt
JoshuaSBrown: I responded to the same point below.

src/CMakeLists.txt
AndrewGaspar: ?

```
~/Code/cmake-playground via 🛆 v3.18.2 
at 11:03:38 ❯ cat CMakeLists.txt 
cmake_minimum_required(VERSION 3.16)

project(cmake-playground LANGUAGES NONE)

set(ENABLE_MPI "Off")
if (ENABLE_MPI)
    message(STATUS "Yes MPI")
else()
    message(STATUS "No MPI")
endif()%                                                                                                                                               

~/Code/cmake-playground via 🛆 v3.18.2 
at 11:03:45 ❯ cmake -S. -Bbuild
-- No MPI
-- Configuring done
-- Generating done
-- Build files have been written to: /Users/agaspar/Code/cmake-playground/build
```

Yurlungur (approved):

src/CMakeLists.txt
AndrewGaspar: Regardless, I don't see why changing it from `if(<variable>)` to `if(${<variable>})` would change whether cmake would compare the value as case insensitive.

AndrewGaspar: Doesn't seem to be a version issue - I tested on CMake 3.12 (the minimum supported CMake version) and "Off" is falsey there, too.

JoshuaSBrown: Ah, I stand corrected, I don't know what problem I ran into in the past, I will correct this.

JoshuaSBrown: It actually looks like not including ${} is better practice.

AndrewGaspar: Didn't look like this was resolved yet? Or maybe you just hadn't pushed the changes yet.

JoshuaSBrown: Ya, I just haven't pushed.

src/CMakeLists.txt
```diff
@@ -244,7 +243,7 @@ if (CMAKE_CXX_COMPILER_ID STREQUAL "XL")
 endif()
 
 
-if (ENABLE_MPI)
+if ( "${ENABLE_MPI}" )
```
AndrewGaspar:
```suggestion
if (ENABLE_MPI)
```

cmake/machinecfg/RZAnsel.cmake
JoshuaSBrown: @pgrete

pgrete: My original intention for this was flexibility, e.g., when also changing and I probably didn't want to introduce more variable.
I don't really feel very strong about this.
From my point of view these machine files are suggestions and/or help to jumpstart a new user on a machine.
Eventually, I see everyone using their own machine file for downstream codes anyway (so different machine files can use different/custom logic and there's no right or wrong in writing them)

cmake/machinecfg/RZAnsel.cmake
```diff
@@ -0,0 +1,246 @@
...
+    # Location of CUDA
+    set(CUDAToolkit_ROOT /usr/tce/packages/cuda/cuda-${RZANSEL_CUDA_VERSION} 
+      CACHE STRING "CUDA Location")
+
+    # All of this code ensures that the CUDA build uses the correct nvcc, and
+    # that we don't have to depend on the user's environment. As of this
+    # writing, nvcc_wrapper can only find nvcc by way of the PATH variable -
+    # there's no way to specify it in the command line. Therefore, we need to
+    # make sure that the expected nvcc is in the PATH when nvcc_wrapper is
+    # executed.
```
pgrete: I think nvcc_wraper can use the `CUDA_ROOT` environment variable to find nvcc:
```bash
# Default to whatever is in the path
nvcc_compiler=nvcc
if [ ! -z $CUDA_ROOT ]; then
  nvcc_compiler="$CUDA_ROOT/bin/nvcc"
fi
```

cmake/machinecfg/RZAnsel.cmake
```diff
@@ -0,0 +1,246 @@
...
+    # This only holds for the length of the cmake process, but is necessary
+    # for making sure the compiler checks pass
+    set(ENV{PATH} "${CUDAToolkit_ROOT}/bin:$ENV{PATH}")
+
+    # We must make sure nvcc is in the path for the compilation and link stages,
+    # so these launch rules add it to the path. This unfortunately caches the
+    # PATH environment variable at the time of configuration. I couldn't figure
+    # out how to evaluate the PATH variable at build time. Ideally, changing
+    # your PATH shouldn't change the build anyway, so I think this is low
+    # impact, but it does differ from typical CMake usage.
```
pgrete: Not sure this is true (see comment above on using nvcc from `$CUDA_ROOT`

docs/building.md
```diff
@@ -270,6 +270,69 @@ Once you've configured your build directory, you can build with
 LANL Employees - to understand how the project space is built out, see
 https://xcp-gitlab.lanl.gov/eap-oss/parthenon-project
 
+### LNLL RZAnsel (Homogeneous)
+
+Last verified 04 Jan 2021.
+
+#### Allocate Node
+
+RZAnasel is a homogeneous cluster consiting of 2,376 nodes with the IBM Power9
```
pgrete:
```suggestion
[RZAnsel](https://hpc.llnl.gov/hardware/platforms/rzansel) is a homogeneous cluster consisting of 2,376 nodes with the IBM Power9
```

Two typos and may also be nice to link to the machine homepage.

pgrete (commented):
In addition to the inline comments, I wonder how much of the logic in the machine is actually required, e.g., setting paths etc.
On other machines my experience is that cmake is pretty smart in finding the correct libraries if the correct compiler with full path is given (i.e., setting `CMAKE_CXX_COMPILER` is usually enough and RZAnsel generally looks quite similar to Summit from the outside).

cmake/TestSetup.cmake
```diff
@@ -108,52 +108,58 @@ endfunction()
...
+      process_mpi_args(${nproc})
+      add_test( NAME regression_mpi_test:${dir} COMMAND ${Python3_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/run_test.py
+        ${MPIARGS} ${arg}
+        --test_dir ${CMAKE_CURRENT_SOURCE_DIR}/test_suites/${dir}
+        --output_dir "${PROJECT_BINARY_DIR}/tst/regression/outputs/${dir}_mpi"
+        --kokkos_args=${PARTHENON_KOKKOS_TEST_ARGS})
+      set_tests_properties(regression_mpi_test:${dir} PROPERTIES LABELS "${labels}" RUN_SERIAL ON )
+      record_driver("${arg}")
+    else()
+      message(STATUS "MPI not found, not building regression tests with mpi")
```
pgrete: One more question:
When does the new additional conditional actually kick it?
According to our main CMakeLists.txt
```
set(ENABLE_MPI OFF)
set(NUM_MPI_PROC_TESTING "4" CACHE STRING "Number of mpi processors to use when running tests with MPI")
if (NOT PARTHENON_DISABLE_MPI)
  find_package(MPI COMPONENTS CXX)
  if (NOT MPI_FOUND)
    message(FATAL_ERROR "MPI is required but couldn't be found. "
    "If you want to build Parthenon without MPI, please rerun CMake with -DPARTHENON_DISABLE_MPI=ON")
  endif()
  set(ENABLE_MPI ON)
endif()
```
ENABLE_MPI=ON implies that MPI_FOUND is true, otherwise it would have failed already earlier.

JoshuaSBrown: Yes, ENABLE_MPI=ON implies MPI_FOUND is true, but just because MPI is found does not mean the user wants to  compile with it. Maybe I am missing something, but when I was just trying to compile with only cuda support on RZAnsel, this code block was getting executed and causing problems.

cmake/machinecfg/RZAnsel.cmake
JoshuaSBrown: @pgrete I would agree with that statement from the point of view of user specefic machine config files, but I feel like we should have a slightly higher standard for machine config files we actually maintain. Regardless this is not a priority. @AndrewGaspar if we are not going to standardize this, I don't think it is worth changing.

cmake/machinecfg/RZAnsel.cmake
JoshuaSBrown: @AndrewGaspar do you want to comment, this section was basically copied from Darwin machine config.

cmake/machinecfg/RZAnsel.cmake
pgrete: Trying on Summit I remember now why I made that choice for a single parameter: convenience/laziness (I only wanted to change one variable on the command line rather than one variable per option)

And now that I think even more about this, I'm wondering why, e.g.,  the `RZANSEL_CUDA` or `RZANSEL_CXX_COMPILER` variables are required at all (instead of setting `Kokkos_ENABLE_CUDA` or `CMAKE_CXX_COMPILER`, respectively).
From my point of view the main purpose of the machine files is to set reasonable defaults for a specific machine. For this reason, those defaults (the values) can be very machine specific (e.g., full path to compiler etc) but don't necessarily need to/should be new variables (if the required functionality is already provided by Parthenon default variables).

Bottom line, I'm now in favor of drastically simplifying the machine file (if you like I can give it a shot).

cmake/TestSetup.cmake
pgrete: I see, I guess I practically never compile without MPI and this didn't come across.
I assume the the problem is that I missed that case in the Summit example.
If "mpi" is not in the string then `PARTHENON_DISABLE_MPI=ON` should be set and everything works as expected.
I'll update the Summit file including the docs shortly.

cmake/machinecfg/RZAnsel.cmake
JoshuaSBrown: I think we should continue this conversation on this issue. https://github.com/lanl/parthenon/issues/372

cmake/machinecfg/RZAnsel.cmake
AndrewGaspar: Oh, sure, this change was added in Kokkos 3.2.00, and I think when I originally wrote this code, we weren't using Kokkos 3.2.00. Using `CUDA_ROOT` is a better solution.

cmake/machinecfg/RZAnsel.cmake
AndrewGaspar: `RZANSEL_CXX_COMPILER` isn't a user option. It's just a variable to store the host compiler.

The reason I made `RZANSEL_CUDA` separate from `Kokkos_ENABLE_CUDA` is because at some point my intention was to actually just build out Kokkos in the spack environment as well, in which case `Kokkos_ENABLE_CUDA` would do nothing to Kokkos and merely control which environment you use in RZAnsel.

What I didn't want to do was make the machine configuration depend on a bunch of variables which aren't defined (along with their defaults) until later in the CMake configuration. So I figured I'd scope the variable names to the machine.

cmake/machinecfg/RZAnsel.cmake
```diff
@@ -0,0 +1,246 @@
...
+set(RZANSEL_GCC_PREFERRED "GCC8")
+set(RZANSEL_COMPILER_PREFERRED "GCC")
+set(RZANSEL_GCC8_VERSION "8.3.1")
+set(RZANSEL_VARIANT_PREFIX "_no-mpi")
+set(RZANSEL_MPI_DATE "2020.08.19")
+set(MPIEXEC_EXECUTABLE "/usr/tcetmp/bin/jsrun" CACHE STRING "Mpi executable to use on RZansel system")
+set(RZANSEL_CUDA_VERSION "10.1.243")
+
+if (NOT MACHINE_VARIANT) 
+  set(MACHINE_VARIANT "cuda-mpi")
```
AndrewGaspar: If we're going to do this `MACHINE_VARIANT` thing, it should just be a cache variable with this as the default.

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@
 - [[PR 382]](https://github.com/lanl/parthenon/pull/382) Adds output on fail for fast ci implementation on Darwin.
 - [[PR 362]](https://github.com/lanl/parthenon/pull/362) Small fix to clean regression tests output folder on reruns
 - [[PR 403]](https://github.com/lanl/parthenon/pull/403) Cleanup Codacy warnings
+- [[PR 377]](https://github.com/lanl/parthenon/pull/377) New machine configuration file for LNLL's RZAnsel cluster
```
AndrewGaspar:
```suggestion
- [[PR 377]](https://github.com/lanl/parthenon/pull/377) New machine configuration file for LLNL's RZAnsel cluster
```

JoshuaSBrown:
@pgrete @AndrewGaspar I think I addressed the relevant issues.

cmake/TestSetup.cmake
```diff
@@ -108,7 +108,7 @@ endfunction()
 # test output will be sent to /tst/regression/outputs/dir_mpi
 # test property labels: regression, mpi-yes
 function(setup_test_mpi nproc dir arg extra_labels)
-  if( MPI_FOUND )
+  if( MPI_FOUND)
```
pgrete:
```suggestion
  if( MPI_FOUND )
```

cmake/machinecfg/RZAnsel.cmake
```diff
@@ -0,0 +1,214 @@
...
+#       Possible Values: "GCC", "GCC8"
+# - `RZANSEL_CUDA` - Build for CUDA
+#       Default: ON 
+# - `RZANSEL_PROJECT_PREFIX`
+#   Description: [ADVANCED] Point to an alternative parthenon-project path
+#       Default: /usr/gapps/parthenon_shared/parthenon-project
+
+cmake_minimum_required(VERSION 3.15)
+
+message(STATUS "Loading machine configuration for LNLL RZAnsel.\n"
```
pgrete:
```suggestion
message(STATUS "Loading machine configuration for LLNL RZAnsel.\n"
```

cmake/machinecfg/RZAnsel.cmake
```diff
@@ -0,0 +1,214 @@
...
+
+# It would be nice if we could let this variable float with the current code
+# checkout, but unfortunately CMake caches enough other stuff (like find
+# packages), that it's easier to just issue a warning if the view date is
+# different from the "latest" date.
+set(RZANSEL_VIEW_DATE ${RZANSEL_VIEW_DATE_LATEST}
+  CACHE STRING "RZAnsel dependency view being used")
+if (NOT RZANSEL_VIEW_DATE_LATEST STREQUAL RZANSEL_VIEW_DATE)
+    message(WARNING "Your current build directory was configured with a \
+    set of RZAnasel dependencies from ${RZANSEL_VIEW_DATE}, but your current \
```
pgrete:
```suggestion
    set of RZAnsel dependencies from ${RZANSEL_VIEW_DATE}, but your current \
```

cmake/machinecfg/RZAnsel.cmake
```diff
@@ -0,0 +1,214 @@
...
+set(NUM_RANKS 4) 
+set(GPU_COUNT "4")
+
+if (${MACHINE_VARIANT} MATCHES "cuda")
+  set(RZANSEL_CUDA_DEFAULT ON)
+else()
+  set(RZANSEL_CUDA_DEFAULT OFF)
+endif()
+
+set(RZANSEL_CUDA ${RZANSEL_CUDA_DEFAULT} CACHE BOOL "Build for CUDA")
```
pgrete, line 58:
```suggestion
if (${MACHINE_VARIANT} MATCHES "cuda")
  set(RZANSEL_CUDA ON CACHE BOOL "Build for CUDA")
else()
  set(RZANSEL_CUDA OFF CACHE BOOL "Build for CUDA")
endif()
```

I don't see `RZANSEL_CUDA_DEFAULT` being used anywhere else. So if I'm not missing anything one could skip that intermediate variable.

cmake/machinecfg/RZAnsel.cmake
```diff
@@ -0,0 +1,214 @@
...
+
+if (${MACHINE_VARIANT} MATCHES "mpi")
+  set(RZANSEL_VARIANT_PREFIX "")
+  set(NUM_MPI_PROC_TESTING ${NUM_RANKS} CACHE STRING "CI runs tests with 2 MPI ranks by default.")
+  set(MPIEXEC_EXECUTABLE "/usr/tcetmp/bin/jsrun" CACHE STRING "Command to launch MPI applications")
+  set(TEST_NUMPROC_FLAG "-c" CACHE STRING "Flag to set number of processes")
+  if (${MACHINE_VARIANT} MATCHES "cuda")
+    string(APPEND TEST_MPIOPTS "-a 1 -n 1 -g ${NUM_GPU_DEVICES_PER_NODE} -r 1 -d packed --smpiargs='-gpu'")
+  else()
+    string(APPEND TEST_MPIOPTS "-a 1 -n 1 -r 1 -d packed")
```
pgrete: Would you mind double checking this (that -c controls the number of processes) is true?
On Summit with `jsrun`  the parameter have different meaning. `-c` controls the number of cores allocated per resource set. The total number of tasks (i.e., MPI ranks) is controlled by the combination of `-n` (number or resource sets) and `-a` (number of tasks per resource set).

pgrete (commented):
Thanks for the update. I got a couple of final comments (typos, one suggestion to remove a variable, and one question on launching tasks on RZAnsel).

cmake/machinecfg/RZAnsel.cmake
JoshuaSBrown: Thanks for the feedback, I'll be sure to check, I probably did something wrong. I was getting errors when trying to run the tests when using -a 4, but it might just be that some of the tests are not set up to run on 4 ranks.

AndrewGaspar (approved):

JoshuaSBrown:
@pgrete I could use some help with this, either there is some logic in parthenon that needs to be corrected or the tests simply are not configured to run with 4 gpus and 4 mpi procs:

This first scenario should be the one that works:

I can verify that nvidia-smi indicates 4 tasks are launched. 

```
    Start 34: regression_mpi_test:advection_performance
1/6 Test #34: regression_mpi_test:advection_performance ...***Failed   68.00 sec


test_dir=['/g/g15/brown338/Software/parthenon/tst/regression/test_suites/advection_performance']
output_dir='/g/g15/brown338/Software/parthenon/build/tst/regression/outputs/advection_performance_mpi'
driver=['/g/g15/brown338/Software/parthenon/build/example/advection/advection-example']
driver_input=['/g/g15/brown338/Software/parthenon/tst/regression/test_suites/advection_performance/parthinput.advection_performance']
kokkos_args=['--kokkos-num-devices=1 --kokkos-threads=1']
num_steps=5
mpirun=['/usr/tcetmp/bin/jsrun']
mpirun_opts=['-a', '4', "-c 1 -n 1 -g 1 -r 1 -d packed --smpiargs='-gpu'"]
coverage=False
*****************************************************************
Beginning Python regression testing script
*****************************************************************

Initializing Test Case
Using:
driver at:       /g/g15/brown338/Software/parthenon/build/example/advection/advection-example
driver input at: /g/g15/brown338/Software/parthenon/tst/regression/test_suites/advection_performance/parthinput.advection_performance
test folder:     /g/g15/brown338/Software/parthenon/tst/regression/test_suites/advection_performance
output sent to:  /g/g15/brown338/Software/parthenon/build/tst/regression/outputs/advection_performance_mpi

Make output folder in test if does not exist
*****************************************************************
Preparing Test Case Step 1
*****************************************************************

*****************************************************************
Running Driver
*****************************************************************

Command to execute driver
/usr/tcetmp/bin/jsrun -a 4 -c 1 -n 1 -g 1 -r 1 -d packed --smpiargs='-gpu' /g/g15/brown338/Software/parthenon/build/example/advection/advection-example -i /g/g15/brown338/Software/parthenon/tst/regression/test_suites/advection_performance/parthinput.advection_performance parthenon/mesh/nx1=256 parthenon/meshblock/nx1=256 parthenon/mesh/nx2=256 parthenon/meshblock/nx2=256 parthenon/mesh/nx3=256 parthenon/meshblock/nx3=256 --kokkos-num-devices=1 --kokkos-threads=1

*****************************************************************
Subprocess error message
*****************************************************************

b'### PARTHENON ERROR
  Message:     ### FATAL ERROR in Mesh constructor
Too few mesh blocks: nbtotal (1) < nranks (4)

  File:        ../src/mesh/mesh.cpp
  Line number: 450
### PARTHENON ERROR
  Message:     ### FATAL ERROR in Mesh constructor
Too few mesh blocks: nbtotal (1) < nranks (4)

  File:        ../src/mesh/mesh.cpp
  Line number: 450
### PARTHENON ERROR
  Message:     ### FATAL ERROR in Mesh constructor
Too few mesh blocks: nbtotal (1) < nranks (4)

  File:        ../src/mesh/mesh.cpp
  Line number: 450
### PARTHENON ERROR
  Message:     ### FATAL ERROR in Mesh constructor
Too few mesh blocks: nbtotal (1) < nranks (4)

  File:        ../src/mesh/mesh.cpp
  Line number: 450
'

```

```
    Start 34: regression_mpi_test:advection_performance
1/6 Test #34: regression_mpi_test:advection_performance ...***Failed   66.39 sec


test_dir=['/g/g15/brown338/Software/parthenon/tst/regression/test_suites/advection_performance']
output_dir='/g/g15/brown338/Software/parthenon/build/tst/regression/outputs/advection_performance_mpi'
driver=['/g/g15/brown338/Software/parthenon/build/example/advection/advection-example']
driver_input=['/g/g15/brown338/Software/parthenon/tst/regression/test_suites/advection_performance/parthinput.advection_performance']
kokkos_args=['--kokkos-num-devices=4 --kokkos-threads=1']
num_steps=5
mpirun=['/usr/tcetmp/bin/jsrun']
mpirun_opts=['-a', '4', "-c 1 -n 1 -g 1 -r 1 -d packed --smpiargs='-gpu'"]
coverage=False
*****************************************************************
Beginning Python regression testing script
*****************************************************************

Initializing Test Case
Using:
driver at:       /g/g15/brown338/Software/parthenon/build/example/advection/advection-example
driver input at: /g/g15/brown338/Software/parthenon/tst/regression/test_suites/advection_performance/parthinput.advection_performance
test folder:     /g/g15/brown338/Software/parthenon/tst/regression/test_suites/advection_performance
output sent to:  /g/g15/brown338/Software/parthenon/build/tst/regression/outputs/advection_performance_mpi

Make output folder in test if does not exist
*****************************************************************
Preparing Test Case Step 1
*****************************************************************

*****************************************************************
Running Driver
*****************************************************************

Command to execute driver
/usr/tcetmp/bin/jsrun -a 4 -c 1 -n 1 -g 1 -r 1 -d packed --smpiargs='-gpu' /g/g15/brown338/Software/parthenon/build/example/advection/advection-example -i /g/g15/brown338/Software/parthenon/tst/regression/test_suites/advection_performance/parthinput.advection_performance parthenon/mesh/nx1=256 parthenon/meshblock/nx1=256 parthenon/mesh/nx2=256 parthenon/meshblock/nx2=256 parthenon/mesh/nx3=256 parthenon/meshblock/nx3=256 --kokkos-num-devices=4 --kokkos-threads=1

*****************************************************************
Subprocess error message
*****************************************************************

b''

*****************************************************************
Error detected while running subprocess command
*****************************************************************

  File "/g/g15/brown338/Software/parthenon/tst/regression/utils/test_case.py", line 237, in Run
    proc = subprocess.run(run_command, check=True, stdout=PIPE, stderr=PIPE)
  File "/usr/gapps/parthenon_shared/parthenon-project/views/rzansel/ppc64le/gcc8/2021-01-04/lib/python3.8/subprocess.py", line 512, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['/usr/tcetmp/bin/jsrun', '-a', '4', '-c', '1', '-n', '1', '-g', '1', '-r', '1', '-d', 'packed', "--smpiargs='-gpu'", '/g/g15/brown338/Software/parthenon/build/example/advection/advection-example', '-i', '/g/g15/brown338/Software/parthenon/tst/regression/test_suites/advection_performance/parthinput.advection_performance', 'parthenon/mesh/nx1=256', 'parthenon/meshblock/nx1=256', 'parthenon/mesh/nx2=256', 'parthenon/meshblock/nx2=256', 'parthenon/mesh/nx3=256', 'parthenon/meshblock/nx3=256', '--kokkos-num-devices=4', '--kokkos-threads=1']' returned non-zero exit status 134.

During handling of the above exception, another exception occurred:

  File "/g/g15/brown338/Software/parthenon/tst/regression/run_test.py", line 152, in <module>
    main(**vars(args))
  File "/g/g15/brown338/Software/parthenon/tst/regression/run_test.py", line 76, in main
    test_manager.Run()
  File "/g/g15/brown338/Software/parthenon/tst/regression/utils/test_case.py", line 247, in Run
    raise TestManagerError('\nReturn code {0} from command \'{1}\''
utils.test_case.TestManagerError: 
Return code 134 from command '/usr/tcetmp/bin/jsrun -a 4 -c 1 -n 1 -g 1 -r 1 -d packed --smpiargs='-gpu' /g/g15/brown338/Software/parthenon/build/example/advection/advection-example -i /g/g15/brown338/Software/parthenon/tst/regression/test_suites/advection_performance/parthinput.advection_performance parthenon/mesh/nx1=256 parthenon/meshblock/nx1=256 parthenon/mesh/nx2=256 parthenon/meshblock/nx2=256 parthenon/mesh/nx3=256 parthenon/meshblock/nx3=256 --kokkos-num-devices=4 --kokkos-threads=1'
```

```
    Start 34: regression_mpi_test:advection_performance
1/6 Test #34: regression_mpi_test:advection_performance ...***Failed   68.12 sec


test_dir=['/g/g15/brown338/Software/parthenon/tst/regression/test_suites/advection_performance']
output_dir='/g/g15/brown338/Software/parthenon/build/tst/regression/outputs/advection_performance_mpi'
driver=['/g/g15/brown338/Software/parthenon/build/example/advection/advection-example']
driver_input=['/g/g15/brown338/Software/parthenon/tst/regression/test_suites/advection_performance/parthinput.advection_performance']
kokkos_args=['--kokkos-num-devices=1 --kokkos-threads=1']
num_steps=5
mpirun=['/usr/tcetmp/bin/jsrun']
mpirun_opts=['-a', '4', "-c 1 -n 1 -g 4 -r 1 -d packed --smpiargs='-gpu'"]
coverage=False
*****************************************************************
Beginning Python regression testing script
*****************************************************************

Initializing Test Case
Using:
driver at:       /g/g15/brown338/Software/parthenon/build/example/advection/advection-example
driver input at: /g/g15/brown338/Software/parthenon/tst/regression/test_suites/advection_performance/parthinput.advection_performance
test folder:     /g/g15/brown338/Software/parthenon/tst/regression/test_suites/advection_performance
output sent to:  /g/g15/brown338/Software/parthenon/build/tst/regression/outputs/advection_performance_mpi

Make output folder in test if does not exist
*****************************************************************
Preparing Test Case Step 1
*****************************************************************

*****************************************************************
Running Driver
*****************************************************************

Command to execute driver
/usr/tcetmp/bin/jsrun -a 4 -c 1 -n 1 -g 4 -r 1 -d packed --smpiargs='-gpu' /g/g15/brown338/Software/parthenon/build/example/advection/advection-example -i /g/g15/brown338/Software/parthenon/tst/regression/test_suites/advection_performance/parthinput.advection_performance parthenon/mesh/nx1=256 parthenon/meshblock/nx1=256 parthenon/mesh/nx2=256 parthenon/meshblock/nx2=256 parthenon/mesh/nx3=256 parthenon/meshblock/nx3=256 --kokkos-num-devices=1 --kokkos-threads=1

*****************************************************************
Subprocess error message
*****************************************************************

b'### PARTHENON ERROR
  Message:     ### FATAL ERROR in Mesh constructor
Too few mesh blocks: nbtotal (1) < nranks (4)

  File:        ../src/mesh/mesh.cpp
  Line number: 450
### PARTHENON ERROR
  Message:     ### FATAL ERROR in Mesh constructor
Too few mesh blocks: nbtotal (1) < nranks (4)

  File:        ../src/mesh/mesh.cpp
  Line number: 450
### PARTHENON ERROR
  Message:     ### FATAL ERROR in Mesh constructor
Too few mesh blocks: nbtotal (1) < nranks (4)

  File:        ../src/mesh/mesh.cpp
  Line number: 450
### PARTHENON ERROR
  Message:     ### FATAL ERROR in Mesh constructor
Too few mesh blocks: nbtotal (1) < nranks (4)

  File:        ../src/mesh/mesh.cpp
  Line number: 450
'

```

@jonahm-LANL I could use your input here as welll.

pgrete:
This simple `advection_performance` test is currently setup to measure the overhead in overdecomposition, i.e., for a fixed mesh size (here 256^3) the number of meshblocks into which the Mesh is separated is successively increased (by decreasing the block size).
Thus, this test assumes that the number of compute elements (e.g., a GPU) remains constant and (at the same time) that there's only one compute element (as the baseline performance is one MeshBlock covering the entire Mesh).
We'd need to setup various different tests to capture various performance aspects (specifically thinking of the proxy app here given that it'll allow us to increase to compute versus infrastructure related pieces in the code to more realistic levels).

AndrewGaspar:
@pgrete Yeah, that's all well and good - we should definitely ensure we're holding the num of GPUs constant when we're doing performance testing, which you can control using `NUM_GPU_DEVICES_PER_NODE`. I think all Josh is trying to ask is how do you get jsrun to correctly partition the GPUs.

pgrete:
> @pgrete Yeah, that's all well and good - we should definitely ensure we're holding the num of GPUs constant when we're doing performance testing, which you can control using `NUM_GPU_DEVICES_PER_NODE`. I think all Josh is trying to ask is how do you get jsrun to correctly partition the GPUs.

Sorry, my previous comment was probably not clear.
I was trying to say that *this specific advection_performance* performance test makes only sense when run on a single GPU as the baseline in that test is using a single MeshBlock for the entire Mesh. Thus, independent of CPU or GPU that test can only be run using a single rank (that's also where the error message comes from).
Thus, this test should not be run with
`/usr/tcetmp/bin/jsrun -a 4 -c 1 -n 1 -g 4 -r 1 -d packed --smpiargs='-gpu' /g/g15/brown338`
but rather with
`/usr/tcetmp/bin/jsrun -a 1 -c 1 -n 1 -g 1 -r 1 -d packed --smpiargs='-gpu' /g/g15/brown338`

It may be worth to add safety check (i.e., ranks == 1) to that test similar to the advection_convergence test:
```python
        # make sure we can evenly distribute the MeshBlock sizes
        err_msg = "Num ranks must be multiples of 2 for convergence test." 
        assert parameters.num_ranks == 1 or parameters.num_ranks % 2 == 0, err_msg                                             
        # ensure a minimum block size of 4
        assert lin_res[0] / parameters.num_ranks >= 4, "Use <= 8 ranks for convergence test."
```

JoshuaSBrown:
Thanks, everyone for the feedback, @pgrete do you want to take a last look over?

pgrete:
I think making sure that the `advection_performance` test in its current version is only run by a single rank is good!
With respect to the `advection_convergence` this change is not required (this was probably not clear from our discussion on Matrix). In fact, `advection_convergence` is designed to work with both a single rank as well as with 2, 4 and 8 ranks. So I think we should not set that number to 1.

JoshuaSBrown:
> I think making sure that the `advection_performance` test in its current version is only run by a single rank is good!
> With respect to the `advection_convergence` this change is not required (this was probably not clear from our discussion on Matrix). In fact, `advection_convergence` is designed to work with both a single rank as well as with 2, 4 and 8 ranks. So I think we should not set that number to 1.

Ok, well I'm thoroughly confused, because when I run the convergence tests with 4 ranks and 4 gpus, it is still only making use of a single gpu. I'm also pretty sure I have the correct command because the restart test will actually utilize all four gpu's with the same command.

pgrete:
> > I think making sure that the `advection_performance` test in its current version is only run by a single rank is good!
> > With respect to the `advection_convergence` this change is not required (this was probably not clear from our discussion on Matrix). In fact, `advection_convergence` is designed to work with both a single rank as well as with 2, 4 and 8 ranks. So I think we should not set that number to 1.
> 
> Ok, well I'm thoroughly confused, because when I run the convergence tests with 4 ranks and 4 gpus, it is still only making use of a single gpu. I'm also pretty sure I have the correct command because the restart test will actually utilize all four gpu's with the same command.

Let met double check in practice. At least on first sight I don't see a difference that may cause that behavior in the parameter/CMake files.

JoshuaSBrown:
```

/usr/tcetmp/bin/jsrun -a 4 -c 1 -n 1 -g 4 -r 1 -d packed --smpiargs='-gpu' /g/g15/brown338/Software/parthenon/build/example/advection/advection-example -i /g/g15/brown338/Software/parthenon/tst/regression/test_suites/restart/parthinput.restart parthenon/job/problem_id=gold --kokkos-num-devices=4 --kokkos-threads=1

```

Gives me 

```
Mon Feb  1 13:10:48 2021
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 440.95.01    Driver Version: 440.95.01    CUDA Version: 10.2     |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|===============================+======================+======================|
|   0  Tesla V100-SXM2...  On   | 00000004:04:00.0 Off |                    0 |
| N/A   29C    P0    51W / 300W |    446MiB / 16160MiB |      5%      Default |
+-------------------------------+----------------------+----------------------+
|   1  Tesla V100-SXM2...  On   | 00000004:05:00.0 Off |                    0 |
| N/A   29C    P0    50W / 300W |    446MiB / 16160MiB |      6%      Default |
+-------------------------------+----------------------+----------------------+
|   2  Tesla V100-SXM2...  On   | 00000035:03:00.0 Off |                    0 |
| N/A   29C    P0    49W / 300W |    446MiB / 16160MiB |      8%      Default |
+-------------------------------+----------------------+----------------------+
|   3  Tesla V100-SXM2...  On   | 00000035:04:00.0 Off |                    0 |
| N/A   30C    P0    48W / 300W |    446MiB / 16160MiB |      9%      Default |
+-------------------------------+----------------------+----------------------+

+-----------------------------------------------------------------------------+
| Processes:                                                       GPU Memory |
|  GPU       PID   Type   Process name                             Usage      |
|=============================================================================|
|    0    132066      C   ...ild/example/advection/advection-example   435MiB |
|    1    132067      C   ...ild/example/advection/advection-example   435MiB |
|    2    132068      C   ...ild/example/advection/advection-example   435MiB |
|    3    132069      C   ...ild/example/advection/advection-example   435MiB |
+-----------------------------------------------------------------------------+
```

But with advection covergence:

```
41: Test command: /usr/gapps/parthenon_shared/parthenon-project/views/rzansel/ppc64le/gcc8/2021-01-04/bin/python3.8 "/g/g15/brown338/Software/parthenon/tst/regression/run_test.py" "--mpirun" "/usr/tcetmp/bin/jsrun" "--mpirun_opts=-a" "--mpirun_opts=4" "--mpirun_opts=-c 1 -n 1 -g 4 -r 1 -d packed --smpiargs='-gpu'" "--driver" "/g/g15/brown338/Software/parthenon/build/example/advection/advection-example" "--driver_input" "/g/g15/brown338/Software/parthenon/tst/regression/test_suites/advection_convergence/parthinput.advection" "--num_steps" "25" "--test_dir" "/g/g15/brown338/Software/parthenon/tst/regression/test_suites/advection_convergence" "--output_dir" "/g/g15/brown338/Software/parthenon/build/tst/regression/outputs/advection_convergence_mpi" "--kokkos_args=--kokkos-num-devices=4 --kokkos-threads=1"
41: Test timeout computed to be: 1500
41: 
41: 
41: test_dir=['/g/g15/brown338/Software/parthenon/tst/regression/test_suites/advection_convergence']
41: output_dir='/g/g15/brown338/Software/parthenon/build/tst/regression/outputs/advection_convergence_mpi'
41: driver=['/g/g15/brown338/Software/parthenon/build/example/advection/advection-example']
41: driver_input=['/g/g15/brown338/Software/parthenon/tst/regression/test_suites/advection_convergence/parthinput.advection']
41: kokkos_args=['--kokkos-num-devices=4 --kokkos-threads=1']
41: num_steps=25
41: mpirun=['/usr/tcetmp/bin/jsrun']
41: mpirun_opts=['-a', '4', "-c 1 -n 1 -g 4 -r 1 -d packed --smpiargs='-gpu'"]
41: coverage=False
41: *****************************************************************
41: Beginning Python regression testing script
41: *****************************************************************
41: 
41: Initializing Test Case
41: Using:
41: driver at:       /g/g15/brown338/Software/parthenon/build/example/advection/advection-example
41: driver input at: /g/g15/brown338/Software/parthenon/tst/regression/test_suites/advection_convergence/parthinput.advection
41: test folder:     /g/g15/brown338/Software/parthenon/tst/regression/test_suites/advection_convergence
41: output sent to:  /g/g15/brown338/Software/parthenon/build/tst/regression/outputs/advection_convergence_mpi
41: 
41: Make output folder in test if does not exist
41: *****************************************************************
41: Preparing Test Case Step 1
41: *****************************************************************
41: 
41: *****************************************************************
41: Running Driver
41: *****************************************************************
41: 
41: Command to execute driver
41: /usr/tcetmp/bin/jsrun -a 4 -c 1 -n 1 -g 4 -r 1 -d packed --smpiargs='-gpu' /g/g15/brown338/Software/parthenon/build/example/advection/advection-example -i /g/g15/brown338/Software/parthenon/tst/regression/test_suites/advection_convergence/parthinput.advection parthenon/mesh/nx1=32 parthenon/meshblock/nx1=32 parthenon/mesh/nx2=1 parthenon/meshblock/nx2=1 parthenon/mesh/nx3=1 parthenon/meshblock/nx3=1 Advection/vy=0.0 Advection/vz=0.0 --kokkos-num-devices=4 --kokkos-threads=1
41: 
41: *****************************************************************
41: Subprocess error message
41: *****************************************************************
41: 
41: b'### PARTHENON ERROR
41:   Message:     ### FATAL ERROR in Mesh constructor
41: Too few mesh blocks: nbtotal (1) < nranks (4)
41: 
41:   File:        ../src/mesh/mesh.cpp
41:   Line number: 450
41: ### PARTHENON ERROR
41:   Message:     ### FATAL ERROR in Mesh constructor
41: Too few mesh blocks: nbtotal (1) < nranks (4)
41: 
41:   File:        ../src/mesh/mesh.cpp
41:   Line number: 450
41: ### PARTHENON ERROR
41:   Message:     ### FATAL ERROR in Mesh constructor
41: Too few mesh blocks: nbtotal (1) < nranks (4)
41: 
41:   File:        ../src/mesh/mesh.cpp
41:   Line number: 450
41: ### PARTHENON ERROR
41:   Message:     ### FATAL ERROR in Mesh constructor
41: Too few mesh blocks: nbtotal (1) < nranks (4)
41: 
41:   File:        ../src/mesh/mesh.cpp
41:   Line number: 450
41: '
41: 
41: *****************************************************************
41: Error detected while running subprocess command
41: *****************************************************************
41: 
41: Traceback (most recent call last):
41:   File "/g/g15/brown338/Software/parthenon/tst/regression/utils/test_case.py", line 235, in Run
41:     proc = subprocess.run(run_command, check=True, stdout=PIPE, stderr=PIPE)
41:   File "/usr/gapps/parthenon_shared/parthenon-project/views/rzansel/ppc64le/gcc8/2021-01-04/lib/python3.8/subprocess.py", line 512, in run
41:     raise CalledProcessError(retcode, process.args,
41: subprocess.CalledProcessError: Command '['/usr/tcetmp/bin/jsrun', '-a', '4', '-c', '1', '-n', '1', '-g', '4', '-r', '1', '-d', 'packed', "--smpiargs='-gpu'", '/g/g15/brown338/Software/parthenon/build/example/advection/advection-example', '-i', '/g/g15/brown338/Software/parthenon/tst/regression/test_suites/advection_convergence/parthinput.advection', 'parthenon/mesh/nx1=32', 'parthenon/meshblock/nx1=32', 'parthenon/mesh/nx2=1', 'parthenon/meshblock/nx2=1', 'parthenon/mesh/nx3=1', 'parthenon/meshblock/nx3=1', 'Advection/vy=0.0', 'Advection/vz=0.0', '--kokkos-num-devices=4', '--kokkos-threads=1']' returned non-zero exit status 134.
41: 
41: During handling of the above exception, another exception occurred:
41: 
41: Traceback (most recent call last):
41:   File "/g/g15/brown338/Software/parthenon/tst/regression/run_test.py", line 152, in <module>
41:     main(**vars(args))
41:   File "/g/g15/brown338/Software/parthenon/tst/regression/run_test.py", line 76, in main
41:     test_manager.Run()
41:   File "/g/g15/brown338/Software/parthenon/tst/regression/utils/test_case.py", line 245, in Run
41:     raise TestManagerError('\nReturn code {0} from command \'{1}\''
41: utils.test_case.TestManagerError: 
41: Return code 134 from command '/usr/tcetmp/bin/jsrun -a 4 -c 1 -n 1 -g 4 -r 1 -d packed --smpiargs='-gpu' /g/g15/brown338/Software/parthenon/build/example/advection/advection-example -i /g/g15/brown338/Software/parthenon/tst/regression/test_suites/advection_convergence/parthinput.advection parthenon/mesh/nx1=32 parthenon/meshblock/nx1=32 parthenon/mesh/nx2=1 parthenon/meshblock/nx2=1 parthenon/mesh/nx3=1 parthenon/meshblock/nx3=1 Advection/vy=0.0 Advection/vz=0.0 --kokkos-num-devices=4 --kokkos-threads=1'
1/1 Test #41: regression_mpi_test:advection_convergence ...***Failed   68.48 sec

```

pgrete:
Yes, I also noticed that while updating/testing on Summit following my previous comment.
In that process I also noticed that the `advection_performance` test, in fact, should run with more MPI processes as I originally added the logic to make the MeshBlocks smaller based on the number of MPI processes involved.
That now makes me believe that the logic behind `parameters.num_ranks` is broken, i.e., it may always be 1.
This led me to  `test_case.py` where
```python
        argstrings = ['-np','-n']
        if len(set(argstrings) & set(self.parameters.mpi_opts)) > 1:
          print('Warning! You have set both "-n" and "-np" in your MPI options.')
          print(self.parameters.mpi_opts)
        for s in argstrings:
          if s in self.parameters.mpi_opts:
            index = self.parameters.mpi_opts.index(s)
            if index < len(self.parameters.mpi_opts) - 1:
              try:
                self.parameters.num_ranks = int(self.parameters.mpi_opts[index+1])
              except ValueError:
                pass
```
so I know wonder if that logic still works (in general and specifically for the `jsrun` based parameters used on Summit.
I'm going to call it a day but if you like to dig deeper today, my best bet is somewhere around these pieces.
Otherwise, I'll dig deeper tomorrow.

Bottom line: There's definitely a bug somewhere that resulted in the things not working as we expected them to be.

JoshuaSBrown:
Alright this should be good to go.

JoshuaSBrown:
All tests pass.

docs/building.md
```diff
@@ -385,6 +385,69 @@ Once you've configured your build directory, you can build with
...
 https://re-git.lanl.gov/eap-oss/parthenon-project
 
+### LNLL RZAnsel (Homogeneous)
+
+Last verified 04 Jan 2021.
+
+#### Allocate Node
+
+[RZAnsel](https://hpc.llnl.gov/hardware/platforms/rzansel) is a homogeneous cluster consisting of 2,376 nodes with the IBM Power9
+architecture with 44 nodes per core and 4 Nvidia Volta GPUs per node. To
```
jlippuner, line 395:
```suggestion
architecture with 44 cores per node and 4 Nvidia Volta GPUs per node. To
```

docs/building.md
```diff
@@ -385,6 +385,69 @@ Once you've configured your build directory, you can build with
...
+allocate an interactive node: 
+
+E.g.
+```bash
+$ lalloc 1
+```
+
+#### Set-Up Environment (Optional, but Still Recommended, for Non-CUDA Builds)
```
jlippuner, line 403: I am confused by this (also for the Darwin instructions).

Are these whole instructions for non-CUDA builds? If so, what are the instructions for CUDA builds? Or is it only optional (but still recommended) for non-CUDA builds but required for CUDA builds?

docs/building.md
```diff
@@ -385,6 +385,69 @@ Once you've configured your build directory, you can build with
 LANL Employees - to understand how the project space is built out, see
 https://re-git.lanl.gov/eap-oss/parthenon-project
 
+### LNLL RZAnsel (Homogeneous)
+
+Last verified 04 Jan 2021.
```
jlippuner, line 390: Is this up-to-date?

docs/building.md
jlippuner: Alright, the last sentence answers this question, I think.

Maybe we should call this section `Set-Up Environment (required for CUDA builds, optional (but recommended) for non-CUDA builds`.

jlippuner (approved):
Nice work! I just tested it on RZAnsel and it seems to work as expected.

I made a few minor comments/suggestions but nothing to hold up merging.

docs/building.md
JoshuaSBrown: I'll fix these in a separate PR then.

docs/building.md
JoshuaSBrown: @AndrewGaspar, @jlippuner raises a good point why isn't the build configuration simply optional, as far as I can tell there is nothing in there but links to ninja, cmake, the compiler, and git. Is it because of the dependence of the cuda on the compiler?

PR #381, opened 2020-11-28
Features for downstream

jdolence:
## PR Summary

Two main improvements:

1. Adding a new overload to `DataCollection::Add` that allows new `MeshBlockData` or `MeshData` objects to be created from a base object, but including only specific named variables.
2. Add a slightly more general `SumData` function in `Update`

Also changed a lot of the API to `T *` instead of `std::shared_ptr<T> &`.  `const T *` would have been better, but the caching we do for packing prevents the usage of `const`.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


src/interface/mesh_data.hpp
```diff
@@ -79,14 +79,14 @@ class MeshData {
   MeshData() = default;
 
   Mesh *GetMeshPointer() const { return pmy_mesh_; }
-  auto GetParentPointer() { return GetMeshPointer(); }
+  auto GetParentPointer() const { return GetMeshPointer(); }
```
JoshuaSBrown, line 82: 👍🏻

src/interface/mesh_data.hpp
```diff
@@ -79,14 +79,14 @@ class MeshData {
...
 
   void SetMeshPointer(Mesh *pmesh) { pmy_mesh_ = pmesh; }
   void SetMeshPointer(const std::shared_ptr<MeshData<T>> &other) {
     pmy_mesh_ = other->GetMeshPointer();
   }
 
-  void SetAllowedDt(const Real dt) {
+  void SetAllowedDt(const Real dt) const {
```
JoshuaSBrown, line 89: 👍🏻

src/interface/meshblock_data.hpp
```diff
@@ -60,14 +60,14 @@ class MeshBlockData {
...
+  std::shared_ptr<MeshBlock> GetBlockPointer() const {
     if (pmy_block.expired()) {
       PARTHENON_THROW("Invalid pointer to MeshBlock!");
     }
     return pmy_block.lock();
   }
-  auto GetParentPointer() { return GetBlockPointer(); }
-  void SetAllowedDt(const Real dt) { GetBlockPointer()->SetAllowedDt(dt); }
+  auto GetParentPointer() const { return GetBlockPointer(); }
+  void SetAllowedDt(const Real dt) const { GetBlockPointer()->SetAllowedDt(dt); }
```
JoshuaSBrown, line 70: 👍🏻

src/interface/meshblock_data.hpp
```diff
@@ -197,7 +230,7 @@ class MeshBlockData {
   //
   const CellVariableVector<T> &GetCellVariableVector() const { return varVector_; }
   const MapToCellVars<T> &GetCellVariableMap() const { return varMap_; }
-  CellVariable<T> &Get(std::string label) {
+  CellVariable<T> &Get(std::string label) const {
```
JoshuaSBrown: label can be made const reference.

src/interface/meshblock_data.hpp
```diff
@@ -206,7 +239,7 @@ class MeshBlockData {
     return *(it->second);
   }
 
-  CellVariable<T> &Get(const int index) { return *(varVector_[index]); }
+  CellVariable<T> &Get(const int index) const { return *(varVector_[index]); }
```
JoshuaSBrown, line 242: 👍🏻

src/interface/update.hpp
```diff
@@ -34,48 +34,50 @@ namespace parthenon {
...
-TaskStatus FluxDivergence(std::shared_ptr<T> &in, std::shared_ptr<T> &dudt_obj);
+TaskStatus FluxDivergence(T *in, T *dudt_obj);
 
 template <typename F, typename T>
-void UpdateData(const std::vector<F> &flags, T &in, T &dudt, const Real dt, T &out) {
-  const auto &in_pack = in->PackVariables(flags);
-  auto out_pack = out->PackVariables(flags);
-  const auto &dudt_pack = dudt->PackVariables(flags);
+TaskStatus SumData(const std::vector<F> &flags, T *in1, T *in2, const Real w1,
+                   const Real w2, T *out) {
```
JoshuaSBrown: I think that in1 and in2 could be made const, though that would require adding const to the methods in the meshblock_data class and what ever other class is passed in. The PackVariables methods that I have seen are creating a copy and returning it so I don't really see why this can't be done.

jdolence: It's possible I'm misinterpreting something, but my belief is that the `PackVariables` calls don't work with a `const T *` because the caching modifies the internal state of the object.  So the `PackVariables` calls can't be marked `const`.  I _think_ you could use `T* const in1`, for example, since that just implies the pointer can't be changed, rather than implying that the object pointed to can't be changed.  I'm not sure marking it `T* const` is very helpful, though.

tst/unit/test_data_collection.cpp
```diff
@@ -0,0 +1,118 @@
...
+    }
+    AND_WHEN("We want only a subset of variables in a new MeshBlockData") {
+      // reset vars
+      par_for(
+          loop_pattern_flatrange_tag, "init vars", DevExecSpace(), 0, 0,
+          KOKKOS_LAMBDA(const int i) {
+            v1(0) = 111;
+            v2(0) = 222;
+            v3(0) = 333;
+          });
```
JoshuaSBrown: Is this reset necessary? 

Ok, never mind I see, the second variable is changed, but then why are you resetting v1 and v3?

src/interface/mesh_data.hpp
```diff
@@ -118,15 +118,16 @@ class MeshData {
     }
   }
 
-  void Copy(const std::shared_ptr<MeshData<T>> src) {
+  template <typename... Args>
+  void Copy(const std::shared_ptr<MeshData<T>> src, Args &&... args) {
     const int nblocks = src->NumBlocks();
```
JoshuaSBrown, line 126: Shared pointers do not guarantee you you are not pointing to null. It would be safer to make sure that src is valid before calling it's method.

JoshuaSBrown (changes requested):
This is great, an improvement in const correctness and a nice unit test. I would also like to see a unit test for the Sum method, then I will be happy.

jdolence:
> This is great, an improvement in const correctness and a nice unit test. I would also like to see a unit test for the Sum method, then I will be happy.

Thanks.  I'll address the other comments, but I'd push back on the unit test for `SumData`.  The advection tests already make use of `SumData` in several forms and I'd say provide adequate coverage of the method until/unless a problem is encountered that motivates a specific unit test.  A unit test of the method as written would be quite painful to produce, I think, because it's fairly interwoven with other pieces of machinery.  What do you think @JoshuaSBrown?

jdolence:
Just a note to mention that the full pipeline was run and passes.

JoshuaSBrown:
> > This is great, an improvement in const correctness and a nice unit test. I would also like to see a unit test for the Sum method, then I will be happy.
> 
> Thanks. I'll address the other comments, but I'd push back on the unit test for `SumData`. The advection tests already make use of `SumData` in several forms and I'd say provide adequate coverage of the method until/unless a problem is encountered that motivates a specific unit test. A unit test of the method as written would be quite painful to produce, I think, because it's fairly interwoven with other pieces of machinery. What do you think @JoshuaSBrown?

There are advantageous to writing unit tests other than code coverage, it typically helps with the development of class interface. However, in this case I take your point. It's not worth it.

src/interface/update.hpp
JoshuaSBrown: Oh, I see what you mean, that's unfortunate, I skimmed over the helper methods in meshblock_data. Generally I think it is good to use const liberally even if it is only fixing the pointer address, but like you said in this instance the benefits are... minimal.

JoshuaSBrown (approved):

docs/interface/state.md
```diff
@@ -60,6 +60,8 @@ The ```MeshBlockData``` class provides a means of organizing and accessing simul
...
+
+The overload `Add(const std::string &label, MeshBlockData<T> &src, const std::vector<std::string> &names)` provides the same functionality as the above `Add` function, but for a subset of variables provided in the vector of names.  This feature allows downstream applications to allocate storage in a more targeted fashion, as might be desirable to hold source terms for particular equations, for example.
```
Yurlungur, line 65: This is great. A much-needed feature. :+1:

src/interface/update.hpp
```diff
@@ -34,48 +34,50 @@ namespace parthenon {
...
 template <typename T>
-TaskStatus FluxDivergence(std::shared_ptr<T> &in, std::shared_ptr<T> &dudt_obj);
+TaskStatus FluxDivergence(T *in, T *dudt_obj);
 
 template <typename F, typename T>
-void UpdateData(const std::vector<F> &flags, T &in, T &dudt, const Real dt, T &out) {
-  const auto &in_pack = in->PackVariables(flags);
-  auto out_pack = out->PackVariables(flags);
-  const auto &dudt_pack = dudt->PackVariables(flags);
+TaskStatus SumData(const std::vector<F> &flags, T *in1, T *in2, const Real w1,
```
Yurlungur: I like this. This is more general and more useful.

Nitpick on the name: This isn't really a sum, it's more like a fuse-multiply-add. How about `TaskStatus FMAData` or something similar?

Also should `w1` and `w2` default to `1.0`?

Yurlungur (approved):
Thanks for this. This is broadly good cleanup and useful changes.

AndrewGaspar:
> Also changed a lot of the API to T * instead of std::shared_ptr<T> &. const T * would have been better, but the caching we do for packing prevents the usage of const.

GOOD

src/interface/update.hpp
jdolence: It's not really a FMA either, is it?  How about `TaskStatus WeightedSumData`?  I think I like the idea to default the weights to 1...

jdolence: @Yurlungur pushed a change here.   Look OK?

Yurlungur: :+1: Yep I like it.

jdolence:
@pgrete, this PR has breaking changes so I want to make sure you see them before this is merged.  Should be a quick review, I think.

src/interface/mesh_data.hpp
pgrete: Wouldn't `if (src.get() == nullptr) {` be more explicit?

src/interface/mesh_data.hpp
```diff
@@ -171,6 +175,13 @@ class MeshData {
     return true;
   }
 
+  bool Contains(const std::vector<std::string> &names) const {
+    for (const auto &b : block_data_) {
+      if (!b->Contains(names)) false;
```
pgrete: this is missing a return statement isn't it?

src/interface/meshblock_data.hpp
```diff
@@ -79,6 +79,39 @@ class MeshBlockData {
...
+          Add(v->AllocateCopy());
+        }
+        found = true;
+      }
+      if (!found) {
+        PARTHENON_THROW("MeshBlockData::Copy did not find " + name);
+      }
+    }
+  }
+
```
pgrete, line 114: What's our current policy with respect to having the same name being used for both a cell variables as well as a SparseVariable?
If that's possible the error handling logic may need an update.

pgrete (changes requested):
Looks good! I only have some minor comments and I think the missing return is actually a bug. that would result in that function always returning true.

src/interface/meshblock_data.hpp
jdolence: The intent, I believe, is to disallow any two variables from having the same name, sparse or otherwise.

src/interface/mesh_data.hpp
jdolence: haha.  nice catch.

jdolence:
fixed those issues @pgrete.  thanks for the review (and catching that bug!).  Mind approving so I can merge once the tests pass?

jdolence:
Given the nature of @pgrete's comments and the fact that I've addressed them, I'm just going to pull the trigger.  Thanks for the helpful reviews all!

PR #383, opened 2020-11-30
WIP: Operationalize Metadata::Sparse for single processor runs

AndrewGaspar:
## PR Summary

Just sharing - currently doesn't work.

TODO
- [x] Add `InvalidSparseID` constant
- [x] Get rid of `MeshBlockDataIterator`
- [ ] Unify `MeshBlockData::Copy` and subcontainer constructors
- [ ] Compare `SparseInfo` in HDF5 diff script

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted


src/interface/meshblock_data.hpp
```diff
@@ -397,24 +397,22 @@ class MeshBlockData {
...
-  vpack_types::VarList<T> MakeList_(std::vector<std::string> &names);
+  struct VariableListResult {
+    vpack_types::VarList<T> vars;
+    std::vector<std::string> expanded_names;
+  };
+
+  VariableListResult GetVariablesByName(const std::vector<std::string> &names,
+                                        const std::vector<int> &sparse_ids = {});
+  VariableListResult GetVariablesByFlag(const std::vector<MetadataFlag> &flags);
+  VariableListResult GetAllVariables();
```
Yurlungur: This is an improvement!

src/interface/variable_pack.hpp
```diff
@@ -56,6 +57,7 @@ class PackIndexMap {
   void insert(std::pair<std::string, vpack_types::IndexPair> keyval) {
     map_[keyval.first] = keyval.second;
   }
+  bool Has(std::string const &key) const { return map_.count(key) > 0; }
```
Yurlungur, line 120: This is also an improvement.

src/interface/variable_pack.hpp
```diff
@@ -56,6 +57,7 @@ class PackIndexMap {
...
 
  private:
   std::map<std::string, vpack_types::IndexPair> map_;
```
Yurlungur: With the above change that uses `map_.count`, should this be an unordered map so the count call is constant time?

src/interface/meshblock_data.cpp
```diff
@@ -234,25 +234,25 @@ std::shared_ptr<MeshBlockData<T>> MeshBlockData<T>::SparseSlice(int id) {
 ///        indices represent inclusive bounds for, e.g., a sparse or tensor-valued
 ///        variable.
 template <typename T>
-VariableFluxPack<T> MeshBlockData<T>::PackVariablesAndFluxesHelper_(
-    const std::vector<std::string> &var_names, const std::vector<std::string> &flx_names,
-    const vpack_types::VarList<T> &vars, const vpack_types::VarList<T> &fvars,
-    PackIndexMap &vmap, vpack_types::StringPair &key) {
-  key = std::make_pair(var_names, flx_names);
-  auto kvpair = varFluxPackMap_.find(key);
+PackVariablesAndFluxesResult<T>
```
Yurlungur: What's the motivation for returning the map by value instead of by reference? This seems equivalent to me?

src/interface/meshblock_data.cpp
```diff
@@ -338,25 +342,24 @@ MeshBlockData<T>::PackVariablesAndFluxes(const std::vector<MetadataFlag> &flags)
...
 ///        variable.
 template <typename T>
-VariablePack<T>
-MeshBlockData<T>::PackVariablesHelper_(const std::vector<std::string> &names,
-                                       const vpack_types::VarList<T> &vars,
-                                       PackIndexMap &vmap) {
-  // this->GetSparseMap
-  auto kvpair = varPackMap_.find(names);
+PackVariablesResult<T>
+MeshBlockData<T>::PackListedVariables(VariableListResult &&variables) {
```
Yurlungur: Changing the naming here and removing the "Helper" functions seems to make this a more public API. Is that intentional?

src/interface/variable_pack.hpp
```diff
@@ -272,45 +337,101 @@ VariableFluxPack<T> MakeFluxPack(const vpack_types::VarList<T> &vars,
...
+};
+
+template <typename T>
+using PackVariablesResult =
+    PackVariablesResultGeneric<VariablePack<T>, std::vector<std::string>>;
+
+template <typename T>
+using PackVariablesAndFluxesResult = PackVariablesResultGeneric<
+    VariableFluxPack<T>, std::pair<std::vector<std::string>, std::vector<std::string>>>;
 } // namespace parthenon
```
Yurlungur: Can you explain in a little more detail what the intent of this is? These class names imply verbage, but they seem to be just containers?

example/calculate_pi/calculate_pi.cpp
```diff
@@ -37,11 +37,57 @@ namespace calculate_pi {
...
+    }
+
+    // If part of the block falls inside and other parts fall outside, then we need to
+    // expand the variable before computing on it. If it's wholly inside or outside, then
+    // there's no need for expansion.
+
+    // TODO(agaspar): Change this to only expand variables that fall on the radius.
+    // Currently we're expanding all variables that fall outside the circle at all. if
+    // (!inside || !outside) {
+    if (inside && !outside) {
```
jlippuner, line 74: I don't quite understand the logic here. The comment above (not TODO) says that we extend if there are corners inside and outside, but it seems that we would also extend if all the corners are outside with this logic

src/interface/meshblock_data.cpp
jlippuner: I think the motivation is to avoid output variables. If I remember correctly, the Google style guide says to never use non-const references, instead to use pointers when a function changes its argument. This function really has 3 outputs: the VariableFluxPack, the PackIndexMap, and the key. One is returned as a return value, the others are set as output variables passed by reference. This is stylistically not very elegant, so Andrew introduced the `PackVariablesAndFluxesResult` type that combines all 3 outputs into one struct that can be returned, thus removing the need for output variables.

src/interface/meshblock_data.cpp
jlippuner: It's declared private, so it's not intended to be a public API. But at the same time, the `Helper_` suffix seems unnecessary to me.

src/interface/variable_pack.hpp
jlippuner: That's a good point. It doesn't seem like we iterate over all elements. We only insert and access elements, which are all O(log(N)) for `map` and average O(1), worst O(N) for `unordered_map`. I'll change it to `unordered_map`.

src/interface/variable_pack.hpp
jlippuner: They are just containers. As mentioned in a previous comment, the intent is to gather 3 different output values into one struct so they can all be returned from the function instead of returning one and setting the other two as non-const reference output variables.

src/refinement/amr_criteria.cpp
```diff
@@ -55,7 +55,12 @@ AMRFirstDerivative::AMRFirstDerivative(ParameterInput *pin, std::string &block_n
 }
 
 AmrTag AMRFirstDerivative::operator()(std::shared_ptr<MeshBlockData<Real>> &rc) {
-  ParArrayND<Real> q = rc->Get(field).data;
+  ParArrayND<Real> q;
+  try {
+    q = rc->Get(field).data;
+  } catch (std::invalid_argument const &) {
+    return AmrTag::derefine;
```
jlippuner: Not sure if this is what we want. We probably want `AmrTag::same`. @Yurlungur thoughts?

jlippuner (commented):
Taking over from @AndrewGaspar. Thanks @Yurlungur for your initial pass a while ago. I tried to reply as best as I could to your comments.

src/interface/meshblock_data.cpp
Yurlungur: I don't really agree that returning a massive struct by value is stylistically better than modifying input variables by reference. Personally, I find this uglier, but I'm willing to be over-ruled.

One bigger question though. Which version is more performant?

src/interface/variable_pack.hpp
Yurlungur: Ah, I think I understand. I got confused by the names starting with `Pack` could we invert the naming order to something like `ResultForPackVariables` and `ResultForPackVariablesAndFluxes`?

src/refinement/amr_criteria.cpp
Yurlungur: @jdolence wrote this so he may have more insight. But I am inclined to agree. I think this should be `AmrTag::same`.

jlippuner:
Retry DeepCode

jlippuner:
Deprecated by new sparse interface

PR #386, opened 2020-12-02
Handle variable name conflicts between packages

Yurlungur:
## PR Summary

This WIP PR is intended to help resolve issues that might arise between packages. This was motivated by an internal discussion between myself, @jdolence , @nmsriram , and Chad Meyer. 

### In particular, @jdolence argued that:

1. When two packages request the same variable, parthenon actually adds it twice. This may be a problem if two packages add a variable of the same name but with different metadata.
2. Following that last comment, there needs to be some kind of application level logic that uniquely determines which variables/metadata agree the right ones to actually instantiate. 

### I suggest the following solution, which I am working on implementing here:

When we add variables to a StateDescriptor in parthenon, each package has its own. But there is a reconciliation step, when we loop through packages and add variables. We add the following metadata flags:
- `Private`: private metadata means a variable is package specific. The package name is prepended to the variable name, like a namespace. Two different variables can have the same name if they live in different packages.
- `Provides`: variables with this metadata are shared, and demand they be the only ones to provide a variable. The metadata of these variables takes priority. If two packages provide the same variable with the Provides metadata, an error is raised at runtime. Provides is on by default unless overriden by another flag.
- `Requires`: variables with this metadata flag are not added by a package. The package simply declares it expects the variable to exist and if it doesn't, an error is raised.
- `Overridable`: Variables with this metadata flag are provided by the package that registers them unless another package provides said variable (with the Provides flag), in which case, that other package determines what happens to that variable. If two packages request an `Overridable` variable, but it is not provided, it's undefined behaviour and Parthenon warns as such.

To enforce this, we add a validation step between package initialization and variable allocation. Variables are allocated from the state descriptor here in MeshBlock. We can easily insert a validation step right above this that loops through packages and resolves conflicts as per the above metadata hints.

I also add a mechanism for spitting out which variables are active and with what metadata and saving this to file.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


Yurlungur:
This is now ready for review.

### I have made the following changes:

- The metadata flags and resolution machinery have been added as described above. This ended up being quite involved and is implemented in `src/interface/state_descriptor.cpp` They are now tested in `tst/unit/test_state_descriptor.cpp`. Feedback on this implementation would be welcome.
- Because this machinery is redundant with the `DerivedOwnership` machinery @jdolence put in a while ago, I removed `DerivedOwnership`.
- I added a type alias `Dictionary<T> = std::unordered_map<std::string,T>` which shows up in many places throughout the code
- I changed the `std::map`s in `state_descriptor` to `std::unordered_map`. This should be marginally more performant but otherwise is equivalent.
- The logic for `AddField` for sparse data is vastly simplified by making `sparseMetadataMap_` a `Dictionary<std::unordered_map<int,Metadata>>` so I changed it.
- To ease debugging, I have added the capability to print out all metadata flags for each variable. The code now does so at mesh construction time.

@jdolence @nmsriram , Chad Meyer, you may be interested in this. I'm also interested to hear what the athena team, @pgrete @forrestglines @jmstone thinks.

Finally, @AndrewGaspar if you could take a peek at the changes I made to `Metadata`, I would appreciate a sanity check on the changes I made there.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 - [[PR 358]](https://github.com/lanl/parthenon/pull/358) Generalize code that interfaces with downstream apps to work with both `MeshData` and `MeshBlockData`.
 - [[PR 335]](https://github.com/lanl/parthenon/pull/335) Support for project-relative `MACHINE_CFG` with `@PAR_ROOT@`
 - [[PR 328]](https://github.com/lanl/parthenon/pull/328) New `MeshBlock` packing interface using `DataCollection`s of `MeshData` and `MeshBlockData`.
+- [[PR 386]](https://github.com/lanl/parthenon/pull/386) New machinery in `StateDescriptor` to reolve conflicts between variables declared in different packages
```
AndrewGaspar:
```suggestion
- [[PR 386]](https://github.com/lanl/parthenon/pull/386) New machinery in `StateDescriptor` to resolve conflicts between variables declared in different packages
```

src/interface/metadata.hpp
```diff
@@ -124,6 +131,12 @@ class MetadataFlag {
   constexpr int InternalFlagValue() const { return flag_; }
 #endif
 
+  friend std::ostream &operator<<(std::ostream &os, const Metadata &m);
+  friend std::ostream &operator<<(std::ostream &os, const MetadataFlag &flag) {
```
AndrewGaspar, line 135: TIL you can have inline friend functions.

src/interface/metadata.hpp
```diff
@@ -172,18 +185,21 @@ class Metadata {
   explicit Metadata(const std::vector<MetadataFlag> &bits, const int sparse_id)
       : shape_({1}), sparse_id_(sparse_id) {
     SetMultiple(bits);
+    // If you pass in a sparse ID, the Sparse metadata flag is automatically added
+    Set(Sparse);
```
AndrewGaspar: I get the motivation here. However, given that I intend to kill sparse IDs in the metadata, I'd prefer you don't make this change since it would encourage people to not mark their sparse variables as sparse. Means there are more places for me to potentially make mistakes (adding `Sparse` where it's missing).

src/interface/metadata.hpp
```diff
@@ -257,7 +288,7 @@ class Metadata {
   }
 
   // Operators
-  bool operator==(const Metadata &b) const {
+  bool HasSameBits(const Metadata &b) const {
```
AndrewGaspar:
```suggestion
  bool HasSameFlags(const Metadata &b) const {
```

src/interface/metadata.hpp
```diff
@@ -276,8 +307,15 @@ class Metadata {
         return false;
       }
     }
+    return true;
+  }
 
-    return std::tie(a.shape_, a.sparse_id_) == std::tie(b.shape_, b.sparse_id_);
+  bool SparseEqual(const Metadata &b) const {
+    return (HasSameBits(b) && std::equal(shape_.begin(), shape_.end(), b.shape_.begin()));
```
AndrewGaspar: Why not the terser:

```suggestion
    return HasSameBits(b) && shape_ == b.shape_;
```

src/interface/state_descriptor.cpp
```diff
@@ -0,0 +1,347 @@
...
+      PARTHENON_THROW("Unknown dependency");
+    }
+  }
+
+  template <typename Collection, typename AdderPrivate, typename AdderProvides>
+  void SortCollection(const std::string &package, const Collection &c,
+                      const AdderPrivate &add_private,
+                      const AdderProvides &add_provides) {
+    for (auto &pair : c) {
+      std::string var = pair.first;
```
AndrewGaspar:
```suggestion
      std::string const &var = pair.first;
```

src/interface/state_descriptor.cpp
```diff
@@ -0,0 +1,347 @@
...
+    for (auto &pair : packages) {
+      auto &package = pair.second;
+      if (package->SwarmPresent(swarm)) {
+        for (auto &pair : package->AllSwarmValues(swarm)) {
+          state->AddSwarmValue(pair.first, swarm, pair.second);
+        }
+        return;
+      }
+    }
+  };
```
AndrewGaspar: This is a typical use case for inheritance, which I think would be better than a loose set of lambda functions. It would also cut down on template instantiations.

You could create a`VariableProvider` interface that looks something like:
```c++
class VariableProvider {
  friend class DependencyTracker;

  virtual void AddPrivateVariable(const string &package, const string &var, Metadata &metadata) = 0;
  virtual void AddProvidesVariable(const string &package, const string &var, Metadata &metadata) = 0;
  virtual void AddOverridableVariable(const string &package, const string &var) = 0;
};
```

You would would sublcass that for dense, sparse, and swarm variables. Then `SortCollection`, `Sort`, and `CheckOverridable` just take references to `VariableProvider` objects, and result in a single instantiation of those functions. Also better error messages. `SortCollection` will still take a template argument for the collection, but that's a relatively small function.

AndrewGaspar: With my PR, the different handling for dense vs. sparse variables should go away.

src/interface/state_descriptor.cpp
```diff
@@ -0,0 +1,347 @@
...
+  if (swarmValueMetadataMap_[swarm_name].count(value_name) > 0) {
+    throw std::invalid_argument("Swarm value " + value_name + " already exists!");
+  }
+  swarmValueMetadataMap_[swarm_name][value_name] = m;
+
+  return true;
+}
+
+bool StateDescriptor::AddField(const std::string &field_name, const Metadata &m_in) {
+  Metadata m = m_in; // horrible hack to force const correctness
```
AndrewGaspar: This comment needs elaboration. No idea what this "hack" does.

AndrewGaspar: Oh, because of the `Associate` calls below? Well, you could definitely reduce the scope of this "hack", but I'm honestly not really fully clear on the point of `Associate`. As far as I can see nobody ever calls `getAssociated` and does anything useful with it - it's only ever used as essentially a boolean to decide if a different name should be associated with the `Metadata`.

The fact that you've decided you need to make a copy of the `Metadata` at all because you don't want this function to mutate the `Metadata` argument proves its uselessness - I think the whole point of it originally was so that, after calling `AddField`, you'd know which field the metadata was called with (for what purpose, I'm unsure). I'd say just delete `Associate`, `getAssociated`, and `associated_`, and I think you'll be happier.

src/interface/state_descriptor.hpp
```diff
@@ -158,42 +121,69 @@ class StateDescriptor {
...
   Metadata &SwarmMetadata(const std::string &swarm_name) {
     return swarmMetadataMap_[swarm_name];
   }
 
-  // get all metadata for this physics
-  const std::map<std::string, Metadata> &AllMetadata() { return metadataMap_; }
-
-  bool FlagsPresent(std::vector<MetadataFlag> const &flags, bool matchAny = false) {
+  template <typename F>
+  void MetadataLoop(F func) {
```
AndrewGaspar: Make this private - be deliberate about adding new public functions.

CMakeLists.txt
```diff
@@ -18,6 +18,10 @@ include(cmake/MachineCfg.cmake)
 
 project(parthenon VERSION 0.3.0 LANGUAGES C CXX)
 
+if (${CMAKE_VERSION} VERSION_GREATER_EQUAL 3.19.0)
+  cmake_policy(SET CMP0110 NEW)
+endif()
```
AndrewGaspar, line 23: Setting policies is a no-op on CMake versions prior to the policy introduction.

```suggestion
# Explicitly enable support for test names with whitespace and special characters
cmake_policy(SET CMP0110 NEW)
```

CHANGELOG.md
AndrewGaspar: Wouldn't this be better describing the new capability? I.e. something about adding support for private fields and field interfaces (via provides, depends, and overrides)

Don't use that exact language, I just want this to emphasize what the new capability is for developers rather than emphasizing the implementation of it.

AndrewGaspar (changes requested):

CHANGELOG.md
Yurlungur: Yes this is a good point. :+1:

src/interface/metadata.hpp
Yurlungur: what about throwing an error instead?

src/interface/metadata.hpp
Yurlungur: Does that actually work? `shape_` is a `std::vector` does the comparator work for those?

CMakeLists.txt
Yurlungur: I attempted this and got cmake errors. It was not a no-op.

src/interface/state_descriptor.cpp
Yurlungur: Sure, this could be re-written as inheritence. Stylistically, I just prefer a functional programming style. I'll change it.

Yeah, I know your PR will clean a lot of this up. But I wanted to introduce the desired behaviour.

nmsriram:
Looks good @Yurlungur.  I like this metadata approach to solve the problem. 

Only suggestion is to perhaps use `Defers` instead of `Overridable`.

src/interface/state_descriptor.cpp
Yurlungur: We hashed this out on Mattermost. The implementation I have here is a bug. However, the logic in `AddField` for associated variables is untested in parthenon and used in riot, so we need tests here to catch this kind of bug. 

Also the current design of `AddField` means `Metadata` can't be passed in as a const reference. It must either be made non-const, or the logic in `Associate` needs to change slightly.

docs/interface/Metadata.md
```diff
@@ -6,6 +6,30 @@ Variables can be tagged with a variety of `MetadataFlag` values. These flags
...
+  live in different packages.
+- `Provides`: variables with this metadata are shared, and demand they
+  be the only ones to provide a variable. The metadata of these
+  variables takes priority. If two packages provide the same variable
+  with the Provides metadata, an error is raised at runtime. Provides
+  is on by default unless overriden by another flag.
+- `Requires`: variables with this metadata flag are not added by a
+  package. The package simply declares it expects the variable to
+  exist and if it doesn't, an error is raised.
+- `Overridable`: Variables with this metadata flag are provided by the
```
jdolence, line 26: So say a package adds a variable with this `Overridable` flag.  Would the implication be, at least sometimes, that the package would provide some means of filling in values?  Like a `FillDerived` function, for example?  If so, how does the package know whether to fill in values?  I guess it just checks for this flag?

src/interface/metadata.hpp
AndrewGaspar: Yeah, that would be fine.

src/interface/metadata.hpp
AndrewGaspar: https://en.cppreference.com/w/cpp/container/vector/operator_cmp

src/interface/state_descriptor.cpp
AndrewGaspar: Yeah, I was just letting you know that hopefully this will get less messy in the near term.

CMakeLists.txt
AndrewGaspar: hm... I did this as well, but with cmake 3.18 and didn't get an error. Maybe it wasn't a no-op on previous versions of cmake? I'll take your word for it.

src/interface/metadata.hpp
Yurlungur: Huh. Neat.

CMakeLists.txt
Yurlungur: yeah the version I tried it on was older... whatever version the CI is using is what gave me problems. That might be 3.12.0 or something.

src/interface/metadata.hpp
```diff
@@ -217,6 +233,20 @@ class Metadata {
     return None;
   }
 
+  MetadataFlag Dependency() const {
```
jdolence: can we change the name of this?  not sure to what.  to me this name suggests that it provides some means of having the current variable depend on other variables.

src/interface/state_descriptor.cpp
```diff
@@ -0,0 +1,347 @@
...
+          }
+        } else {
+          PARTHENON_THROW("Variable " + var + " provided by multiple packages");
+        }
+      }
+      if (is_sparse) sparse_added_provides[var][sparse_id] = true;
+      provided_vars.insert(var);
+      add_provides(package, var, metadata);
+    } else if (dependency == Metadata::Requires) {
+      depends_vars.insert(var);
```
jdolence, line 66: when a package depends on a variable, does it not also depend on at least some properties of it's metadata.  for example, a package may require something like "velocity" with shape vector(1,3), but some package may provide "velocity" as a scalar.  That's a bit contrived, but I think you see the point.  Maybe this is captured in some other way I haven't seen yet...

jdolence: just another quick note on this.  this may be a place where some notion of needing "at least" these properties might be useful.  btw, this could certainly be for later.

src/interface/state_descriptor.cpp
```diff
@@ -0,0 +1,347 @@
...
+    }
+  }
+
+  template <typename Adder>
+  void CheckOverridable(Adder add_to_state) {
+    std::unordered_set<std::string> cache;
+    for (auto &pair : overridable_vars) {
+      auto &var = pair.first;
+      auto &count = pair.second;
+      if (provided_vars.count(var) == 0) {
```
jdolence: Since you wrote the `Provided` function below, maybe it should be used for things like this?

src/interface/state_descriptor.cpp
```diff
@@ -0,0 +1,347 @@
...
+    }
+  }
+  return false;
+}
+
+void StateDescriptor::ValidateMetadata() {
+  MetadataLoop([&](Metadata &m) {
+    auto dependency = m.Dependency();
+    if (dependency == Metadata::None) {
+      m.Set(Metadata::Provides);
```
jdolence, line 282: So if you don't set one of the new flags, it defaults to `Provides`?  I think that seems correct.  The only reasonable alternative would be to require that a flag be provided and throw here if none is given.  Just thought I'd bring this up in case there's a good argument for one or the other.  I think I'd lean _slightly_ toward requiring that a flag be provided, just to force some explicitness and consistency in downstream apps, but it's only a slight lean.

src/interface/state_descriptor.cpp
```diff
@@ -0,0 +1,347 @@
...
+  std::unordered_set<std::string> provided_vars;
+  std::unordered_set<std::string> depends_vars;
+
+  Dictionary<int> overridable_vars;
+  Dictionary<std::unordered_map<int, bool>> sparse_added_provides;
+  Dictionary<std::unordered_map<int, bool>> sparse_added_depends;
+  Dictionary<std::vector<Metadata>> overridable_meta;
+
+  template <typename AdderPrivate, typename AdderProvides>
+  void Sort(const std::string &package, const std::string &var, const Metadata &metadata,
```
jdolence: Is this really sorting?

src/interface/state_descriptor.cpp
```diff
@@ -0,0 +1,347 @@
...
+      if (overridable_meta.at(var).size() == 1) {
+        overridable_vars[var] += 1; // using value initalization of ints = 0
+      }
+    } else {
+      PARTHENON_THROW("Unknown dependency");
+    }
+  }
+
+  template <typename Collection, typename AdderPrivate, typename AdderProvides>
+  void SortCollection(const std::string &package, const Collection &c,
```
jdolence: Same question.  Is sorting really what's going on here?

src/interface/state_descriptor.cpp
jdolence: That _might_ be more clear.  The lambdas are a little gross, but I guess it's not as clear to me that breaking this out into a new class will make it significantly better.  Certainly could.  One hesitation I'd have in suggesting you worry too much about this right now is that @AndrewGaspar's work on sparse will already clean up part of this, and then I'm not sure but some comments on an impending cleanup of `Swarm`/`SwarmVariables` might clean up that part.  Will the latter help here?  If so, then a lot of this boilerplate will go away anyway, even without adding a new class.  So I guess what I'm saying is that the best case is that this is cleaned up _and_ no new class is created.  Not sure what the prospects are for this, though.  Just a thought.

src/interface/state_descriptor.hpp
```diff
@@ -33,8 +35,6 @@ class MeshBlockData;
 template <typename T>
 class MeshData;
 
-enum class DerivedOwnership { shared, unique };
```
jdolence, line 36: glad we're finally dealing with this in a much more comprehensive way.

src/mesh/meshblock.cpp
```diff
@@ -143,38 +144,37 @@ void MeshBlock::Initialize(int igid, int ilid, LogicalLocation iloc,
   precon = std::make_unique<Reconstruction>(shared_from_this(), pin);
 
   // Add field properties data
+  // TOOD(JMM): Should packages be resolved for state descriptors in
+  // properties?
```
jdolence, line 148: Actually, I think we should just remove `properties`.  This is obviously not for this PR, but I doubt anyone will object.

jdolence (changes requested):

jdolence:
Thanks for putting this together @Yurlungur!  I'm really happy to see us finally address this issue!

docs/interface/Metadata.md
Yurlungur: Right exactly. Overridable means the package provides some means for filling in values and it has to check for this flag.

src/interface/state_descriptor.cpp
Yurlungur: Yeah this is a good point and it's not currently captured... I don't think it would be hard to add something. But expressing it in the right way might be kind of tricky. I say we massage this MR and merge it in and attack that question a bit later.

src/interface/metadata.hpp
Yurlungur: How about something like `Role`?

src/interface/state_descriptor.cpp
Yurlungur: Yeah I went with automatically being `Provides` because that doesn't require downstream changes and because it seems to be the intuitively "right" thing. You're right, though that, enforcing explicitness has merit.

I don't mind changing this, but I'll leave it as is unless you feel strongly.

src/interface/state_descriptor.cpp
Yurlungur: It's sorting into categories. I guess a better name would be `Categorize`.

src/interface/state_descriptor.cpp
Yurlungur: It could simplify substantially, I agree. But I'm not sure the need for some flexibility will entirely disappear.

Yurlungur:
Thanks, @jdolence @nmsriram ! And thanks for your comments @AndrewGaspar ! I think I have addressed all your comments, @jdolence , @AndrewGaspar . Can you take another look?

Also can someone from the Athena team take a look? @pgrete @forrestglines ?

jdolence (approved):

AndrewGaspar (approved):

src/interface/state_descriptor.cpp
```diff
@@ -0,0 +1,379 @@
...
+      auto &var = pair.first;
+      auto &count = pair.second;
+      if (!Provided(var)) {
+        if (count > 1) {
+          std::stringstream ss;
+          ss << "Variable " << var
+             << " registered as overridable multiple times, but never provided."
+             << " This results in undefined behaviour as to which package will provide"
+             << " it." << std::endl;
+          PARTHENON_DEBUG_WARN(ss);
```
pgrete, line 120: Is this intentional, i.e., is there a "use case" for this?
I'm worried that this will just work for downstream codes and then at some point (e.g., when we change some defaults) either continue to work (for the right or wrong reasons) or stop working.
So unless there's a reason to only throw a warning I'm in favor of making this fail hard.

src/interface/state_descriptor.cpp
pgrete: Do we already have an issue for this? I didn't find one but based on this conversation I think we should have one.

tst/unit/test_state_descriptor.cpp
```diff
@@ -0,0 +1,270 @@
...
+        auto pkg3 = ResolvePackages(packages);
+        AND_THEN("The provides package is available") {
+          REQUIRE(pkg3->FieldPresent("dense"));
+          REQUIRE(pkg3->FieldMetadata("dense") == m_provides);
+        }
+      }
+    }
+
+    WHEN("We add an overridable variable and nothing else") {
+      pkg1->AddField("dense", m_overridable);
```
pgrete, line 209: What is actually happening here?
In principle we already added "dense" with different MetaData above or am I missing where the packages are cleared?

pgrete (commented):
Thanks for the changes and the thorough testing implementation @Yurlungur 
This is definitely an improvement over the previous (not really existent) handling.

I left a few minor comments that in principle should not hold off merging but I'd like to get some feedback to.

src/interface/state_descriptor.cpp
Yurlungur: I fixed it in this PR.

Yurlungur: Ah but I didn't remove the comment "horrible hack." I'll change that.

src/interface/state_descriptor.cpp
Yurlungur: This is intentional. My thinking is that, for example, three packages might all want, say, Pressure but all register it as overridable. It feels to me like that should run, but you need to pick one to provide. Only one of them should provide it and there's no way to automatically pick. That's why the warning.

We could change it to an error, that would certainly enforce being more deliberate. But it would make it harder to "flawlessly" turn off some package which provides.

tst/unit/test_state_descriptor.cpp
Yurlungur: `catch2` automatically clears everything between `WHEN`s. Basically `GIVEN` is always called around each new `WHEN` block.

src/interface/state_descriptor.cpp
Yurlungur: Also, this does break `associate` in downstream codes, if you're using it. But it's an easy fix.

tst/unit/test_state_descriptor.cpp
pgrete: That's for the clarification. That's good to know!

src/interface/state_descriptor.cpp
pgrete: I didn't consider it from that angle yet.
I guess my main concern is related to #396 in the sense that, if everyone overrides, multiple different definitions may float around and it's not guaranteed which one is picked.

pgrete:
One more general comment: I think in the long run it'd be great if we could rename `StateDescriptor` to `Packages`.
This would (from my point of view) be more closely aligned with how we handle/define/use `packages` and `Packages_t` in the code as well as our daily language when we talk about it.

src/interface/state_descriptor.cpp
pgrete: It's not used in AthenaPK so shouldn't be an issue for us. Just stumbled across the "horrible hack" :D

src/interface/state_descriptor.cpp
Yurlungur: Ah, yes. This is an issue. How about this? I'll make a comment on the same issue, and submit a PR later to make sure dimensions, flags, etc, are consistent.

pgrete: Sounds good.

Yurlungur:
Thanks, @pgrete ! As discussed, #396 should lead to another PR.

> One more general comment: I think in the long run it'd be great if we could rename `StateDescriptor` to `Packages`.
> This would (from my point of view) be more closely aligned with how we handle/define/use `packages` and `Packages_t` in the code as well as our daily language when we talk about it.

This is a good idea. I will make an issue for further discussion.

Yurlungur:
I think all concerns have been addressed. Thanks all for your feedback. Pulling the trigger.

PR #390, opened 2020-12-08
Resolve @PAR_ROOT@ to parthenon root, not project root.

AndrewGaspar:
## PR Summary

Resolve @PAR_ROOT@ to parthenon root, not project root.

Also:
- Add a guard in MachineCfg.cfg to prevent double inclusion
- Update the date of views in Darwin.cmake
- Issue a warning if MachineCfg.cmake is included after project is called

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


cmake/MachineCfg.cmake
```diff
@@ -40,6 +49,13 @@ if (USE_MACHINE_CFG)
 endif()
 
 if (MACHINE_CFG_FULL)
+  if (CMAKE_PROJECT_NAME)
+    message(WARNING "MachineCfg.cmake should be included before calling the \
+      project() function. If you're using parthenon as a subdirectory, try \
+      adding an include for it in your top level CMakeLists.txt prior to \
+      calling project().")
+  endif()
```
JoshuaSBrown, line 57: Are you referring to this? https://cmake.org/cmake/help/latest/variable/CMAKE_PROJECT_PROJECT-NAME_INCLUDE_BEFORE.html#variable:CMAKE_PROJECT_%3CPROJECT-NAME%3E_INCLUDE_BEFORE

Yurlungur: I don't quite understand. When would this code be hit? Wouldn't the function have returned already?

Yurlungur (approved):
Thanks for making these changes.

JoshuaSBrown (approved):
Looks good to me.

cmake/MachineCfg.cmake
JoshuaSBrown: No, it's only returned if the machine config file is included more than once. This warning gets triggered if the project name is defined before you include the machine config file.

Yurlungur: The `if (CMAKE_PROJECT_NAME)` part I understand. But this is encased in an `if (MACHINE_CFG_FULL)` clause too. Won't that always return false? Or am I missing something?

JoshuaSBrown: No because if its the first time through the file MACHINE_CFG_FULL is only defined after the return logic.

Yurlungur: right so if it's undefined, the if clause will read it as false, and this won't happen. But if it's defined as true, then the list would have already exited out, right?

JoshuaSBrown: But it gets defined on line 40 before the if clause on line 51 but after the return statement on line 15.

Yurlungur: Oh I see. I missed that. Sorry for the confusion.

JoshuaSBrown: Ya, no problem its easy to miss.

AndrewGaspar: > Are you referring to this? https://cmake.org/cmake/help/latest/variable/CMAKE_PROJECT_PROJECT-NAME_INCLUDE_BEFORE.html#variable:CMAKE_PROJECT_%3CPROJECT-NAME%3E_INCLUDE_BEFORE

Hm... this seems more like an alternative to `MACHINE_CFG`. `MACHINE_CFG` really needs to be included before *any* project command is called (or, at least, any language is enabled), not just before the `project(parthenon)` call.

AndrewGaspar:
@pgrete Could you look at `MachineCfg.cmake` real quick?

cmake/MachineCfg.cmake
```diff
@@ -21,11 +26,15 @@ elseif (DEFINED ENV{MACHINE_CFG})
...
+  # included from a project using Parthenon.
+  get_filename_component(PAR_ROOT ${CMAKE_CURRENT_LIST_DIR} DIRECTORY)
+
   # The `@PAR_ROOT@` string allows the devleper to specify a relative path to
-  # the project directory via an environment variable.
+  # the parthenon root directory via an environment variable.
   string(
-    REPLACE "@PAR_ROOT@" ${CMAKE_CURRENT_SOURCE_DIR}
-      USE_MACHINE_CFG ${USE_MACHINE_CFG})
+    REPLACE "@PAR_ROOT@" ${PAR_ROOT} USE_MACHINE_CFG ${USE_MACHINE_CFG})
```
pgrete, line 37: Out of curiosity, what is actually happening here? For what kind of scenario do we need this line?

If I understand correctly 
- USE_MACHINE_CFG could be sth like `"../myfile/machine.cmake"`
- `${PAR_ROOT}` now always contains the `/path/to/parthenon`  because from this file (`MachineCfg.cmake`) is included from the `/path/to/parthenon/CMakeLists.txt` and the 
- `@PAR_ROOT@` is replaced by whatever `PAR_ROOT` is set to by some external project that uses parthenon.
  - so if `@PAR_ROOT@` has the full path, it does already need to match `${PAR_ROOT}` by construction (otherwise the file that included this `MachineCfg.cmake` would not be the Parthenon `CmakeLists.txt` but some other, which is likely not intended behavior)
  - if `@PAR_ROOT@` contains a relative path then it'd replace the relative path of `USE_MACHINE_CFG` with the full path only if a custom `machine.cmake` was used in the Parthnenon subdirectory

We get the full path below anyway, so I'm just curious if this line is still required.

pgrete (commented):
@AndrewGaspar In principle this looks fine and the include guard is definitely useful.
I just have a final comment/question because I'm not sure I fully follow the logic.
If that's of no concern (and that line is required for some scenario as is) feel free to merge.

cmake/MachineCfg.cmake
pgrete: ping @AndrewGaspar (I just went through the backlog and except for this comment this looks good)

JoshuaSBrown:
I can confirm that this pr helps with configuring applications built on top of parthenon, it prevents redundant inclusion of machine config. 

@pgrete for something like AthenaPK, if the parthenon machine config file is going to be used this line needs be added before the project command is called, otherwise the wrong compilers will be registered. 

```
cmake_minimum_required(VERSION 3.13)                                                
                                                                                    
include(external/parthenon/cmake/MachineCfg.cmake)                                  
                                                                                    
project(athenaPK LANGUAGES CXX) 
```

PR #391, opened 2020-12-09
Add `GetGlobalSparseId` and `GetLocalSparseId` methods to `VariablePack`

clellsolomon:
Add GetGlobalSparseId and GetLocalSparseId to VariablePack
    
* For now both these functions dispatch to GetSparse
    
* When sparse variables are actually being used, GetGlobalSparseID
   should return the global id and GetLocalSparseID should return the
   variable-pack-local id

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Current develop
 
 ### Added (new features/APIs/variables/...)
-- [[PR 391]](https://github.com/lanl/parthenon/pull/391) Add `VariablePack<T>::GetGlobalSparseID` and `VariablePack<T>::GetGlobalSparseID` to return global sparse ids and pack-local sparse ids, repsectively.
+- [[PR 391]](https://github.com/lanl/parthenon/pull/391) Add `VariablePack<T>::GetGlobalSparseId` and `VariablePack<T>::GetGlobalSparseId` to return global sparse ids and pack-local sparse ids, repsectively.
 - [[PR 381]](https://github.com/lanl/parthenon/pull/381) Overload `DataCollection::Add` to build `MeshData` and `MeshBlockData` objects with a subset of variables.
 - [[PR 378]](https://github.com/lanl/parthenon/pull/378) Add Kokkos profiling regions throughout the code to allow the collection characteristic application profiles
 - [[PR 358]](https://github.com/lanl/parthenon/pull/358) Generalize code that interfaces with downstream apps to work with both `MeshData` and `MeshBlockData`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Current develop
 
 ### Added (new features/APIs/variables/...)
-- [[PR 391]](https://github.com/lanl/parthenon/pull/391) Add `VariablePack<T>::GetGlobalSparseId` and `VariablePack<T>::GetGlobalSparseId` to return global sparse ids and pack-local sparse ids, repsectively.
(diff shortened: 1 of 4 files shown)


AndrewGaspar:
Does "Global" mean the actual sparse ID and "local" mean the index of the sparse ID in the pack?

So if you have sparse IDs 17 and 33 for a variable, you'd have mapping in the variable pack of 0: 17 and 1: 33. Would 0 and 1 be the "local" IDs and 17 and 33 would be the "global" IDs?

clellsolomon:
@AndrewGaspar : Yes, that is correct.

AndrewGaspar:
I don't like "global" and "local" in this context, because currently "global" is used to refer to block IDs that are unique across all processors, as opposed to rank-"local" IDs, which is unique on just the rank where the block is located.

What about `GetSparseIndex` vs. `GetSparseID`?

clellsolomon:
@AndrewGaspar : Yes, that is good.  I will make the change. 

Just to be clear, `GetSparseIndex` would refer to what I called "local" and `GetGlobalID` would be what I called "global"?

AndrewGaspar:
> Just to be clear, GetSparseIndex would refer to what I called "local" and GetGlobalID* would be what I called "global"?

`GetSparseID`, but yes.

src/interface/variable_pack.hpp
```diff
@@ -95,9 +95,9 @@ class VariablePack {
     return sparse_ids_(n);
   }
   KOKKOS_FORCEINLINE_FUNCTION
-  int GetGlobalSparseId(const int n) const { return GetSparse(n); }
+  int GetSparseId(const int n) const { return GetSparse(n); }
   KOKKOS_FORCEINLINE_FUNCTION
-  int GetLocalSparseId(const int n) const { return GetSparse(n); }
+  int GetSparseIndex(const int n) const { return GetSparse(n); }
```
AndrewGaspar, line 100: Shouldn't this just be `return n;`? I might be missing something.

clellsolomon: `n` is the offset of a variable in the pack and the sparse index that identifies which sparse item that variable belongs with can be different.  

@jdolence: can you confirm that my understanding is correct as I'm just coming into this.

Yurlungur: Not @jdolence but I think thats right

Yurlungur (approved):

src/interface/variable_pack.hpp
jdolence: yes that's right.  it is definitely not `n` in general.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Current develop
 
 ### Added (new features/APIs/variables/...)
+- [[PR 391]](https://github.com/lanl/parthenon/pull/391) Add `VariablePack<T>::GetGlobalSparseId` and `VariablePack<T>::GetGlobalSparseId` to return global sparse ids and pack-local sparse ids, repsectively.
```
jdolence: little typo here.  please fix.

jdolence (changes requested):

jdolence (approved):

jdolence:
This should have zero impact on the Athena folks and is a very small change, so I'll pull the trigger as soon as the CI passes.  Blame me if this breaks something unexpectedly.

AndrewGaspar (approved):

PR #393, opened 2020-12-10
Small refactor of some driver code

jdolence:
## PR Summary

Some pretty straightforward changes to make it easier to use the RK `Integrator` object in something that derives from the `EvolutionDriver` because it wants to combine RK integration with other kinds of operations.

Be aware, this very likely breaks downstream codes, though in very easily fixable ways.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


jdolence:
Full suite run and passes.  Should be a quick review, but @pgrete one more warning that this will probably break AthenaPK.  The fix is trivial though.

src/driver/multistage.hpp
```diff
@@ -25,33 +26,37 @@
...
+        integrator(std::make_unique<StagedIntegrator>(pin)) {}
+  // An application driver that derives from this class must define this
+  // function, which defines the application specific list of tasks and
+  // there dependencies that must be executed.
+  virtual TaskCollection MakeTaskCollection(BlockList_t &blocks, int stage) = 0;
+  virtual TaskListStatus Step();
 
- private:
+ protected:
+  std::unique_ptr<StagedIntegrator> integrator;
```
clellsolomon, line 50: Question: Why not hold the `StagedIntegrator` directly instead of by `unique_ptr`?  Is there an overarching design reason not to?

clellsolomon (approved):
@jdolence : Only a minor question/curiosity...perhaps only for my own edification.  Also, is there a downstream MR in place for us that makes the requisite changes?

src/driver/multistage.hpp
```diff
@@ -25,33 +26,37 @@
 
 namespace parthenon {
 
-struct Integrator {
-  Integrator() = default;
-  Integrator(int nstages, std::vector<Real> beta) : nstages(nstages), beta(beta) {}
+struct StagedIntegrator {
+  StagedIntegrator() = default;
+  StagedIntegrator(int nstages, std::vector<Real> beta) : nstages(nstages), beta(beta) {}
```
JoshuaSBrown: Make beta const reference to avoid a copy.

src/driver/multistage.hpp
```diff
@@ -25,33 +26,37 @@
...
-  MultiStageDriver(ParameterInput *pin, ApplicationInput *app_in, Mesh *pm);
-  std::vector<std::string> stage_name;
-  Integrator *integrator;
-  ~MultiStageDriver() { delete integrator; }
+  MultiStageDriver(ParameterInput *pin, ApplicationInput *app_in, Mesh *pm)
+      : EvolutionDriver(pin, app_in, pm),
+        integrator(std::make_unique<StagedIntegrator>(pin)) {}
+  // An application driver that derives from this class must define this
+  // function, which defines the application specific list of tasks and
+  // there dependencies that must be executed.
```
JoshuaSBrown: I think "there" should be "the"

JoshuaSBrown (approved):
LGTM minor suggestions.

src/driver/multistage.hpp
jdolence: not a great one.  sometimes it might make sense to pass the integrator object into a task, as is done in the particles example.  since task arguments are captured by value, this was an attempt to discourage copying the object.  Worst case is you copy the `unique_ptr`, but preferably you just pass in the pointer, as is also done in the particles example.

I don't have strong feelings about this at all.  It would also work to just store the `StagedIntegrator` object directly and take it's address when passing it into a task.  Is that preferable?

src/driver/multistage.hpp
jdolence: just deleted this unused constructor

pgrete (changes requested):
Looks generally good (if I understood correctly).
From what I took from the code there's not the distinction, between the a multistage driver that operates on BlockLists (the `MultiStageDriver`) and on MeshBlocks `MultiStageBlockTaskDriver` (where the latter derives from the former), correct?
Having this documented somewhere would be great.
In addition, I think it'd be good to make the naming consistent, e.g., `MultiStageBlockListDriver` and `MultiStageBlockDriver` (or similar).

src/driver/multistage.hpp
clellsolomon: You cannot copy a `unique_ptr` so you you cannot pass it directly by value.  It would have to be passed by pointer or preferably (IMO) by `StagedIntegrator &` or `StagedIntegrator const&`.  I generally tend to prefer passing by reference when the object is guaranteed/required to exist, as is the case here.  My thinking is that one should not have a `MultiStageDriver` without a valid integrator.  The `unique_ptr` should not be `nullptr`, right?

If I were writing the code, I would store it directly here and prefer passing by reference (not pointer) to the task if needed.  

Less important here as the driver is unlikely to be copied (I think), but generally relevant, is that by *not* providing copy ctor/asgn this class also becomes uncopyable because its member is not copyable.  

Just food for thought.  I'm not holding the PR up over this.

clellsolomon (commented):
@jdolence : Just some more thoughts on storing by `unique_ptr`.

jdolence:
> @jdolence : Just some more thoughts on storing by `unique_ptr`.

Ah yes, you're right of course.  That's what I get for responding quickly without thinking.  Can't copy a `unique_ptr`...that's kind of its thing.  I'm remembering a bit better now how I ended up with this.  Originally, the integrator was  just a raw pointer with the object created via a call to new, which meant the driver destructor had to explicitly free it -- gross.  Meanwhile, for reasons that I can't really remember, I couldn't get our tasking code to capture by `&` or by `const &` in the time I was willing to give it, so for now at least we're stuck capturing by value.  With that, I really wanted to make sure the integrator was passed into tasks as a pointer (to avoid silly unnecessary copies).  So combining all these things led me to `unique_ptr` because it's memory managed and would lead to a compile time error unless you explicitly call `get()` to retrieve the pointer for passing into a task.

Anyway, I agree completely that it would be preferable if our tasking could be made to capture by value or by reference as appropriate which would eliminate the need for this `unique_ptr`.  At some point, maybe this can be addressed.

jdolence:
> Looks generally good (if I understood correctly).
> From what I took from the code there's not the distinction, between the a multistage driver that operates on BlockLists (the `MultiStageDriver`) and on MeshBlocks `MultiStageBlockTaskDriver` (where the latter derives from the former), correct?
> Having this documented somewhere would be great.
> In addition, I think it'd be good to make the naming consistent, e.g., `MultiStageBlockListDriver` and `MultiStageBlockDriver` (or similar).

thanks for the feedback @pgrete.  Good point about the docs.  I've updated those.  The distinction between the two is now described there.  To summarize, `MultiStageDriver` requires applications write something that defines a `TaskCollection` (with all it's generality) whereas the `MultiStageBlockTaskDriver` requires applications write something the defines a simpler `TaskList` for a given `MeshBlock` and stage.  The latter populates a `TaskCollection` with a single `TaskRegion` with `number_of_meshblocks` asynchronously executed `TaskList`s, but that's all hidden from downstream codes for simplicity.  Basically, the `MultiStageBlockTaskDriver` provides Athena++ like functionality, while the `MultiStageDriver` provides more general functionality.

Given all that, I'm not convinced the current names are the best, but I also don't think the names you suggested clarify things.  The justification for the current naming scheme is that `MultiStageDriver` provides a very general capability for driving things in a multi-stage integration scheme, whereas `MultiStageBlockTaskDriver` provides the more narrowly focused capability of multi-stage integration with tasks tied to `MeshBlocks`.  So `MultiStageDriver` provides a superset of the capabilities of `MultiStageBlockTaskDriver`.

Anyway, I'm open to other suggestions for names, but what I've got is the best I can come up with without thinking very hard.  Let me know if you still want me to change something.

Yurlungur (approved):

pgrete (approved):
Sounds good. Let's merge this in and I'll open issues to keep track of of the unique_ptr "issue" and naming

PR #394, opened 2020-12-11
Add const correctness to `Params.Get`

clellsolomon:
Add const correctness to `Params.Get`

It was returning a const reference to something it owns, so it should be const.  The only thing preventing it was some of the error checks that can be rewritten to support const-ness.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/src/interface/params.hpp
+++ b/src/interface/params.hpp
@@ -53,9 +53,8 @@ class Params {
   const T &Get(const std::string &key) const {
     auto const it = myParams_.find(key);
     PARTHENON_REQUIRE_THROWS(it != myParams_.end(), "Key " + key + " doesn't exist");
-    PARTHENON_REQUIRE_THROWS(
-        !(myTypes_.find(key)->second.compare(std::string(typeid(T).name()))),
-        "WRONG TYPE FOR KEY '" + key + "'");
+    PARTHENON_REQUIRE_THROWS(!(myTypes_.at(key).compare(std::string(typeid(T).name()))),
+                             "WRONG TYPE FOR KEY '" + key + "'");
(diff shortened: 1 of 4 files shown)


src/interface/params.hpp
```diff
@@ -50,11 +50,12 @@ class Params {
...
   template <typename T>
-  const T &Get(const std::string &key) {
-    auto it = myParams_.find(key);
+  const T &Get(const std::string &key) const {
+    auto const it = myParams_.find(key);
     PARTHENON_REQUIRE_THROWS(it != myParams_.end(), "Key " + key + " doesn't exist");
-    PARTHENON_REQUIRE_THROWS(!(myTypes_[key].compare(std::string(typeid(T).name()))),
-                             "WRONG TYPE FOR KEY '" + key + "'");
+    PARTHENON_REQUIRE_THROWS(
+        !(myTypes_.find(key)->second.compare(std::string(typeid(T).name()))),
```
Yurlungur: should this be `.find()->second`? Or would `.at(key)` be better?

Yurlungur (approved):
Thanks for doing this @clellsolomon . One minor nitpick below. Otherwise looks good.

src/interface/params.hpp
clellsolomon: Yes, it would be.   I'll make the change.

clellsolomon: Not only is it better, what I wrote is a bug because `find(key)` could return end which I then de referenced.  Doh!

nmsriram (approved):
Thanks for making the changes.  These look good.

PR #395, opened 2020-12-14
Joshua s brown/app for parthenon

JoshuaSBrown:
## PR Summary

This branch is part of the larger changes to automate the generation of performance metrics as part of the ci. Here I introduce the pathenon metrics app. This app is a python script that should be authenticated to push files to parthenon's github repository or the wiki from within the ci.

Currently, the app is set to push images to a figures branch so as not to cause bloating of the main repository.

This pr is WIP for now I would like to discuss features that should be available to this application. Right now it is setup to 
simply upload files in an authenticated fashion. Though the permissions have not yet been enabled.

E.g. If you wanted to upload a log file you would do something like this within in the ci:

```
python3 pathenon_metrics_app.py --pem file.pem --upload metrics.log --branch metrics
``` 

One additional feature that should probably be added is the ability to upload to the wiki instead of Pathenon's main repository. 

Files with the following extensions will by default be uploaded to a figures branch in the main repository: '.png', '.jpg', '.jpeg', '.tiff', '.bmp', '.gif', this branch should never be merged with main or develop. 

Image files (graphs) showing performance metrics that exist in the parthenon figures branch can be directly linked in the wiki that was the ci can automatically update figures in the wiki as needed. 

It might be worth also continually squashing commits in the figures branch to prevent repo bloat. 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


JoshuaSBrown:
Waiting for https://github.com/spack/spack/pull/20669.

JoshuaSBrown:
I'm adding documentation now but this pr is ready for review.

JoshuaSBrown:
There are a few points I want to mention, this application is far from feature complete but I think at this stage it is useful enough to merge. E.g. some of the other things that need to be considered moving forward. https://github.com/lanl/parthenon/issues/248

Also, right now the performance metrics app does not report a failed status if the performance decreases. I was not sure if this was needed or how best to implement it. As it is whoever looks over the pr will have to check the figures to determine if there is a significant regression.

Yurlungur:
> Also, right now the performance metrics app does not report a failed status if the performance decreases. I was not sure if this was needed or how best to implement it. As it is whoever looks over the pr will have to check the figures to determine if there is a significant regression.

I think pass/fail here really depends on how significant of a regression it is. This might be best handled by humans squinting at plots.

JoshuaSBrown:
> > Also, right now the performance metrics app does not report a failed status if the performance decreases. I was not sure if this was needed or how best to implement it. As it is whoever looks over the pr will have to check the figures to determine if there is a significant regression.
> 
> I think pass/fail here really depends on how significant of a regression it is. This might be best handled by humans squinting at plots.

:) That was my thought too, though it might be helpful to show a performance difference as a plot as well.

JoshuaSBrown:
Metrics app is currently failing because of a copyright check that was merged into develop. This pr will let me disable the feature. https://github.com/lanl/parthenon/pull/452

scripts/darwin/build_fast.sh
```diff
@@ -3,19 +3,48 @@
...
+SCRIPT=$(realpath "$0")
+SOURCE=$(dirname "$SCRIPT")
+GITHUB_APP_PEM="$1"
+export CI_COMMIT_SHA="$2"
+export CI_COMMIT_BRANCH="$3"
+CI_JOB_URL="$4"
+export GITHUB_APP_PEM="$GITHUB_APP_PEM"
+
+trap 'catch $? $LINENO' ERR
+catch() {
```
Yurlungur: I had no idea `bash` had `try-catch` blocks. That's cool.

scripts/darwin/metrics.sh
```diff
@@ -0,0 +1,48 @@
...
+export CI_COMMIT_BRANCH="${4}"
+echo "Metrics PEM file $GITHUB_APP_PEM"
+
+trap 'catch $? $LINENO' ERR
+catch() {
+  echo "Error $1 occurred on $2"
+  echo "BUILD FAILED"
+  # For some reason the ERR variable is not recongized as a numeric type when quoted, 
+  # exit requires err be recognized as a numeric type, furthermore the quotes
+  # are needed as good practice
```
Yurlungur: I thought quotes vs `${}` were fairly equivalent. Why are the quotes better practice?

scripts/python/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+# This program was produced under U.S. Government contract 89233218CNA000001 for Los
+# Alamos National Laboratory (LANL), which is operated by Triad National Security, LLC
+# for the U.S. Department of Energy/National Nuclear Security Administration. All rights
+# in the program are reserved by Triad National Security, LLC, and the U.S. Department
+# of Energy/National Nuclear Security Administration. The Government is granted for
+# itself and others acting on its behalf a nonexclusive, paid-up, irrevocable worldwide
+# license in this material to reproduce, prepare derivative works, distribute copies to
+# the public, perform publicly and display publicly, and to permit others to do so.
+#=========================================================================================
+
```
Yurlungur: I'm a little uncomfortable with a `__init__.py` script directly in the `scripts/python` directory. This means that our entire directory is a python "package." Is this something we absolutely need? If we're going this way, I'd rather make an explicit package named `parthenon` or something. So the structure would be
```
scripts/python/parthenon/__init__.py
```
we could then put a `setup.py` file in `scripts/python` that could "install" the parthenon scripts into a pythonpath, for example.

scripts/python/app.py
```diff
@@ -0,0 +1,600 @@
...
+import pem
+import pycurl
+from git import Repo
+import git
+
+class Node:
+
+    """
+    Class used to store branch contents in terms of files, directories or misc
+    """
```
Yurlungur: I do this all the time too. But I think technically, there should be no line break between the opening quotes and the first line of text.

scripts/python/app.py
```diff
@@ -0,0 +1,600 @@
...
+        for fil in self.files:
+            print("File " + fil)
+        for mis in self.misc:
+            print("Misc " + mis)
+        for node in self.dirs:
+            node.printTree()
+
+
+
+class App:
```
Yurlungur: The name `App` is not very descriptive. We may want to think of a different name. Something like "PerformanceReporter" maybe?

scripts/python/parthenon_metrics_app.py
```diff
@@ -0,0 +1,791 @@
...
+
+
+def main(**kwargs):
+
+    app = ParthenonApp()
+    app.initialize(
+        kwargs.pop('wiki'),
+        kwargs.pop('ignore'),
+        kwargs.pop('permissions'),
+        kwargs.pop('create'))
```
Yurlungur: I'm not a huge fan of `pop` here, since it deletes these values from the kwargs dict forever. Is that the intended behaviour? You could also just do `kwargs['wiki']` to get that value?

scripts/python/parthenon_metrics_app.py
```diff
@@ -0,0 +1,791 @@
...
+        app.printTargetBranch(branch)
+
+
+# Execute main function
+if __name__ == '__main__':
+
+    parser = argparse.ArgumentParser(
+        "python3 parthenon_metrics_app.py -p file.pem")
+
+    desc = ('Path to the (permissions file/permissions string) which authenticates the application. If not provided will use the env variable PARTHENON_METRICS_APP_PEM.')
```
Yurlungur: this is a bit long for a single line. Can we break it up?

Yurlungur (approved):
Wow this was a lot of work! I have several stylistic comments below. But overall, looks great.

.gitlab-ci-darwin.yml
```diff
@@ -1,26 +1,49 @@
...
 
 .gcc-mpi-cuda-performance-regression:
   variables: 
     CMAKE_CXX_COMPILER: $CI_PROJECT_DIR/external/Kokkos/bin/nvcc_wrapper
   script:
-    - env -i bash --norc --noprofile ./scripts/darwin/build_fast.sh
+    - env
+    - git fetch --all
+    - printf 'env -i bash --norc --noprofile %s/scripts/darwin/build_fast_target.sh  "%s" "%s" "%s" "%s"\n' ${CI_PROJECT_DIR} ${GITHUB_APP_PEM} ${CI_COMMIT_SHA} ${CI_COMMIT_BRANCH} ${CI_JOB_URL}> execute_main.sh
+    - printf 'env -i bash --norc --noprofile %s/scripts/darwin/metrics.sh "%s" "%s/build" "%s" "%s"\n' ${CI_PROJECT_DIR} ${GITHUB_APP_PEM} ${CI_PROJECT_DIR} ${CI_COMMIT_SHA} ${CI_COMMIT_BRANCH}>> execute_main.sh
```
jlippuner: What's the purpose of `execute_main.sh`? Does it actually get executed by something or is it just a log file? Do these two lines exactly duplicate the two commands below? If so, we should only have them once so that they don't need to be kept in sync.

If `execute_main.sh` is just a log file, maybe a combination of `set -x` and `exec > execute_main.sh` can be used in the commands below to record what's being done. Or maybe we can just execute `execute_main.sh` below instead of  providing the full commands again. If you want to stop execution on the first error, I think `set -e` will do that for you.

.gitlab-ci-darwin.yml
```diff
@@ -1,26 +1,49 @@
...
-      - ${CI_PROJECT_DIR}/build/tst/regression/outputs/advection_performance/performance.png
-      - ${CI_PROJECT_DIR}/build/tst/regression/outputs/advection_performance_mpi/performance.png
+.gcc-mpi-cuda-performance-regression-target-branch:
+  script:
+    - git fetch --all
+    - printf 'env -i bash --norc --noprofile %s/scripts/darwin/build_fast_target.sh  "%s" "%s/build" "%s" "%s" "%s %s"\n' ${CI_PROJECT_DIR} ${GITHUB_APP_PEM} ${CI_PROJECT_DIR} ${CI_COMMIT_SHA} ${CI_COMMIT_BRANCH} ${CI_JOB_URL} ${CI_JOB_TOKEN} > execute_target.sh
+    - |
+      STATUS=$(env -i bash --norc --noprofile ${CI_PROJECT_DIR}/scripts/darwin/build_fast_target.sh "${GITHUB_APP_PEM}" "${CI_PROJECT_DIR}/build"  "${CI_COMMIT_SHA}" "${CI_COMMIT_BRANCH}" "${CI_JOB_URL}" "${CI_JOB_TOKEN}")
+      echo "PASS_OR_FAIL ${STATUS}"
+      if [[ "$STATUS" == *"FAILED"* ]]; then exit 1; fi
```
jlippuner: Same comment as above. If these commands are always the same, they should only appear once.

.gitlab-ci-darwin.yml
```diff
@@ -1,26 +1,49 @@
...
+  allow_failure: false
+  when: manual
+  except:
+    - schedules
 
 parthenon-power9-gcc-mpi-cuda-perf-manual:
   extends: .gcc-mpi-cuda-performance-regression
   stage: performance-regression
-  when: manual
+  needs: [parthenon-power9-gcc-mpi-cuda-perf-manual-target-branch]
```
jlippuner: So will this still be manually triggered because its dependency must be manually triggered?

CHANGELOG.md
```diff
@@ -24,6 +24,8 @@
 - [[PR 390]](https://github.com/lanl/parthenon/pull/390) Resolve @PAR_ROOT@ to parthenon root rather than the location of the current source directory
 - [[PR 443]](https://github.com/lanl/parthenon/pull/443) Fix Darwin machine config - use spectrum mpi
 - [[PR 444]](https://github.com/lanl/parthenon/pull/444) Writes performance metrics to file for advection test
+- [[PR 432]](https://github.com/lanl/parthenon/pull/432) Fix advection-performance build configuration
```
jlippuner: i'm confused. This is PR 395 not 432

scripts/darwin/build_fast.sh
```diff
@@ -3,19 +3,48 @@
 # Load system env only
 source /etc/bashrc
 source /etc/profile
-source /projects/parthenon-int/parthenon-project/.bashrc
+source /projects/parthenon-int/parthenon-project/spack/share/spack/setup-env.sh
```
jlippuner: What's the reason for this change?

scripts/darwin/build_fast.sh
```diff
@@ -3,19 +3,48 @@
...
+  echo "BUILD FAILED"
+  # For some reason the ERR variable is not recongized as a numeric type when quoted, 
+  # exit requires err be recognized as a numeric type, furthermore the quotes
+  # are needed as good practice
+  exit "$(($ERR))"
+}
+
+module load gcc/9.3.0
+spack compiler find
+spack env activate darwin-ppc64le-gcc9-2021-02-08
```
jlippuner: What if we update the view, then we'd have to update this as well? Are we not using cmake (and thus the cmake machine config file) because we want to build "fast"?

scripts/darwin/build_fast.sh
```diff
@@ -3,19 +3,48 @@
...
+echo "build fast $GITHUB_APP_PEM"
+"${SOURCE}"/../python/parthenon_metrics_app.py -p "${GITHUB_APP_PEM}"  --status "pending" --status-context "Parthenon Metrics App" --status-description "Building parthenon" --status-url "${CI_JOB_URL}"
+
 # Calculate number of available cores
 export J=$(( $(nproc --all) )) && echo Using ${J} cores during build
 
+source /projects/parthenon-int/parthenon-project/.bashrc
```
jlippuner: Ah, it's down here now. Any reason we can't load this at the beginning?

scripts/darwin/build_fast_target.sh
```diff
@@ -0,0 +1,88 @@
+#!/bin/bash
```
jlippuner: There seems to be quite a bit of overlap between this script and `build_fast.sh`. Might it be possible to combine those into one script and differentiate between the two cases with an (optional) argument?

scripts/darwin/metrics.sh
```diff
@@ -0,0 +1,48 @@
...
+  exit "$(($ERR))"
+}
```
jlippuner: Again overlap with previous scripts. would be nice to combine those. Maybe even by just sourcing a common script

scripts/python/__init__.py
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env python3
+#=========================================================================================
+# (C) (or copyright) 2020. Triad National Security, LLC. All rights reserved.
```
jlippuner: 2020 or 2020-2021?

scripts/python/parthenon_metrics_app.py
```diff
@@ -0,0 +1,791 @@
...
+
+    branch = kwargs.pop('branch')
+
+    if isinstance(branch, list):
+        branch = branch[0]
+
+    if 'upload' in kwargs:
+        value = kwargs.pop('upload')
+        if isinstance(value, list):
+            value = value[0]
```
jlippuner: Maybe define a helper function for this and use it everywhere you do this check? For example:
```python
def get_val(kwargs, name):
  value = kwargs[name]
  if isinstance(value, list):
    if len(value) == 1:
      return value[0]
    else:
      raise ValueError("Expected exactly 1 value for %s, but got %i" % (name, len(value)))
  else:
    return value
```

scripts/python/app.py
```diff
@@ -0,0 +1,600 @@
...
+        c.setopt(pycurl.VERBOSE, 0)
+        c.setopt(c.WRITEDATA, buffer_temp)
+        c.setopt(c.HTTPHEADER, header)
+        c.perform()
+        c.close()
+
+        js_obj = json.loads(buffer_temp.getvalue())
+
+        if isinstance(js_obj, list):
+            js_obj = js_obj[0]
```
jlippuner: I recommend making this into a re-usable helper function that just takes a URL (and maybe other inputs) and returns the json dict. This would reduce quite a bit of code duplication. You might even combine it with the POST requests with an optional post argument. Or have two helper functions, one for GET and one for POST. It might also be worthwhile to store the Curl instance and re-use it

scripts/python/parthenon_metrics_app.py
```diff
@@ -0,0 +1,791 @@
...
+    parser = argparse.ArgumentParser(
+        "python3 parthenon_metrics_app.py -p file.pem")
+
+    desc = ('Path to the (permissions file/permissions string) which authenticates the application. If not provided will use the env variable PARTHENON_METRICS_APP_PEM.')
+
+    parser.add_argument('--permissions', '-p',
+                        default="",
+                        type=str,
+                        nargs=1,
+                        required=False,
```
jlippuner: Here and elsewhere: Isn't `required=False` the default?

scripts/python/parthenon_metrics_app.py
```diff
@@ -0,0 +1,791 @@
...
+
+    parser = argparse.ArgumentParser(
+        "python3 parthenon_metrics_app.py -p file.pem")
+
+    desc = ('Path to the (permissions file/permissions string) which authenticates the application. If not provided will use the env variable PARTHENON_METRICS_APP_PEM.')
+
+    parser.add_argument('--permissions', '-p',
+                        default="",
+                        type=str,
+                        nargs=1,
```
jlippuner: Here and elsewhere: Is this necessary? The default action `store` consumes 1 argument.

scripts/python/parthenon_metrics_app.py
```diff
@@ -0,0 +1,791 @@
...
+        if isinstance(value, list):
+            value = value[0]
+        if not value is None:
+            app.upload(value, branch)
+
+    if 'status' in kwargs:
+        value = kwargs.pop('status')
+        if isinstance(value, list):
+            value = value[0]
+        if not value is None:
```
jlippuner: This is splitting hairs, but I'd prefer `if value is not None:` (and elsewhere)

scripts/python/parthenon_metrics_app.py
```diff
@@ -0,0 +1,791 @@
...
+                    target_branch = branch
+            app.analyze(
+                value,
+                branch,
+                target_branch,
+                kwargs.pop('post_analyze_status'),
+                kwargs.pop('generate_figures_on_analysis'))
+
+    check = kwargs.pop('check_branch_metrics_uptodate')
+    if check:
```
jlippuner: What's the reason for storing `check`?

scripts/python/parthenon_metrics_app.py
```diff
@@ -0,0 +1,791 @@
...
+                        "Performance Metrics for test %s is uptodate: %s" %
+                        test_isUpToDate)
+                    if not test_isUpToDate:
+                        isUpToDate = False
+            else:
+                print(
+                    "Branch (%s) is not available on github." %
+                    target_branch)
+                isUpToDate = False
+            print("Performance Metrics are uptodate: %s" % isUpToDate)
```
jlippuner: This could be moved to the end of the function to avoid duplicating it in the `else` branch

scripts/python/app.py
```diff
@@ -0,0 +1,600 @@
...
+        if isinstance(js_obj, list):
+                # Cycle through list to try to find the right object
+            for obj in js_obj:
+                contents[obj['name']] = obj['sha']
+
+        return contents
+
+    def upload(self, file_name, branch=None, use_wiki=False):
+        """
+        This method attempts to upload a file to the specefied branch.
```
jlippuner:
```suggestion
        This method attempts to upload a file to the specified branch.
```

scripts/python/app.py
```diff
@@ -0,0 +1,600 @@
...
+
+        If the file is found to already exist it will be updated. Image files will by default be placed
+        in a figures branch of the main repository, so as to not bloat the repositories commit history.
+        """
+        if isinstance(file_name, list):
+            file_name = file_name[0]
+        if branch is None:
+            branch = self._default_branch
+        if file_name.lower().endswith(('.png', '.jpg', '.jpeg', '.tiff', '.bmp', '.gif')):
+            print("Image file detected")
```
jlippuner: All `print` statements: Might we want to turn those off at some point if we don't need the debug output anymore? Maybe this could be replaced with a wrapper that we could turn off with a single switch

scripts/python/app.py
```diff
@@ -0,0 +1,600 @@
...
+                print("Unless the ignore flag is used.")
+                branch = self._default_image_branch
+                self._use_wiki = False
+
+        if self._use_wiki or use_wiki:
+            if branch != "master":
+                print(
+                    "Files can only be uploaded to the wiki repositories master branch")
+                return
+            else:
```
jlippuner: Since there is a return above, this `else` is not necessary. Will save some indentation...

scripts/python/app.py
```diff
@@ -0,0 +1,600 @@
...
+                destination = self._parthenon_wiki_dir + "/" + \
+                    os.path.basename(os.path.normpath(file_name))
+                if not filecmp.cmp(file_name, destination):
+                    shutil.copy(file_name, destination)
+                repo.index.add([str(self._parthenon_wiki_dir + "/" +
+                                    os.path.basename(os.path.normpath(file_name)))])
+                repo.index.commit(commit_msg)
+                repo.git.push("--set-upstream", "origin", repo.head.reference)
+                return
+        else:
```
jlippuner: Same as above

scripts/python/app.py
```diff
@@ -0,0 +1,600 @@
...
+                    'content': encoded_file.decode('ascii')
+                }
+
+            else:
+                custom_data = {
+                    'message': self._name + " uploading file " + os.path.basename(os.path.normpath(file_name)),
+                    'name': self._name,
+                    'content': encoded_file.decode('ascii'),
+                    'branch': branch
+                }
```
jlippuner:
```suggestion
            custom_data = {
                'message': "%s %s file %s" % (self._name, "overwriting" if file_found else "uploading", os.path.basename(os.path.normpath(file_name)),
                'name': self._name,
                'branch': branch,
                'content': encoded_file.decode('ascii')
            }
            if file_found:
                custom_data['sha'] = contents[os.path.basename(os.path.normpath(file_name))]
```

scripts/python/app.py
```diff
@@ -0,0 +1,600 @@
...
+                    " in the main repository.")
+                print("Unless the ignore flag is used.")
+                branch = self._default_image_branch
+                self._use_wiki = False
+
+        if self._use_wiki or use_wiki:
+            if branch != "master":
+                print(
+                    "Files can only be uploaded to the wiki repositories master branch")
+                return
```
jlippuner: This is not an exception?

scripts/python/app.py
```diff
@@ -0,0 +1,600 @@
...
+        return self._branch_current_commit_sha.get(target_branch)
+
+    def branchExist(self, branch):
+        """This method will determine if a branch exists on the github repository by pinging the
+        github api
+        """
+        branches = self.getBranches()
+        if branch in branches:
+            return True
+        return False
```
jlippuner:
```suggestion
        return branch in self.getBranches()
```

scripts/python/app.py
```diff
@@ -0,0 +1,600 @@
...
+            custom_data = {"branch": branch}
+            buffer_temp2 = BytesIO(json.dumps(custom_data).encode('utf-8'))
+            c = pycurl.Curl()
+            c.setopt(pycurl.VERBOSE, 0)
+            c.setopt(c.URL, self._repo_url + "/contents/" + node.getPath())
+            c.setopt(c.READDATA, buffer_temp2)
+            c.setopt(c.WRITEDATA, buffer_temp)
+            c.setopt(c.HTTPHEADER, self._header)
+            c.perform()
+            c.close()
```
jlippuner: Maybe we could re-use the `Curl` object?

scripts/python/parthenon_metrics_app.py
```diff
@@ -0,0 +1,791 @@
...
+                            # Overwrite the existing content with the new
+                            # content
+                            data_grp['mesh_blocks'] = copy.deepcopy(
+                                data_grp2.get('mesh_blocks'))
+                            data_grp['zone_cycles'] = copy.deepcopy(
+                                data_grp2.get('zone_cycles'))
+                            return
+            else:
+                # Then the test was not found so we are going to append to it
+                self._data['data'].append(new_data['data'])
```
jlippuner: Would this work? Please double check where the `return` should go.

```suggestion
        if isinstance(new_data, list):
            if len(new_data) == 1:
                new_data = new_data[0]
            else:
                raise ValueError("Expected exactly 1 new data")
                
        dat_list = self._data if isinstance(self._data, list) else [self._data]
        for json_obj in dat_list:
            if json_obj.get('commit sha') == new_data.get('commit sha'):
                for data_grp in json_obj.get('data'):
                    for data_grp2 in new_data.get('data'):
                        if data_grp.get('test') == new_data.get('test'):
                            # Overwrite the existing content with the new content
                            data_grp['mesh_blocks'] = copy.deepcopy(
                                data_grp2.get('mesh_blocks'))
                            data_grp['zone_cycles'] = copy.deepcopy(
                                data_grp2.get('zone_cycles'))
                return
            else:
                # Then the test was not found so we are going to append to it
                json_obj['data'].append(new_data['data'])
```

scripts/python/parthenon_metrics_app.py
```diff
@@ -0,0 +1,791 @@
...
+                                if data_grp.get('test') == test:
+                                    mesh_blocks = data_grp.get('mesh_blocks')
+                                    cycles = data_grp.get('zone_cycles')
+
+                        if new_datetime > recent_datetime:
+                            recent_datetime = new_datetime
+                            for data_grp in json_obj.get('data'):
+                                if data_grp.get('test') == test:
+                                    mesh_blocks = data_grp.get('mesh_blocks')
+                                    cycles = data_grp.get('zone_cycles')
```
jlippuner: Please try to remove code duplication

scripts/python/parthenon_metrics_app.py
```diff
@@ -0,0 +1,791 @@
...
+                for data_grp in json_obj.get('data'):
+                    if data_grp.get('test') == test:
+                        mesh_blocks = data_grp.get('mesh_blocks')
+                        if isinstance(mesh_blocks, str):
+                            mesh_blocks = np.array(
+                                mesh_blocks.strip("[").strip("]").split()).astype(
+                                np.float)
+                        return mesh_blocks
+            list_ind = list_ind + 1
+        return None
```
jlippuner: Please merge, if possible

scripts/python/parthenon_metrics_app.py
```diff
@@ -0,0 +1,791 @@
...
+                        if data_grp.get('test') == test:
+                            mesh_blocks = data_grp.get('mesh_blocks')
+                            cycles = data_grp.get('zone_cycles')
+
+                if new_datetime > recent_datetime:
+                    recent_datetime = new_datetime
+                    for data_grp in json_obj.get('data'):
+                        if data_grp.get('test') == test:
+                            mesh_blocks = data_grp.get('mesh_blocks')
+                            cycles = data_grp.get('zone_cycles')
```
jlippuner: Code duplication

jlippuner (changes requested):
Thanks @JoshuaSBrown for these great improvements. I'm very happy to see all the helpful comments!

I added a number of comments and questions. One recurring theme is that there is quite a bit of code duplication. I added some suggestions to reduce it and highlighted in some cases. I think getting rid of the code duplication will vastly improve readability and maintainability.

Also, I'd suggest using 2-space-indent for the Python scripts to keep the lines at a manageable length and help readability.

scripts/python/app.py
Yurlungur: [We could use Python's built in logging functionality.](https://docs.python.org/3/howto/logging.html)

scripts/darwin/metrics.sh
JoshuaSBrown: Codacy gave me a bunch of reasons, I think the difference between "" and ${} is how they are expanded, seems to be a pretty good discussion here if you are interested https://unix.stackexchange.com/questions/4899/var-vs-var-and-to-quote-or-not-to-quote

scripts/python/__init__.py
JoshuaSBrown: > I'm a little uncomfortable with a `__init__.py` script directly in the `scripts/python` directory. This means that our entire directory is a python "package." Is this something we absolutely need? If we're going this way, I'd rather make an explicit package named `parthenon` or something. So the structure would be
> 
> ```
> scripts/python/parthenon/__init__.py
> ```
> 
> we could then put a `setup.py` file in `scripts/python` that could "install" the parthenon scripts into a pythonpath, for example.

I think that is a good idea, I'm still a little ignorant of python best practices.

scripts/python/app.py
JoshuaSBrown: So I did actually have a reason for this. I have opened an EIDR trying to get the App file open sourced as a standalone item. It is meant to be generic. The actual implementation of App is in the parthenon_metrics_app.py that is directly relevant to parthenon. In all actuallity the App class could be used to create several different apps. If you want something a little more specific I could change it to GithubApp but beyond that I'm not sure what else would be appropriate.

scripts/python/parthenon_metrics_app.py
JoshuaSBrown: Sounds good.

.gitlab-ci-darwin.yml
JoshuaSBrown: > What's the purpose of `execute_main.sh`? Does it actually get executed by something or is it just a log file? Do these two lines exactly duplicate the two commands below? If so, we should only have them once so that they don't need to be kept in sync.
> 
> If `execute_main.sh` is just a log file, maybe a combination of `set -x` and `exec > execute_main.sh` can be used in the commands below to record what's being done. Or maybe we can just execute `execute_main.sh` below instead of providing the full commands again. If you want to stop execution on the first error, I think `set -e` will do that for you.

That's a really good point, the execute main was used to help with debugging the ci, it is a lot easier to do it this way because it save the state of the ci variables, otherwise I have to go look them up, but ya I could simply execute main so I'm not duplicating, great idea.

.gitlab-ci-darwin.yml
JoshuaSBrown: > So will this still be manually triggered because its dependency must be manually triggered?

That is correct. :)

CHANGELOG.md
JoshuaSBrown: > i'm confused. This is PR 395 not 432

That's because it is an error :) good catch.

scripts/darwin/build_fast.sh
JoshuaSBrown: > What's the reason for this change?

That's a good question. It is generally not necessary to call spack directly when running with the bashrc file because CMake imports all the dependencies from the machine config file. However, the python application also has dependencies these are not automatically found. What I could try to do is change the logic in the bashrc file if needed, so the application dependencies are loaded there. That way all the build dependencies are handled in the same file.

This has it's own set of problems though. The spack environment that is loaded is dependent on what is in the machine config file for Dawin. There is a date associated with the set of dependencies. The bashrc does not know which environment to pick before hand.

scripts/darwin/build_fast.sh
JoshuaSBrown: Ah, this is a good point. Hmm how to make it consistent with CMake machine config but only specify it once.

scripts/darwin/build_fast_target.sh
JoshuaSBrown: That's a possibility.

scripts/darwin/build_fast.sh
JoshuaSBrown: There shouldn't be other than simply putting things closer to where they are being used.

scripts/python/__init__.py
JoshuaSBrown: Good catch

scripts/python/parthenon_metrics_app.py
JoshuaSBrown: Good idea.

scripts/python/app.py
JoshuaSBrown: That is a good idea.

scripts/python/parthenon_metrics_app.py
JoshuaSBrown: Good point I didn't realize I could word it that way.

scripts/python/parthenon_metrics_app.py
JoshuaSBrown: Will have to check.

scripts/python/parthenon_metrics_app.py
JoshuaSBrown: In this case it was intentional. Only one permissions file should be passed in, there is no logic for dealing with a second, and one should definately be passed in if you want to do anything useful with the application.

JoshuaSBrown:
@jlippuner @Yurlungur thanks for the feedback it is super helpful I realized this was a large pr, after working on it for several months I definitely was experiencing some fatigue with this having a fresh set of eyes is really great. What is so painful with this pr is sometimes I can only catch bugs while running the ci which slows down the turn around time of refactoring. I will do my best to address your concerns promptly.

scripts/darwin/metrics.sh
Yurlungur: Ah. The globbing thing makes sense.

scripts/python/__init__.py
Yurlungur: This also means we can put scripts we don't want to install one level above `scripyts/python/parthenon` so they're not bundled.

scripts/python/app.py
Yurlungur: I would prefer `GitHubApp`. `App` is just too generic---my initial instinct would be to read the name as "oh this is the parthenon application," which is confusing.

.gitlab-ci-darwin.yml
JoshuaSBrown: So it turns out this is a bit difficult to make DRY. The ci yaml files do not really have a notion of functions. It would be nice if I could just put the logic in the shell script for handling the error codes but this does not work because when I call the script using the "env" command the error code gets lost. It is important that the right exit code is called from within the yaml file because this is how we know if there was a problem or not. At this point as nasty as it seems I will be leaving the "if logic" as is.

scripts/darwin/build_fast.sh
JoshuaSBrown: I just made sure to pull in the latest env detected this should be a pretty safe default because it should be backwards compatible.

.gitlab-ci-darwin.yml
JoshuaSBrown: Ok, I figured out how to do this with anchors and templates.

.gitlab-ci-darwin.yml
JoshuaSBrown: Done.

scripts/darwin/build_fast_target.sh
JoshuaSBrown: Done.

scripts/darwin/metrics.sh
JoshuaSBrown: Done

scripts/python/parthenon_metrics_app.py
JoshuaSBrown: Removed all instances, this is the default.

scripts/python/parthenon_metrics_app.py
JoshuaSBrown: Done.

scripts/python/parthenon_metrics_app.py
JoshuaSBrown: Done.

scripts/python/parthenon_metrics_app.py
JoshuaSBrown: Done.

scripts/python/app.py
JoshuaSBrown: Done

scripts/python/app.py
JoshuaSBrown: Fixed

scripts/python/app.py
JoshuaSBrown: Good ideas.

JoshuaSBrown: Done

scripts/python/app.py
JoshuaSBrown: Done

scripts/python/app.py
JoshuaSBrown: Done.

scripts/python/app.py
JoshuaSBrown: Done

scripts/python/app.py
JoshuaSBrown: Fixed.

scripts/python/parthenon_metrics_app.py
JoshuaSBrown: Done

JoshuaSBrown:
ping @pgrete @jlippuner, the only reason the ci is not passing is related to this issue. https://github.com/lanl/parthenon/issues/489 which should probably be addressed in a separate pr.

scripts/python/packages/parthenon_performance_app/README.md
```diff
@@ -0,0 +1,16 @@
+# Parthenon Python Metrics Github Application
+
+This package is responsible for running the performance metrics of the pathernon project.
```
pgrete:
```suggestion
This package is responsible for running the performance metrics of the Parthenon project.
```

scripts/python/packages/parthenon_performance_app/parthenon_performance_app/parthenon_performance_advection_analyzer.py
```diff
@@ -0,0 +1,123 @@
...
+                now
+                ):
+
+        if not os.path.isfile(
+                regression_outputs + "/advection_performance/performance_metrics.txt"):
+            raise Exception(
+                "Cannot analyze advection_performance, missing performance metrics file: " +
+                regression_outputs +
+                "/advection_performance/performance_metrics.txt")
+#        super().cloneWikiRepo()
```
pgrete:
```suggestion
```

pgrete (approved):
What a Herculean task! Nice work.
I don't have much to add as everything looks pretty self consistent and adding more tests is also up to me (and others).

Two (slightly related questions):
1. I saw that you updated the Kokkos submodule to 3.3.1. Was this required for some feature? I'm not objecting at all -- just curious.
2. How flexible is the app if I want to use it downstream to also keep track of `AthenaPK` performance (potentially in a separate wiki/repo)?

Assuming that the tests pass once we decided what do do about #489 I fully approve.

JoshuaSBrown:
> What a Herculean task! Nice work.
> I don't have much to add as everything looks pretty self consistent and adding more tests is also up to me (and others).
>
> Two (slightly related questions):
> 
>     1. I saw that you updated the Kokkos submodule to 3.3.1. Was this required for some feature? I'm not objecting at all -- just curious.

No, good catch, this keeps sneaking in, whenever I get lazy with my git calls. I'll be sure to fix it.  
 
>     2. How flexible is the app if I want to use it downstream to also keep track of `AthenaPK` performance (potentially in a separate wiki/repo)?

Depends, which part of the python tool chain you are referring too. The base class that interacts with github through RESTful api calls should be very adaptable, I tried to design it that way and am trying to get it as it's own repository. The parthenon performance derived class less so but you should be able to borrow large chunks from it. 

> 
> Assuming that the tests pass once we decided what do do about #489 I fully approve.

Sounds, good thanks for taking the time to look through it.

JoshuaSBrown:
@jlippuner ping

jlippuner (approved):
Looks great! Sorry it took me so long to review and thanks for this huge effort of work!

pgrete:
Shall we squash and merge?
I'd like to see this in sooner than later as I think it'd be a great addition for some current and upcoming PRs (e.g., #508 )

jlippuner:
> Shall we squash and merge?
> I'd like to see this in sooner than later as I think it'd be a great addition for some current and upcoming PRs (e.g., #508 )

I think @JoshuaSBrown said he needed to remove the Kokkos upgrade from this branch before merging. I think this is ready to go, but I'd let @JoshuaSBrown actually pull the trigger

JoshuaSBrown:
> > Shall we squash and merge?
> > I'd like to see this in sooner than later as I think it'd be a great addition for some current and upcoming PRs (e.g., #508 )
> 
> I think @JoshuaSBrown said he needed to remove the Kokkos upgrade from this branch before merging. I think this is ready to go, but I'd let @JoshuaSBrown actually pull the trigger

Ya, I will get this in today, thanks for all the feedback there are a few tweaks but with the approvals I will just merge + squash when they are fixed.

PR #400, opened 2020-12-24
Extend packages for customized outputs and their control

clellsolomon:
## PR Summary

Extends `StateDescriptor`, `Mesh`, `Driver`, and `ApplicationInput` to support pre- and post-step user operations.  These could be used, for example, to provide custom edits on a per-package basis before and or after a step.  

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Current develop
 
 ### Added (new features/APIs/variables/...)
-- [[PR 400]](https://github.com/lanl/parthenon/pull/400) Extend `StateDescriptor` for customizable output via user-customizable function pointer `OutputDiagnosticsMesh`
+- [[PR 400]](https://github.com/lanl/parthenon/pull/400) Extend `StateDescriptor` for customizable output via user-customizable function pointers `PreStepDiagnosticsMesh` and `PostStepDiagnosticsMesh`
 - [[PR 391]](https://github.com/lanl/parthenon/pull/391) Add `VariablePack<T>::GetSparseId` and `VariablePack<T>::GetSparseIndex` to return global sparse ids and pack-local sparse index, repsectively.
 - [[PR 381]](https://github.com/lanl/parthenon/pull/381) Overload `DataCollection::Add` to build `MeshData` and `MeshBlockData` objects with a subset of variables.
 - [[PR 378]](https://github.com/lanl/parthenon/pull/378) Add Kokkos profiling regions throughout the code to allow the collection characteristic application profiles
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Current develop
 
 ### Added (new features/APIs/variables/...)
+- [[PR 400]](https://github.com/lanl/parthenon/pull/400) Extend `StateDescriptor` for customizable output via user-customizable function pointer `OutputDiagnosticsMesh`
 - [[PR 391]](https://github.com/lanl/parthenon/pull/391) Add `VariablePack<T>::GetSparseId` and `VariablePack<T>::GetSparseIndex` to return global sparse ids and pack-local sparse index, repsectively.
 - [[PR 381]](https://github.com/lanl/parthenon/pull/381) Overload `DataCollection::Add` to build `MeshData` and `MeshBlockData` objects with a subset of variables.
 - [[PR 378]](https://github.com/lanl/parthenon/pull/378) Add Kokkos profiling regions throughout the code to allow the collection characteristic application profiles
@@ -12,6 +13,7 @@
 - [[PR 386]](https://github.com/lanl/parthenon/pull/386) Introduce `Private`, `Provides`, `Requires`, and `Overridable` variable metadata, allowing fine-grained control of conflict resolution between packages.
 
 ### Changed (changing behavior/API/variables/...)
+- [[PR 400]](https://github.com/lanl/parthenon/pull/400) Change `Mesh`, `ApplicationInput`, and `Driver` to suppport pre- and post- step user work
 - [[PR 394]](https://github.com/lanl/parthenon/pull/332) Make `Params.Get` const-correct.
 - [[PR 332]](https://github.com/lanl/parthenon/pull/332) Rewrote boundary conditions to work on GPUs with variable packs. Re-enabled user-defined boundary conditions via `ApplicationInput`.
 
--- a/src/application_input.hpp
+++ b/src/application_input.hpp
@@ -37,10 +37,10 @@ struct ApplicationInput {
 
   // Mesh functions
   std::function<void(ParameterInput *)> InitUserMeshData = nullptr;
-  std::function<void(Mesh *, ParameterInput *, SimTime const &)> PreStepMeshUserWorkInLoop =
-      nullptr;
-  std::function<void(Mesh *, ParameterInput *, SimTime const &)> PostStepMeshUserWorkInLoop =
-      nullptr;
+  std::function<void(Mesh *, ParameterInput *, SimTime const &)>
+      PreStepMeshUserWorkInLoop = nullptr;
+  std::function<void(Mesh *, ParameterInput *, SimTime const &)>
+      PostStepMeshUserWorkInLoop = nullptr;
   std::function<void(Mesh *, ParameterInput *, SimTime &)> UserWorkAfterLoop = nullptr;
   BValFunc boundary_conditions[BOUNDARY_NFACES] = {nullptr, nullptr, nullptr,
                                                    nullptr, nullptr, nullptr};
--- a/src/application_input.hpp
+++ b/src/application_input.hpp
@@ -37,10 +37,17 @@ struct ApplicationInput {
 
   // Mesh functions
   std::function<void(ParameterInput *)> InitUserMeshData = nullptr;
+
   std::function<void(Mesh *, ParameterInput *, SimTime const &)>
       PreStepMeshUserWorkInLoop = nullptr;
   std::function<void(Mesh *, ParameterInput *, SimTime const &)>
       PostStepMeshUserWorkInLoop = nullptr;
+
+  std::function<void(Mesh *, ParameterInput *, SimTime const &)>
+      PreStepDiagnosticsInLoop = nullptr;
+  std::function<void(Mesh *, ParameterInput *, SimTime const &)>
+      PostStepDiagnosticsInLoop = nullptr;
+
   std::function<void(Mesh *, ParameterInput *, SimTime &)> UserWorkAfterLoop = nullptr;
   BValFunc boundary_conditions[BOUNDARY_NFACES] = {nullptr, nullptr, nullptr,
                                                    nullptr, nullptr, nullptr};
--- a/src/application_input.hpp
+++ b/src/application_input.hpp
@@ -37,7 +37,8 @@ struct ApplicationInput {
 
   // Mesh functions
   std::function<void(ParameterInput *)> InitUserMeshData = nullptr;
-  std::function<void()> MeshUserWorkInLoop = nullptr;
+  std::function<void(Mesh *, ParameterInput *, SimTime &)> PreStepMeshUserWorkInLoop = nullptr;
+  std::function<void(Mesh *, ParameterInput *, SimTime &)> PostStepMeshUserWorkInLoop = nullptr;
   std::function<void(Mesh *, ParameterInput *, SimTime &)> UserWorkAfterLoop = nullptr;
   BValFunc boundary_conditions[BOUNDARY_NFACES] = {nullptr, nullptr, nullptr,
                                                    nullptr, nullptr, nullptr};
--- a/src/application_input.hpp
+++ b/src/application_input.hpp
@@ -37,8 +37,10 @@ struct ApplicationInput {
 
   // Mesh functions
   std::function<void(ParameterInput *)> InitUserMeshData = nullptr;
-  std::function<void(Mesh *, ParameterInput *, SimTime &)> PreStepMeshUserWorkInLoop = nullptr;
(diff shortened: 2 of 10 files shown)


src/application_input.hpp
```diff
@@ -37,7 +37,10 @@ struct ApplicationInput {
 
   // Mesh functions
   std::function<void(ParameterInput *)> InitUserMeshData = nullptr;
-  std::function<void()> MeshUserWorkInLoop = nullptr;
+  std::function<void(Mesh *, ParameterInput *, SimTime &)> PreStepMeshUserWorkInLoop =
```
clellsolomon: Question: Does it break anything to take the `Mesh` and `ParameterInput` by references?  Although this is inconsistent with the some of the other functions, the should be required to exist (be non-`NULL`) and, I think, a reference would better capture that requirement.

Yurlungur: Passing by reference wouldn't break anything. Pointers are passed around everywhere mostly out of convenience and to follow the design of Athena++ on which Parthenon is based.

Mesh is created and held as a `std::unique_ptr` I believe, so `pmesh.get()` makes sense as a thing to pass around. But so would `*pmesh`.

Yurlungur:
Thanks for this, @clellsolomon per-package work before/after loop is a good idea.

src/application_input.hpp
clellsolomon: Would the preference be to leave it consistent (as it is, I think) with the Athena++ style?

Yurlungur:
One thing that occurs to me is we might want an application level setting too, in `ApplicationInputs`.

clellsolomon:
@Yurlungur: I'm not quite sure what you mean by

> One thing that occurs to me is we might want an application level setting too, in `ApplicationInputs`.

The `ApplicationInputs` do hold `std::functions` for these.  Can you please clarify?

src/application_input.hpp
Yurlungur: Yeah, I think the preference would be leave it consistent for now. But it's also not a huge deal either way.

Yurlungur:
> @Yurlungur: I'm not quite sure what you mean by
> 
> > One thing that occurs to me is we might want an application level setting too, in `ApplicationInputs`.
> 
> The `ApplicationInputs` do hold `std::functions` for these. Can you please clarify?

Sorry my bad. That was not clear. I will explain in a review.

src/application_input.hpp
```diff
@@ -37,7 +37,10 @@ struct ApplicationInput {
 
   // Mesh functions
   std::function<void(ParameterInput *)> InitUserMeshData = nullptr;
-  std::function<void()> MeshUserWorkInLoop = nullptr;
+  std::function<void(Mesh *, ParameterInput *, SimTime const &)>
+      PreStepMeshUserWorkInLoop = nullptr;
+  std::function<void(Mesh *, ParameterInput *, SimTime const &)>
+      PostStepMeshUserWorkInLoop = nullptr;
```
Yurlungur, line 44: My thought earlier was maybe there should be an application-wide `OutputDiagnostics` which could go here in `ApplicationInput` as well as the `UserWorkInLoop` stuff.

This is optional though. I think having package-level diagnostics and working application-level work in loop is already a major improvement.

src/interface/state_descriptor.hpp
```diff
@@ -188,6 +183,10 @@ class StateDescriptor {
     if (FillDerivedMesh != nullptr) FillDerivedMesh(rc);
   }
 
+  void OutputDiagnostics(SimTime const &simtime, MeshData<Real> *rc) const {
+    if (OutputDiagnosticsMesh != nullptr) OutputDiagnosticsMesh(simtime, rc);
+  }
+
```
Yurlungur: I like the addition of `OutputDiagnostics` at the package level. But I notice it's not actually called anywhere in the driver/mesh. Maybe we could add a loop over packages at the driver level (either before or after loop) to spit out the diagnostics, so that the application doesn't have to write that bit of boiler plate?

Yurlungur (approved):
Thanks for this @clellsolomon this is definitely an improvement. A few minor comments below.

One other comment is that I notice we don't have any documentation on how to set `ApplicationData` or the function pointers in `StateDescriptor` at the moment. If you're feeling motivated, that would be a welcome addition to this PR too. But if not, we can add that later---it's not a big deal.

src/interface/state_descriptor.hpp
clellsolomon: Thanks for the review @Yurlungur.   I suppose this boils down to whether or not we want to "impose" this to be some sort of default behavior.  I had expected that the downstream use would chose if/how/when to output any package diagnostics.  I'm not opposed to some default call to all packages' `OutputDiagnostics` somewhere, but I'm not entirely sure where it should go.  Can you offer guidance?

src/application_input.hpp
clellsolomon: Ok, I have added `Pre`/`PostStepDiagnostics` for the application.  The mesh has default implementation that cycle each package and call their `Pre`/`PostStepDiagnostics` functions.

src/interface/state_descriptor.hpp
clellsolomon: Now has default implementation to cycle the packages and all the `Pre`/`PostStepDiagnostics`.

Yurlungur (approved):
Thanks for making these changes, @clellsolomon ! This looks great!

src/application_input.hpp
```diff
@@ -37,7 +37,17 @@ struct ApplicationInput {
 
   // Mesh functions
   std::function<void(ParameterInput *)> InitUserMeshData = nullptr;
```
jlippuner, line 39: Don't we also need to pass a pointer to `Mesh` to `InitUserMeshData`? Otherwise there's no way to actually manipulate the mesh, is there?

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Current develop
 
 ### Added (new features/APIs/variables/...)
+- [[PR 400]](https://github.com/lanl/parthenon/pull/400) Extend `StateDescriptor` for customizable output via user-customizable function pointer `OutputDiagnosticsMesh`
```
jlippuner: We now have pre- and post-step diagnostic hooks, right?

src/mesh/mesh.hpp
```diff
@@ -147,8 +147,22 @@ class Mesh {
...
   std::function<void(Mesh *, ParameterInput *, SimTime &)> UserWorkAfterLoop =
       &UserWorkAfterLoopDefault;
-  static void UserWorkInLoopDefault(); // called in main after each cycle
-  std::function<void()> UserWorkInLoop = &UserWorkInLoopDefault;
+  static void UserWorkInLoopDefault(Mesh *, ParameterInput *,
+                                    SimTime const &); // called in main after each cycle
+  std::function<void(Mesh *, ParameterInput *, SimTime const &)> PreStepUserWorkInLoop =
+      &UserWorkInLoopDefault;
+  std::function<void(Mesh *, ParameterInput *, SimTime const &)> PostStepUserWorkInLoop =
+      &UserWorkInLoopDefault;
```
jlippuner: Could this lead to some hard to debug unexpected behavior? If I understand this correctly, if the user doesn't explicitly specify `PreStepMeshUserWorkInLoop` or `PostStepMeshUserWorkInLoop`, the default is that those call `UserWorkInLoopDefault`. But that means that `UserWorkInLoopDefault` would be called **twice** per time step. This could have unintended side effects. Or am I missing something?

I think we should either only call `UserWorkInLoopDefault` by default as `PostStepUserWorkInLoop` and have `PreStepUserWorkInLoop` do nothing by default, or split up `UserWorkInLoopDefault` into a pre and post default.

src/pgen/default_pgen.cpp
```diff
@@ -49,11 +49,25 @@ void Mesh::InitUserMeshDataDefault(ParameterInput *pin) {
 //  \brief Function called once every time step for user-defined work.
 //========================================================================================
 
-void Mesh::UserWorkInLoopDefault() {
+void Mesh::UserWorkInLoopDefault(Mesh *, ParameterInput *, SimTime const &) {
   // do nothing
   return;
 }
```
jlippuner, line 55: See comment in `mesh.hpp` about this potentially being called twice per time step (as opposed to what the comment says).

jlippuner (changes requested):
This is a very important improvement that makes the infrastructure a lot more flexible. Please see my inline comments for details.

Additionally, I think we also need to update the Meshblock user hooks. They are now somewhat out of sync with the Mesh hooks (Mesh now has pre and post step work in loop, but block only has `MeshUserWorkInLoop`, same for output/diagnostic hooks). And also the function interface to the Meshblock hooks is not complete, we should probably pass the `ParameterInput` pointer to all of them and they should all also get a pointer to the `MeshBlock` they're operating on.

I think making those changes to the Meshblock hooks would fit well into this PR, but I'm also ok with making them in a separate PR. I'd also be happy to contribute those changes myself either to this PR or a new PR, just let me know.

CHANGELOG.md
clellsolomon: yes

src/application_input.hpp
clellsolomon: This line wasn't changed, but I don't know off the top of my head.  Maybe something to look at?

src/mesh/mesh.hpp
clellsolomon: I don't think you are missing anything.  I didn't feel the need for separate `Pre`- and `PostStepMeshUserWorkInLoop` because the default is a no-op.  If there is a reason to have different default behaviors, then I would definitely agree that there should be separate defaults.  At a minimum I agree the offending comment should be removed. 

Alternatively, we could initialize to `nullptr` instead of setting he default no-op function. 
 Obviously, this would require some check before calling.  We have this pattern in the `StateDescriptor` but seem to use default no-op functions here.  :shrug:

clellsolomon (commented):
Thanks for your review @jlippuner.  The questions you pose are good ones.  This PR got a bit convoluted in that I had originally intended to do diagnostics work as part of the `PreStepMeshUserWorkInLoop` and `PostStepMeshUserWorkInLoop` and then broke the diagnostics out separately so that reasonable defaults could be implemented without requiring a coupling to other work that one might want to perform over the mesh.  Please let me know what you think regarding may comments about the default `PreStepMeshUserWorkInLoop` and `PostStepMeshUserWorkInLoop`.

I see your point about the equivalent mesh block functions, but I think I'd prefer to defer it to another PR.  I think you could do mesh block work through the equivalent mesh functions because the mesh has access to its blocks.  Without a more prevalent use case, I'm just not sure which option makes more sense.  Thoughts?

src/application_input.hpp
Yurlungur: I'm not sure this is fully hooked into the infrastructure. I'd leave it for now and we can fix it when there's a need.

src/mesh/mesh.hpp
Yurlungur: Choose your favorite path. We haven't been consistent.

src/application_input.hpp
jlippuner: That's fine with me

src/mesh/mesh.hpp
jlippuner: Actually, I think you're right that it is not an issue. It's only `UserWorkInLoopDefault` that is called twice, but the user has no way of overriding that default function. I was concerned that an unsuspecting user would implement that method and then it gets executed twice, but for that to happen, they would have to assign their function to both `PreStepUserWorkInLoop` and `PostStepUserWorkInLoop` and then they would want it to be executed twice per time step. So I think we're all good.

jlippuner (approved):
Thanks @clellsolomon for addressing my comments. I agree with doing the Meshblock hooks in a different PR and maybe removing some of them altogether since one has access to all Meshblocks from the Mesh hooks.

src/driver/driver.cpp
```diff
@@ -61,17 +61,23 @@ DriverStatus EvolutionDriver::Execute() {
...
-    // pmesh->UserWorkInLoop();
+
     tm.ncycle++;
     tm.time += tm.dt;
     pmesh->mbcnt += pmesh->nbtotal;
     pmesh->step_since_lb++;
 
+    pmesh->PostStepUserWorkInLoop(pmesh, pinput, tm);
+    pmesh->PostStepUserDiagnosticsInLoop(pmesh, pinput, tm);
+
```
pgrete: Are the function call after 
```
    tm.ncycle++;
    tm.time += tm.dt;
```
on purpose?
Conceptually, I would put them before because in case I add have some userwork that should be done in a specific cycle/time, I could filter for the same cycle/time in both pre and post functions, e.g.,
```
PreStepUserDiagnosticsInLoop (pmesh, pinput, tm) {
  if (tm.ncycle % 10 == 0) {
    my_magic_diagnostic_before_step();
  }
}
```
and PostStep correspondingly.

Otherwise, I'd need `ncycle` and `ncycle+1`, which I personally find less intuitive.

pgrete (changes requested):
Thanks for those changes. This functionality was definitely missing.
Apart from the comment I left inline I'm wondering if the documentation was not committed  yet or if I'm blind (as it's marked of in the PR checklist).
Finally, I'm wondering how much effort it would be to add some small test functions to one of the example to illustrate usage of this code.

src/driver/driver.cpp
clellsolomon: I think I did this because the diagnostic and work is occurring after the step has happened, so the time the problem state has reached is "effectively" at the updated time.  But, one could use the `PreStep` stuff on the next cycle to do that provided the problem hasn't terminated.

clellsolomon: @pgrete : I have moved those calls to before the update to time and cycle.

pgrete (approved):
Looks good! Thanks for addressing the comments.

PR #402, opened 2020-12-29
Fix initital timestep calculation for `MeshData` functions

### LINKED ISSUE #401: Initial timestep not calculated for `MeshData` based functions
For apps using `MeshData`based `EstimateTimestep` functions (rather than `MeshBlockData` based) the initial timestep is not calculated.

pgrete:
## PR Summary

Takes `MeshData` based functions into account when calculating the first timestep.

Fixes #401 

## PR Checklist

- [x] Code passes cpplint
- [does not apply] New features are documented.
- [?] Adds a test for any bugs fixed. Adds tests for new features.
- [x] Code is formatted

This "bug" made me wonder if we should split/duplicate/extend the advection example so that we test both the `MeshData` and `MeshBlockData` based machinery in Parthenon.
Maybe @jlippuner planned proxy app (going for AMR performance with small MeshBlocks where we definitely need the `MeshData` functions could exclusively use the `MeshData` framework and we keep the current advection example to only use the `MeshBlockData` based interface.

Any opinions?

### DIFF
--- a/src/interface/mesh_data.hpp
+++ b/src/interface/mesh_data.hpp
@@ -13,6 +13,7 @@
 #ifndef INTERFACE_MESH_DATA_HPP_
 #define INTERFACE_MESH_DATA_HPP_
 
+#include <algorithm>
 #include <limits>
 #include <map>
 #include <memory>
@@ -88,7 +89,7 @@ class MeshData {
 
   void SetAllowedDt(const Real dt) const {
     for (const auto &pbd : block_data_) {
-      pbd->SetAllowedDt(dt);
+      pbd->SetAllowedDt(std::min(dt, pbd->GetBlockPointer()->NewDt()));
     }
   }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 - [[PR 332]](https://github.com/lanl/parthenon/pull/332) Rewrote boundary conditions to work on GPUs with variable packs. Re-enabled user-defined boundary conditions via `ApplicationInput`.
 
 ### Fixed (not changing behavior/API/variables/...)
+- [[\#401]](https://github.com/lanl/parthenon/issues/401) Fix missing initial timestep for MeshData functions
 - [[PR 387]](https://github.com/lanl/parthenon/pull/387) Add missing const that was needed
 - [[PR 353]](https://github.com/lanl/parthenon/pull/353) Fixed small error in input\_parameter logic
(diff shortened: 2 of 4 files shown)


Yurlungur (approved):
Thanks for this, @pgrete . I do think it would be a good idea to split the `advection` example in the way you suggest.

But one thing I don't understand... shouldn't the `dt` from `MeshData` be the same as the minimum `dt` over `MeshBlocks`s? Also, should we perhaps offer the ability to only use `MeshData` or only use `MeshBlockData` for this loop, to potentially save some kernel launches?

src/interface/mesh_data.hpp
```diff
@@ -88,7 +89,7 @@ class MeshData {
 
   void SetAllowedDt(const Real dt) const {
```
JoshuaSBrown, line 90: I feel like the name of this method needs to be changed because it is no longer just a set. E.g. ConsolidateAllowedDt though even that is not entirely clear either.

JoshuaSBrown (approved):
Other than possibly adjusting the name of the set method this looks good to me. Thanks for fixing this!

src/interface/mesh_data.hpp
pgrete: I agree (and would even go further as some naming doesn't seem to match the implementation any more).
Should I fix this here or in a separate PR?

pgrete:
> But one thing I don't understand... shouldn't the `dt` from `MeshData` be the same as the minimum `dt` over `MeshBlocks`s?

I assume that you're referring to
```diff
    for (const auto &pbd : block_data_) {
-      pbd->SetAllowedDt(dt);
+      pbd->SetAllowedDt(std::min(dt, pbd->GetBlockPointer()->NewDt()));
    }
  }
```

This is required because the default `dt` is the `numeric_limit<Real>`, thus, if the existing new timestep contained in `NewDt()` is not honored then it would be set to the max value (effectively ignoring the new one).

> Also, should we perhaps offer the ability to only use `MeshData` or only use `MeshBlockData` for this loop, to potentially save some kernel launches?

Are kernel launches here refer to function calls? If a function is not enrolled then nothing should happen (though I'm not sure if the compiler is smart enough to skip the entire loop completely in that case).

src/interface/mesh_data.hpp
JoshuaSBrown: I'm fine if you merge and open a WIP pr to fix it.

pgrete: Sounds good. I'll merge then (one thing off the table).

PR #406, opened 2021-01-06
Add stochastic subgrid example

jlippuner:
## PR Summary

Add an example that performs dummy FLOP work for a random number of iterations in each cell. The number of iterations is sampled from a power-law distribution whose coefficient and minimum and maximum numbers of iterations can be specified in the input file.

## PR Checklist

- [x] Code passes cpplint
- [n/a] Adds a test for any bugs fixed. Adds tests for new features.
- [x] Code is formatted


jlippuner:
@Yurlungur @jdolence This is ready for you to take a look and see if the dummy work is adequate for our purposes (see function `DoLotsOfWork` on line 338 of `advanced_advection_package.cpp`. Thanks!

example/advanced_advection/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+# for the U.S. Department of Energy/National Nuclear Security Administration. All rights
+# in the program are reserved by Triad National Security, LLC, and the U.S. Department
+# of Energy/National Nuclear Security Administration. The Government is granted for
+# itself and others acting on its behalf a nonexclusive, paid-up, irrevocable worldwide
+# license in this material to reproduce, prepare derivative works, distribute copies to
+# the public, perform publicly and display publicly, and to permit others to do so.
+#=========================================================================================
+
+get_property(DRIVER_LIST GLOBAL PROPERTY DRIVERS_USED_IN_TESTS)
+if( "advanced_advection-example" IN_LIST DRIVER_LIST OR NOT PARTHENON_DISABLE_EXAMPLES)
```
Yurlungur: I don't love the name `advanced_advection-example`. Could we come up with something slightly more descriptive? Like `heterogeneous-advection-example` or something?

example/advanced_advection/advanced_advection_driver.cpp
```diff
@@ -0,0 +1,235 @@
...
+      auto &mc1 = pmesh->mesh_data.GetOrAdd(stage_name[stage], i);
+      tr[i].AddTask(none, parthenon::cell_centered_bvars::SetBoundaries, mc1);
+    }
+  } else {
+    TaskRegion &tr = tc.AddRegion(num_task_lists_executed_independently);
+    for (int i = 0; i < blocks.size(); i++) {
+      auto &sc1 = blocks[i]->meshblock_data.Get(stage_name[stage]);
+      tr[i].AddTask(none, &MeshBlockData<Real>::SetBoundaries, sc1.get());
+    }
+  }
```
Yurlungur: I'm not sure what to do about this if logic here. It's in the simple advection example. But for the more complicated one, I vote we pick one... if you're using sparse variables, I say no packs. If you're not using sparse variables, I saw do the packs.

example/advanced_advection/advanced_advection_package.cpp
```diff
@@ -0,0 +1,523 @@
...
+    lambda = std::min(lambda, x2);
+  if ((pin->GetOrAddInteger("parthenon/mesh", "nx3", 1) > 1) && ang_2 != 0.0)
+    lambda = std::min(lambda, x3);
+
+  // If cos_a2 or cos_a3 = 0, need to override lambda
+  if (ang_3_vert) lambda = x2;
+  if (ang_2_vert) lambda = x3;
+
+  // Initialize k_parallel
+  Real k_par = 2.0 * (M_PI) / lambda;
```
Yurlungur: I'm not sure this actually works... should lambda not be the diagonal accross the domain? I.e., it should be a pythagorean rule?

example/advanced_advection/advanced_advection_package.cpp
```diff
@@ -0,0 +1,523 @@
...
+  pkg->AddParam<>("cos_a2", cos_a2);
+  pkg->AddParam<>("cos_a3", cos_a3);
+  pkg->AddParam<>("sin_a2", sin_a2);
+  pkg->AddParam<>("sin_a3", sin_a3);
+
+  // initialize the alias tables to sample the number of iterations from a discrete power
+  // law distributions, for description of the Alias Method, see
+  // https://en.wikipedia.org/wiki/Alias_method
+  // https://www.keithschwarz.com/darts-dice-coins/
+  // https://gist.github.com/Liam0205/0b5786e9bfc73e75eb8180b5400cd1f8
```
Yurlungur: Why do you hand implement this instead of using the stdlib? Just curious.

example/advanced_advection/advanced_advection_package.cpp
```diff
@@ -0,0 +1,523 @@
...
+  const IndexRange ib = pack.cellbounds.GetBoundsI(IndexDomain::interior);
+  const IndexRange jb = pack.cellbounds.GetBoundsJ(IndexDomain::interior);
+  const IndexRange kb = pack.cellbounds.GetBoundsK(IndexDomain::interior);
+
+  auto prob_table = pkg->Param<Kokkos::View<Real *>>("prob_table");
+  auto alias_table = pkg->Param<Kokkos::View<int *>>("alias_table");
+  int N_min = pkg->Param<int>("N_min");
+
+  par_for(
+      parthenon::loop_pattern_mdrange_tag, "ComputeNumIter", parthenon::DevExecSpace(), 0,
```
Yurlungur: Why the mdrange loop pattern tag?

example/advanced_advection/advanced_advection_package.cpp
```diff
@@ -0,0 +1,523 @@
...
+        for (int r = 0; r < num_iter; ++r) {
+          double odd = 0.0;
+          double even = 0.0;
+
+          for (int n = 0; n < num_vars; ++n)
+            (n % 2 == 0 ? even : odd) += sqrt(n + 1) * v(in + n, k, j, i);
+
+          double a = exp((odd + even) / (fmax(1.0, abs(odd * even))));
+          double b = exp((odd - even) / (fmax(1.0, abs(odd * even))));
+          v(out, k, j, i) += log(a * b) / (log(a) + log(b));
```
Yurlungur: I'm not sure if this matters, but it might be good to change these exponentials to `pow(10,x)` and the logarithms to `log10`. Riot uses base 10... and if I recall, that's slower.

Actually to be most precise, riot uses `logf(x)*ILOG10` where `constexpr Real ilog10 = 1./log(10);`

src/pgen/default_pgen.cpp
```diff
@@ -89,7 +89,8 @@ void Mesh::UserWorkAfterLoopDefault(Mesh *mesh, ParameterInput *pin, SimTime &tm
 //========================================================================================
 
 std::unique_ptr<MeshBlockApplicationData>
-MeshBlock::InitApplicationMeshBlockDataDefault(ParameterInput *pin) {
+MeshBlock::InitApplicationMeshBlockDataDefault(MeshBlock * /*pmb*/,
```
Yurlungur, line 92: good catch regarding these

Yurlungur (approved):
Thanks for this, @jlippuner ! Overall, this looks great! I have a few comments/questions/suggestions, but they're all non-blocking.

Yurlungur:
@jlippuner some profiling information for FillDerrived in RIOT. The following should be considered a lower bound on cost, but I think it's reasonable. The average number of iterations (in the test problem I ran) is about 4 with a long tail.
- On CPU, we have about 1 solve/ms.
- On GPU, we have about 133 solves/ms.

Note these times are over all cells. e.g., I took total time and divide by total number of cells.

example/advanced_advection/advanced_advection_package.cpp
jlippuner: Good question. I just copied this from the `advection` example. Maybe @pgrete can shed some light on this since it appears he wrote this code in the `advection` example.

example/advanced_advection/advanced_advection_package.cpp
jlippuner: Excellent question. I'm always for using stdlib and not reinventing (or re-implementing) the wheel. I did look at `std::discrete_distribution` and considered using it, but eventually decided to implement the alias method myself for several reasons:

1. It's not clear to me which method `std::discrete_distribution` uses internally. It says it's `operator()` method is amortized constant, so it could be less efficient than the alias method, which is guaranteed constant.
2. It's also not clear to me if `std::discrete_distribution` keeps any dynamic internal state and thus if it is thread-safe, which we would need, otherwise we'd have to make many copies of it, one for each thread.
3. Most importantly, since we sample the number of iterations on the device, we would need `std::discrete_distribution` to work on the GPU in parallel. I didn't try it and it's possible that it works, but I decided I'd rather know exactly what's happening on the GPU and how much memory we use. With the alias method implemented ourselves, we just read from two small `Kokkos::view`s (the probability and alias tables) of which we only need one copy and the actual sampling just involves two table look ups without any dynamic state, so we know it's thread-safe (given that Kokkos' random number generator is thread-safe) and lightweight on the device.

It may be interesting to try to use `std::discrete_distribution` on the GPU, but I suspect we will get some additional UVM transfers because we can't tell `std::discrete_distribution` to allocate its internal state on the device, as we can with `Kokkos::view`s.

example/advanced_advection/advanced_advection_package.cpp
jlippuner: No particular reason. That's just what I saw in other examples. Should I use something else?

example/advanced_advection/advanced_advection_package.cpp
pgrete: I recycled the code from the Athena++ pgen so unless I made a copy/paste error this should work.

example/advanced_advection/advanced_advection_package.cpp
jlippuner: I'm happy to change to base 10. For the log, should I use `log10` or multiply by `1/log(10)`? Where do you get the `constexpr` version of the `log` function to define the constant `ilog10`? When I try to do this, it complains that `cannot call non-constexpr function "log(double)"`

example/advanced_advection/advanced_advection_package.cpp
Yurlungur: Ok. I must not fully understand what's going on here. If it works in advection and Athena++ it's probably fine.

example/advanced_advection/advanced_advection_package.cpp
Yurlungur: I see. Interesting.

Yeah I don't think `std::discrete_distribution` would work on GPU. I think the strategy there would be to call it on CPU and precompute random numbers, and then fill a kokkos view with them.

Would it be worth saving this alias method to a `parthenon` utility so it's available in the future and to downstream codes?

example/advanced_advection/advanced_advection_package.cpp
Yurlungur: If it runs without specifying a loop pattern, we probably want to not specify one and let the default tag do its work. But that might depend on the dimension of the loop.... I think only only `mdrange` is supported for 5D and 6D loops. (Which is something we should get around to fixing at some point.)

Yurlungur: Sorry I forgot about that detail until just now. I guess let's leave this for now. We can fix it when we add support for 5d and 6d loops with the other tags.

example/advanced_advection/advanced_advection_package.cpp
Yurlungur: I think I computed `1/log(10)` and then copy-pasted the number to get around the lack of a constexpr for log. Let's just multiply by `1/log(10)`. That probably is sufficient.

example/advanced_advection/advanced_advection_package.cpp
jlippuner: I think that's a good idea. I'll split it off.

example/advanced_advection/CMakeLists.txt
jlippuner: Thanks @Yurlungur for the suggestion `stochastic_subgrid`!

jlippuner:
This is good to go once CI passes. I need one more review. @pgrete @JoshuaSBrown @cielling ?

example/stochastic_subgrid/parthenon_app_inputs.cpp
```diff
@@ -0,0 +1,255 @@
...
+  const auto &sin_a3 = pkg->Param<Real>("sin_a3");
+  const auto &profile = pkg->Param<std::string>("profile");
+
+  auto q_h = q.GetHostMirror();
+
+  auto cellbounds = pmb->cellbounds;
+
+  IndexRange ib = cellbounds.GetBoundsI(IndexDomain::entire);
+  IndexRange jb = cellbounds.GetBoundsJ(IndexDomain::entire);
+  IndexRange kb = cellbounds.GetBoundsK(IndexDomain::entire);
```
JoshuaSBrown: Can be made const.

example/stochastic_subgrid/parthenon_app_inputs.cpp
```diff
@@ -0,0 +1,255 @@
...
+      const auto &k_par = pkg->Param<Real>("k_par");
+      const auto &cos_a2 = pkg->Param<Real>("cos_a2");
+      const auto &cos_a3 = pkg->Param<Real>("cos_a3");
+      const auto &sin_a2 = pkg->Param<Real>("sin_a2");
+      const auto &sin_a3 = pkg->Param<Real>("sin_a3");
+      const auto &profile = pkg->Param<std::string>("profile");
+
+      IndexRange ib = pmb->cellbounds.GetBoundsI(IndexDomain::interior);
+      IndexRange jb = pmb->cellbounds.GetBoundsJ(IndexDomain::interior);
+      IndexRange kb = pmb->cellbounds.GetBoundsK(IndexDomain::interior);
```
JoshuaSBrown: Can be made const.

example/stochastic_subgrid/parthenon_app_inputs.cpp
```diff
@@ -0,0 +1,255 @@
...
+
+  auto q_h = q.GetHostMirror();
+
+  auto cellbounds = pmb->cellbounds;
+
+  IndexRange ib = cellbounds.GetBoundsI(IndexDomain::entire);
+  IndexRange jb = cellbounds.GetBoundsJ(IndexDomain::entire);
+  IndexRange kb = cellbounds.GetBoundsK(IndexDomain::entire);
+
+  auto coords = pmb->coords;
```
JoshuaSBrown: const reference?

example/stochastic_subgrid/stochastic_subgrid_package.cpp
```diff
@@ -0,0 +1,462 @@
...
+// *************************************************//
+// define the "physics" package Advect, which      *//
+// includes defining various functions that control*//
+// how parthenon functions and any tasks needed to *//
+// implement the "physics"                         *//
+// *************************************************//
+
+namespace stochastic_subgrid_package {
+
+std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
```
JoshuaSBrown, line 45: This is a large function I would recommend breaking this up into smaller local functions in an unnames namespace. I think a lot of the variables in this function can be declared const, that would help clarify where things are being changed vs being used.

example/stochastic_subgrid/stochastic_subgrid_package.cpp
```diff
@@ -0,0 +1,462 @@
...
+  pkg->CheckRefinementBlock = CheckRefinement;
+  pkg->EstimateTimestepBlock = EstimateTimestepBlock;
+
+  return pkg;
+}
+
+AmrTag CheckRefinement(MeshBlockData<Real> *rc) {
+  auto pmb = rc->GetBlockPointer();
+  // refine on advected, for example.  could also be a derived quantity
+  auto v = rc->Get("advected").data;
```
JoshuaSBrown, line 217: Can `v` be made const reference?

example/stochastic_subgrid/stochastic_subgrid_package.cpp
```diff
@@ -0,0 +1,462 @@
...
+
+  IndexRange ib = pmb->cellbounds.GetBoundsI(IndexDomain::entire);
+  IndexRange jb = pmb->cellbounds.GetBoundsJ(IndexDomain::entire);
+  IndexRange kb = pmb->cellbounds.GetBoundsK(IndexDomain::entire);
```
JoshuaSBrown: make const.

example/stochastic_subgrid/stochastic_subgrid_package.cpp
```diff
@@ -0,0 +1,462 @@
...
+
+// this is the package registered function to fill derived
+void DoLotsOfWork(MeshBlockData<Real> *rc) {
+  auto pmb = rc->GetBlockPointer();
+  auto pkg = pmb->packages.Get("stochastic_subgrid_package");
+
+  IndexRange ib = pmb->cellbounds.GetBoundsI(IndexDomain::entire);
+  IndexRange jb = pmb->cellbounds.GetBoundsJ(IndexDomain::entire);
+  IndexRange kb = pmb->cellbounds.GetBoundsK(IndexDomain::entire);
+
```
JoshuaSBrown: make const

example/stochastic_subgrid/stochastic_subgrid_package.cpp
```diff
@@ -0,0 +1,462 @@
...
+  auto pmb = rc->GetBlockPointer();
+  auto pkg = pmb->packages.Get("stochastic_subgrid_package");
+  const auto &cfl = pkg->Param<Real>("cfl");
+  const auto &vx = pkg->Param<Real>("vx");
+  const auto &vy = pkg->Param<Real>("vy");
+  const auto &vz = pkg->Param<Real>("vz");
+
+  IndexRange ib = pmb->cellbounds.GetBoundsI(IndexDomain::interior);
+  IndexRange jb = pmb->cellbounds.GetBoundsJ(IndexDomain::interior);
+  IndexRange kb = pmb->cellbounds.GetBoundsK(IndexDomain::interior);
```
JoshuaSBrown: const

example/stochastic_subgrid/stochastic_subgrid_package.cpp
```diff
@@ -0,0 +1,462 @@
...
+
+  auto &coords = pmb->coords;
```
JoshuaSBrown: const

example/stochastic_subgrid/parthinput.stochastic_subgrid
```diff
@@ -0,0 +1,91 @@
...
+nx2 = 16
+nx3 = 1
+
+<parthenon/time>
+nlim = -1
+tlim = 1.0
+integrator = rk2
+
+<Advection>
+cfl = 0.45
```
JoshuaSBrown, line 54: What is cfl? Could we use something a little more descriptive?

example/stochastic_subgrid/stochastic_subgrid_driver.cpp
```diff
@@ -0,0 +1,204 @@
...
+    }
+  }
+
+  {
+    TaskRegion &tr = tc.AddRegion(num_partitions);
+    for (int i = 0; i < num_partitions; i++) {
+      auto &mc1 = pmesh->mesh_data.GetOrAdd(stage_name[stage], i);
+      tr[i].AddTask(none, parthenon::cell_centered_bvars::SetBoundaries, mc1);
+    }
+  }
```
JoshuaSBrown: There is a lot of repetition here, I would recommend placing in a local function.

example/stochastic_subgrid/stochastic_subgrid_driver.cpp
```diff
@@ -0,0 +1,204 @@
...
+  pin->CheckDesired("Advection", "vx");
+  pin->CheckDesired("Advection", "refine_tol");
+  pin->CheckDesired("Advection", "derefine_tol");
+}
+
+using stochastic_subgrid_package::ComputeNumIter;
+
+// See the advection.hpp declaration for a description of how this function gets called.
+TaskCollection StochasticSubgridDriver::MakeTaskCollection(BlockList_t &blocks,
+                                                           const int stage) {
```
JoshuaSBrown, line 59: This method is really long, I would recommend breaking into small local functions.

JoshuaSBrown: I think the method can also be made const, unless I'm missing something.

src/utils/alias_method.cpp
```diff
@@ -0,0 +1,87 @@
...
+  // make probability table and alias table, this implementation is based on
+  // https://gist.github.com/Liam0205/0b5786e9bfc73e75eb8180b5400cd1f8
+  auto host_prob_table = Kokkos::create_mirror_view(prob_table);
+  auto host_alias_table = Kokkos::create_mirror_view(alias_table);
+
+  Real prob_sum = 0.0;
+  for (size_t i = 0; i < probabilities.size(); ++i) {
+    // printf("p[%lu] = %.8e\n", i, probabilities[i]);
+    prob_sum += probabilities[i];
+  }
```
JoshuaSBrown: Can you use std accumulate?

src/utils/alias_method.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+  std::queue<int> under_full, over_full;
+  for (size_t i = 0; i < probabilities.size(); ++i) {
+    host_alias_table(i) = -1;
+
+    host_prob_table(i) = Real(probabilities.size()) * probabilities[i] / prob_sum;
+    if (host_prob_table(i) < 1.0)
+      under_full.push(i);
+    else
+      over_full.push(i);
```
JoshuaSBrown: This is more stylistic but I prefer putting opening and closing braces to prevent accidentally shifting lines and changing logic.

example/stochastic_subgrid/stochastic_subgrid_package.cpp
```diff
@@ -0,0 +1,462 @@
...
+
+// Compute fluxes at faces given the constant velocity field and
+// some field "advected" that we are pushing around.
+// This routine implements all the "physics" in this example
+TaskStatus CalculateFluxes(std::shared_ptr<MeshBlockData<Real>> &rc) {
+  Kokkos::Profiling::pushRegion("Task_Advection_CalculateFluxes");
+  auto pmb = rc->GetBlockPointer();
+  IndexRange ib = pmb->cellbounds.GetBoundsI(IndexDomain::interior);
+  IndexRange jb = pmb->cellbounds.GetBoundsJ(IndexDomain::interior);
+  IndexRange kb = pmb->cellbounds.GetBoundsK(IndexDomain::interior);
```
JoshuaSBrown: Make const

JoshuaSBrown (changes requested):
It looks good to me, it is very ordered and clean pr. I would like to see a few of the larger functions broken up into smaller sub functions to make the logic a little clearer. There was a little repetition as well which would benefit from the same. Everything else I think is just small stuff. Does it make sense to add a regression test for this example or is it basically covering the same logic as the advection examples? Actually, I think this would make a really good performance tests sense you are testing the performance when the workload is unevenly distributed. If we can pick the random seeds it could be really useful for testing the stability of the performance.

example/stochastic_subgrid/stochastic_subgrid_package.cpp
jlippuner: I don't think breaking it up makes sense, because there is no code reuse here that could be extracted. So breaking up this function into smaller functions is the same as breaking it up into scopes with comment headers that delineate different sections. That's what I've done now. I think that's a better way to go because it doesn't unnecessarily bloat the code with functions that are called only once.

I've rearranged the reading and adding parameters to the package to keep things as local as possible and I've also added const to the variables that are not changed.

example/stochastic_subgrid/stochastic_subgrid_package.cpp
jlippuner: Hmm, good question. I think it could, but I don't think we gain anything. It goes into a `KOKKOS_LAMBDA` that captures it by value and thus makes a copy and I think the Kokkos documentation encourages to use `view` by value since it's very lightweight and can do reference counting properly that way.

example/stochastic_subgrid/parthinput.stochastic_subgrid
jlippuner: CFL is the Courant-Friedrichs-Lewy condition. This is extremely common and widely used in hydrodynamics and computational fluid dynamics. Although I'm usually always for more descriptive variable names, because "cfl" is used so ubiquitously, I think it would actually be more confusing to name it something different, because people expect it to be called "cfl".

JoshuaSBrown: Ok, I guess my ignorance of this knowledge domain is showing. :)

example/stochastic_subgrid/stochastic_subgrid_package.cpp
JoshuaSBrown: Fair enough this is on the side of stylistic preference, you do gain one more advantage by placing things in functions, you can fit more into your screen.

example/stochastic_subgrid/stochastic_subgrid_package.cpp
jlippuner: Hah, I thought you missed this one :P

example/stochastic_subgrid/parthinput.stochastic_subgrid
jlippuner: It's a fair question, though. I'm sure there are some variables that could be more descriptive.

example/stochastic_subgrid/stochastic_subgrid_package.cpp
jlippuner: Actually, scopes are even better in this regard, I think. You don't have to waste lines pulling in variables you need from the outer scope and you can place the "function body" right where you need it (so no jumping around), and you can fold it just the same to make space for more code on the screen :)

example/stochastic_subgrid/stochastic_subgrid_driver.cpp
jlippuner: As per the previous discussion, I prefer to not break up this function into smaller functions because there would be no re-use of those functions. Instead, I added some scoping with header comments to break up the function into manageable sections.

This particular method could be made `const` indeed, however, it's overriding a virtual method in the base class, which is not `const`. I'm not sure if the method in the base class should be const or not. There may be a good reason that we don't want it to be const so that other derived classes have the freedom to do non-`const` things in these methods.

src/utils/alias_method.cpp
jlippuner: Good idea!

example/stochastic_subgrid/stochastic_subgrid_driver.cpp
JoshuaSBrown: There is a workaround, add a const method with the same name to the base class, and have the non constant method in your specefic case call the const method. 

https://riptutorial.com/cplusplus/example/16974/avoiding-duplication-of-code-in-const-and-non-const-getter-methods-

This is non-blocking, but I think it would make it easier to follow in the future, if people are not changing the state.

src/utils/alias_method.cpp
jlippuner: Even better:
```cxx
(host_prob_table(i) < 1.0 ? under_full : over_full).push(i);
```

JoshuaSBrown: under_full is missing a push but ya that is clear. 

Oh, nevermind.

jlippuner:
@JoshuaSBrown Thanks for all your diligent comments. I think I've implemented or responded to all of them. Please review those changes.

example/stochastic_subgrid/stochastic_subgrid_driver.cpp
```diff
@@ -0,0 +1,206 @@
...
+                               mdudt.get(), beta * dt, mc1.get());
+    }
+
+    auto add_boundary_task = [&](const auto &func) {
+      TaskRegion &tr = tc.AddRegion(num_partitions);
+      for (int i = 0; i < num_partitions; i++) {
+        auto &mc1 = pmesh->mesh_data.GetOrAdd(stage_name[stage], i);
+        tr[i].AddTask(none, func, mc1);
+      }
+    };
```
JoshuaSBrown, line 157: Yes, this is nicer, it clearly limits the scope of local variables.

JoshuaSBrown (approved):
Beautiful pr, one last point pleas update the copyright date on all the files you have changed some of them are out of date. Oh, and you never answered me about creating a performance regression test?

jlippuner:
> Beautiful pr, one last point pleas update the copyright date on all the files you have changed some of them are out of date. Oh, and you never answered me about creating a performance regression test?

Thanks and good catch! I'll update that and then merge once CI passes.

JoshuaSBrown:
> > Beautiful pr, one last point pleas update the copyright date on all the files you have changed some of them are out of date. Oh, and you never answered me about creating a performance regression test?
> 
> Thanks and good catch! I'll update that and then merge once CI passes.

**and you never answered me about creating a performance regression test?**

jlippuner:
> > > Beautiful pr, one last point pleas update the copyright date on all the files you have changed some of them are out of date. Oh, and you never answered me about creating a performance regression test?
> > 
> > 
> > Thanks and good catch! I'll update that and then merge once CI passes.
> 
> **and you never answered me about creating a performance regression test?**

Oh yes, sorry I forgot about that. That might be worthwhile. Although I'm not sure we want two performance regression tests. Maybe we should use this one instead of the advection example. I'm not sure. I think that would be something to discuss with the team. I'd stay it's beyond the scope of this PR, though.

PR #410, opened 2021-01-08
Joshua s brown/fix python cpplint to call python3

### LINKED ISSUE #398: Detect absence of `python`
I have `python2` and `python3` but not `python`, which results in a lot of build errors. We should deal with this. At least warn if there is no `python` and better yet detect what versions are available and choose one that exists.

JoshuaSBrown:
## PR Summary

Fixes the issue raised by @jlippuner about the cpp linter. In this pr cmake will warn if python3 is not found it also specifically calls python3 in the Lint.cmake file. 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


Yurlungur (approved):
It looks like if a python3 interpreter isn't found, the lint target is just not build but the build process doesn't die? If so, that's great.

cmake/Lint.cmake
```diff
@@ -52,15 +52,22 @@ function(lint_target TARGET_NAME)
     get_target_property(TARGET_SOURCES ${TARGET_NAME} SOURCES)
     get_target_property(TARGET_SOURCE_DIR ${TARGET_NAME} SOURCE_DIR)
 
-    set(TARGET_OUTPUTS)
-    foreach(SOURCE ${TARGET_SOURCES})
-        lint_file(${TARGET_SOURCE_DIR} ${SOURCE} ${SOURCE}.lint)
-        list(APPEND TARGET_OUTPUTS ${SOURCE}.lint)
-    endforeach()
+    if( NOT Python3_Interpreter_FOUND)
+      find_package(Python3 REQUIRED COMPONENTS Interpreter)
```
AndrewGaspar: `REQUIRED` will result in a `FATAL_ERROR` if Python3 isn't found, which doesn't seem to be your intention based on the check after this. You probably want to remove `REQUIRED`.

AndrewGaspar (changes requested):

cmake/Lint.cmake
JoshuaSBrown: Good point!

AndrewGaspar (approved):

jlippuner (approved):
Thanks @JoshuaSBrown! I tested this on my system without `python` and it works (the master branch fails, so this fixes the issue).

PR #413, opened 2021-01-13
Add machine configuration for LANL Snow

AndrewGaspar:
## PR Summary

Adds a Snow configuration file.

Also adds srun to a few places... Snow seems to require anything that uses MPI (including things like h5py indirectly) to run under srun. @pgrete probably wants to take a look at this.

I also added a label to the unit tests.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


AndrewGaspar:
cc @cielling

Yurlungur (approved):
Thanks for setting this up!

cmake/TestSetup.cmake
```diff
@@ -126,7 +133,12 @@ function(setup_test_mpi nproc dir arg extra_labels)
       --test_dir ${CMAKE_CURRENT_SOURCE_DIR}/test_suites/${dir}
       --output_dir "${PROJECT_BINARY_DIR}/tst/regression/outputs/${dir}_mpi"
       --kokkos_args=${PARTHENON_KOKKOS_TEST_ARGS})
-    set_tests_properties(regression_mpi_test:${dir} PROPERTIES LABELS "${labels}" RUN_SERIAL ON )
+    set_tests_properties(
+      regression_mpi_test:${dir}
+      PROPERTIES
+        LABELS "${labels}"
```
pgrete, line 152: What happened to the RUN_SERIAL option?

tst/unit/CMakeLists.txt
```diff
@@ -50,4 +50,7 @@ ParseAndAddCatchTests(unit_tests)
 
 get_property(ALL_TESTS DIRECTORY . PROPERTY TESTS)
 
+foreach(TEST ${ALL_TESTS})
+    set_property(TEST "${TEST}" APPEND PROPERTY LABELS unit)
+endforeach()
```
pgrete: Is this required? If yes, why didn't we have this before?

docs/building.md
```diff
@@ -268,7 +268,60 @@ Once you've configured your build directory, you can build with
...
+```
+
+If you didn't follow the "Set-Up Environment" section, you need to specify the
+`MACHINE_CFG` file, as well.
+
+```bash
+$ cmake -S. -Bbuild -DMACHINE_CFG=cmake/machinecfg/Snow.cmake
+```
+
+Parthenon is built with the Intel compilers by default on Snow. To build with
```
pgrete, line 323: If Intel is the default I also suggest to add additional compiler flags (e.g., to use the simd statements)

cmake/TestSetup.cmake
```diff
@@ -51,10 +51,17 @@ function(setup_test dir arg extra_labels)
...
+    set(PREFIX_SERIAL_REGRESSION_TEST ${TEST_MPIEXEC})
+    set(SUFFIX_SERIAL_REGRESSION_TEST --mpirun ${TEST_MPIEXEC})
+  endif()
+  add_test(
+    NAME regression_test:${dir}
+    COMMAND ${PREFIX_SERIAL_REGRESSION_TEST} ${Python3_EXECUTABLE} "${CMAKE_CURRENT_SOURCE_DIR}/run_test.py" 
+      ${arg} --test_dir "${CMAKE_CURRENT_SOURCE_DIR}/test_suites/${dir}"
+      --output_dir "${PROJECT_BINARY_DIR}/tst/regression/outputs/${dir}"
+      --kokkos_args=${PARTHENON_KOKKOS_TEST_ARGS}
+      ${SUFFIX_SERIAL_REGRESSION_TEST})
```
JoshuaSBrown: I think if we are going to use mpi related variables in this function we should probably change the names for the functions, because we already have setup_test_mpi, this is now using mpi variable but it's not in the name, we could change the function names to setup_test_serial and setup_test_parallel.

cmake/TestSetup.cmake
JoshuaSBrown: I think that logic was moved to the setup_test function above.

tst/unit/CMakeLists.txt
JoshuaSBrown: Ya, this should not be required, CATCH2 will register the labels with CMake that already appear in the TEST_CASE macros, were you having problems with that?

docs/building.md
AndrewGaspar: I think we already are - we're "linking" `OpenMP::OpenMP_CXX` by default.

cmake/TestSetup.cmake
AndrewGaspar: Well, not quite - the intention with setting PROCESSORS was to be better than RUN_SERIAL, since it lets you still get parallel test execution, but on different processors.

However, my one concern with this was the interaction with GPU execution.

tst/unit/CMakeLists.txt
AndrewGaspar: Ah, I didn't realize catch2 automatically labeled them all with unit_tests. I'll remove this.

cmake/TestSetup.cmake
AndrewGaspar: I'm thinking I should keep `RUN_SERIAL` if `Kokkos_ENABLE_CUDA` since cmake can't really control "GPU affinity". Thoughts?

This change makes the tests run a lot faster on CPU builds, so I'd like to keep it there.

cmake/TestSetup.cmake
AndrewGaspar: `setup_serial_test`?

I probably should change `if(SERIAL_WITH_MPI)` to `if(SERIAL_WITH_MPI AND NOT DISABLE_MPI)` or something.

cmake/TestSetup.cmake
JoshuaSBrown: That's fine by me.

JoshuaSBrown (approved):
Thanks Andrew this is great!

cmake/TestSetup.cmake
```diff
@@ -47,17 +47,24 @@ endfunction()
...
-    ${arg} --test_dir "${CMAKE_CURRENT_SOURCE_DIR}/test_suites/${dir}"
-    --output_dir "${PROJECT_BINARY_DIR}/tst/regression/outputs/${dir}"
-    --kokkos_args=${PARTHENON_KOKKOS_TEST_ARGS})
+  if (SERIAL_WITH_MPIEXEC)
+    set(PREFIX_SERIAL_REGRESSION_TEST ${TEST_MPIEXEC})
+    set(SUFFIX_SERIAL_REGRESSION_TEST --mpirun ${TEST_MPIEXEC})
+  endif()
+  add_test(
+    NAME regression_test:${dir}
+    COMMAND ${PREFIX_SERIAL_REGRESSION_TEST} ${Python3_EXECUTABLE} "${CMAKE_CURRENT_SOURCE_DIR}/run_test.py"
```
pgrete: I'm not an expert in MPI contexts, but if I read this correct this means that we'd run python with h5py (that potentially and initializes mpi4py) to run another program (our examples) also with mpi.
Do I read this correctly? While this may work this somehow asks for trouble, doesn't it?
Also, if that's to only fix the `import h5py` issues, there may be other ways to address that (see https://github.com/h5py/h5py/issues/917 and related issues)

docs/building.md
pgrete: Out of curiosity, would you mind double checking that `-fopenmp-simd` and `-qopenmp-simd`, respectively, are added as compile flags?
I'm not sure linking against OpenMP makes this happen as `openmp-simd` is independent of openmp threading.

cmake/TestSetup.cmake
AndrewGaspar: Actually, it may not be necessary here.

AndrewGaspar: Let me check.

AndrewGaspar: *it being running python under mpiexec. Still need it for the test execution for sure.

AndrewGaspar: Removed the mpiexec call for running the python script - I didn't actually need that.

docs/building.md
AndrewGaspar: `-qopenmp` and `-fopenmp` are applied, and the documentation in icpc for `-qopenmp-simd` says this:

```
-qopenmp-simd
          Enables OpenMP* SIMD compilation.  Enabled by default with
          -qopenmp.  Use -qno-openmp-simd to disable.
```

cmake/TestSetup.cmake
pgrete: Great!

docs/building.md
pgrete: Thanks for confirming.

pgrete (approved):

Issue #423, opened 2021-01-20
Package names not unique


Yurlungur:
@clellsolomon discovered that one can create a package with two different "names" via something like:
```C++
Packages_t packages;
packages["Name1"] = std::make_shared<StateDescriptor>("Name2");
```
Unfortunately, there appears to be a hidden assumption within the code that "Name1" and "Name2" are the same... at least in the package dependencies machinery I wrote in #386 . I see fixing this as having two parts:

 - [ ] Remove this implicit assumption in the code added in #386 
 - [ ] Redesign `Packages_t` and `StateDescriptor` so that this mistake cannot be made. This change would be API breaking, but the changes should be easy to fix.

Interested in the thoughts of those who have thought about `src/interface` such as @jdolence and @nmsriram and maintainers of downstream codes e.g., @pgrete and/or @forrestglines

PR #424, opened 2021-01-20
Jmm/remove implicit logic

Yurlungur:
## PR Summary

Resolves the bug found by @clellsolomon and reported in #423. In short, the issue is that a package can have a different label in `StateDescriptor` than key in `Packages_t`, however they were assumed to be identical in the code I wrote in MR #386. This MR undoes that assumption by explicitly using keys instead of package labels. I will submit an MR in the future that resolves the ambiguity where each package has two "names."

## PR Checklist

- [x] Code passes cpplint
- [x] New features are documented. (None)
- [x] Code is formatted


Yurlungur:
On @AndrewGaspar 's suggestion, I will close this and simply merge in #425 as appropriate.

PR #425, opened 2021-01-21
Remove ambiguity in package names.

### LINKED ISSUE #423: Package names not unique

Yurlungur:
## PR Summary

This is related to issue #423 and fixes (in my opinion) an underlying design flaw. In a downstream code, @clellsolomon discovered that there are really *two* names for each package. There's the label set in the `StateDescriptor` constructor, e.g.,
```C++
auto pkg = std::make_shared<StateDescriptor>(name1);
```
and then there's the *key* in the `std::unordered_map` used for `Packages_t`:
```C++
Packages_t packages;
packages[name2] = pkg;
```
It's implicitly assumed that `name1` and `name2` are the same, but it's not enforced anywhere. And even if we enforced it, it's strange that we require the user to enter this name twice. Here I propose and implement one solution, which I think is the preferred one. I have changed `Packates_t` from a `std::unordered_map` to it's own class, which adds a package with the same key as the label in the package. E.g., you now call:
```C++
packages.Add(pkg); // registers the package with name pkg->label()
```
and you can access the package via
```C++
packages.Get("package name");
```
which has bounds checking, so an error is thrown if the package doesn't exist.

I like this solution because it:
1. Removes the aforementioned ambiguity with regards to package names
2. "Encapsulates" package names to some degree to the package itself, rather than the application. I.e., you can imagine having an application that initializes 3 or 4 packages. The packages can be swapped in or out and the application doesn't necessarily need to know what they're called.

The disadvantages of this solution are:
1. It is API breaking because I removed the `operator[]` from `Packages_t` and replaced it with `Add` and `Get`. However, the changes to downstream codes should be very easy.
2. The application has no control over package names. If the names conflict, the code throws an error.

I'm interested in people's thoughts.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


CHANGELOG.md
```diff
@@ -26,12 +26,14 @@ Date: 01/19/2021
...
 
 ### Changed (changing behavior/API/variables/...)
+- [[PR 425]](https://github.com/lanl/parthenon/pull/425) Remove ambiguity in package names. `Packages_t` no longer has an `operator[]` method. This has been replaced with `Add` and `Get`.
 - [[PR 393]](https://github.com/lanl/parthenon/pull/393) Small refactor to make driver code more flexible for downstream apps.
 - [[PR 400]](https://github.com/lanl/parthenon/pull/400) Change `Mesh`, `ApplicationInput`, and `Driver` to suppport pre- and post- step user work
 - [[PR 394]](https://github.com/lanl/parthenon/pull/394) Make `Params.Get` const-correct.
 - [[PR 332]](https://github.com/lanl/parthenon/pull/332) Rewrote boundary conditions to work on GPUs with variable packs. Re-enabled user-defined boundary conditions via `ApplicationInput`.
 
 ### Fixed (not changing behavior/API/variables/...)
+- [[PR 424]](https://github.com/lanl/parthenon/pull/424) Remove implicit assumption that package labels and keys in `Packages_t` are identical
```
AndrewGaspar: Remove this.

CHANGELOG.md
```diff
@@ -26,12 +26,14 @@ Date: 01/19/2021
 - [[PR 386]](https://github.com/lanl/parthenon/pull/386) Introduce `Private`, `Provides`, `Requires`, and `Overridable` variable metadata, allowing fine-grained control of conflict resolution between packages.
 
 ### Changed (changing behavior/API/variables/...)
+- [[PR 425]](https://github.com/lanl/parthenon/pull/425) Remove ambiguity in package names. `Packages_t` no longer has an `operator[]` method. This has been replaced with `Add` and `Get`.
```
AndrewGaspar: You need to move this up to the `Current develop` section.

CHANGELOG.md
AndrewGaspar:
```suggestion
```

src/interface/state_descriptor.hpp
```diff
@@ -245,7 +245,34 @@ class StateDescriptor {
...
+    }
+    return false;
+  }
+  std::shared_ptr<StateDescriptor> &Get(const std::string &name) {
+    return packages_.at(name);
+  }
+  // TODO(JMM): Should this be provided or deleted?
+  // My preference is to delete it, but it's more API breaking.
+  /*
+  std::shared_ptr<StateDescriptor> &operator[](const std::string &name) {
```
AndrewGaspar: On the one hand, I prefer the new method names. On the other hand, I always frown on breaking the API unnecessarily. Of course, we're still pre-1.0, so if there's a better API, we should add it now, so I'd say let's not support operator[].

Remove this commented out section and TODO before merge.

AndrewGaspar (changes requested):

src/interface/state_descriptor.hpp
```diff
@@ -245,7 +245,34 @@ class StateDescriptor {
   Dictionary<Dictionary<Metadata>> swarmValueMetadataMap_;
 };
 
-using Packages_t = Dictionary<std::shared_ptr<StateDescriptor>>;
+class Packages_t {
+ public:
+  Packages_t() = default;
+  bool Add(const std::shared_ptr<StateDescriptor> &package) {
```
AndrewGaspar: Yeah, I'd say make it an error to call `Add` with the same name multiple times. I can't imagine that overwriting is ever the intended behavior. If people really want it, you could add a separate `AddOrOverwrite` method later.

CHANGELOG.md
Yurlungur: Oops good catch.

src/interface/state_descriptor.hpp
Yurlungur: Will do. I would like to hear what the Athena-PK people say before I delete it though, since they're stakeholders in the API. @pgrete @forrestglines ?

CHANGELOG.md
Yurlungur: Oops. Thanks for the catch. That must have somehow gotten caught in the merge after the release.

Yurlungur: Done.

brryan:
This is a nice improvement IMO. As a downstream consumer I think the API-breaking is justified.

It seems to me right now (and like @AndrewGaspar says) that it would be better to just throw an error if you try to add a package that's already added, I'm having a hard time imagining a scenario where you would want the application to proceed when you try to add some package that doesn't get added because of a name conflict.

One issue I worry about though is if you are trying to add two packages that have the same `StateDescriptor` label. What is the right thing to do here? Maybe add the second package with a mangled label (e.g. `package_1`), have `Add` return that new label, and set an internal boolean array inside `Packages_t` that says that there's a mismatch between the `StateDescriptor` label and the `Packages_t` key?

src/interface/state_descriptor.hpp
```diff
@@ -245,7 +245,34 @@ class StateDescriptor {
...
+    const auto &name = package->label();
+    if (packages_.count(name) == 0) {
+      packages_[name] = package;
+      return true;
+    }
+    return false;
+  }
+  std::shared_ptr<StateDescriptor> &Get(const std::string &name) {
```
AndrewGaspar: I feel like this should return `std::shared_ptr<StateDescriptor> const &`. Remember - `const` in this case would only confer _shallow_ immutability. So it would prevent people from changing the slot to a wholly different `shared_ptr<StateDescriptor>`, but still allow them to make changes to the existing `StateDescriptor`. Of course, this doesn't do much because you could then just mutate the `StateDescriptor` to violate invariants, too. Idk, we really ought to bifurcate our public API and internal implementation details more, but that's a discussion for another time.

Yurlungur: Done.

src/interface/state_descriptor.hpp
pgrete: No worries here. Better break the API early than late ;)
I'll do a full review tomorrow.

src/interface/state_descriptor.hpp
Yurlungur: I think I agree. And I've made the change.

Yurlungur:
Thanks, @AndrewGaspar and @brryan for your comments! I've made the changes you suggest, @AndrewGaspar except for deleting the commented out `operator[]` as I want to hear from the Athena-PK folks.

> One issue I worry about though is if you are trying to add two packages that have the same StateDescriptor label. What is the right thing to do here? Maybe add the second package with a mangled label (e.g. package_1), have Add return that new label, and set an internal boolean array inside Packages_t that says that there's a mismatch between the StateDescriptor label and the Packages_t key?

I think for now we should just forbid this, as @AndrewGaspar suggested and throw an error. But I agree in the future we might need to be more clever.

src/interface/state_descriptor.hpp
Yurlungur: Cool, thanks. I'll delete the commented-out code then. :)

Yurlungur:
@AndrewGaspar I think all your review comments should be addressed now.

jdolence (approved):
Thanks for doing this @Yurlungur.  This issue has bothered me for a long time and your resolution seems quite reasonable.

AndrewGaspar (approved):

pgrete (approved):
Thanks for taking care of this!

I was just about to ask if the documentation was updated and then double checked to realize that we don't have docs for this piece (except for "look at the advection example") :smile: 
I think the pointer to the advection example is fine/enough so let's get this in.

PR #428, opened 2021-01-21
Update copyright for all files that LANL employees touched in 2021

AndrewGaspar:
## PR Summary

Fails the build if LANL employees don't do this.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


Yurlungur (approved):

cmake/CheckCopyrightScript.cmake
```diff
@@ -0,0 +1,68 @@
...
+# license in this material to reproduce, prepare derivative works, distribute copies to
+# the public, perform publicly and display publicly, and to permit others to do so.
+#=========================================================================================
+
+# This script checks that a file has an up to date copyright. Only runs for
+# users with @lanl.gov email addresses.
+
+set(SOURCE_DIR ${CMAKE_ARGV3})
+set(GIT_EXECUTABLE ${CMAKE_ARGV4})
+set(INPUT ${CMAKE_ARGV5})
```
JoshuaSBrown, line 19: What does INPUT do?

cmake/CheckCopyrightScript.cmake
```diff
@@ -0,0 +1,68 @@
...
+
+string(TIMESTAMP CURRENT_YEAR "%Y")
+
+# Check if the file has an up to date copyright
+file(READ ${SOURCE_DIR}/${INPUT} CONTENTS)
+string(REGEX MATCH "\\(C\\) \\(or copyright\\) 2020-${CURRENT_YEAR}\\. Triad National Security, LLC\\. All rights reserved\\." HAS_COPYRIGHT ${CONTENTS})
```
JoshuaSBrown: So if I add a new file this year to Parthenon is the copyright still supposed to start from 2020 for that file?

cmake/CheckCopyrightScript.cmake
```diff
@@ -0,0 +1,68 @@
...
+                ${INPUT}
+        WORKING_DIRECTORY ${SOURCE_DIR}
+        RESULT_VARIABLE HAS_CHANGES
+        OUTPUT_QUIET)
+
+    # If the file has changes by a LANL email, the copyright must be updated
+    if (HAS_CHANGES)
+        set(REQUIRES_COPYRIGHT_UPDATE ON)
+    endif()
+endif()
```
JoshuaSBrown: I don't understand this logic, if the file has not been modified why do we need the else statement?

JoshuaSBrown:
Maybe we should also implement a target that will automatically update files that have been modified.

cmake/CheckCopyrightScript.cmake
AndrewGaspar: Ah, good point...

cmake/CheckCopyrightScript.cmake
AndrewGaspar: This was basically to catch all the cases that had slipped through before this change... I suppose we shouldn't need it anymore.

JoshuaSBrown (approved):
LGTM, thanks Andrew!

PR #434, opened 2021-02-02
Runtime configurable nghost

### LINKED ISSUE #431: PostExecute() or PostExecute(DriverStatus)?
@jdolence @AndrewGaspar You were the last to touch lines 53 and 82 in `driver.hpp`. The `Driver` class contains `virtual void PreExecute();` but the `EvolutionDriver` (derived from `Driver`) has `void PostExecute(DriverStatus status);`. This results in lots of compiler warnings like this:

```
src/driver/driver.hpp(82): warning: function "parthenon::Driver::PostExecute()" is hidden by "parthenon::EvolutionDriver::PostExecute" -- virtual function override intended?
```

Is `PostExecute` in `EvolutionDriver` supposed to override the `Driver`'s `PostExecution`? If not, it should have a different name to remove ambiguity and compiler warning.

jdolence:
## PR Summary

The number of ghost zones has been fixed at compile time, but this is inconvenient when a downstream code provides options that require more ghost zones.  This is particularly acute when downstream codes consume parthenon as a library dependency.

This PR removes the compile time constant and replaces it with a runtime parameter that can be set in input files.  If not provided, the default number of ghost zones is set to 2, which was the default in our CMake setup before.

I don't believe there will be any real performance implications but we should test this.

Closes #431.

## PR Checklist

- [x] Code passes cpplint
- [x] Existing tests exercise this already.
- [x] Code is formatted

### DIFF
--- a/docs/inputs.md
+++ b/docs/inputs.md
@@ -3,4 +3,4 @@
    |             Option                    | Default  | Type   | Description |
    | ------------------------------------: | :------- | :----- | :---------- |
    | <parthenon/time><br>perf_cycle_offset | 0        | int    | Skip the first N cycles when calculating the final performance (e.g., zone-cycles/wall_second). Allows to hide the initialization overhead in Parthenon, which usually takes place in the first cycles when Containers are allocated, etc. | 
-   | <parthenon/mesh><br>nghost            | 2        | int    | Number of ghost cells for each mesh block | 
+   | <parthenon/mesh><br>nghost            | 2        | int    | Number of ghost cells for each mesh block on each side. | 
--- a/docs/inputs.md
+++ b/docs/inputs.md
@@ -1,5 +1,6 @@
 ### List of input file parameters (incomplete)
 
-   |             Option               | Default  | Type   | Description |
-   | -------------------------------: | :------- | :----- | :---------- |
-   | parthenon/time/perf_cycle_offset | 0        | int | Skip the first N cycles when calculating the final performance (e.g., zone-cycles/wall_second). Allows to hide the initialization overhead in Parthenon, which usually takes place in the first cycles when Containers are allocated, etc. | 
+   |             Option                    | Default  | Type   | Description |
+   | ------------------------------------: | :------- | :----- | :---------- |
+   | <parthenon/time><br>perf_cycle_offset | 0        | int    | Skip the first N cycles when calculating the final performance (e.g., zone-cycles/wall_second). Allows to hide the initialization overhead in Parthenon, which usually takes place in the first cycles when Containers are allocated, etc. | 
+   | <parthenon/mesh><br>nghost            | 2        | int    | Number of ghost cells for each mesh block | 
--- a/example/calculate_pi/pi_driver.cpp
+++ b/example/calculate_pi/pi_driver.cpp
@@ -1,5 +1,5 @@
 //========================================================================================
-// (C) (or copyright) 2020. Triad National Security, LLC. All rights reserved.
+// (C) (or copyright) 2020-2021. Triad National Security, LLC. All rights reserved.
 //
 // This program was produced under U.S. Government contract 89233218CNA000001 for Los
 // Alamos National Laboratory (LANL), which is operated by Triad National Security, LLC
@@ -111,7 +111,7 @@ void PiDriver::PostExecute(Real pi_val) {
     fs << "rel error = " << (pi_val - M_PI) / M_PI << std::endl;
     fs.close();
   }
-  Driver::PostExecute();
+  Driver::PostExecute(DriverStatus::complete);
 }
 
 template <typename T>
--- a/example/kokkos_pi/kokkos_pi.cpp
+++ b/example/kokkos_pi/kokkos_pi.cpp
@@ -139,8 +139,7 @@ static double sumArray(BlockList_t &blocks, const int &n_block) {
           const int k_grid = idx / n_block2;
           const int j_grid = (idx - k_grid * n_block2) / n_block;
           const int i_grid = idx - k_grid * n_block2 - j_grid * n_block;
-          mySum += inOrOut(0, k_grid + nghost, j_grid + nghost,
-                           i_grid + nghost);
+          mySum += inOrOut(0, k_grid + nghost, j_grid + nghost, i_grid + nghost);
         },
         oneSum);
     Kokkos::fence();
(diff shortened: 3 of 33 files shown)


Yurlungur (approved):
This is a long overdue change. Thanks for doing this.

Yurlungur (commented):
Putting these in globals was a good solution.

jdolence:
> Putting these in globals was a good solution.

yeah thanks to @brryan

pgrete (approved):
Great! Looks. I'll confirm downstream that this works for the intended purpose but even if it's not then this would be different bug and should not prevent this PR.

docs/inputs.md
```diff
@@ -1,5 +1,6 @@
 ### List of input file parameters (incomplete)
 
-   |             Option               | Default  | Type   | Description |
-   | -------------------------------: | :------- | :----- | :---------- |
-   | parthenon/time/perf_cycle_offset | 0        | int | Skip the first N cycles when calculating the final performance (e.g., zone-cycles/wall_second). Allows to hide the initialization overhead in Parthenon, which usually takes place in the first cycles when Containers are allocated, etc. | 
+   |             Option                    | Default  | Type   | Description |
+   | ------------------------------------: | :------- | :----- | :---------- |
+   | <parthenon/time><br>perf_cycle_offset | 0        | int    | Skip the first N cycles when calculating the final performance (e.g., zone-cycles/wall_second). Allows to hide the initialization overhead in Parthenon, which usually takes place in the first cycles when Containers are allocated, etc. | 
+   | <parthenon/mesh><br>nghost            | 2        | int    | Number of ghost cells for each mesh block |
```
jlippuner: Can we maybe clarify that this is the number of ghost cell layers on each face, i.e. the number of total ghost cells in each dimension is twice this number.

example/calculate_pi/pi_driver.cpp
```diff
@@ -111,7 +111,7 @@ void PiDriver::PostExecute(Real pi_val) {
     fs << "rel error = " << (pi_val - M_PI) / M_PI << std::endl;
     fs.close();
   }
-  Driver::PostExecute();
+  Driver::PostExecute(DriverStatus::complete);
```
jlippuner, line 114: Just making sure this should be part of this PR and isn't meant for another PR

src/driver/driver.cpp
```diff
@@ -35,7 +35,7 @@ void Driver::PreExecute() {
   timer_main.reset();
 }
 
-void Driver::PostExecute() {
+void Driver::PostExecute(DriverStatus status) {
```
jlippuner, line 38: So this PR will also close #431?

nmsriram:
@jdolence do we know if there are any performance implications since the compiler can no longer optimize out a constant?

src/mesh/mesh_refinement.cpp
```diff
@@ -50,10 +50,11 @@ MeshRefinement::MeshRefinement(std::weak_ptr<MeshBlock> pmb, ParameterInput *pin
   // Create coarse mesh object for parent grid
   coarse_coords = Coordinates_t(pmb.lock()->coords, 2);
 
-  if (NGHOST % 2) {
+  if (Globals::nghost % 2) {
```
jlippuner: For clarity, I suggest:
```suggestion
  if ((Globals::nghost % 2) != 0) {
```

src/mesh/mesh_refinement.cpp
```diff
@@ -50,10 +50,11 @@ MeshRefinement::MeshRefinement(std::weak_ptr<MeshBlock> pmb, ParameterInput *pin
...
     std::stringstream msg;
     msg << "### FATAL ERROR in MeshRefinement constructor" << std::endl
-        << "Selected --nghost=" << NGHOST << " is incompatible with mesh refinement.\n"
+        << "Selected --nghost=" << Globals::nghost
+        << " is incompatible with mesh refinement.\n"
```
jlippuner:
```suggestion
        << " is incompatible with mesh refinement because it is not a multiple of 2.\n"
```

jlippuner (approved):
This looks great and I agree it's a needed change.

I have just a few minor comments/questions/suggestions. Please address/respond as you see fit, but none of them should hold up merging.

nmsriram:
I guess if it does slow it down tremendously, we can always convert it to a constant in Globals and remove the input block in the input processing.

example/kokkos_pi/kokkos_pi.cpp
```diff
@@ -65,6 +65,8 @@ using namespace parthenon::driver::prelude;
 
 using View2D = Kokkos::View<Real **, Kokkos::LayoutRight, Kokkos::DefaultExecutionSpace>;
 
+const int nghost = 2;
```
nmsriram, line 68: Should this be set in the input file and `Globals::ghost` used in the file?

src/mesh/mesh_refinement.cpp
```diff
@@ -50,10 +50,11 @@ MeshRefinement::MeshRefinement(std::weak_ptr<MeshBlock> pmb, ParameterInput *pin
...
         << "Reconfigure with an even number of ghost cells " << std::endl;
```
nmsriram: Replace "Reconfigure" with "Rerun"

src/reconstruct/reconstruction.cpp
```diff
@@ -80,11 +80,11 @@ Reconstruction::Reconstruction(std::weak_ptr<MeshBlock> wpmb, ParameterInput *pi
...
   if (xorder == 3) {
     int req_nghost = 3;
-    if (NGHOST < req_nghost) {
+    if (Globals::nghost < req_nghost) {
       std::stringstream msg;
       msg << "### FATAL ERROR in Reconstruction constructor" << std::endl
           << "xorder=" << input_recon
-          << " (PPM) reconstruction selected, but nghost=" << NGHOST << std::endl
+          << " (PPM) reconstruction selected, but nghost=" << Globals::nghost << std::endl
           << "Reconfigure with --nghost=XXX with XXX > " << req_nghost - 1 << std::endl;
```
nmsriram: ditto Reconfigure with Rerun
A few other places as well that I'm not going to mark.

nmsriram (approved):
I've made a couple requests that are not binding:
 - use Globals::nghost in the pi _driver.cpp
 - Replace 'Reconfigure with different nghost' with 'Rerun with even nghost' in a few places.

example/calculate_pi/pi_driver.cpp
jdolence: I snuck this in here as an issue that keeps coming up but never with enough importance to generate a PR.

src/driver/driver.cpp
jdolence: oh yeah good call.  thanks!

example/kokkos_pi/kokkos_pi.cpp
jdolence: I actually tried to use `Globals::nghost` and something prevented it from working.  Honestly, I just didn't think it was worth it to track down the issue.  This example doesn't really actually use the parthenon machinery anyway.  I've even been thinking lately (after seeing it trip some people up) that we should remove it from the examples that ship with parthenon.

Issue #433, opened 2021-02-01
Bug in Regression tests


JoshuaSBrown:
I believe there are two issues causing problems, one jsrun uses different parameters to specify the number of ranks, they are not all accounted by test_case.py the python regression testing framework. 
This however does not account for the tests that have been failing on darwin since the buffer-in-one pr was merged. 

https://github.com/lanl/parthenon/pull/359

JoshuaSBrown:
Been failing since commit c4dfc1eef028027cd625538cf3b571862a4858fb

JoshuaSBrown:
This is the test that fails:
```
/projects/opt/ppc64le/p9/openmpi/4.0.2-gcc_9.2.0/bin/mpiexec -n 2 /home/joshuasbrown/Software/parthenon_x86_64/parthenon/build/example/advection/advection-example -i /home/joshuasbrown/Software/parthenon_x86_64/parthenon/tst/regression/test_suites/advection_performance/parthinput.advection_performance parthenon/mesh/nx1=512 parthenon/meshblock/nx1=16 parthenon/mesh/nx2=256 parthenon/meshblock/nx2=16 parthenon/mesh/nx3=256 parthenon/meshblock/nx3=16 --kokkos-num-devices=2 --kokkos-threads=1
```

JoshuaSBrown:
The commit before c4dfc1e passes f47690fcaaeac65da6b6e4a85842e8cee36e16bb

JoshuaSBrown:
Which is really weird because c4dfc1e adds files but doesn't call on any of it's logic.

JoshuaSBrown:
Looking back at the history it seems that the same test has been failing on and off before it became consistent. The earliest failure I can find starts with commit 8eb18ace, which is completely unrelated. So something is off in the logic of the mpi/cuda calls that is triggering this that is sensitive to the system environment.

JoshuaSBrown:
```

[...]

PR #435, opened 2021-02-02
Separate num ranks from other mpi options

pgrete:
## PR Summary

Currently the regression test manually parsed the number of ranks from `-n` or `-np`, which was then in turn used to adjust the MeshBlock sizes in the advection performance and convergence tests so that those test would work with 1+ ranks.

However, the current automated parsing logic was problematic when the number of ranks was not determined by the above options, e.g, when using `jsrun`  with `-a` (and `-n` carrying a different meaning).
Given that we already have a hardcoded CMake value for the number of ranks, this number is now directly used (instead of being derived).

Fixes some part of #377 and #433

## PR Checklist

- [X] Code passes cpplint
- [X] Adds a test for any bugs fixed. Adds tests for new features.
- [X] Code is formatted

### DIFF
--- a/cmake/PythonModuleCheck.cmake
+++ b/cmake/PythonModuleCheck.cmake
@@ -19,9 +19,7 @@ function(required_python_modules_found module_list)
   if(${Python3_Interpreter_FOUND})
     set(IMPORT_ERROR 0)
     foreach(module IN LISTS module_list )
-      # Run test under mpiexec equivalent - some modules like h5py may require
-      # it.
-      execute_process(COMMAND ${TEST_MPIEXEC} ${Python3_EXECUTABLE} -c "import ${module}"
+      execute_process(COMMAND ${Python3_EXECUTABLE} -c "import ${module}"
         RESULT_VARIABLE IMPORT_MODULE ERROR_QUIET)
     
       if(NOT ${IMPORT_MODULE} EQUAL 0)
--- a/cmake/TestSetup.cmake
+++ b/cmake/TestSetup.cmake
@@ -98,12 +98,12 @@ function(process_mpi_args nproc)
   endif()
   # use custom numproc flag
   if (TEST_NUMPROC_FLAG)
-    list(APPEND TMPARGS "--mpirun_ranks=${TEST_NUMPROC_FLAG}")
+    list(APPEND TMPARGS "--mpirun_ranks_flag=${TEST_NUMPROC_FLAG}")
   # use CMake determined numproc flag
   else()
-    list(APPEND TMPARGS "--mpirun_ranks=${MPIEXEC_NUMPROC_FLAG}")
+    list(APPEND TMPARGS "--mpirun_ranks_flag=${MPIEXEC_NUMPROC_FLAG}")
   endif()
-  list(APPEND TMPARGS "--mpirun_ranks=${nproc}")
+  list(APPEND TMPARGS "--mpirun_ranks_num=${nproc}")
   # set additional options from machine configuration
   foreach(MPIARG ${TEST_MPIOPTS})
     list(APPEND TMPARGS "--mpirun_opts=${MPIARG}")
--- a/cmake/TestSetup.cmake
+++ b/cmake/TestSetup.cmake
@@ -98,12 +98,12 @@ function(process_mpi_args nproc)
   endif()
   # use custom numproc flag
   if (TEST_NUMPROC_FLAG)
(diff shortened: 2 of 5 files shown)


pgrete:
@AndrewGaspar could you please confirm that this still works on Snow.
I just reverted the TEST_MPIEXEC call in the python module changes as this should also not be required any more (following our final discussion #413 ) and causes problem on Summit (`jsrun` cannot be called on the dev nodes).

cmake/TestSetup.cmake
```diff
@@ -98,12 +98,12 @@ function(process_mpi_args nproc)
...
-    list(APPEND TMPARGS "--mpirun_opts=${TEST_NUMPROC_FLAG}")
+    list(APPEND TMPARGS "--mpirun_ranks=${TEST_NUMPROC_FLAG}")
   # use CMake determined numproc flag
   else()
-    list(APPEND TMPARGS "--mpirun_opts=${MPIEXEC_NUMPROC_FLAG}")
+    list(APPEND TMPARGS "--mpirun_ranks=${MPIEXEC_NUMPROC_FLAG}")
   endif()
-  list(APPEND TMPARGS "--mpirun_opts=${nproc}")
+  list(APPEND TMPARGS "--mpirun_ranks=${nproc}")
   # set additional options from machine configuration
```
JoshuaSBrown: This is a good idea.

JoshuaSBrown (approved):

tst/regression/utils/test_case.py
```diff
@@ -106,20 +106,10 @@ def __init__(self,run_test_path,**kwargs):
...
-        for s in argstrings:
-          if s in self.parameters.mpi_opts:
-            index = self.parameters.mpi_opts.index(s)
-            if index < len(self.parameters.mpi_opts) - 1:
-              try:
-                self.parameters.num_ranks = int(self.parameters.mpi_opts[index+1])
-              except ValueError:
-                pass
+        if len(self.parameters.mpi_ranks) == 2:
+            self.parameters.num_ranks = int(self.parameters.mpi_ranks[1])
```
Yurlungur: Here's a clear win.

Yurlungur (approved):
Nice catch. Thanks for the fix.

cmake/PythonModuleCheck.cmake
```diff
@@ -19,9 +19,7 @@ function(required_python_modules_found module_list)
   if(${Python3_Interpreter_FOUND})
     set(IMPORT_ERROR 0)
     foreach(module IN LISTS module_list )
-      # Run test under mpiexec equivalent - some modules like h5py may require
-      # it.
-      execute_process(COMMAND ${TEST_MPIEXEC} ${Python3_EXECUTABLE} -c "import ${module}"
+      execute_process(COMMAND ${Python3_EXECUTABLE} -c "import ${module}"
```
AndrewGaspar, line 22: Yes this breaks Snow.

AndrewGaspar:
> @AndrewGaspar could you please confirm that this still works on Snow.
I just reverted the TEST_MPIEXEC call in the python module changes as this should also not be required any more (following our final discussion #413 ) and causes problem on Summit (jsrun cannot be called on the dev nodes).

@JoshuaSBrown Maybe we should just add an option to skip the python module checking. Since I know the Snow environment has all the modules we need, there's no reason to check the modules.

pgrete:
> > @AndrewGaspar could you please confirm that this still works on Snow.
> > I just reverted the TEST_MPIEXEC call in the python module changes as this should also not be required any more (following our final discussion #413 ) and causes problem on Summit (jsrun cannot be called on the dev nodes).
> 
> @JoshuaSBrown Maybe we should just add an option to skip the python module checking. Since I know the Snow environment has all the modules we need, there's no reason to check the modules.

I'd also be in favor of that. This would allow us to keep the logic simple -- especially for our "known" environments.

JoshuaSBrown:
> > > @AndrewGaspar could you please confirm that this still works on Snow.
> > > I just reverted the TEST_MPIEXEC call in the python module changes as this should also not be required any more (following our final discussion #413 ) and causes problem on Summit (jsrun cannot be called on the dev nodes).
> > 
> > 
> > @JoshuaSBrown Maybe we should just add an option to skip the python module checking. Since I know the Snow environment has all the modules we need, there's no reason to check the modules.
> 
> I'd also be in favor of that. This would allow us to keep the logic simple -- especially for our "known" environments.

I'm fine with that.

pgrete:
Sounds good. I'll merge then and we make a new PR with an option to disable module testing.

PR #436, opened 2021-02-02
Update Summit build doc and machine file

pgrete:
## PR Summary

Updates Summit doc and machine file.
Does include #435 so should be merged afterwards.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/cmake/machinecfg/Summit.cmake
+++ b/cmake/machinecfg/Summit.cmake
@@ -18,20 +18,36 @@
 message(STATUS "Loading machine configuration for OLCF's Summit.\n"
   "Supported MACHINE_VARIANT includes 'cuda', 'mpi', and 'cuda-mpi'\n"
   "This configuration has been tested using the following modules: "
-  "module load cuda gcc cmake/3.14.2 python hdf5\n")
+  "module load cuda gcc cmake python hdf5\n")
 
 # common options
 set(Kokkos_ARCH_POWER9 ON CACHE BOOL "CPU architecture")
+set(PARTHENON_DISABLE_OPENMP ON CACHE BOOL "OpenMP support not yet tested in Parthenon.")
+set(CMAKE_BUILD_TYPE "Release" CACHE STRING "Default release build")
+set(MACHINE_VARIANT "cuda-mpi" CACHE STRING "Default build for CUDA and MPI")
 
 # variants
 if (${MACHINE_VARIANT} MATCHES "cuda")
   set(Kokkos_ARCH_VOLTA70 ON CACHE BOOL "GPU architecture")
   set(Kokkos_ENABLE_CUDA ON CACHE BOOL "Enable Cuda")
   set(CMAKE_CXX_COMPILER ${CMAKE_CURRENT_SOURCE_DIR}/external/Kokkos/bin/nvcc_wrapper CACHE STRING "Use nvcc_wrapper")
+else()
+  set(CMAKE_CXX_COMPILER g++ CACHE STRING "Use g++")
+  set(CMAKE_CXX_FLAGS "-fopenmp-simd -ffast-math -fprefetch-loop-arrays" CACHE STRING "Default opt flags")
 endif()
 
+# Setting launcher options independent of parallel or serial test as the launcher always
+# needs to be called from the batch node (so that the tests are actually run on the
+# compute nodes.
+set(TEST_MPIEXEC jsrun CACHE STRING "Command to launch MPI applications")
+set(TEST_NUMPROC_FLAG "-a" CACHE STRING "Flag to set number of processes")
+set(NUM_GPU_DEVICES_PER_NODE "6" CACHE STRING "6x V100 per node")
+set(PARTHENON_ENABLE_GPU_MPI_CHECKS OFF CACHE BOOL "Disable check by default")
+
 if (${MACHINE_VARIANT} MATCHES "mpi")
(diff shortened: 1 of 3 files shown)


cmake/machinecfg/Summit.cmake
```diff
@@ -18,20 +18,36 @@
...
+set(TEST_NUMPROC_FLAG "-a" CACHE STRING "Flag to set number of processes")
+set(NUM_GPU_DEVICES_PER_NODE "6" CACHE STRING "6x V100 per node")
+set(PARTHENON_ENABLE_GPU_MPI_CHECKS OFF CACHE BOOL "Disable check by default")
+
 if (${MACHINE_VARIANT} MATCHES "mpi")
-  set(TEST_MPIEXEC jsrun CACHE STRING "Command to launch MPI applications")
-  set(TEST_NUMPROC_FLAG "-a" CACHE STRING "Flag to set number of processes")
+  # Use a single resource set on a node that includes all cores and GPUs.
+  # GPUs are automatically assigned round robin when run with more than one rank.
   list(APPEND TEST_MPIOPTS "-n" "1" "-g" "6" "-c" "42" "-r" "1" "-d" "packed" "-b" "packed:7" "--smpiargs='-gpu'")
```
JoshuaSBrown, line 50: So if you are just running with mpi I'm guessing the gpu's just won't be called even though they are passed to mpi?

tst/regression/run_test.py
```diff
@@ -138,6 +138,11 @@ class TestError(RuntimeError):
                         nargs=1,
                         help='change MPI run wrapper command (e.g. for job schedulers)')
 
+    parser.add_argument('--mpirun_ranks',
+                        default=[],
+                        action='append',
+                        help='add option to specify number of ranks')
+
```
JoshuaSBrown: Why not add

```
default=1,
nargs=1,
```

Won't there always be a number passed in with the flag?

cmake/machinecfg/Summit.cmake
pgrete: What you do you refer to with "GPUs being called"?
These option ensure that there's a resource set for a node that includes all resources (all CPU cores and all GPUs).
The number of ranks that are launched in the tests is controlled via the `-a` flag and the number of GPUs controlled via the `kokkos-num-devices` argument.

tst/regression/utils/test_case.py
```diff
@@ -106,20 +106,10 @@ def __init__(self,run_test_path,**kwargs):
...
-        for s in argstrings:
-          if s in self.parameters.mpi_opts:
-            index = self.parameters.mpi_opts.index(s)
-            if index < len(self.parameters.mpi_opts) - 1:
-              try:
-                self.parameters.num_ranks = int(self.parameters.mpi_opts[index+1])
-              except ValueError:
-                pass
+        if len(self.parameters.mpi_ranks) == 2:
+            self.parameters.num_ranks = int(self.parameters.mpi_ranks[1])
```
JoshuaSBrown: When will num_ranks be different from mpi_ranks? If the array is empty I see it but what about if mpi_ranks == 4?

tst/regression/run_test.py
pgrete: While in principle I'd be great to enforce this, I'm not sure we can given our logic in the following.
For example, for mpirun the effective line for `run_test.py` reads
`run_test.py --mpirun=mpirun --mpirun_ranks=-n --mpirun_ranks=4 ...`
whereas for jsrun is reads like
`run_test.py --mpirun=jsrun --mpirun_ranks=-a --mpirun_ranks=4 ...`

One option would be to split mpirun_ranks even further and separate the flag (`-n`, `-a`, ...) from the the actual number.

tst/regression/utils/test_case.py
pgrete: As mentioned in the comment above mpirun_ranks will always contain two entries. One for the flag and one for the number.

tst/regression/run_test.py
Yurlungur: I would be in favor of splitting `mpirun_ranks` into `mpirun_rank_flag` and `mpirun_num_ranks` or something.

Yurlungur (approved):
Overall looks good, assuming you tested it. My comment below should be considered non-blocking.

tst/regression/run_test.py
pgrete: Done now #435

pgrete:
> Overall looks good, assuming you tested it. My comment below should be considered non-blocking.

Yes, I tested on Summit and the tests work as described in the build doc and also run on multiple GPUs.

tst/regression/run_test.py
JoshuaSBrown: Great.

tst/regression/utils/test_case.py
JoshuaSBrown: Oh, I completely parsed over len, disregard my comment.

cmake/machinecfg/Summit.cmake
JoshuaSBrown: I'm saying why do you need "-g 6" if you are just running with the "mpi" variant, but I guess it doesn't really matter. If Kokkos is disabled anyway the gpus won't be utilized. Regardless, it's a non-issue.

JoshuaSBrown (approved):

PR #437, opened 2021-02-02
WIP: Explicit template instantiation

brryan:
## PR Summary

We have some templated classes that are used throughout the code, usually with the same template parameter, e.g. `ParArrayND<Real>`. This PR explicitly instantiates some of these template for common types. In particular, after this first pass I shaved 23 seconds off of my compile time (for an Ubuntu VM on a laptop).

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/src/interface/variable_pack.cpp
+++ b/src/interface/variable_pack.cpp
@@ -21,14 +21,14 @@ template class VariableFluxPack<Real>;
 template void FillVarView<Real>(const vpack_types::VarList<Real> &vars,
                                 PackIndexMap *vmap, ViewOfParArrays<Real> &cv,
                                 ParArray1D<int> &sparse_assoc,
-                                ParArray1D<int> &vector_component, bool coarse = false);
+                                ParArray1D<int> &vector_component, bool coarse);
 template void FillFluxViews(const vpack_types::VarList<Real> &vars, PackIndexMap *vmap,
                             const int ndim, ViewOfParArrays<Real> &f1,
                             ViewOfParArrays<Real> &f2, ViewOfParArrays<Real> &f3);
 template VariableFluxPack<Real> MakeFluxPack(const vpack_types::VarList<Real> &vars,
                                              const vpack_types::VarList<Real> &flux_vars,
-                                             PackIndexMap *vmap = nullptr);
+                                             PackIndexMap *vmap);
 template VariablePack<Real> MakePack(const vpack_types::VarList<Real> &vars,
-                                     PackIndexMap *vmap = nullptr, bool coarse = false);
+                                     PackIndexMap *vmap, bool coarse);
 
 } // namespace parthenon
--- a/src/interface/variable_pack.cpp
+++ b/src/interface/variable_pack.cpp
@@ -12,6 +12,7 @@
 //=======================================================================================
 
 #include "variable.hpp"
+
 #include "variable_pack.hpp"
 
 namespace parthenon {
--- a/src/interface/variable_pack.cpp
+++ b/src/interface/variable_pack.cpp
@@ -1,0 +1,34 @@
+//========================================================================================
+// (C) (or copyright) 2020-2021. Triad National Security, LLC. All rights reserved.
+//
+// This program was produced under U.S. Government contract 89233218CNA000001 for Los
+// Alamos National Laboratory (LANL), which is operated by Triad National Security, LLC
+// for the U.S. Department of Energy/National Nuclear Security Administration. All rights
+// in the program are reserved by Triad National Security, LLC, and the U.S. Department
+// of Energy/National Nuclear Security Administration. The Government is granted for
+// itself and others acting on its behalf a nonexclusive, paid-up, irrevocable worldwide
+// license in this material to reproduce, prepare derivative works, distribute copies to
+// the public, perform publicly and display publicly, and to permit others to do so.
+//=======================================================================================
+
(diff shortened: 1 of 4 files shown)


brryan:
Adding variable packs to the mix shaves another 12 seconds off compile time on my end. 

I could have easily done this not quite right (or extremely wrong) so I invite C++ experts to suggest changes

src/interface/variable_pack.cpp
```diff
@@ -0,0 +1,34 @@
...
+#include "variable.hpp"
+
+namespace parthenon {
+
+template class VariablePack<Real>;
+template class VariableFluxPack<Real>;
+template void FillVarView<Real>(const vpack_types::VarList<Real> &vars,
+                                PackIndexMap *vmap, ViewOfParArrays<Real> &cv,
+                                ParArray1D<int> &sparse_assoc,
+                                ParArray1D<int> &vector_component, bool coarse = false);
```
Yurlungur: sometimes I get errors when including default arguments in the implementation file. Should they be removed?

Yurlungur (changes requested):
This is a nice improvement. Thanks for doing it! Can you also add a change log?

src/interface/variable_pack.cpp
brryan: I'm not familiar with that issue but if you've had problems lets just remove the default args (I just did this)

AndrewGaspar (changes requested):
You need to prove this doesn't result in runtime performance regression first. I'm not sure what the interaction is between explicit template instantiations and forceinline - it wouldn't surprise me if it's ignored.

src/interface/variable_pack.cpp
```diff
@@ -0,0 +1,35 @@
+//========================================================================================
+// (C) (or copyright) 2020-2021. Triad National Security, LLC. All rights reserved.
```
AndrewGaspar, line 2: Just `2021`.

src/parthenon_arrays.cpp
```diff
@@ -0,0 +1,39 @@
+//========================================================================================
+// (C) (or copyright) 2020-2021. Triad National Security, LLC. All rights reserved.
```
AndrewGaspar, line 2:
```suggestion
// (C) (or copyright) 2021. Triad National Security, LLC. All rights reserved.
```

src/interface/variable_pack.cpp
AndrewGaspar:
```suggestion
// (C) (or copyright) 2021. Triad National Security, LLC. All rights reserved.
```

src/parthenon_arrays.cpp
```diff
@@ -0,0 +1,39 @@
...
+template class ParArrayNDGeneric<Kokkos::View<Real ******>>;
+template class ParArrayNDGeneric<Kokkos::View<int ******>>;
+
+template class ParArrayNDGeneric<device_view_t<Real, LayoutWrapper>>;
+template class ParArrayNDGeneric<device_view_t<int, LayoutWrapper>>;
+
+template struct FaceArray<Real>;
+template struct EdgeArray<Real>;
+
+} // namespace parthenon
```
AndrewGaspar, line 39: Wait... I don't see any new `extern template`s in any header files. I don't see how this would improve compile time. You're instantiating these templates in an object file, but no other source file knows this is the case, so they would still be instantiated in every object file they're used in.

Could you re-run your compile performance tests?

AndrewGaspar:
> You need to prove this doesn't result in runtime performance regression first. I'm not sure what the interaction is between explicit template instantiations and forceinline - it wouldn't surprise me if it's ignored.

Testing on godbolt, looks like we still get inlining, even without forceinline, so we should be fine. https://godbolt.org/z/x4vabr

brryan:
> > You need to prove this doesn't result in runtime performance regression first. I'm not sure what the interaction is between explicit template instantiations and forceinline - it wouldn't surprise me if it's ignored.
> 
> Testing on godbolt, looks like we still get inlining, even without forceinline, so we should be fine. https://godbolt.org/z/x4vabr

Thanks for checking this @agaspar. What I did should be analogous to what's already in e.g. `interface/variable.cpp`, e.g. `template class CellVariable<Real>;`, so if this PR causes performance issues the problem will be more widespread in the code I think.

src/parthenon_arrays.cpp
brryan: I wondered about how the compiler would know to use the explicit instantiations, for just the reason you say, but was under the impression compilers could play sneaky tricks here. But I just re-ran my compile time test and the code is compiling at the new, faster speed with these explicit instantiations commented out! I'm changing this back to WIP until I can dig into this further.

If we do need `extern template`s everywhere this change may not be worth it if that introduced too much clutter.

AndrewGaspar: You don't need them "everywhere", you just need to put them in preferably the same header as where the template type is declared.

AndrewGaspar:
> What I did should be analogous to what's already in e.g. interface/variable.cpp, e.g. template class CellVariable<Real>;

There aren't any member functions on `CellVariable` that are called in performance critical sections of the code, so it's not a big deal. But it would be really bad if `Kokkos::View::operator()` wasn't inlined.

jlippuner:
Will revisit if needed

PR #440, opened 2021-02-04
Add abstraction for allocating a unique_ptr to device memory

brryan:
## PR Summary

It can be convenient to allocate memory on the device, for example when selecting derived classes at runtime or avoiding large copies when launching compute kernels. This PR adds functions for allocating device memory for an object (either uninitialized or a copy of an object in host memory) and returning a `unique_ptr` to that device object. The custom destructor cleans up the device memory when the `unique_ptr` releases its memory. 

I got this implementation from the slide deck "Runtime Polymorphism in Kokkos Applications"

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


pgrete:
@brryan I' not sure I fully understand the use case.
How is this different from using a plain `ParArray1D`  (and it's mirror)?
I think an example where and how it would be used would be helpful.

src/kokkos_abstraction.hpp
```diff
@@ -635,6 +635,36 @@ struct SpaceInstance<Kokkos::Cuda> {
...
+  void operator()(T *ptr) {
+    Kokkos::parallel_for(
+        Kokkos::RangePolicy<DevExecSpace>(0, 1),
+        KOKKOS_LAMBDA(const int i) { ptr->~T(); });
+    Kokkos::kokkos_free<DevMemSpace>(ptr);
+  }
+};
+
+template <typename T>
+std::unique_ptr<T, DeviceDeleter> DeviceAllocate() {
```
Yurlungur: This is cool. I knew this was possible with unique pointers, but I didn't fully understand how to make it work.

tst/unit/kokkos_abstraction.cpp
```diff
@@ -698,3 +698,34 @@ TEST_CASE("Overlapping SpaceInstances", "[wrapper][performance]") {
         "Few Threads Long Kernel");
   }
 }
+
+TEST_CASE("Test Device Allocation", "[wrapper]") {
+
+  struct MyTestStruct {
+    int i;
+  };
+  parthenon::ParArray1D<int> buffer("Testing buffer", 1);
```
Yurlungur: Can we add a test here showing the application for inheritance?

Yurlungur (changes requested):
This is awesome @brryan . This would make treating inheritence for, e.g., the particle boundaries stuff or for coordinate systems. The one thing I'd like to see is a test of this applied to inheritence to show how it works in a simple example.

AndrewGaspar:
I think this needs more justification, and if you can provide a good example of a use case, then we can merge. But I have some concerns.

So the main value here over a 0D `Kokkos::View` is that it runs the destructor in the device context. In exchange, it gives up the benefit of a 0D `Kokkos::View` of attaching the lifetime of the object to all outstanding async kernels.

Additionally, I think that this is probably an anti-pattern - launching an individual kernel to deallocate a single object seems like it could make it easy to step into a performance pitfall. Also, what kind of things do you want to do inside a device-side destructor anyway?

Can you provide more context?

src/kokkos_abstraction.hpp
```diff
@@ -635,6 +635,36 @@ struct SpaceInstance<Kokkos::Cuda> {
 };
 #endif
 
+struct DeviceDeleter {
```
AndrewGaspar, line 643: You probably should make the execution space a member of the deleter since at some point we intend to do more stuff with streams.

Yurlungur:
> I think this needs more justification, and if you can provide a good example of a use case, then we can merge. But I have some concerns.
> 
> So the main value here over a 0D `Kokkos::View` is that it runs the destructor in the device context. In exchange, it gives up the benefit of a 0D `Kokkos::View` of attaching the lifetime of the object to all outstanding async kernels.
> 
> Additionally, I think that this is probably an anti-pattern - launching an individual kernel to deallocate a single object seems like it could make it easy to step into a performance pitfall. Also, what kind of things do you want to do inside a device-side destructor anyway?
> 
> Can you provide more context?

@brryan should jump in and correct me if I'm wrong. I believe the intent here is to make polymorphism work on GPU. I.e., if you have
```C++
class DerivedClass : public BaseClass {
public:
  KOKKOS_FUNCTION blah { // stuff }
};
```
Then this method lets you put a `DerivedClass` into a pointer pointing to `BaseClass` on GPU. You can't do that with views, because you can't make a view of pointers.

brryan:
Thanks for the comments @AndrewGaspar and @Yurlungur 

> > I think this needs more justification, and if you can provide a good example of a use case, then we can merge. But I have some concerns.
> > So the main value here over a 0D `Kokkos::View` is that it runs the destructor in the device context. In exchange, it gives up the benefit of a 0D `Kokkos::View` of attaching the lifetime of the object to all outstanding async kernels.
> > Additionally, I think that this is probably an anti-pattern - launching an individual kernel to deallocate a single object seems like it could make it easy to step into a performance pitfall. Also, what kind of things do you want to do inside a device-side destructor anyway?
> > Can you provide more context?
> 
> @brryan should jump in and correct me if I'm wrong. I believe the intent here is to make polymorphism work on GPU. I.e., if you have
> 
> ```c++
> class DerivedClass : public BaseClass {
> public:
>   KOKKOS_FUNCTION blah { // stuff }
> };
> ```
> 
> Then this method lets you put a `DerivedClass` into a pointer pointing to `BaseClass` on GPU. You can't do that with views, because you can't make a view of pointers.

Yes @Yurlungur, thanks, that's just what this is for. I wanted to do essentially exactly what you wrote in code above for particle boundaries (for which we want to choose device functions at runtime based on boundary condition choices in parameter input files). My understanding is this may also help you with some code patterns you were leveraging `std::variant` for in downstream contexts. 

@AndrewGaspar I agree this is not performant code but the idea (at least for my use cases) is that really this is only going to be called on initialization/finalization. 

> > Also, what kind of things do you want to do inside a device-side destructor anyway?

The destructor I wrote out is only to let the host-side `unique_ptr` deallocate the device-side object -- is that what you're referring to?

I'll expand the test to showcase inheritance, and expand the docs to warn people to use these functions only sparingly.

src/kokkos_abstraction.hpp
brryan: `DevExecSpace` is a type, I think, so I'm not sure how to store this -- would it help to template `DeviceDeleter` on `ExecSpace` and `MemSpace`?

brryan:
@pgrete Sorry, forgot to tag you in the above, thanks for taking a look -- this is largely for inheritance patterns which `View`s don't support. An example is now shown in the unit test.

src/kokkos_abstraction.hpp
AndrewGaspar: Yeah, that's right.

src/kokkos_abstraction.hpp
```diff
@@ -635,6 +635,36 @@ struct SpaceInstance<Kokkos::Cuda> {
...
+  template <typename T>
+  void operator()(T *ptr) {
+    Kokkos::parallel_for(
+        Kokkos::RangePolicy<DevExecSpace>(0, 1),
+        KOKKOS_LAMBDA(const int i) { ptr->~T(); });
```
AndrewGaspar: > The destructor I wrote out is only to let the host-side unique_ptr deallocate the device-side object -- is that what you're referring to?

You're calling the destructor for `T` on the device here.

AndrewGaspar: Oh, yes, that's what you're saying.

I get why you need to run the constructor on the device, in order to set the virtual function pointers. I'd need to see a pretty compelling case why you need to call the destructor on the device. I wonder if it makes sense to require `std::is_trivial_destructible`.

src/kokkos_abstraction.hpp
brryan: I'm not familiar with the stream stuff but I added the template parameters to `DeviceAllocate` and `DeviceCopy` on the assumption that they will be needed there as well.

src/kokkos_abstraction.hpp
brryan: Right, yes, I forgot about this line, thanks for the reminder. 

I don't see a need now to call the destructor on the device. I think if we're going to remove this destructor line, though, we should check that `T` is trivially destructible, which I will add now.

src/kokkos_abstraction.hpp
brryan: I'm not familiar with the stream stuff but I added the template parameters to DeviceAllocate and DeviceCopy on the assumption that they will be needed there as well.

pgrete:
> Thanks for the comments @AndrewGaspar and @Yurlungur
> 
> > > I think this needs more justification, and if you can provide a good example of a use case, then we can merge. But I have some concerns.
> > > So the main value here over a 0D `Kokkos::View` is that it runs the destructor in the device context. In exchange, it gives up the benefit of a 0D `Kokkos::View` of attaching the lifetime of the object to all outstanding async kernels.
> > > Additionally, I think that this is probably an anti-pattern - launching an individual kernel to deallocate a single object seems like it could make it easy to step into a performance pitfall. Also, what kind of things do you want to do inside a device-side destructor anyway?
> > > Can you provide more context?
> > 
> > 
> > @brryan should jump in and correct me if I'm wrong. I believe the intent here is to make polymorphism work on GPU. I.e., if you have
> > ```c++
> > class DerivedClass : public BaseClass {
> > public:
> >   KOKKOS_FUNCTION blah { // stuff }
> > };
> > ```
> > 
> > 
> > Then this method lets you put a `DerivedClass` into a pointer pointing to `BaseClass` on GPU. You can't do that with views, because you can't make a view of pointers.
> 
> Yes @Yurlungur, thanks, that's just what this is for. I wanted to do essentially exactly what you wrote in code above for particle boundaries (for which we want to choose device functions at runtime based on boundary condition choices in parameter input files). My understanding is this may also help you with some code patterns you were leveraging `std::variant` for in downstream contexts.
> 
> @AndrewGaspar I agree this is not performant code but the idea (at least for my use cases) is that really this is only going to be called on initialization/finalization.

I'm wondering if we should ask the Kokkos team for additional input here.
If this is a more general/common use case then it's likely that there some best practices have already been established.

I just want to be cautious because (despite all warnings in terms of "should be used with care"/"is not performant") once a feature is there it will be (ab)used in all kinds of ways almost certainly.

brryan:
@pgrete I copied this approach from [these slides](https://www.osti.gov/biblio/1592283) published in 2019. I think this is from the Kokkos team -- at least, Christian Trott is on the author list. They provide the same warning about efficiency: `Constructing many polymorphic device objects per-timestep is problematic, make persistent if possible`

Polymorphism would be really useful to me and I don't see a better way to handle this than calling the constructor on device in a small kernel. On some level we can only do so much to prevent downstream consumers from writing non-performant code, unfortunately. I'm happy to try to make this concern more explicit in the code but I don't have any good ideas right now (except maybe renaming `DeviceCopy` to `DeviceAllocateCopy`)?

Yurlungur (approved):

src/kokkos_abstraction.hpp
AndrewGaspar: Oh, you don't need the `ES` anymore if you're not calling the destructor.

src/kokkos_abstraction.hpp
```diff
@@ -635,6 +638,38 @@ struct SpaceInstance<Kokkos::Cuda> {
...
+  void operator()(T *ptr) {
+    Kokkos::kokkos_free<MS>(ptr);
+  }
+};
+
+template <typename T, typename ES = DevExecSpace, typename MS = DevMemSpace>
+std::unique_ptr<T, DeviceDeleter<ES, MS>> DeviceAllocate() {
+  PARTHENON_REQUIRE(std::is_trivially_destructible<T>::value,
+                    "DeviceAllocate only supports trivially destructible classes!");
+  auto *p = static_cast<T *>(Kokkos::kokkos_malloc<MS>(sizeof(T)));
```
AndrewGaspar: _Technically_ you should put this pointer (and in `DeviceCopy`) into `unique_ptr` before calling any of the Kokkos functions, since they may throw. If you don't, then this would leak should you catch the exception and continue running. However, this is a pretty minor issue - exceptions almost always result in code termination in parthenon.

src/kokkos_abstraction.hpp
brryan: nice catch, thanks (fixed)

src/kokkos_abstraction.hpp
brryan: That's a good point, this is a pretty generic function so it's best to do things right here. Done.

src/kokkos_abstraction.hpp
```diff
@@ -635,6 +638,42 @@ struct SpaceInstance<Kokkos::Cuda> {
...
+  template <typename T>
+  void operator()(T *ptr) {
+    Kokkos::kokkos_free<MS>(ptr);
+  }
+};
+
+template <typename T, typename ES = DevExecSpace, typename MS = DevMemSpace>
+std::unique_ptr<T, DeviceDeleter<MS>> DeviceAllocate() {
+  PARTHENON_REQUIRE(std::is_trivially_destructible<T>::value,
+                    "DeviceAllocate only supports trivially destructible classes!");
```
AndrewGaspar: Sorry, should have caught this earlier.
```suggestion
  static_assert(std::is_trivially_destructible<T>::value,
                    "DeviceAllocate only supports trivially destructible classes!");
```

src/kokkos_abstraction.hpp
```diff
@@ -635,6 +638,42 @@ struct SpaceInstance<Kokkos::Cuda> {
...
+  Kokkos::parallel_for(
+      Kokkos::RangePolicy<ES>(0, 1), KOKKOS_LAMBDA(const int i) { new (p) T(); });
+  Kokkos::fence();
+  return up;
+}
+
+template <typename T, typename ES = DevExecSpace, typename MS = DevMemSpace>
+std::unique_ptr<T, DeviceDeleter<MS>> DeviceCopy(const T &host_object) {
+  PARTHENON_REQUIRE(std::is_trivially_destructible<T>::value,
+                    "DeviceCopy only supports trivially destructible classes!");
```
AndrewGaspar:
```suggestion
  static_assert(std::is_trivially_destructible<T>::value,
                    "DeviceCopy only supports trivially destructible classes!");
```

src/kokkos_abstraction.hpp
brryan: thanks!

AndrewGaspar (approved):

pgrete (approved):
Thanks for taking care of this and also adding the additional pointer to the Sandia presentation.
The only additional thing that I noticed is the global fence, which may at some point be a another performance penalty when we move to streams.
Though I don't think we need to solve this here and now so I just open an issue so that we keep it in mind.

PR #444, opened 2021-02-10
Output performance metrics for ci

JoshuaSBrown:
## PR Summary

This pr allows the ci to run performance regression tests on the advection performance test, because it outputs the performance metrics in a parsable file format. 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/tst/regression/test_suites/advection_performance/advection_performance.py
+++ b/tst/regression/test_suites/advection_performance/advection_performance.py
@@ -3,7 +3,7 @@
 # Copyright(C) 2020 The Parthenon collaboration
 # Licensed under the 3-clause BSD License, see LICENSE file for details
 #========================================================================================
-# (C) (or copyright) 2020. Triad National Security, LLC. All rights reserved.
+# (C) (or copyright) 2020-2021. Triad National Security, LLC. All rights reserved.
 #
 # This program was produced under U.S. Government contract 89233218CNA000001 for Los
 # Alamos National Laboratory (LANL), which is operated by Triad National Security, LLC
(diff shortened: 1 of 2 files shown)


Yurlungur (approved):
:+1: but should this wait until the advanced advection test is in master?

JoshuaSBrown:
> 👍 but should this wait until the advanced advection test is in master?

No, that pr is orthogonal to this one, but thanks for pointing it out.

tst/regression/test_suites/advection_performance/advection_performance.py
```diff
@@ -78,8 +78,13 @@ def Analyse(self, parameters):
 
         perfs = np.array(perfs)
 
-        # Plot results
+        # Save performance metrics to file
+        with open('performance_metrics.txt', 'w') as writer:
+          writer.write("zone-cycles/s | Normalize Overhead | Meshblock Size\n")
+          for ind, mb_size in enumerate(mb_sizes):
+              writer.write("%14e %14e %14e\n" % (perfs[ind],perfs[0]/perfs[ind],mb_size))
```
pgrete:
```suggestion
              writer.write("%.2e %.2e %d\n" % (perfs[ind],perfs[0]/perfs[ind],mb_size))
```
I'm not sure `%14e` has any effect. With `%.2e` we'll get three significant digits total, which should be fine in terms of accuracy for these kind of tests.

pgrete (approved):
I'm hitting approve so that there are two approval and merging is enabled, but the suggested change should go in first (unless I'm missing the intend of using `%14e`.

tst/regression/test_suites/advection_performance/advection_performance.py
dholladay00: Are these values standard metrics to print across multiple tests? If so, I think we should standardize the format. If not, @pgrete's suggestions aren't unreasonable as this would allow us to tell % level changes in performance which is sufficient for our purposes.

JoshuaSBrown: Sounds good, and yes they are.

PR #451, opened 2021-02-17
Remove ran2 interface (and source)

pgrete:
## PR Summary

We currently provide the interface to ran2 and have the source in the tree, but don't compile it.
The alternative to compiling it would be to remove it (so that people use the std random number generators).
I'm in favor of the latter but wanted to hear opinions first as we currently provide the interface in our API.

PS: I tagged the people that I think may be affected most downstream (if at all).

## PR Checklist

- [x] Code passes cpplint
- [ x New features are documented.
- [x] Code is formatted


JoshuaSBrown:
I do not have an opinion on this, but I approve this pr. It makes sense that we should be compiling the code included with parthenon if we are going to keep it.

JoshuaSBrown (approved):

Yurlungur (approved):
My preference is to remove it. But I'm willing to compile it if that's what people want to do.

pgrete:
As discussed during the call, we'll remove the custom interface and source.

PR #452, opened 2021-02-18
Allow option for disabling copyright checks while building

JoshuaSBrown:
## PR Summary

The build failures are problematic when running things in the ci this provides users with the option to disable to copyright checks. It might also be of use while developing before files are pushed. 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


Yurlungur (approved):
Nice little quality of life improvement thanks.

pgrete (commented):
I also think this makes sense and I'm wondering if it'd be worth to go even one step further, i.e., disable this by default and add a cmake configure step to the "style" check we currently perform anyway.
That way code checks would also work everywhere and and all "style" related checks are independent.

JoshuaSBrown:
> I also think this makes sense and I'm wondering if it'd be worth to go even one step further, i.e., disable this by default and add a cmake configure step to the "style" check we currently perform anyway.
> That way code checks would also work everywhere and and all "style" related checks are independent.

Do I understand this correctly? You are suggesting simply turning on as part of the ci during each commit as a check? If that's the case I think that is a good idea as long the license is being checked somewhere before things get merged I'm fine with changing the default.

jlippuner (approved):
Great improvement and much needed!

I agree with @pgrete that I'd prefer to have the copyright check turned off by default and only checked as part of CI. In fact, I would even want this for the linting. Sometimes it's a bit annoying if one is writing code and testing it and it has to be squeaky clean before it even builds... For rapid development, code should build if it's valid C++ and then style (i.e. copyright, linting) gets checked before merging PR.

pgrete:
> Great improvement and much needed!
> 
> I agree with @pgrete that I'd prefer to have the copyright check turned off by default and only checked as part of CI. In fact, I would even want this for the linting. Sometimes it's a bit annoying if one is writing code and testing it and it has to be squeaky clean before it even builds... For rapid development, code should build if it's valid C++ and then style (i.e. copyright, linting) gets checked before merging PR.

+1 from me disable linting in CI, too.
So the suggested fix would be 
- disable linting and copyright check in our main `CMakeLists.txt` (so that it's not automatically called also in downstream codes)
- no changes  in the CI runner (given that the default is now `off`, CI should always run independent of linting and copyright issues
- extend the "Check Formatting / Check C++ Formatting (push)" GitHub Action based test (e.g., https://github.com/lanl/parthenon/pull/452/checks?check_run_id=1946624095) so that it configures and compiles with linting and copyright check on. This will prevent accidental merges that don't follow style and copyright.

jlippuner:
I'm not quite sure how to add linting and copyright check to the GitHub Check Formatting action, so I added the copyright check to the GitLab CI instead. The GitLab CI already runs `cpplint` in its Style pipeline and I added another stage to run CMake with the copyright check turned on.

So the changes I've made are:
- Disable copyright check by default
- Disable linting by default
- Add GitLab CI stage to check copyright
- Also check copyright (when checked) on files in `tst`

I'm going to push a commit that breaks the linting and copyright to test the CI catches it. Then I will revert that commit.

jlippuner:
@JoshuaSBrown @pgrete @Yurlungur This is ready to go if you guys are happy with my changes. It seems to be working as intended now (at least as I understand the intention ;)

Yurlungur (approved):
I'm happy. Let's pull the trigger.

JoshuaSBrown:
Thanks @jlippuner for the tweaks!

PR #453, opened 2021-02-20
Fix bounds and use vector

pgrete:
## PR Summary

Array bounds were incorrect and resulted in the following issue with `maxlevel > 4`:
```
#0  0x0000200000bbfbf0 in raise () from /lib64/libc.so.6
#1  0x0000200000bc1f6c in abort () from /lib64/libc.so.6
#2  0x0000200000c08d10 in __libc_message () from /lib64/libc.so.6
#3  0x0000200000c19be0 in free () from /lib64/libc.so.6
#4  0x00002000009d7968 in operator delete (ptr=<optimized out>) at /sw/summit/gcc/6.4.0/src/gcc-6.4.0/libstdc++-v3/libsupc++/del_op.cc:49
#5  0x00002000009d7a28 in operator delete[] (ptr=<optimized out>) at /sw/summit/gcc/6.4.0/src/gcc-6.4.0/libstdc++-v3/libsupc++/del_opv.cc:35
#6  0x00000000100bc5bc in std::default_delete<int []>::operator()<int> (this=<optimized out>, __ptr=0x375a0510) at /autofs/nccs-svm1_sw/summit/gcc/6.4.0/include/c++/6.4.0/bits/unique_ptr.h:110
```

I can confirm that the code now does not crash any more with this issue.

Also changed `unique_ptr` to `vector` as discussed on Matrix (there's practically no advantage to use a `unique_ptr` based array here and `vector` is a safer option).

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


JoshuaSBrown (approved):
@pgrete this is great thanks for fixing this!!!

src/mesh/mesh.cpp
```diff
@@ -811,12 +811,9 @@ void Mesh::OutputMeshStructure(const int ndim,
...
 
   // compute/output number of blocks per level, and cost per level
-  auto nb_per_plevel = std::make_unique<int[]>(max_level);
-  auto cost_per_plevel = std::make_unique<int[]>(max_level);
-  for (int i = 0; i <= max_level; ++i) {
-    nb_per_plevel[i] = 0;
-    cost_per_plevel[i] = 0;
-  }
+  auto nb_per_plevel = std::vector<int>(max_level + 1);
+  auto cost_per_plevel = std::vector<int>(max_level + 1);
```
jlippuner: Not strictly necessary, but if you want to be very clear and verbose, you can explicitly pass the value to initialize to the constructor. Also, `auto` is not needed anymore ;)

```suggestion
  std::vector<int> nb_per_plevel(max_level + 1, 0);
  std::vector<int> cost_per_plevel(max_level + 1, 0);
```

src/mesh/mesh.cpp
```diff
@@ -835,12 +832,9 @@ void Mesh::OutputMeshStructure(const int ndim,
...
   // compute/output number of blocks per rank, and cost per rank
   std::cout << "Number of parallel ranks = " << Globals::nranks << std::endl;
-  auto nb_per_rank = std::make_unique<int[]>(Globals::nranks);
-  auto cost_per_rank = std::make_unique<int[]>(Globals::nranks);
-  for (int i = 0; i < Globals::nranks; ++i) {
-    nb_per_rank[i] = 0;
-    cost_per_rank[i] = 0;
-  }
+  auto nb_per_rank = std::vector<int>(Globals::nranks);
+  auto cost_per_rank = std::vector<int>(Globals::nranks);
```
jlippuner:
```suggestion
  std::vector<int> nb_per_rank(Globals::nranks, 0);
  std::vector<int> cost_per_rank(Globals::nranks, 0);
```

jlippuner (approved):
Looks great! I have some hair-splitting stylistic suggestions, feel free to accept or not.

src/mesh/mesh.cpp
Yurlungur: I prefer this also.

Yurlungur (approved):
Good catch!

PR #459, opened 2021-02-24
WIP Joshua s brown/make template

JoshuaSBrown:
This is simply to provide an example of how templates could be used in the swarm classes. @brryan hopefully you can use some of this. This does not fully compile yet because I cannot spend any more time on this, but it should show you how to move forward. 

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted


brryan:
@JoshuaSBrown Thanks for trying this out! I pulled this into another branch and fiddled with it a little bit but I'm stuck in a new place:
```c++
inline void Swarm::Add_(const std::string & label) {                                                 
  int vec_type = RealVec;                                                                            
  if( std::is_same<T,int> ) {                                                                        
    vec_type = IntVec;                                                                               
  }                                                                                                  
  auto var = std::make_shared<ParticleVariable<int>>(label, nmax_pool_, metadata);                   
  std::get<vec_type>(Vectors_).push_back(var);                                                       
  std::get<vec_type>(Maps_)[label] = var;                                                            
}      
```

This gives an error 
```
/home/brryan/github/parthenon/src/interface/swarm.hpp:349:30: error: the value of 'vec_type' is not usable in a constant expression
  349 |   std::get<vec_type>(Vectors_).push_back(var);
  ```

The only way I see around that is to have something like
```c++
if( std::is_same<T,int> ) {
    auto var = std::make_shared<ParticleVariable<int>>(label, nmax_pool_, metadata);                   
  std::get<IntVec>(Vectors_).push_back(var);                                                       
  std::get<IntVec>(Maps_)[label] = var;
} else {
  auto var = std::make_shared<ParticleVariable<Real>>(label, nmax_pool_, metadata);                   
  std::get<RealVec>(Vectors_).push_back(var);                                                       
  std::get<RealVec>(Maps_)[label] = var;
}
```

but that's basically the code duplication we already have, right? I may not be seeing something but this seems like a fair bit of work so I'm not sure it's worth it unless we can really remove all code duplication, since it wasn't too too many duplicated lines.

JoshuaSBrown:
There is an error in that piece, should be is_same<T,int>::value, if that doesn't fix the problem, you might try setting vec_type to a const int and using a ternary operator to assign to it.

brryan (approved):

PR #463, opened 2021-02-27
Add `PARTHENON_ENABLE_TESTING` and `PARTHENON_ENABLE_PYTHON_MODULE_CHECK` and downstream testing instructions

pgrete:
## PR Summary

Another quality of life improvement.

The originally used `BUILD_TESTING` is a CTest option (and thus could also be used in downstream codes in other contexts).
I introduced a `PARTHENON_ENABLE_TESTING` option (that defaults to the `BUILD_TESTING` value), which allows downstream codes to easily disable the Parthenon tests.

I addition, there's a new `PARTHENON_ENABLE_PYTHON_MODULE_CHECK` variable (defaults to `PARTHENON_ENABLE_TESTING`) to disable the module check (which sometimes causes problem on cluster where the development node environment is different from the batch node and different from the compute node).

Finally, there a new internal variable `PARTHENON_TEST_DIR` that is used to use the absolute path to `run_test.py` in the Parthenon repo.
This, in combination with some additional internal refactoring, now allows downstream codes to reuse the regression test infrastructure, see https://gitlab.com/theias/hpc/jmstone/athena-parthenon/athenapk/-/merge_requests/6

I also used the opportunity to add documentation on how to use the test infrastructure downstream.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


cmake/TestSetup.cmake
```diff
@@ -25,9 +25,12 @@ if(Python3_Interpreter_FOUND)
     message(FATAL_ERROR "Python version requirements not satisfied for running regression tests.")
   endif()
 endif()
-# Ensure all required packages are present
-include(${PROJECT_SOURCE_DIR}/cmake/PythonModuleCheck.cmake)
-required_python_modules_found("${REQUIRED_PYTHON_MODULES}")
+
+if (PARTHENON_ENABLE_PYTHON_MODULE_CHECK)
+  # Ensure all required packages are present
+  include(${PROJECT_SOURCE_DIR}/cmake/PythonModuleCheck.cmake)
```
JoshuaSBrown: Can you change PROJECT_SOURCE_DIR to PARTHENON_SOURCE_DIR?

CMakeLists.txt
```diff
@@ -226,6 +228,7 @@ if (NOT TARGET Kokkos::kokkos)
 endif()
 
 # Build Tests and download Catch2
+set(PARTHENON_TEST_DIR ${CMAKE_CURRENT_SOURCE_DIR}/tst/regression CACHE INTERNAL "Absolute path dir Parthenon regression test directory.")
```
JoshuaSBrown: I would change CMAKE_CURRENT_SOURCE_DIR to PARTHENON_SOURCE_DIR.

docs/README.md
```diff
@@ -48,6 +48,40 @@ The former is considered the safer option as it prevents accidental overwriting
...
+            __init__.py  # <-- empty file
+            my_first_test/
+                __init.py__  # <-- empty file
+                my_first_test.py
+```
+3. Contents of `tst/regression/CMakeLists.txt`
+```
+
+# import Parthenon setup_test_serial and setup_test_parallel
+include(${PROJECT_SOURCE_DIR}/external/parthenon/cmake/TestSetup.cmake)
```
JoshuaSBrown, line 79: Might want to point out somewhere above that the user also needs to have the parthenon submodule in external folder in the root project source directory. I know it may seem obvious... but it would probably make it more clear.

JoshuaSBrown: Based on your documentation I would add something like. For easiest use the downstream author should have a directory structure with the following form:

```
project_foo/external/parthenon - (parthenon submodule)
project_foo/bin - (location of downstream binaries/apps/executables)
```

Maybe add something like this to your existing description.

tst/regression/utils/test_case.py
```diff
@@ -138,8 +140,8 @@ def __checkAndGetRegressionTestFolder(self,test_dir):
         else:
             return os.path.abspath(test_dir)
 
-    def __checkRegressionTestScript(self,test_base_name):
-        python_test_script = os.path.join(self.__run_test_py_path,'test_suites',test_base_name,test_base_name + ".py")
+    def __checkRegressionTestScript(self, test_dir, test_base_name):
+        python_test_script = os.path.join(test_dir, test_base_name + ".py")
```
JoshuaSBrown, line 144: Nice, this is cleaner.

JoshuaSBrown (approved):
LGTM. Thanks @pgrete

CMakeLists.txt
Yurlungur: Actually, I think it has to be `CMAKE_CURRENT_SOURCE_DIR`. The intent here is for downstream codes to be able to build tests too.

Yurlungur (approved):
This is awesome. Thanks for setting this up @pgrete ! One question: with the `PARTHENON_SOURCE_DIR` and `CMAKE_CURRENT_SOURCE_DIR` stuff... can we build regression tests for a downstream code *and* parthenon at the same time? or is it either-or? Doesn't really matter which is the case... but it should be documented.

CMakeLists.txt
JoshuaSBrown: I don't think that is accurate, because he is using PARTHENON_TEST_DIR to point to the run_test.py script, which will always be in the Parthenon repo if I'm not mistaken.

cmake/TestSetup.cmake
pgrete: Done.

CMakeLists.txt
pgrete: You're both right and both versions did achieve this. The important piece was that it's an absolute path to the `run_test.py` provided by `Parthenon`.
I now removed that variable entirely and directly use the `parthenon_SOURCE_DIR` in `TestSetup.cmake`

docs/README.md
pgrete: I added the info to the parthenon directory location and made the path to the bin in the example absolute.

pgrete:
> This is awesome. Thanks for setting this up @pgrete ! One question: with the `PARTHENON_SOURCE_DIR` and `CMAKE_CURRENT_SOURCE_DIR` stuff... can we build regression tests for a downstream code _and_ parthenon at the same time? or is it either-or? Doesn't really matter which is the case... but it should be documented.

I can confirm that the tests now completely integrate, e.g.,
```
$ ctest -N
Test project /home/pgrete/src/athenapk/build
  Test  #1: "unit_tests:Adding MeshBlockData objects to a DataCollection"
  Test  #2: "unit_tests:Just check everything"
  Test  #3: "unit_tests:Task Object Lifecycle"
  Test  #4: "unit_tests:Can create a vector-valued face-variable"
  Test  #5: "unit_tests:Add and Get is called"
  Test  #6: "unit_tests:reset is called"
  Test  #7: "unit_tests:when hasKey is called"
  Test  #8: "unit_tests:Physical constants"
  Test  #9: "unit_tests:Checking IndexShape indices"
  Test #10: "unit_tests:Checking IndexShape cell counts"
  Test #11: "unit_tests:par_for loops"
  Test #12: "unit_tests:nested par_for loops"
  Test #13: "unit_tests:Overlapping SpaceInstances"
  Test #14: "unit_tests:Device Object Allocation"
  Test #15: "unit_tests:Built-in flags are registered"
  Test #16: "unit_tests:A Metadata flag is allocated"
  Test #17: "unit_tests:A Metadata struct is created"
  Test #18: "unit_tests:Metadata created with a sparse ID must be sparse"
  Test #19: "unit_tests:ParArrayND"
  Test #20: "unit_tests:ParArrayND with LayoutLeft"
  Test #21: "unit_tests:Time simple stencil operations"
  Test #22: "unit_tests:Check registry pressure"
  Test #23: "unit_tests:Check many arrays"
  Test #24: "unit_tests:Can pull variables from containers based on Metadata"
  Test #25: "unit_tests:Coarse variable from meshblock_data for cell variable"
  Test #26: "unit_tests:Swarm memory management"
  Test #27: "unit_tests:Test required/desired checking from inputs"
  Test #28: "unit_tests:Parthenon Error Checking"
  Test #29: "unit_tests:Check that integer division, rounded up, works"
  Test #30: "unit_tests:Check that partitioning a container works"
  Test #31: "unit_tests:Test Add/Get in Packages_t"
  Test #32: "unit_tests:Test Associate in StateDescriptor"
  Test #33: "unit_tests:Test reqendency resolution in StateDescriptor"
  Test #34: "performance_tests:Catch2 Container Iterator Performance"
  Test #35: regression_test:advection_performance
  Test #36: regression_mpi_test:advection_performance
  Test #37: regression_test:restart
  Test #38: regression_mpi_test:restart
  Test #39: regression_test:calculate_pi
  Test #40: regression_mpi_test:calculate_pi
  Test #41: regression_test:advection_convergence
  Test #42: regression_mpi_test:advection_convergence
  Test #43: regression_test:output_hdf5
  Test #44: regression_mpi_test:output_hdf5
  Test #45: regression_test:advection_outflow
  Test #46: regression_mpi_test:advection_outflow
  Test #47: regression_test:convergence
  Test #48: regression_test:performance

Total Tests: 48
```
where the last two tests are downstream tests.
The only important piece is that the binary dir is not changed before Parthenon is included/added (which is originally did and resulted in wrong paths as all the Parthenon example binaries also ended up in my binary dir and not in the test directories).
I added an additional note on that in the README.
Feel free to merge when you think that everything is addressed now.

pgrete:
I'm interpreting the thumbsup as acknowledgement and hit merge now.

PR #467, opened 2021-03-04
[WIP] Make parthenon input class methods DRY

JoshuaSBrown:
## PR Summary

This pr is incomplete, submitting to test triggering of CI. 

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted


Yurlungur:
This is a good idea... but will be API breaking. Perhaps we could leave the old API in place also, but it just calls the templated version?

JoshuaSBrown:
> This is a good idea... but will be API breaking. Perhaps we could leave the old API in place also, but it just calls the templated version?

That's reasonable, what about if I added deprecated to the old methods to encourage switching over without breaking. The template api will be cleaner.

Yurlungur:
> what about if I added deprecated to the old methods to encourage switching over without breaking

What did you have in mind? A comment above each method saying its deprecated makes sense to me. I would advise against adding a runtime warning, as that will make simulation output difficult to read.

JoshuaSBrown:
> > what about if I added deprecated to the old methods to encourage switching over without breaking
> 
> What did you have in mind? A comment above each method saying its deprecated makes sense to me. I would advise against adding a runtime warning, as that will make simulation output difficult to read.

No, compile time 
```
[[deprecated]]
```
https://en.cppreference.com/w/cpp/language/attributes/deprecated

Yurlungur:
> > > what about if I added deprecated to the old methods to encourage switching over without breaking
> > 
> > 
> > What did you have in mind? A comment above each method saying its deprecated makes sense to me. I would advise against adding a runtime warning, as that will make simulation output difficult to read.
> 
> No, compile time
> 
> ```
> [[deprecated]]
> ```
> 
> https://en.cppreference.com/w/cpp/language/attributes/deprecated

Oh cool. I wasn't aware of that feature. Yep that makes sense to me.

jlippuner:
Will revisit if needed

PR #475, opened 2021-03-18
Support low storage, two stage integrators

pgrete:
## PR Summary

Add supports for the low storage, two stage integrator implemented in Athena++.
Specifically, two `gamma` scalars are required and a new function that updates two containers and the flux divergence in one call.
I'm not really sure what to do about documentation and testing.

For testing: the integration in AthenaPK builds on this, see [here](https://gitlab.com/theias/hpc/jmstone/athena-parthenon/athenapk/-/jobs/artifacts/main/file/build-cuda/tst/regression/outputs/convergence/linearwave-errors.png?job=cuda-regression) so this works, but to test this in Parthenon we'd need to modify an existing example (removing the current more flexible [but potentially untested] integration scheme) or add a new one.

For documentation: Currently we don't really have any documentation on the functions in `update.cpp` nor the integrators, which is why I didn't add anything specific. At some point we may want to add a more extended documentation.

Anyway, I'm open for suggestion (with respect to the requires for this PR to get merged).

For documentation, 
## PR Checklist

- [x] Code passes cpplint
- [?] New features are documented.
- [?] Adds a test for any bugs fixed. Adds tests for new features.
- [x] Code is formatted


JoshuaSBrown:
Since I am a little removed from the Physics use cases if you can make me understand and then put that into documentation I would be happy. At the moment I don't really get how this part of the code is being used.

I'm guessing the update functions are helper methods that can be used to change variables in different ways. Having a code snippet illustrating a use case where this method would be called would be nice. I haven't read the Athena++ paper methods section so maybe I should do that first. I guess what I could use some clarification on is how general are these update methods?

pgrete:
> Since I am a little removed from the Physics use cases if you can make me understand and then put that into documentation I would be happy. At the moment I don't really get how this part of the code is being used.
> 
> I'm guessing the update functions are helper methods that can be used to change variables in different ways. Having a code snippet illustrating a use case where this method would be called would be nice. I haven't read the Athena++ paper methods section so maybe I should do that first. I guess what I could use some clarification on is how general are these update methods?

I guess that's the big question here because currently effectively no "update-related" function (neither existing nor new) is documented (apart from "look at how we use it in the example").
This is why I only added a brief doc to the function declaration.

In principle, the functions are generic (in the sense that they are calculating weighted average of variables in containers) and the weight themselves are specific to a certain type of integrator (which again are kind of generic in the sense of just providing a temporal integration scheme that is not tied to specifics physics).

How should we proceed? What should be added/changed to get this merged?

Yurlungur (approved):
Sorry it took me so long to get to this. This all seems fine to me. 

I don't really want to ask you to introduce documentation for integrators just for this... although at some point we should really have some. 

As far as tests, perhaps we could modify the advection example to use one of these integrators optionally? But I'm going to say this looks simple enough I won't object to just getting this in without tests, if it's being used downstream in Athena.

pgrete:
Sounds good.
For reference (and posterity), here's the convergence plot of AthenaPK 
image
making use of those functions: 
```c++
    auto update = tl.AddTask(
        none, parthenon::Update::UpdateWithFluxDivergence<MeshData<Real>>, mu0.get(),
        mu1.get(), integrator->gam0[stage - 1], integrator->gam1[stage - 1],
        integrator->beta[stage - 1] * integrator->dt);
```
https://gitlab.com/theias/hpc/jmstone/athena-parthenon/athenapk/-/blob/c0a822fc498598c53efe8f481a872db26369718f/src/hydro/hydro_driver.cpp#L115

With respect to the advection example, the "problem" is that this kind of integrator would change to logic of creating registers (`MeshBlockData`) for stages in the task list.
The current example is more flexible (but also uses more memory as potentially unused `MeshBlockData` objects are created in order to allow for the more general separate averaging and update tasks).
An updated example would only create one additional `MeshBlockData` object that is reused across all stages (and specific to the given implementation of the integrator).
If people think it's worth having a potentially simpler, separate example I'm happy to implement it. It'll be very straightforward.
Too bad we cannot put a simple hydro example in here ;)

Yurlungur:
Ah, I see. A simple example would be nice to have, but I would consider waiting for it non-blocking.

pgrete:
@JoshuaSBrown anything I should address here before this could be merged?

JoshuaSBrown (approved):
Sorry for the delay.

PR #476, opened 2021-03-23
Prepare for C++17 and sync CMake min version

### LINKED ISSUE #544: Failing CI Pipeline looks like cmake version needs to be updated
https://gitlab.com/theias/hpc/jmstone/athena-parthenon/parthenon-ci-mirror/-/jobs/1349331713

pgrete:
## PR Summary

Prepares for bumping C++ version to 17 and sync minimum CMake version required with the Kokkos submodule.
To use C++17 only the `CMAKE_CXX_STANDARD` variable and the `target_compile_feature` of Parthenon would need to be updated.

C++17 would allow us to use `constexpr if`s inside kernels to reduce the kernel size (e.g., for kernels where lots of inline functions are compiled in by default but in each invocation only a subset of those is actually called).

~~**Note**, to compile/run with C++17 on Cuda either~~
- ~~compile with clang (works as is)~~
- ~~compile with `nvcc` and Cuda >= 11.2 (for Cuda <11.1 `Kokkos` does not compile as we require the experimental "relaxed constexpr" feature of nvcc, which is buggy in earlier `nvcc` versions in combination with C++17, see https://github.com/kokkos/kokkos/issues/3496 )~~

Update: This now works with Cuda 11 and nvcc. The problem was that only Kokkos didn't compile with "relaxed constexpr".
I added a corresponding note to our `CMakeLists.txt`

This also disabled OpenMP host parallelism when Cuda is used with nvcc.
For some unknown reason having `-fopenmp` results in the code not compiling any more with many errors like
```
/usr/include/c++/10.2.0/bits/stl_pair.h(460): error: argument list for class template "std::pair" is missing/usr/include/c++/10.2.0/bits/stl_pair.h(460): error: expected a ")"/usr/include/c++/10.2.0/bits/stl_pair.h(460): error: template parameter "_T1" may not be redeclared in this scope
...
/usr/include/c++/10.2.0/bits/unordered_map.h(1193): error: expected a ")"Error limit reached.
100 errors detected in the compilation of "/home/pgrete/src/parthenon/src/interface/sparse_variable.cpp".
```

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


CMakeLists.txt
```diff
@@ -125,13 +125,19 @@ endif()
...
 if (NOT PARTHENON_DISABLE_OPENMP)
-  find_package(OpenMP COMPONENTS CXX)
-  if (NOT OpenMP_FOUND)
-    message(FATAL_ERROR "OpenMP is required but couldn't be found. "
-    "If you want to build Parthenon without OpenMP, please rerun CMake with -DPARTHENON_DISABLE_OPENMP=ON")
+  # Using Host OpenMP and Cuda currently does not compile with C++17 standard.
+  # We may need to revisit this logic when OpenMP target for devices will be used in the future.
+  if (Kokkos_ENABLE_CUDA)
+    message(STATUS "Parthenon is not using Host OpenMP because Cuda is enabled")
+  else()
```
Yurlungur, line 137: :+1:

src/CMakeLists.txt
```diff
@@ -259,7 +259,14 @@ if (ENABLE_HDF5)
...
 
-if (Kokkos_ENABLE_CUDA)
+# For Cuda with NVCC (<11.2) and C++17 Kokkos currently does not work/compile with
+# relaxed-constexpr, see https://github.com/kokkos/kokkos/issues/3496
+# However, Parthenon heavily relies on it and there is no harm in compiling Kokkos
+# without and Parthenon with (via Max Katz on the Kokkos Slack channel).
+# Therefore, we don't use the Kokkos_ENABLE_CUDA_CONSTEXPR option add the flag manually.
+# Also, not checking for NVIDIA as nvcc_wrapper is identified as GNU so we just make sure
+# the flag is not added when compiling with Clang for Cuda.
+if (Kokkos_ENABLE_CUDA AND NOT CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
```
Yurlungur, line 253: :+1:

Yurlungur (approved):
Assuming everything works on all target systems, no objections. But we do need to ensure that.

.gitlab-ci-ias.yml
```diff
@@ -39,7 +39,7 @@ prep-cuda:
     - gpu
     - springdale7
   script:
-    - module load cudatoolkit/10.2 gcc/7.3.0
+    - module load gcc/7.3.0 cudatoolkit/11.1
```
jlippuner: GCC 7.3.0 is very old... How hard would it be to update this?

jlippuner (approved):
Can't wait to use `if constexpr` :D

I agree we need to make sure it works on our target machines. The CI seems to have failed because of a 3-hr timeout. If that's not a fluke and the tests now take more than 3 hrs, we need to understand why.

Yurlungur:
> I agree we need to make sure it works on our target machines. The CI seems to have failed because of a 3-hr timeout. If that's not a fluke and the tests now take more than 3 hrs, we need to understand why.

@brryan mentioned that the machines at LLNL may not support C++17. We should double check that before proceeding.

brryan:
> > I agree we need to make sure it works on our target machines. The CI seems to have failed because of a 3-hr timeout. If that's not a fluke and the tests now take more than 3 hrs, we need to understand why.
> 
> @brryan mentioned that the machines at LLNL may not support C++17. We should double check that before proceeding.

Retaining the ability to run in a C++14 mode, maybe through a compile time flag, might be advantageous for some considerable time. That might mostly prohibit the use of C++17 features within parthenon itself, though.

.gitlab-ci-ias.yml
pgrete: This machine is beyond my control. Nevertheless, this is actually less of a problem as nvcc itself doesn't support the most recent GCC versions anyway.

pgrete:
> > > I agree we need to make sure it works on our target machines. The CI seems to have failed because of a 3-hr timeout. If that's not a fluke and the tests now take more than 3 hrs, we need to understand why.
> > 
> > 
> > @brryan mentioned that the machines at LLNL may not support C++17. We should double check that before proceeding.
> 
> Retaining the ability to run in a C++14 mode, maybe through a compile time flag, might be advantageous for some considerable time. That might mostly prohibit the use of C++17 features within parthenon itself, though.

Yes and no. While downstream codes can use C++17 features with Parthenon being at C++14, this does not apply to kernels (i.e., using `constexpr if` within a kernel as the wrapper lives in Parthenon.

pgrete:
Quite generally, I revisited the changes in this PR and think that they should go in independent of C++17 support (i.e., the CMake update, the logic for Cuda <-> host OpenMP, and CI update).
Therefore, I reverted setting the standard back to 14, but kept all the other changes so that using C++17 in a downstream code requires only two lines of changes in the Parthenon source of that project.
When can then revisit later to actually make the update (in a separate PR) when all machine we are about properly support C++17.
Does this sound like a plan?

brryan:
> Does this sound like a plan?

That's a good point about the kernel wrappers. 

Would it be OK to include something like
```cmake
option(PARTHENON_ENABLE_CPP17 "Enable C++17" OFF)
#...
set(CMAKE_CXX_STANDARD_REQUIRED ON)
if (NOT PARTHENON_ENABLE_CPP17)
  set(CMAKE_CXX_STANDARD 14)
else()
  set(CMAKE_CXX17_STANDARD_COMPILE_OPTION "-std=c++17")
endif()
```
so C++17 can be enabled without modifying the source? This might make experimenting with C++17 downstream a bit easier with respect to version control/submodules. 

(I don't quite understand `CMAKE_CXX_STANDARD` versus `CMAKE_CXX17_STANDARD_COMPILE_OPTION` so the above may not be quite right)

Yurlungur:
> Does this sound like a plan?

This sounds great to me.

> so C++17 can be enabled without modifying the source? This might make experimenting with C++17 downstream a bit easier with respect to version control/submodules.

Is this necessary? I think C++17 should automatically be supported for in-tree builds if the downstream code sets the standard before including the parthenon cmake. Is that how you're doing it for athena-pk, @pgrete ?

On the other hand, it might be nice to try experimental features of C++ in parthenon proper, in which case @brryan 's suggestion might be nice.

pgrete:
I'll introduce a variable as @brryan suggested.
Originally, I tried setting it at a higher level (i.e, in AthenaPKs CMakeLists.txt) but it didn't work (for Parthenon and Kokkos) as the the standard `set()` overrides it at the subproject level.
In principle, this could be solved by making `CMAKE_CXX_STANDARD` a `CACHE` variable, which (as far as I read) is not recommended. In the end there's a reason why a standard is defined at the per (sub)project level (so that the correct standard is being is for *this* project) so making is a free floating cache variables that potentially controls  the behavior of multiple projects simultaneously could cause unintended problems.
In principle, we could use more fine grained control by specifically adding the cxx features to the Parthenon target, but I think that's likely overkill.

pgrete:
Added a `PARTHENON_ENABLE_CPP17` option (default off).
Should be ready for prime time now.

Yurlungur (approved):

Yurlungur:
In that case, let's merge at your convenience, @pgrete !

pgrete:
@par-hermes format

pgrete:
(Sorry for the noise with the force pushes. For some reason both CI and check formatting were reporting "waiting for status" despite being run successful. Not sure what's going on but potentially also affecting #519 )

JoshuaSBrown:
> (Sorry for the noise with the force pushes. For some reason both CI and check formatting were reporting "waiting for status" despite being run successful. Not sure what's going on but potentially also affecting #519 )

I've noticed that as well, I'm wondering if that will go away when a new pr is created. If not I will need to look into it to see what's going on. That is check is showing Andrew's implementation instead of the newer formatting.

pgrete:
Somehow, the check now showed as passed and I was able to manually pull the trigger.
I consider the problem solved for now.

JoshuaSBrown:
> Somehow, the check now showed as passed and I was able to manually pull the trigger.
> I consider the problem solved for now.

Ya, I found the problem, it was related to protected branch permissions settings.

PR #478, opened 2021-03-25
Extend HDF5 output for yt

forrestglines:
## PR Summary

Adds some meta data and AMR level information to the HDF5 outputs so that the python analysis library `yt` can read it. Development of the companion `yt` frontend is at https://github.com/forrestglines/yt/tree/athenapk-frontend

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted

### DIFF
--- a/docs/outputs.md
+++ b/docs/outputs.md
@@ -59,3 +59,114 @@ which will produce a ```png``` image per dump suitable for encoding into a movie
 ## Visualization software
 
 Both [ParaView](https://www.paraview.org/) and [VisIt](https://wci.llnl.gov/simulation/computer-codes/visit/) are capable of opening and visualizing Parthenon graphics dumps.  In both cases, the ```.xdmf``` files should be opened.  In ParaView, select the "XDMF Reader" when prompted.
+
+## Preparing outputs for ```yt```
+
+Parthenon HDF5 outputs can be read with the python visualization library
+[yt](https://yt-project.org/) as certain variables are named when adding
+fields via ```StateDescriptor::AddField```. Variable names are added as a
+```std::vector<std::string>``` in the variable metadata. These labels are
+optional and are only used for output to HDF5. 4D variables are named with a
+list of names for each row while 3D variables are named with a single name.
+For example, the following configurations are acceptable:
+
+```c++
+auto pkg = std::make_shared<StateDescriptor>("Hydro");
+
+/* ... */
+const int nhydro = 5;
+std::vector<std::string> cons_labels(nhydro);
+cons_labels[0]="Density";
+cons_labels[1]="MomentumDensity1";
+cons_labels[2]="MomentumDensity2";
+cons_labels[3]="MomentumDensity3";
+cons_labels[4]="TotalEnergyDensity";
+Metadata m({Metadata::Cell, Metadata::Independent, Metadata::FillGhost},
+           std::vector<int>({nhydro}), cons_labels);
+pkg->AddField("cons", m);
+
+const int ndensity = 1;
+std::vector<std::string> density_labels(ndensity);
+density_labels[0]="Density";
+m = Metadata({Metadata::Cell, Metadata::Derived}, std::vector<int>({ndensity}), density_labels);
+pkg->AddField("dens", m);
+
+const int nvelocity = 3;
+std::vector<std::string> velocity_labels(nvelocity);
+velocity_labels[0]="Velocity1";
+velocity_labels[1]="Velocity2";
+velocity_labels[2]="Velocity3";
+m = Metadata({Metadata::Cell, Metadata::Derived}, std::vector<int>({nvelocity}), velocity_labels);
+pkg->AddField("vel", m);
+
+const int npressure = 1;
+std::vector<std::string> pressure_labels(npressure);
+pressure_labels[0]="Pressure";
+m = Metadata({Metadata::Cell, Metadata::Derived}, std::vector<int>({npressure}), pressure_labels);
+pkg->AddField("pres", m);
+```
+
+The `yt` frontend needs either the hydrodynamic conserved variables or
+primitive compute derived quantities. The conserved variables must have the
+names ```"Density"```, ```"MomentumDensity1"```, ```"MomentumDensity2"```,
+```"MomentumDensity3"```, ```"TotalEnergyDensity"``` while the primitive
+variables must have the names ```"Density"```, ```"Velocity1"```,
+```"Velocity2"```, ```"Velocity3"```, ```"Pressure"```. Either of these sets of
+variables must be named and present in the output, with the primitive variables
+taking precedence over the conserved variables when computing derived
+quantities such as specific thermal energy. In the above example, including
+either ```"cons"``` or ```"dens"```, ```"vel"```, and ```"pres"``` in the  HDF5
+output would allow ```yt``` to read the data.
+
+Additional parameters can also be packaged into the HDF5 file to help ```yt```
+interpret the data, namely adiabatic index and code unit information. These are
+identified by passing ```true``` as an optional boolean argument when adding
+parameters via ```StateDescriptor::AddParam```. For example, 
+```c++
+pkg->AddParam<double>("CodeLength", 100,true);
+pkg->AddParam<double>("CodeMass", 1000,true);
+pkg->AddParam<double>("CodeTime", 1,true);
+pkg->AddParam<double>("AdibaticIndex", 5./3.,true);
+
+pkg->AddParam<int>("IntParam", 0,true);
+pkg->AddParam<std::string>("EquationOfState", "Adiabatic",true);
+```
+adds the parameters ```CodeLength```, ```CodeMass```, ```CodeTime```,
+```AdiabaticIndex```, ```IntParam```, and ```EquationOfState``` to the HDF5
+output. Currently, only ```int```, ```float```, and ```std::string```
+parameters can be included with the HDF5.
+
+Code units can be defined for ```yt``` by including the parameters
+```CodeLength```, ```CodeMass```, and ```CodeTime```, which specify the code
+units used by Parthenon in terms of centimeters, grams, and seconds by writing
+the parameters.  In the above example, these parameters dictate ```yt``` to
+interpret code lengths in the data in units of 100 centimeters (or 1 meter per
+code unit), code masses in units of 1000 grams (or 1 kilogram per code units)
+and code times in units of seconds (or 1 second per code time).
+Alternatively, this unit information can also be supplied to the ```yt```
+frontend when loading the data. If code units are not defined in the HDF5 file
+or at load time, ```yt``` will assume that the data is in ```CGS```.
+
+The adiabatic index can also be specified via the parameter
+```AdiabaticIndex```, defined at load time for ```yt```, or left as its default
+```5./3.```.
+
+For example, the following methods are valid to load data with ```yt```
+```python
+filename = "parthenon.out0.00000.phdf"
+
+#Read units and adiabatic index from the HDF5 file or use defaults
+ds = yt.load(filename)
+
+#Specify units and adiabatic index explicitly
+units_override = {"length_unit" : (100, "cm"),
+                  "time_unit"   : (1,   "s"),
+                  "mass_unit"   : (1000,"g")}
+
+ds = yt.load(filename,units_override=units_override,gamma=5./3.)
+```
+
+Currently, the ```yt``` frontend for Parthenon is hosted on the
+```athenapk-frontend``` [on this ```yt```
+fork](https://github.com/forrestglines/yt/tree/athenapk-frontend). In the
+future, the Parthenon frontend will be included in the main ```yt``` repo.
--- a/scripts/python/phdf.py
+++ b/scripts/python/phdf.py
@@ -108,27 +108,28 @@ class phdf:
             self.CellsPerBlock = np.prod(self.MeshBlockSize)
             self.TotalCells = self.NumBlocks * self.CellsPerBlock
 
+            #Read in Params
+            self.Params = dict(f["Params"].attrs)
+
             # Read in coordinates
-            tmpx = f['/Locations/x'][:,:]
-            self.x = np.zeros((self.NumBlocks,self.MeshBlockSize[0]))
-            for bId in range(self.NumBlocks):
-                for cId in range(self.MeshBlockSize[0]):
-                    self.x[bId,cId] = 0.5*(tmpx[bId,cId]+tmpx[bId,cId+1])
-            self.xf = tmpx
+            def load_coord(coord_i):
+                coord_name = ["x","y","z"][coord_i]
 
-            tmpy = f['/Locations/y'][:,:]
-            self.y = np.zeros((self.NumBlocks,self.MeshBlockSize[1]))
-            for bId in range(self.NumBlocks):
-                for cId in range(self.MeshBlockSize[1]):
-                    self.y[bId,cId] = 0.5*(tmpy[bId,cId]+tmpy[bId,cId+1])
-            self.yf = tmpy
+                tmp = f['/Locations/' + coord_name][:,:]
+                vol_loc = '/VolumeLocations/' + coord_name
+                if vol_loc in f:
+                    coord = f[vol_loc][:,:]
+                else:
+                    coord = np.zeros((self.NumBlocks,self.MeshBlockSize[coord_i]))
+                    for bId in range(self.NumBlocks):
+                        for cId in range(self.MeshBlockSize[coord_i]):
+                            coord[bId,cId] = 0.5*(tmp[bId,cId]+tmp[bId,cId+1])
+                coordf=tmp
+                return tmp,coord,coordf
 
-            tmpz = f['/Locations/z'][:,:]
-            self.z = np.zeros((self.NumBlocks,self.MeshBlockSize[2]))
-            for bId in range(self.NumBlocks):
-                for cId in range(self.MeshBlockSize[2]):
-                    self.z[bId,cId] = 0.5*(tmpz[bId,cId]+tmpz[bId,cId+1])
-            self.zf = tmpz
+            tmpx, self.x, self.xf = load_coord(0)
+            tmpy, self.y, self.yf = load_coord(1)
+            tmpz, self.z, self.zf = load_coord(2)
 
             # fill in self.offset and block bounds
             self.offset = [0,0,0]
@@ -151,6 +152,8 @@ class phdf:
                     tmpy[ib,iOffsets[2]]-eps, tmpy[ib,iOffsets[3]]+eps,
                     tmpz[ib,iOffsets[4]]-eps, tmpz[ib,iOffsets[5]]+eps
                     ]
+            #Save info 
+            self.Info = dict(f["/Info"].attrs)
                     
             # generate self.offset, isGhost and BlockIdx arrays
             self.GenAuxData()
@@ -161,6 +164,7 @@ class phdf:
 
             self.Variables = [k for k in f.keys()]
             self.varData = {k:None for k in self.Variables}
+
         except:
             raise FileNotFoundError(
                 errno.ENOENT, os.strerror(errno.ENOENT), filename)
@@ -291,6 +295,25 @@ class phdf:
 
         raise ValueError("Unable to map cell")
 
+    def findBlockIdxInOther(self,other,ib,tol=1e-10,verbose=False):
+        """
+        Given an meshblock index in my data, find the meshblock index in a different file.
+        """
+
+        myibBounds = np.array(self.BlockBounds[ib])
+
+        # now hunt in other file
+        for ib1 in range(other.NumBlocks):
+            ib1Bounds = np.array(other.BlockBounds[ib1])
+
+            if( np.all( np.abs(myibBounds - ib1Bounds) < tol)):
+                return ib1
+
+        if verbose:
+            print(f"Block id: {ib} with bounds {myibBounds} not found in {other.file}")
+        return None #block index not found
+
+
     def Get(self, variable, flatten=True):
         """
         Reads data for the named variable from file.
@@ -386,4 +409,3 @@ if __name__ == "__main__":
         l = ba.Get("c.c.bulk.bulk_modulus")
         print('mod=',l.shape)
         print(help(ba))
-
--- a/scripts/python/phdf_diff.py
+++ b/scripts/python/phdf_diff.py
@@ -26,15 +26,16 @@ import argparse
 def Usage():
     print("""
 
-    Usage: %s [-quiet] [-brief] [-all] [-one] [--tol=eps] file1.phdf file2.phdf
+    Usage: %s [-quiet] [-brief] [-all] [-one] [--tol=eps] [-ignore_metadata] file1.phdf file2.phdf
 
-           -all: report all diffs at all positions
-           -one: Quit after first different variable
-         -brief: Only report if files are different
-                 Overrides --all
-         -quiet: Only report if files are different and
-                 don't print any extraneous info.
-      --tol=eps: set tolerance to eps.  Default 1.0e-12
+                  -all: report all diffs at all positions
+                  -one: Quit after first different variable
+                -brief: Only report if files are different
+                        Overrides --all
+                -quiet: Only report if files are different and
+                        don't print any extraneous info.
+             --tol=eps: set tolerance to eps.  Default 1.0e-12
+      -ignore_metadata: Ignore differences in metadata
 
     This example takes two hdf files and compares them to see if there are
     differences in the state variables.
@@ -53,6 +54,7 @@ def processArgs():
     parser.add_argument('-o', '-one', action='store_true', help='Only report data for first different variable.')
     parser.add_argument('-b', '-brief', action='store_true', help='Only report if files are different.  Overrides -all')
     parser.add_argument('-q', '-quiet', action='store_true', help='Only report if files are different.  No other output. Overrides -all')
+    parser.add_argument('-i', '-ignore_metadata', action='store_true', help='Ignore differences in metadata. Overrides -all')
     parser.add_argument('files', nargs='*')
 
     return parser.parse_args()
@@ -64,8 +66,162 @@ def addPath():
     #sys.path.insert(0,myPath+'/../vis/python')
     #sys.path.insert(0,myPath+'/vis/python')
 
-def compare(files, all=False, brief=True, quiet=False, one=False, tol=1.0e-12):
-    """ compares two hdf files """
+def compare_metadata( f0, f1, tol=1.0e-12):
+    """ compares metadata of two hdf files f0 and f1. Returns 0 if the files are equivalent.
+
+        Error codes:
+            10 : Times in hdf files differ
+            11 : Attribute names in Info of hdf files differ
+            12 : Values of attributes in Info of hdf files differ
+            13 : Attribute names in Params of hdf files differ
+            14 : Values of attributes in Params of hdf files differ
+            15 : Meta variables (Locations, VolumeLocations, LogicalLocations, Levels) differ
+    """
+    ERROR_TIME_DIFF=10
+    ERROR_INFO_ATTRS_DIFF=11
+    ERROR_INFO_VALUES_DIFF=12
+    ERROR_PARAMS_ATTR_DIFF=13
+    ERROR_PARAMS_VALUES_DIFF=14
+    ERROR_META_VARS_DIFF=15
+
+
+    #Compare the time in both files
+    errTime = np.abs(f0.Time- f1.Time)
+    if errTime > tol:
+        print(f"""
+        Time of outputs differ by {f0.Time - f1.Time}
+
+        Quitting...
+        """)
+        return(ERROR_TIME_DIFF)
+
+    #Compare the names of attributes in /Info, except "Time"
+    f0_Info = { key:value for key,value in f0.Info.items() if key != "Time" and key != "BlocksPerPE" }
+    f1_Info = { key:value for key,value in f1.Info.items() if key != "Time" and key != "BlocksPerPE" }
+    if sorted(f0_Info.keys()) != sorted(f1_Info.keys()):
+        print("""
+        Names of attributes in '/Info' of differ
+
+        Quitting...
+        """)
+        return(ERROR_INFO_ATTRS_DIFF)
+
+    #Compare the values of attributes in /Info
+    info_diffs = list( key for key in f0_Info.keys() if np.any(f0_Info[key] != f1_Info[key]))
+    if len(info_diffs) > 0:
+        print("\nValues of attributes in '/Info' differ\n")
+        print("Differing attributes: ", info_diffs )
+        print("\nQuitting...\n")
+        return(ERROR_INFO_VALUES_DIFF)
+    else:
+        print('  %20s: no diffs'%"Info")
+
+    f0_Params = f0.Params
+    f1_Params = f1.Params
+    #Comparing all params at once might work except for float point Params
+    if sorted(f0_Params.keys()) != sorted(f1_Params.keys()):
+        print("""
+        Names of attributes in '/Params' of differ
+
+        Quitting...
+        """)
+        return(ERROR_PARAMS_ATTRS_DIFF)
+
+    #Check that the values of non-floats in Params match
+    params_nonfloats_diffs = list(key for key in f0_Params.keys()
+            if not isinstance(f0_Params[key], float) and f0_Params[key] != f1_Params[key] )
+
+
+    #Check that the values of floats Params match
+    params_floats_diffs = list(key for key in f0_Params.keys()
+            if isinstance(f0_Params[key], float) and np.abs(f0_Params[key] - f1_Params[key]) > tol )
+
+    if len(params_nonfloats_diffs) > 0 or len(params_floats_diffs) > 0:
+        if len(params_nonfloats_diffs) > 0:
+            print("\nValues of non-float attributes in '/Params' differ\n")
+            print("Differing attributes: ",params_nonfloats_diffs  )
+            print("\nQuitting...\n")
+        elif len(params_floats_diffs) > 0:
+            print("\nValues of float attributes in '/Params' differ\n")
+            for key in params_floats_diffs:
+                print(f"Param {key} differs by {f0_Params[key] - f1_Params[key]} ")
+            print("\nQuitting...\n")
+        return(ERROR_PARAMS_VALUES_DIFF)
+    else:
+        print('  %20s: no diffs'%"Params")
+
+    # Now go through all variables in first file
+    # and hunt for them in second file.
+    #
+    # Note that indices don't match when blocks
+    # are different
+    no_meta_variables_diff = True
+
+    otherBlockIdx = list(f0.findBlockIdxInOther(f1,i) for i in range(f0.NumBlocks))
+
+    for var in set(f0.Variables+f1.Variables):
+        if var in ['Locations', 'VolumeLocations']:
+            for key in f0.fid[var].keys():
+                #Compare raw data of these variables
+                val0 = f0.fid[var][key]
+                val1 = f1.fid[var][key]
+
+                #Sort val1 by otherBlockIdx
+                val1 = val1[otherBlockIdx]
+
+                # Compute norm error, check against tolerance
+                errVal = np.abs(val0 - val1)
+                errMag = np.linalg.norm(errVal)
+                if(errMag > tol):
+                    no_meta_variables_diff = False
+                    if not quiet: print("")
+                    print(f'Metavariable {var}/{key} differs between {f0.file} and {f1.file}')
+                    if not quiet: print("")
+                else:
+                    print('  %18s/%s: no diffs'%(var,key))
+        if var in ['LogicalLocations', 'Levels']:
+            #Compare raw data of these variables
+            val0 = np.array(f0.fid[var])
+            val1 = np.array(f1.fid[var])
+
+            #Sort val1 by otherBlockIdx
+            val1 = val1[otherBlockIdx]
+
+            #As integers, they should be identical
+            if np.any(val0 != val1):
+                no_meta_variables_diff = False
+                if not quiet: print("")
+                print(f'Metavariable {var} differs between {f0.file} and {f1.file}')
+                if not quiet: print("")
+            else:
+                print('  %20s: no diffs'%var)
+
+    if not no_meta_variables_diff:
+        return(ERROR_META_VARS_DIFF)
+    return(0)
+
+def compare(files, all=False, brief=True, quiet=False, one=False, tol=1.0e-12, check_metadata=True):
+    """ compares two hdf files. Returns 0 if the files are equivalent.
+
+        Error codes:
+            1  : Can't open file 0
+            2  : Can't open file 1
+            3  : Total number of cells differ
+            4  : Variable data in files differ
+
+        Metadata Error codes:
+            10 : Times in hdf files differ
+            11 : Attribute names in Info of hdf files differ
+            12 : Values of attributes in Info of hdf files differ
+            13 : Attribute names in Params of hdf files differ
+            14 : Values of attributes in Params of hdf files differ
+            15 : Meta variables (Locations, VolumeLocations, LogicalLocations, Levels) differ
+    """
+
+    ERROR_NO_OPEN_F0 = 1
+    ERROR_NO_OPEN_F1 = 2
+    ERROR_CELLS_DIFFER = 3
+    ERROR_DATA_DIFFER = 4
 
     #**************
     # import Reader
@@ -80,6 +236,7 @@ def compare(files, all=False, brief=True, quiet=False, one=False, tol=1.0e-12):
 
 
     # Load first file and print info
+    f0 = phdf(files[0])
     try:
         f0 = phdf(files[0])
         if not quiet: print(f0)
@@ -87,7 +244,7 @@ def compare(files, all=False, brief=True, quiet=False, one=False, tol=1.0e-12):
         print("""
         *** ERROR: Unable to open %s as phdf file
         """%files[0])
-        return(1)
+        return(ERROR_NO_OPEN_F0)
 
     # Load second file and print info
     try:
@@ -97,7 +254,7 @@ def compare(files, all=False, brief=True, quiet=False, one=False, tol=1.0e-12):
         print("""
         *** ERROR: Unable to open %s as phdf file
         """%files[1])
-        return(2)
+        return(ERROR_NO_OPEN_F1)
 
     # rudimentary checks
     if f0.TotalCellsReal != f1.TotalCellsReal:
@@ -108,7 +265,15 @@ def compare(files, all=False, brief=True, quiet=False, one=False, tol=1.0e-12):
 
         Quitting...
         """)
-        return(3)
+        return(ERROR_CELLS_DIFFER)
+
+    if check_metadata:
+        if not quiet: print("Checking metadata")
+        metadata_status = compare_metadata(f0,f1)
+        if( metadata_status != 0):
+            return metadata_status
+    else:
+        if not quiet: print("Ignoring metadata")
 
     # Now go through all variables in first file
     # and hunt for them in second file.
@@ -136,9 +301,10 @@ def compare(files, all=False, brief=True, quiet=False, one=False, tol=1.0e-12):
         otherLocations[idx] = f0.findIndexInOther(f1,idx)
     if not quiet: print(f0.TotalCells,'cells mapped')
 
-    for var in f0.Variables:
-        if var == 'Locations' or var == 'Info':
+    for var in set(f0.Variables+f1.Variables):
+        if var in ['Locations', 'VolumeLocations', 'LogicalLocations', 'Levels', 'Info', 'Params']:
             continue
+
         #initialize info values
         same = True
         errMax = -1.0
@@ -203,11 +369,11 @@ def compare(files, all=False, brief=True, quiet=False, one=False, tol=1.0e-12):
 
         if one and not same:
             break
-
+    
     if no_diffs:
       return(0)
     else:
-      return(4)
+      return(ERROR_DATA_DIFFER)
 
 if __name__ == "__main__":
     addPath()
@@ -218,6 +384,9 @@ if __name__ == "__main__":
     brief=input.b
     quiet=input.q
     one = input.o
+    ignore_metadata = input.i
+
+    check_metadata = not ignore_metadata
 
     # set all only if brief not set
     if brief or quiet:
@@ -236,5 +405,5 @@ if __name__ == "__main__":
         Usage()
         sys.exit(1)
 
-    ret = compare(files, all, brief, quiet, one, tol)
+    ret = compare(files, all, brief, quiet, one, tol, check_metadata)
     sys.exit(ret)
--- a/src/interface/metadata.hpp
+++ b/src/interface/metadata.hpp
@@ -181,6 +181,13 @@ class Metadata {
     SetMultiple(bits);
   }
 
+  /// returns a metadata with bits and shape set
+  explicit Metadata(const std::vector<MetadataFlag> &bits, const std::vector<int> &shape,
+                    const std::vector<std::string> component_labels)
+      : shape_(shape), sparse_id_(-1), component_labels_(component_labels) {
+    SetMultiple(bits);
+  }
+
   /// returns a metadata with bits and sparse id set
   explicit Metadata(const std::vector<MetadataFlag> &bits, const int sparse_id)
       : shape_({1}), sparse_id_(sparse_id) {
@@ -346,12 +353,18 @@ class Metadata {
   void Associate(const std::string &name) { associated_ = name; }
   const std::string &getAssociated() const { return associated_; }
 
+  const std::vector<std::string> getComponentLabels() const noexcept {
+    return component_labels_;
+  }
+
  private:
   /// the attribute flags that are set for the class
   std::vector<bool> bits_;
   std::vector<int> shape_;
   std::string associated_;
   int sparse_id_;
+
+  std::vector<std::string> component_labels_;
 
   /*--------------------------------------------------------*/
   // Setters for the different attributes of metadata
--- a/src/interface/params.hpp
+++ b/src/interface/params.hpp
@@ -19,6 +19,7 @@
 #include <string>
 #include <typeindex>
 #include <typeinfo>
+#include <utility>
 #include <vector>
 
 #include "utils/error_checking.hpp"
@@ -41,7 +42,7 @@ class Params {
   void Add(const std::string &key, T value) {
     PARTHENON_REQUIRE_THROWS(!(hasKey(key)), "Key " + key + " already exists");
     myParams_[key] = std::unique_ptr<Params::base_t>(new object_t<T>(value));
-    myTypes_[key] = std::string(typeid(value).name());
+    myTypes_.emplace(make_pair(key, std::type_index(typeid(value))));
   }
 
   /// Updates existing object
@@ -49,7 +50,7 @@ class Params {
   template <typename T>
   void Update(const std::string &key, T value) {
     PARTHENON_REQUIRE_THROWS((hasKey(key)), "Key " + key + "missing.");
-    PARTHENON_REQUIRE_THROWS(!(myTypes_.at(key).compare(std::string(typeid(T).name()))),
+    PARTHENON_REQUIRE_THROWS(myTypes_.at(key) == std::type_index(typeid(T)),
                              "WRONG TYPE FOR KEY '" + key + "'");
     myParams_[key] = std::unique_ptr<Params::base_t>(new object_t<T>(value));
   }
@@ -63,7 +64,7 @@ class Params {
   const T &Get(const std::string &key) const {
     auto const it = myParams_.find(key);
     PARTHENON_REQUIRE_THROWS(it != myParams_.end(), "Key " + key + " doesn't exist");
-    PARTHENON_REQUIRE_THROWS(!(myTypes_.at(key).compare(std::string(typeid(T).name()))),
+    PARTHENON_REQUIRE_THROWS(myTypes_.at(key) == std::type_index(typeid(T)),
                              "WRONG TYPE FOR KEY '" + key + "'");
     auto typed_ptr = dynamic_cast<Params::object_t<T> *>((it->second).get());
     return *typed_ptr->pValue;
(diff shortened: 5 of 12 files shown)


Yurlungur:
@forrestglines will you submit the yt frontend as a PR to yt when it's ready?

src/outputs/parthenon_hdf5.cpp
```diff
@@ -307,7 +307,85 @@ void PHDF5Output::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm) {
...
       const hsize_t xDims[1] = {3};
       writeH5AI32("MeshBlockSize", meshblock_size,
-                  H5S::FromHIDCheck(H5Screate_simple(1, xDims, NULL)), myDSet);
+                  H5S::FromHIDCheck(H5Screate_simple(1, xDims, NULL)), myDSet);  
+    }
+
+    { // TODO(forrestglines) - should this data be confined to yt-Info
+      // Store yt-necessary attributes
+      H5S const localDSpace = H5S::FromHIDCheck(H5Screate(H5S_SCALAR));
+      H5D const ytInfo =
```
Yurlungur: I actually see no reason to tag this as "needed for yt." Who knows, other vis tools might find this output useful.

src/outputs/parthenon_hdf5.cpp
```diff
@@ -307,7 +307,85 @@ void PHDF5Output::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm) {
...
+
+        Real code_length = 1.;
+        writeH5AF64("CodeLength", &code_length, localDSpace, downstreamInfo); 
+
+        Real code_mass = 1.;
+        writeH5AF64("CodeMass", &code_mass, localDSpace, downstreamInfo); 
+
+        Real code_time = 1.;
+        writeH5AF64("CodeTime", &code_time, localDSpace, downstreamInfo); 
+      }
```
Yurlungur: Let's make sure the yt reader is robust against the absence of this stuff

src/outputs/parthenon_hdf5.cpp
forrestglines: After writing the prototype, I agree. It's a bit convoluted to wrap these into a separate dataset and should belong in the `Info` dataset with everything else. I think we should always write this data rather than create two classes of outputs - one that's readable by `yt` (or other codes) and one that's not

forrestglines:
> @forrestglines will you submit the yt frontend as a PR to yt when it's ready?

I'll ask the `yt-dev` mailing list about their standard for inclusion. I would be in favor of adding it.

Yurlungur:
My experience is they're very inclusive.

forrestglines:
Here's my current design for downstream hooks for labeling individual components of variables and including parameters as attributes in the HDF5 file.

To label individual components, I've added an optional `std::vector<std:string> component_labels_` to the `Metadata` class. This vector is used to name components along the first dimension of a variable - `"Density"`,`"MomentumDensity1"`, etc. for AthenaPK. If this vector is not specified, the variable is treated as one component with the same name as the variable label.

 I explored adding this vector to the `StateDescriptor` or to `CellVariable`, but it turned out to affect too many lines of code.
  
To add parameters to the HDF5 file, I added a `std::vector<std::string> outputing_params_` to `StateDescriptor`, which is used as a simple list of parameters already in `Params`.  I also added a optional `bool outputing` to `StateDescriptor::AddParam`, which defaults to `false`, to mark if a parameter should be included with the HDF5 file. Right now, only `float`, `double`, and `std:string` parameters are writable to HDF5.

See [this](https://gitlab.com/theias/hpc/jmstone/athena-parthenon/athenapk/-/blob/forrestglines/hdf5-for-yt/src/hydro/hydro.cpp#L161) for how I add component labels in AthenaPK and [here](https://gitlab.com/theias/hpc/jmstone/athena-parthenon/athenapk/-/blob/forrestglines/hdf5-for-yt/src/hydro/hydro.cpp#L134) for how I add `AdiabaticIndex` and `EquationOfState` as parameters for the HDF5 file.

forrestglines:
Do I need to add to the `xml` in the `xdmf` file? The `xdmf` file is not used within `yt`

Yurlungur:
> Here's my current design for downstream hooks for labeling individual components of variables and including parameters as attributes in the HDF5 file.
> 
> To label individual components, I've added an optional `std::vector<std:string> component_labels_` to the `Metadata` class. This vector is used to name components along the first dimension of a variable - `"Density"`,`"MomentumDensity1"`, etc. for AthenaPK. If this vector is not specified, the variable is treated as one component with the same name as the variable label.
> 
> I explored adding this vector to the `StateDescriptor` or to `CellVariable`, but it turned out to affect too many lines of code.
> 
> To add parameters to the HDF5 file, I added a `std::vector<std::string> outputing_params_` to `StateDescriptor`, which is used as a simple list of parameters already in `Params`. I also added a optional `bool outputing` to `StateDescriptor::AddParam`, which defaults to `false`, to mark if a parameter should be included with the HDF5 file. Right now, only `float`, `double`, and `std:string` parameters are writable to HDF5.
> 
> See [this](https://gitlab.com/theias/hpc/jmstone/athena-parthenon/athenapk/-/blob/forrestglines/hdf5-for-yt/src/hydro/hydro.cpp#L161) for how I add component labels in AthenaPK and [here](https://gitlab.com/theias/hpc/jmstone/athena-parthenon/athenapk/-/blob/forrestglines/hdf5-for-yt/src/hydro/hydro.cpp#L134) for how I add `AdiabaticIndex` and `EquationOfState` as parameters for the HDF5 file.

This seems like a reasonable solution. Can the output name for a variable be different than the code name? i.e., is there a way to translate variable names to the canonical yt names?

> Do I need to add to the xml in the xdmf file? The xdmf file is not used within yt

Hmm... I don't think so... I think this information can be handed to paraview separately... Although a paraview expert like @nmsriram is perhaps better equipped to answer.

forrestglines:
> This seems like a reasonable solution. Can the output name for a variable be different than the code name? i.e., is there a way to translate variable names to the canonical yt names?

Yes, these names are only defined for HDF5 outputs for `yt`. These labels are independent of any variable names already in use.

forrestglines:
@par-hermes format

JoshuaSBrown:
Is this ready for review?

forrestglines:
@JoshuaSBrown It was ready - but then I decided to extend the testing so that the HDF5 comparison checks the metadata of the HDF5 files as well.

That extension should be ready today or tomorrow but we'll need to rebless the gold data as well.

forrestglines:
The extension to HDF5 metadata testing is finished. The python scripts now test that both files contain the same data in `Info`, allowing exceptions like differences in blocks per processing element and the order of mesh blocks in the file.

We do need to rebless the all the HDF5 gold data since metadata has been added to these files. However, this PR is ready for review.

docs/outputs.md
```diff
@@ -48,3 +48,105 @@ which will produce a ```png``` image per dump suitable for encoding into a movie
...
+used for output to HDF5. 4D variables are named with a list of names for each
+row while 3D variables are named with a single name.  For example, the
+following configurations are acceptable:
+
+```c++
+auto pkg = std::make_shared<StateDescriptor>("Hydro");
+
+/* ... */
+
+std::vector<std::string> cons_labels(nhydro);
```
JoshuaSBrown: I would add:
```
const int nhydro = 5;
```

docs/outputs.md
```diff
@@ -48,3 +48,105 @@ which will produce a ```png``` image per dump suitable for encoding into a movie
...
+cons_labels[0]="Density";
+cons_labels[1]="MomentumDensity1";
+cons_labels[2]="MomentumDensity2";
+cons_labels[3]="MomentumDensity3";
+cons_labels[4]="TotalEnergyDensity";
+Metadata m({Metadata::Cell, Metadata::Independent, Metadata::FillGhost},
+           std::vector<int>({5}), cons_labels);
```
JoshuaSBrown: Remove magic number and use I'm guessing ```nhydro```?

docs/outputs.md
```diff
@@ -48,3 +48,105 @@ which will produce a ```png``` image per dump suitable for encoding into a movie
...
+velocity_labels[1]="Velocity2";
+velocity_labels[2]="Velocity3";
+m = Metadata({Metadata::Cell, Metadata::Derived}, std::vector<int>({3}), velocity_labels);
+pkg->AddField("vel", m);
+
+std::vector<std::string> pressure_labels(1);
+pressure_labels[0]="Pressure";
+m = Metadata({Metadata::Cell, Metadata::Derived}, std::vector<int>({1}), pressure_labels);
+pkg->AddField("pres", m);
+```
```
JoshuaSBrown: Maybe I'm being a bit picky, but I would recommend for the purpose of being explicit removing the magic numbers and setting them equal to a value in each of these cases.

docs/outputs.md
```diff
@@ -48,3 +48,105 @@ which will produce a ```png``` image per dump suitable for encoding into a movie
...
+pkg->AddParam<double>("AdibaticIndex", 5./3.,true);
+
+pkg->AddParam<int>("IntParam", 0,true);
+pkg->AddParam<std::string>("EquationOfState", "Adiabatic",true);
+```
+all function to add parameters to the HDF5 output. Currently, only ```int```,
+```float```, and ```std::string``` parameters can be included with the HDF5.
+
+Defining ```CodeLength```, ```CodeMass```, and ```CodeTime```, all in CGS,
+allows ```yt``` to intupret units for the data. The above example declares the
```
JoshuaSBrown: interpret

docs/outputs.md
```diff
@@ -48,3 +48,105 @@ which will produce a ```png``` image per dump suitable for encoding into a movie
...
+data as being in meters, kilograms, and seconds. Alternatively, this unit
+information can be also supplied to the ```yt``` frontend when loading the
```
JoshuaSBrown: can also be

docs/outputs.md
```diff
@@ -48,3 +48,105 @@ which will produce a ```png``` image per dump suitable for encoding into a movie
...
 
 Both [ParaView](https://www.paraview.org/) and [VisIt](https://wci.llnl.gov/simulation/computer-codes/visit/) are capable of opening and visualizing Parthenon graphics dumps.  In both cases, the ```.xdmf``` files should be opened.  In ParaView, select the "XDMF Reader" when prompted.
+
+## Preparing outputs for ```yt```
+
+Parthenon HDF5 outputs can be read with the python visualization library
+[yt](https://yt-project.org/) as certain variables are named during when adding
+fields via ```StateDescriptor::AddField```. Variable names are added as a
+```std::vector<std::string>``` in the variable metadata. These labels are only
+used for output to HDF5. 4D variables are named with a list of names for each
```
Yurlungur: Are 5D variables supported? Are they simply rasterized?

docs/outputs.md
```diff
@@ -48,3 +48,105 @@ which will produce a ```png``` image per dump suitable for encoding into a movie
...
+cons_labels[2]="MomentumDensity2";
+cons_labels[3]="MomentumDensity3";
+cons_labels[4]="TotalEnergyDensity";
+Metadata m({Metadata::Cell, Metadata::Independent, Metadata::FillGhost},
+           std::vector<int>({5}), cons_labels);
+pkg->AddField("cons", m);
+
+std::vector<std::string> density_labels(1);
+density_labels[0]="Density";
+m = Metadata({Metadata::Cell, Metadata::Derived}, std::vector<int>({1}), density_labels);
```
Yurlungur: Are these labels optional? Or are they now required?

docs/outputs.md
```diff
@@ -48,3 +48,105 @@ which will produce a ```png``` image per dump suitable for encoding into a movie
...
+m = Metadata({Metadata::Cell, Metadata::Derived}, std::vector<int>({3}), velocity_labels);
+pkg->AddField("vel", m);
+
+std::vector<std::string> pressure_labels(1);
+pressure_labels[0]="Pressure";
+m = Metadata({Metadata::Cell, Metadata::Derived}, std::vector<int>({1}), pressure_labels);
+pkg->AddField("pres", m);
+```
+
+For the ```yt``` frontend needs either the hydrodynamic conserved variables or
```
Yurlungur:
```suggestion
The `yt` frontend needs either the hydrodynamic conserved variables or
```

src/interface/params.hpp
```diff
@@ -41,15 +42,15 @@ class Params {
   void Add(const std::string &key, T value) {
     PARTHENON_REQUIRE_THROWS(!(hasKey(key)), "Key " + key + " already exists");
     myParams_[key] = std::unique_ptr<Params::base_t>(new object_t<T>(value));
-    myTypes_[key] = std::string(typeid(value).name());
+    myTypes_.insert(make_pair(key, std::type_index(typeid(value))));
```
Yurlungur: This seems like an improvement! But out of curiosity, why did it need to be changed now?

src/interface/state_descriptor.hpp
```diff
@@ -49,8 +49,11 @@ class StateDescriptor {
   explicit StateDescriptor(std::string const &label) : label_(label) {}
 
   template <typename T>
-  void AddParam(const std::string &key, T value) {
+  void AddParam(const std::string &key, T value, const bool outputing = false) {
```
Yurlungur: The name `outputing` feels odd here. Why not just `output`?

Yurlungur (approved):
Great work @forrestglines . Overall looks good. I have a few nitpicks below.

What's the status of the yt reader being in yt mainline?

forrestglines:
> What's the status of the yt reader being in yt mainline?

I have a fork for the `yt` frontend working, but it might change once we start doing analysis and include magnetic fields in our simulations. When we're sure that we have everything in the yt frontend working, I'll push that frontend to the yt mainline.

src/interface/params.hpp
forrestglines: Two reasons:

1. I changed the `myTypes_` map to hold `std::type_index` instead of `std::string` in order to be more sure of some type checking. This change likely wasn't needed for what I was trying to accomplish, but I imagine comparing `std::type_index` is safer than comparing strings.
2. The `[]` operator for maps (I think just for assignment) requires the type to have a default constructor. However, `std::type_index` doesn't have a default constructor since there isn't a default type. Hence using `insert` instead.

src/interface/state_descriptor.hpp
forrestglines: I used `outputing` to match the `outputing_params` vector that I've added. `OutputParams` is an entirely separate object, so I avoided using `output` as an verb here to avoid confusion. Would `include_with_output` be too verbose or should I just use `output`?

Yurlungur: I like `include_with_output`

docs/outputs.md
forrestglines: Currently not supported. The frontend is built around identifying either the conserved or primitive variables in  3D space, either being packed together in one 4D array, split up into individual 3D arrays, or some combination.

I'm not sure how we should interpret 5D data. Should the extra dimension be different species (`yt` as a whole doesn't really deal with multifluid) or should we interpolate them?

docs/outputs.md
forrestglines: They're optional. If they're not specified, the HDF5 output will label the entire field by the field name, i.e. `cons` and `dens` here. I'll update the documentation.

docs/outputs.md
Yurlungur: I was more thinking about, e.g., tensor quantities. i.e., the Reynolds stress for example. But multiple species (which I think we might implement as sparse variables) would be another example.

scripts/python/phdf.py
```diff
@@ -110,24 +110,33 @@ def __init__(self, filename):
...
-            for bId in range(self.NumBlocks):
-                for cId in range(self.MeshBlockSize[0]):
-                    self.x[bId,cId] = 0.5*(tmpx[bId,cId]+tmpx[bId,cId+1])
+            if '/VolumeLocations/x' in f:
+                self.x = f['/VolumeLocations/x'][:,:]
+            else:
+                self.x = np.zeros((self.NumBlocks,self.MeshBlockSize[0]))
+                for bId in range(self.NumBlocks):
+                    for cId in range(self.MeshBlockSize[0]):
+                        self.x[bId,cId] = 0.5*(tmpx[bId,cId]+tmpx[bId,cId+1])
```
JoshuaSBrown: I think you could put this logic into a function and pass in an index, and simply return the the numpy array. This would reduce the repetition below.

scripts/python/phdf_diff.py
```diff
@@ -110,12 +110,43 @@ def compare(files, all=False, brief=True, quiet=False, one=False, tol=1.0e-12):
...
+            """)
+            return(6)
+
+        info_diffs = list(np.any(f0_Info[key] != f1_Info[key]) for key in f0_Info.keys())
+        infos_differ = np.any( info_diffs )
+        if infos_differ:
+            print("\n'/Info' of outputs differ\n")
+            print("Differing attributes: ",list( (key) for key in f0_Info.keys() if np.any(f0_Info[key] != f1_Info[key]))  )
+            print("\nQuitting...\n")
+            return(7)
```
JoshuaSBrown: I would move this logic into a separate method, but that is more preference.

scripts/python/phdf_diff.py
```diff
@@ -110,12 +110,43 @@ def compare(files, all=False, brief=True, quiet=False, one=False, tol=1.0e-12):
...
+        f0_Info = { key:value for key,value in f0.Info.items() if key != "Time" and key != "BlocksPerPE" }
+        f1_Info = { key:value for key,value in f1.Info.items() if key != "Time" and key != "BlocksPerPE" }
+
+        errTime = np.abs(f0.Time- f1.Time)
+        if errTime > tol:
+            print("""
+            Time of outputs differ
+
+            Quitting...
+            """)
```
JoshuaSBrown: Because it is floating point I would recommend actually printing the difference in the error message.

scripts/python/phdf_diff.py
```diff
@@ -136,9 +167,50 @@ def compare(files, all=False, brief=True, quiet=False, one=False, tol=1.0e-12):
...
+                    val1 = val1[otherBlockIdx]
+
+                    #As integers, they should be identical
+                    if val0 != val0:
+                        no_meta_variables_diff = False
+                        if not quiet: print("")
+                        print(f'Metavariable {var} differs between {f0.file} and {f1.file}')
+                        if not quiet: print("")
+                    else:
+                        print('  %20s: no diffs'%var)
```
JoshuaSBrown: Again I would recommend packaging this block of logic into a separate method, ... preference. 
for floating point message I would print the difference.

scripts/python/phdf_diff.py
```diff
@@ -204,6 +276,9 @@ def compare(files, all=False, brief=True, quiet=False, one=False, tol=1.0e-12):
         if one and not same:
             break
 
+    if not no_meta_variables_diff:
+        return(8)
```
JoshuaSBrown: Are the return values defined somewhere in the documentation or am I simply missing them.

src/interface/metadata.hpp
```diff
@@ -346,13 +353,17 @@ class Metadata {
   void Associate(const std::string &name) { associated_ = name; }
   const std::string &getAssociated() const { return associated_; }
 
+  const std::vector<std::string> getComponentLabels() const { return component_labels_; }
```
JoshuaSBrown: This method can be made noexcept.

src/interface/params.hpp
JoshuaSBrown: Could you use emplace instead instead of insert.

src/interface/params.hpp
```diff
@@ -41,15 +42,15 @@ class Params {
...
   }
 
   /// Updates existing object
   /// Throws an error if the key is not already in use
   template <typename T>
   void Update(const std::string &key, T value) {
     PARTHENON_REQUIRE_THROWS((hasKey(key)), "Key " + key + "missing.");
-    PARTHENON_REQUIRE_THROWS(!(myTypes_.at(key).compare(std::string(typeid(T).name()))),
+    PARTHENON_REQUIRE_THROWS(myTypes_.at(key) == std::type_index(typeid(T)),
```
JoshuaSBrown, line 53: This is clearer, nice!

src/outputs/parthenon_hdf5.cpp
```diff
@@ -372,14 +461,99 @@ void PHDF5Output::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm) {
...
+      local_count[1] = global_count[1] = nx3;
+      WRITEH5SLAB("z", tmpData.data(), gVLocations, local_start, local_count,
+                  global_count, property_list);
+    }
+
+    {
+      // Write Levels and Logical Locations with the level for each Meshblock
+      // loclist contains levels and logical locations for all meshblocks on
+      // all ranks
+      std::vector<LogicalLocation> loclist = pm->GetLocList();
```
JoshuaSBrown: Can  you make this a const reference, doesn't look like you are altering it's contents.

src/interface/state_descriptor.hpp
```diff
@@ -67,10 +70,17 @@ class StateDescriptor {
   // infers type correctly.
   template <typename T>
   const T &Param(const std::string &key, T value) {
-    params_.Get(key, value);
+    return params_.Get(key, value);
+  }
+
+  const std::type_index &ParamType(const std::string &key) {
+    return params_.GetType(key);
   }
```
JoshuaSBrown: These methods look like they should be made const.

src/interface/state_descriptor.hpp
```diff
@@ -67,10 +70,17 @@ class StateDescriptor {
...
 
   Params &AllParams() { return params_; }
+
+  std::vector<std::string> &IncludeWithOutputParams() { return includeWithOutputParams_; }
+
```
JoshuaSBrown: Can be made const noexcept.

JoshuaSBrown: It looks like you are also only adding to includeWithOutputParams_ from within the Statedescriptor class so you could probably return by const reference as well.

src/outputs/parthenon_hdf5.cpp
```diff
@@ -274,45 +279,129 @@ void PHDF5Output::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm) {
...
+          pm->mesh_size.x2min, pm->mesh_size.x2max, pm->mesh_size.x2rat,
+          pm->mesh_size.x3min, pm->mesh_size.x3max, pm->mesh_size.x3rat};
+      const hsize_t rootGridDomain_size[1] = {9};
+      writeH5AF64("RootGridDomain", rootGridDomain,
+                  H5S::FromHIDCheck(H5Screate_simple(1, rootGridDomain_size, NULL)),
+                  infoDSet);
+
+      // RootGridSize - int[3] number of cells on the root grid
+      int rootGridSize[3] = {pm->mesh_size.nx1, pm->mesh_size.nx2, pm->mesh_size.nx3};
+      const hsize_t rootGridSize_size[1] = {3};
```
JoshuaSBrown: I would reverse the definition of rootGridDomain_size and rootGridSize_size so they appear before rootGridDomain and rootGridSize respectively, that way you can use them when defining both structures and get rid of some magic numbers.

src/outputs/parthenon_hdf5.cpp
```diff
@@ -274,45 +279,129 @@ void PHDF5Output::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm) {
...
+      writeH5AI32("RootGridSize", rootGridSize,
+                  H5S::FromHIDCheck(H5Screate_simple(1, rootGridSize_size, NULL)),
+                  infoDSet);
+
+      // BoundaryConditions
+      std::vector<std::string> boundaryConditions(BOUNDARY_NFACES);
+      for (int i = 0; i < BOUNDARY_NFACES; i++) {
+        boundaryConditions[i] = GetBoundaryString(pm->mesh_bcs[i]);
+      }
```
JoshuaSBrown: In general I believe it will be more performant to call reserve with push_back because you get to avoid a default construction call. Not that it will really matter here.

src/outputs/parthenon_hdf5.cpp
```diff
@@ -274,45 +279,129 @@ void PHDF5Output::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm) {
...
+      // VariablesNames - Names of variables within each dataset
+      std::vector<std::string> variableNames;
+
+      for (int i = 0; i < ciX.vars.size(); i++) { // for every block
+        const std::vector<std::string> component_labels =
+            ciX.vars[i]->metadata().getComponentLabels();
+
+        if (component_labels.size() > 0) {
+          numVariables[i] = component_labels.size();
+          for (auto &label : component_labels) {
```
JoshuaSBrown: make const auto

src/outputs/parthenon_hdf5.cpp
```diff
@@ -274,45 +279,129 @@ void PHDF5Output::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm) {
...
+          ss << state->label() << "/" << param_name;
+
+          if (type == typeid(int)) {
+            int param = state->Param<int>(param_name);
+            writeH5AI32(ss.str().c_str(), &param, localDSpace, infoDSet);
+          } else if (type == typeid(double)) {
+            double param = state->Param<double>(param_name);
+            writeH5AF64(ss.str().c_str(), &param, localDSpace, infoDSet);
+          } else if (type == typeid(std::string)) {
+            std::string param = state->Param<std::string>(param_name);
```
JoshuaSBrown: can param be made a const reference?

src/outputs/parthenon_hdf5.cpp
```diff
@@ -274,45 +279,129 @@ void PHDF5Output::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm) {
...
+          const std::type_index type = state->ParamType(param_name);
+
+          std::stringstream ss;
+          ss << state->label() << "/" << param_name;
+
+          if (type == typeid(int)) {
+            int param = state->Param<int>(param_name);
+            writeH5AI32(ss.str().c_str(), &param, localDSpace, infoDSet);
+          } else if (type == typeid(double)) {
+            double param = state->Param<double>(param_name);
```
JoshuaSBrown: make param const?

src/outputs/parthenon_hdf5.cpp
```diff
@@ -274,45 +279,129 @@ void PHDF5Output::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm) {
...
+      const hsize_t variableNames_size[1] = {variableNames.size()};
+      writeH5ASTRINGS("VariableNames", variableNames,
+                      H5S::FromHIDCheck(H5Screate_simple(1, variableNames_size, NULL)),
+                      infoDSet);
+
+      for (auto &package : pm->packages.AllPackages()) {
+        std::shared_ptr<StateDescriptor> state = package.second;
+        // Write IncludeWithOutputParams
+        std::vector<std::string> outputing_params = state->IncludeWithOutputParams();
+        for (auto &param_name : outputing_params) {
```
JoshuaSBrown: make param_name const reference

src/outputs/parthenon_hdf5.hpp
```diff
@@ -256,6 +256,27 @@ static void writeH5ASTRING(const char *name, const std::string &pData,
   PARTHENON_HDF5_CHECK(H5Awrite(attribute, atype, pData.c_str()));
 }
 
+static void writeH5ASTRINGS(const char *name, const std::vector<std::string> &pData,
+                            const hid_t &dSpace, const hid_t &dSet) {
+  int max_name_length = 0;
+  for (auto s : pData) {
```
JoshuaSBrown: make s const reference.

src/outputs/parthenon_hdf5.hpp
```diff
@@ -256,6 +256,27 @@ static void writeH5ASTRING(const char *name, const std::string &pData,
...
+    max_name_length = (max_name_length > s.length()) ? max_name_length : s.length();
+  }
+
+  std::vector<const char *> c_strs;
```
JoshuaSBrown, line 267: call reserve after declaring c_strs;

src/outputs/parthenon_hdf5.cpp
```diff
@@ -372,14 +461,99 @@ void PHDF5Output::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm) {
...
+    }
+
+    {
+      // Write Levels and Logical Locations with the level for each Meshblock
+      // loclist contains levels and logical locations for all meshblocks on
+      // all ranks
+      std::vector<LogicalLocation> loclist = pm->GetLocList();
+
+      std::vector<std::int64_t> levels(pm->nbtotal);
+      std::vector<std::int64_t> logicalLocations(pm->nbtotal * 3);
```
JoshuaSBrown: call reserve instead and use with push_back.

JoshuaSBrown (changes requested):
Thanks for doing this @forrestglines this is a nice piece of work, the only thing that I would really like to see is some tests that account for the extra output.

docs/outputs.md
forrestglines: Unfortunately, `yt` doesn't have any way to navigate tensors seemlessly. Each component would have to be named. From the way the output by Parthenon is implement and how I've implemented the `yt` frontend works currently, I think that would work without little to no modification to the `yt` frontend. 

We can keep that in mind if we want a separate `yt` test in `Parthenon` once the `yt` frontend is firmly solidified.

scripts/python/phdf_diff.py
forrestglines: It looks like these return values were made sequential starting from 1, using the standard return value of 0 to the command line to indicate no errors. I don't think the nonzero return values are treated specially anywhere. Would enumerating the return values at the top of the file be sufficient?

forrestglines:
@par-hermes format

forrestglines:
> Thanks for doing this @forrestglines this is a nice piece of work, the only thing that I would really like to see is some tests that account for the extra output.

The new metadata checking in the python scripts checks for the new metadata added, assuming that the gold data has those extra attributes.

The one new piece that isn't exercised in the testing are the naming of variable components via `AddField` and marking parameters for output via `AddParam`. I could add that to the advection example and that functionality would be covered.

forrestglines:
@par-hermes format

scripts/python/phdf_diff.py
JoshuaSBrown: Yes, I think so, simply having a summary in the docstring of what each exit code means would be nice so you don't have to go digging through the logic.

scripts/python/phdf.py
```diff
@@ -109,35 +109,23 @@ def __init__(self, filename):
...
+                    coord = np.zeros((self.NumBlocks,self.MeshBlockSize[coord_i]))
+                    for bId in range(self.NumBlocks):
+                        for cId in range(self.MeshBlockSize[coord_i]):
+                            coord[bId,cId] = 0.5*(tmp[bId,cId]+tmp[bId,cId+1])
+                coordf=tmp
+                return tmp,coord,coordf
+
+            tmpx, self.x, self.xf = load_coord(0)
+            tmpy, self.y, self.yf = load_coord(1)
+            tmpz, self.z, self.zf = load_coord(2)
```
JoshuaSBrown: Beautiful :)

scripts/python/phdf_diff.py
```diff
@@ -64,7 +64,90 @@ def addPath():
...
+    #Check in the metadata in Info is identical, except use tol for comparing "Time"
+    f0_Info = { key:value for key,value in f0.Info.items() if key != "Time" and key != "BlocksPerPE" }
+    f1_Info = { key:value for key,value in f1.Info.items() if key != "Time" and key != "BlocksPerPE" }
+
+    errTime = np.abs(f0.Time- f1.Time)
+    if errTime > tol:
+        print(f"""
+        Time of outputs differ by {f0.Time - f1.Time}
+
+        Quitting...
```
JoshuaSBrown, line 94: 👍🏻

forrestglines:
@JoshuaSBrown I've covered all the changes you requested. I also added the new metadata functionality to the advection example so that it gets tested during regression testing along with the rest of the HDF5 output.

forrestglines:
This PR is ready to merge except for final approval and reblessing of the gold data for HDF5 outputs. I don't know how to go about reblessing.

docs/outputs.md
```diff
@@ -59,3 +59,108 @@ which will produce a ```png``` image per dump suitable for encoding into a movie
 ## Visualization software
 
 Both [ParaView](https://www.paraview.org/) and [VisIt](https://wci.llnl.gov/simulation/computer-codes/visit/) are capable of opening and visualizing Parthenon graphics dumps.  In both cases, the ```.xdmf``` files should be opened.  In ParaView, select the "XDMF Reader" when prompted.
+
+## Preparing outputs for ```yt```
+
+Parthenon HDF5 outputs can be read with the python visualization library
+[yt](https://yt-project.org/) as certain variables are named during when adding
```
jlippuner:
```suggestion
[yt](https://yt-project.org/) as certain variables are named when adding
```

docs/outputs.md
```diff
@@ -59,3 +59,108 @@ which will produce a ```png``` image per dump suitable for encoding into a movie
...
+```c++
+pkg->AddParam<double>("CodeLength", 100,true);
+pkg->AddParam<double>("CodeMass", 1000,true);
+pkg->AddParam<double>("CodeTime", 1,true);
+pkg->AddParam<double>("AdibaticIndex", 5./3.,true);
+
+pkg->AddParam<int>("IntParam", 0,true);
+pkg->AddParam<std::string>("EquationOfState", "Adiabatic",true);
+```
+all function to add parameters to the HDF5 output. Currently, only ```int```,
```
jlippuner: I don't understand "all function to add parameters to the HDF5 output." here

docs/outputs.md
```diff
@@ -59,3 +59,108 @@ which will produce a ```png``` image per dump suitable for encoding into a movie
...
+```float```, and ```std::string``` parameters can be included with the HDF5.
+
+Defining ```CodeLength```, ```CodeMass```, and ```CodeTime```, all in CGS,
+allows ```yt``` to interpret units for the data. The above example declares the
+data as being in meters, kilograms, and seconds. Alternatively, this unit
```
jlippuner: I am confused. The above example doesn't specify units explicitly, so wouldn't that mean they're in centimeters, grams, seconds by default? Where are meters and kilograms coming from?

scripts/python/phdf.py
```diff
@@ -109,26 +109,23 @@ def __init__(self, filename):
...
-            for bId in range(self.NumBlocks):
-                for cId in range(self.MeshBlockSize[2]):
-                    self.z[bId,cId] = 0.5*(tmpz[bId,cId]+tmpz[bId,cId+1])
-            self.zf = tmpz
+            def load_coord(coord_i):
+                coord_name = ["x","y","z"][coord_i]
+
+                tmp = f[f'/Locations/{coord_name}'][:,:]
+                if f'/VolumeLocations/{coord_name}' in f:
+                    coord = f[f'/VolumeLocations/{coord_name}'][:,:]
```
jlippuner: Hmm, the fact that `f` is used as the file and as the prefix for formatting strings is a bit confusing here... Can we maybe do this?
```suggestion
                tmp = f['/Locations/' + coord_name][:,:]
                vol_loc = '/VolumeLocations/' + coord_name
                if vol_loc in f:
                    coord = f[vol_loc][:,:]
```

scripts/python/phdf_diff.py
```diff
@@ -64,8 +66,121 @@ def addPath():
...
+    #
+    # Note that indices don't match when blocks
+    # are different
+    no_meta_variables_diff = True
+
+    otherBlockIdx = list(f0.findBlockIdxInOther(f1,i) for i in range(f0.NumBlocks))
+
+    for var in set(f0.Variables+f1.Variables):
+        if var in ['Locations', 'VolumeLocations', 'LogicalLocations', 'Levels']:
+            if var in ['Locations', 'VolumeLocations']:
```
jlippuner: This seems unnecessarily complicated to me. Can't we just make the first block `for var in ['Locations', 'VolumeLocations']:` and the second block `for var in ['LogicalLocations', 'Levels']:` and then have a check that `var` exists in the files? That seems more readable to me.

src/interface/metadata.hpp
```diff
@@ -181,6 +181,13 @@ class Metadata {
     SetMultiple(bits);
   }
 
+  /// returns a metadata with bits and shape set
+  explicit Metadata(const std::vector<MetadataFlag> &bits, const std::vector<int> &shape,
+                    const std::vector<std::string> component_labels)
+      : shape_(shape), sparse_id_(-1), component_labels_(component_labels) {
```
jlippuner, line 187: Are there any consistency rules between the `shape` and the number of `component_labels` that we should/could enforce?

src/interface/params.hpp
```diff
@@ -83,12 +84,18 @@ class Params {
     return Get<T>(key);
   }
 
+  const std::type_index &GetType(const std::string &key) const {
+    auto const it = myParams_.find(key);
+    PARTHENON_REQUIRE_THROWS(it != myParams_.end(), "Key " + key + " doesn't exist");
+    return myTypes_.at(key);
```
jlippuner:
```suggestion
    return it->second;
```

src/interface/state_descriptor.hpp
```diff
@@ -49,8 +49,11 @@ class StateDescriptor {
   explicit StateDescriptor(std::string const &label) : label_(label) {}
 
   template <typename T>
-  void AddParam(const std::string &key, T value) {
+  void AddParam(const std::string &key, T value, const bool include_with_output = false) {
```
jlippuner: Is there a reason to not just include all parameters with the output and not give the user an option?

src/interface/state_descriptor.hpp
```diff
@@ -249,6 +261,8 @@ class StateDescriptor {
   Dictionary<std::unordered_map<int, Metadata>> sparseMetadataMap_;
   Dictionary<Metadata> swarmMetadataMap_;
   Dictionary<Dictionary<Metadata>> swarmValueMetadataMap_;
+
+  std::vector<std::string> includeWithOutputParams_;
```
jlippuner: Maybe use a `std::set` to avoid duplicates? Or get rid of this altogether and just write out all params, as suggested above.

src/outputs/parthenon_hdf5.cpp
```diff
@@ -274,45 +279,132 @@ void PHDF5Output::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm) {
...
+          if (type == typeid(int)) {
+            const int &param = state->Param<int>(param_name);
+            writeH5AI32(ss.str().c_str(), &param, localDSpace, infoDSet);
+          } else if (type == typeid(double)) {
+            const double &param = state->Param<double>(param_name);
+            writeH5AF64(ss.str().c_str(), &param, localDSpace, infoDSet);
+          } else if (type == typeid(std::string)) {
+            const std::string &param = state->Param<std::string>(param_name);
+            writeH5ASTRING(ss.str().c_str(), param, localDSpace, infoDSet);
+          }
```
jlippuner: We should probably have `uint32_t`, `uint64_t`, `int64_t`, and `float` here as well. This will become a bit easier with my PR #486 that makes writing attributes a template function on the type.

src/outputs/parthenon_hdf5.cpp
```diff
@@ -372,14 +464,102 @@ void PHDF5Output::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm) {
...
+        logicalLocations.push_back(loc.lx2);
+        logicalLocations.push_back(loc.lx3);
+      }
+
+      // Only write levels on rank 0
+      local_count[0] = (Globals::my_rank == 0) ? pm->nbtotal : 0;
+      local_count[1] = 0;
+      local_count[2] = 0;
+      local_count[3] = 0;
+      local_count[4] = 0;
```
jlippuner: This is not necessary (same for `global_count`) since only the first entry will be ready when we create a rank 1 space

src/outputs/parthenon_hdf5.hpp
```diff
@@ -256,6 +256,28 @@ static void writeH5ASTRING(const char *name, const std::string &pData,
   PARTHENON_HDF5_CHECK(H5Awrite(attribute, atype, pData.c_str()));
 }
 
+static void writeH5ASTRINGS(const char *name, const std::vector<std::string> &pData,
+                            const hid_t &dSpace, const hid_t &dSet) {
+  int max_name_length = 0;
+  for (const auto &s : pData) {
+    max_name_length = (max_name_length > s.length()) ? max_name_length : s.length();
```
jlippuner: Should use `std::max` here, but it appears that `max_name_length` doesn't get used. I wouldn't worry about this function, though, because I have a similar, more general, function to write vector attributes in my PR #486, so I'll remove this in that PR.

jlippuner (approved):
Nice work! This is a good addition to the code to allow easy plotting of Parthenon output with yt.

I added some questions and nitpicks, but nothing blocking.

docs/outputs.md
```diff
@@ -59,3 +59,108 @@ which will produce a ```png``` image per dump suitable for encoding into a movie
...
+
+const int nvelocity = 3;
+std::vector<std::string> velocity_labels(nvelocity);
+velocity_labels[0]="Velocity1";
+velocity_labels[1]="Velocity2";
+velocity_labels[2]="Velocity3";
+m = Metadata({Metadata::Cell, Metadata::Derived}, std::vector<int>({nvelocity}), velocity_labels);
+pkg->AddField("vel", m);
+
+const in npressure = 1;
```
JoshuaSBrown:
```suggestion
const int npressure = 1;
```

docs/outputs.md
```diff
@@ -59,3 +59,108 @@ which will produce a ```png``` image per dump suitable for encoding into a movie
...
+
+#Specify units and adiabatic index explicitly
+units_override = {"length_unit" : (100, "cm"),
+                  "time_unit"   : (1,   "s"),
+                  "mass_unit"   : (1000,"g")}
+
+ds = yt.load(filename,units_override=units_override,gamma=5./3.)
+```
+
+Currently the ```yt``` frontend for Parthenon is hosted on the
```
JoshuaSBrown:
```suggestion
Currently, the ```yt``` frontend for Parthenon is hosted on the
```

scripts/python/phdf_diff.py
```diff
@@ -26,15 +26,16 @@
...
-                 don't print any extraneous info.
-      --tol=eps: set tolerance to eps.  Default 1.0e-12
+                  -all: report all diffs at all positions
+                  -one: Quit after first different variable
+                -brief: Only report if files are different
+                        Overrides --all
+                -quiet: Only report if files are different and
+                        don't print any extraneous info.
+             --tol=eps: set tolerance to eps.  Default 1.0e-12
+      -ignore_metadata: Ignore differences in metadata
```
JoshuaSBrown, line 38: I don't expect any changes, but do you know why some of these are using double dashes and others are not is this a convention? I always thought the double dashes were for the long version and single dashes for short versions of the same argument, it looks like here the double dash prefix is being used only in cases where an assignment is involved. Could probably create an issue (low priority), this should probably be uniform between our python scripts.

src/interface/state_descriptor.hpp
forrestglines: That would be an interesting idea. We don't have that many parameters, so it wouldn't be that much of a file size change to write them all to disk. It might clutter the output a bit but it would simplify the code.

The one obstacle is that currently we don't have methods to write all types of parameters as an HDF5 attribute. The simple integer, floating point, and string types are easy to write, but more complicated ones like the equation of state class that we have in AthenaPK.

I think it would make sense to write all simple parameters in a separate `Params` dataset as long as we're ok with all that data in the HDF5 file. We could limit it to simple types for now and in the future let parameter types implement a `WriteAsHDF5Attribute` if they also need to be added to the HDF5 file. We don't have any use cases for that though.

For this PR I can make all `int`, `double`, and `std::string`  parameters write as an attribute and in your PR we can add other simple types.

src/interface/params.hpp
JoshuaSBrown: ?

JoshuaSBrown (approved):

docs/outputs.md
forrestglines: I'll change to "add the parameters `CodeLength`, `CodeMass`, `CodeTime`, `AdiabaticIndex`, `IntParam`, and `EquationOfState` to the HDF5 output" since it's more descriptive.

docs/outputs.md
forrestglines: You specify `CodeLength`, `CodeMass`, and `Code Time` in terms of centimeters in a code length, grams in a code mass, and seconds in a code time. So in the example you have 100 cm in a code length (1 code length is 1 m ), 1000 g in a code mass (1 code mass is 1 kg), and 1 s in a code time (1 code time is 1 s). I'll make that more clear

scripts/python/phdf.py
forrestglines: That's fine here. Usually I avoid concatenations with strings and prefer the python string formatting, but here it is confusing with the file `f`.

src/interface/state_descriptor.hpp
JoshuaSBrown: If you do end up using the boolean, I would recommend changing it to an enum class which would be more descriptive:

```
enum class IncludeWithOutput {
    True,
    False
};

AddParam<std::string>("Parthenon is Awesome", IncludeWithOutput::True);
```

scripts/python/phdf_diff.py
forrestglines: Yes. This was the result of this code block moving around.

src/outputs/parthenon_hdf5.cpp
forrestglines: Yes. I only added the types that already had a convenient macro to write the attribute. In your PR with the template function, we can add the other obvious types.

docs/outputs.md
```diff
@@ -59,3 +59,114 @@ which will produce a ```png``` image per dump suitable for encoding into a movie
...
+output. Currently, only ```int```, ```float```, and ```std::string```
+parameters can be included with the HDF5.
+
+Code units can be defined for ```yt``` by including the parameters
+```CodeLength```, ```CodeMass```, and ```CodeTime```, which specify the code
+units used by Parthenon in terms of centimeters, grams, and seconds by writing
+the parameters.  In the above example, these parameters dictate ```yt``` to
+interpret code lengths in the data in units of 100 centimeters (or 1 meter per
+code unit), code masses in units of 1000 grams ( or 1 kilogram per code units)
+and code times in units of seconds ( or 1 second per code time).
```
jlippuner: Thanks, I get it now :) Two small fixes:
```suggestion
code unit), code masses in units of 1000 grams (or 1 kilogram per code units)
and code times in units of seconds (or 1 second per code time).
```

src/outputs/parthenon_hdf5.cpp
jlippuner: I'm fine with this.

src/interface/state_descriptor.hpp
jlippuner: I realized this later. Yes, I think just writing all params with simple type to the outputfile and ignoring complex types is the right way to go.

jlippuner: That's a good suggestion. Although it would be a non-issue if we just output all parameters of simple type as discussed above.

jlippuner (approved):

scripts/python/phdf_diff.py
forrestglines: I also don't see any consistency here. I think `./tst/regression/run_test.py` is the only other python script to do argument parsing.

src/interface/state_descriptor.hpp
forrestglines: I'm removing the boolean, otherwise it's a good suggestion.

src/interface/state_descriptor.hpp
forrestglines: Removing this vector with the new commit.

scripts/python/phdf.py
```diff
@@ -109,26 +109,24 @@ def __init__(self, filename):
...
+
+                tmp = f['/Locations/' + coord_name][:,:]
+                vol_loc = '/VolumeLocations/' + coord_name
+                if vol_loc in f:
+                    coord = f[vol_loc][:,:]
+                else:
+                    coord = np.zeros((self.NumBlocks,self.MeshBlockSize[coord_i]))
+                    for bId in range(self.NumBlocks):
+                        for cId in range(self.MeshBlockSize[coord_i]):
+                            coord[bId,cId] = 0.5*(tmp[bId,cId]+tmp[bId,cId+1])
```
pgrete, line 126: Wouldn't sth like
```suggestion
                    coord = 0.5*(tmp[:,:-1] + tmp[:,1:])
```
also work here as both for loops cover the entire range?

scripts/python/phdf_diff.py
pgrete: I also think consistency would be nice and also not necessary for this PR -> "good first issue"

scripts/python/phdf_diff.py
```diff
@@ -64,8 +66,120 @@ def addPath():
...
+        Error codes:
+            10 : Times in hdf files differ
+            11 : Attribute names in Info of hdf files differ
+            12 : Values of attribute names in Info of hdf files differ
+            13 : Meta variables (Locations, VolumeLocations, LogicalLocations, Levels) differ
+    """
+    ERROR_TIME_DIFF=10
+    ERROR_INFO_ATTRS_DIFF=12
+    ERROR_INFO_VALUES_DIFF=13
+    ERROR_META_VARS_DIFF=14
```
pgrete:
```suggestion
    ERROR_INFO_ATTRS_DIFF=11
    ERROR_INFO_VALUES_DIFF=12
    ERROR_META_VARS_DIFF=13
```

Probably a leftover from an earlier iteration, but the indices between doc and code don't match.

scripts/python/phdf_diff.py
```diff
@@ -64,8 +66,120 @@ def addPath():
...
+
+    #Check in the metadata in Info is identical, except use tol for comparing "Time"
+    f0_Info = { key:value for key,value in f0.Info.items() if key != "Time" and key != "BlocksPerPE" }
+    f1_Info = { key:value for key,value in f1.Info.items() if key != "Time" and key != "BlocksPerPE" }
+
+    errTime = np.abs(f0.Time- f1.Time)
+    if errTime > tol:
+        print(f"""
```
pgrete, line 91: Nice, f-strings, I approve ;)

scripts/python/phdf_diff.py
```diff
@@ -64,8 +66,120 @@ def addPath():
...
+
+    for var in set(f0.Variables+f1.Variables):
+        if var in ['Locations', 'VolumeLocations']:
+            for key in f0.fid[var].keys():
+                #Compare raw data of these variables
+                val0 = f0.fid[var][key]
+                val1 = f1.fid[var][key]
+
+                #Sort val1 by otherBlockIdx
+                val1 = val1[otherBlockIdx]
```
pgrete, line 170: Leaving a note for myself here to follow up on what's happening here.
Haven't seen that kind of indexing with a list yet.

scripts/python/phdf_diff.py
```diff
@@ -64,8 +66,120 @@ def addPath():
...
+        if var in ['LogicalLocations', 'Levels']:
+            #Compare raw data of these variables
+            val0 = f0.fid[var]
+            val1 = f1.fid[var]
+
+            #Sort val1 by otherBlockIdx
+            val1 = val1[otherBlockIdx]
+
+            #As integers, they should be identical
+            if val0 != val0:
```
pgrete:
```suggestion
            if val0 != val1:
```
Also, does this like work without an `.any()` ?

scripts/python/phdf_diff.py
```diff
@@ -64,8 +66,120 @@ def addPath():
...
+            10 : Times in hdf files differ
+            11 : Attribute names in Info of hdf files differ
+            12 : Values of attribute names in Info of hdf files differ
+            13 : Meta variables (Locations, VolumeLocations, LogicalLocations, Levels) differ
+    """
+
+    ERROR_NO_OPEN_F0 = 1
+    ERROR_NO_OPEN_F1 = 2
+    ERROR_CELLS_DIFFER = 3
+    ERROR_DATA_DIFFER = 4
```
pgrete: :+1: I like this.

forrestglines:
@par-hermes format

scripts/python/phdf.py
pgrete: Also, may be worth adding a warning here (printed) that this assume uniform Cartesian grids so that we don't miss it when adding other coordinate systems.

pgrete (approved):
Nice work! I like the flexible design through the state descriptor.
My major comments mostly concern the Python piece which may be worth double checking.

src/interface/metadata.hpp
forrestglines: Not any simple rules right now. Currently, the `yt` interface would accept any number of `component_labels` as long as it could reshape the raw array into right `nlabels*nz*ny*nx` shape. Since we're not sure at the moment how downstream codes will arrange or label their components, I'm not sure there's a simple rule that we can impose at the moment.

pgrete:
Additional note on updating the "gold standard":

1. In an environment where current `develop` passes create the new gold standard and put into a single file named `parthenon_regression_gold_v#.tgz`  with `#` being the "next" number (here 4)
2. Go to https://github.com/lanl/parthenon/releases and create a new release with the file as binary attachment and branch name `regression-gold-v#`, cf., e.g., https://github.com/lanl/parthenon/releases/tag/regression-gold-v3
3. Edit the root `CMakeLists.txt` (in this PR) and update both number and SHA512 hash:
```
set(REGRESSION_GOLD_STANDARD_VER 3 CACHE STRING "Version of gold standard to download and use")
set(REGRESSION_GOLD_STANDARD_HASH
  "SHA512=2445000a031aafc9a85a684aa7e82fb5ccda05530c5652ab432a8fb254642c7b18252ff55d8306b1120aba898713712b28b804eb30272b06278698046a6461cc"
  CACHE STRING "Hash of default gold standard file to download")
```
4. IIRC that's it: Fingers crossed ;)

scripts/python/phdf_diff.py
forrestglines: Good catch, I just fixed this

scripts/python/phdf_diff.py
forrestglines: `otherBlockIdx` gives the indexing of mesh blocks on `f1` in the same order as the meshblocks in `f0`, since meshblocks are free to rearrange in order due to changing MPI ranks.

scripts/python/phdf_diff.py
forrestglines: It should be `np.any( val0 != val1)` for full correctness. It might be checking equivalence by reference here

forrestglines:
@pgrete Thanks! I'll rebless the tests and merge once the CI passes in this PR.

forrestglines:
>     2\. Go to https://github.com/lanl/parthenon/releases and create a new release with the file as binary attachment and branch name `regression-gold-v#`, cf., e.g., https://github.com/lanl/parthenon/releases/tag/regression-gold-v3

So do I use the tag `regression-gold-v4` and target branch `develop` or the branch in this PR?

PR #479, opened 2021-03-30
Add `Update` function to `Params` to update the value of an existing key

pgrete:
## PR Summary

Small quality of life improvement: Add `Update` function to `Params` to update the value of an existing key, which was previously not possible
Eventually, it'd be nice to have a separate `MeshVariables` dictionary with transparent reductions at the driver level.
For now `Params`  with `Update` does the trick to enable mesh level parameters in downstream codes.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


JoshuaSBrown (approved):
This looks good to me.

src/interface/params.hpp
```diff
@@ -39,11 +39,21 @@ class Params {
...
     myTypes_[key] = std::string(typeid(value).name());
   }
 
+  /// Updates existing object
+  /// Throws an error if the key is not already in use
+  template <typename T>
+  void Update(const std::string &key, T value) {
+    PARTHENON_REQUIRE_THROWS((hasKey(key)), "Key " + key + "missing.");
+    PARTHENON_REQUIRE_THROWS(!(myTypes_.at(key).compare(std::string(typeid(T).name()))),
+                             "WRONG TYPE FOR KEY '" + key + "'");
```
jlippuner, line 53: Hmm, I guess we can't use `std::is_same` or any other compile-time check here because we don't have the type of the existing key at compile time, right?

tst/unit/test_unit_params.cpp
```diff
@@ -23,24 +23,36 @@
...
-      auto const &const_params = params;
-      REQUIRE(params.Get<double>(key) == Approx(value));
-    }
-    WHEN("the same key is provided a second time") {
-      REQUIRE_THROWS_AS(params.Add(key, value), std::runtime_error);
-    }
-
-    WHEN("attempting to get the key but casting to a different type") {
-      REQUIRE_THROWS_AS(params.Get<int>(key), std::runtime_error);
+    THEN("we can add it to Params") {
```
jlippuner, line 31: Should there be a test to make sure it throws if we try to update a non-existent key?

jlippuner (approved):
Looks good. I have a couple of small comments but nothing blocking

src/interface/params.hpp
Yurlungur: Yeah it's a runtime thing.

Yurlungur (approved):
This seems fine to me. My one concern would be that a "modify" function makes params more like global variables... which has kind of a bad code smell. But I do think this is a useful functionality.

I wonder if others have thoughts. @nmsriram ? @jdolence ?

pgrete:
> This seems fine to me. My one concern would be that a "modify" function makes params more like global variables... which has kind of a bad code smell. But I do think this is a useful functionality.

I agree that this is not ideal. I added a "Global/Mesh variables in downstream codes"  item to the agenda for the next call (though people should still feel free to already comment here). I'm fairly sure this use case will come up more often with more and more stuff being added to downstream codes so a clean implementation of this kind of functionality would be desirable.

tst/unit/test_unit_params.cpp
pgrete: Yes, I just added that test.

PR #480, opened 2021-03-31
particles-regression fixes

brryan:
## PR Summary

@pgrete ran into some issues standing up a new particle example. This fix adds one non-obvious line to his example problem, and fixes another bug that subsequently arose when particles have integer data fields.

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted

### DIFF
--- a/example/particle_leapfrog/particle_leapfrog.cpp
+++ b/example/particle_leapfrog/particle_leapfrog.cpp
@@ -158,12 +158,20 @@ TaskStatus WriteParticleLog(MeshBlock *pmb) {
 }
 
 // initial particle position: x,y,z,vx,vy,vz
-constexpr int num_test_particles = 4;
+/*constexpr int num_test_particles = 4;
 const std::array<std::array<Real, 6>, num_test_particles> particles_ic = {{
     {-0.1, 0.2, 0.3, 1.0, 0.0, 0.0},   // along x direction
     {0.4, -0.1, 0.3, 0.0, 1.0, 0.0},  // along y direction
     {-0.1, 0.3, 0.2, 0.0, 0.0, 1.0},  // along z direction
     {0.12, 0.2, -0.3, 1.0, 1.0, 1.0}, // along diagonal
+}};*/
+constexpr int num_test_particles = 1;
+const std::array<std::array<Real, 6>, num_test_particles> particles_ic = {{
+    {-0.1, 0.2, 0.3, 1.0, 0.0, 0.0}
+    /*,   // along x direction
+    {0.4, -0.1, 0.3, 0.0, 1.0, 0.0},  // along y direction
+    {-0.1, 0.3, 0.2, 0.0, 0.0, 1.0},  // along z direction
+    {0.12, 0.2, -0.3, 1.0, 1.0, 1.0}, // along diagonal*/
 }};
 
 void ProblemGenerator(MeshBlock *pmb, ParameterInput *pin) {
--- a/example/particle_leapfrog/particle_leapfrog.cpp
+++ b/example/particle_leapfrog/particle_leapfrog.cpp
@@ -158,20 +158,12 @@ TaskStatus WriteParticleLog(MeshBlock *pmb) {
 }
 
 // initial particle position: x,y,z,vx,vy,vz
-/*constexpr int num_test_particles = 4;
+constexpr int num_test_particles = 4;
 const std::array<std::array<Real, 6>, num_test_particles> particles_ic = {{
     {-0.1, 0.2, 0.3, 1.0, 0.0, 0.0},   // along x direction
     {0.4, -0.1, 0.3, 0.0, 1.0, 0.0},  // along y direction
     {-0.1, 0.3, 0.2, 0.0, 0.0, 1.0},  // along z direction
     {0.12, 0.2, -0.3, 1.0, 1.0, 1.0}, // along diagonal
-}};*/
-constexpr int num_test_particles = 1;
-const std::array<std::array<Real, 6>, num_test_particles> particles_ic = {{
-    {-0.1, 0.2, 0.3, 1.0, 0.0, 0.0}
-    /*,   // along x direction
-    {0.4, -0.1, 0.3, 0.0, 1.0, 0.0},  // along y direction
-    {-0.1, 0.3, 0.2, 0.0, 0.0, 1.0},  // along z direction
-    {0.12, 0.2, -0.3, 1.0, 1.0, 1.0}, // along diagonal*/
 }};
 
 void ProblemGenerator(MeshBlock *pmb, ParameterInput *pin) {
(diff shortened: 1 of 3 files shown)


src/interface/swarm.cpp
```diff
@@ -852,7 +852,7 @@ void Swarm::UnloadBuffers_() {
           }
           for (int i = 0; i < int_vars_size; i++) {
             vint(i, sid) = static_cast<int>(
-                bdvar.recv[nid]((real_vars_size + bid) * particle_size + i));
+                bdvar.recv[nid](bid * particle_size + real_vars_size + i));
```
brryan, line 855: I unrolled this "indexing" (real + int vars in the same 1D buffer) incorrectly -- thanks for making this apparent @pgrete!

example/particle_leapfrog/particle_leapfrog.cpp
```diff
@@ -276,6 +276,9 @@ TaskStatus TransportParticles(MeshBlock *pmb, const StagedIntegrator *integrator
           x(n) += vx(n) * 0.5 * dt;
           y(n) += vy(n) * 0.5 * dt;
           z(n) += vz(n) * 0.5 * dt;
+
+          bool on_current_mesh_block = true;
+          swarm_d.GetNeighborBlockIndex(n, x(n), y(n), z(n), on_current_mesh_block);
```
brryan, line 282: This call is necessary because it sets an internal `Swarm` variable saying that this particle is off the current meshblock and boundary conditions/communication need to be applied to it. This is not great design (at *least* I should have added a comment to my example saying why this needs to be there) but I did it this way to avoid having the `Swarm` internally call boundary checking functions many times.

brryan:
@pgrete OK I think I (mostly) understand what is going on here. There needed to be a call to `GetNeighborBlockIndex` to let the `Swarm` know particles had moved off their `MeshBlock`s and needed to communicate this. Bad design on my part, happy to discuss a better path forward here. Sorry for not making this clearer in my particles example. But thank you for writing a whole particles example yourself to see how things work! You also exposed a bug in my code which is happily fixed now.

This *almost* works -- here's the final output:
```
0 , 0 , 0 , -0.0566987 , 0.2 , 0.3 , 1 , 0 , 0
0 , 0 , 1 , 0.4 , -0.0566987 , 0.3 , 0 , 1 , 0
0 , 0 , 3 , 0.163301 , 0.243301 , -0.256699 , 1 , 1 , 1
0 , 0 , 2 , -0.1 , 0.3 , 0.243301 , 0 , 0 , 1
```
whereas that should be
```
0 , 0 , 0 , -0.1 , 0.2 , 0.3 , 1 , 0 , 0
0 , 0 , 1 , 0.4 , -0.1 , 0.3 , 0 , 1 , 0
0 , 0 , 3 , 0.12 , 0.2 , -0.3 , 1 , 1 , 1
0 , 0 , 2 , -0.1 , 0.3 , 0.2 , 0 , 0 , 1
```

Looking at the code a bit, it seems like the particles aren't being transported for the last timestep... I'm not sure why this is yet but I'll keep looking. `dt` is also a bit suspicious towards the end:

```
cycle=89 time=9.6345326171019008e-01 dt=1.0825317547305485e-02 zone-cycles/wsec_step=1.89e+07 wsec_step=2.17e-04
0 , 0 , 0 , -0.0824202 , 0.2 , 0.3 , 1 , 0 , 0
0 , 0 , 1 , 0.4 , -0.0824202 , 0.3 , 0 , 1 , 0
0 , 0 , 3 , 0.13758 , 0.21758 , -0.28242 , 1 , 1 , 1
0 , 0 , 2 , -0.1 , 0.3 , 0.21758 , 0 , 0 , 1
cycle=90 time=9.7427857925749561e-01 dt=1.0825317547305485e-02 zone-cycles/wsec_step=1.99e+07 wsec_step=2.06e-04
0 , 0 , 0 , -0.0715948 , 0.2 , 0.3 , 1 , 0 , 0
0 , 0 , 1 , 0.4 , -0.0715948 , 0.3 , 0 , 1 , 0
0 , 0 , 3 , 0.148405 , 0.228405 , -0.271595 , 1 , 1 , 1
0 , 0 , 2 , -0.1 , 0.3 , 0.228405 , 0 , 0 , 1
cycle=91 time=9.8510389680480115e-01 dt=1.0825317547305485e-02 zone-cycles/wsec_step=1.99e+07 wsec_step=2.06e-04
0 , 0 , 0 , -0.0607695 , 0.2 , 0.3 , 1 , 0 , 0
0 , 0 , 1 , 0.4 , -0.0607695 , 0.3 , 0 , 1 , 0
0 , 0 , 3 , 0.15923 , 0.23923 , -0.26077 , 1 , 1 , 1
0 , 0 , 2 , -0.1 , 0.3 , 0.23923 , 0 , 0 , 1
cycle=92 time=9.9592921435210668e-01 dt=4.0707856478933158e-03 zone-cycles/wsec_step=1.97e+07 wsec_step=2.07e-04
0 , 0 , 0 , -0.0566987 , 0.2 , 0.3 , 1 , 0 , 0
0 , 0 , 1 , 0.4 , -0.0566987 , 0.3 , 0 , 1 , 0
0 , 0 , 3 , 0.163301 , 0.243301 , -0.256699 , 1 , 1 , 1
0 , 0 , 2 , -0.1 , 0.3 , 0.243301 , 0 , 0 , 1
cycle=93 time=1.0000000000000000e+00 dt=8.1415712957866315e-03 zone-cycles/wsec_step=6.05e+05 wsec_step=6.77e-03
```

Also, if you're happy with these changes do feel free to merge back into your `pgrete/particles-regression` branch.

brryan:
@pgrete OK I figured out what was going on with the slightly off particle timesteps. Because this `Transport` call just pushes particles by `dt` without checking to see whether the particles would go beyond `t+dt` for the step (which by itself is fine), the task list from the previous example (where you loop over transport->send->receive until there are zero sends) leads to an extra particle push every time you send a particle because it gets pushed again by the block it arrives at. To avoid this we can actually simplify the particle logic a lot, because with the way you limit the simulation timestep a particle should never cross to more than one neighbor per timestep. I made this reduction and the example works, I think we could get away without even having a custom `Step` function with this logic. This works with MPI as well, I recast `StopCommunicationMesh` as `InitializeCommunicationMesh` that just resets all the MPI flags. Maybe this should be a `Swarm` method...

JoshuaSBrown:
I'm guessing this pr is WIP?

brryan:
I think this is ready, though it's only getting merged in to `pgrete/particles-regression`

example/particle_leapfrog/particle_leapfrog.cpp
pgrete: I see. In that case I suggest to rename the function to expose that it also make some internal changes (which I naively would not expect from a `Get` function.
Similarly, that function is currently marked `const`, which I'm now also wondering if it's semantically correct as it acutally updates the content of the private `blockIndex_`.

pgrete:
Thanks for the cleanup. i can confirm that this fixes the particle communication in the example.
I will merge continue to report in the original PR.

PR #481, opened 2021-04-01
Cleanup unused/untested/not fully ported code (mostly OpenMP and reconstruction)

pgrete:
## PR Summary

Raised by a question from @JoshuaSBrown on Matrix regarding the requirement for a thread safe MPI lib, I cleaned up some code that I had my eye on for some time now:
- Removed all leftover OpenMP logic that in Athena++ was required for coarse grained parallelism (threads for blocks). Parthenon does not support coarse grained parallelism at this point and (re)introducing it in combination with Kokkos requires more complex/low level changes, see #190). **Fine grained OpenMP parallelism (threads for loops) managed by Kokkos is not affected by this PR**
- Removed the non Kokkos version of `WeightedAve` leftover from Athena++. We have those functions in `update.#pp` and already use those.
- Removed non Kokkos versions of various reconstruction methods (and the associated logic) as they don't work on devices in their current form. The Kokkosized inline functions for first and second order reconstruction (that are also used in the advection example) are still there and working. I'm also happy to add the inline version of PPM that is currently used in AthenaPK to Parthenon.
- Removed unused defines that had no meaning (nor use) in Parthenon such as `NWAVE` or `MAX_NSTAGE`

My main intention is to keep the code in Parthenon as "clean" and maintainable as possible.
Given that we started to deviate from the Athena++ basis already in many places (and will continue to do so), I prefer to remove more and more leftover unused code so that we can focus on a clean implementation aligned with the current Parthenon basis when (re)introducing new features.

These changes may potentially affect downstream codes (e.g., in terms of header files included or "host only" code for reconstruction), which is why I'm tagging/requesting many people for review (though only one team member of each code should actually have a closer look).

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


JoshuaSBrown:
Wow, well I didn't mean to make you do a bunch of extra work when I asked the question but now that you have this is great.

src/utils/show_config.cpp
```diff
@@ -32,12 +32,9 @@ namespace parthenon {
 void ShowConfig() {
   // To match configure.py output: use 2 space indent for option, value output starts on
   // column 30
-  std::cout << "This Athena++ executable is configured with:" << std::endl;
+  std::cout << "This Parthenon library is configured with:" << std::endl;
```
Yurlungur, line 35: whoops good catch

Yurlungur (approved):
This is a welcome change! I will have to double check that Phoebus and Riot build with all this stuff removed, but I would be shocked if they didn't.

JoshuaSBrown (approved):

jlippuner (approved):
Very nice! It's oddly satisfying to see an almost all-red PR :)

jlippuner:
I'll go ahead and merge this, it's been sitting for a while with enough approvals

Issue #466, opened 2021-03-04
Tiling when using Kokkos MDRange and others


JoshuaSBrown:
It will be worth finding reasonable defaults once the code base matures a bit more for the tiling parameter that can be specified alongside the different Kokkos policies. Using the defaults can lead to significant performance degradation.

Yurlungur:
And this may be hardware or meshblock size dependent.

pgrete:
Closing as this has been addressed in #487 
Feel free to reopen if performance is still not as expected.

PR #482, opened 2021-04-01
Add history output

pgrete:
## PR Summary

Introduce a history file capability similar to what is available in Athena++.
Works across packages, see documentation and advection example on how to use.
The output for the advection example now looks like
```
#  History data
# [1]=time     [2]=dt       [3]=total_advected[4]=max_advected[5]=min_advected
  0.00000e+00  4.68750e-03  1.39160e-02  1.00000e+00  0.00000e+00
  2.53125e-01  4.68750e-03  1.39160e-02  9.38903e-01  0.00000e+00
  5.01563e-01  4.68750e-03  1.39160e-02  7.44601e-01  6.35449e-16
  7.54687e-01  4.68750e-03  1.39160e-02  5.89906e-01  1.27520e-09
  1.00000e+00  3.12500e-03  1.39160e-02  4.80914e-01  1.45889e-07
```

This also includes the fix I mention on Matrix with cleaning the cached block list of the `MeshData` container.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


Yurlungur (commented):
I only skimmed this since it doesn't look like you're ready for review. I like this. And I also like the idea of enrolling history functionality in packages.

pgrete:
> I only skimmed this since it doesn't look like you're ready for review. I like this. And I also like the idea of enrolling history functionality in packages.

Correct, but I only mentioned it verbally that documentation and example is still missing here.
I now added both and this is ready for review.

docs/interface/state.md
```diff
@@ -25,6 +25,52 @@ The reasoning for providing `FillDerived*` and `EstimateTimestep*` function poin
...
+
+```c++
+// List (vector) of HistoryOutputVar that will all be enrolled as output variables
+parthenon::HstVar_list hst_vars = {};
+
+// Add a callback function
+hst_vars.emplace_back(parthenon::HistoryOutputVar({UserHistoryOperation::sum, MyHstFunction, "my label"}));
+
+// add callbacks for HST output identified by the `hist_str`
+pkg->AddParam<>(parthenon::hist_str, hst_vars);
```
Yurlungur, line 44: I take it the `parthenon::hist_str` is the param name and it's called in a loop over packages? I don't love the name `hist_str`. It sounds like it has a deeper meaning. Maybe `parthenon::hist_param` or `parthenon::hist_param_key` or  something?

Yurlungur: I take it the `parthenon::hist_str` is the param name and it's called in a loop over packages? I don't love the name `hist_str`. It sounds like it has a deeper meaning. Maybe `parthenon::hist_param` or `parthenon::hist_param_key` or  something?

docs/interface/state.md
```diff
@@ -25,6 +25,52 @@ The reasoning for providing `FillDerived*` and `EstimateTimestep*` function poin
...
+```
+
+Here, `HistoryOutputVar` is a `struct` containing the global (over all blocks of all ranks) reduction operation, `MyHstFunction` is a callback function (see below), and `"my label"` is the string to
```
Yurlungur, line 47: Why does the constructor for `HistoryOutputVar` require a list initialization? We could add a constructor maybe that removes the need for those `{` and `}`?

docs/interface/state.md
```diff
@@ -25,6 +25,52 @@ The reasoning for providing `FillDerived*` and `EstimateTimestep*` function poin
...
+Real MyHstFunction(MeshData<Real> *md);
+```
+i.e., they will always work on `MeshData`.
+*Note*, currently history output will always be calculated for the "base" container.
+More specifically, the output machinery will automatically use (or create if non existent)
+a single "base" `MeshData` object containing *all* blocks of a rank.
+This simplifies the the logic for reductions over all blocks of a rank and also (generally)
+resuls in better performance as the number of kernel calls is reduced.
+However, this also implies the expectation that the "base" container holds the most recent
+data at the end of a timestep.
```
Yurlungur, line 72: This seems fine.

example/advection/advection_package.cpp
```diff
@@ -276,6 +298,50 @@ void PostFill(MeshBlockData<Real> *rc) {
...
 }
 
+// Example of how to enroll a history function.
+// Templating is *NOT* required and just implemented here to reuse this function
+// for testing of the UserHistoryOperations curently available in Parthenon (Sum, Min,
+// Max), which translate to the MPI reduction being called over all ranks. T should be
+// either Kokkos::Sum, Kokkos::Min, or Kokkos::Max.
+template <typename T>
+Real AdvectionHst(MeshData<Real> *md) {
+  auto hydro_pkg =
```
Yurlungur: This should probably say `advection_pkg`

example/advection/advection_package.cpp
```diff
@@ -276,6 +298,50 @@ void PostFill(MeshBlockData<Real> *rc) {
...
+
+  Real result = 0.0;
+  T reducer(result);
+
+  // We choose to apply volume weighting when using the sum reduction.
+  // Downstream this choice will be done on a variable by variable basis and volume
+  // weighting needs to be applied in the reduction region.
+  bool volume_weighting = std::is_same<T, Kokkos::Sum<Real, DevExecSpace>>::value;
+
+  Kokkos::parallel_reduce(
```
Yurlungur: why not use `par_reduce` since we have it?

example/advection/parthinput.advection
```diff
@@ -75,3 +76,7 @@ dt = 0.05
 variables = advected, one_minus_advected, & # comments are ok
             one_minus_advected_sq, & # on every (& characters are ok in comments)
             one_minus_sqrt_one_minus_advected_sq # line
+
+<parthenon/output3>
+file_type = hst
+dt = 0.05
```
Yurlungur, line 82: Newline

src/interface/mesh_data.hpp
```diff
@@ -172,6 +172,7 @@ class MeshData {
   }
 
   void ClearCaches() {
+    block_data_.clear();
```
Yurlungur, line 175: :+1: 

Worth noting this is not required for the history MR necessarily. It's a fix of an older bug, as discussed on the Matrix channel.

src/outputs/outputs.hpp
```diff
@@ -124,6 +125,23 @@ class OutputType {
...
 };
 
+//----------------------------------------------------------------------------------------
+// Helper definitions to enroll user output variables
+
+// Function signature for currently supported user output functions
+using HstFun_t = std::function<Real(MeshData<Real> *md)>;
+
+// Container
+struct HistoryOutputVar {
```
Yurlungur, line 135: I would add a constructor to this struct. e.g.,
```C++
HistoryOutputVar(const UserHistoryOperation &hst_op_, const HstFun_t &hst_fun_, const std::string &label_)
  : hst_op(hst_op_), hst_fun(hst_fun_), label(label_) { }
```
just to minimize unnecessary copies and remove the need for the list initialization

tst/regression/test_suites/output_hdf5/parthinput.advection
```diff
@@ -70,3 +70,7 @@ dt = 1.0
 variables = advected, one_minus_advected, & # comments are ok
             one_minus_advected_sq, & # on every (& characters are ok in comments)
             one_minus_sqrt_one_minus_advected_sq # line
+
+<parthenon/output1>
+file_type = hst
+dt = 0.25
```
Yurlungur: newline

Yurlungur (approved):
This is an important feature and I'm really happy to see it re-enabled. I also like the design you came up with. Some minor nitpicks below. Consider them non-blocking.

Yurlungur (approved):

docs/interface/state.md
```diff
@@ -25,6 +25,52 @@ The reasoning for providing `FillDerived*` and `EstimateTimestep*` function poin
 
 In Parthenon, each `Mesh` and `MeshBlock` owns a `Packages_t` object, which is a `std::map<std::string, std::shared_ptr<StateDescriptor>>`.  The object is intended to be populated with a `StateDescriptor` object per package via an `Initialize` function as in the advection example [here](../example/advection/advection.cpp).  When Parthenon makes use of the `Packages_t` object, it iterates over all entries in the `std::map`.  Note that it's often useful to add a `StateDescriptor` to the `Packages_t` object for the overall application, allowing for a convenient way to define global parameters, for example.
 
+## History output
+
+Parthenon allows packages to enroll an arbitrary number of "history" functions that are all
+called at the interval according to the input parameters,
+see [output documention](../outputs.md#History-Files).
+
+To entroll functions create a list of callback function with the appropriate reduction operation:
```
JoshuaSBrown: typo `entroll`

example/advection/advection_package.cpp
pgrete: Because it's slow, see #466 ;) I updated to par_reduce here and will fix `par_reduce` in a separate PR.

example/advection/advection_package.cpp
pgrete: Yes, also it's not even used in this function :D So I removed it.

docs/interface/state.md
pgrete: Good point. I think `parthenon::hist_param_key` is more verbose and easier to understand so I changed that.

docs/interface/state.md
pgrete: It doesn't require. I was just lazy ;) I added the constructor.

pgrete:
I addressed the review comments.
@JoshuaSBrown  or @jlippuner would be great if one of you could give it a look for a second review.

example/advection/advection_package.cpp
```diff
@@ -276,6 +298,45 @@ void PostFill(MeshBlockData<Real> *rc) {
...
+Real AdvectionHst(MeshData<Real> *md) {
+  auto pmb = md->GetBlockData(0)->GetBlockPointer();
+
+  // Packing variable over MeshBlock as the function is called for MeshData, i.e., a
+  // collection of blocks
+  const auto &advected_pack = md->PackVariables(std::vector<std::string>{"advected"});
+
+  IndexRange ib = advected_pack.cellbounds.GetBoundsI(IndexDomain::interior);
+  IndexRange jb = advected_pack.cellbounds.GetBoundsJ(IndexDomain::interior);
+  IndexRange kb = advected_pack.cellbounds.GetBoundsK(IndexDomain::interior);
```
JoshuaSBrown: make const

example/advection/advection_package.cpp
```diff
@@ -276,6 +298,45 @@ void PostFill(MeshBlockData<Real> *rc) {
...
+
+  Real result = 0.0;
+  T reducer(result);
+
+  // We choose to apply volume weighting when using the sum reduction.
+  // Downstream this choice will be done on a variable by variable basis and volume
+  // weighting needs to be applied in the reduction region.
+  bool volume_weighting = std::is_same<T, Kokkos::Sum<Real, DevExecSpace>>::value;
```
JoshuaSBrown: make const

src/outputs/history.cpp
```diff
@@ -42,7 +44,108 @@ namespace parthenon {
...
+
+  // Loop over all packages of the application
+  for (const auto &pkg : pm->packages.AllPackages()) {
+    // Check if the package has enrolled functions which are stored in the
+    // Params under the `hist_param_key` name.
+    const auto &params = pkg.second->AllParams();
+    if (!params.hasKey(hist_param_key)) {
+      continue;
+    }
+    auto hist_vars = params.Get<HstVar_list>(hist_param_key);
```
JoshuaSBrown: make const reference if possible.

src/outputs/history.cpp
```diff
@@ -42,7 +44,108 @@ namespace parthenon {
...
+
+    for (auto &hist_var : hist_vars) {
```
JoshuaSBrown: make const

src/outputs/history.cpp
```diff
@@ -42,7 +44,108 @@ namespace parthenon {
...
+        PARTHENON_MPI_CHECK(MPI_Reduce(MPI_IN_PLACE, &result, 1, MPI_PARTHENON_REAL,
+                                       usr_op, 0, MPI_COMM_WORLD));
+      } else {
+        PARTHENON_MPI_CHECK(MPI_Reduce(&result, &result, 1, MPI_PARTHENON_REAL, usr_op, 0,
+                                       MPI_COMM_WORLD));
+      }
+#endif
+
+      all_results.emplace_back(result);
+      all_labels.emplace_back(hist_var.label);
```
JoshuaSBrown, line 100: Call reserve before the loop since you alread know its size.

JoshuaSBrown (approved):
Thanks for this doing this, this is a nice chunk of code. I approve :) with some non blocking changes requested.

src/outputs/history.cpp
pgrete: The history output collects results from all packages. Thus, a single `reserve` would mean to query all packages first (adding another for loop over packages) or call a reserve for each package.
I don't expect this to be a performance issues (as the history output is usually only called everyone once in a while and the number of variables is typically small, say < 10-20). Thus, there should be no overhead for adding a `string` and one `double` in the loop.
So I prefer to keep the code here as is as it's logically the simplest.

JoshuaSBrown: Sounds good!

PR #485, opened 2021-04-09
Add Mesh argument to InitUserMeshData

forrestglines:
## PR Summary

Adds `Mesh*` argument to `InitUserMeshData` which points to the mesh calling `InitUserMeshData`.

Allows access to that user hooks like `EnrollUserExplicitSourceFunction` during problem generation.

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted


forrestglines:
@par-hermes format

jlippuner (approved):
Very useful, thanks!

src/mesh/mesh.hpp
```diff
@@ -170,6 +170,12 @@ class Mesh {
 
   void OutputMeshStructure(const int dim, const bool dump_mesh_structure = true);
 
+  // Typically called within InitUserMeshData during problem generation
+  void EnrollUserRefinementCondition(AMRFlagFunc amrflag);
+  void EnrollUserMeshGenerator(CoordinateDirection dir, MeshGenFunc my_mg);
+  void EnrollUserExplicitSourceFunction(SrcTermFunc my_func);
+  void EnrollUserTimeStepFunction(TimeStepFunc my_func);
```
Yurlungur, line 177: Are these functions meaningfully hooked up at the moment? Should some of them be removed? For example, I'm not sure how source functions would fit within a `mesh` object now.

Yurlungur (approved):

src/mesh/mesh.hpp
forrestglines: Now that I look more into it, they might not be fully hooked up. If we decide to use these hooks, we'll hook them up in Parthenon as we start using them soon in AthenaPK. Otherwise, we might decide to remove them and go through the `StateDescriptor` objects instead.

pgrete (commented):
Here's what I think about the function touched in this PR:
```c++
InitUserMeshData(this, pin);
```
Access to the Mesh is/will be useful as it allows to also update data in the package/params.
```c++
  void EnrollUserTimeStepFunction(TimeStepFunc my_func);
  void EnrollUserRefinementCondition(AMRFlagFunc amrflag);
  void EnrollUserExplicitSourceFunction(SrcTermFunc my_func);
```
The required underlying functionality in Parthenon is current not used/implemented.
Also, I think this should be done (and already is done in AthenaPK) at the package level rather than the Mesh level.
-> Remove interface and functionality form Parthenon.
```c++
  void EnrollUserMeshGenerator(CoordinateDirection dir, MeshGenFunc my_mg);
```
There seem to be quite a few reference to that overall machinery but I'm not sure how much this is actually used (e.g., the default ones). May be worth (at some point) to double check if this still works or not and then potentially remove it.

What do people think?

forrestglines:
> What do people think?

I lean towards removing the old hooks and relying on the new `StateDescriptor` interface. The one issue would be porting old problem generators that use those hooks, but that isn't a huge code change.

Yurlungur:
I would argue for going through `StateDescriptor` at the package level as much as possible and removing the machinery otherwise.

The mesh generator may be the exception. I am not familiar with how that works in `Athena++` and whether or not it's even hooked up in parthenon is not obvious to me.

pgrete:
Now that I think about this again, I think what we want/need is probably not a `InitUserMeshData`, but a `InitUserPackageData`.
From a conceptual point of view there should be no function changing anything in the `Mesh` as this would imply that something downstream is not properly/cleanly handled by a package (which, then, would imply that the package is not properly encapsulated).

Yurlungur:
Isn't all package data essentially `InitUserPackageData`?

pgrete:
> Isn't all package data essentially `InitUserPackageData`?

Hehe, I guess that depends on the terminology.
The more specific term is probably `InitProblemPackageData`

The "`User`" above referred to that this functionality (in the original `Athena++` code) could be enrolled by the "Athena++ User" by  specifying functions in a custom problem generator (but that not necessarily override the behavior of the driver/hydro/mhd integration).
Here, this would be kind of a custom/user callback for package initialization.
Having said that, this functionality may not need to (should not?) live in Parthenon, but could be custom callback in the downstream code.

Yurlungur:
I see

> Having said that, this functionality may not need to (should not?) live in Parthenon, but could be custom callback in the downstream code.

Thanks to the split in `Parthenon::Init` one can also do this in the downstream application's main function and register things that way.

pgrete:
Closing as this has been addressed with #498 and in downstream package functionality.

PR #486, opened 2021-04-13
Overhaul HDF5 output

jlippuner:
## PR Summary

This PR is preparing the HDF5 output facility for the sparse variable update where sparse variables are not allocated on all blocks. Therefore, the list of variables that should be written to an HDF5 file (either an output or restart file) can now be different from meshblock to meshblock. Since we write one global HDF5 output file, we thus need to first assemble a global list of variables to be written to the output file. Additionally, some extra output is needed for sparse variables to record if it is allocated or not an a particular meshblock.

Sparse variables are written to the HDF5 output files as a dense N-dimensional array, just like dense variables. The blocks that don't have the sparse variable allocated just write zeros. This will make it easier to read the output files into external tools (e.g. visualization) that expect dense fields. To reduce the output file size, this PR enables writing compressed HDF5 files.

This PR makes the following changes:
- Combine writing restart and HDF5 output files into one code path, namely `PHDF5Output` and remove the `RestartOutput` class
- Assemble a global list of variables (sparse and dense) written to the output file, which is needed to collectively agree on the list and order of sparse fields for writing the associated sparse metadata to the HDF5 file (see next point)
- Write a dataset called `SparseInfo` that is a 2D array of bools that record whether a particular sparse field is allocated on a particular block
- Add the option to write compressed HDF5 files. This is enabled by default, but it requires a very recent version of HDF5 that has a bug fix which would otherwise crash when trying to write an HDF5 file with compression and parallel I/O. CMake checks the HDF5 version and won't allow configuring Parthenon if the bug would be hit (it will display a useful error message on how to resolve the issue, namely use a more recent version of HDF5 or disable compression).

Note: This PR will not contain a test for writing sparse fields that are not allocated on all blocks, because it's currently not possible to allocate sparse fields only on certain blocks. Such a test will be part of the sparse update PR that will allow allocating sparse fields only on certain blocks.

This PR includes a small update tho the `CellVariable` interface to query it whether it is a sparse field. This is just an interface update for now without the underlying full implementation, because we need the interface update in the HDF5 output infrastructure. The full implementation of the sparse stuff will be done in the sparse update PR #383.

In my very limited test, I found that enabling HDF5 compression not only drastically reduces the file size, but also improves I/O performance considerably. But more testing should be done with more realistic output.

Still to do:
- [x] Clean up HDF5 code, and especially remove macros
- [ ] ~~Implementing reading sparse information from restart files (this may have to be done in the sparse update PR since it might require more of the sparse infrastructure)~~ This will have to be done in #383 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


Yurlungur:
One concern I might have about HDF5 compression is that it might reduce resiliency. What happens if the I/O fails partway?

JoshuaSBrown:
> One concern I might have about HDF5 compression is that it might reduce resiliency. What happens if the I/O fails partway?

How would that be any worse than if the file is not compressed? I would think the benefits of reducing write times etc would actually be an improvement in resiliency.

Yurlungur:
> > One concern I might have about HDF5 compression is that it might reduce resiliency. What happens if the I/O fails partway?
> 
> How would that be any worse than if the file is not compressed? I would think the benefits of reducing write times etc would actually be an improvement in resiliency.

My experience in the past has been that an I/O failure with compression can destroy the whole file, whereas an I/O failure without may only destroy an individual dataset. I.e., if we care about trying to recover a file, compression makes that harder.

pgrete:
While you're at it I'd like to add an item to the wish list:
Add an option/switch to write the data itself in single precision (e.g., for "analysis" outputs that cannot be used for restarts -- or at least for restarts that *try* to be bitwise identical).

jlippuner:
> > > One concern I might have about HDF5 compression is that it might reduce resiliency. What happens if the I/O fails partway?
> > 
> > 
> > How would that be any worse than if the file is not compressed? I would think the benefits of reducing write times etc would actually be an improvement in resiliency.
> 
> My experience in the past has been that an I/O failure with compression can destroy the whole file, whereas an I/O failure without may only destroy an individual dataset. I.e., if we care about trying to recover a file, compression makes that harder.

When you say failure, are you thinking of some catastrophic hardware or external failure (node or filesystem crashing) or are you thinking about a software failure in the Parthenon I/O routines? If the later, the output test we have should give us reasonable confidence that Parthenon won't have failures in writing HDF5 files.

jlippuner:
This is ready for review now

CMakeLists.txt
```diff
@@ -54,6 +54,12 @@ option(TEST_INTEL_OPTIMIZATION "Test intel optimization and vectorization" OFF)
 option(TEST_ERROR_CHECKING "Enables the error checking unit test. This test will FAIL" OFF)
 option(CODE_COVERAGE "Enable code coverage reporting" OFF)
 
+# HDF5 compression level option, value between 0 and 9
+set(PARTHENON_HDF5_COMPRESSION_LEVEL 5 CACHE STRING "Compression level for HDF5 output and restart files, valid range is 0-9, default is 5, set to 0 to turn off compression")
+if(NOT PARTHENON_HDF5_COMPRESSION_LEVEL MATCHES "^[0-9]$")
+  message(FATAL_ERROR "PARTHENON_HDF5_COMPRESSION_LEVEL must be a number between 0 and 9")
+endif()
+
```
forrestglines: So compression levels are set at compile time? If I make a restart file with one compression level then recompile with a different compression, then I can assume that the new binary won't be able to restart with the old compression level?

src/interface/params.hpp
```diff
@@ -108,6 +112,40 @@ class Params {
...
+  void WriteAllToHDF5(const std::string &prefix, const HDF5::H5G &group) const {
+    WriteTVecToHDF5<bool>(prefix, group);
+    WriteTVecToHDF5<int32_t>(prefix, group);
+    WriteTVecToHDF5<int64_t>(prefix, group);
+    WriteTVecToHDF5<uint32_t>(prefix, group);
+    WriteTVecToHDF5<uint64_t>(prefix, group);
+    WriteTVecToHDF5<float>(prefix, group);
+    WriteTVecToHDF5<double>(prefix, group);
+    WriteTVecToHDF5<std::string>(prefix, group);
+  }
```
forrestglines: I see how this does work, but should `WriteToHDF5` give feedback if you try to write a parameters to HDF5 but with the wrong type? In this current implementation, if you called `WriteToHDF5` somewhere else with the wrong type, you would silently write nothing to file. Changing it to an error though would break `WriteAllToHDF5` though.

Would to make sense in the context of the rest of the code to return an error in `WriteToHDF5` if there's a type mismatch then use a switch statement on `myTypes_.at(key)` to call the correct template of `WriteTVecToHD?

src/outputs/parthenon_hdf5.cpp
```diff
@@ -192,436 +365,652 @@ void PHDF5Output::genXDMF(std::string hdfFile, Mesh *pm, SimTime *tm) {
...
+        variableNames.push_back(label);
+      }
+    } else {
+      numVariables.push_back(1);
+      variableNames.push_back(vi.label);
+    }
+  }
+
+  HDF5WriteAttribute("NumVariables", numVariables, info_group);
+  HDF5WriteAttribute("VariableNames", variableNames, info_group);
```
forrestglines:
```suggestion
  // NumComponents - number of components within each dataset
  std::vector<size_t> numComponents;
  // ComponentNames - Names of components within each dataset
  std::vector<std::string> componentNames;

  for (const auto &vi : all_unique_vars) {
    const auto &component_labels = vi.component_labels;

    if (component_labels.size() > 0) {
      numComponents.push_back(component_labels.size());
      for (const auto &label : component_labels) {
        componentNames.push_back(label);
      }
    } else {
      numComponents.push_back(1);
      componentNames.push_back(vi.label);
    }
  }

  HDF5WriteAttribute("NumComponents", numComponents, info_group);
  HDF5WriteAttribute("ComponentNames", componentNames, info_group);
```
This naming convention I introduced is a bit confusing, especially `VariableNames`. What do you think about renaming these as "Components" as they're named in the metadata and elsewhere in the code?

src/parthenon_manager.cpp
```diff
@@ -222,6 +222,9 @@ void ParthenonManager::RestartPackages(Mesh &rm, RestartReader &resfile) {
   size_t nCells = bsize[0] * bsize[1] * bsize[2];
 
   // Get list of variables, assumed same for all blocks
+
+  // TODO(JL) this doesn't work anymore, will update this in sparse update
+
```
forrestglines: How much is broken? Just when using sparse variables?

forrestglines (commented):
Looks good, much cleaner to write to HDF5 now. I had a few suggestions and questions but I think we can merge this in soon. I'll verify that my `yt` frontend still works and tweak it if needed.

CMakeLists.txt
jlippuner: Good question. Fortunately, HDF5 handles all the compression and decompression. This flag only specifies with what compression data is written. It can still read data written with any other compression level (or no compression). So if you compile with compression level 5, write a restart dump, then compile with level 0 or anything else, you can still use the previous restart dump. Parthenon never sees the compressed data, because it's all handled by the HDF5 library.

src/interface/params.hpp
jlippuner: That's a good point. I think I should rename `WriteToHDF5` and `WriteTVecToHDF5`, because it seems confusing. These functions don't write specific parameters you specify, instead, they write all parameters that match the type you give it. So if you ask it to write all, say, `int`'s and there are no `int`'s, that's fine. I don't think there should be a warning for that.

Also, I'm thinking of `WriteToHDF5` and `WriteTVecToHDF5` as purely internal functions that should only be used by `WriteAllToHDF5`. The reason they are separate functions is to reduce code duplication.

Let me see if I can give them a better name and encapsulate them better, so they won't be available where they shouldn't be used.

src/parthenon_manager.cpp
jlippuner: Nothing is broken right now, even when using current sparse variables. But with the sparse update, you will be able to write true sparse variables in a restart file, but won't be able to read them back correctly. So this needs to be updated in that PR, but it shouldn't make a difference as long as we don't have true sparse variables yet. I'll change this to "this won't work" instead of "doesn't work anymore".

src/outputs/parthenon_hdf5.cpp
jlippuner: Good idea. If you don't mind, I think I'll change it to `num_components` and `component_names` instead to go with the snake case style used in that file

src/interface/params.hpp
forrestglines: > Also, I'm thinking of `WriteToHDF5` and `WriteTVecToHDF5` as purely internal functions that should only be used by `WriteAllToHDF5`. The reason they are separate functions is to reduce code duplication.

I realized that later in the review - I think you can just make those two `private`, that would be fine

src/outputs/parthenon_hdf5.cpp
forrestglines: For naming the C++ variables? Yes, you should do that.

forrestglines:
I just checked, nothing in this commit breaks the existing `yt` frontend. I'll have to update it though when `VariableNames` changes to `ComponentNames`

src/interface/params.hpp
jlippuner: I've made them `private` and given them a much more descriptive name

src/outputs/parthenon_hdf5.cpp
jlippuner: Done. Also updated gold to v7 and yes, the name in the HDF5 metadata changed

scripts/python/phdf_diff.py
```diff
@@ -66,89 +66,116 @@ def addPath():
...
+    return x if isinstance(x, np.ndarray) else [x]
+
+def compare_attributes(dict0, dict1):
+    keys0 = set(dict0.keys())
+    keys1 = set(dict1.keys())
+    union = keys0.union(keys1)
+    intersect = keys0.intersection(keys1)
+
+    # keys that only show up in one set
+    diff_keys = [k for k in union if (k not in keys0) or (k not in keys1)]
```
JoshuaSBrown: This is nicely written!

JoshuaSBrown: How come you don't just search to see if the keys are not in the intersection won't that give you the keys that are not in both?

JoshuaSBrown: Or you could just use this https://www.programiz.com/python-programming/methods/set/symmetric_difference

jlippuner: Oh cool, I didn't know about `symmetric_difference`, will use it

src/interface/variable.hpp
```diff
@@ -48,9 +48,11 @@ class CellVariable {
  public:
   /// Initialize a 6D variable
   CellVariable<T>(const std::string &label, const std::array<int, 6> dims,
-                  const Metadata &metadata)
+                  const Metadata &metadata, int sparse_id = -1)
```
JoshuaSBrown, line 51: Can you make sparse_id const? Actually const reference would be better.

src/interface/variable.hpp
```diff
@@ -76,6 +78,11 @@ class CellVariable {
   ///< retrieve metadata for variable
   Metadata metadata() const { return m_; }
 
+  /// Get Sparse ID (-1 if not sparse)
+  int GetSparseID() const { return sparse_id_; }
+
+  bool IsSparse() const { return sparse_id_ >= 0; }
```
JoshuaSBrown: Can you make these noexcept? Or just inline them?

src/mesh/mesh.cpp
```diff
@@ -582,14 +582,14 @@ Mesh::Mesh(ParameterInput *pin, ApplicationInput *app_in, RestartReader &rr,
...
 
   // All ranks read HDF file
-  nbnew = rr.GetAttr<int>("Mesh", "nbnew");
-  nbdel = rr.GetAttr<int>("Mesh", "nbdel");
-  nbtotal = rr.GetAttr<int>("Mesh", "nbtotal");
-  root_level = rr.GetAttr<int32_t>("Mesh", "rootLevel");
+  nbnew = rr.GetAttr<int32_t>("Info", "NBNew");
+  nbdel = rr.GetAttr<int32_t>("Info", "NBDel");
+  nbtotal = rr.GetAttr<int32_t>("Info", "NumMeshBlocks");
+  root_level = rr.GetAttr<int32_t>("Info", "RootLevel");
```
JoshuaSBrown: Nice! I wish we could use enums here instead but alas that's an issue on its own.

src/mesh/mesh.cpp
```diff
@@ -582,14 +582,14 @@ Mesh::Mesh(ParameterInput *pin, ApplicationInput *app_in, RestartReader &rr,
...
 
-  auto bc = rr.ReadAttr1DReal("Mesh", "bc");
+  auto bc = rr.GetAttrVec<std::string>("Info", "BoundaryConditions");
```
JoshuaSBrown: can bc be a const reference?

src/mesh/mesh.cpp
```diff
@@ -613,23 +613,23 @@ Mesh::Mesh(ParameterInput *pin, ApplicationInput *app_in, RestartReader &rr,
...
   EnrollBndryFncts_(app_in);
 
-  std::vector<Real> bounds = rr.ReadAttr1DReal("Mesh", "bounds");
-  mesh_size.x1min = bounds[0];
-  mesh_size.x2min = bounds[1];
-  mesh_size.x3min = bounds[2];
-  mesh_size.x1max = bounds[3];
-  mesh_size.x2max = bounds[4];
-  mesh_size.x3max = bounds[5];
+  std::vector<Real> grid_dim = rr.GetAttrVec<Real>("Info", "RootGridDomain");
```
JoshuaSBrown: Can grid_dim be made const reference? At least const I think doesn't look like it is being changed.

src/mesh/mesh.cpp
```diff
@@ -613,23 +613,23 @@ Mesh::Mesh(ParameterInput *pin, ApplicationInput *app_in, RestartReader &rr,
...
+
+  mesh_size.x3min = grid_dim[6];
+  mesh_size.x3max = grid_dim[7];
+  mesh_size.x3rat = grid_dim[8];
 
   // initialize
   loclist = std::vector<LogicalLocation>(nbtotal);
 
-  auto blockSize = rr.ReadAttr1DI32("Mesh", "blockSize");
+  auto blockSize = rr.GetAttrVec<int32_t>("Info", "MeshBlockSize");
```
JoshuaSBrown: Same here const reference if possible.

src/mesh/mesh.cpp
```diff
@@ -758,7 +758,7 @@ Mesh::Mesh(ParameterInput *pin, ApplicationInput *app_in, RestartReader &rr,
   int nbe = nbs + nb - 1;
 
   // read in xmin from file
-  auto xmin = rr.ReadDataset<double>("/Blocks/xmin");
+  auto xmin = rr.ReadDataset<Real>("/Blocks/xmin");
```
JoshuaSBrown: Maybe I'm blind but where is xmin being used?

src/outputs/outputs.cpp
```diff
@@ -235,6 +224,26 @@ Outputs::Outputs(Mesh *pm, ParameterInput *pin, SimTime *tm) {
       // read cartesian mapping option
       op.cartesian_vector = false;
 
+      // read single precision output option
+      bool is_hdf5_output = (op.file_type.compare("rst") == 0) ||
+                            (op.file_type.compare("ath5") == 0) ||
+                            (op.file_type.compare("hdf5") == 0);
```
JoshuaSBrown: I would write this differently,

```
std::unordered_set<std::string> hdf5_extensions = {"rst", "ath5", "hdf5"};
const bool is_hdf5_output = hdf5_extensions.count(op.file_type) == 1;
```
Looks like in c++20 set gets a contains keyword that would be nicer but guess we aren't there.

src/outputs/outputs.cpp
```diff
@@ -252,20 +261,21 @@ Outputs::Outputs(Mesh *pm, ParameterInput *pin, SimTime *tm) {
         pnew_type = new FormattedTableOutput(op);
       } else if (op.file_type.compare("vtk") == 0) {
         pnew_type = new VTKOutput(op);
-      } else if (op.file_type.compare("rst") == 0) {
-        pnew_type = new RestartOutput(op);
-        num_rst_outputs++;
-      } else if (op.file_type.compare("ath5") == 0 || op.file_type.compare("hdf5") == 0) {
+      } else if (is_hdf5_output) {
+        bool restart = (op.file_type.compare("rst") == 0);
```
JoshuaSBrown: make restart const. 

Also I think that comparing strings with == is a little bit more readable. Is there a reason you use compare? 

```
op.file_type == "rst"
```

src/outputs/outputs.cpp
```diff
@@ -252,20 +261,21 @@ Outputs::Outputs(Mesh *pm, ParameterInput *pin, SimTime *tm) {
...
+        if (restart) {
+          num_rst_outputs++;
+        }
 #ifdef HDF5OUTPUT
-        pnew_type = new PHDF5Output(op);
+        pnew_type = new PHDF5Output(op, restart);
```
JoshuaSBrown, line 298: Can we change this to a unique_ptr?

src/outputs/outputs.hpp
```diff
@@ -177,44 +179,24 @@ class VTKOutput : public OutputType {
...
 //----------------------------------------------------------------------------------------
 //! \class PHDF5Output
-//  \brief derived OutputType class for Athena HDF5 files
+//  \brief derived OutputType class for Athena HDF5 files or restart dumps
 
 class PHDF5Output : public OutputType {
  public:
   // Function declarations
-  explicit PHDF5Output(const OutputParameters &oparams) : OutputType(oparams) {}
+  PHDF5Output(const OutputParameters &oparams, bool restart)
```
JoshuaSBrown, line 191: Make bool const reference

src/outputs/parthenon_hdf5.cpp
```diff
@@ -15,17 +15,193 @@
...
+namespace HDF5 {
+
+// template specializations for std::string
+template <>
+void HDF5WriteAttribute(const std::string &name, const std::vector<std::string> &values,
+                        hid_t location) {
+  std::vector<const char *> char_ptrs(values.size());
+  for (size_t i = 0; i < values.size(); ++i) {
+    char_ptrs[i] = values[i].c_str();
+  }
```
JoshuaSBrown, line 48: I would use a range based for loop, secondly I would call reserve instead of initializing, then I would use emplace.

src/outputs/parthenon_hdf5.cpp
```diff
@@ -15,17 +15,193 @@
...
+} // namespace HDF5
+
+namespace {
+
+std::string PackStrings(const std::vector<std::string> &strs, char delimiter) {
+  std::string pack;
+  for (const auto &s : strs) {
+    pack += s + delimiter;
+  }
+  return pack;
```
JoshuaSBrown: It makes no sense to me that the std libary does not have a function that will do this... ah.

JoshuaSBrown: I think the project might benefit from moving stuff like this into a string helpers file. String manipulation is likely to be needed in multiple places.

src/interface/variable.hpp
jlippuner: For small data types, pass by value is preferable and `const` is not needed, because we get a copy.

src/interface/variable.hpp
jlippuner: They should be inlined automatically, but I guess adding `inline` can't hurt.

src/mesh/mesh.cpp
jlippuner: I'll make it const, but reference seems confusing, because `GetAttrVec` returns a value, not a reference. It can't return a reference because it just creates the vector and returns it, it doesn't store the vector.

src/mesh/mesh.cpp
jlippuner: See previous comment.

src/mesh/mesh.cpp
jlippuner: Same as above

src/mesh/mesh.cpp
jlippuner: Good catch, it doesn't seem to be used

src/outputs/outputs.cpp
jlippuner: I don't think using a `set` or `undordered_set` makes sense here. That's too much overhead to check 3 possibilities. If the number of options explodes, we can revisit this. But I am removing the `.compare`. Not sure why that was used...

src/outputs/outputs.cpp
jlippuner: I agree. I don't know who wrote the initial version using `compare`.

src/outputs/outputs.cpp
jlippuner: Alas, no. This is a custom-made linked list and the pointers get passed around and manipulated a lot. At least it seems they all get properly freed in the destructor of the `Outputs` class. But I agree that this is not pretty. We should be using an STL container here. This should be relatively straightforward to fix, but it's beyond the scope of this PR. Might be a good "first issue".

src/outputs/outputs.hpp
jlippuner: POD (int, bool, double, etc.) should be passed by value, not reference. Passing by reference can be slightly less performant since it requires two memory reads (first reading the pointer, then following it) and a pointer is actually a bigger word than a bool/int.

src/outputs/parthenon_hdf5.cpp
jlippuner: Normally, I agree that it's better to use range based for loops. However, in this case, we need the index so we can assign the relevant entry in `char_ptrs`. You're right that we could use `emplace_back`. However, since a `const char *` is default-constructible, we don't have to and in this case, using `reserve` and `emplace_back` seems too convoluted and would even add an extra line. We could use `std::transform`, but I think the plain index based for loop is the most readable and easy to understand. And there are no performance reasons to not use it.

src/outputs/parthenon_hdf5.cpp
jlippuner: Good idea. I've moved into a separate utils file and combined it with the string trimming utils.

src/outputs/outputs.hpp
JoshuaSBrown: Ah, thanks for the correction, I would just make it const then.

src/mesh/mesh.cpp
JoshuaSBrown: Sounds good!

src/interface/variable.hpp
JoshuaSBrown: Yep, yep, your right. I saw your other comment, will need to make sure I remember that.

jlippuner:
Thanks @JoshuaSBrown for your helpful comments. I've addressed or responded to all of them. Please let me know if you disagree with any of my responses.

@forrestglines It appears your approval was reset after a few minor updates (mostly stylistic). Would you mind approving again?

src/outputs/parthenon_hdf5.cpp
```diff
@@ -15,17 +15,193 @@
...
+    if (tab == curr) {
+      unpack.push_back("");
+    } else {
+      unpack.push_back(std::string(curr, tab - curr));
+      curr = tab + 1;
+    }
+  }
+
+  return unpack;
+}
```
JoshuaSBrown: Wouldn't it be a little simpler to use something like this:

```
std::string input = "abc,def,ghi";
std::istringstream ss(input);
std::vector<std::string> unpack;
std::string token;
while(std::getline(ss, token, ',')) {
    unpack.emplace_back(token);
}
```

You could just check the last character to see if it is a delimiter.

src/outputs/parthenon_hdf5.cpp
```diff
@@ -15,17 +15,193 @@
...
+  std::vector<std::string> component_labels; // may be empty
+  int vlen;
+  bool is_sparse;
+  bool is_vector;
+
+ private:
+  VarInfo() = default;
+
+ public:
+  static VarInfo Decode(const std::string &compact_labels, int info_code) {
```
JoshuaSBrown: make info_code const?

src/outputs/parthenon_hdf5.cpp
```diff
@@ -15,17 +15,193 @@
...
+    VarInfo res;
+
+    // unpack compact_labels
+    auto labels = UnpackStrings(compact_labels, '\t');
```
JoshuaSBrown: labels -> const?

src/outputs/parthenon_hdf5.cpp
```diff
@@ -175,12 +349,11 @@ void PHDF5Output::genXDMF(std::string hdfFile, Mesh *pm, SimTime *tm) {
     dims[2] = nx2;
     dims[3] = nx1;
     dims[4] = 1;
-    for (auto &v : ciX.vars) {
-      const int vlen = v->GetDim(4);
+    for (auto &vinfo : var_list) {
```
JoshuaSBrown: I think vinfo can be made const?

src/outputs/parthenon_hdf5.hpp
```diff
@@ -11,14 +11,14 @@
 //========================================================================================
 #ifndef OUTPUTS_PARTHENON_HDF5_HPP_
 #define OUTPUTS_PARTHENON_HDF5_HPP_
-// Definitions common to parthenon restart and parthenon output for HDF5
 
-// options for building
-#include "config.hpp"
+#ifndef HDF5OUTPUT
+#error "parthenon_hdf5.hpp requires HDF5 output to be enabled"
```
JoshuaSBrown, line 16: Oh, I didn't know there was an error pragma, that's neat.

src/outputs/parthenon_hdf5.cpp
jlippuner: That's a good idea, thanks!

src/outputs/parthenon_hdf5.cpp
```diff
@@ -192,436 +330,652 @@ void PHDF5Output::genXDMF(std::string hdfFile, Mesh *pm, SimTime *tm) {
...
+    buffer_lengths[Globals::my_rank * 2 + 1] = static_cast<int>(code_buffer.size());
+
+    PARTHENON_MPI_CHECK(MPI_Allgather(MPI_IN_PLACE, 2, MPI_INT, buffer_lengths.data(), 2,
+                                      MPI_INT, MPI_COMM_WORLD));
+
+    // now do an Allgatherv combining label_buffer and vlen_buffer from all ranks
+    std::vector<int> label_lengths(Globals::nranks, 0);
+    std::vector<int> label_offsets(Globals::nranks, 0);
+    std::vector<int> code_lengths(Globals::nranks, 0);
+    std::vector<int> code_offsets(Globals::nranks, 0);
```
JoshuaSBrown, line 756: You are assigning 0's then overwriting them, I would use reserve here with push back instead.

JoshuaSBrown (approved):
A nice piece of work, definitely an improvement, thanks for the refactor. I had a several non blocking comments which I think you have addressed anyway. As long as the ci is passing I think this is good to go.

src/outputs/parthenon_hdf5.cpp
jlippuner: This is a style question. I prefer to not make function arguments that are passed by value `const`. Because the function has a copy of the data anyway, it cannot modify the value outside. So adding const doesn't really add anything and it just makes the function signature longer and a little less readable.

src/outputs/parthenon_hdf5.cpp
jlippuner: Done

src/outputs/parthenon_hdf5.cpp
jlippuner: Good catch

src/outputs/parthenon_hdf5.cpp
jlippuner: Since the size is fixed and known ahead of time, I think this is slightly more straightforward and readable than using `reserve` and `push_back`

src/outputs/parthenon_hdf5.cpp
JoshuaSBrown: I disagree on this point, it shows intent. Anytime you see const in the function signature you know it will be the same wherever you see it in the function body, but I leave it at your discretion, it is mostly preference.

jlippuner: Fair point. I guess it boils down to stylistic preference

Yurlungur:
@jlippuner my apologies on taking some time to get to this. I'll have a review for you by the end of the week.

jlippuner:
> @jlippuner my apologies on taking some time to get to this. I'll have a review for you by the end of the week.

Thanks! I know it takes time to go through a big PR.

CMakeLists.txt
```diff
@@ -160,6 +166,23 @@ if (NOT PARTHENON_DISABLE_HDF5)
...
   endif()
 
+  if ((PARTHENON_HDF5_COMPRESSION_LEVEL GREATER 0) AND ENABLE_MPI)
+    # make sure we use an HDF5 version that doesn't have a known bug when used with compression and
+    # parallel I/O. See:
+    # https://forum.hdfgroup.org/t/crash-when-writing-parallel-compressed-chunks/6186
+    #
+    # It appears that the bug is fixed in 1.10.7 and 1.12.0, assume that later versions are ok as
+    # well, so we check that we have at least 1.10.7
+    if (HDF5_VERSION VERSION_LESS "1.10.7")
```
Yurlungur, line 171: I appreciate that you support lower HDF5 versions if you disable compression or MPI! :+1:

CMakeLists.txt
```diff
@@ -54,6 +54,12 @@ option(TEST_INTEL_OPTIMIZATION "Test intel optimization and vectorization" OFF)
 option(TEST_ERROR_CHECKING "Enables the error checking unit test. This test will FAIL" OFF)
 option(CODE_COVERAGE "Enable code coverage reporting" OFF)
 
+# HDF5 compression level option, value between 0 and 9
+set(PARTHENON_HDF5_COMPRESSION_LEVEL 5 CACHE STRING "Compression level for HDF5 output and restart files, valid range is 0-9, default is 5, set to 0 to turn off compression")
+if(NOT PARTHENON_HDF5_COMPRESSION_LEVEL MATCHES "^[0-9]$")
+  message(FATAL_ERROR "PARTHENON_HDF5_COMPRESSION_LEVEL must be a number between 0 and 9")
+endif()
```
Yurlungur: Here's a thought. Does compression need to be set at compile time? or could it be set at run time?

example/advection/parthenon_app_inputs.cpp
```diff
@@ -117,18 +117,18 @@ void UserWorkAfterLoop(Mesh *mesh, ParameterInput *pin, SimTime &tm) {
...
           Real ref_val;
-          if (profile.compare("wave") == 0) {
+          if (profile == "wave") {
             Real x =
                 cos_a2 * (pmb->coords.x1v(i) * cos_a3 + pmb->coords.x2v(j) * sin_a3) +
                 pmb->coords.x3v(k) * sin_a2;
             Real sn = std::sin(k_par * x);
             ref_val = 1.0 + amp * sn * vel;
-          } else if (profile.compare("smooth_gaussian") == 0) {
+          } else if (profile == "smooth_gaussian") {
```
Yurlungur, line 126: Didn't we change this to be device-side?

tst/regression/test_suites/advection_outflow/advection_outflow.py
```diff
@@ -40,7 +40,6 @@ def Analyse(self, parameters):
 
         delta = compare([
             'outflow.out0.00001.phdf',
-            parameters.parthenon_path + '/tst/regression/gold_standard/outflow.out0.00001.phdf'],
-            check_metadata=False)
+            parameters.parthenon_path + '/tst/regression/gold_standard/outflow.out0.00001.phdf'])
```
Yurlungur, line 43: I think this test shouldn't check metadata. Why are you setting `check_metadata=True`?

scripts/python/packages/parthenon_tools/parthenon_tools/phdf_diff.py
```diff
@@ -66,89 +66,116 @@ def addPath():
     #sys.path.insert(0,myPath+'/../vis/python')
     #sys.path.insert(0,myPath+'/vis/python')
 
-def compare_metadata( f0, f1, tol=1.0e-12):
+def ensure_list(x):
```
Yurlungur, line 69: This is a substantial cleanup of this file. Nice!

src/interface/params.hpp
```diff
@@ -24,6 +24,10 @@
 
 #include "utils/error_checking.hpp"
 
+#ifdef HDF5OUTPUT
```
Yurlungur: Was this the define we used previously? Should it be named something like `PARTHENON_HDF5_OUTPUT`?

src/interface/params.hpp
```diff
@@ -108,6 +112,44 @@ class Params {
     std::cout << std::endl;
   }
 
+#ifdef HDF5OUTPUT
+
+ private:
+  // will write all params with type T to the given HDF5 group as an attribute
+  template <typename T>
+  void WriteToHDF5AllParamsOfTypeT(const std::string &prefix,
```
Yurlungur: I don't love the `T` in this name. I don't think we need it.

src/interface/variable.hpp
```diff
@@ -48,9 +48,11 @@ class CellVariable {
  public:
   /// Initialize a 6D variable
   CellVariable<T>(const std::string &label, const std::array<int, 6> dims,
-                  const Metadata &metadata)
+                  const Metadata &metadata, int sparse_id = -1)
       : data(label, dims[5], dims[4], dims[3], dims[2], dims[1], dims[0]),
-        mpiStatus(false), m_(metadata), label_(label) {
+        mpiStatus(false), m_(metadata),
+        label_(label + (sparse_id >= 0 ? "_" + std::to_string(sparse_id) : "")),
```
Yurlungur, line 54: So using non-negative sparse id's isn't unreasonable to me. But it might be safer to use a signal nan or sparse id or something like that?

src/mesh/mesh.cpp
```diff
@@ -582,14 +582,14 @@ Mesh::Mesh(ParameterInput *pin, ApplicationInput *app_in, RestartReader &rr,
...
+  nbnew = rr.GetAttr<int32_t>("Info", "NBNew");
+  nbdel = rr.GetAttr<int32_t>("Info", "NBDel");
+  nbtotal = rr.GetAttr<int32_t>("Info", "NumMeshBlocks");
+  root_level = rr.GetAttr<int32_t>("Info", "RootLevel");
 
-  auto bc = rr.ReadAttr1DReal("Mesh", "bc");
+  const auto bc = rr.GetAttrVec<std::string>("Info", "BoundaryConditions");
   for (int i = 0; i < 6; i++) {
-    block_bcs[i] = static_cast<BoundaryFlag>(bc[i]);
+    block_bcs[i] = GetBoundaryFlag(bc[i]);
```
Yurlungur, line 592: :+1:

src/mesh/mesh.cpp
Yurlungur: Should the `int` vs `int32_t` setting here be compile-time?

src/outputs/outputs.cpp
```diff
@@ -235,8 +224,27 @@ Outputs::Outputs(Mesh *pm, ParameterInput *pin, SimTime *tm) {
...
+        op.single_precision_output = false;
+
+        if (pin->DoesParameterExist(op.block_name, "single_precision_output")) {
+          std::stringstream warn;
+          warn << "### WARNING Output option single_precision_output only applies to "
+                  "HDF5 outputs or restarts. Ignoring it for output block '"
+               << op.block_name << "'" << std::endl;
+          PARTHENON_WARN(warn);
+        }
+      }
```
Yurlungur, line 244: 👍

src/outputs/parthenon_hdf5.cpp
```diff
@@ -15,17 +15,158 @@
...
+  static constexpr int vector_flag = (1 << 21);
+
+  std::string label;
+  std::vector<std::string> component_labels; // may be empty
+  int vlen;
+  bool is_sparse;
+  bool is_vector;
+
+ private:
+  VarInfo() = default;
```
Yurlungur: Why is the default constructor private?

src/outputs/parthenon_hdf5.cpp
```diff
@@ -15,17 +15,158 @@
...
+    // unpack compact_labels
+    const auto labels = string_utils::UnpackStrings(compact_labels, '\t');
+
+    if (labels.size() == 0) {
+      std::stringstream msg;
+      msg << "### ERROR: Got no labels" << std::endl;
+      PARTHENON_FAIL(msg);
+    }
+
+    // first label in compact labe is the variable label, the rest are the component
```
Yurlungur:
```suggestion
    // first label in compact label is the variable label, the rest are the component
```

src/outputs/parthenon_hdf5.cpp
```diff
@@ -192,436 +330,652 @@ void PHDF5Output::genXDMF(std::string hdfFile, Mesh *pm, SimTime *tm) {
...
+  PARTHENON_HDF5_CHECK(H5Pset_alignment(acc_file, 524288, 262144));
+
+  PARTHENON_MPI_CHECK(MPI_Info_set(FILE_INFO_TEMPLATE, "access_style", "write_once"));
+  PARTHENON_MPI_CHECK(MPI_Info_set(FILE_INFO_TEMPLATE, "collective_buffering", "true"));
+  PARTHENON_MPI_CHECK(MPI_Info_set(FILE_INFO_TEMPLATE, "cb_block_size", "1048576"));
+  PARTHENON_MPI_CHECK(MPI_Info_set(FILE_INFO_TEMPLATE, "cb_buffer_size", "4194304"));
+
+  /* tell the HDF5 library that we want to use MPI-IO to do the writing */
+  PARTHENON_HDF5_CHECK(H5Pset_fapl_mpio(acc_file, MPI_COMM_WORLD, FILE_INFO_TEMPLATE));
+  PARTHENON_HDF5_CHECK(H5Pset_fapl_mpio(acc_file, MPI_COMM_WORLD, MPI_INFO_NULL));
```
Yurlungur, line 418: Is this the best place for this stuff? Should it be runtime or compile time set with the defaults you chose?

src/outputs/parthenon_hdf5.cpp
```diff
@@ -192,436 +330,652 @@ void PHDF5Output::genXDMF(std::string hdfFile, Mesh *pm, SimTime *tm) {
...
-            variableNames.push_back(label);
-          }
-        } else {
-          numVariables[i] = 1;
-          variableNames.push_back(ciX.vars[i]->label());
-        }
-      }
+  // now open the file
+  H5F const file = H5F::FromHIDCheck(
+      H5Fcreate(filename.c_str(), H5F_ACC_TRUNC, H5P_DEFAULT, acc_file));
```
Yurlungur: Error checking?

src/outputs/parthenon_hdf5.cpp
```diff
@@ -192,436 +330,652 @@ void PHDF5Output::genXDMF(std::string hdfFile, Mesh *pm, SimTime *tm) {
...
+
+    for (size_t b = 0; b < pm->block_list.size(); ++b) {
+      auto &pmb = pm->block_list[b];
+
+      for (int i = out_ib.s; i <= out_ib.e + offset; ++i) {
+        loc_x[idx_x++] = face ? pmb->coords.x1f(0, 0, i) : pmb->coords.x1v(0, 0, i);
+      }
+
+      for (int j = out_jb.s; j <= out_jb.e + offset; ++j) {
+        loc_y[idx_y++] = face ? pmb->coords.x2f(0, j, 0) : pmb->coords.x2v(0, j, 0);
```
Yurlungur, line 640: nice use of the ternary operator here

src/outputs/parthenon_hdf5.cpp
```diff
@@ -192,436 +330,652 @@ void PHDF5Output::genXDMF(std::string hdfFile, Mesh *pm, SimTime *tm) {
...
+
+  // local list of unique vars
+  std::set<VarInfo> all_unique_vars;
+  for (auto &pmb : pm->block_list) {
+    auto ci = get_MeshBlockDataIterator(pmb);
+    for (auto &v : ci.vars) {
+      VarInfo vinfo(v);
+      all_unique_vars.insert(vinfo);
+    }
+  }
```
Yurlungur, line 723: So I'm not sure which is better, but an alternate approach to looping over meshblocks would be to query the state descriptor, which will be identical between meshblocks and contain relevant variable information.

There's a package called `parthenon::resolved_state`, which contains a unique list of all variables:
https://github.com/lanl/parthenon/blob/develop/src/interface/state_descriptor.cpp#L333

Yurlungur: This approach would also save you an allgather. But it would require parsing the variable labels kind of "by hand"

src/outputs/parthenon_hdf5.cpp
```diff
@@ -192,436 +330,652 @@ void PHDF5Output::genXDMF(std::string hdfFile, Mesh *pm, SimTime *tm) {
...
+          auto v_h = v->data.GetHostMirrorAndCopy();
+          for (int k = out_kb.s; k <= out_kb.e; ++k) {
+            for (int j = out_jb.s; j <= out_jb.e; ++j) {
+              for (int i = out_ib.s; i <= out_ib.e; ++i) {
+                for (int l = 0; l < vlen; ++l) {
+                  tmpData[index++] = static_cast<OutT>(v_h(l, k, j, i));
+                }
+              }
+            }
           }
```
Yurlungur, line 892: I don't love this loop being done on the host. I wonder if there's a way to do it on device. We could do the index math ahead of time, then in a kernel copy to a temporary view, that's then copied to host in a single htod instad of nvars htods?

src/parthenon_manager.cpp
```diff
@@ -255,8 +259,18 @@ void ParthenonManager::RestartPackages(Mesh &rm, RestartReader &resfile) {
...
+          // Note index l transposed to interior
+          for (int k = out_kb.s; k <= out_kb.e; ++k) {
+            for (int j = out_jb.s; j <= out_jb.e; ++j) {
+              for (int i = out_ib.s; i <= out_ib.e; ++i) {
+                for (int l = 0; l < vlen; ++l) {
+                  v_h(l, k, j, i) = tmp[index++];
+                }
+              }
+            }
+          }
```
Yurlungur, line 272: see my comment above about host-mirrors and kernels. Less important for loading restarts than it is for writing output though.

Yurlungur (approved):
Overall looks good. This was clearly a major effort. Thanks for going through it, @jlippuner ! It's a large enough MR that I don't think a review will catch bugs, but I trust you've run the regression tests through their paces? Have we tested the case of a restart with independent sparse variables?

Some non-blocking comments below. My biggest hangups are on performance, which can be addressed potentially---and may not need to be if the code is already fast.

Yurlungur:
This is a pretty big change so it would be good if someone from the AthenaPK team took a look. @pgrete @forrestglines @jmstone

CMakeLists.txt
jlippuner: The compression does not need to be set at compile time. It can be set at run-time (in fact, it could be changed during runtime). The reason I made it a compile time option is so that we can make sure the version of HDF5 used doesn't have the known bug. However, I could make that option simply enable/disable compression but not set the compression level and then we could set the compression level through the input file (individually for each HDF5 output/restart file). If one tries to use compression with a build that has compression disabled, we can generate an error or warning.

I think that's actually a pretty good idea. I'll go ahead and make that change.

tst/regression/test_suites/advection_outflow/advection_outflow.py
jlippuner: Why don't we want to check metadata? I assumed setting `check_metadata=False` here was a relic from trying to make the test pass with an outdated gold standard. I've updated the gold standard files and so the test passes now with checking metadata.

src/interface/variable.hpp
jlippuner: Good point. We can't use NaN because it's an int. But I'll make a static constant `invalid_sparse_id` (probably `-INT_MAX` or something like that). Then we could also add support for negative sparse ids. Why not :)

example/advection/parthenon_app_inputs.cpp
jlippuner: Will be merged in if that change is in develop already

src/interface/params.hpp
jlippuner: Agreed

src/interface/params.hpp
jlippuner: The only other constant was `PARTHENON_DISABLE_HDF5`. But the CMake build system uses `ENABLE_HDF5` internally. I've changed `config.hpp.in` to use that directly, instead of defining `HDF5OUTPUT` when `ENABLE_HDF5` is on. This removes a layer of indirection and makes the interaction between CMake and the C++ code easier to follow, I think. All references to `HDF5OUTPUT` are now replaced with simply `ENABLE_HDF5`.

src/interface/variable.hpp
jlippuner: Actually, I think this change will better fit into #383, because I'm already changing and renaming some interfaces there.

tst/regression/test_suites/advection_outflow/advection_outflow.py
Yurlungur: In the case of this test, we're testing the physical output. So in my mind, checking the metadata makes the test more brittle for no reason. But yes, it is true that it will pass with updated gold files.

src/mesh/mesh.cpp
jlippuner: Yeah, I'm not super happy with this. Using `int32_t` here instead of `int` is basically being overly cautious in case we run on platforms where `int != int32_t` and we want to read and HDF5 file written on a platform where `int == int32_t` (or the other way around).

Although, I just realize that this means we'd also have to explicitly cast to `int32_t` when writing these attributes, which we currently don't do. They're declared as `int` and get written as such. So I think we should remove all references to `int32_t` instead and just use `int`. If we ever run into an issue of a platform where `int != int32_t`, we'll deal with it then.

src/outputs/parthenon_hdf5.cpp
jlippuner: The reason is because `VarInfo` should not default-constructed because then it would not be initialized. But I was too lazy to add a general constructor and need to initialize the fields individually in the `Decode` static constructor. But I should probably add a proper constructor and use that.

src/outputs/parthenon_hdf5.cpp
jlippuner: Good catch. This line has moved, so I'll fix it in my update

src/outputs/parthenon_hdf5.cpp
jlippuner: That's a good question. This code is copy-paste from what was there before to write HDF5 files. I'm assuming it's originally from Athena++. @pgrete @forrestglines any thoughts on this?

src/outputs/parthenon_hdf5.cpp
jlippuner: Good idea. Added.

src/outputs/parthenon_hdf5.cpp
jlippuner: So the issue here is that not all mesh blocks have the same variable fields. For example, we might have a sparse variable called "Sparse". Meshblock 0 has "Sparse_0", meshblock 1 has "Sparse_13" and "Sparse_2975", and meshblock 2 has "Sparse_0", "Sparse_13", and "Sparse_27". We now need to get a list of all the sparse IDs that "Sparse" has on any meshblock (globally). This is different from the resolved state that resolves name clashes, as far as I understand it. I don't think the StateDescriptor currently knows about the sparse IDs that may be allocated on any number of meshblocks, or does it? It just knows that there is a sparse variable called "Sparse", right? (At least, this will be true in #383, because there the sparse_id is removed from the Metadata.)

src/outputs/parthenon_hdf5.cpp
jlippuner: Hmm, interesting suggestion. In order to keep memory requirements manageable, I guess we would need to launch one kernel per meshblock, right? That would come with the complication of moving the mesh-block specific list of variables to the device. But I guess that would just be a variable pack per meshblock, although it would potentially have to be rebuilt each time because which sparse ids are allocated on a particular block can change at any time.

It's not obvious to me that this would actually improve performance significantly. But I guess we'd have to try to find out. I'm not sure if I'd want to hold up this PR (and thus the sparse implementation) to find out what the performance difference would be, though.

src/outputs/parthenon_hdf5.cpp
pgrete: I suggest to keep it as is for now and potentially update based on actual profiling data.
In principle these variables could also be set a runtime through ROMIO hints but overall (when running at scale) those numbers are likely machine/filesystem and simulation dependent.

pgrete: > I suggest to keep it as is for now and potentially update based on actual profiling data.
> In principle these variables could also be set a runtime through ROMIO hints but overall (when running at scale) those numbers are likely machine/filesystem and simulation dependent.

I now added #513 for this.

pgrete:
> This is a pretty big change so it would be good if someone from the AthenaPK team took a look. @pgrete @forrestglines @jmstone

@forrestglines would you mind reviewing this as you're more familiar with the hdf5 output?

src/outputs/parthenon_hdf5.cpp
jlippuner: Sounds good, thanks!

forrestglines:
> @forrestglines would you mind reviewing this as you're more familiar with the hdf5 output?
@jlippuner Yes, I can rereview this one today

src/mesh/mesh.cpp
pgrete: One additional comment related to data types/encoding.
Forrest and I stumbled across https://github.com/lanl/parthenon/issues/515 (which I personally see as an argument for being aware of encoding issues -- though, not directly related to the `int` <-> `int32_t` discussion here).

src/outputs/parthenon_hdf5.cpp
Yurlungur: How do you declare which sparse ids should exist at all? That has to be stored somewhere. I would think the `StateDescriptor` would know?

tst/regression/test_suites/advection_outflow/advection_outflow.py
jlippuner: That's a good point and I agree that this test is about the physics output. One thing to consider, though, is that if the HDF5 metadata changes, we have to update the gold files anyway because the `output` test actually tests the output and so I'd say we don't want to ignore metadata there. We could only avoid a gold update if a PR changes the metadata output of the `advection_outflow` test but not of the `output` test (which is possible, but would seem rare to me). I'd say let's leave metadata check turned on for now and if we run into a PR that only changes metadata for this test and not for the output test (which would require a gold update), then I'd be fine with turning off metadata check for this test.

src/outputs/parthenon_hdf5.cpp
jlippuner: Currently, the `SparseVariable` holds the allocated sparse ids on a per meshblock basis (`MeshblockData` has a list of `SparseVariable`s). So the `StateDescriptor` doesn't currently know which sparse ids exist. It only knows the names of the sparse variables. But you may be right that the `StateDescriptor` should know all the sparse ids that are allocated. However, we will still need a global allgather to exchange that information between all the ranks, unless we want to make allocating a new sparse id a global operation.

In fact, maybe there is a case to be made that all the sparse ids that can be used during the simulation have to be specified when the sparse variable is created. Then the list of possible sparse ids (per sparse variable) would live on the entire mesh (all ranks) from the very beginning, which will simply some things. We would still allocate particular sparse ids on individual blocks as needed, but at least the pool of ids is known globally ahead of time.

However, if we go this route, this would be implemented in #383 and then we could change this logic in the output code in that PR, so I think we should still merge what's in this PR now, even though it might be short lived...

src/outputs/parthenon_hdf5.cpp
jlippuner: I've made issue #516 for this, since this PR needs to move along and I need to focus on actual sparse implementation first.

src/parthenon_manager.cpp
jlippuner: See #516

jlippuner:
> Overall looks good. This was clearly a major effort. Thanks for going through it, @jlippuner ! It's a large enough MR that I don't think a review will catch bugs, but I trust you've run the regression tests through their paces? Have we tested the case of a restart with independent sparse variables?
> 
> Some non-blocking comments below. My biggest hangups are on performance, which can be addressed potentially---and may not need to be if the code is already fast.

Thanks again for your helpful suggestions. A restart test with independent sparse vars is coming in #383.

tst/regression/test_suites/advection_outflow/advection_outflow.py
Yurlungur: @forrestglines hdf5 for yt was one such example. The `output` test failed because it tested metadata, but the `outflow` test still passed.

src/outputs/parthenon_hdf5.cpp
Yurlungur: Waiting until #383 seems fine to me.

As discussed on the mattermost channel, I think delcaring which sparse IDs are globally available should be a global operation at the beginning of a simulation, in the `Initialize` functions for each package.

src/outputs/parthenon_hdf5.cpp
Yurlungur: That seems reasonable

forrestglines (approved):
Looks good!

tst/regression/test_suites/advection_outflow/advection_outflow.py
jlippuner: Yes, but in that case the gold files had to be updated because the `output` test failed. So if one updates the `output` gold files anyway, one might as well also update the `outflow` gold files. I'd we way only revert back to not checking metadata if we run into a change that only changes metadata for the `outflow` test and not the `output` test, because in that case it would save us from having to do a new gold release. But if we have to do a new gold release (because `output` test fails), it will also include new files for `outflow` test and so it's ok to keep checking metadata for `outflow` test. Does this make sense?

PR #488, opened 2021-04-14
update Dockerfile.nvcc to use HDF5 version 1.10.7

jlippuner:
## PR Summary

Update `Dockerfile.nvcc` for gitlab CI to use HDF5 version 1.10.7.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -5,12 +5,9 @@ before_script:
     - export OMP_PROC_BIND=close
     - export OMP_PLACES=cores
     - export OMP_NUM_THREADS=1
-    - export CMAKE_VERSION=3.12.4
     - export CTEST_OUTPUT_ON_FAILURE=1
     - export MACHINE_CFG=${PWD}/cmake/machinecfg/CI.cmake
     - export J=$(( $(nproc --all) / 4  + 1 )) && echo Using ${J} cores during build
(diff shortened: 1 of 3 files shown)


JoshuaSBrown (approved):

JoshuaSBrown (changes requested):
Can you add this dependency version requirement to CMake? Seems like it might be appropriate.

pgrete:
> Can you add this dependency version requirement to CMake? Seems like it might be appropriate.

I think this will first become relevant in the other PR where compression can be used.
Also that requirement should then depend on whether compression support is built in or not.

jlippuner:
> Can you add this dependency version requirement to CMake? Seems like it might be appropriate.

A version check is added to CMake in PR #486 where we really need the newer version of HDF5.

jlippuner:
> > Can you add this dependency version requirement to CMake? Seems like it might be appropriate.
> 
> I think this will first become relevant in the other PR where compression can be used.
> Also that requirement should then depend on whether compression support is built in or not.

Yup, the HDF5 version will only be checked if compression and parallel I/O is enabled, because that's what triggers the bug in older HDF5 versions.

JoshuaSBrown (approved):

pgrete (approved):

PR #493, opened 2021-04-20
Use subviews for buffers

pgrete:
## PR Summary

Before:

Screenshot from 2021-04-20 22-40-29

After:

Screenshot from 2021-04-20 22-42-49

## PR Checklist

- [ ] Code passes cpplint
- [ ] New features are documented.
- [ ] Adds a test for any bugs fixed. Adds tests for new features.
- [ ] Code is formatted
- [ ] Changes are summarized in CHANGELOG.md
- [ ] (@lanl.gov employees) Update copyright on changed files

PR #494, opened 2021-04-21
Jmm/one buffer

Yurlungur:
## PR Summary

Built on the discussions in the hackathon and on @pgrete 's unification of communication buffers in #493 . This combines the flux and coarse cell buffers into a single kokkos view. A couple of things to note:
- This reduces the number of allocations per variable per meshblock by 3.
- This relies on `LayoutRight` in our kokkos views.
- This is wasteful of some memory because the coarse buffer is now the same size as the dense buffer.
- This is built on #493 and should go in after it.

I haven't done any profiling yet.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/src/interface/variable.cpp
+++ b/src/interface/variable.cpp
@@ -1,5 +1,5 @@
 //========================================================================================
-// (C) (or copyright) 2020. Triad National Security, LLC. All rights reserved.
+// (C) (or copyright) 2020-2021. Triad National Security, LLC. All rights reserved.
 //
 // This program was produced under U.S. Government contract 89233218CNA000001 for Los
 // Alamos National Laboratory (LANL), which is operated by Triad National Security, LLC
--- a/src/interface/variable.cpp
+++ b/src/interface/variable.cpp
@@ -86,8 +86,6 @@ CellVariable<T>::AllocateCopy(const bool allocComms, std::weak_ptr<MeshBlock> wp
 /// Initialize a 6D variable
 template <typename T>
 void CellVariable<T>::allocateComms(std::weak_ptr<MeshBlock> wpmb) {
-  if (wpmb.expired()) return;
-  std::shared_ptr<MeshBlock> pmb = wpmb.lock();
   std::string base_name = label();
 
   // TODO(JMM): Note that this approach assumes LayoutRight. Otherwise
@@ -100,7 +98,8 @@ void CellVariable<T>::allocateComms(std::weak_ptr<MeshBlock> wpmb) {
   if (IsSet(Metadata::Independent)) {
     n_outer += (GetDim(1) > 1) + (GetDim(2) > 1) + (GetDim(3) > 1);
   }
-  n_outer += (pmb->pmy_mesh->multilevel);
+  n_outer += (!wpmb.expired() && wpmb.lock()->pmy_mesh != nullptr &&
+              wpmb.lock()->pmy_mesh->multilevel);
   comm_data_ = ParArray7D<T>(base_name + ".comm_data", n_outer, GetDim(6), GetDim(5),
                              GetDim(4), GetDim(3), GetDim(2), GetDim(1));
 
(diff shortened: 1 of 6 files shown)


src/interface/variable.cpp
```diff
@@ -85,28 +86,42 @@ CellVariable<T>::AllocateCopy(const bool allocComms, std::weak_ptr<MeshBlock> wp
...
-  if (pmb->pmy_mesh->multilevel)
-    coarse_s = ParArrayND<T>(base_name + ".coarse", GetDim(6), GetDim(5), GetDim(4),
-                             pmb->c_cellbounds.ncellsk(IndexDomain::entire),
-                             pmb->c_cellbounds.ncellsj(IndexDomain::entire),
-                             pmb->c_cellbounds.ncellsi(IndexDomain::entire));
+  if (pmb->pmy_mesh->multilevel) {
+    // This wastes about 1/2 a meshblock in memory
+    coarse_s = ParArrayND<T>(Kokkos::subview(comm_data_, offset++, Kokkos::ALL(),
+                                             Kokkos::ALL(), Kokkos::ALL(), Kokkos::ALL(),
+                                             Kokkos::ALL(), Kokkos::ALL()));
```
forrestglines, line 125: Getting around the wasted memory would probably mean allocating from fluxes and this coarse data from one long array, right? However, I don't think Kokkos supports changing layouts like that with a subview.

src/interface/variable.cpp
```diff
@@ -85,28 +86,42 @@ CellVariable<T>::AllocateCopy(const bool allocComms, std::weak_ptr<MeshBlock> wp
...
-    if (GetDim(2) > 1)
-      flux[X2DIR] = ParArrayND<T>(base_name + ".fluxX2", GetDim(6), GetDim(5), GetDim(4),
-                                  GetDim(3), GetDim(2), GetDim(1));
-    if (GetDim(3) > 1)
-      flux[X3DIR] = ParArrayND<T>(base_name + ".fluxX3", GetDim(6), GetDim(5), GetDim(4),
-                                  GetDim(3), GetDim(2), GetDim(1));
+    n_outer += (GetDim(1) > 1) + (GetDim(2) > 1) + (GetDim(3) > 1);
   }
+  n_outer += (pmb->pmy_mesh->multilevel);
+  comm_data_ = ParArray7D<T>(base_name + ".comm_data", n_outer, GetDim(6), GetDim(5),
```
forrestglines, line 103: Is `comm_data_` only the collection of 4 arrays - the 3 fluxes and the coarse array? Does the benefit of combining these allocations outweight the cost of wasted memory for `coarse_s`? Could `coarse_s` remain it's own array?

forrestglines: This is less relevant for this PR, but is allocating space for all the fluxes necessary? Could the flux divergence be computed in each dimension one by one so that only one flux array is needed in memory? If possible, this would allow larger problem sizes but might be too big a code change.

forrestglines (commented):
I'm curious to see the fix to the unit test, but once that's in we should merge

src/interface/variable.cpp
Yurlungur: Yeah, I tried that actually. Kokkos yelled at me because types.

src/interface/variable.cpp
Yurlungur: > Is comm_data_ only the collection of 4 arrays - the 3 fluxes and the coarse array? Does the benefit of combining these allocations outweight the cost of wasted memory for coarse_s? Could coarse_s remain it's own array?

That could be done. I'm not sure what the right trade-off is. I would say we accept the memory overhead for now, and we can split out the coarse buffer later if we desire.

> This is less relevant for this PR, but is allocating space for all the fluxes necessary? Could the flux divergence be computed in each dimension one by one so that only one flux array is needed in memory? If possible, this would allow larger problem sizes but might be too big a code change.

This is an interesting idea. I agree not for this PR, but something to think about.

PR #496, opened 2021-04-21
[WIP] Hackathon integration

Yurlungur:
## PR Summary

This branch contains all changes and tuning done in the Hackathon. We will merge our individual hackathon branches into this one. Then, when we're happy with all performance tweaks, this can be merged into `develop`.

Tagging @pgrete @forrestglines @jlippuner @felker for ongoing discussion

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted


Yurlungur:
Summary of hackathon:
- Todo for @pgrete Track down order of git hashes to test individually
- Todo for @Yurlungur try to hook the hackathon integration into phoebus
- We will try to wrap up remaining hackathon stuff by the next parthenon call
- Todo for whomever: figure out what's going wrong with the CI
- Todo for whomever: get trace data incrementally over hashes

felker:
@brryan 
More info from @tomidakn:
> Although MPI_Iprobe does nothing for us (it checks arrival of any message but we do not use the result), this is ABSOLUTELY NECESSARY for the performance of Athena++. Although non-blocking MPI communications look like multi-tasking running behind our code, actually they are not. The network interface card can run autonomously from the CPU, but to move the data between the memory and the network interface and initiate/complete communications, MPI has to do something using CPU. So to process communications, we have to allow MPI to use CPU . Theoretically MPI can use multi-thread for this (OpenMPI can be configured so) but it is not common because of performance and compatibility issues. Instead, MPI processes communications whenever any MPI function is called. MPI_Iprobe is one of the cheapest function in MPI and by calling this occasionally MPI can process communications "as if it is in the background". Using only MPI_Test, the communications were very slow. I suspect that MPI_Test changes the ordering of the messages internally (I guess it tries to promote the message it is Testing), and if we call MPI_Test for different messages, they are left half done. So if we remove them, I am sure we will see significant performance drop. I could not dig it up right now, Collela or Woodward mentioned this in a paper.

> Note that, this depends on the MPI implementation. In fact, on Fugaku, they use truly multi-threadead MPI based on OpenMPI, and they assign dedicated CPU cores for the MPI control processes (although this concept seems causing many issues right now, and I am very frustrated with it). In this case, we have to remove MPI_Iprobe to gain the best performance, because calling MPI_Iprobe is only additional load to slow down MPI.

> Also, calling MPI_Iprobe once per MPI_Test is just based on my experience. It is possible that there is better calling strategy. At least, we do not have to call it twice as far as I tested.

jlippuner:
All merged into develop

PR #497, opened 2021-04-21
Tracer particles example

brryan:
## PR Summary

Tracer particles that follow a grid-based quantity are worth supporting. In addition, an example that includes both cell-centered variables and particle swarms could help stress any outstanding issues with MPI communication; in particular, variables and particles appear to share MPI tags (though I could be mistaken). 

This PR adds a tracer particle example that advects a cell-centered variable with a sinusoidal perturbation using a first-order upwind method, and advects tracer particles in a separate first-order-in-time step. This works with two meshblocks with both 1 and 2 MPI ranks. 

image

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted


example/particle_tracers/particle_tracers.cpp
```diff
@@ -0,0 +1,515 @@
...
+
+// *************************************************//
+// define the Particles package, including         *//
+// initialization and update functions.            *//
+// *************************************************//
+
+std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
+  auto pkg = std::make_shared<StateDescriptor>("particles_package");
+
+  Real vmatx = pin->GetOrAddReal("Material", "vx", 1.0);
```
Yurlungur: I don't love the name `Material` here. What about `background`?

example/particle_tracers/particle_tracers.cpp
```diff
@@ -0,0 +1,515 @@
...
+
+  // Add swarm of tracer particles
+  std::string swarm_name = "tracers";
+  Metadata swarm_metadata;
+  pkg->AddSwarm(swarm_name, swarm_metadata);
+  Metadata real_swarmvalue_metadata({Metadata::Real});
+  pkg->AddSwarmValue("id", swarm_name, Metadata({Metadata::Integer}));
+
+  // Add density field
+  std::string field_name = "density";
```
Yurlungur: `advected`?

example/particle_tracers/particle_tracers.cpp
```diff
@@ -0,0 +1,515 @@
...
+    for (int n = 0; n < swarm->vbswarm->bd_var_.nbmax; n++) {
+      auto &nb = pmb->pbval->neighbor[n];
+      swarm->vbswarm->bd_var_.flag[nb.bufid] = BoundaryStatus::waiting;
+    }
+  }
+
+  return TaskStatus::complete;
+}
+
+} // namespace Particles
```
Yurlungur, line 319: Could these tasks be moved into the main Parthenon infrastructure? They could take a swarm name maybe as input?

example/particle_tracers/particle_tracers.cpp
```diff
@@ -0,0 +1,515 @@
...
+
+  auto density_h = density.GetHostMirror();
+  for (int k = kb.s; k <= kb.e; k++) {
+    for (int j = jb.s; j <= jb.e; j++) {
+      for (int i = ib.s; i <= ib.e; i++) {
+        density_h(k, j, i) = density_mean + density_amp * sin(2. * M_PI * coords.x1v(i));
+      }
+    }
+  }
+  density.DeepCopy(density_h);
```
Yurlungur: Could we put this in a `par_for`? Just to avoid a host copy? I know it doesn't matter since it's initialization but it would still be nicer I think.

Yurlungur (approved):
This is a really nice, simple, cute example of particles. I like it! Some pedantic comments below. All non-blocking.

Also, some minor thoughts:
- You need a changelog. :)
- Should we create a regression test for this? Probably not worth it since you already have a test stressing the particle machinery elsewhere.

example/particle_tracers/particle_tracers.cpp
brryan: Yeah that's good, switched to `Background`

example/particle_tracers/particle_tracers.cpp
brryan: :+1:

example/particle_tracers/particle_tracers.cpp
brryan: Yeah that's better (also will `coords.x1v(i)` even work on the host if you compile for GPU? I assume not). I think there was going to be a reason this should be done on the host at some point, but there isn't now.

example/particle_tracers/particle_tracers.cpp
brryan: Yes I think that should be done. Right now I think the thing to do (this issue shows up in other particles examples) is merge this in to @pgrete's branch, and then make that change there everywhere all at once.

brryan:
@Yurlungur Thanks for looking at this. Changelog updated. Let's hold off on a regression test though, since we already have two particles regression test PRs in the works.

example/particle_tracers/particle_tracers.cpp
Yurlungur: Works for me

example/particle_tracers/particle_tracers.cpp
Yurlungur: coords should work on both host and device I think. At least for now.

example/particle_tracers/particle_tracers.cpp
```diff
@@ -0,0 +1,511 @@
...
+  auto coords = pmb->coords;
+
+  const Real advected_mean = 1.0;
+  const Real advected_amp = 0.5;
+  PARTHENON_REQUIRE(advected_mean > advected_amp, "Cannot have negative densities!");
+
+  pmb->par_for(
+      "Init advected profile", kb.s, kb.e, jb.s, jb.e, ib.s, ib.e,
+      KOKKOS_LAMBDA(const int k, const int j, const int i) {
+        advected(k, j, i) = advected_mean + advected_amp * sin(2. * M_PI * coords.x1v(i));
```
pgrete, line 347: Should the sine wave be continuous? Then we should divide by the xmax-xmin.

example/particle_tracers/particle_tracers.cpp
```diff
@@ -0,0 +1,511 @@
...
+  // number = advected*volume
+  Real number_meshblock =
+      advected_mean * (x_max - x_min) -
+      advected_amp / (2. * M_PI) * (cos(2. * M_PI * x_max) - cos(2. * M_PI * x_min));
+  number_meshblock *= (y_max - y_min) * (z_max - z_min);
+  Real number_mesh = advected_mean * (x_max_mesh - x_min_mesh);
+  number_mesh -= advected_amp / (2. * M_PI) *
+                 (cos(2. * M_PI * x_max_mesh) - cos(2. * M_PI * x_min_mesh));
+  number_mesh *= (y_max_mesh - y_min_mesh) * (z_max_mesh - z_min_mesh);
+
```
pgrete, line 375: I'm not really sure what's going on here, but we can fix that in the other PR.

example/particle_tracers/particle_tracers.cpp
```diff
@@ -0,0 +1,511 @@
...
+        Real val;
+        do {
+          x(n) = x_min + rng_gen.drand() * (x_max - x_min);
+          val = advected_mean + advected_amp * sin(2. * M_PI * x(n));
+        } while (val < rng_gen.drand() * (advected_mean + advected_amp));
+
+        y(n) = y_min + rng_gen.drand() * (y_max - y_min);
+        z(n) = z_min + rng_gen.drand() * (z_max - z_min);
+
+        rng_pool.free_state(rng_gen);
```
pgrete, line 403: Does this work as intended for 1d or 2D? In the other functions above there are `ndim` based conditionals.

example/particle_tracers/particle_tracers.cpp
```diff
@@ -0,0 +1,511 @@
...
+      auto receive =
+          tl.AddTask(send, &SwarmContainer::Receive, sc.get(), BoundaryCommSubset::all);
+
+      auto deposit = tl.AddTask(receive, Particles::DepositTracers, pmb.get());
+
+      auto defrag = tl.AddTask(deposit, Particles::Defrag, pmb.get());
+    }
+  }
+
+  return tc;
```
pgrete, line 508: Now that I see this, I wondering if this kind of example would be a perfect example to demonstrate how two packages are used in one executable, e.g., using the advection examples and the tracer example as individual packages combined in a new application.

pgrete (approved):
I left a few comments, that I don't think need/should be addressed here and fixed other pieces directly (should now compile and pass the test suite).
I suggest to merge this in to the target branch (my regression branch) as soon as possible and then continue to fix that branch (e.g., separating the common tasks to the main infrastructure).
Then merge that branch into original particles mpi branch, rebase and fix that for a final test to eventually merge.

Does this sound like a plan? @brryan @Yurlungur

pgrete:
FYI CI is not passing because this PR still uses the old docker image that I updated to Cuda 11 some time ago.
This will be fixed with a later rebase of the parent branches.
Any objections to merge this?

Yurlungur:
Thanks for digging in @pgrete that plan works for me! no objections from me for moving forward

example/particle_tracers/particle_tracers.cpp
brryan: I guess it doesn't *have* to be, but it probably should be. Right now we are relying on specific x1max, x1min in the parameter file. This should be changed, thanks

example/particle_tracers/particle_tracers.cpp
brryan: It's the definite integrals of the initial sine wave of the `advected` field over each meshblock and the whole mesh, so given a number of tracers for the whole problem I can get the number of tracers on each meshblock without any MPI reduction calls.

example/particle_tracers/particle_tracers.cpp
brryan: This example should only work in 1D, I think I should just enforce that (unless you think it's worthwhile to go to 2D and 3D now).

example/particle_tracers/particle_tracers.cpp
brryan: Yeah I agree I think this is a good place to do that. My one comment is that I like that this example provides an almost maximally simple advection implementation (as opposed to the `advection` example which at least for me is a little hard to understand at first). I could move the advection code (e.g. `CalculateFluxes`) to a separate `advection` package in this example?

brryan:
Thanks for merging this in @pgrete. I will try to find the time to make the changes you suggested in `pgrete/particles-regression` tomorrow

Issue #421, opened 2021-01-18
AMR performance 7/N: Simplying communication routines


pgrete:
There exist some call in the logic of our communication routine that I don't fully follow (and which may be leftover from some initial design ideas).
For example, the `resetBoundary()` function for each `Variable` resets the `var_cc` data pointer.
I'm not sure, this is still required and if the overall communication logic could not be simplified.

@jdolence I've been tagging you here as discussed last Wednesday as (IIRC) you're most familiar with these pieces/the original design.

pgrete:
Looks like @Yurlungur may also have some good input on this issue.
Just found this comment in `variable.cpp` while trying to figure out what the difference between `bvars` and `bvars_main_int` is and if this is still need in Parthenon (or a remnant of the GRMHD code in Athena++)
```c++
  // TODO(JMM): This means RestrictBoundaries()
  // is called on EVERY stage, regardless of what
  // stage needs it.
  // The fix is to refactor BoundaryValues
  // to expose calls at either the `Variable`
  // or `MeshBlockData` and `MeshData` level.
  pmb->pbval->bvars.push_back(vbvar);
  pmb->pbval->bvars_main_int.push_back(vbvar)
```

Yurlungur:
Yes. Let me refresh my memory on how this works. There's some optimization here that needs to be done.

pgrete:
Three more riddles ;)
### 1 `coarse_s`
Why does `  ParArrayND<T> coarse_s; // used for sending coarse boundary calculation`, see variable.cpp
```c++
  if (pmb->pmy_mesh->multilevel)
    coarse_s = ParArrayND<T>(base_name + ".coarse", GetDim(6), GetDim(5), GetDim(4),
                             pmb->c_cellbounds.ncellsk(IndexDomain::entire),
                             pmb->c_cellbounds.ncellsj(IndexDomain::entire),
                             pmb->c_cellbounds.ncellsi(IndexDomain::entire));
```
have this shape?
This looks like it's covering the entire MeshBlock. Is this required? I'm not sure I fully follow.

### 2 `bvars_main_int`
Are ` bvars` and  `bvars_main_int` (see `bvals.hpp`)
```c++
  // variable-length arrays of references to BoundaryVariable instances
  // containing all BoundaryVariable instances:
  std::vector<std::shared_ptr<BoundaryVariable>> bvars;
  // subset of bvars that are exchanged in the main TimeIntegratorTaskList
[...]

PR #498, opened 2021-04-23
Cleanup unused user hooks and variables

pgrete:
## PR Summary

I originally planned to update the boundary comm in `Mesh::Initialize` in this branch, but realized that update would be convoluted by leftover original design.
Thus , this PR turned into a cleanup PR first, which now also includes some of the changes discussed in #421 and #485 

Just housekeeping changes:
- Cleaned up Mesh::Initialize (made internal logic more readable and removed artifacts, e.g., the `call` variable, from already removed openmp loop)
- Removed unused `bvars_main_int` vector from boundary function (was used in Athena++ to distinguish between which vars need to be updated in the main driver. We handle this via `Metadata::Independent`)
- Removed `gr_amr` phase from boundary comm (does not apply to Parthenon)
- Removed user hooks that were not linked to any callback as those are handled at the package level
- Fix compiler warnings and errors with clang++

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


example/stochastic_subgrid/stochastic_subgrid_package.cpp
```diff
@@ -319,8 +319,8 @@ void DoLotsOfWork(MeshBlockData<Real> *rc) {
           for (int n = 0; n < num_vars; ++n)
             (n % 2 == 0 ? even : odd) += sqrt(n + 1) * v(in + n, k, j, i);
 
-          Real a = pow(10.0, (odd + even) / (fmax(1.0, abs(odd * even))));
-          Real b = pow(10.0, (odd - even) / (fmax(1.0, abs(odd * even))));
+          Real a = pow(10.0, (odd + even) / (fmax(1.0, std::abs(odd * even))));
+          Real b = pow(10.0, (odd - even) / (fmax(1.0, std::abs(odd * even))));
```
JoshuaSBrown: Why are we just doing this for the abs methods?

src/bvals/boundary_flag.cpp
```diff
@@ -126,8 +126,10 @@ void CheckBoundaryFlag(BoundaryFlag block_flag, CoordinateDirection dir) {
     default:
       break;
     }
+    break;
+  default:
+    PARTHENON_FAIL(msg);
```
JoshuaSBrown, line 131: Nice.

src/outputs/parthenon_hdf5.cpp
```diff
@@ -310,13 +310,13 @@ void PHDF5Output::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm) {
...
       const hsize_t xDims[1] = {3};
-      int meshblock_size[xDims[0]] = {nx1, nx2, nx3};
+      int meshblock_size[3] = {nx1, nx2, nx3};
       writeH5AI32("MeshBlockSize", meshblock_size,
                   H5S::FromHIDCheck(H5Screate_simple(1, xDims, NULL)), infoDSet);
 
       // RootGridDomain - float[9] array with xyz mins, maxs, rats (dx(i)/dx(i-1))
       const hsize_t rootGridDomain_size[1] = {9};
-      Real rootGridDomain[rootGridDomain_size[0]] = {
+      Real rootGridDomain[9] = {
```
JoshuaSBrown: I prefer to have named constants that are used in place of magic numbers even when it seems obvious, but this is more preference.

src/mesh/mesh.cpp
```diff
@@ -970,15 +968,6 @@ void Mesh::EnrollBndryFncts_(ApplicationInput *app_in) {
...
 
-//----------------------------------------------------------------------------------------
-//! \fn void Mesh::EnrollUserRefinementCondition(AMRFlagFunc amrflag)
-//  \brief Enroll a user-defined function for checking refinement criteria
-
-void Mesh::EnrollUserRefinementCondition(AMRFlagFunc amrflag) {
-  if (adaptive) AMRFlag_ = amrflag;
-  return;
-}
-
```
JoshuaSBrown, line 981: Nice!

src/mesh/mesh.cpp
```diff
@@ -1013,24 +1002,6 @@ void Mesh::EnrollUserMeshGenerator(CoordinateDirection dir, MeshGenFunc my_mg) {
...
-//----------------------------------------------------------------------------------------
-//! \fn void Mesh::EnrollUserTimeStepFunction(TimeStepFunc my_func)
-//  \brief Enroll a user-defined time step function
-
-void Mesh::EnrollUserTimeStepFunction(TimeStepFunc my_func) {
-  UserTimeStep_ = my_func;
-  return;
-}
-
```
JoshuaSBrown, line 1033: Nice!

src/mesh/mesh.cpp
```diff
@@ -1042,85 +1013,76 @@ void Mesh::ApplyUserWorkBeforeOutput(ParameterInput *pin) {
...
+// \!fn void Mesh::Initialize(bool init_problem, ParameterInput *pin)
 // \brief  initialization before the main loop
 
-void Mesh::Initialize(int res_flag, ParameterInput *pin, ApplicationInput *app_in) {
+void Mesh::Initialize(bool init_problem, ParameterInput *pin, ApplicationInput *app_in) {
   Kokkos::Profiling::pushRegion("Mesh::Initialize");
-  bool iflag = true;
-  int inb = nbtotal;
+  bool init_done = true;
+  const int inb = nbtotal;
```
JoshuaSBrown: I wonder if this line is really even necessary. We could just use nbtotal it's a bit more descriptive and it's used on line 1084 anyway. Regardless, whatever is used it should be made consistent throughout the method.

Oh, never mind I see nbtotal is changing. Maybe add comment
```
\\ initial number of blocks
```

src/mesh/mesh.cpp
```diff
@@ -1042,85 +1013,76 @@ void Mesh::ApplyUserWorkBeforeOutput(ParameterInput *pin) {
...
       }
-    } // omp parallel
+    }
 
-    if (!res_flag && adaptive) {
-      iflag = false;
-      int onb = nbtotal;
+    if (init_problem && adaptive) {
+      init_done = false;
+      const int onb = nbtotal;
```
JoshuaSBrown: What does the o stand for in onb? A comment would be helpful.

src/mesh/mesh.cpp
```diff
@@ -1042,85 +1013,76 @@ void Mesh::ApplyUserWorkBeforeOutput(ParameterInput *pin) {
...
-// \!fn void Mesh::Initialize(int res_flag, ParameterInput *pin)
+// \!fn void Mesh::Initialize(bool init_problem, ParameterInput *pin)
 // \brief  initialization before the main loop
 
-void Mesh::Initialize(int res_flag, ParameterInput *pin, ApplicationInput *app_in) {
+void Mesh::Initialize(bool init_problem, ParameterInput *pin, ApplicationInput *app_in) {
   Kokkos::Profiling::pushRegion("Mesh::Initialize");
-  bool iflag = true;
-  int inb = nbtotal;
+  bool init_done = true;
```
JoshuaSBrown, line 1021: This is more descriptive nice!

JoshuaSBrown (approved):
I have a few comments that are optional regardless this is an improvement so thanks for the work!

example/stochastic_subgrid/stochastic_subgrid_package.cpp
pgrete: These were the ones that clang++ complained about. IIRC because the default `abs` was assuming an `int` datatype here.

src/outputs/parthenon_hdf5.cpp
pgrete: I generally agree. Here I'm not sure I could come up with a good name and the number is only used in two adjacent lines for defining the array dims so I'm leaning towards keeping it as they are here.

src/mesh/mesh.cpp
pgrete: Good point. If there are no objections I'll add that comment with the next PR tomorrow that changes the buffer calls here.

src/mesh/mesh.cpp
pgrete: See above, I'll add the comment in the next iteration of this function.

example/stochastic_subgrid/stochastic_subgrid_package.cpp
JoshuaSBrown: Because someone might come along and remove the std namespace I would add a comment, or I would call fabs instead.

src/outputs/parthenon_hdf5.cpp
```diff
@@ -310,13 +310,13 @@ void PHDF5Output::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm) {
       // open vector space
       // write mesh block size
       const hsize_t xDims[1] = {3};
-      int meshblock_size[xDims[0]] = {nx1, nx2, nx3};
+      int meshblock_size[3] = {nx1, nx2, nx3};
```
Yurlungur: Why specify 3 for the array length? We can just let the compiler figure it out
```suggestion
      int meshblock_size[] = {nx1, nx2, nx3};
```

src/outputs/parthenon_hdf5.cpp
Yurlungur: My vote would be don't put a number here at all. Just let the compiler figure it out.

src/outputs/parthenon_hdf5.cpp
```diff
@@ -326,8 +326,7 @@ void PHDF5Output::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm) {
 
       // RootGridSize - int[3] number of cells on the root grid
       const hsize_t rootGridSize_size[1] = {3};
-      int rootGridSize[rootGridSize_size[0]] = {pm->mesh_size.nx1, pm->mesh_size.nx2,
-                                                pm->mesh_size.nx3};
+      int rootGridSize[3] = {pm->mesh_size.nx1, pm->mesh_size.nx2, pm->mesh_size.nx3};
```
Yurlungur:
```suggestion
      int rootGridSize[] = {pm->mesh_size.nx1, pm->mesh_size.nx2, pm->mesh_size.nx3};
```

Yurlungur (approved):
This is a welcome change. The simpler we can make `bvals` the better.

example/stochastic_subgrid/stochastic_subgrid_package.cpp
pgrete: updated to fabs

src/outputs/parthenon_hdf5.cpp
pgrete: fixed

src/outputs/parthenon_hdf5.cpp
pgrete: fixed

Issue #489, opened 2021-04-15
Tests on darwin now failing


JoshuaSBrown:
```
13/46 Test #13: unit_tests:Overlapping SpaceInstances ..............................Child aborted***Exception:   0.29 sec
 Cuda ExecSpace Error: MDRange tile dims exceed maximum number of threads per block - choose smaller tile dimsKokkos::OpenMP::initialize WARNING: OMP_PROC_BIND environment variable not set
  In general, for best performance with OpenMP 4.0 or better set OMP_PROC_BIND=spread and OMP_PLACES=threads
  For best performance with OpenMP 3.1 set OMP_PROC_BIND=true
  For unit testing set OMP_PROC_BIND=false
Filters: Overlapping SpaceInstances
 Tile dimensions exceed Cuda limits
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
unit_tests is a Catch v2.11.1 host application.
Run with -? for options
-------------------------------------------------------------------------------
Overlapping SpaceInstances
  Many Threads Short Kernel
-------------------------------------------------------------------------------
../tst/unit/kokkos_abstraction.cpp:692
...............................................................................
../tst/unit/kokkos_abstraction.cpp:692: FAILED:
due to a fatal error condition:
  SIGABRT - Abort (abnormal termination) signal
===============================================================================
test cases: 1 | 1 failed
assertions: 1 | 1 failed

```

pgrete:
Same on the CI machine. This is a plain Kokkos error and I want to discuss during the call what we about it (e.g., should we add hardware depended safety checks or change the test case).

JoshuaSBrown:
Still causing failure, looks like some other thresholds in the performance testing need to be updated:
```
Start 13: unit_tests:Overlapping SpaceInstances
13/46 Test #13: unit_tests:Overlapping SpaceInstances ..............................***Failed    0.27 sec
Kokkos::OpenMP::initialize WARNING: OMP_PROC_BIND environment variable not set
  In general, for best performance with OpenMP 4.0 or better set OMP_PROC_BIND=spread and OMP_PLACES=threads
  For best performance with OpenMP 3.1 set OMP_PROC_BIND=true
  For unit testing set OMP_PROC_BIND=false
Filters: Overlapping SpaceInstances
Many Threads Short Kernel
time default: 1.0375e-05
[...]

PR #500, opened 2021-04-26
Update docker file (for latest nsys)

pgrete:
## PR Summary

The updated cuda version on the host seems to have resulted in the docker `nsys` stalling (with tests timing out after three hours of "finalizing").
This updates also the docker image to use Cuda 11.3 with the latest version of `nsys`.
For reference: We use `nsys profile` to ensure there are not unintended host to device or vice versa copies during execution.

## PR Checklist


pgrete:
I now also updated the thresholds for the Overlapping SpaceInstances so that they are less restrictive (but still check the SpaceInstances themselves work), see also #489

Yurlungur (approved):

pgrete:
Just for reference: I manually triggered the full test suite and it passed.

tst/unit/kokkos_abstraction.cpp
```diff
@@ -622,8 +622,9 @@ struct SmallNLongTBufferPack {
   template <typename TimeType>
   static void test_time(const TimeType time_default, const TimeType time_spaces,
                         const int nspaces) {
```
jlippuner, line 624: So we don't need `nspaces` anymore? In that case we should do:
```suggestion
                        const int /*nspaces*/) {
```

jlippuner (approved):
Looks good. One tiny comment.

tst/unit/kokkos_abstraction.cpp
pgrete: Will handle that in a separate PR not not trigger CI again.

felker: Curious, is this because of `-Wunused-parameter` ? Why not remove the `const int` too?

PR #502, opened 2021-04-27
Use subviews of a single view for fluxes

Yurlungur:
Same as PR #494 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


src/bvals/bvals_interfaces.hpp
```diff
@@ -3,7 +3,7 @@
 // Copyright(C) 2014 James M. Stone <jmstone@princeton.edu> and other code contributors
 // Licensed under the 3-clause BSD License, see LICENSE file for details
 //========================================================================================
-// (C) (or copyright) 2020. Triad National Security, LLC. All rights reserved.
+// (C) (or copyright) 202-2021. Triad National Security, LLC. All rights reserved.
```
pgrete:
```suggestion
// (C) (or copyright) 2020-2021. Triad National Security, LLC. All rights reserved.
```

pgrete (changes requested):
Is it possible that this PR is half a mixture of the original coarse buffer plus fluxes with coarse buffers being pulled out again (but not all names and comments updated yet?

Yurlungur:
Thanks for the catch, @pgrete fixed.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 - [[PR 487]](https://github.com/lanl/parthenon/pull/487) Add default tiling matching `i` index range to MDRange loop pattern.
 
 ### Infrastructure (changes irrelevant to downstream codes)
+- [[PR 494]](https://github.com/lanl/parthenon/pull/494) Use subviews of a single view for fluxes and coarse cells
```
```suggestion
- [[PR 502]](https://github.com/lanl/parthenon/pull/502) Use subviews of a single view for fluxes
```

src/interface/variable.hpp
```diff
@@ -106,6 +106,7 @@ class CellVariable {
  private:
   Metadata m_;
   std::string label_;
+  ParArray7D<T> flux_data_; // unified par array for the fluxes and coarse buffer
```
pgrete:
```suggestion
  ParArray7D<T> flux_data_; // unified par array for the fluxes
```

src/interface/variable.cpp
```diff
@@ -85,28 +86,42 @@ CellVariable<T>::AllocateCopy(const bool allocComms, std::weak_ptr<MeshBlock> wp
 /// Initialize a 6D variable
 template <typename T>
 void CellVariable<T>::allocateComms(std::weak_ptr<MeshBlock> wpmb) {
```
pgrete, line 88: More a general comment and not related to this PR: I think it's counter intuitive that that the fluxes are allocated within `Comms` or am I missing something?

src/interface/variable.cpp
```diff
@@ -85,28 +86,42 @@ CellVariable<T>::AllocateCopy(const bool allocComms, std::weak_ptr<MeshBlock> wp
...
-    if (GetDim(2) > 1)
-      flux[X2DIR] = ParArrayND<T>(base_name + ".fluxX2", GetDim(6), GetDim(5), GetDim(4),
-                                  GetDim(3), GetDim(2), GetDim(1));
-    if (GetDim(3) > 1)
-      flux[X3DIR] = ParArrayND<T>(base_name + ".fluxX3", GetDim(6), GetDim(5), GetDim(4),
-                                  GetDim(3), GetDim(2), GetDim(1));
+    n_outer += (GetDim(1) > 1) + (GetDim(2) > 1) + (GetDim(3) > 1);
   }
+  flux_data_ = ParArray7D<T>(base_name + ".flux_data", n_outer, GetDim(6), GetDim(5),
+                             GetDim(4), GetDim(3), GetDim(2), GetDim(1));
```
pgrete: I'm not sure what happens when an array with one dim of 0 is created. I think this should be moved within the conditional again.
```suggestion
```

src/interface/variable.cpp
```diff
@@ -85,28 +86,42 @@ CellVariable<T>::AllocateCopy(const bool allocComms, std::weak_ptr<MeshBlock> wp
...
-  // set up fluxes
   std::string base_name = label();
+
+  // TODO(JMM): Note that this approach assumes LayoutRight. Otherwise
+  // the stride will mess up the types.
+
+  // Compute size of unified flux_data object and create it. A unified
+  // flux_data_ object reduces the number of memory allocations per
+  // variable per meshblock from 5 to 3.
+  int n_outer = 0;
```
pgrete: Probably not required any more with the change below.
```suggestion
```

src/interface/variable.cpp
```diff
@@ -85,28 +86,42 @@ CellVariable<T>::AllocateCopy(const bool allocComms, std::weak_ptr<MeshBlock> wp
...
   if (IsSet(Metadata::Independent)) {
-    flux[X1DIR] = ParArrayND<T>(base_name + ".fluxX1", GetDim(6), GetDim(5), GetDim(4),
-                                GetDim(3), GetDim(2), GetDim(1));
-    if (GetDim(2) > 1)
-      flux[X2DIR] = ParArrayND<T>(base_name + ".fluxX2", GetDim(6), GetDim(5), GetDim(4),
-                                  GetDim(3), GetDim(2), GetDim(1));
-    if (GetDim(3) > 1)
-      flux[X3DIR] = ParArrayND<T>(base_name + ".fluxX3", GetDim(6), GetDim(5), GetDim(4),
-                                  GetDim(3), GetDim(2), GetDim(1));
+    n_outer += (GetDim(1) > 1) + (GetDim(2) > 1) + (GetDim(3) > 1);
```
pgrete:
```suggestion
  flux_data_ = ParArray7D<T>(base_name + ".flux_data", wpmb->pmy_mesh->ndim, GetDim(6),
                               GetDim(5), GetDim(4), GetDim(3), GetDim(2), GetDim(1));
```

Note, I haven't tried this change yet and I noticed the new `pmb->pmy_mesh != nullptr` below.
Can we call this function without a Mesh? then using `ndim` may not work after all.

src/interface/variable.cpp
```diff
@@ -85,28 +86,42 @@ CellVariable<T>::AllocateCopy(const bool allocComms, std::weak_ptr<MeshBlock> wp
...
+  int offset = 0;
+  if (IsSet(Metadata::Independent)) {
+    for (int d = X1DIR; d <= X3DIR; ++d) {
+      if (GetDim(d) > 1) {
+        flux[d] = ParArrayND<T>(
+            Kokkos::subview(flux_data_, offset++, Kokkos::ALL(), Kokkos::ALL(),
+                            Kokkos::ALL(), Kokkos::ALL(), Kokkos::ALL(), Kokkos::ALL()));
+      }
+    }
+  }
```
pgrete: Probably a question of style, but I think the original, explicit version above is easier to parse.

```suggestion
  // set up fluxes
  if (IsSet(Metadata::Independent)) {
    flux[X1DIR] = ParArrayND<T>(Kokkos::subview(flux_data_, 0, Kokkos::ALL(), Kokkos::ALL(),
                                Kokkos::ALL(), Kokkos::ALL(), Kokkos::ALL(), Kokkos::ALL()));
    if (GetDim(2) > 1)
      flux[X2DIR] = ParArrayND<T>(Kokkos::subview(flux_data_, 1, Kokkos::ALL(), Kokkos::ALL(),
                                Kokkos::ALL(), Kokkos::ALL(), Kokkos::ALL(), Kokkos::ALL()));
    if (GetDim(3) > 1)
      flux[X3DIR] =ParArrayND<T>(Kokkos::subview(flux_data_, 2, Kokkos::ALL(), Kokkos::ALL(),
                                Kokkos::ALL(), Kokkos::ALL(), Kokkos::ALL(), Kokkos::ALL()));
  }
```

pgrete (changes requested):
Feel free to disregard the nitpicking/style comments. Only the allocation itself should likely be done within  the Metadata conditional.

src/interface/variable.cpp
Yurlungur: Maybe this is worth a longer discussion. I think the logic here is that coarse, comms, and fluxes, are all one-copy and all only needed for independent variables. But perhaps we should decouple them.

Yurlungur: Not going to fix it here though. ;)

src/interface/variable.cpp
Yurlungur: n_outer will always be at least 1. Because if you get to this point in the code, there will be at least 1 flux.

Yurlungur: I don't think you can get a dim0 here. `n_outer >= 1` always, so this will always have nonzero shape.

src/interface/variable.cpp
Yurlungur: See my note below.

src/interface/variable.cpp
Yurlungur: I wanted to keep this decoupled from `mesh` as much as possible. Variables are created without an underlying mesh in unit testing, and I think in general, it's a good idea to decouple the infrastructure as much as we can to enable self-contained tests. Relying on the existence of a mesh pointer will break the current unit tests.

Yurlungur: that said, I can see doing:
```C++
int n_outer = 1;
n_outer += (GetDim(2) > 1) + (GetDim(3) > 1);
```
to make it more clear that `n_outer` is always >= 1.

Alternatively I could add an assert to check that `n_outer >= 1` before creating the flux array.

src/interface/variable.cpp
Yurlungur: I disagree, but I don't feel strongly about it. I can change it back.

JoshuaSBrown (approved):
I don't see any problems with this pr. Thanks @Yurlungur.

src/interface/variable.cpp
pgrete: Why is always `n_outer >= 1` ? It's initialized to 0 and only updated in the `IsSet(Metadata::Independent)` conditional.

How about keeping the init value to 0 outside the conditional, then within set it to  `n_outer = 1 + (GetDim(2) > 1) + (GetDim(3) > 1);` and allocate the View.

Finally, this would also simplify the subview allocation below, which could then be a single for loop without conditionals (neither for `IsSet(Metadata::Independent)`  nor `GetDim` along
```c++
for (int d = X1DIR; d <= n_nouter; d++) {
        flux[d] = ParArrayND<T>(
            Kokkos::subview(flux_data_, d - 1, Kokkos::ALL(), Kokkos::ALL(),
                            Kokkos::ALL(), Kokkos::ALL(), Kokkos::ALL(), Kokkos::ALL()));
}
```

src/interface/variable.cpp
pgrete: We could reduce the logic from an `if for if` to a single `for` reusing the information from the previous condition here, see above.

src/interface/variable.cpp
Yurlungur: Yes. I like this solution. I will make the change.

src/interface/variable.cpp
Yurlungur: This might be fixed in @jlippuner 's new PR.

src/interface/variable.cpp
Yurlungur: Cleaned up. In fact, cleaned up even further by combining some conditionals.

src/interface/variable.cpp
Yurlungur: Resolved as discussed above

src/interface/variable.cpp
Yurlungur: Resolved as discussed above

Yurlungur:
Thanks for your comments, @pgrete. I believe all issues should be resolved now.

Yurlungur:
@par-hermes format

pgrete (approved):
Perfect!

PR #511, opened 2021-05-06
Regression analysis additions/improvements

forrestglines:
## PR Summary

A few additions and improvements to the python scripts for regression analysis.

- A "relative" comparison option for `phdf_diff.py`
   - Uses an error metric |a-b|/|a| instead of |a-b|
- Speed improvements to `phdf_diff.py` accomplished by using more numpy arrays
- Adds `compare_analytic.py` script, which compares components in a `phdf` file to user provided analytic functions
- Adds `--analyze` option to `run_test.py` which allows running the analysis for regression tests without rerunning the simulation
   - Useful for debugging analysis of long runtime tests
- Adds `GetComponents` method to `phdf.py`, which allows getting data for a list of components (which are named subsets of variables. Used for AthenPK and yt)
- Adds `GetVolumeLocations` and `GetFaceLocations` to `phdf.py`, which return Z,Y,X full dimensional arrays for cell-centered/face-centered locations
   - Useful generating location data that matches the dimensionality of `GetVar` and `GetComponents`

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted


forrestglines:
More potential improvements:

- [x] Method in `phdf` to read a single/list of components
- [x] Method in `phdf` to return a dictionary mapping of component names to component columns

forrestglines:
@par-hermes format

forrestglines:
@par-hermes format

scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
```diff
@@ -395,6 +432,106 @@ def Get(self, variable, flatten=True):
...
+
+    def __GenLocMeshGrid(self, z, y, x, flatten):
+        """
+        Returns Z[grid_idx,k,j,i], Y[grid_idx,k,j,i], X[grid_idx,k,j,i] full
+        arrays of locations from 2D arrays z[grid_idx,k], y[grid_idx,j],
+        x[grid_idx,i]
+        """
+
+        # loc[grid_idx,k,j,i]
+        loc_shape = (x.shape[0], z.shape[1], y.shape[1], x.shape[1])
```
jlippuner, line 508: Might be good to add a check that `x.shape[0] == y.shape[0] == z.shape[0]`

scripts/python/packages/parthenon_tools/parthenon_tools/phdf_diff.py
```diff
@@ -419,26 +426,37 @@ def compare(
...
-    if not quiet:
-        print(f0.TotalCells, "cells mapped")
+    # Share every coordinate
+    for grid_idx in range(loc.shape[1]):
+        loc[:, grid_idx] = np.meshgrid(
+            locations_z[grid_idx],
+            locations_y[grid_idx],
+            locations_x[grid_idx],
+            indexing="ij",
+        )
```
jlippuner: Could we use the new `GetVolumeLocations` function from `phdf.py` here? This code seems to be duplicating that function

scripts/python/packages/parthenon_tools/parthenon_tools/phdf_diff.py
```diff
@@ -452,75 +470,80 @@ def compare(
...
+
+        # Rearrange val1 to match ordering of meshblocks in val0
+        val1 = val1[otherBlockIdx]
+
+        # compute error at every point
+        if relative:
+            err_val = np.abs((val0 - val1) / val0)
+            # Set error values where val0==0 to 0
+            # Numpy masked arrays would be more robust here, but they are very slow
+            err_val[val0 == 0] = 0
```
jlippuner: For symmetry (so that comparison results are the same if `f0` and `f1` are swapped), I'd propose a small change:
```suggestion
            denom = 0.5 * (np.abs(val0) + np.abs(val1))
            err_val = np.abs(val0 - val1) / denom
            # Set error values where denom==0 to 0
            # Numpy masked arrays would be more robust here, but they are very slow
            err_val[denom == 0] = 0
```

tst/regression/run_test.py
```diff
@@ -74,14 +74,16 @@ def main(**kwargs):
 
     test_manager = tc.TestManager(run_test_py_path, **kwargs)
 
-    print("Make output folder in test if does not exist")
+    if not "analyze" in kwargs or kwargs["analyze"] == False:
```
jlippuner: This is splitting hair, but I'd prefer `if "analyze" not in kwargs` instead.

tst/regression/utils/compare_analytic.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+def l2_rel_err(gold, test):
+    mask = gold != 0
+    return np.linalg.norm((gold[mask] - test[mask]) / gold[mask])
+
+
+def linf_rel_err(gold, test):
+    mask = gold != 0
+    return np.max(np.abs((gold[mask] - test[mask]) / gold[mask]))
```
jlippuner: Here I'm ok with only dividing by `gold` instead of `abs(gold) + abs(test)`, because we don't need to treat `gold` and `test` on equal footing.

tst/regression/utils/compare_analytic.py
```diff
@@ -0,0 +1,102 @@
...
+    filename, analytic_components, err_func=l2_err, tol=1e-12, quiet=False
+):
+    """Compares data in filename to analytic gold data in analytic_components.
+
+    Arguments:
+    filename: Filename of ".phdf" file to compare data
+    analytic_components: Dictionary keying component names to analytic functions.
+        Each analytic function in the dictionary takes the arguments:
+
+        def analytic_func(Z,Y,x,t)
```
jlippuner:
```suggestion
        def analytic_func(Z,Y,X,t)
```

jlippuner (approved):
Nice improvement, thanks! I just have a handful of non-blocking comments.

scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
```diff
@@ -81,106 +81,143 @@ def __init__(self, filename):
...
-                self.offset[1],
-                self.MeshBlockSize[1] - self.offset[1],
-                self.offset[2],
-                self.MeshBlockSize[2] - self.offset[2],
+                coord = np.zeros((self.NumBlocks, self.MeshBlockSize[coord_i]))
+                for bId in range(self.NumBlocks):
+                    for cId in range(self.MeshBlockSize[coord_i]):
+                        coord[bId, cId] = 0.5 * (tmp[bId, cId] + tmp[bId, cId + 1])
+            coordf = tmp
+            return tmp, coord, coordf
```
JoshuaSBrown: Aren't tmp and coordf always going to be the same, why return three items when you can return two, is it because it makes it easier to split up the return values?

scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
```diff
@@ -395,6 +432,106 @@ def Get(self, variable, flatten=True):
...
+        number of blocks, and Nz, Ny, and Nx are the number of cells in the z,
+        y, and x directions respectively.
+        """
+
+        # Check if these components exist
+        missing_components = [
+            component
+            for component in components
+            if not component in self.ComponentsDatasetMap
+        ]
```
JoshuaSBrown, line 469: I'm not familiar with this style of syntax, is this treating the stuff in brackets as a single line? Otherwise my intuition tells me the if statement is missing an indentation?

scripts/python/packages/parthenon_tools/parthenon_tools/phdf_diff.py
JoshuaSBrown: I was going to suggest the exact same thing, @jlippuner great minds think alike. ;)

tst/regression/utils/compare_analytic.py
```diff
@@ -0,0 +1,102 @@
...
+#   Compares data in filename to a dictionary of analytic functions for some
+#   components
+################################################################################
+
+
+def compare_analytic(
+    filename, analytic_components, err_func=l2_err, tol=1e-12, quiet=False
+):
+    """Compares data in filename to analytic gold data in analytic_components.
+
```
JoshuaSBrown: I think it might make sense to put this file in: scripts/python/packages/parthenon_tools/parthenon_tools, that way it will be installed with the other tools. Is there a particular reason you placed it here?

JoshuaSBrown (approved):
Thanks for doing this @forrestglines this is a nice addition, pending the suggested changes I approve.

scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
Yurlungur: I believe that inside a list, you can have line breaks for free. @forrestglines is using a list comprehension here to build the `missing_components` list on the fly.

scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
```diff
@@ -395,6 +432,106 @@ def Get(self, variable, flatten=True):
...
+        Returns components as a dictionary of {component:component_data}
+
+        Throws an exception if the components do not exist in the file
+
+        Default is to return a flat array of length TotalCells.  However if
+        flatten is set to False, a 4D for each component array is returned that
+        has dimensions [NumBlocks, Nz, Ny, Nx] where NumBlocks is the total
+        number of blocks, and Nz, Ny, and Nx are the number of cells in the z,
+        y, and x directions respectively.
+        """
```
Yurlungur: I'm a little confused by the name "component" here. Can we add in a docstring somewhere what this word means precisely?

It looks to me like it means either a scalar dataset OR it means, e.g., `vector_x`?

scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
```diff
@@ -395,6 +432,106 @@ def Get(self, variable, flatten=True):
...
+                        component_data[component] = dataset[..., idx]
+
+        return component_data
+
+    def __GenLocMeshGrid(self, z, y, x, flatten):
+        """
+        Returns Z[grid_idx,k,j,i], Y[grid_idx,k,j,i], X[grid_idx,k,j,i] full
+        arrays of locations from 2D arrays z[grid_idx,k], y[grid_idx,j],
+        x[grid_idx,i]
+        """
```
Yurlungur, line 502: Nice improvement! :+1:

scripts/python/packages/parthenon_tools/parthenon_tools/phdf_diff.py
Yurlungur: Thirded.

tst/regression/utils/compare_analytic.py
Yurlungur: I would prefer the denominator be the average.

tst/regression/utils/compare_analytic.py
Yurlungur: I see why it's in regression, but I agree. I think this is more broadly useful and I would like access to it "downstream and installed."

tst/regression/utils/compare_analytic.py
```diff
@@ -0,0 +1,102 @@
...
+################################################################################
+#   Compare_analytic
+#
+#   Compares data in filename to a dictionary of analytic functions for some
+#   components
+################################################################################
+
+
+def compare_analytic(
```
Yurlungur: Oh this is a fantastic new capability. :+1:

Yurlungur (approved):
This is absolutely a fantastic improvement. Nice work, @forrestglines ! Some non-blocking comments below.

scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
forrestglines: Yep, line breaks work across any bracketing ie. `( ) [ ] { }`. This is just a list comprehension don't remember formatting like this - the python formatter may have done it.

tst/regression/run_test.py
forrestglines: I didn't realize that `not in` was valid python. Definitely reads more naturally.

tst/regression/utils/compare_analytic.py
forrestglines: I agree. This started before the other tools were put together as an installable package. `compare_analytic.py` belongs with the other tools.

forrestglines:
@par-hermes format

scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
forrestglines: These are components within a dataset (or spread across multiple datasets). For AthenaPK, these are `Density`, `MomentumDensity1`,`TotalEnergyDensity`, etc.  They get written to `Info/ComponentNames`. There's documentation on how these get added: https://github.com/lanl/parthenon/blob/develop/docs/outputs.md#preparing-outputs-for-yt I'll add to the docstring.

tst/regression/utils/compare_analytic.py
forrestglines: I ended up reimplementing these error functions as a single function. I'm using the average for the denominator now.

forrestglines:
@par-hermes format

scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
Yurlungur: So do we need to add the component names for yt for this machinery to work?

forrestglines:
@par-hermes format

scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
forrestglines: No, if you don't provide component names, Parthenon will give default component names that match the OutputDatasetNames. Calling `GetComponents` with dataset names in that case will return the dataset in its scalar/vector shape. `compare_analytic` will also happily accept those dataset names and managing components of the datasets is up to the user when providing the analytic functions.

I originally implemented the component names for `yt`, but later found them useful for `AthenaPK` tests.

PR #517, opened 2021-05-12
Remove shape duplication

jlippuner:
Remove optional dims argument when adding variable to `MeshBlockData` and use the shape from `Metadata` instead.

This way, the variable shape only exists in one place and it can't become inconsistent with the shape stored in `Metadata`.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/src/interface/meshblock_data.hpp
+++ b/src/interface/meshblock_data.hpp
@@ -163,36 +163,6 @@ class MeshBlockData {
   ///
   /// This function will eventually look at the metadata flags to
   /// identify the size of the first dimension based on the
-  /// topological location.
-  ///
-  /// @param label the name of the variable
-  /// @param metadata the metadata associated with the variable
-  /// @param dims the size of each element
-  ///
-  /// TODO(JMM): DO NOT make these strings const reference.
-  /// passing in C-style string literals misbehaves
-  void Add(const std::string &label, const Metadata &metadata,
-           const std::vector<int> &dims);
-
-  ///
-  /// Allocate and add a variable<T> to the container
-  ///
-  /// This function will eventually look at the metadata flags to
-  /// identify the size of the first dimension based on the
-  /// topological location.
-  ///
-  /// @param labelVector the array of names of variables
-  /// @param metadata the metadata associated with the variable
-  /// @param dims the size of each element
-  ///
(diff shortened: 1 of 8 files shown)


src/interface/meshblock_data.hpp
```diff
@@ -461,8 +431,7 @@ class MeshBlockData {
   MapToVariablePack<T> coarseVarPackMap_; // cache for varpacks over coarse arrays
   MapToVariableFluxPack<T> varFluxPackMap_;
 
-  void calcArrDims_(std::array<int, 6> &arrDims, const std::vector<int> &dims,
-                    const Metadata &metadata);
+  std::array<int, 6> calcArrDims_(const Metadata &metadata);
```
Yurlungur, line 434: Are there any performance implications of this change? or does the move operator save us?

Yurlungur (approved):
Looks good! I'm pretty sure this shouldn't have any downstream implications, since downstream codes should be setting shape in metadata anyway.

src/interface/meshblock_data.hpp
jlippuner: Good question. The compiler should do a return value optimization here, meaning it shouldn't make an unnecessary copy of the `std::array`, so I think there shouldn't be any performance penalty. Maybe it would even perform slightly better because we save a pointer lookup :)

src/interface/metadata.hpp
```diff
@@ -166,67 +166,65 @@ class Metadata {
...
-      : associated_(associated), shape_(shape) {
+  Metadata() = default;
+
+  // There are 4 optional arguments: shape, sparse_id, component_labels, and associated,
+  // so we'll need 16 constructors to provide all possible variants
+
+  // 5 constructors, this is the general constructor called by all other constructors, so
+  // we do some sanity checks here
+  Metadata(const std::vector<MetadataFlag> &bits, const std::vector<int> &shape = {1},
+           int sparse_id = -1, const std::vector<std::string> component_labels = {},
```
JoshuaSBrown: I think you can make component_labels a reference to avoid a copy.

JoshuaSBrown (approved):
Nice little improvement.

src/interface/metadata.hpp
jlippuner: Good catch, this should be a ref. Thanks!

PR #519, opened 2021-05-13
Add SHA256 checksum to bash uploader script

JoshuaSBrown:
## PR Summary

This is a small pr. Adds a checksum check, for the bash uploader script. This should make it more secure. We can verify that the file has not been tampered with. 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


JoshuaSBrown:
ping @jlippuner @pgrete @Yurlungur this is a really small pr that should not affect anything, please take a look when you get a chance.

pgrete (commented):
AFAIK the hash may change rather frequently (which we require us to update it manually every time).
Codecov provides some best practices on how to get the hash automatically, see https://about.codecov.io/blog/validating-the-bash-script-on-ci/
Of course, this does not help in a situation where their GitHub and website are compromised simultaneously.
I think using the automated validation is a is a good balance between maintainability and security.

JoshuaSBrown:
> AFAIK the hash may change rather frequently (which we require us to update it manually every time).
> Codecov provides some best practices on how to get the hash automatically, see https://about.codecov.io/blog/validating-the-bash-script-on-ci/
> Of course, this does not help in a situation where their GitHub and website are compromised simultaneously.
> I think using the automated validation is a is a good balance between maintainability and security.

That is some great feedback. Will refactor.

cmake/CodeCovBashDownloadScript.cmake
```diff
@@ -0,0 +1,13 @@
+#=========================================================================================
+# (C) (or copyright) 2020-2021. Triad National Security, LLC. All rights reserved.
```
jlippuner, line 2: Shouldn't this be only 2021?

jlippuner (approved):
This looks good. My only concern is what was mentioned already that the hash might change frequently. Do we really have to download the bash script every time we run cmake? I would expect that the scripts are backwards compatible, so we could just put one known-good script in the Parthenon repo and use that. Later we can update and check manually, if need be.

cmake/CodeCovBashDownloadScript.cmake
JoshuaSBrown: Good point.

JoshuaSBrown:
> This looks good. My only concern is what was mentioned already that the hash might change frequently. Do we really have to download the bash script every time we run cmake? I would expect that the scripts are backwards compatible, so we could just put one known-good script in the Parthenon repo and use that. Later we can update and check manually, if need be.

That is a possibility, though there might be license implications to directly including third party packages in the repo, I should check.

JoshuaSBrown:
@jlippuner I think I will go with your solution, it is a little easier to implement, one can always update the files when necessary. I will place the script in the external folder. As long as the license is included with the script there shouldn't be a problem.

Yurlungur (commented):
I agree with @pgrete and @jlippuner 's comments. I'll wait for the new version before doing a full review.

pgrete:
> @jlippuner I think I will go with your solution, it is a little easier to implement, one can always update the files when necessary. I will place the script in the external folder. As long as the license is included with the script there shouldn't be a problem.

I suggest to do some additional research first because as far as I understand I'm not convinced this will nor (not be required).
The main point of using that external script is so that the results can be uploaded and processed by codecov.
Thus, it's a 100% external dependency and is changed by codecov when they change their parsing infrastructure.
I'd be surprised if they keep backwards compatibility, i.e., allowing us to use an outdated script when we'd rerun coverage (with upload to codecov) using some commit in the past.
Moreover, I'm not sure there's even a usecase for the latter, i.e., I  cannot imagine a situation where we need to run code coverage for an old commit and upload the results to codecov.

JoshuaSBrown:
> > @jlippuner I think I will go with your solution, it is a little easier to implement, one can always update the files when necessary. I will place the script in the external folder. As long as the license is included with the script there shouldn't be a problem.
> 
> I suggest to do some additional research first because as far as I understand I'm not convinced this will nor (not be required).
> The main point of using that external script is so that the results can be uploaded and processed by codecov.
> Thus, it's a 100% external dependency and is changed by codecov when they change their parsing infrastructure.
> I'd be surprised if they keep backwards compatibility, i.e., allowing us to use an outdated script when we'd rerun coverage (with upload to codecov) using some commit in the past.
> Moreover, I'm not sure there's even a usecase for the latter, i.e., I cannot imagine a situation where we need to run code coverage for an old commit and upload the results to codecov.

I messaged the CodeCov people to see if they maintain backwards compatibility.

pgrete:
> > > @jlippuner I think I will go with your solution, it is a little easier to implement, one can always update the files when necessary. I will place the script in the external folder. As long as the license is included with the script there shouldn't be a problem.
> > 
> > 
> > I suggest to do some additional research first because as far as I understand I'm not convinced this will nor (not be required).
> > The main point of using that external script is so that the results can be uploaded and processed by codecov.
> > Thus, it's a 100% external dependency and is changed by codecov when they change their parsing infrastructure.
> > I'd be surprised if they keep backwards compatibility, i.e., allowing us to use an outdated script when we'd rerun coverage (with upload to codecov) using some commit in the past.
> > Moreover, I'm not sure there's even a usecase for the latter, i.e., I cannot imagine a situation where we need to run code coverage for an old commit and upload the results to codecov.
> 
> I messaged the CodeCov people to see if they maintain backwards compatibility.

Maybe I'm missing sth very basic, but what would be *our* usecase to go back to an old commit, run coverage on that commmit, and then upload it with an old version of the codecov script?
Even if we ever want to run code coverage for an old commit (for which I also cannot imagine a usecase right now), I still expect that the resulting report should processed/uploaded with the interface codecov uses at the time of running the coverage.

JoshuaSBrown:
> > > > @jlippuner I think I will go with your solution, it is a little easier to implement, one can always update the files when necessary. I will place the script in the external folder. As long as the license is included with the script there shouldn't be a problem.
> > > 
> > > 
> > > I suggest to do some additional research first because as far as I understand I'm not convinced this will nor (not be required).
> > > The main point of using that external script is so that the results can be uploaded and processed by codecov.
> > > Thus, it's a 100% external dependency and is changed by codecov when they change their parsing infrastructure.
> > > I'd be surprised if they keep backwards compatibility, i.e., allowing us to use an outdated script when we'd rerun coverage (with upload to codecov) using some commit in the past.
> > > Moreover, I'm not sure there's even a usecase for the latter, i.e., I cannot imagine a situation where we need to run code coverage for an old commit and upload the results to codecov.
> > 
> > 
> > I messaged the CodeCov people to see if they maintain backwards compatibility.
> 
> Maybe I'm missing sth very basic, but what would be _our_ usecase to go back to an old commit, run coverage on that commmit, and then upload it with an old version of the codecov script?
> Even if we ever want to run code coverage for an old commit (for which I also cannot imagine a usecase right now), I still expect that the resulting report should processed/uploaded with the interface codecov uses at the time of running the coverage.

None, that's not what I was concerned about, I was concerned that their public interface would change at some point. Generally, breaking changes like that are rare but I thought that was where your hesitation was stemming from. If you are not concerned about the public interface changing, then what is wrong with storing a version of the code covarage script in the parthenon repo as opposed to downloading it each time?

pgrete:
> > > > > @jlippuner I think I will go with your solution, it is a little easier to implement, one can always update the files when necessary. I will place the script in the external folder. As long as the license is included with the script there shouldn't be a problem.
> > > > 
> > > > 
> > > > I suggest to do some additional research first because as far as I understand I'm not convinced this will nor (not be required).
> > > > The main point of using that external script is so that the results can be uploaded and processed by codecov.
> > > > Thus, it's a 100% external dependency and is changed by codecov when they change their parsing infrastructure.
> > > > I'd be surprised if they keep backwards compatibility, i.e., allowing us to use an outdated script when we'd rerun coverage (with upload to codecov) using some commit in the past.
> > > > Moreover, I'm not sure there's even a usecase for the latter, i.e., I cannot imagine a situation where we need to run code coverage for an old commit and upload the results to codecov.
> > > 
> > > 
> > > I messaged the CodeCov people to see if they maintain backwards compatibility.
> > 
> > 
> > Maybe I'm missing sth very basic, but what would be _our_ usecase to go back to an old commit, run coverage on that commmit, and then upload it with an old version of the codecov script?
> > Even if we ever want to run code coverage for an old commit (for which I also cannot imagine a usecase right now), I still expect that the resulting report should processed/uploaded with the interface codecov uses at the time of running the coverage.
> 
> None, that's not what I was concerned about, I was concerned that their public interface would change at some point. 

Isn't that the point of downloading every time? If their interface changes, the downloaded file would not work any more and the only (?) way for us to know whether the interface changed is to check for this every time we run it.

Maybe we're talking about different things. All my comments were targeted at the script provided by codecov to upload our results (created interdependently of that script using a set of tools within our control) so that it can be processed at the time of upload on their systems.


> Generally, breaking changes like that are rare but I thought that was where your hesitation was stemming from. If you are not concerned about the public interface changing, then what is wrong with storing a version of the code covarage script in the parthenon repo as opposed to downloading it each time?

See above. As far as I understand we'd have to download sth anyway (at least the hashes) to ensure that the version we're trying to use to push the results is up to date.

JoshuaSBrown:
@pgrete ok this should make you happy.

JoshuaSBrown:
This needs to be retriggered via the web on ci/gitlab/gitlab.com to test the coverage functionality, it won't run by default on a branch other than develop.

pgrete (commented):
Thanks for the update. I triggered the full test suite.

With the more simple testing now, do we still need that separate file or would the three line version used in the codecov blog entry also work, e.g.,
```shell
curl -fLso codecov https://codecov.io/bash;
VERSION=$(grep -o 'VERSION=\"[0-9\.]*\"' codecov | cut -d'"' -f2);
shasum -a 256 -c --ignore-missing <(curl -s "https://raw.githubusercontent.com/codecov/codecov-bash/${VERSION}/SHA256SUM")
```

JoshuaSBrown:
> Thanks for the update. I triggered the full test suite.
> 
> With the more simple testing now, do we still need that separate file or would the three line version used in the codecov blog entry also work, e.g.,
> 
> ```shell
> curl -fLso codecov https://codecov.io/bash;
> VERSION=$(grep -o 'VERSION=\"[0-9\.]*\"' codecov | cut -d'"' -f2);
> shasum -a 256 -c --ignore-missing <(curl -s "https://raw.githubusercontent.com/codecov/codecov-bash/${VERSION}/SHA256SUM")
> ```

I intentionally used cmake functionality to avoid external dependence on external tools as it guarantees portability. I don't see the advantage of switching it, that will tie it to those commands albeit there is not likely to be a problem as grep curl etc are pretty ubiquitous. 

I think this is a preference thing.

jlippuner:
> > Thanks for the update. I triggered the full test suite.
> > With the more simple testing now, do we still need that separate file or would the three line version used in the codecov blog entry also work, e.g.,
> > ```shell
> > curl -fLso codecov https://codecov.io/bash;
> > VERSION=$(grep -o 'VERSION=\"[0-9\.]*\"' codecov | cut -d'"' -f2);
> > shasum -a 256 -c --ignore-missing <(curl -s "https://raw.githubusercontent.com/codecov/codecov-bash/${VERSION}/SHA256SUM")
> > ```
> 
> I intentionally used cmake functionality to avoid external dependence on external tools as it guarantees portability. I don't see the advantage of switching it, that will tie it to those commands albeit there is not likely to be a problem as grep curl etc are pretty ubiquitous.
> 
> I think this is a preference thing.

I agree with @JoshuaSBrown. Using just 3 lines would be nicer, but I think using CMake commands is the better option here for portability.

JoshuaSBrown:
@par-hermes format

pgrete (approved):
Outvoted and approved :smiley:

PR #522, opened 2021-05-14
Fix order of ComponentNames

### LINKED ISSUE #521: ComponentNames out of order
The HDF5 overhaul has led to the arrays `NumComponents` and `ComponentNames` in the output being out of order compared to `OutputDatasetNames`. 

I'm not sure how things should function with sparse variables in the output, but there should be enough information in `NumComponents`, `ComponentNames`, and `OutputDatasetNames` to build a dictionary of dictionaries such as `{"DatasetName":{"ComponentName":idx}}`

forrestglines:
## PR Summary

Fixes out of order `ComponentNames` and `NumComponents` introduced in  #486.

There should be enough information in `NumComponents`, `ComponentNames`, and `OutputDatasetNames` to build a dictionary of dictionaries such as `{"DatasetName":{"ComponentName":idx}}`.

Fixes #521 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


forrestglines:
@jlippuner Could you take a look at this change? I discovered that the use of `std::set` led to `ComponentNames` being in the different order compared to `OutputDatasetNames`. I made a quick fix that works but it's not elegant. I'm also not sure if it's correct in the context of sparse variables.

Do you see a better solution to access the component labels with the variable metadata for each `std::string` in `output_params.variables` from this function? I'm struggling to see where to access them.

Yurlungur (approved):

jlippuner:
> @jlippuner Could you take a look at this change? I discovered that the use of `std::set` led to `ComponentNames` being in the different order compared to `OutputDatasetNames`. I made a quick fix that works but it's not elegant. I'm also not sure if it's correct in the context of sparse variables.
> 
> Do you see a better solution to access the component labels with the variable metadata for each `std::string` in `output_params.variables` from this function? I'm struggling to see where to access them.

Sorry, only saw this now. Yes, there is indeed a mismatch between `OutputDatasetNames` and `ComponentNames`. However, I think we need to fix `OutputDatasetNames` rather than rearranging `ComponentNames`, because the `ComponentNames` is sorted alphabetically and it's the same order as the `SparseInfo` data.

I'll push what I think should fix the issue.

jlippuner:
@forrestglines I think the order is now consistent. I also moved the check for empty component labels to the `VarInfo` creator.

jlippuner (approved):

src/outputs/parthenon_hdf5.cpp
```diff
@@ -103,8 +104,10 @@ struct VarInfo {
 
   VarInfo(const std::string &label, const std::vector<std::string> &component_labels,
           int vlen, bool is_sparse, bool is_vector)
-      : label(label), component_labels(component_labels), vlen(vlen),
-        is_sparse(is_sparse), is_vector(is_vector) {
+      : label(label),
+        component_labels(component_labels.size() > 0 ? component_labels
+                                                     : std::vector<std::string>{label}),
```
JoshuaSBrown: Please use a different signature for component_labels the argument vs the member.

JoshuaSBrown: The same for the other arguments.

forrestglines: Yes - I renamed the members in `VarInfo` to have underscores

forrestglines:
@par-hermes format

PR #524, opened 2021-05-17
Enforce Metadata constraints, add `Metadata::WithFluxes`, fix Metadata checks

jlippuner:
## PR Summary

Adds a `Metadata::WithFluxes` flag. Documents and enforces some constraints on Metadata flags, and fixes a potential issue where the boundary variable is accessed without checking that it exists.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/src/interface/meshblock_data.cpp
+++ b/src/interface/meshblock_data.cpp
@@ -54,7 +54,7 @@ void MeshBlockData<T>::Add(const std::string &label, const Metadata &metadata) {
     int varIndex = metadata.GetSparseId();
     sparseMap_[label]->Add(varIndex, arrDims);
     auto &v = sparseMap_[label]->Get(varIndex);
-    v->allocateComms(pmy_block);
+    v->AllocateFluxesAndBdryVar(pmy_block);
   } else if (metadata.Where() == Metadata::Edge) {
     // add an edge variable
     std::cerr << "Accessing unliving edge array in stage" << std::endl;
@@ -77,7 +77,7 @@ void MeshBlockData<T>::Add(const std::string &label, const Metadata &metadata) {
   } else {
     auto sv = std::make_shared<CellVariable<T>>(label, arrDims, metadata);
     Add(sv);
-    sv->allocateComms(pmy_block);
+    sv->AllocateFluxesAndBdryVar(pmy_block);
   }
 }
 
--- a/src/interface/meshblock_data.cpp
+++ b/src/interface/meshblock_data.cpp
@@ -561,12 +561,12 @@ template <typename T>
 TaskStatus MeshBlockData<T>::SendFluxCorrection() {
   Kokkos::Profiling::pushRegion("Task_SendFluxCorrection");
   for (auto &v : varVector_) {
-    if (v->HasFluxes() && v->HasBoundaryVars()) {
+    if (v->IsSet(Metadata::WithFluxes) && v->IsSet(Metadata::FillGhost)) {
       v->vbvar->SendFluxCorrection();
     }
   }
   for (auto &sv : sparseVector_) {
-    if (sv->HasFluxes() && sv->HasBoundaryVars()) {
+    if (sv->IsSet(Metadata::WithFluxes) && sv->IsSet(Metadata::FillGhost)) {
       CellVariableVector<T> vvec = sv->GetVector();
       for (auto &v : vvec) {
         v->vbvar->SendFluxCorrection();
@@ -582,13 +582,13 @@ TaskStatus MeshBlockData<T>::ReceiveFluxCorrection() {
   Kokkos::Profiling::pushRegion("Task_ReceiveFluxCorrection");
   int success = 0, total = 0;
   for (auto &v : varVector_) {
-    if (v->HasFluxes() && v->HasBoundaryVars()) {
+    if (v->IsSet(Metadata::WithFluxes) && v->IsSet(Metadata::FillGhost)) {
       if (v->vbvar->ReceiveFluxCorrection()) success++;
       total++;
     }
   }
   for (auto &sv : sparseVector_) {
-    if (sv->HasFluxes() && sv->HasBoundaryVars()) {
+    if (sv->IsSet(Metadata::WithFluxes) && sv->IsSet(Metadata::FillGhost)) {
       CellVariableVector<T> vvec = sv->GetVector();
       for (auto &v : vvec) {
         if (v->vbvar->ReceiveFluxCorrection()) success++;
@@ -607,13 +607,13 @@ TaskStatus MeshBlockData<T>::SendBoundaryBuffers() {
   // sends the boundary
   debug = 0;
   for (auto &v : varVector_) {
-    if (v->HasBoundaryVars()) {
+    if (v->IsSet(Metadata::FillGhost)) {
       v->resetBoundary();
       v->vbvar->SendBoundaryBuffers();
     }
   }
   for (auto &sv : sparseVector_) {
-    if (sv->HasBoundaryVars()) {
+    if (sv->IsSet(Metadata::FillGhost)) {
       CellVariableVector<T> vvec = sv->GetVector();
       for (auto &v : vvec) {
         v->resetBoundary();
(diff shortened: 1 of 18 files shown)


src/interface/sparse_variable.hpp
```diff
@@ -99,6 +99,9 @@ class SparseVariable {
 
   bool IsSet(const MetadataFlag flag) { return metadata_.IsSet(flag); }
 
+  inline bool HasFluxes() const { return metadata_.IsSet(Metadata::WithFluxes); }
+  inline bool HasBoundaryVars() const { return metadata_.IsSet(Metadata::FillGhost); }
```
Yurlungur: Hmm now that I'm seeing these in action, I'm not sure I like them. This feels like it obscures the meaning of `WithFluxes` and `FillGhost` more than it helps... now it's harder to grep through the code to find where metadata is used. I think I prefer just making the `IsSet` calls... But I don't feel strongly about it.

src/interface/metadata.hpp
```diff
@@ -54,10 +48,44 @@
...
+  PARTHENON_INTERNAL_FOR_FLAG(Overridable)                                               \
+  /************************************************/                                     \
+  /**  SHAPE: Neither or one (but not both) can be specified */                          \
+  /**  a vector quantity, i.e. a rank 1 contravariant tensor */                          \
   PARTHENON_INTERNAL_FOR_FLAG(Vector)                                                    \
   /**  a rank-2 tensor */                                                                \
   PARTHENON_INTERNAL_FOR_FLAG(Tensor)                                                    \
+  /************************************************/                                     \
+  /**  DATATYPE: Exactly one must be specified (default is Real) */                      \
+  /** Boolean-valued quantity */                                                         \
```
Yurlungur, line 69: Not all data types are supported for all variables. Bool and integer variables are only used in swarms.

src/interface/metadata.hpp
```diff
@@ -54,10 +48,44 @@
...
+  /** is an independent, evolved variable */                                             \
+  PARTHENON_INTERNAL_FOR_FLAG(Independent)                                               \
+  /** is a derived quantity (ignored) */                                                 \
+  PARTHENON_INTERNAL_FOR_FLAG(Derived)                                                   \
+  /************************************************/                                     \
+  /**  COMMUNICATION: SharedComms can only be set if FillGhost is set */                 \
+  /** Do boundary communication */                                                       \
+  PARTHENON_INTERNAL_FOR_FLAG(FillGhost)                                                 \
+  /** Communication arrays are a copy: hint to destructor */                             \
+  PARTHENON_INTERNAL_FOR_FLAG(SharedComms)                                               \
```
Yurlungur: I wouldn't worry about this now, but are we actually using this? With comms all reference counted, I'm not sure we are

src/interface/metadata.hpp
```diff
@@ -423,10 +482,15 @@ class Metadata {
...
-            (bit == None));
+  /// count the number of set flags from the given list
+  int CountSet(const std::vector<MetadataFlag> &flags) const {
+    int num = 0;
+    for (const auto f : flags) {
+      if (IsSet(f)) {
+        ++num;
+      }
+    }
+    return num;
```
Yurlungur, line 481: I like this

src/interface/variable.cpp
```diff
@@ -98,28 +98,30 @@ void CellVariable<T>::allocateComms(std::weak_ptr<MeshBlock> wpmb) {
...
+    // enroll CellCenteredBoundaryVariable object
+    vbvar->bvar_index = pmb->pbval->bvars.size();
+    // TODO(JMM): This means RestrictBoundaries()
+    // is called on EVERY stage, regardless of what
+    // stage needs it.
+    // The fix is to refactor BoundaryValues
+    // to expose calls at either the `Variable`
+    // or `MeshBlockData` and `MeshData` level.
+    pmb->pbval->bvars.push_back(vbvar);
+  }
```
Yurlungur: This section is going to conflict with my `buffer-in-one` PR.

Yurlungur (approved):
Some non-blocking comments below.

src/interface/sparse_variable.hpp
jlippuner: I'm fine with just using `IsSet`. At least with `WithFluxes` it should be clearer what we're testing instead of `Independent`

src/interface/metadata.hpp
jlippuner: You're right, it doesn't seem to be used. I'm happy to remove it unless there are plans to use it in the future

src/interface/metadata.hpp
jlippuner: You're right, it doesn't seem to be used. I'm happy to remove it unless there are plans to use it in the future

src/interface/metadata.hpp
jlippuner: It looks like `Metadata` doesn't know whether it's used for swarms or fields, so I think I'll leave this part. But I'll add a check in `CellVariable` to make sure the `Metadata` says `Real`.

src/interface/variable.cpp
jlippuner: Thanks for the heads up. So should I change anything or will you deal with it in your PR (or whichever gets merged second, I guess).

src/interface/metadata.hpp
Yurlungur: I think removing it is better. :+1: :+1: :+1:

src/interface/variable.cpp
Yurlungur: We can just fix it in whatever gets merged second. It's a minor thing.

src/interface/meshblock_data.cpp
```diff
@@ -53,10 +53,8 @@ void MeshBlockData<T>::Add(const std::string &label, const Metadata &metadata) {
     }
     int varIndex = metadata.GetSparseId();
     sparseMap_[label]->Add(varIndex, arrDims);
-    if (metadata.IsSet(Metadata::FillGhost)) {
-      auto &v = sparseMap_[label]->Get(varIndex);
-      v->allocateComms(pmy_block);
-    }
+    auto &v = sparseMap_[label]->Get(varIndex);
+    v->allocateComms(pmy_block);
```
pgrete: Why is the check for FillGhost not required any more?
Would this lead to allocation of comms for all variables?

src/interface/meshblock_data.cpp
```diff
@@ -79,9 +77,7 @@ void MeshBlockData<T>::Add(const std::string &label, const Metadata &metadata) {
   } else {
     auto sv = std::make_shared<CellVariable<T>>(label, arrDims, metadata);
     Add(sv);
-    if (metadata.IsSet(Metadata::FillGhost)) {
-      sv->allocateComms(pmy_block);
-    }
+    sv->allocateComms(pmy_block);
```
pgrete: Same as above regarding why no FillGhost check is required any more.

pgrete (approved):
Looks good and should be ready for a merge (pending the FillGhost <-> allocComms question).

src/interface/meshblock_data.cpp
jlippuner: I've moved the check inside `CellVariable::allocateComms`. This function now does the following:
1. Allocate fluxes if and only if `Metadata::WithFluxes` is set
2. Allocate `CellCenteredBoundaryVariable` if and only if `Metadata::FillGhost` is set

Previously, `allocateComms` was only called if `Metadata::FillGhost` was true, so that means if `WithFluxes` was set but `FillGhost` was not, the fluxes would *not* be allocated. Now they are. That's what we want, right?

src/interface/meshblock_data.cpp
jlippuner: See above

src/interface/meshblock_data.cpp
jlippuner: I just pushed another commit to rename `allocateComms` to something more descriptive.

pgrete (approved):
Thanks for the final tweak! This makes it a lot easier to follow when looking a the code.
I'll pull the trigger momentarily.

JoshuaSBrown:
Reopening the branch to help debug ci issues.

PR #538, opened 2021-06-07
fix coarse buffer vs amr incompatibility. Resolves issue #537.

### LINKED ISSUE #537: Inconsistent assumptions about coarse buffers for variables
Currently, a coarse buffer is only created in a variable if `Metadata::FillGhost` is set. However, variables are added to mesh refinement's variables to refine based on whether or not they have `Metadata::Independent`. Given the way we currently create variables, this should probably be changed to `Metadata::FillGhost`.

Yurlungur:
## PR Summary

Resolves #537 . In recent metadata updates, we decided variables should only contain coarse buffers if `FillGhost` is set. However, mesh refinement is currently performed if a variable is `Independent`. This causes segfaults. This PR changes the relevant metadata flag for adding variables to the mesh refinement.

There is an open question here about how we should count variables for load-balancing, as these don't need to be the same as the variables for `var_cc_`. I have left them the same for now, as that appears to be an assumption in `amr_loadbalance.cpp` but I'm not sure that's correct in general.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 - [[PR 522]](https://github.com/lanl/parthenon/pull/522) Corrected ordering of `OutputDatasetNames` to match `ComponentNames`
 
 ### Fixed (not changing behavior/API/variables/...)
+- [[PR 537]](https://github.com/lanl/parthenon/pull/538) Fix inconsistent treatment of coarse buffers.
 - [[PR 487]](https://github.com/lanl/parthenon/pull/487) Add default tiling matching `i` index range to MDRange loop pattern.
 - [[PR 531]](https://github.com/lanl/parthenon/pull/531) Work around in parthenon_hdf5.cpp for GCC 7.3.0
 
--- a/docs/interface/Metadata.md
+++ b/docs/interface/Metadata.md
@@ -96,9 +96,11 @@ exclusive and required. All variables should be either independent or
 derived.
 
 - `Metadata::Independent` implies the variable is part of independent
-  state
+  state. In particular, implies data is in restart files 
+  and is prolongated/restricted during remeshing. 
+  Buffers for a coarse grid are allocated for independent variables.
 - `Metadata::Derived` implies the variable can be calculated, given
-  the independent state
+  the independent state. This is the default.
 
 ### Communication
 
--- a/src/interface/metadata.hpp
+++ b/src/interface/metadata.hpp
@@ -73,7 +73,7 @@
   /** Real-valued quantity */                                                            \
   PARTHENON_INTERNAL_FOR_FLAG(Real)                                                      \
   /************************************************/                                     \
-  /** INDEPENDENT: Exactly one must be specified (default is Independent) */             \
+  /** INDEPENDENT: Exactly one must be specified (default is Derived) */                 \
   /** is an independent, evolved variable */                                             \
   PARTHENON_INTERNAL_FOR_FLAG(Independent)                                               \
   /** is a derived quantity (ignored) */                                                 \
@@ -206,7 +206,7 @@ class Metadata {
       DoBit(Real, true);
     }
     if (CountSet({Independent, Derived}) == 0) {
-      DoBit(Independent, true);
+      DoBit(Derived, true);
     }
 
     // check if all flag constraints are satisfied, throw if not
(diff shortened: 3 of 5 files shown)


Yurlungur:
Full test suite should be run, as I changed behaviour that may effect regression tests.

jdolence:
Yeah there is definitely something to discuss here, I think.  Up until now, the assumption has been that AMR operates on the `Metadata::Independent` variables.  There are good reasons for that in the context where the independent variables are conserved, but perhaps that model is too restrictive.  I definitely don't think we should tie AMR to `Metadata::FillGhost`, though whether we use `Metadata::Independent` or try to provide a new flag like `Metadata::AMR` is an open question to me.

Yurlungur:
I am going to mark this as WIP until we can have the discussion that needs to be had. Perhaps over the Parthenon call if it can wait that long.

jlippuner (approved):
This looks reasonable. We should probably rename `nindependent`, though.

Yurlungur:
After a discussion with @jdolence I think the solution above is wrong. The main issue is that `FIllGhost` is not necessarily the same as needing coarse buffers. At remeshing steps, coarse buffers are required... even if ghost zone communication never happens. Therefore, independent variables *should* be registered for refinement, not `FIllGhost` variables. Thanks to `bvals_cc_in_one`, the `pmr` mesh variables are never actually called.

Therefore, I am going to modify this PR to do the following:
1. Register the coarse buffer for `Metadata::Independent` and `Metadata::FillGhost`
2. Only register comms vars for `Metadata::FillGhost`
3. Update the documentation to reflect this reality
4. I would like to only register `Metadata::Independent` variables with the `MeshRefinement` object, but this isn't possible, because the old `bvals_refine` code for getting boundary conditions uses `MeshRefinement`. To remove this, we need to write a `Prolongate` call for `MeshData` and remove the old `bvals_in_one` machinery. We will also need sparse variables to behave like dense ones. In the mean time, I'm registering `FillGhost` variables in the `MeshRefinement` object. 
5. I am changing `Metadata::Derived` to be the default, not `Metadata::Independent`. I think accidentally having an `Independent` variable when you think your variable is derived is worse than the other way around, as it will cause an unexpected performance hit. Whereas if you have a `Metadata::Derived` variable that you think is `Metadata::Independent`, presumably your code will crash in an obvious way.

I'm interested in people's thoughts. Especially @jdolence @jmstone @pgrete and @forrestglines . @jlippuner my apologies, this means you'll have to re-do your review.

jlippuner (approved):
Looks good

forrestglines (approved):
Looks good. Doesn't seem to break anything in AthenaPK, everything still works.

forrestglines:
>     5. I am changing `Metadata::Derived` to be the default, not `Metadata::Independent`. I think accidentally having an `Independent` variable when you think your variable is derived is worse than the other way around, as it will cause an unexpected performance hit. Whereas if you have a `Metadata::Derived` variable that you think is `Metadata::Independent`, presumably your code will crash in an obvious way.

I think this is the correct default for the same reasons you give.

forrestglines:
> There is an open question here about how we should count variables for load-balancing, as these don't need to be the same as the variables for `var_cc_`. I have left them the same for now, as that appears to be an assumption in `amr_loadbalance.cpp` but I'm not sure that's correct in general.

I'm confused about the context for this. Are we counting variables to estimate the cost of rebalancing, or are we counting to estimate the cost of each grid to determine how we should load balance? For the first, case I see that  `Metadata:Independent` would be the better flag since we only move those variables during a rebalance. If this variable counting is for the second scenario, I don't see how the number of variables is a great measure of cost for each grid.

Yurlungur:
> > There is an open question here about how we should count variables for load-balancing, as these don't need to be the same as the variables for `var_cc_`. I have left them the same for now, as that appears to be an assumption in `amr_loadbalance.cpp` but I'm not sure that's correct in general.
> 
> I'm confused about the context for this. Are we counting variables to estimate the cost of rebalancing, or are we counting to estimate the cost of each grid to determine how we should load balance? For the first, case I see that `Metadata:Independent` would be the better flag since we only move those variables during a rebalance. If this variable counting is for the second scenario, I don't see how the number of variables is a great measure of cost for each grid.

I actually don't know the answer. I haven't looked into that part of the code too deeply yet. We do need to look eventually though.

PR #539, opened 2021-06-08
Fix restart bug

Yurlungur:
## PR Summary

Downstream, I tried to restart a simulation in Phoebus, and got a very cryptic segfault. I found two issues causing the problem:
1. One loop bound was wrong---Parthenon is currently looping over the *max* number of vector elements in all variables. It should loop over the number of elements in a *given* variable.
2. Obfuscating the issue, if the code crashes, because the HDF5 file is not closed, hdf5 errors can pollute the output *and* corrupt the restart file. I found this even causes a problem in the advection example, where sometimes the restart would just hang for certain combinations of gcc and hdf5 versions.

In this PR, I fix the loop bounds and I call the destructor for the restart reader after reading is finished, which closes the HDF5 file.

To catch this in the future, I recommend we have an integration test where we restart with variables of different sizes. However, that's a bigger lift than this little bug fix, and this bug is blocking functionality. I would recommend we get this in, and I will create an issue that that reminds us to put in a test at a later date.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/src/parthenon_manager.cpp
+++ b/src/parthenon_manager.cpp
@@ -151,7 +151,7 @@ void ParthenonManager::ParthenonInitPackagesAndMesh() {
 
     // close hdf5 file to prevent HDF5 hangs and corrupted files
     // if code dies after restart
-    this->restartReader = nullptr;
+    restartReader = nullptr;
   }
 
   // add root_level to all max_level
--- a/src/parthenon_manager.cpp
+++ b/src/parthenon_manager.cpp
@@ -255,7 +255,7 @@ void ParthenonManager::RestartPackages(Mesh &rm, RestartReader &resfile) {
           for (int k = out_kb.s; k <= out_kb.e; ++k) {
             for (int j = out_jb.s; j <= out_jb.e; ++j) {
               for (int i = out_ib.s; i <= out_ib.e; ++i) {
-                for (int l = 0; l < vlen; ++l) {
+		      for (int l = 0; l < v_h.GetDim(4); ++l) {
                   v_h(l, k, j, i) = tmp[index++];
                 }
               }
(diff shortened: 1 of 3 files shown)


Yurlungur:
Created an issue for the testing. #539.

src/parthenon_manager.cpp
```diff
@@ -148,6 +148,10 @@ void ParthenonManager::ParthenonInitPackagesAndMesh() {
 
     // Read package data from restart file
     RestartPackages(*pmesh, *restartReader);
+
+    // close hdf5 file to prevent HDF5 hangs and corrupted files
+    // if code dies after restart
+    this->restartReader = nullptr;
```
jdolence: I don't understand why this fixes a bug, but I'll take your word for it.  Why explicitly use `this->` though?

src/parthenon_manager.cpp
```diff
@@ -251,7 +255,7 @@ void ParthenonManager::RestartPackages(Mesh &rm, RestartReader &resfile) {
           for (int k = out_kb.s; k <= out_kb.e; ++k) {
             for (int j = out_jb.s; j <= out_jb.e; ++j) {
               for (int i = out_ib.s; i <= out_ib.e; ++i) {
-                for (int l = 0; l < vlen; ++l) {
+                for (int l = 0; l < v_h.GetDim(4); ++l) {
```
jdolence, line 256: Good catch!

jdolence (approved):
Thanks for the fix!  If we can drop the `this` pointer, that would be great, but otherwise this should be merged in ASAP.

src/parthenon_manager.cpp
Yurlungur: I copied it from the destructor. Almost certainly don't need the `this` pointer.

Yurlungur:
Thanks, @jdolence ! I think our issues are addressed now.

pgrete (approved):
Good catch. I just pushed a tiny updated (removing the destructor from `ParthenonFinalize` as it's not needed any more).
If you agree with this changes, feel free to hit merge.

PR #552, opened 2021-06-21
Add missing fstream include to particle_leapfrog example

maximumcats:
## PR Summary

The particle_leapfrog example uses std::ofstream to create a file but doesn't do `#include <fstream>`, which is a problem at least with gcc 8.3 for some configurations.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 - [[PR 522]](https://github.com/lanl/parthenon/pull/522) Corrected ordering of `OutputDatasetNames` to match `ComponentNames`
 
 ### Fixed (not changing behavior/API/variables/...)
+- [[PR 552]](https://github.com/lanl/parthenon/pull/552) Fix missing include for fstream
 - [[PR 537]](https://github.com/lanl/parthenon/pull/538) Fix inconsistent treatment of coarse buffers.
 - [[PR 539]](https://github.com/lanl/parthenon/pull/539) Fix restart indexing/hdf5 bugs
 - [[PR 487]](https://github.com/lanl/parthenon/pull/487) Add default tiling matching `i` index range to MDRange loop pattern.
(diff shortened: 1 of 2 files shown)


Yurlungur (approved):
Thanks for the catch, @maxpkatz ! I think to trigger the CI, we need a branch on the main repo. Is this ready for merge? If so, I'll create a branch to trigger the CI and then we can merge it in.

maximumcats:
Yes I think this is OK. Note that this is actually on the main repo since one of you gave me write access to it at some point, and I saw the comment about that aspect of the CI in the contributing guidelines so I just put it here :-D Let me know if you'd like me to not do that in the future.

Since this appears to be a new example, does it make sense to skip this in the CHANGELOG, or should every PR have an entry in that log?

Yurlungur:
Oh oops thought this was a fork! No this i great. CI should trigger automatically then. Last thing we need is a changelog edit. I'll go ahead and add that. Then with a second approval we can merge.

jdolence (approved):

jlippuner (approved):
How was this not caught earlier?

jlippuner:
> How was this not caught earlier?

Ah, only specific compilers are affected...

maximumcats:
I am not sure if it is the compiler or if it was my specific configuration -- I was using
```
cmake -DKokkos_ENABLE_SERIAL=On -DCMAKE_CXX_COMPILER=g++ -DPARTHENON_DISABLE_HDF5=ON
```
I wouldn't be shocked if, for example, it worked before because of the HDF5 dependency bringing in fstream somewhere.

Yurlungur:
> Since this appears to be a new example, does it make sense to skip this in the CHANGELOG, or should every PR have an entry in that log?

Every PR should have an entry in the changelog, I think. No worries this time. I added it.

PR #556, opened 2021-07-08
Iterative Tasks

jdolence:
## PR Summary

This PR extends the tasking capability.  There are two new features.  First, within a region with multiple task lists, tasks can be marked as "Regional" which indicates that they should not be considered complete for the purposes of dependent tasks until all lists have that task complete.  Second, there is now a mechanism to associate tasks with an iterative pattern that needs to be executed multiple times until some criteria are satisfied.  Combining these two, we can now support global iterative solvers like Jacobi in a simple way.  A new example that uses the Jacobi method to solve the Poisson equation has been added to demonstrate the capability.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


example/poisson/poisson_package.cpp
```diff
@@ -0,0 +1,151 @@
...
+        dv(b, 0, k, j, i) /= 2.0 * ndim;
+        dv(b, 0, k, j, i) -= v(b, iphi, k, j, i);
+      });
+
+  parthenon::par_for(
+      DEFAULT_LOOP_PATTERN, "UpdatePhi", DevExecSpace(), 0, u->NumBlocks() - 1, kb.s,
+      kb.e, jb.s, jb.e, ib.s, ib.e,
+      KOKKOS_LAMBDA(const int b, const int k, const int j, const int i) {
+        v(b, iphi, k, j, i) += dv(b, 0, k, j, i);
+      });
```
Yurlungur: This is not flux conservative so it won't work for AMR. That's fine though I think. AMR won't work anyway without higher-order prolongation/restriction.

src/tasks/task_list.hpp
```diff
@@ -46,28 +47,102 @@ class TaskList {
...
+  void ClearIteration(const std::string &label) {
+    auto task = task_list_.begin();
+    while (task != task_list_.end()) {
+      if (task->GetLabel() == label) {
+        task = task_list_.erase(task);
+      } else {
+        ++task;
+      }
+    }
+  }
```
Yurlungur: I wonder if it's time to move beyond a task list and into, say, an ordered map, with the task label as a key. Because looping through the task list looks very slow to me.

No need to change that here. Just food for thought.

src/tasks/task_list.hpp
```diff
@@ -46,28 +47,102 @@ class TaskList {
...
     }
+    for (auto &label : completed_iters_) {
+      ClearIteration(label);
+    }
+    completed_iters_.clear();
     ClearComplete();
     if (IsComplete()) return TaskListStatus::complete;
     return TaskListStatus::running;
   }
+  bool Validate() const {
```
Yurlungur, line 247: If this always throws for invalid tasks, is there any reason to return anything?

src/tasks/task_list.hpp
```diff
@@ -89,6 +164,52 @@ class TaskList {
                          [=]() mutable -> TaskStatus { return (obj->*func)(args...); });
   }
 
+  template <class F, class... Args>
+  TaskID AddIterativeTask(const TaskType &type, const std::string &label,
```
Yurlungur: Could `AddIterativeTask` be an overload of just `AddTask`? Or would the templating not like that?

src/tasks/task_list.hpp
```diff
@@ -89,6 +164,52 @@ class TaskList {
...
+      throw_with_max_iters_[label] = false;
+    }
+    if (warn_with_max_iters_.count(label) == 0) {
+      warn_with_max_iters_[label] = true;
+    }
+    TaskID id(tasks_added_ + 1);
+    task_list_.push_back(Task(
+        id, dep,
+        [=, func = std::forward<F>(func)]() mutable -> TaskStatus {
+          return func(args...);
```
Yurlungur:
```suggestion
          return func(std::forward<Args>(args)...);
```
to maintain perfect forwarding

src/tasks/task_list.hpp
```diff
@@ -89,6 +164,52 @@ class TaskList {
...
+    max_iterations_[label] = max;
+  }
+
+  void SetFailWithMaxIterations(const std::string &label, bool flag) {
+    throw_with_max_iters_[label] = flag;
+  }
+
+  void SetWarnWithMaxIterations(const std::string &label, bool flag) {
+    warn_with_max_iters_[label] = flag;
+  }
```
Yurlungur: This part of the machinery feels a little odd... in fact having iterative tasks tied to a label, rather than being a separate manager object in themselves is a little odd to me. But it does express what we want to express.

src/tasks/task_list.hpp
```diff
@@ -46,28 +47,102 @@ class TaskList {
...
           MarkTaskComplete(task.GetID());
+          if (task.GetType() == TaskType::completion_criteria) {
+            IterationComplete(task.GetLabel());
+          }
+        } else if (task.GetStatus() == TaskStatus::iterate &&
+                   task.GetType() == TaskType::completion_criteria) {
+          ResetIteration(task.GetLabel());
         }
       }
     }
```
Yurlungur: This is beginning to get a little hard to follow, but I'm not sure there's anything to be done about that.

Yurlungur (approved):
Cool! This looks great! And very useful. It's a fairly simple change to the task list machinery. I have a few comments below.

Also I notice tests aren't passing, so we should deal with that before merge.

example/poisson/poisson_package.cpp
jdolence: Yeah, this is a garbage solver all around, but the example is just meant to demonstrate the iterative task usage.  In fact, this explicitly assumes that `dx` is fixed for all blocks, so AMR would be out anyway.

src/tasks/task_list.hpp
jdolence: I don't know.  I'd like to see some evidence that this is a bottle neck in practice before doing something more complicated.  There are other ways to avoid these sorts of loops too, I think.

src/tasks/task_list.hpp
jdolence: originally, this didn't throw.  the call is in an assert.  i wanted to validate that iterations always had a completion criteria when in debug mode, but skip this in release.  when I tested that, it aborted for invalid lists, but didn't give a helpful message at all.  so I added this throw.  I still only want this called in debug mode, and the nicest way I know how to do that is by having the call in an assert, and that means it needs to return a bool, I think.

src/tasks/task_list.hpp
jdolence: I didn't try.  It might be OK, but I kind of liked the explicitness of calling out these kinds of tasks as iterative.  I'd be curious if anyone else has an opinion on this.  I'd be fine with being overruled and try the overload.

src/tasks/task_list.hpp
jdolence: yeah, let me think about what that looks like.

jdolence: ok @Yurlungur.  That was a little painful, but I refactored into something that perhaps looks more like you imagined, at least from the perspective of the interface.  I'm not sure whether it made the code in `task_list.hpp` more or less clear.  While I _do_ like that downstream codes won't have to provide labels and such now, my worry is that this interface obfuscates the fact that the tasks associated with some iterative collection can be interleaved with other tasks in a region that may be part of another iterative pattern or otherwise.  So, very curious to hear what you think.  Seems like both approaches have some merits.  Not at all sure which one we should stick with.  Thoughts?

src/tasks/task_list.hpp
jdolence: fixed this and all the others.

example/poisson/poisson_package.cpp
Yurlungur: maybe worth putting a warning about that somewhere in the example?

src/tasks/task_list.hpp
Yurlungur: I agree---don't worry about it for now. It's just something I noticed.

src/tasks/task_list.hpp
Yurlungur: `PARTHENON_REQUIRE_THROWS` is always called. You need to change it to `PARTHENON_DEBUG_REQUIRE`

src/tasks/task_list.hpp
Yurlungur: We could also add the type of task as a tag. e.g.,
```C++
AddTask(TaskType::Iterative,...);
AddTask(TaskType::Normal,...);
//etc
```

Yurlungur: But on the other hand that's API breaking so NVM

example/poisson/poisson_driver.cpp
```diff
@@ -0,0 +1,90 @@
...
+  auto &md = pmesh->mesh_data.GetOrAdd("base", 0);
+  // make a mesh_data container for dphi
+  auto &mdelta = pmesh->mesh_data.GetOrAdd("delta", 0);
+
+  TaskRegion &solver_region = tc.AddRegion(1);
+  TaskList &tl = solver_region[0];
+
+  auto start_recv = tl.AddTask(none, &MeshData<Real>::StartReceiving, md.get(),
+                               BoundaryCommSubset::all);
+  auto &solver = tl.AddIteration();
```
Yurlungur: Should `AddIteration` except a human readable string as a useful debugging tag?

src/tasks/task_list.hpp
```diff
@@ -85,8 +158,80 @@ class TaskList {
...
+      tl_->tasks_added_++;
+      return id;
+    }
+    TaskList *tl_;
+    int key_;
+    unsigned int max_iterations_;
+    unsigned int count_ = 0;
+    bool throw_with_max_iters_ = false;
+    bool warn_with_max_iters_ = true;
+  };
```
Yurlungur: I like this quite a bit... but I don't love that it's nested. Is there some way to not nest the classes?

jdolence: I don't know.  I did this because it allows these objects to directly access/modify private variables in the enclosing class (TaskLIst).  That was the most obvious thing to do, but there might be a better approach.

src/tasks/task_list.hpp
Yurlungur: Hmm interesting. I actually like the new way it looks for applications better. I agree it somewhat obfuscates that the tasks can be interwoven, but it also makes interacting with bits in a given iterative solve more clear and intuitive I feel. So I lean towards that.

On the other hand, I really dislike the nested ness of the classes. Can we not just have them flat? Or maybe make them friends?

src/tasks/task_list.hpp
Yurlungur: Make the classes flat, but make them friends of each other. That should give private access.

jdolence: ok I separated the classes.  curious what you think

src/tasks/task_list.hpp
```diff
@@ -36,6 +36,75 @@ namespace parthenon {
...
+  template <class T, class... Args>
+  TaskID AddTask(TaskID const &dep, TaskStatus (T::*func)(Args...), T *obj,
+                 Args &&... args) {
+    return this->AddTask_(TaskType::iterative, dep, [=]() mutable -> TaskStatus {
+      return (obj->*func)(std::forward<Args>(args)...);
+    });
+  }
+
+  template <class T, class... Args>
+  TaskID AddCompletionTask(TaskID const &dep, T &&func, Args &&... args) {
```
Yurlungur: Should this be `SetCompletionTask`? Can we ever have more than one completion task? How does that work?

src/tasks/task_list.hpp
```diff
@@ -142,6 +211,14 @@ class TaskList {
     return valid;
   }
 
+  TaskID AddTask(Task tsk) {
+    TaskID id(tasks_added_ + 1);
+    tsk.SetID(id);
+    task_list_.push_back(tsk);
+    tasks_added_++;
+    return id;
+  }
```
Yurlungur: Hmm I see. Now the task id needs to be kind of exposed and tracked. In principle I think this is fine.

src/tasks/task_list.hpp
Yurlungur: I like it. I don't think it's too burdensome to return the task id in the way you do now. Seems entirely reasonable and clear how things are passed around.

src/tasks/task_list.hpp
jdolence: yeah that's fair.  I'll change it

Yurlungur: We might also wanna check that there's only ever one completion task.

Yurlungur: (Per set of iterative tasks, obviously)

src/tasks/task_list.hpp
```diff
@@ -27,54 +30,202 @@
...
+  template <class F, class... Args>
+  TaskID AddTask_(const TaskType &type, TaskID const &dep, F &&func, Args &&... args) {
+    TaskID id(0);
+    id = task_list_impl::AddTaskHelper(
+        tl_, Task(
+                 id, dep,
+                 [=, func = std::forward<F>(func)]() mutable -> TaskStatus {
+                   return func(std::forward<Args>(args)...);
+                 },
+                 type, key_));
```
Yurlungur, line 106: It's just rearranging things, but it seems to me that this bit of nastiness could be hidden by having a task constructor overload that automatically sets the task id to something nonsensical, which we could check for when we validate, to ensure it was overwritten properly.

example/poisson/parthinput.poisson
```diff
@@ -0,0 +1,62 @@
...
+x2min = -1.0
+x2max = 1.0
+ix2_bc = outflow
+ox2_bc = outflow
+
+nx3 = 1
+x3min = -0.5
+x3max = 0.5
+ix3_bc = periodic
+ox3_bc = periodic
```
Yurlungur, line 39: Is outflow the right BC here?

example/poisson/poisson_package.cpp
```diff
@@ -0,0 +1,156 @@
...
+  parthenon::par_reduce(
+      parthenon::loop_pattern_mdrange_tag, "CheckConvergence", DevExecSpace(), 0,
+      v.GetDim(5) - 1, kb.s, kb.e, jb.s, jb.e, ib.s, ib.e,
+      KOKKOS_LAMBDA(const int b, const int k, const int j, const int i, Real &eps) {
+        Real reps = std::abs(dv(b, 0, k, j, i) / v(b, 0, k, j, i));
+        Real aeps = std::abs(dv(b, 0, k, j, i));
+        eps = std::max(eps, std::min(reps, aeps));
+      },
+      Kokkos::Max<Real>(max_err));
+
```
Yurlungur, line 144: I understand you didn't think about MPI for this. But presumably a real application would need an MPI reduce here too? Maybe we should add a comment to that effect? And raise an error if the user tries to use MPI?

Yurlungur (commented):
I really like this interface now. And I'm happy to see it's working.

example/poisson/poisson_package.cpp
jdolence: But I _did_ think about MPI.  This all works in parallel.  The iteration completes only when _all_ completion_criteria tasks return `TaskStatus::complete`

example/poisson/parthinput.poisson
jdolence: Sort of.  I'm actually using a Dirichlet boundary for the Poisson solver.  Initially, Phi = 0 everywhere (including in the boundary ghost cells).  Then the driver never calls anything to update the boundaries, so they stay fixed at zero while the solver iterates.

Yurlungur: Ah ok this makes sense

example/poisson/poisson_package.cpp
Yurlungur: Ah I see. This makes sense.

jdolence:
@pgrete I think this is getting pretty close.  I _think_ this might be useful for the kind of reduction stuff we talked about.  Curious what you think.

example/poisson/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
+#=========================================================================================
+# (C) (or copyright) 2020. Triad National Security, LLC. All rights reserved.
```
jlippuner:
```suggestion
# (C) (or copyright) 2021. Triad National Security, LLC. All rights reserved.
```

example/poisson/main.cpp
```diff
@@ -0,0 +1,53 @@
+//========================================================================================
+// (C) (or copyright) 2020. Triad National Security, LLC. All rights reserved.
```
jlippuner:
```suggestion
// (C) (or copyright) 2021. Triad National Security, LLC. All rights reserved.
```

example/poisson/main.cpp
```diff
@@ -0,0 +1,53 @@
...
+  // Redefine parthenon defaults
+  pman.app_input->ProcessPackages = poisson_example::ProcessPackages;
+  pman.app_input->ProblemGenerator = poisson_example::ProblemGenerator;
+
+  // call ParthenonInit to initialize MPI and Kokkos, parse the input deck, and set up
+  auto manager_status = pman.ParthenonInit(argc, argv);
+  if (manager_status == ParthenonStatus::complete) {
+    pman.ParthenonFinalize();
+    return 0;
+  }
```
jlippuner, line 32: This is outside the scope of this PR, but looking at the `ParthenonManager` code, it looks like `ParthenonInit` can never return `ParthenonStatus::complete`. So we should probably remove it here and in other places too.

example/poisson/parthenon_app_inputs.cpp
```diff
@@ -0,0 +1,76 @@
+//========================================================================================
+// (C) (or copyright) 2020-2021. Triad National Security, LLC. All rights reserved.
```
jlippuner:
```suggestion
// (C) (or copyright) 2021. Triad National Security, LLC. All rights reserved.
```

example/poisson/parthinput.poisson
```diff
@@ -0,0 +1,62 @@
+# ========================================================================================
+#  Athena++ astrophysical MHD code
+#  Copyright(C) 2014 James M. Stone <jmstone@princeton.edu> and other code contributors
+#  Licensed under the 3-clause BSD License, see LICENSE file for details
+# ========================================================================================
+#  (C) (or copyright) 2020-2021. Triad National Security, LLC. All rights reserved.
```
jlippuner:
```suggestion
#  (C) (or copyright) 2021. Triad National Security, LLC. All rights reserved.
```

example/poisson/poisson_driver.cpp
```diff
@@ -0,0 +1,103 @@
+//========================================================================================
+// (C) (or copyright) 2020-2021. Triad National Security, LLC. All rights reserved.
```
jlippuner:
```suggestion
// (C) (or copyright) 2021. Triad National Security, LLC. All rights reserved.
```

example/poisson/poisson_driver.hpp
```diff
@@ -0,0 +1,43 @@
+//========================================================================================
+// (C) (or copyright) 2020-2021. Triad National Security, LLC. All rights reserved.
```
jlippuner:
```suggestion
// (C) (or copyright) 2021. Triad National Security, LLC. All rights reserved.
```

example/poisson/poisson_package.cpp
```diff
@@ -0,0 +1,168 @@
+//========================================================================================
+// (C) (or copyright) 2020-2021. Triad National Security, LLC. All rights reserved.
```
jlippuner:
```suggestion
// (C) (or copyright) 2021. Triad National Security, LLC. All rights reserved.
```

example/poisson/poisson_package.hpp
```diff
@@ -0,0 +1,31 @@
+//========================================================================================
+// (C) (or copyright) 2020. Triad National Security, LLC. All rights reserved.
```
jlippuner:
```suggestion
// (C) (or copyright) 2021. Triad National Security, LLC. All rights reserved.
```

src/basic_types.hpp
```diff
@@ -34,7 +34,7 @@ using Real = double;
 #endif
 #endif
```
jlippuner, line 36: Please update copyright at top

src/tasks/task_list.hpp
```diff
@@ -16,8 +16,11 @@
```
jlippuner, line 16: Please update copyright

src/tasks/task_types.hpp
```diff
@@ -23,20 +23,51 @@
```
jlippuner, line 23: Please update copyright

tst/regression/test_suites/poisson/parthinput.poisson
```diff
@@ -0,0 +1,62 @@
+# ========================================================================================
+#  Athena++ astrophysical MHD code
+#  Copyright(C) 2014 James M. Stone <jmstone@princeton.edu> and other code contributors
+#  Licensed under the 3-clause BSD License, see LICENSE file for details
+# ========================================================================================
+#  (C) (or copyright) 2020-2021. Triad National Security, LLC. All rights reserved.
```
jlippuner:
```suggestion
#  (C) (or copyright) 2021. Triad National Security, LLC. All rights reserved.
```

tst/regression/test_suites/poisson/poisson.py
```diff
@@ -0,0 +1,51 @@
+# ========================================================================================
+# Parthenon performance portable AMR framework
+# Copyright(C) 2020 The Parthenon collaboration
+# Licensed under the 3-clause BSD License, see LICENSE file for details
+# ========================================================================================
+# (C) (or copyright) 2020-2021. Triad National Security, LLC. All rights reserved.
```
jlippuner:
```suggestion
# (C) (or copyright) 2021. Triad National Security, LLC. All rights reserved.
```

src/utils/error_checking.hpp
```diff
@@ -195,6 +195,11 @@ inline void warn(std::stringstream const &message, const char *const filename,
   warn(message.str().c_str(), filename, linenumber);
 }
```
jlippuner, line 197: Please update copyright

example/poisson/parthenon_app_inputs.cpp
```diff
@@ -0,0 +1,76 @@
...
+// license in this material to reproduce, prepare derivative works, distribute copies to
+// the public, perform publicly and display publicly, and to permit others to do so.
+//========================================================================================
+
+#include <sstream>
+#include <string>
+
+#include <parthenon/package.hpp>
+
+//#include "poisson_driver.hpp"
```
jlippuner:
```suggestion
```

example/poisson/parthenon_app_inputs.cpp
```diff
@@ -0,0 +1,76 @@
...
+  IndexRange ib = cellbounds.GetBoundsI(IndexDomain::entire);
+  IndexRange jb = cellbounds.GetBoundsJ(IndexDomain::entire);
+  IndexRange kb = cellbounds.GetBoundsK(IndexDomain::entire);
+
+  auto coords = pmb->coords;
+  PackIndexMap imap;
+  std::vector<std::string> vars({"density", "potential"});
+  auto q = data->PackVariables(vars, imap);
+  int irho = imap["density"].first;
+  int iphi = imap["potential"].first;
```
jlippuner:
```suggestion
  const std::vector<std::string> vars({"density", "potential"});
  const auto &q = data->PackVariables(vars, imap);
  const int irho = imap["density"].first;
  const int iphi = imap["potential"].first;
```

example/poisson/poisson_driver.cpp
```diff
@@ -0,0 +1,103 @@
...
+        start_recv, parthenon::cell_centered_bvars::ReceiveBoundaryBuffers, md);
+
+    auto setb =
+        solver.AddTask(recv | update, parthenon::cell_centered_bvars::SetBoundaries, md);
+
+    auto clear = solver.AddTask(send | setb, &MeshData<Real>::ClearBoundary, md.get(),
+                                BoundaryCommSubset::all);
+
+    auto check = solver.SetCompletionTask(
+        update | clear, poisson_package::CheckConvergence<MeshData<Real>>, md.get(),
```
jlippuner: Listing `update` here is not necessary, because `clear` depends on `update` (via `setb`).

example/poisson/poisson_driver.cpp
```diff
@@ -0,0 +1,103 @@
...
+        mdelta.get());
+    solver_region.AddRegionalDependencies(0, i, check);
```
jlippuner, line 96: Would be good to have an explanatory comment for this line

example/poisson/poisson_driver.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+    auto print = tl.AddTask(check, poisson_package::PrintComplete);
```
jlippuner, line 98: `print` is not being used, as far as I can see

src/interface/mesh_data.hpp
```diff
@@ -229,6 +229,20 @@ class MeshData {
...
+    }
+    return TaskStatus::complete;
+  }
+
+  TaskStatus ClearBoundary(BoundaryCommSubset phase) {
+    for (const auto &pbd : block_data_) {
+      pbd->ClearBoundary(phase);
+    }
+    return TaskStatus::complete;
+  }
```
jlippuner: Why can we ignore the returned `TaskStatus` from `pbd->StartReceiving` and `pbd->ClearBoundary` here? Shouldn't we only return `TaskStatus::complete` if all of the respective `MeshBlockData` tasks return complete?

example/poisson/poisson_package.cpp
```diff
@@ -0,0 +1,168 @@
...
+  Kokkos::Profiling::pushRegion("Task_Poisson_UpdatePhi");
+  auto pm = u->GetParentPointer();
+
+  IndexRange ib = u->GetBoundsI(IndexDomain::interior);
+  IndexRange jb = u->GetBoundsJ(IndexDomain::interior);
+  IndexRange kb = u->GetBoundsK(IndexDomain::interior);
+
+  PackIndexMap imap;
+  std::vector<std::string> vars({"density", "potential"});
+  auto v = u->PackVariables(vars, imap);
```
jlippuner:
```suggestion
  const std::vector<std::string> vars({"density", "potential"});
  const auto &v = u->PackVariables(vars, imap);
```

example/poisson/poisson_package.cpp
```diff
@@ -0,0 +1,168 @@
...
+  const int irho = imap["density"].first;
+  const int iphi = imap["potential"].first;
+  std::vector<std::string> phi_var({"potential"});
+  PackIndexMap imap2;
+  auto dv = du->PackVariables(phi_var, imap2);
```
jlippuner:
```suggestion
  const std::vector<std::string> phi_var({"potential"});
  PackIndexMap imap2;
  const auto &dv = du->PackVariables(phi_var, imap2);
```

example/poisson/poisson_package.cpp
```diff
@@ -0,0 +1,168 @@
...
+  Kokkos::Profiling::pushRegion("Task_Poisson_UpdatePhi");
+  auto pm = u->GetParentPointer();
+
+  IndexRange ib = u->GetBoundsI(IndexDomain::interior);
+  IndexRange jb = u->GetBoundsJ(IndexDomain::interior);
+  IndexRange kb = u->GetBoundsK(IndexDomain::interior);
+
+  std::vector<std::string> vars({"potential"});
+  PackIndexMap imap;
+  auto v = u->PackVariables(vars, imap);
```
jlippuner:
```suggestion
  const std::vector<std::string> vars({"potential"});
  PackIndexMap imap;
  const auto &v = u->PackVariables(vars, imap);
```

example/poisson/poisson_package.cpp
```diff
@@ -0,0 +1,168 @@
...
+        }
+        dv(b, idphi, k, j, i) /= 2.0 * ndim;
+        dv(b, idphi, k, j, i) -= v(b, iphi, k, j, i);
+      });
+
+  parthenon::par_for(
+      DEFAULT_LOOP_PATTERN, "UpdatePhi", DevExecSpace(), 0, dv.GetDim(5) - 1, kb.s, kb.e,
+      jb.s, jb.e, ib.s, ib.e,
+      KOKKOS_LAMBDA(const int b, const int k, const int j, const int i) {
+        v(b, iphi, k, j, i) += dv(b, 0, k, j, i);
```
jlippuner: Why is variable index hard-coded to 0 here?

example/poisson/poisson_package.cpp
```diff
@@ -0,0 +1,168 @@
...
+  IndexRange ib = u->GetBoundsI(IndexDomain::interior);
+  IndexRange jb = u->GetBoundsJ(IndexDomain::interior);
+  IndexRange kb = u->GetBoundsK(IndexDomain::interior);
+
+  std::vector<std::string> vars({"potential"});
+  PackIndexMap imap;
+  auto v = u->PackVariables(vars, imap);
+  const int iphi = imap["potential"].first;
+  PackIndexMap imap2;
+  auto dv = du->PackVariables(vars, imap2);
```
jlippuner:
```suggestion
  const auto &dv = du->PackVariables(vars, imap2);
```

jlippuner (commented):
This is a great addition to the code. I'm breaking up my review into two parts. I went through most files and will go through the remaining two tomorrow. So far my comments (albeit many) are mostly minor.

example/poisson/poisson_driver.cpp
jdolence: Yeah nothing depends on this task, but my preference is still to capture its ID.

src/interface/mesh_data.hpp
jdolence: Both of those can only return `TaskStatus::complete`.  Still, it's a good point.  I now check the return status and throw if any of them return anything other than `TaskStatus::complete`.

example/poisson/poisson_package.cpp
jdolence: all accesses to these trivial one-variable packs started with hard-coded 0's, which is what revealed the issue that @Yurlungur tracked down (see #579 and #580).  I just missed this zero.  Replaced now.

jdolence:
Thanks for the review @jlippuner.  I think I've addressed all your comments so far.  I'll look out for more tomorrow.

src/tasks/task_list.hpp
```diff
@@ -27,54 +30,249 @@
...
 
+class TaskList;
+namespace task_list_impl {
+TaskID AddTaskHelper(TaskList *, Task);
+} // namespace task_list_impl
+
+class IterativeTasks {
+ public:
+  IterativeTasks(TaskList *tl, int key) : tl_(tl), key_(key) {
+    max_iterations_ = std::numeric_limits<unsigned int>::max();
```
jlippuner: What about using 0 for "no limit"? I'm assuming that an iteration limit of 0 would really make sense, because then none of the tasks get executed. An while UINT_MAX is effectively unlimited, I like the (commonly used) convention of using 0 as no limit.

src/tasks/task_list.hpp
```diff
@@ -27,54 +30,249 @@
...
+  TaskID SetCompletionTask(TaskID const &dep, TaskStatus (T::*func)(Args...), T *obj,
+                           Args &&... args) {
+    return AddTask_(TaskType::completion_criteria, check_interval_, dep,
+                    [=]() mutable -> TaskStatus {
+                      return (obj->*func)(std::forward<Args>(args)...);
+                    });
+  }
+
+  void SetMaxIterations(const unsigned int max) { max_iterations_ = max; }
+  void SetCheckInterval(const unsigned int chk) { check_interval_ = chk; }
```
jlippuner: I think it would be safer if these functions take an `int` instead of `unsigned int`. If the user does `int` arithmetic on the number of iterations and erroneously ends up with a negative value that gets passed into one of these functions, it will silently overflow into a huge positive number. Maybe something like this:

```suggestion
  void SetMaxIterations(const int max) {
    PARTHENON_REQUIRE_THROWS(max > 0, "Max number of iterations must be positive");
    max_iterations_ = std::static_cast<const unsigned int>(max);
    
    if (max_iterations_ > 1000000) {
      PARTHENON_WARN("Large number of max iterations in iterative task, is this intentional? Use SetUnlimitedIterations() to set no limit");
    }
  }
  void SetUnlimitedIterations() {
    max_iterations_ = 0; // or std::numeric_limits<unsigned int>::max() if not using 0 for unlimited
  }
  void SetCheckInterval(const int chk) {
    PARTHENON_REQUIRE_THROWS(chk > 0, "Check interval must be positive");
    check_interval_ = std::static_cast<const unsigned int>(chk);
    
    if ((max_iterations_ > 0) && (check_interval_ >= max_iterations_)) {
      PARTHENON_WARN("Check interval is greater or equal to max iterations, is this intentional?");
    } else if (check_interval_ > 1000000) {
      PARTHENON_WARN("Large check interval in iterative task, is this intentional?");
    }
  }
```

src/tasks/task_list.hpp
```diff
@@ -27,54 +30,249 @@
...
+                 id, dep,
+                 [=, func = std::forward<F>(func)]() mutable -> TaskStatus {
+                   return func(std::forward<Args>(args)...);
+                 },
+                 type, key_));
+    return id;
+  }
+  TaskList *tl_;
+  int key_;
+  unsigned int max_iterations_;
```
jlippuner: Maybe set default value here?

src/tasks/task_list.hpp
```diff
@@ -27,54 +30,249 @@
...
   TaskList() = default;
   bool IsComplete() { return task_list_.empty(); }
   int Size() { return task_list_.size(); }
-  void Reset() {
-    tasks_added_ = 0;
-    task_list_.clear();
-    tasks_completed_.clear();
+  void MarkRegional(const TaskID &id) {
+    for (auto &task : task_list_) {
+      if (task.GetID() == id) {
```
jlippuner, line 125: Oof, I don't like this. This has the potential to lead to really bad performance issues if we ever have an unexpectedly large number of tasks. Is `std::list` really the best container for the task list? I think I'd be more comfortable with an `unordered_map`... Thoughts?

src/tasks/task_list.hpp
```diff
@@ -27,54 +30,249 @@
...
+    while (task != task_list_.end()) {
+      if (task->GetKey() == key) {
+        task = task_list_.erase(task);
+      } else {
+        ++task;
+      }
+    }
+  }
+  void ResetIteration(const int key) {
+    iter_tasks[key].IncrementCount();
```
jlippuner, line 187:
```suggestion
    PARTHENON_REQUIRE_THROWS(key < iter_tasks.size(), "Invalid iteration key");
    iter_tasks[key].IncrementCount();
```

src/tasks/task_list.hpp
```diff
@@ -27,54 +30,249 @@
...
+      }
+      ClearIteration(key);
+      return;
+    }
+    for (auto &task : task_list_) {
+      if (task.GetKey() == key) {
+        if (CheckDependencies(task.GetID())) {
+          MarkTaskComplete(task.GetID());
+        }
+        task.SetStatus(TaskStatus::incomplete);
```
jlippuner, line 205: So we always want to leave the task with the incomplete status, right (so this is not inside an `else` clause)? What about if we return above if the max iterations have been reached, will it end up with incomplete?

src/tasks/task_list.hpp
```diff
@@ -85,53 +283,165 @@ class TaskList {
...
     std::cout << "TaskList::Print():" << std::endl;
     for (auto &t : task_list_) {
       std::cout << "  " << i << "  " << t.GetID().to_string() << "  "
-                << t.GetDependency().to_string() << std::endl;
+                << t.GetDependency().to_string() << " "
+                << (t.GetStatus() == TaskStatus::incomplete)
+                << (t.GetStatus() == TaskStatus::complete)
+                << (t.GetStatus() == TaskStatus::skip)
+                << (t.GetStatus() == TaskStatus::iterate)
+                << (t.GetStatus() == TaskStatus::fail) << std::endl;
```
jlippuner, line 315: Can a task have multiple statuses at once? If so, I'd suggest adding some text here to output which status is being printed, if not, it would be good to convert the status to a string.

src/tasks/task_types.hpp
```diff
@@ -23,20 +23,51 @@
...
-  TaskID GetID() { return myid_; }
-  TaskID GetDependency() { return dep_; }
-  void SetComplete() { complete_ = true; }
-  bool IsComplete() { return complete_; }
+  Task(const TaskID &id, const TaskID &dep, std::function<TaskStatus()> func)
+      : myid_(id), dep_(dep), type_(TaskType::single), key_(-1), func_(std::move(func)) {}
+  Task(const TaskID &id, const TaskID &dep, std::function<TaskStatus()> func,
+       const TaskType &type, const int key)
+      : myid_(id), dep_(dep), type_(type), key_(key), func_(std::move(func)) {
+    assert(key_ >= 0);
```
jlippuner, line 37: Should we also have `assert(type != TaskType::single);`?

src/tasks/task_types.hpp
```diff
@@ -23,20 +23,51 @@
...
+  }
+  Task(const TaskID &id, const TaskID &dep, std::function<TaskStatus()> func,
+       const TaskType &type, const int key, const int interval)
+      : myid_(id), dep_(dep), type_(type), key_(key), func_(std::move(func)),
+        interval_(interval) {
+    assert(key_ >= 0);
```
jlippuner, line 44: And here too?
And maybe also `assert(interval_ > 0);`?

src/tasks/task_types.hpp
```diff
@@ -23,20 +23,51 @@
...
+  TaskID GetDependency() const { return dep_; }
+  TaskStatus GetStatus() const { return status_; }
+  void SetStatus(const TaskStatus &status) { status_ = status; }
+  TaskType GetType() const { return type_; }
+  int GetKey() const { return key_; }
+  void SetRegional() { regional_ = true; }
+  bool IsRegional() const { return regional_; }
 
  private:
   TaskID myid_, dep_;
```
jlippuner: Why are these not const? I don't understand why we need to be able to change the ID of a task. But in any case, there seems to be no mechanism to change `dep_`, so at least that one can be const, right?

src/tasks/task_types.hpp
```diff
@@ -23,20 +23,51 @@
...
-  bool lb_time, complete_ = false;
+  const TaskType type_;
+  const int key_;
+  TaskStatus status_ = TaskStatus::incomplete;
+  bool regional_ = false;
+  bool lb_time_ = false;
   std::function<TaskStatus()> func_;
+  int calls_ = 0;
+  int interval_ = 1;
```
jlippuner: Looks like `interval_` should be const as well, unless we're missing a `SetInterval` function.

jlippuner (approved):
Looks great. I have some more mostly minor comments/questions.

src/tasks/task_list.hpp
jdolence: wouldn't this just needlessly introduce extra logic elsewhere?

src/tasks/task_list.hpp
jdolence: We've had conversations about this sort of thing in the past.  The conclusion at that time was not to refactor this until someone somewhere actually sees the current implementation meaningfully impact performance.  Certainly it's beyond the scope of this PR, I think.

src/tasks/task_list.hpp
jdolence: It's always set in the constructor -- is that not sufficient?

src/tasks/task_list.hpp
jdolence: I got rid of the `unsigned int`s, but did something a bit simpler.

src/tasks/task_list.hpp
jdolence: Yeah that's right.  If the max iterations were reached, all the tasks get deleted.

src/tasks/task_list.hpp
jlippuner: I think there's only one place where we check whether maximum number of iterations has been reached, isn't there? In that case, we'd just have to check there if MaxIterations is 0 and if so, we never consider max iterations reached. But it's ok to leave it this way if you think it's easier.

src/tasks/task_list.hpp
jlippuner: It is sufficient. This is just a stylistic point here, because all the other fields are having their default values set in this section. would just be nice to be consistent.

src/tasks/task_list.hpp
jlippuner: Fair enough

src/tasks/task_list.hpp
jdolence: tasks only have one status.  This isn't a bad idea, but I'm out of steam on this one.  This function is really only used by someone tinkering with the tasking machinery, so they can just read what's being printed.

jdolence:
@pgrete do you want to have a look at this or should we go ahead and merge?

forrestglines:
> @pgrete do you want to have a look at this or should we go ahead and merge?

@jdolence  I was going to review this one today but I got distracted. I should have my review ready by tomorrow morning.

example/poisson/main.cpp
forrestglines: Is there a time when it should? Like when you attempt to restart a simulation but the last restart dump is already at the final time?

jdolence:
Awesome, thanks @forrestglines!

forrestglines (approved):
@jdolence Looks good. I don't have many suggestions.

I'm pretty excited about this addition. These iterative tasks would enable some IMEX schemes that would otherwise not be possible -- particularly those with a global implicit equation.

jdolence:
Thanks for the look @forrestglines. Regarding the status thing, it looks like there's a path where this should be the return status already, but we're doing something silly in our arg parser.  When you run it and pass a `-h` flag, it's supposed to print a help message and then _should_ return `complete` but seems to return `error` right now.  I'd suggest we just have it properly return `complete` and then this whole thing is a non-issue.

I'm pulling the trigger now.  Thanks everyone!

PR #557, opened 2021-07-09
plug in InitUserMeshBlockData

Yurlungur:
## PR Summary

As discussed in the call today and on the Matrix channel, there's a need for initializing variables and setting grid quantities once per re-meshing, but not more frequently than that. To enable this, I re-enable `InitUserMeshBlockData`, which is set in the application input object. This is now called at the beginning of `Mesh::Initialize` before anything else. `Mesh::Initialize` is called at the beginning of a simulation and upon each remeshing. This means data set in this function is expected to remain constant except when remeshing, and is available to problem generators, boundary conditions, and `FillDerived`.

I also hook it in to the `calculate_pi` example to show it works.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/docs/amr.md
+++ b/docs/amr.md
@@ -34,3 +34,21 @@ In each refinement block, you are required to provide a ``method`` which is a st
 
 ## Package-specific Criteria
 As a package developer, you can define a tagging function that takes a ``Container`` as an argument and returns an integer in {-1,0,1} to indicate the block should be derefined, left alone, or refined, respectively.  This function should be registered in a ``StateDescriptor`` object by assigning the ``CheckRefinement`` function pointer to point at the packages function.  An example is demonstrated [here](../example/calculate_pi/pi.cpp).
+
+## Ensuring your data is consistent after re-meshing
+
+WHen re-meshing happens, a few operations have, which can be plugged in to in various ways. The operations performed (in order) are:
+- The function `InitUserMeshBlockData` is called. This function can be set by setting it in the `ApplicationInputs` field of the problem generator:
+```C++
+void MyInitUserMeshBlockData(MeshBlock *pmb, ParameterInput *pin) {
+  // Do something on a meshblock
+}
+pman.app_input.InitUserMeshBlockData = MyInitUserMeshBlockData;
+// continue with initialization...
+```
+- When the mesh is being generated at initialization, the problem generator is called after every re-meshing.
+- Prolongation, restriction, physical boundaries, and ghost zone communication are performed
+- The `FillDerived` functions set per-package and per-application are called.
+
+If you have a function that you would like called every cycle, you may wish to put it in `FillDerived`.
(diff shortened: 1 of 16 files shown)


Yurlungur:
@par-hermes format

Yurlungur:
I have checked that this fits our needs downstream.

brryan (commented):
Looks good to me. Only comment is would it be worth mentioning this usage of `InitUserMeshBlockData` in e.g. `~/github/parthenon/docs/amr.md`? From the existing documentation it's not immediately apparent (to me at least) that `InitUserMeshBlockData` is called after every remeshing step.

Yurlungur:
Oh good call, @brryan . Yes documentation should be added.

docs/amr.md
```diff
@@ -34,3 +34,21 @@ In each refinement block, you are required to provide a ``method`` which is a st
 
 ## Package-specific Criteria
 As a package developer, you can define a tagging function that takes a ``Container`` as an argument and returns an integer in {-1,0,1} to indicate the block should be derefined, left alone, or refined, respectively.  This function should be registered in a ``StateDescriptor`` object by assigning the ``CheckRefinement`` function pointer to point at the packages function.  An example is demonstrated [here](../example/calculate_pi/pi.cpp).
+
+## Ensuring your data is consistent after re-meshing
+
+WHen re-meshing happens, a few operations have, which can be plugged in to in various ways. The operations performed (in order) are:
```
brryan: `have` -> `happen`. I'm being picky, sorry

Yurlungur: Fixed. :+1:

brryan (approved):

src/application_input.hpp
```diff
@@ -52,7 +52,7 @@ struct ApplicationInput {
   // MeshBlock functions
   std::function<std::unique_ptr<MeshBlockApplicationData>(MeshBlock *, ParameterInput *)>
       InitApplicationMeshBlockData = nullptr;
-  std::function<void(ParameterInput *)> InitUserMeshBlockData = nullptr;
+  std::function<void(MeshBlock *, ParameterInput *)> InitUserMeshBlockData = nullptr;
```
jlippuner: I think passing a pointer to the `MeshBlock` is the right thing to do here. But because we also have a `MeshBlockData` class, the name of the function might be a bit confusing (since it's user data for the `MeshBlock` and not the `MeshBlockData`). What about renaming the function to `InitMeshBlockUserData`?

jlippuner (approved):
Looks great! I just have a small suggestion to rename the function to avoid potential confusion with `MeshBlockData`, but I'll defer to your judgement.

src/application_input.hpp
Yurlungur: I like this idea. I'll make the change.

PR #558, opened 2021-07-09
Boundary bugfix(es) and exposing `FluxDiv_` interface

pgrete:
## PR Summary

Also adjust `dt` calc as the original version resulted in a `nan` when it was initially set to `max`

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


src/driver/driver.cpp
```diff
@@ -156,7 +156,9 @@ void EvolutionDriver::InitializeBlockTimeSteps() {
 void EvolutionDriver::SetGlobalTimeStep() {
   // don't allow dt to grow by more than 2x
   // consider making this configurable in the input
-  tm.dt *= 2.0;
+  if (tm.dt < 0.1 * std::numeric_limits<Real>::max()) {
+    tm.dt *= 2.0;
+  }
```
Yurlungur, line 161: Wow... you hit this? I'm impressed.

src/bvals/boundary_conditions.cpp
```diff
@@ -136,7 +136,7 @@ void OutflowOuterX2(std::shared_ptr<MeshBlockData<Real>> &rc, bool coarse) {
 void OutflowInnerX3(std::shared_ptr<MeshBlockData<Real>> &rc, bool coarse) {
   std::shared_ptr<MeshBlock> pmb = rc->GetBlockPointer();
   auto bounds = coarse ? pmb->c_cellbounds : pmb->cellbounds;
-  int ref = bounds.GetBoundsJ(IndexDomain::interior).s;
+  int ref = bounds.GetBoundsK(IndexDomain::interior).s;
```
Yurlungur: Good catch!

src/driver/driver.cpp
pgrete: Tracking down another problem I used 
```
#include <fenv.h> 
feenableexcept(FE_DIVBYZERO | FE_INVALID | FE_OVERFLOW);
```
and immediately found this one directly during init

Yurlungur: heh nice

src/interface/update.hpp
```diff
@@ -33,6 +33,23 @@ namespace parthenon {
 
 namespace Update {
 
+// Calculate the flux divergence for a specific component l of a variable v
+KOKKOS_FORCEINLINE_FUNCTION
+Real FluxDiv_(const int l, const int k, const int j, const int i, const int ndim,
```
Yurlungur: We should consider renaming `FluxDiv_` to `FluxDiv` or `FluxDivHelper` or something if it's publicly exposed.

Yurlungur:
@pgrete Changing k-bounds initialized the uninitialized variables in the ghost zones at the initial time. I just pushed to your branch.

src/interface/update.hpp
pgrete: Done.

Yurlungur (approved):

pgrete:
What do you think about updating the advection example so that the background has a value of "1" versus "0" currently and then I'd add a `PARTHENON_REQUIRE(advected != 0)` to the `SquareIt` function that is enrolled as `FillDerived`?
This would mean updating the regression data, but I think it's worth adding as is causes almost no overhead an will increase test coverage (namely, that all cells are always initialized).

Yurlungur:
> What do you think about updating the advection example so that the background has a value of "1" versus "0" currently and then I'd add a `PARTHENON_REQUIRE(advected != 0)` to the `SquareIt` function that is enrolled as `FillDerived`?
> This would mean updating the regression data, but I think it's worth adding as is causes almost no overhead an will increase test coverage (namely, that all cells are always initialized).

I think this is a good idea, so long as the advected field now has whatever form it currently has + 1. ;)

jlippuner (approved):
Looks great and great catch!

Since we're making changes to the boundary conditions file anyway, I thought this would be a good opportunity to reduce some code duplication. I've made one generic boundary condition function covering all 12 cases (X1, X2, X3) x (inner, outer) x (outflow, reflection). Hopefully, this would make it easier to avoid such copy&paste errors in the future (if the boundary condition code needs to change). Please let me know what you think.

Yurlungur:
> Looks great and great catch!
> 
> Since we're making changes to the boundary conditions file anyway, I thought this would be a good opportunity to reduce some code duplication. I've made one generic boundary condition function covering all 12 cases (X1, X2, X3) x (inner, outer) x (outflow, reflection). Hopefully, this would make it easier to avoid such copy&paste errors in the future (if the boundary condition code needs to change). Please let me know what you think.

In principal, I'm in favor of reduced code, but I worry that I can't quite parse whether or not the generic version covers all 12 specialized versions. I'm not sure we have an automated test to catch this. Perhaps we should add one.

jlippuner:
> > Looks great and great catch!
> > Since we're making changes to the boundary conditions file anyway, I thought this would be a good opportunity to reduce some code duplication. I've made one generic boundary condition function covering all 12 cases (X1, X2, X3) x (inner, outer) x (outflow, reflection). Hopefully, this would make it easier to avoid such copy&paste errors in the future (if the boundary condition code needs to change). Please let me know what you think.
> 
> In principal, I'm in favor of reduced code, but I worry that I can't quite parse whether or not the generic version covers all 12 specialized versions. I'm not sure we have an automated test to catch this. Perhaps we should add one.

We should definitely have a test. However, I think it should be relatively straightforward to check. There's only a few lines that were different between the expanded 12 cases. That being said, I'm happy to move the code duplication reduction to a new PR if it would take too long to review as part of this PR.

Yurlungur:
>  That being said, I'm happy to move the code duplication reduction to a new PR if it would take too long to review as part of this PR.

That's up to @pgrete I suppose. My preference would be to add an automated test before merging that ensures these are working as intended. Advecting off of each face in 3d would be a. reasonable regression test, I think. We've clearly just been burned by our lack of testing in this area.

src/bvals/boundary_conditions.cpp
```diff
@@ -78,168 +80,100 @@ TaskStatus ApplyBoundaryConditionsOnCoarseOrFine(std::shared_ptr<MeshBlockData<R
 
 namespace BoundaryFunction {
 
-// TODO(JMM): These are all awfully similar. There's gotta be a way to
-// save some code here.
-void OutflowInnerX1(std::shared_ptr<MeshBlockData<Real>> &rc, bool coarse) {
+template <CoordinateDirection DIR, bool INNER, bool REFLECT>
```
pgrete: For the sake of readability, how about using enum classes instead of the `bool`. Given that it's all compile time logic and bool/enum/int typically have the same size, it should also not affect performance.

src/bvals/boundary_conditions.cpp
```diff
@@ -78,168 +80,100 @@ TaskStatus ApplyBoundaryConditionsOnCoarseOrFine(std::shared_ptr<MeshBlockData<R
...
+void GenericBC(std::shared_ptr<MeshBlockData<Real>> &rc, bool coarse) {
+  // make sure DIR is X[123]DIR so we don't have to check again
+  static_assert(DIR == X1DIR || DIR == X2DIR || DIR == X3DIR, "DIR must be X[123]DIR");
+
+  // convenient shorthands
+  const bool X1 = (DIR == X1DIR);
+  const bool X2 = (DIR == X2DIR);
+  const bool X3 = (DIR == X3DIR);
```
pgrete:
```suggestion
  constexpr bool X1 = (DIR == X1DIR);
  constexpr bool X2 = (DIR == X2DIR);
  constexpr bool X3 = (DIR == X3DIR);
```

pgrete (commented):
I like the reduced code (see also comment on readability).
I also think that extending the test coverage is a good idea.
I have an idea to use dc reconstruction with a cfl of 1, so that we an check inflow, outflow, reflect exactly as as there'd be no diffusion.

src/bvals/boundary_conditions.cpp
jlippuner: Good idea. I'll make that change

jlippuner:
Waiting on test for boundary condition before merging

kurtsansom:
What kind of test are you looking for? simple unit test exercising the templated boundary condition function?

Yurlungur:
> What kind of test are you looking for? simple unit test exercising the templated boundary condition function?

Yes exactly. Just to make sure it behaves the same as the older non-templated code. (And/or is correct. ;) )

pgrete:
This is now ready for new comments/review.
I now added checks for the new infrastructure (though user boundary conditions are not covered yet, see below).
I tried to be least intrusive to the advection example.
The key changes are:
- the FillDerived function now has a conditional that checks if all values are initialized (i.e., making sure the corners are filled, too)
- a new regression test, testing periodic, reflecting and outflow conditions enabled
  -  by migrating individual cells in all directions
  -  through an optional `v`(elocity) vector (so that reflecting boundaries are covered)
  - a new flux branch in the flux functions that translates cell data for cfl=1.0 (this works as the content of individual cells don't cross)file:///home/pgrete/test_reflect.mp4

  - the reflection test looks like: 

https://user-images.githubusercontent.com/22685568/133004699-1207cdd0-7073-4cb7-89ca-39be52fb38ef.mp4

Regarding the user boundary conditions:
In principle, I'd like to test all allowed permutations of boundary conditions as that combination of user and outflow threw me off in first place.
From a design point of view, we'd have to extend the advection example with 6 different user functions.
I imagine those just setting the `advected` field to a non-zero value and let the now implemented check in `FillDerived`/`SquareIt` handle it.
Do you think it's worth adding this change here or should we merge this sooner than later and take care of it in a separate PR?

pgrete:
Ping @Yurlungur  for final re-review of latest changes as discussed on Thursday.

example/advection/advection_package.cpp
```diff
@@ -171,11 +176,18 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
...
+                  Metadata::FillGhost, Metadata::Vector},
+                 std::vector<int>({3}), std::vector<std::string>{"vx", "vy", "vz"});
+    pkg->AddField(std::string("v"), m);
+  }
   if (fill_derived) {
     field_name = "one_minus_advected";
     m = Metadata({Metadata::Cell, Metadata::Derived, Metadata::OneCopy},
                  std::vector<int>({num_vars}));
     pkg->AddField(field_name, m);
+    pkg->AllFields();
```
Yurlungur: What is this line supposed to be doing? Doesn't this return the const reference to the dictionary? Why the empty call?

example/advection/advection_package.cpp
```diff
@@ -304,6 +316,25 @@ void SquareIt(MeshBlockData<Real> *rc) {
...
         v(out + n, k, j, i) = v(in + n, k, j, i) * v(in + n, k, j, i);
       });
+
+  // The following block/logic is also just added for regression testing.
+  // More specifically, the "smooth_gaussian" profile is initially != 0 everywhere, but
+  // initialializes IndexDomain::interior.
+  // FillDerived (here, SquareIt) is called after the ghost cells are exchanged and over
+  // IndexDomain::entire.
+  // Thus, no 0 (from initializing Kokkos views) should be left if all faces/corners/edges
+  // are correct, which is what we check in the loop below.
```
Yurlungur, line 325: High value comment :+1:

example/advection/advection_package.cpp
```diff
@@ -449,15 +487,23 @@ TaskStatus CalculateFluxes(std::shared_ptr<MeshBlockData<Real>> &rc) {
...
+              } else {
+                v.flux(X1DIR, n, k, j, i) = qr(n, i) * vx;
+              }
+              // Custom flux function to move isolated, cells around. Just used for
+              // bvals testing.
+            } else {
+              v.flux(X1DIR, n, k, j, i) =
+                  ql(idx_v, i) > 0.0 ? ql(n, i) * ql(idx_v, i) : 0.0;
+              v.flux(X1DIR, n, k, j, i) +=
+                  qr(idx_v, i) < 0.0 ? qr(n, i) * qr(idx_v, i) : 0.0;
```
Yurlungur, line 503: I like the `+=` syntax here. Feels nicer than the if-else block.

example/advection/advection_package.cpp
```diff
@@ -481,15 +527,23 @@ TaskStatus CalculateFluxes(std::shared_ptr<MeshBlockData<Real>> &rc) {
...
+                if (vy > 0.0) {
+                  v.flux(X2DIR, n, k, j, i) = ql(n, i) * vy;
+                } else {
+                  v.flux(X2DIR, n, k, j, i) = qr(n, i) * vy;
+                }
+                // Custom flux function to move isolated, cells around. Just used for
+                // bvals testing.
+              } else {
+                v.flux(X2DIR, n, k, j, i) =
+                    ql(idx_v + 1, i) > 0.0 ? ql(n, i) * ql(idx_v + 1, i) : 0.0;
```
Yurlungur: rather than `+1` or `+2`, one could do `idx_v + XNDIR - 1` which might be a little cleaner. Not a big deal though.

example/advection/advection_package.cpp
```diff
@@ -53,7 +58,7 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
 
   auto profile_str = pin->GetOrAddString("Advection", "profile", "wave");
   if (!((profile_str == "wave") || (profile_str == "smooth_gaussian") ||
-        (profile_str == "hard_sphere"))) {
+        (profile_str == "hard_sphere") || (profile_str == "block"))) {
```
Yurlungur, line 61: I don't love the name `block` for the test profile. But I don't have a better name either.

tst/regression/test_suites/bvals/bvals.py
```diff
@@ -0,0 +1,129 @@
...
+import utils.test_case
+
+""" To prevent littering up imported folders with .pyc files or __pycache_ folder"""
+sys.dont_write_bytecode = True
+
+
+class TestCase(utils.test_case.TestCaseAbs):
+    def Prepare(self, parameters, step):
+
+        # Step 1 (default vals in parameter file) is reflecting BC
```
Yurlungur: I know it's default to reflecting, but I still might be explicit here, in case the parameter file changes.

Yurlungur (approved):
LGTM @pgrete some minor comments below. Really fantastic to have this test in place, along with the reduced code.

example/advection/advection_package.cpp
pgrete: Good catch. Looks like a debugging leftover. I'm removing it.

tst/regression/test_suites/bvals/bvals.py
pgrete: Good point. Will do.

example/advection/advection_package.cpp
pgrete: Done.

pgrete:
Addressed comment and triggering automerge now.

Issue #547, opened 2021-06-17
Clean up particles implementation


brryan:
Work constraints and the size of PR #404 motivated merging the particles MPI capability into `develop` while some of the code isn't yet fully cleaned up. As @pgrete points out, some outstanding issues are

 - [x] `InitializeCommunicationMesh` should be a `Swarm` method
 - [ ] `FinalizeCommunicationMesh` should be a `Swarm` method
 - [x] `Defrag` could just be a task in `Swarm`, probably
 - [ ] `SwarmContainer` should almost certainly be removed
 - [x] `Metadata::Provides` is not being correctly set by default for `Swarm`s

I should make these changes before adding new particles features.

brryan:
I made some progress on this but there are a few sticking points at the moment:
 - I remade `InitializeCommunicationMesh` as a `Swarm` method `ResetCommunication` but I can't remake the similar `FinalizeCommunicationMesh` until #582 gets merged in.
 - The `Metadata` issue is weird. I switched to using `Metadata::None` as the brace initializer contents for `Metadata`. It's not enough to just provide an empty `{}` to the constructor because this is ambiguous. I'm not sure what the best path forward right now is... 
 -  I'd like to think more about `Swarm`s and `MeshBlockPack`s before I modify the `SwarmContainer` infrastructure. Should `Swarm`s be held by the `mesh_data` member variable? Etc.

PR #563, opened 2021-07-15
Physical boundary options for particles

brryan:
## PR Summary

Up until now periodic boundaries have been hardcoded for particles. We want to relax this assumption.

The approach I adopted in this PR was to derive classes for specific boundary conditions based on a generic boundary conditions object, allocate these on the device, and then call the specific boundary condition method from the set of allocated objects. Whether this is the most performant way to do this I'm not sure, but I wasn't sure what else to do while keeping the boundary conditions settable at runtime, except maybe for templating swarms on six boundary condition values (!)

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Fix issues in #547


brryan:
@pgrete I didn't resolve all the issues in #547 yet but the items still open there merit further discussion I think.

CHANGELOG.md
```diff
@@ -76,11 +77,12 @@ Date: 03/30/2021
...
 - [[PR 440]](https://github.com/lanl/parthenon/pull/440) Add abstraction for allocating a `unique_ptr` to an object in device memory
+- [[PR 439]](https://github.com/lanl/parthenon/pull/439) Add split initialization of environment (`ParthenonInitEnv()`) and packages and mesh (`ParthenonInitPackagesAndMesh()`) to `ParthenonManager`
 - [[PR 438]](https://github.com/lanl/parthenon/pull/438) More diagnostic runtime output (AMR/Loadbalance and mesh structure) controlled via `parthenon/time/ncycle_out_mesh` input parameter (default 0 - off)
 - [[PR 412]](https://github.com/lanl/parthenon/pull/412) Add capability to use host (pinned) memory for communication buffers (via `PARTHENON_ENABLE_HOST_COMM_BUFFERS` - default OFF)
+- [[PR 406]](https://github.com/lanl/parthenon/pull/406) Add `stochastic_subgrid` example that performs a random amount of work per cell (drawn from a power law distribution)
+- [[PR 404]](https://github.com/lanl/parthenon/pull/404) Add capability to communicate particles across meshblocks/MPI processes
```
forrestglines: Do the PR tags need to be in order? That isn't necessarily the chronological order in which they were added. Doesn't make a big difference though.

src/interface/swarm.cpp
```diff
@@ -404,6 +492,15 @@ void Swarm::SetNeighborIndices1D_() {
...
 
+  // Initialize array in event of zero neighbors
+  for (int k = 0; k < 4; k++) {
+    for (int j = 0; j < 4; j++) {
+      for (int i = 0; i < 4; i++) {
+        neighborIndices_h(k, j, i) = 0;
+      }
+    }
+  }
+
```
forrestglines, line 493: The array isn't filled by zeros by default, nor is there a "Fill" function? It doesn't matter that much through

src/interface/swarm_boundaries.hpp
```diff
@@ -0,0 +1,199 @@
...
+
+class ParticleBoundIX1Reflect : public ParticleBound {
+ public:
+  KOKKOS_INLINE_FUNCTION void Apply(const int n, double &x, double &y, double &z,
+                                    const SwarmDeviceContext &swarm_d) const override {
+    if (x < swarm_d.x_min_global_) {
+      x = swarm_d.x_min_global_ + (swarm_d.x_min_global_ - x);
+    }
+  }
+};
```
forrestglines: Where does the velocity get reversed upon reflection?

What happens when a particle exits through a domain edge or corner?

forrestglines (approved):
Different boundary conditions for particles is much appreciated.

I had one comment about how velocities are handled with reflecting boundary conditions. I believe velocities also have to be reflected along with positions.

CHANGELOG.md
brryan: Hm that's odd, I didn't reorder the `CHANGELOG` myself, only added this commit to it. Let me switch these back just because this isn't an intended change.

brryan: Also why this PR adds `404`, particle MPI communication, to the Release 0.5.0 section, I have no idea (it's already listed in the current develop section)

src/interface/swarm_boundaries.hpp
brryan: Oops, good catch! Yes this is an error. Let me see about fixing this and then adding a unit test for the boundaries...

src/interface/swarm.cpp
brryan: I don't think we can assume `ParArrayND`s (or their `host` equivalents) are filled with zeros, and I don't know that we have a `Fill` function. `Fill` might be nice to have at some point.

src/interface/swarm_boundaries.hpp
brryan: OK this is actually more subtle, because particle velocity isn't a default swarm parameter (only `x`, `y`, and `z` are). So parthenon itself doesn't know which variables to change the direction of. I've removed the default reflecting boundary condition and these will now need to be implemented manually. Reintroducing default reflecting boundaries would probably require adding default velocity variables to `Swarm`s. 

Also, I couldn't find a clear path forward for implementing unit tests of particle boundaries. The `Swarm::ApplyBoundaries_` function requires a large amount of parthenon to be built up because it depends on `bvals`, and I'm not sure just testing the boundary classes themselves is particularly useful since they're so simple (I'd be more worried about how e.g. particles the move across corners are moved back onto the mesh by multiple boundary calls, but this is already captured by the `periodic` boundaries in the particles examples, at least if there are enough particles...)

src/interface/swarm.cpp
brryan: @Yurlungur alerted me to [this](https://github.com/kokkos/kokkos/wiki/View#642-initialization): `A View's entries are initialized to zero by default.` We're a little close to magic number territory here though (`-1` having a special meaning) so at the moment I just feel more comfortable being explicit here if thats OK

example/particle_tracers/particle_tracers.cpp
```diff
@@ -150,7 +150,7 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
 
   // Add swarm of tracer particles
   std::string swarm_name = "tracers";
-  Metadata swarm_metadata({Metadata::Provides});
+  Metadata swarm_metadata({Metadata::None});
```
Yurlungur: why this change?

src/interface/swarm.cpp
```diff
@@ -35,6 +38,12 @@ SwarmDeviceContext Swarm::GetDeviceContext() const {
   const IndexRange &ib = pmb->cellbounds.GetBoundsI(IndexDomain::interior);
   const IndexRange &jb = pmb->cellbounds.GetBoundsJ(IndexDomain::interior);
   const IndexRange &kb = pmb->cellbounds.GetBoundsK(IndexDomain::interior);
+  context.ib_s_ = ib.s;
+  context.jb_s_ = jb.s;
+  context.kb_s_ = kb.s;
```
Yurlungur, line 40: why not put the whole index bounds into the context?

src/interface/swarm.cpp
```diff
@@ -35,6 +38,12 @@ SwarmDeviceContext Swarm::GetDeviceContext() const {
...
+  context.dx1_ = (pmb->coords.x1f(ib.e + 1) - pmb->coords.x1f(ib.s)) / (ib.e - ib.s + 1);
+  context.dx2_ = (pmb->coords.x2f(jb.e + 1) - pmb->coords.x2f(jb.s)) / (jb.e - jb.s + 1);
+  context.dx3_ = (pmb->coords.x3f(kb.e + 1) - pmb->coords.x3f(kb.s)) / (kb.e - kb.s + 1);
```
Yurlungur: Why not just use `coords.Dx(direction)`? Heck, you could put the coords object in the context too.

src/interface/swarm.cpp
```diff
@@ -78,6 +87,73 @@ Swarm::Swarm(const std::string &label, const Metadata &metadata, const int nmax_
...
+  } else if (bcs[5] != BoundaryFlag::user) {
+    msg << "ox3 boundary flag " << static_cast<int>(bcs[5]) << " not supported!";
+    PARTHENON_THROW(msg);
+  }
+
+  for (int n = 0; n < 6; n++) {
+    pbounds.bounds[n] = bounds[n].get();
+    PARTHENON_REQUIRE(pbounds.bounds[n] != nullptr,
+                      "Null device boundary condition pointer!");
+  }
```
Yurlungur: I think this error message could be more useful for the confused downstream user. Something like:
```suggestion
  for (int n = 0; n < 6; n++) {
    pbounds.bounds[n] = bounds[n].get();
    std::stringstream msg;
    msg << "Boundary condition on face " << n << " missing.\n"
           << "Please set it to `outflow`, `periodic`, or `user` in the input deck.\n"
           << "If you set it to user, you must also manually set "
           << "the swarm boundary pointer in your application."
           << std::endl;
    PARTHENON_REQUIRE(pbounds.bounds[n] != nullptr, msg);
  }
```

src/interface/swarm.cpp
```diff
@@ -775,14 +873,50 @@ void Swarm::LoadBuffers_(const int max_indices_size) {
...
+    int total_sent_particles = 0;
+
+    for (int n = 0; n <= max_active_index_; n++) {
+      if (mask_h(n)) {
+        // This particle should be "sent"
+        if (blockIndex_h(n) >= 0) {
+          total_sent_particles++;
+        }
+      }
+    }
```
Yurlungur: I don't fully understand what's going on here. Some indexing logic is being done on the host, but then the actual boundary conditions are applied on device?

src/interface/swarm_boundaries.hpp
```diff
@@ -0,0 +1,139 @@
...
+#define INTERFACE_SWARM_BOUNDARIES_HPP_
+
+#include "swarm_device_context.hpp"
+
+namespace parthenon {
+
+class ParticleBound {
+ public:
+  KOKKOS_INLINE_FUNCTION virtual void Apply(const int n, double &x, double &y, double &z,
+                                            // const SwarmDeviceContext &context) = 0;
```
Yurlungur: why not pure?

src/interface/swarm_boundaries.hpp
```diff
@@ -0,0 +1,139 @@
...
+                                            const SwarmDeviceContext &context) const {
+    PARTHENON_FAIL("Calling base class");
```
Yurlungur:
```suggestion
    PARTHENON_FAIL("You are calling the ParticleBound abstract base class Apply method");
```

Yurlungur (commented):
Overall LGTM @brryan . A few comments below. And also one big question:

How do you expect downstream codes to set boundaries in a swarm? Swarms themselves are only created long after `Package::Initialize`. Should we have boundaries be set in `ApplicationInputs`? Is this already done and I missed it?

Yurlungur (approved):

example/particle_tracers/particle_tracers.cpp
brryan: Right now we don't use any specific metadata for swarms, but `Provides` is required. This is normally the default, but `Metadata({})` is ambiguous and a compiler error so we need to pass *something*. `None` is correct since this doesn't have cell-specific data, but `Provides` is also correct... Honestly I'm not sure what the best solution to this right now is. We may ultimately want `Metadata` in `Swarm`s at least for the `Provides`/`Requires` flags with multiple packages.

Yurlungur: Hmm... what about being just extremely explicit. I.e., `swarm_metadata({Metadata::Provides, Metadata::None});`? That way it's clear it's both things?

pgrete: +1 on that suggestion (here and above and below). It would also be in line with the other variable handling in packages (in the sense that this swarm provides the "tracers" particles).

src/interface/swarm.cpp
pgrete: +1 for the suggestion.

src/interface/swarm.cpp
pgrete: I think this suggestions is actually a requirement.
Once we have more than one options for coordinate system this hard coded version would not be correct any more.

pgrete: Having said that I now have an additional comment: If constant dx are an assumption then it'd be great to have sth along `PARTHENON_REQUIRE(uniform_cartessian)` at the appropriate place.
I prefer to die loud when called in an unintended/not tested case.

src/interface/swarm.cpp
pgrete: In addition, why not execute this loop on the device?
Both working arrays are already on the device so this could be a `par_reduce`or am I missing sth?

src/interface/swarm.cpp
```diff
@@ -775,14 +873,50 @@ void Swarm::LoadBuffers_(const int max_indices_size) {
...
 
-  // Prepare buffers for send operations
-  LoadBuffers_(max_indices_size);
+    if (total_sent_particles > 0) {
+      ParArrayND<int> new_indices("new indices", total_sent_particles);
+      int sent_particle_index = 0;
+      for (int n = 0; n <= max_active_index_; n++) {
+        if (mask_h(n)) {
+          if (blockIndex_h(n) >= 0) {
+            new_indices(sent_particle_index) = n;
```
pgrete: Does this work on a device? `ParArrayND`  should be allocated by default in device memory, but this loop is not a kernel.
If it passes so far, then this may hint at a test case not covered yet.

src/interface/swarm.cpp
```diff
@@ -775,14 +873,50 @@ void Swarm::LoadBuffers_(const int max_indices_size) {
...
+    // Prepare buffers for send operations
+    LoadBuffers_(max_indices_size);
+
+    // Send buffer data
+    vbswarm->Send(phase);
+  }
 
-  // Send buffer data
-  vbswarm->Send(phase);
   return true;
```
pgrete: This function always reaches this point and returns true, doesn't it?  So could be made `void` instead of `bool`.

src/interface/swarm.cpp
```diff
@@ -861,58 +995,86 @@ void Swarm::UnloadBuffers_() {
...
 
-  // Populate buffers
-  vbswarm->Receive(phase);
+  if (nneighbor == 0) {
+    // Do nothing; no boundaries to receive
+    return true;
+  } else {
+    // Ensure all local deep copies marked BoundaryStatus::completed are actually received
+    GetBlockPointer()->exec_space.fence();
+    auto pmb = GetBlockPointer();
```
```suggestion
    pmb->exec_space.fence();
```
`pmb` is already defined outside the conditional, isn't it?

src/interface/swarm.cpp
```diff
@@ -861,58 +995,86 @@ void Swarm::UnloadBuffers_() {
...
 
-  // Transfer data from buffers to swarm memory pool
-  UnloadBuffers_();
+    return all_boundaries_received;
+  }
+}
 
-  auto &bdvar = vbswarm->bd_var_;
-  bool all_boundaries_received = true;
+bool Swarm::ResetCommunication() {
```
pgrete: Why `bool`? The function always returns `true`, doesn't it?

src/interface/swarm.cpp
```diff
@@ -19,6 +19,9 @@
 #include "mesh/mesh.hpp"
 #include "swarm.hpp"
 
+using Kokkos::kokkos_malloc;
+using Kokkos::parallel_for;
+
```
```suggestion
```
Not used in this file.

src/interface/swarm.hpp
```diff
@@ -236,9 +193,17 @@ class Swarm {
   int num_particles_sent_;
   bool finished_transport;
 
+  // Class to store raw pointers to boundary conditions on device. Copy locally for
+  // compute kernel capture.
+  ParticleBoundaries pbounds;
+
```
pgrete: Does this work as intended on a device? IIRC member variables are not captured automatically. Maybe I also haven't fully understood the difference between `pbounds.bounds` and `bounds`.

src/interface/swarm_boundaries.hpp
```diff
@@ -0,0 +1,139 @@
...
+      x = swarm_d.x_max_global_ - (swarm_d.x_min_global_ - x);
+    }
+  }
+};
+
+class ParticleBoundIX1Outflow : public ParticleBound {
+ public:
+  KOKKOS_INLINE_FUNCTION void Apply(const int n, double &x, double &y, double &z,
+                                    const SwarmDeviceContext &swarm_d) const override {
+    swarm_d.MarkParticleForRemoval(n);
```
pgrete: Why is there no check required if the particle actually crossed the boundary?
If the function is only called if a particle already crossed it, then why is the check required for periodic boundary conditions?

src/interface/swarm_container.cpp
```diff
@@ -74,6 +74,24 @@ void SwarmContainer::Remove(const std::string &label) {
...
 
+TaskStatus SwarmContainer::Defrag(double min_occupancy) {
+  Kokkos::Profiling::pushRegion("Task_SwarmContainer_Defrag");
+  PARTHENON_REQUIRE_THROWS(min_occupancy >= 0. && min_occupancy <= 1.,
+                           "Max fractional occupancy of swarm must be >= 0 and <= 1");
+
+  for (auto &s : swarmVector_) {
+    s->SetupPersistentMPI();
+    if (s->GetNumActive() > 0 &&
+        s->GetNumActive() / (s->GetMaxActiveIndex() + 1) < min_occupancy) {
```
pgrete: Does this work as intended?
`s->GetNumActive() / (s->GetMaxActiveIndex() + 1)` is integer arithmetic, isn't it?

src/interface/swarm_device_context.hpp
```diff
@@ -0,0 +1,112 @@
...
+
+    return blockIndex_(n);
+  }
+
+  KOKKOS_INLINE_FUNCTION
+  int GetMyRank() const { return my_rank_; }
+
+  KOKKOS_INLINE_FUNCTION
+  void Xtoijk(const Real &x, const Real &y, const Real &z, int &i, int &j, int &k) const {
+    if (ndim_ > 0) {
```
pgrete: Is there a case where `ndim_` is not `>0`?

src/interface/swarm_device_context.hpp
```diff
@@ -0,0 +1,112 @@
...
+          k = static_cast<int>(std::floor((z - z_min_) / dx3_)) + kb_s_;
+        } else {
+          k = kb_s_;
+        }
+      } else {
+        j = jb_s_;
+      }
+    } else {
+      i = ib_s_;
+    }
```
pgrete: Does the logic here work as intended?
For example, `k` is only set if `ndim_ > 2`, but even in a 2d sim, we still carry around the third dimension in terms of indices (but `k=0` in that case).
So I had expected a function similar to
```suggestion
      i = static_cast<int>(std::floor((x - x_min_) / dx1_)) + ib_s_;
      j = (ndim_ > 1) ? static_cast<int>(std::floor((y - y_min_) / dx2_)) + jb_s_ : jb_s_;
      k = (ndim_ > 2) ? static_cast<int>(std::floor((z - z_min_) / dx3_)) + kb_s_ : kb_s_;
```

However, if the function is as intended then some inline comments would be great to highlight that it's okay for some vars to not being set.

src/interface/swarm_device_context.hpp
```diff
@@ -0,0 +1,112 @@
...
+    is_on_current_mesh_block = (blockIndex_(n) == this_block_);
+
+    return blockIndex_(n);
+  }
+
+  KOKKOS_INLINE_FUNCTION
+  int GetMyRank() const { return my_rank_; }
+
+  KOKKOS_INLINE_FUNCTION
+  void Xtoijk(const Real &x, const Real &y, const Real &z, int &i, int &j, int &k) const {
```
pgrete, line 69: Following up on my earlier comment: This function could be templated (or otherwise safeguarded) for use only with Cartesian coordinates.

src/kokkos_abstraction.hpp
```diff
@@ -785,6 +785,16 @@ std::unique_ptr<T, DeviceDeleter<MS>> DeviceCopy(const T &host_object) {
...
+template <typename T, typename ES = DevExecSpace, typename MS = DevMemSpace>
+T *RawDeviceCopy(const T &host_object) {
+  auto p = static_cast<T *>(Kokkos::kokkos_malloc<MS>(sizeof(T)));
+  Kokkos::parallel_for(
+      Kokkos::RangePolicy<ES>(0, 1),
+      KOKKOS_LAMBDA(const int i) { new (p) T(host_object); });
+  Kokkos::fence();
+  return p;
+}
+
```
pgrete: Is this called somewhere?

pgrete (changes requested):
Thanks for working on the open items @brryan 
I have a couple of suggestions and questions that would be great to get replies to before merging.
I'll monitor the PR so that I can react to your replies on short notice.

src/interface/swarm.cpp
brryan: I have ongoing concerns about the size of this context object, and since we don't need `*b.e`, I'd rather leave them out until we do need them (or the whole device context picture changes...)

example/particle_tracers/particle_tracers.cpp
brryan: Switched to `swarm_metadata({Metadata::Provides, Metadata::None});` for all three particles examples

src/interface/swarm.cpp
brryan: Yes I think we need to be carrying the coords object around in general. I made this change. 

When more grids are added we will need to revisit the swarm design to accomodate them but at the moment I don't think we have a flag to check in a `PARTHENON_REQUIRE(uniform_cartesian);` statement. Such a check should definitely go in at that time.

src/interface/swarm.cpp
brryan: Yes good point! This should just be a `par_reduce`. Unfortunately the logic for filling the elements in `new_indices` is serial at the moment (hence done on the host), I would like to find a smarter way to get that onto the device as well at some point...

src/interface/swarm.cpp
brryan: Good catch. Yes I just noticed in working out the `par_reduce` from the last comment I addressed that this code really isn't run anywhere right now. If any of your boundaries are periodic `nneighbor > 0` so the code uses the full MPI buffer logic. 

This logic is tested if I change all the boundaries to `outflow`. I think I need to add another particles regression test with `particle_leapfrog`, set all the boundaries to outflow, and run it for long enough that some of the particles leave the domain and are removed.

src/interface/swarm.cpp
brryan: Nice catch, thanks. Updated here and in `swarm_container.cpp`.

src/interface/swarm.cpp
brryan: Fixed

src/interface/swarm.hpp
brryan: My understanding is member variables are captured automatically (e.g. how `SwarmDeviceContext` puts a bunch of data in a struct that is then all copied onto the device when captured). If this weren't working on device it should fail when the particles regression is run on GPU on the darwin CI right?

`pbounds.bounds` is the array of device pointers to `ParticleBoundary` objects, and `bounds` is the array of `unique_ptr`s that own those pointers and handle automatic deletion at cleanup time. I'll rename `Swarm::bounds` to... um... `bounds_uptrs` and `pbounds` to `bounds_d` (in analogy to `swarm_d`) and `ParticleBoundaries` to `BoundaryDeviceContext`?

src/interface/swarm_boundaries.hpp
brryan: Yes good catch this logic is wrong, I need to add the checks. This will also be exercised with a modified `particle_leapfrog` regression test where only *some* of the particles should cross the boundaries and be removed.

src/interface/swarm_container.cpp
brryan: Whoa nice catch! Yes this was broken. I fixed it (`1` -> `1.0`) and checked the result with `printf`s.

brryan: one thumbs up isnt enough :+1: :+1:

src/interface/swarm.cpp
Yurlungur: Yeah ok fair enough.

src/interface/swarm_device_context.hpp
brryan: No -- the relevant line in `mesh/mesh.cpp`: `ndim((mesh_size.nx3 > 1) ? 3 : ((mesh_size.nx2 > 1) ? 2 : 1))`. I'll remove this conditional.

src/interface/swarm_device_context.hpp
brryan: Yes you are right, in 1D sims `k` was not being set which is bad. I could have added a `k = kb_s_;` below the `j = jb_s_` line but your solution is much cleaner so I've gone ahead with that :)

src/kokkos_abstraction.hpp
brryan: No but I figured it might be useful in downstream codes (at one point I found myself in a spot where getting a `unique_ptr` back was frustrating since there it was returned as a temporary object). I don't think I'm using that anymore if you'd rather remove this.

src/interface/swarm_boundaries.hpp
brryan: I thought that would work but when I do that, meaning
```c++
 KOKKOS_INLINE_FUNCTION virtual void Apply(const int n, double &x, double &y, double &z,
                                             const SwarmDeviceContext &context) = 0
 ```
and remove the `override`s from the derived classes the compiler thinks the derived classes are abstract for some reason:
```
/home/brryan/github/parthenon/src/kokkos_abstraction.hpp:769:67: error: invalid new-expression of abstract class type 'parthenon::ParticleBoundIX2Outflow'
  769 |       Kokkos::RangePolicy<ES>(0, 1), KOKKOS_LAMBDA(const int i) { new (p) T(); });
      |                                                                   ^~~~~~~~~~~
In file included from /home/brryan/github/parthenon/src/interface/swarm.hpp:37,
                 from /home/brryan/github/parthenon/src/interface/state_descriptor.hpp:29,
                 from /home/brryan/github/parthenon/src/application_input.hpp:24,
                 from /home/brryan/github/parthenon/src/mesh/mesh.hpp:33,
                 from /home/brryan/github/parthenon/src/interface/swarm.cpp:19:
/home/brryan/github/parthenon/src/interface/swarm_boundaries.hpp:79:7: note:   because the following virtual functions are pure within 'parthenon::ParticleBoundIX2Outflow':
   79 | class ParticleBoundIX2Outflow : public ParticleBound {
      |       ^~~~~~~~~~~~~~~~~~~~~~~
/home/brryan/github/parthenon/src/interface/swarm_boundaries.hpp:22:39: note: 	'virtual void parthenon::ParticleBound::Apply(int, double&, double&, double&, const parthenon::SwarmDeviceContext&)'
   22 |   KOKKOS_INLINE_FUNCTION virtual void Apply(const int n, double &x, double &y, double &z,
      |
```
Do you recognize that?

File: src/interface/swarm.cpp (hunk 14 above)
Comment (brryan) [reply src/interface/swarm.cpp]: Yep, sloppiness on my part, thanks -- fixed

File: src/interface/swarm_device_context.hpp (hunk 21 above)
Comment (brryan) [reply src/interface/swarm_device_context.hpp:69]: Yes when we go beyond uniform cartesian grids this will need to be redesigned... but depending on what profiling looks like this approach to the swarm device context may need to be redesigned first anyway. 

A safeguard would be good to add now, I just added 
```c++
  PARTHENON_REQUIRE_THROWS(typeid(Coordinates_t) == typeid(UniformCartesian),                        
    "SwarmDeviceContext only supports a uniform Cartesian mesh!");
```
to the `Swarm` constructor, does that look good to you? (Can't use `typeid` in device code i.e. inside `SwarmDeviceContext` calls)

src/interface/swarm.cpp
brryan: OK, added a `particle_leapfrog_outflow` test. It runs for 25 cycles on one MPI rank with all outflow boundaries and two of the particles leave the domain but two stick around.

src/interface/swarm.cpp
pgrete: We'll only have one device context per swarm per block, won't we? In that case it's probably a long way until the size becomes a bottleneck compared to the other data.
I'm not suggesting to make this change here, but probably sooner than later (for consistency reasons within the code base).

src/interface/swarm.cpp
```diff
@@ -861,58 +998,84 @@ void Swarm::UnloadBuffers_() {
...
+  const int nneighbor = pmb->pbval->nneighbor;
 
-  // Populate buffers
-  vbswarm->Receive(phase);
+  if (nneighbor == 0) {
+    // Do nothing; no boundaries to receive
+    return true;
+  } else {
+    // Ensure all local deep copies marked BoundaryStatus::completed are actually received
+    GetBlockPointer()->exec_space.fence();
```
```suggestion
    pmb->exec_space.fence();
```

src/interface/swarm_device_context.hpp
pgrete: Yes, I think a single check early on is fine.
Potentially you could add comments to the places in the code that'd need to be changed (right now, after you wrote the code they may be more obvious than a couple to hours/days/weeks/... down the road when make/need that change), but it's just a suggestion.

src/interface/swarm_device_context.hpp
```diff
@@ -0,0 +1,108 @@
...
+  Real y_min_;
+  Real y_max_;
+  Real z_min_;
+  Real z_max_;
+  Real x_min_global_;
+  Real x_max_global_;
+  Real y_min_global_;
+  Real y_max_global_;
+  Real z_min_global_;
+  Real z_max_global_;
```
pgrete, line 98: I was actually wondering if we need those or if the local ones are fine.
The boundary functions are only called for blocks that contain a boundary, so for those `x_max` is `x_max_global` and for all other blocks the boundary function the global numbers don't matter, do they?

src/kokkos_abstraction.hpp
pgrete: If you already use it downstream, then we should keep it (and somehow test it).
If it's not used I'm in favor if removing it so that it can be introduced/tested when required.

src/interface/swarm_device_context.hpp
```diff
@@ -0,0 +1,108 @@
...
+    return blockIndex_(n);
+  }
+
+  KOKKOS_INLINE_FUNCTION
+  int GetMyRank() const { return my_rank_; }
+
+  KOKKOS_INLINE_FUNCTION
+  void Xtoijk(const Real &x, const Real &y, const Real &z, int &i, int &j, int &k) const {
+    i = static_cast<int>(
+            std::floor((x - x_min_) / coords_.Dx(CoordinateDirection::X1DIR))) +
```
pgrete, line 71: This works for uniform cartesian.
I suggest (again not blocking), to use `coords_.dx1v(i)` (and similarly `coords_dx2v(j)` and `coords_dx3v(k)`) as its the more common usage (at the same computational cost).

pgrete (approved):
Thanks for the update. Looks good to me!

src/kokkos_abstraction.hpp
brryan: You're right, this shouldn't be floating around untested. I'm not using this now so I've removed it.

src/interface/swarm_device_context.hpp
brryan: months/years/... :) Yes, good idea, I've added some TODOs in the relevant parts of `SwarmDeviceContext` for later on.

src/interface/swarm_device_context.hpp
brryan: I'm not sure we can do the usual `coords_.dx1v(i)` here, since `i` is what we're calculating. Technically it would work I think because `i`, `j`, `k` aren't actually used in `uniform_cartesian.hpp`'s `dx*v()` functions I don't think, but it could be quite confusing. Since the behavior is a bit different I slightly prefer keeping the unusual `Dx()` call here.

Thinking more on this `Xtoijk` in the general case would have to become a nasty binary search over the dx's of each zone in each direction to figure out what your `i` actually is... an alternative would be to have each particle keep a record of the zone index its currently in, and require particle push functions to update that appropriately. What this actually makes me think is that particle support for non-uniform meshes will be even further out in the future.

src/interface/swarm.cpp
brryan: Nice catch, thanks

src/interface/swarm_boundaries.hpp
brryan: Checks added, regression test added

src/interface/swarm_device_context.hpp
brryan: My reading is the boundary functions are called for all blocks regardless of whether there is a physical boundary. Either the boundary is physical or MPI, and we need to call the boundary functions after at least some `Receive`s if, for example, a particle is communicated from the X1R boundary periodically to the X1L boundary and needs its position updated from xmax + delta to xmin + delta. For that, these variables are explicitly used in `swarm_boundaries.hpp` e.g. `x = swarm_d.x_max_global_ - (swarm_d.x_min_global_ - x);`

src/interface/swarm_boundaries.hpp
Yurlungur: I think you still need the `override` but you can let the base method be pure.

example/particle_leapfrog/particle_leapfrog.cpp
```diff
@@ -67,7 +67,7 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
   pkg->AddParam<>("write_particle_log_nth_cycle", write_particle_log_nth_cycle);
 
   std::string swarm_name = "my particles";
-  Metadata swarm_metadata({Metadata::Provides});
+  Metadata swarm_metadata({Metadata::Provides, Metadata::None});
```
Yurlungur, line 70: :+1:

src/interface/swarm_device_context.hpp
Yurlungur: Not now, but maybe for the future, `Xtoijk` should be part of the `Coords` class. Then the implementation details can be punted to the coordinate system.

Yurlungur (approved):
LGTM

src/interface/swarm_device_context.hpp
brryan: Yeah good idea

pgrete: My bad. Using `dx1v` defeats the purpose here. I agree that it'd be nice to have this functionality in `Coords` at some point.

Issue #571, opened 2021-07-23
Full Regression test suite currently failing due to sparse update


Yurlungur:
As discussed on the Parthenon call. @jlippuner to fix.

pgrete:
@jlippuner do you already know what's going on?
I saw that you started looking into this and I'm not sure I understand where the issue is.
On my machine the tests pass.

jlippuner:
> @jlippuner do you already know what's going on?
> I saw that you started looking into this and I'm not sure I understand where the issue is.
> On my machine the tests pass.

Yes, I'm looking into it. I have a training workshop today, so might not be able to get to it until tomorrow. It's not clear to me either, what's going on. There's a difference in the "File" metadata attribute in the HDF5 file that's causing the issue. I added some verbose output in another branch to investigate further.

jlippuner:
Alright, I see the issue. The tests are run with MPI on the CI machine, which adds the following default values to the input file:
```
<loadbalancing>
balancer  = default # Default value added at run time
tolerance = 0.5     # Default value added at run time
interval  = 10      # Default value added at run time
```
The gold files were generated without MPI and so they lack these entries.

We could just make `phdf_diff` ignore the `File` attribute and so it would not compare that the two sims were run with the same parameter file. But having that comparison may be useful. So maybe the better solution would be to write the default load balancing parameters to the file even when running without MPI. What do folks think?

pgrete:
oh interesting, very subtle, that's what it passed locally (I also ran without MPI).
I'd say, let's also write those default parameters for serial runs because we can restart serial runs with MPI and vice versa. Thus, the output file should also be consistent from my point of view.

Yurlungur:
I am fine with writing them in the file. That said, I lean towards ignoring this edge case in the tests. It bothers me that the tests are so brittle to these details.

jlippuner:
Alright, I think I agree with both @pgrete and @Yurlungur. I'm changing it to always write the load balancing info, and I'm also turning off comparison of the File metadata by default. However, I've added an option to turn it on for debugging purposes (i.e. if one is trying to figure out why two files differ, one can turn on checking the input file contained in the HDF5 metadata and maybe that will help). See #575

PR #535, opened 2021-06-01
Sparse interface update

jlippuner:
## PR Summary

This PR updates various interfaces connected to sparse variables, metadata, variable packing, etc. It also adds the capability of creating a sparse field variable without allocating its data. This capability is currently turned off, i.e. all sparse fields get allocated on all blocks, just like the previous behavior, because this PR does not yet implement the capability to communicate a sparse field from a block where it's allocated to a neighboring block where it's not allocated. This will be added in a subsequent PR. So sparse variables in this PR are actually dense, just like before, but the bookkeeping interface has been changed to allow for them to not be allocated everywhere.

Therefore, downstream codes should work as before even with sparse variables, but they will need to be updated to use the new interfaces. Most notably, how sparse pools are added to the `StateDescriptor`.

Here are some of the biggest changes in this PR.
- Removes `MeshBlockDataIterator` and instead makes public the existing interface in `MeshBlockData` to get variables by name or flags
- Simplifies and streamlines the `PackVariables` and `PackVariablesAndFluxes` interfaces (to reduce the amount of boilerplate code). The optional `PackIndexMap` and string keys output arguments are now passed as pointers (which can be NULL if the caller doesn't need that output) instead of non-const references. The Google style guide stipulates that function output arguments should be pointers and it reduces the number of overloaded functions necessary.
- Removes public interface to add variables directly to a `MeshBlockData`. All variables (dense and sparse) now have to be added to a `StateDescriptor` and a `MeshBlockData` instance needs to be initialized with a resolved `StateDescriptor`. This way, all the `MeshBlockData` instances will have the metadata for the same list of variable fields (dense and sparse) and so we don't have to worry about assembling a global list of variables. Currently, each `MeshBlockData` also allocates the data for all variables known to the resolved `StateDescriptor`, but in a subsequent PR, this will only be true for dense variables and sparse variable fields will be allocated and deallocated as needed by the runtime.
- Moves `sparse_id` from `Metadata` to `CellVariable`
[...]

PR #572, opened 2021-07-23
fix meshblock pack issues coming from template shadowing and add a test

### LINKED ISSUE #570: Sparse interface update broke meshblockpacks
The following code snippet now fails
```C++
/ /MeshBlockData *rc provided above somewhere
PackIndexMap imap;
rc->PackVariables(names, imap);
```
with errors that look like
```bash
/parthenon/src/interface/mesh_data.hpp:296:5: error: no matching function for call to ‘parthenon::MeshBlockData<double>::PackVariables(const std::vector<std::__cxx11::basic_string<char> >&, parthenon::PackIndexMap&, parthenon::PackIndexMap&, std::vector<std::__cxx11::basic_string<char> >&)’
```
This function signature is obviously wrong. Somehow the `PackIndexMap` is getting passed in twice. I believe this is because the API

We didn't catch this because for `MeshBlockPack` we apparently don't test this particular call signature anywhere. This needs to change, as it is used everywhere in the downstream codes.

Yurlungur:
## PR Summary

Downstream, @brryan noticed that calls to `MeshData::PackVariables` with a `PackIndexMap`, for example
```C++
// MeshData *rc defined above
PackIndexMap imap;
rc->PackVariables(names, imap);
```
cause compile-time errors. The issue was introduced in the sparse interface update, #535 , where this machinery changed significantly. In short, in that update, explicit overloads were replaced with variatic templates. Unfortunately, C++ does not resolve these templates appropriately, and they shadow each other, causing bugs. We didn't catch this until now because templated functions are optimized out if they're not called.

This MR resolves #570 . I replace the variatics with explicit overloads again, with a few changes to reduce line number as best I can. I also add a call to this machinery in the calculate pi example, so this issue should be caught by regression tests in the future.

Note that the full regression test suite is currently failing. (See #571 ). So we should either merge this in without triggering the full suite on the CI, or we should wait for #571 to be resolved. I lean towards the former. Thoughts? 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/src/interface/mesh_data.hpp
+++ b/src/interface/mesh_data.hpp
@@ -289,50 +289,211 @@ class MeshData {
 
   template <typename... Args>
   const auto &PackVariablesImpl(PackIndexMap *map_out, std::vector<std::string> *key_out,
-                                Args &&... args) {
+                                bool coarse, Args &&... args) {
     auto pack_function = [&](std::shared_ptr<MeshBlockData<T>> meshblock_data,
                              PackIndexMap &map, std::vector<std::string> &key) {
-      return meshblock_data->PackVariables(std::forward<Args>(args)..., map, key);
+      return meshblock_data->PackVariables(std::forward<Args>(args)..., map, key, coarse);
     };
     return pack_on_mesh_impl::PackOnMesh<VariablePack<T>>(
         varPackMap_, block_data_, pack_function, map_out, key_out);
   }
 
  public:
-  template <typename... Args>
-  const auto &PackVariablesAndFluxes(Args &&... args, PackIndexMap &map,
+  // DO NOT use variatic templates here. They shadow each other
+
+  // Pack by separate variable and flux names
+  const auto &PackVariablesAndFluxes(const std::vector<std::string> &var_names,
+                                     const std::vector<std::string> &flx_names,
+                                     const std::vector<int> &sparse_ids,
+                                     PackIndexMap &map, vpack_types::StringPair &key) {
+    return PackVariablesAndFluxesImpl(&map, &key, var_names, flx_names, sparse_ids);
+  }
+  const auto &PackVariablesAndFluxes(const std::vector<std::string> &var_names,
+                                     const std::vector<std::string> &flx_names,
+                                     const std::vector<int> &sparse_ids,
+                                     PackIndexMap &map) {
+    return PackVariablesAndFluxesImpl(&map, nullptr, var_names, flx_names, sparse_ids);
+  }
+  const auto &PackVariablesAndFluxes(const std::vector<std::string> &var_names,
+                                     const std::vector<std::string> &flx_names,
+                                     const std::vector<int> &sparse_ids,
                                      vpack_types::StringPair &key) {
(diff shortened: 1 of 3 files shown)


src/interface/mesh_data.hpp
```diff
@@ -289,50 +289,211 @@ class MeshData {
...
+                                     vpack_types::StringPair &key) {
+    return PackVariablesAndFluxesImpl(nullptr, &key, sparse_ids);
+  }
+  const auto &PackVariablesAndFluxes(const std::vector<int> &sparse_ids,
+                                     PackIndexMap &map) {
+    return PackVariablesAndFluxesImpl(&map, nullptr, sparse_ids);
+  }
+  const auto &PackVariablesAndFluxes(const std::vector<int> &sparse_ids) {
+    return PackVariablesAndFluxesImpl(nullptr, nullptr, sparse_ids);
+  }
```
pgrete, line 407: Are these actually required?
If I'm not missing something they're matching the
```c++
  const auto &PackVariablesAndFluxes(const std::vector<Elem> &names_or_flags,
                                     PackIndexMap &map, vpack_types::StringPair &key)
```
above, aren't they?

pgrete (approved):
Pending the minor question, looks good! Thanks.
I'm also fine with merging this in while #571 is not fixed yet, but we should still run the full test suite here manually and see if nothing else new fails.

src/interface/mesh_data.hpp
Yurlungur: These exclude either the `PackIndexMap` or the key, or both. Whereas the one you quoted has both of them. They're just syntactic sugar overloads.

Yurlungur:
Thanks @pgrete . Agreed, I'll run the full test suite

src/interface/mesh_data.hpp
```diff
@@ -289,50 +289,211 @@ class MeshData {
...
+  const auto &PackVariablesAndFluxes(vpack_types::StringPair &key) {
+    return PackVariablesAndFluxesImpl(nullptr, &key);
+  }
+  const auto &PackVariablesAndFluxes(PackIndexMap &map) {
+    return PackVariablesAndFluxesImpl(&map, nullptr);
+  }
+  const auto &PackVariablesAndFluxes() {
+    return PackVariablesAndFluxesImpl(nullptr, nullptr);
+  }
```
brryan, line 420: What does this return with no arguments? (For some reason I can't find the two-argument version of `PackVariablesAndFluxesImpl` to look at the code itself)

brryan (approved):
Just one question for my own edification. Looks good, thanks for adding a typical use case to the `calculate_pi` example.

src/interface/mesh_data.hpp
Yurlungur: It should pack all variables and all fluxes.

Yurlungur:
> Pending the minor question, looks good! Thanks.
> I'm also fine with merging this in while #571 is not fixed yet, but we should still run the full test suite here manually and see if nothing else new fails.

Looks like the only error is the `Input` metadata diff

PR #577, opened 2021-07-27
Update invalid indices to allow for no-op loops

brryan:
## PR Summary

When variable pack index pairs are requested for variables that don't exist, a set of invalid indices are returned. This provides a convenient way to turn loops over those variables' indices into no-ops without additional checking. We use this extensively downstream. To have loops like
```c++
for (int i = pack.first; i <= pack.second; i++) {
  /* Work we don't want to do for invalid packs */
}
```
work, though, we need `pack.second` to be more negative than `pack.first`. The invalid indices were changed from `0, -1` to `-1, -1` by #535, which made sense since `0` isn't necessarily invalid, but this PR changes them to `-1, -2` so the above pattern still works. A bit dangerous (as the code comments now say) but it's a very effective pattern downstream in performance-critical code. 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Current develop
 
 ### Added (new features/APIs/variables/...)
+- [[PR 577]](https://github.com/lanl/parthenon/pull/577) Update invalid indices to allow for no-op loops
 - [[PR 564]](https://github.com/lanl/parthenon/pull/564) Add EstimateTimestep to particles example task list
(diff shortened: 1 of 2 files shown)


Yurlungur (approved):
I like this change.

jlippuner (approved):
Short and sweet :)

pgrete (approved):
LGTM

PR #578, opened 2021-07-28
Add some profiling regions to tasks in particles example

maximumcats:
## PR Summary

Profiling regions are added to the particles example for easier visualizations in tools like Nsight Systems.

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Current develop
 
 ### Added (new features/APIs/variables/...)
+- [[PR 578]](https://github.com/lanl/parthenon/pull/578) Add some profiling regions to tasks in particles example
 - [[PR 577]](https://github.com/lanl/parthenon/pull/577) Update invalid indices to allow for no-op loops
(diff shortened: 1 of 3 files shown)


Yurlungur:
Great to have this! Ping some of us when this is ready for review.

maximumcats:
This is good for now; I am sure I will have more coming later but I want to spend a little time understanding the software design a little better before marking up the rest. Also, a way to do this which is a little easier is to create a simple class whose constructor calls pushRegion() and whose destructor calls popRegion(), so you just create a copy of the object at the top of the routine and don't need to remember to destroy it, when it goes out of scope the destructor will trigger the region end, so we can cut down on number of lines of code by half (and write less error-prone code).

Yurlungur:
> Also, a way to do this which is a little easier is to create a simple class whose constructor calls pushRegion() and whose destructor calls popRegion()

Oh that's cool. We should definitely add that.

Yurlungur (approved):

Yurlungur:
Tagging @brryan and @jdolence for review

jlippuner (approved):
Looks great! I like the idea of a timer class that pushes and pops on construction/destruction.

PR #580, opened 2021-08-02
Add basic test for meshblock data and make key private to *Data

### LINKED ISSUE #579: Possible bug in PackVariables call on MeshData objects
I haven't gotten to the bottom of this, but I think there might be a bug in the PackVariables machinery.  It seems like while this works:
```c++
myvar = std::vector<std:string>({"myvar"});
PackIndexMap imap;
auto v = md->PackVariables(myvar, imap);
// use v
```
this does not:
```c++
auto v = md->PackVariables(myvar);
```
i.e. the call where we don't ask for a PackIndexMap back seems to be broken.  Code compiles, but the answers are wrong.  For this single variable case, the index in the map should be (and is when explicitly pulled out) just zero.  So I had some code that called PackVariables without the PackIndexMap argument and then just hardcoded the zero index for the variable when I used the pack.  That code did not work.  Adding the PackIndexMap fixes the issue.  Don't have time to dig into this right now, but wanted to note this issue somewhere.

@jlippuner @Yurlungur @pgrete might all be interested.

Yurlungur:
## PR Summary

Recently there have been several downstream issues with `MeshData::PackVariables`. In particular, see #570, where template shadowing caused some versions of the function to not successfully compile, and #579, where currently calls without `PackIndexMap` are behaving incorrectly.

This MR resolves the above-mentioned issues and hopefully future issues in two ways:
- It implements some units tests of `MeshData`, which I believe are sorely needed. For now, I've only implemented two basic tests, but the framework is now there, and it's better than what we had before.
- I disable the overloads that expose the key used for pack caching. In `MeshData` I remove the overloads completely. In `MeshBlockData`, I make them private, only accessible by `MeshData`. This resolves #579 .

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/src/interface/mesh_data.hpp
+++ b/src/interface/mesh_data.hpp
@@ -240,7 +240,7 @@ class MeshData {
   }
 
   template <class... Args>
-  void Add(Args &&... args) {
+  void Add(Args &&...args) {
     for (const auto &pbd : block_data_) {
       pbd->Add(std::forward<Args>(args)...);
     }
@@ -256,7 +256,7 @@ class MeshData {
   }
 
   template <typename... Args>
-  void Copy(const std::shared_ptr<MeshData<T>> src, Args &&... args) {
+  void Copy(const std::shared_ptr<MeshData<T>> src, Args &&...args) {
     if (src.get() == nullptr) {
       PARTHENON_THROW("src points at null");
     }
@@ -276,7 +276,7 @@ class MeshData {
   template <typename... Args>
   const auto &PackVariablesAndFluxesImpl(PackIndexMap *map_out,
                                          vpack_types::StringPair *key_out,
-                                         Args &&... args) {
+                                         Args &&...args) {
     auto pack_function = [&](std::shared_ptr<MeshBlockData<T>> meshblock_data,
                              PackIndexMap &map, vpack_types::StringPair &key) {
       return meshblock_data->PackVariablesAndFluxes(std::forward<Args>(args)..., map,
@@ -289,7 +289,7 @@ class MeshData {
 
   template <typename... Args>
   const auto &PackVariablesImpl(PackIndexMap *map_out, std::vector<std::string> *key_out,
-                                bool coarse, Args &&... args) {
+                                bool coarse, Args &&...args) {
     auto pack_function = [&](std::shared_ptr<MeshBlockData<T>> meshblock_data,
                              PackIndexMap &map, std::vector<std::string> &key) {
       return meshblock_data->PackVariables(std::forward<Args>(args)..., map, key, coarse);
@@ -305,20 +305,8 @@ class MeshData {
   const auto &PackVariablesAndFluxes(const std::vector<std::string> &var_names,
                                      const std::vector<std::string> &flx_names,
                                      const std::vector<int> &sparse_ids,
-                                     PackIndexMap &map, vpack_types::StringPair &key) {
-    return PackVariablesAndFluxesImpl(&map, &key, var_names, flx_names, sparse_ids);
-  }
-  const auto &PackVariablesAndFluxes(const std::vector<std::string> &var_names,
-                                     const std::vector<std::string> &flx_names,
-                                     const std::vector<int> &sparse_ids,
                                      PackIndexMap &map) {
     return PackVariablesAndFluxesImpl(&map, nullptr, var_names, flx_names, sparse_ids);
-  }
-  const auto &PackVariablesAndFluxes(const std::vector<std::string> &var_names,
-                                     const std::vector<std::string> &flx_names,
-                                     const std::vector<int> &sparse_ids,
-                                     vpack_types::StringPair &key) {
-    return PackVariablesAndFluxesImpl(nullptr, &key, var_names, flx_names, sparse_ids);
   }
   const auto &PackVariablesAndFluxes(const std::vector<std::string> &var_names,
                                      const std::vector<std::string> &flx_names,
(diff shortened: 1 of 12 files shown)


Yurlungur:
I can't trigger on Darwin at the moment, but this should have zero performance implications. This is ready for review.

src/interface/mesh_data.hpp
```diff
@@ -406,92 +354,49 @@ class MeshData {
...
-  }
   const auto &PackVariablesAndFluxes(PackIndexMap &map) {
     return PackVariablesAndFluxesImpl(&map, nullptr);
   }
   const auto &PackVariablesAndFluxes() {
     return PackVariablesAndFluxesImpl(nullptr, nullptr);
   }
 
-  // As above, no variatics because they shadow each other.
+  // As above, NO TEMPLATE ARGUMENTS, as they shadow each other
```
jlippuner: I'm a bit confused by this comment, we still have the `Elem` template arguments...

src/interface/mesh_data.hpp
```diff
@@ -406,92 +354,49 @@ class MeshData {
...
-  const auto &PackVariables(const std::vector<Elem> names_or_flags,
-                            const std::vector<int> &sparse_ids, PackIndexMap &map,
-                            std::vector<std::string> &key, bool coarse = false) {
-    return PackVariablesImpl(&map, &key, coarse, names_or_flags, sparse_ids);
-  }
-  template <typename Elem>
   const auto &PackVariables(const std::vector<Elem> names_or_flags,
                             const std::vector<int> &sparse_ids, PackIndexMap &map,
                             bool coarse = false) {
     return PackVariablesImpl(&map, nullptr, coarse, names_or_flags, sparse_ids);
```
jlippuner: Why are we keeping the `key_out` function argument in `PackVariablesImpl`? It looks like we're always passing `nullptr` now, so we might as well remove it, right? Same goes for `PackVariablesAndFluxes`.

tst/unit/test_mesh_data.cpp
```diff
@@ -0,0 +1,148 @@
+//========================================================================================
+// Athena++ astrophysical MHD code
+// Copyright(C) 2014 James M. Stone <jmstone@princeton.edu> and other code contributors
+// Licensed under the 3-clause BSD License, see LICENSE file for details
+//========================================================================================
+// (C) (or copyright) 2020-2021. Triad National Security, LLC. All rights reserved.
```
jlippuner:
```suggestion
// (C) (or copyright) 2021. Triad National Security, LLC. All rights reserved.
```

tst/unit/test_mesh_data.cpp
```diff
@@ -0,0 +1,148 @@
...
+                  var4(c, k, j, i) = n;
+                }
+              });
+        }
+      }
+
+      THEN("A pack for a scalar with a PackIndexMap works") {
+        PackIndexMap imap;
+        std::vector<std::string> vlist{"v5"};
+        auto pack = mesh_data.PackVariables(vlist, imap);
```
jlippuner:
```suggestion
        const std::vector<std::string> vlist{"v5"};
        const auto &pack = mesh_data.PackVariables(vlist, imap);
```

tst/unit/test_mesh_data.cpp
```diff
@@ -0,0 +1,148 @@
...
+              Real n = i + 1e1 * j + 1e2 * k + 1e3 * c + 1e4 * v + 1e5 * b;
+              if (n != pack(b, vlo, k, j, i)) ltot += 1;
+            },
+            nwrong);
+        REQUIRE(nwrong == 0);
+      }
+
+      THEN("A hardcoded pack for a scalar with no PackIndexMap works") {
+        std::vector<std::string> vlist{"v5"};
+        auto pack = mesh_data.PackVariables(vlist);
```
jlippuner:
```suggestion
        const std::vector<std::string> vlist{"v5"};
        const auto &pack = mesh_data.PackVariables(vlist);
```

jlippuner (approved):
Looks great! Thanks for tracking down this issue, fixing it, and adding the much needed unit test! I just have some very minor comments.

jlippuner:
> I can't trigger on Darwin at the moment, but this should have zero performance implications. This is ready for review.

I just triggered Darwin CI. Were you not able to do it because of the LANL network issues yesterday, or is there another issue with the Darwin CI?

src/interface/mesh_data.hpp
Yurlungur: Oops. This is from my time when I was trying to debug. This comment is wrong and I'll change it back.

tst/unit/test_mesh_data.cpp
Yurlungur: The string can be a const, but the pack cannot be a reference. Since the pack gets copied to device, this will cause memory errors.

tst/unit/test_mesh_data.cpp
Yurlungur: See above.

src/interface/mesh_data.hpp
Yurlungur: Ah good point. I'll remove it.

Yurlungur:
> I just triggered Darwin CI. Were you not able to do it because of the LANL network issues yesterday, or is there another issue with the Darwin CI?

It was just the network issues.

Yurlungur:
Thanks for the useful comments, @jlippuner ! I think it should all be addressed now.

Yurlungur:
@par-hermes format

tst/unit/test_mesh_data.cpp
jlippuner: Really? I thought a copy will be made to transfer to the device because the lambda is capture by value...

Yurlungur: Capturing a reference by value copies the reference, right? Maybe I'm misunderstanding something.

jlippuner: I'm actually not sure. I thought it would feed the const ref into the copy constructor to make a copy, but you may be right. I guess we have to test it.

jlippuner: I think it makes a copy. This code
```c++
#include <cstdio>

int main(int, char**) {
  const int var = 3;
  const int &ref = var;

  printf("&var = %p, &ref = %p\n", &var, &ref);

  auto lam_ref = [&]() {
    printf("In lambda ref:\n");
    printf("&var = %p, &ref = %p\n", &var, &ref);
  };

  auto lam_val = [=]() {
    printf("In lambda val:\n");
    printf("&var = %p, &ref = %p\n", &var, &ref);
  };

  lam_ref();
  lam_val();

  return 0;
}
```
produces this output for me:
```
&var = 0x7ffffc182004, &ref = 0x7ffffc182004
In lambda ref:
&var = 0x7ffffc182004, &ref = 0x7ffffc182004
In lambda val:
&var = 0x7ffffc181fe8, &ref = 0x7ffffc181fec
```
So in the second lambda that captures by reference, both pointers are different, because it made two fresh copies.

jdolence (approved):
Looks good.  Formatting check seems not to have run though???

Yurlungur:
Oh weird. I'll force-retrigger it.

tst/unit/test_mesh_data.cpp
Yurlungur: ah ok interesting. Thanks for checking that. I swear I've been burned by this when working with kokkos before.

PR #582, opened 2021-08-12
Building blocks for solvers

jdolence:
## PR Summary

Adding some utilities that should help construct some generic solvers for parthenon based codes.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


src/solvers/solver_utils.hpp
```diff
@@ -0,0 +1,79 @@
...
+                   const Stencil<T> &s, const IndexRange &ib, const IndexRange &jb,
+                   const IndexRange &kb) {
+  parthenon::par_for(
+      DEFAULT_LOOP_PATTERN, "StencilMatVec", DevExecSpace(), 0, vout.GetDim(5) - 1, kb.s,
+      kb.e, jb.s, jb.e, ib.s, ib.e,
+      KOKKOS_LAMBDA(const int b, const int k, const int j, const int i) {
+        vout(b, ivout, k, j, i) = rhs(b, irhs, k, j, i);
+        for (int n = 0; n < s.nstencil; n++) {
+          vout(b, ivout, k, j, i) +=
+              s.w(n) * vin(b, ivin, k + s.koff(n), j + s.joff(n), i + s.ioff(n));
```
ryosukep: How do we deal with Matvec with different weights in each row? e.g., a_{i,i+1) != a_{j,j+1}

src/utils/all_reduce.hpp
```diff
@@ -0,0 +1,106 @@
...
+  typename std::enable_if<std::is_same<U, Real>::value, TaskStatus>::type
+  StartReduce(int &only, MPI_Op op) {
+#ifdef MPI_PARALLEL
+    if (only == N) {
+      MPI_Iallreduce(MPI_IN_PLACE, &val, 1, MPI_PARTHENON_REAL, op, MPI_COMM_WORLD, &req);
+    }
+#endif
+    active = true;
+    return TaskStatus::complete;
+  }
```
Yurlungur: This seems like an odd pattern... Can you not just template on `T` here and then provide the separate implementations in a `*.cpp` file?

src/utils/all_reduce.hpp
```diff
@@ -0,0 +1,106 @@
...
+
+#include <config.hpp>
+#include <parthenon_mpi.hpp>
+
+namespace parthenon {
+
+enum class Op { MAX, MIN, SUM, PROD, LAND, BAND, LOR, BOR, LXOR, BXOR, MAXLOC, MINLOC };
+#ifndef MPI_PARALLEL
+#define MPI_
+#endif
```
Yurlungur: This is both clever and evil.

src/utils/all_reduce.hpp
```diff
@@ -0,0 +1,106 @@
...
+struct AllReduce {
+  T val;
+#ifdef MPI_PARALLEL
+  MPI_Request req;
+#endif
+  bool active = false;
+
+  template <typename U = T>
+  typename std::enable_if<std::is_same<U, Real>::value, TaskStatus>::type
+  StartReduce(int &only, MPI_Op op) {
```
Yurlungur: I take it this is what you were referring to as far as the compiler insisting on pass by reference?

example/poisson/poisson_driver.cpp
```diff
@@ -59,13 +59,47 @@ TaskCollection PoissonDriver::MakeTaskCollection(BlockList_t &blocks) {
   const int num_partitions = pmesh->DefaultNumPartitions();
   TaskRegion &solver_region = tc.AddRegion(num_partitions);
 
+  // make some storage for doing an Allreduce
+  reductions.resize(1);
```
Yurlungur: Why is this a vector if it's always size 1? Why not make it a variable called `error` or something?

example/poisson/poisson_driver.cpp
```diff
@@ -59,13 +59,47 @@ TaskCollection PoissonDriver::MakeTaskCollection(BlockList_t &blocks) {
...
     auto &mdelta = pmesh->mesh_data.GetOrAdd("delta", i);
 
     TaskList &tl = solver_region[i];
 
+    // pass a pointer to the variable being reduced into
+    auto loc_red = tl.AddTask(none, poisson_package::SumMass<MeshData<Real>>, md.get(),
+                              &reductions[0].val);
+    // make it a regional dependency so dependent tasks can't execute until all lists do
+    // this
+    solver_region.AddRegionalDependencies(0, i, loc_red);
```
Yurlungur: why pass the region index instead of the `tl` reference?

Yurlungur: also why does the local reduce need the region dependency? Shouldn't only the test need it?

example/poisson/poisson_driver.cpp
```diff
@@ -59,13 +59,47 @@ TaskCollection PoissonDriver::MakeTaskCollection(BlockList_t &blocks) {
...
+    // start a non-blocking MPI_Iallreduce
+    TaskID start_global_reduce = tl.AddTask(
+        loc_red, &parthenon::AllReduce<Real>::StartReduce, &reductions[0], i, MPI_SUM);
+    // test the reduction until it completes
+    TaskID finish_global_reduce = tl.AddTask(
+        start_global_reduce, &parthenon::AllReduce<Real>::CheckReduce, &reductions[0], i);
+    solver_region.AddRegionalDependencies(1, i, finish_global_reduce);
```
Yurlungur: I guess not in this case, but in general the receive doesn't need to be in the same place as the send right? There could be intermediate work

src/solvers/solver_utils.hpp
```diff
@@ -0,0 +1,80 @@
...
+    Kokkos::deep_copy(koff, koff_h);
+  }
+};
+
+template <typename PackType, typename T>
+void StencilMatVec(const PackType &vin, const int ivin, const PackType &vout,
+                   const int ivout, const PackType &rhs, const int irhs,
+                   const Stencil<T> &s, const IndexRange &ib, const IndexRange &jb,
+                   const IndexRange &kb) {
+  parthenon::par_for(
```
Yurlungur: We may wish to have a version of this that's a `par_for_inner` so we can call the stencil from device inside another kernel.

example/poisson/poisson_package.cpp
```diff
@@ -49,59 +50,123 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
...
   for (int i = X2DIR; i <= ndim; i++) {
     const Real dy = coords.Dx(i);
     PARTHENON_REQUIRE_THROWS(dx == dy,
-                             "UpdatePhi requires that DX be equal in all directions.");
+                             "ComputeRHS requires that DX be equal in all directions.");
+  }
+
+  Real total;
+  parthenon::par_reduce(
+      parthenon::loop_pattern_mdrange_tag, "ComputeRHS", DevExecSpace(), 0,
```
Yurlungur: Loop name should be `SumMass` or something to avoid ambiguity with the `ComputeRHS` task

example/poisson/poisson_package.cpp
```diff
@@ -49,59 +50,123 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
...
+  const int iphi = imap["potential"].first;
+  const std::vector<std::string> phi_var({"potential"});
+  PackIndexMap imap2;
+  const auto &dv = du->PackVariables(phi_var, imap2);
+  const int idphi = imap2["potential"].first;
+
+  StateDescriptor *pkg = pm->packages.Get("poisson_package").get();
+  const auto stencil = pkg->Param<Stencil_t>("stencil");
+
+  StencilMatVec(v, iphi, dv, idphi, v, irhs, stencil, ib, jb, kb);
```
Yurlungur: Written in this way, you introduce an extra kernel launch. I wonder if we could add a `Stencil` call that computes and fills `dv` and does the update all at once?

src/solvers/solver_utils.hpp
```diff
@@ -0,0 +1,80 @@
...
+    Kokkos::deep_copy(joff, joff_h);
+    Kokkos::deep_copy(koff, koff_h);
+  }
+};
+
+template <typename PackType, typename T>
+void StencilMatVec(const PackType &vin, const int ivin, const PackType &vout,
+                   const int ivout, const PackType &rhs, const int irhs,
+                   const Stencil<T> &s, const IndexRange &ib, const IndexRange &jb,
+                   const IndexRange &kb) {
```
Yurlungur: Why not make `StencilMatVec` into `Stencil::MatVec`? I.e., make it a member function of the stencil object?

Yurlungur (commented):
Nice work.

Some random thoughts below. I like the new `Stencil` thing and I mostly like the new reduction machinery. I don't know how I feel about the task regional dependencies machinery. I *think* I like it. It does seem like our tasking tooling is getting pretty complicated though. We should write down all that it's capable of doing now and how to do it in some documentation.

jdolence:
@pgrete I know you're busy, but this branch has a new task-based global reduction capability that I think might satisfy the need you had.  curious what you think. @forrestglines you're probably aware of the same need so also interested in your opinion.

src/solvers/solver_utils.hpp
jdolence: good idea!  done.

example/poisson/poisson_package.cpp
jdolence: it has to be this way (and was before also) because we're doing Jacobi.  in general though, it's not hard to imagine that the list of available `MatVec` like routines should grow some to accommodate other patterns.  perhaps we should model this after BLAS or something...

example/poisson/poisson_package.cpp
jdolence: oops copy/paste error.  got it.

src/solvers/solver_utils.hpp
jdolence: maybe.  wouldn't that only be appropriate if the stencil was quite large (like larger than the ~standard 7-point stencil in 3d)?

example/poisson/poisson_driver.cpp
jdolence: It does not.  In fact, even though they are written next to each other, the other non-dependent tasks in the list can (and do!) execute until the global reduction finishes.  in other words, even here there _is_ intermediate work.

example/poisson/poisson_driver.cpp
jdolence: There's no need for it to be a vector in this example.  It was one of those things I did to try to indicate that you could have multiple independent reductions going on simultaneously in a task region.

example/poisson/poisson_driver.cpp
jdolence: re index for `tl` reference: it was just a choice.  this actually went in with my last PR.  I actually hate both the first two arguments, but prioritized getting the capability in over cleaning that interface piece up.

re why local reduce...: I think you're misinterpreting what the `AddRegionalDependencies` call is doing.  When you mark a `TaskID` like that, you're saying that dependent tasks (who in this example use `loc_rad` as a dependency) need to wait until all lists that registered that task need to finish before they can execute.  Example:

```c++
auto first = tl.AddTask(none, do_first);
region.AddRegionalDependencies(0, i, first);
auto second = tl.AddTask(first, do_second);
```

This code says that `do_second` won't be called until all the lists that mark `first` as a regional dependency have completed there calls to `do_first`.  The `do_first` call is just a normal task -- it doesn't know or care about the regional dependency thing.  I know it's a little confusing, but this is the basic info that needs to be provided to get this all working.  We can probably change some names to clarify things a bit.

So, back to the code in the Poisson driver.  What the code says is don't do the MPI_Allreduce until I'm done accumulating the result locally.  Does that make sense?

src/utils/all_reduce.hpp
jdolence: it is. ☹️

src/utils/all_reduce.hpp
jdolence: Mwahahahaha.  Haven't even tested that it works, to be honest.

forrestglines:
> @pgrete I know you're busy, but this branch has a new task-based global reduction capability that I think might satisfy the need you had. curious what you think. @forrestglines you're probably aware of the same need so also interested in your opinion.

Sounds interesting. I also need the global reductions. Currently I'm using the same hack as @pgrete. However, my use case is slightly more complicated since I need to do two simultaneous summations.

I'll take a look at it

src/solvers/solver_utils.hpp
```diff
@@ -0,0 +1,126 @@
...
+#include <string>
+#include <vector>
+
+#include "kokkos_abstraction.hpp"
+
+namespace parthenon {
+
+namespace solvers {
+
+struct SparseMatrixAccessor {
```
forrestglines, line 25: A little documentation on what this struct is useful for would be good (I assume it's not yet documented because it's WIP)

src/solvers/solver_utils.hpp
```diff
@@ -0,0 +1,126 @@
...
+            vout(b, ivout, k, j, i) +=
+                spmat(b, n, k, j, i) *
+                vin(b, ivin, k + koff(m), j + joff(m), i + ioff(m));
+          }
+        });
+  }
+};
+
+template <typename T>
+struct Stencil {
```
forrestglines, line 82: Same, documentation would be appreciated here

src/utils/all_reduce.hpp
forrestglines: Wow. Make sure it works on a few different compilers when you do check

src/utils/all_reduce.hpp
```diff
@@ -0,0 +1,106 @@
...
+namespace parthenon {
+
+enum class Op { MAX, MIN, SUM, PROD, LAND, BAND, LOR, BOR, LXOR, BXOR, MAXLOC, MINLOC };
+#ifndef MPI_PARALLEL
+#define MPI_
+#endif
+
+template <typename T, int N = 0>
+struct AllReduce {
+  T val;
```
forrestglines: I see that currently this only reduces on one value. Could this be extended to reduce on multiple values?

example/poisson/poisson_driver.cpp
forrestglines: This is helpful  for an example. We'll also be doing several independent reductions with potentially different operations.

Yurlungur: In that case can we reduce something else, like 1-phi^2 or some nonsense, just to show how that would work?

example/poisson/poisson_driver.cpp
Yurlungur: > re index for `tl` reference: it was just a choice.  this actually went in with my last PR.  I actually hate both the first two arguments, but prioritized getting the capability in over cleaning that interface piece up.

Yeah fair enough. I agree it can stay for now. Its not a huge deal. But we'll want to clean it up Bernie downstream Codes start using it or it'll be too late.

> re why local reduce...:

Ah I see where my confusion lay! Somehow I had it stuck in my head that there was one task region per rank, but of course that's wrong. There's one per mesh block.

example/poisson/poisson_driver.cpp
Yurlungur: Excellent. :+1:. It might be worth emphasizing that in a comment.

src/utils/all_reduce.hpp
jdolence: You don't like my SFINAE???

Yes, you can just use template specialization.  My secret ambition is to have all of parthenon be header only.

The other solution I was thinking about was some kind of map from type to MPI_Datatype.  like `my_map[Real] = MPI_PARTHENON_REAL`.  Obviously that doesn't work as I just wrote it, and in any case still requires that you write out all the types explicitly.  I'll think a bit more about what the right thing to do here is...

example/poisson/poisson_package.cpp
```diff
@@ -49,59 +50,173 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
...
+  }
+
+  Real total;
+  parthenon::par_reduce(
+      parthenon::loop_pattern_mdrange_tag, "SumMass", DevExecSpace(), 0, v.GetDim(5) - 1,
+      kb.s, kb.e, jb.s, jb.e, ib.s, ib.e,
+      KOKKOS_LAMBDA(const int b, const int k, const int j, const int i, Real &sum) {
+        sum += v(b, irho, k, j, i) * std::pow(dx, ndim);
+      },
+      Kokkos::Sum<Real>(total));
```
forrestglines, line 157: So for my reduction needs, my equivalent local kernel reduces two values at once and is somewhat expensive to execute. I don't want to execute the kernel twice, so I don't want two tasks for the local reduction. With the current interface, I think I could still accomplish the local reduction all at once into two different `AllReduce` objects. It would then be only a little awkward to add separate global reduction tasks for the two `AllReduce` objects.

src/solvers/solver_utils.hpp
Yurlungur: Hmm. I'd have to think about that. My thinking was more along the lines off coalescing kernels. Or for example if the stencil is block structured rather than banded, there might be gains.

I guess I'm overthinking it for now. That kinda machinery can be added when a need arises.

src/solvers/solver_utils.hpp
jdolence: yeah, I'll document.  just haven't gotten there yet.  in brief, this class let's us define sparse matrices via all our variable/packing machinery.

example/poisson/poisson_package.cpp
Yurlungur: Hmm. I think naming things similar to blas for continuity is probably useful.

But you're right for now we should focus on the need we have.

src/utils/all_reduce.hpp
jdolence: I haven't tested it yet, but this should be able to support doing reductions on, for example, a `std::vector`.    See the last version of `StartReduce`.

src/utils/all_reduce.hpp
Yurlungur: I like the template solution overall. I think I'd prefer standard template specialization over SFINAE. But obviously it's not a big deal.

This is really a style thing more than anything else.

If you were really evil and we had access to it in the standard, you could return a std::variant. (Don't do that.)

src/solvers/solver_utils.hpp
jdolence: this is addressed now via the `SparseMatrixAccessor` class.  See the Poisson example for usage.

src/utils/all_reduce.hpp
jdolence: just pushed an update.  the vector thing is still untested, but perhaps the intent is clearer now.  would doing a reduction on a std::vector work for you?

src/utils/all_reduce.hpp
jdolence: ok check out my solution now @Yurlungur.  I like this one a lot better.  I even got to use a `void *` and not feel guilty about it 😄

Yurlungur: Yeah I like the new solution. Very old-style MPI if I do say so myself.

src/utils/all_reduce.hpp
jdolence: ok the poisson example now (needlessly) does a reduction on a vector of ints as an example.

jdolence:
I expect this will fail the CI until I update the gold for the Poisson regression test.  I _really_ wish I could update the gold as part of this PR.  Is that really not possible?

pgrete:
> I expect this will fail the CI until I update the gold for the Poisson regression test. I _really_ wish I could update the gold as part of this PR. Is that really not possible?

Why isn't it possible? Everything is versioned anyway so feel free to upload/"release" a new gold standard and edit the `CMakeLists.txt` here accordingly.
All previous version will continue to use the old standard as they don't go away.
Also, in case your  test changes during review here, you can just reupload/update your new gold standard at a later time without updating the version string (just the checksum would change).

jdolence:
> > I expect this will fail the CI until I update the gold for the Poisson regression test. I _really_ wish I could update the gold as part of this PR. Is that really not possible?
> 
> Why isn't it possible? Everything is versioned anyway so feel free to upload/"release" a new gold standard and edit the `CMakeLists.txt` here accordingly.
> All previous version will continue to use the old standard as they don't go away.
> Also, in case your test changes during review here, you can just reupload/update your new gold standard at a later time without updating the version string (just the checksum would change).

Yeah ok.  It's not as bad as I made it sound.  I basically just meant I wish that I could just commit my new gold in this PR and everything would just work instead of having to do the release thing and update the CMakeLists.txt file.  I know this is the best solution we could come up with -- I'm just complaining, so please ignore me ;)

example/poisson/poisson_package.cpp
```diff
@@ -86,22 +195,33 @@ TaskStatus UpdatePhi(T *u, T *du) {
...
-        dv(b, idphi, k, j, i) -= v(b, iphi, k, j, i);
-      });
+  StateDescriptor *pkg = pm->packages.Get("poisson_package").get();
+  if (isp_hi < 0) { // there is no sparse matrix, so we must be using the stencil
+    const auto &stencil = pkg->Param<Stencil_t>("stencil");
+    parthenon::par_for(
+        DEFAULT_LOOP_PATTERN, "StencilJacobi", DevExecSpace(), 0, v.GetDim(5) - 1, kb.s,
+        kb.e, jb.s, jb.e, ib.s, ib.e,
+        KOKKOS_LAMBDA(const int b, const int k, const int j, const int i) {
+          const Real rhs = dV * v(b, irho, k, j, i);
```
brryan, line 232: Was the intention for this to be `v(b, irhs, k, j, i) = dV * v(b, irho, k, j, i);`?

src/solvers/solver_utils.hpp
```diff
@@ -0,0 +1,140 @@
...
+  int ndiag;
+
+  SparseMatrixAccessor(const std::string &label, const int n,
+                       std::vector<std::vector<int>> off)
+      : ioff(label + "_ioff", n), joff(label + "_joff", n), koff(label + "_koff", n),
+        nstencil(n) {
+    assert(off.size() == 3);
+    assert(off[0].size() >= n);
+    assert(off[1].size() >= n);
+    assert(off[2].size() >= n);
```
brryan: Can these `assert`s be `PARTHENON_REQUIRE_THROWS` instead? The Parthenon error checking macros give extra information about e.g. line and file numbers

src/solvers/solver_utils.hpp
```diff
@@ -0,0 +1,140 @@
...
+
+  Stencil(const std::string &label, const int n, std::vector<T> wgt,
+          std::vector<std::vector<int>> off)
+      : w(label + "_w", n), ioff(label + "_ioff", n), joff(label + "_joff", n),
+        koff(label + "_koff", n), nstencil(n) {
+    assert(off.size() == 3);
+    assert(wgt.size() >= n);
+    assert(off[0].size() >= n);
+    assert(off[1].size() >= n);
+    assert(off[2].size() >= n);
```
brryan: Similar complaint about `assert`

src/tasks/task_list.hpp
```diff
@@ -348,22 +351,31 @@ class TaskRegion {
...
+      if (id_complete[reg_id]) continue;
+      if (HasRun(reg_id) && !all_done[reg_id].active) {
         bool done = IsComplete(reg_id);
 #ifdef MPI_PARALLEL
-        int all_done = done;
-        int global_done;
-        MPI_Allreduce(&all_done, &global_done, 1, MPI_INT, MPI_MIN, MPI_COMM_WORLD);
+        all_done[reg_id].val = done;
+        all_done[reg_id].StartReduce(MPI_MIN);
+        int global_done = all_done[reg_id].val;
```
brryan: Is `global_done` no longer needed now that we're checking `all_done[reg_id].active` instead?

brryan (approved):
This looks good, just some small comments from me, I'll go ahead and approve now.

example/poisson/poisson_package.cpp
jdolence: No, actually.  But this reminds me that I never removed "rhs" as a variable, which is no longer required.  Thanks.

src/tasks/task_list.hpp
jdolence: thanks, cleaning this up.

src/utils/all_reduce.hpp
jdolence: I updated how I dealt with this.  It's less insane now, and works for serial builds..

src/solvers/solver_utils.hpp
jdolence: I updated docs

src/solvers/solver_utils.hpp
jdolence: updated

jdolence:
@forrestglines any chance you want to approve (or leave more comments) so this can move along.  I'd like to merge it.

example/poisson/poisson_driver.cpp
```diff
@@ -59,13 +60,84 @@ TaskCollection PoissonDriver::MakeTaskCollection(BlockList_t &blocks) {
...
 
     TaskList &tl = solver_region[i];
 
+    //--- Demo a few reductions
+    // pass a pointer to the variable being reduced into
+    auto loc_red = tl.AddTask(none, poisson_package::SumMass<MeshData<Real>>, md.get(),
+                              &total_mass.val);
+    // make it a regional dependency so dependent tasks can't execute until all lists do
+    // this
+    solver_region.AddRegionalDependencies(0, i, loc_red);
```
forrestglines: This example might be made a little clearer by changing the region id's from hard coded numbers to `regiond_id` or `reg_dep_id` and incrementing them as they're used.

example/poisson/poisson_driver.cpp
```diff
@@ -59,13 +60,84 @@ TaskCollection PoissonDriver::MakeTaskCollection(BlockList_t &blocks) {
...
+                                  [](int *max_rank) {
+                                    std::cout << "Max rank = " << *max_rank << std::endl;
+                                    return TaskStatus::complete;
+                                  },
+                                  &max_rank.val)
+                            : none);
+
+    //--- Begining of tasks related to solving the Poisson eq.
+    auto mat_elem =
+        tl.AddTask(none, poisson_package::SetMatrixElements<MeshData<Real>>, md.get());
```
forrestglines, line 144: Solving the Poisson equation using these sparse matrix elements isn't the only way nor the most straightforward implementation, it's done here to demonstrate these solver building blocks, correct? Can you add the the comment something like "... solving the Poisson eq. using matrix operations to demonstrate stencil solver"

example/poisson/poisson_driver.cpp
```diff
@@ -74,28 +146,93 @@ TaskCollection PoissonDriver::MakeTaskCollection(BlockList_t &blocks) {
...
+    solver_region.AddRegionalDependencies(5, i, check);
+
+    auto print = none;
+    if (i == 0) { // only print once
+      print = tl.AddTask(check, poisson_package::PrintComplete);
+    }
+    //--- End of tasks related to solving the Poisson eq
+
+    // do a vector reduction (everything below here), just for fun
+    // first fill it in
```
forrestglines, line 203: Looks good, I'll be using this in AthenaPK soon

example/poisson/poisson_package.cpp
```diff
@@ -53,14 +54,140 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
...
+  const Real w0 = -2.0 * ndim;
+  parthenon::par_for(
+      DEFAULT_LOOP_PATTERN, "SetMatElem", DevExecSpace(), 0, v.GetDim(5) - 1, kb.s, kb.e,
+      jb.s, jb.e, ib.s, ib.e,
+      KOKKOS_LAMBDA(const int b, const int k, const int j, const int i) {
+        for (int n = isp_lo; n <= isp_hi; n++) {
+          v(b, n, k, j, i) = 1;
+        }
+        v(b, isp_lo + 1, k, j, i) = w0;
+      });
```
forrestglines, line 120: I'm struggling to visualize the actual structure of the matrix v, what's it's full dimensions? `nblocks*27*nz*ny*nx`? So a few elements along the n dimension are 1 with the second being `w0`? And this matrix is repeated across all `k,j,i`?

src/solvers/solver_utils.hpp
forrestglines: I think it would be better that the classes and code themselves are documented with comments explaining how they function and operate, but looking at the rest of the code that's not the norm. So this is ok

forrestglines (approved):
Looks good! We'll be using the new reductions in AthenaPK soon

example/poisson/poisson_driver.cpp
jdolence: done

example/poisson/poisson_driver.cpp
jdolence: this is actually a little subtle in the general case, but not here, so I did this.

example/poisson/poisson_driver.cpp
jdolence: Great, let me know if you run into issues.

example/poisson/poisson_package.cpp
jdolence: The point of these, at some level, is to not make you have to think about the actual sparse matrix.  They represent, very deliberately, the sparse matrices you get out of discretizing equations.  That's why you specify which elements are non-zero via offsets to the i,j,k index of a cell.  With that, it's super easy to fill one in for representing things like difference equations.  You'd never want to try to represent a generic sparse matrix with these.  The values above are what comes out of a simple centered finite difference discretization of the Laplacian.  Each `n` corresponds to either the cell itself or one of it's neighbors, as determined by the i,j,k offset arrays that are set up in the initialization routine.

PR #583, opened 2021-08-13
Fix file numbering logic for writing outputs after restarting a simulation

### LINKED ISSUE #573: Flaw in output logic
As far as I can tell, when restarting from a restart file result in the restart file being written again as well as other outputs also being immediately written (in case the `dt` match for the restart and other output).

pgrete:
## PR Summary

There were two (linked) logical issues with restarting simulations.

1. Restart outputs would always be overridden (with the identical data) after a restart and hdf5 outputs (which were triggered at the same time) would be written again, but with an increased counter if the hdf block came before the rst block in the input file.

Reason: we actually store `start_time` in the output file and the simulation time is read in from the `start_time` on restarts.
This behavior is different from `Athena++` where `start_time` was "passive" and never written to a restart file.
For Parthenon this lead to output being triggered immediately on restart as the first condition on the second line in `outputs.cpp`
```c++
        ((ptype->output_params.dt >= 0.0) &&
         ((tm->time == tm->start_time) || (tm->time >= ptype->output_params.next_time) ||
          (tm->time >= tm->tlim)))) {
      if (first && ptype->output_params.file_type != "hst") 
```
would always evaluate true right after a restart.
I now updated that condition to check for `ncycle == 0`.

2. With the fix above, restart outputs would still be overwritten.

Reason: File number and `next_time` for the output are only updated *after* the restart output is written.
Thus, when restarting the conditions are still true (as they triggered the original file write).

I now moved the advancement of the parameters right after the output filename created from the current parameters, so that the output file actually contains the information for the next real output file.

Note, this may lead to slightly different behavior with respect to output file naming when restarting a simulation compared to previously.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


pgrete:
review ping/bump (this should be a quick one)

Also I just triggered the full pipeline to ensure that there's no hiccup with the filenames.

Yurlungur (approved):
Oops my bad. Thought I already approved.

forrestglines (approved):
Looks good!

PR #588, opened 2021-09-09
Switch from nbmax to nneighbors in particles

brryan:
## PR Summary

@jlippuner noticed that regression tests were segfaulting on Darwin after #563 was merged. Looking at the code, in a few places I was looping over `nbmax` rather than `nneighbors` (in general `nbmax >= nneighbors`) which was generating segfaults when accessing neighbor data. I switched to `nneighbors` for all loop bounds in `interface/swarm.cpp` and this fixed the segfault I was seeing on a Darwin power9.

This bug has been around for a while, I think, and I'm unclear why it only just started showing up, and only for certain builds. 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


jlippuner:
This is getting more and more bizarre. On my local workstation, all 4 leapfrog tests now pass with OpenMP in release mode (no MPI, no CUDA). So this PR fixes the issue on my local workstation. But it appears that all 4 leapfrog tests now fail in the Darwin CI (previously only the outflow test without MPI failed). I just retriggered the CI pipeline on Darwin to make sure the failures are reproducible.

jlippuner:
Hmm, all 4 tests failed again in the Darwin CI. So there's clearly another issue still present.

brryan:
> Hmm, all 4 tests failed again in the Darwin CI. So there's clearly another issue still present.

Very odd, the darwin power9 node I tested this on would have been I thought ~identical to the CI. I'll find a way to reproduce.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Current develop
 
 ### Added (new features/APIs/variables/...)
+- [[PR 588]](https://github.com/lanl/parthenon/pull/588) Switch from nbmax to nneighbors in particles
```
pgrete: This should probably go to the "fixed" section.

src/interface/swarm.cpp
```diff
@@ -853,8 +851,8 @@ void Swarm::LoadBuffers_(const int max_indices_size) {
   auto particle_indices_to_send = particle_indices_to_send_;
   pmb->par_for(
       "Pack Buffers", 0, max_indices_size,
-      KOKKOS_LAMBDA(const int n) {        // Max index
-        for (int m = 0; m < nbmax; m++) { // Number of neighbors
+      KOKKOS_LAMBDA(const int n) {                        // Max index
+        for (int m = 0; m < pmb->pbval->nneighbor; m++) { // Number of neighbors
```
pgrete: I'm not sure `pmb->pbval->nneighbor` woud be available on a device (I'd actually be surprised if it'd would be).
So this could be another issue that still causes problems.

pgrete (changes requested):
Good catch. `nbmax` needs to be block local.
I think I spotted another place that may cause problem sat the moment, see review comment.

CHANGELOG.md
brryan: Agreed, thanks (actually never noticed the other `CHANGELOG` sections, oops)

src/interface/swarm.cpp
brryan: Ah good point, my mistake. Let's see if the Darwin CI passes now...

jlippuner: It looks like this definitely fixed *something*. Now only the leapfrog outflow fails, as before (instead of all 4 leapfrog tests). So we should probably focus on what's different with the outflow test compared to the regular leapfrog test. Although, @brryan did you say you also saw segfaults in the regular leapfrog test if you ran them for longer?

brryan: Yeah looks like that only fixed the bug I introduced earlier in this branch... I'm seeing segfaults in the regular leapfrog test at cycle 27 again. Trying to debug now, although gdb says it's not seeing any debug symbols even when I built with `RelWithDebInfo`

brryan: quick update on the particles failure: I'm getting a segfault on this line:
```c++
      PARTHENON_MPI_CHECK(MPI_Isend(bd_var_.send[n].data(), send_size[n],                            
                                    MPI_PARTHENON_REAL, nb.snb.rank, send_tag[n],                    
                                    MPI_COMM_WORLD, &(bd_var_.req_send[nb.bufid])));
```
(line ~100 in `src/bvals/bvals_swarm.cpp`; I checked the macro expansion and it is the MPI call that fails when I run two mpi processes with xterm and gdb on a power9 node with cuda backend). Not clear to me what the segfault is since I can print out `bd_var_.send[n].extent(0)`, `send_size[n]`, `send_tag[n]`, and I can write to `bd_var_.req_send[nb.bufid]` and `nb.snb.rank=1` (the error is occuring on rank 0). I'll keep looking, just a bit confused right now.

brryan: The error: 
image

pgrete:
Does a successful run from the metrics app also imply that Darwin CI passed?

jlippuner:
> Does a successful run from the metrics app also imply that Darwin CI passed?

Yes it does. Darwin CI passed successfully for this branch now. I'll double check locally and then will approve the PR.

pgrete:
Alright, there's definitely still something not working as expected.
Here are my finding on an x86/V100 test:

The *non-MPI* outflow test intermittently fails with
```
$ ./example/particle_leapfrog/particle-leapfrog -i ../tst/regression/test_suites/particle_leapfrog_outflow/parthinput.particle_leapfrog_outflow
#Variables in use:
# Variables:
# Name	Metadata flags
# Sparse Variables:
# Name	sparse id	Metadata flags
# Swarms:
# Swarm	Value	metadata
my particles
    	vx                  None,Provides,Real,Derived
    	id                  None,Provides,Integer,Derived
    	vz                  None,Provides,Real,Derived
    	vy                  None,Provides,Real,Derived


cycle=0 time=0.0000000000000000e+00 dt=1.0825317547305485e-02 zone-cycles/wsec_step=0.00e+00 wsec_total=1.87e-04 wsec_step=7.09e-01
  what():  cudaMemcpy(dst, src, n, cudaMemcpyDefault) error( cudaErrorIllegalAddress): an illegal memory access was encountered /path/to/parthenon/external/Kokkos/core/src/Cuda/Kokkos_CudaSpace.cpp:100

[nvl-004:13883] *** Process received signal ***
[nvl-004:13883] Signal: Aborted (6)
[nvl-004:13883] Signal code:  (-6)
[nvl-004:13883] [ 0] /lib64/libpthread.so.0(+0xf630)[0x2aecbcb4c630]
[nvl-004:13883] [ 1] /lib64/libc.so.6(gsignal+0x37)[0x2aecbcf71387]
[nvl-004:13883] [ 2] /lib64/libc.so.6(abort+0x148)[0x2aecbcf72a78]
[nvl-004:13883] [ 3] /opt/software/GCCcore/9.3.0/lib64/libstdc++.so.6(+0xa2693)[0x2aecbcdfb693]
[nvl-004:13883] [ 4] /opt/software/GCCcore/9.3.0/lib64/libstdc++.so.6(+0xae106)[0x2aecbce07106]
[nvl-004:13883] [ 5] /opt/software/GCCcore/9.3.0/lib64/libstdc++.so.6(+0xae171)[0x2aecbce07171]
[nvl-004:13883] [ 6] /opt/software/GCCcore/9.3.0/lib64/libstdc++.so.6(+0xae3c5)[0x2aecbce073c5]
[nvl-004:13883] [ 7] /path/to/parthenon/build-test/example/particle_leapfrog/particle-leapfrog[0x41aff0]
[nvl-004:13883] [ 8] /path/to/parthenon/build-test/example/particle_leapfrog/particle-leapfrog[0x5d1774]
[nvl-004:13883] [ 9] /path/to/parthenon/build-test/example/particle_leapfrog/particle-leapfrog[0x5ccf6e]
[nvl-004:13883] [10] /path/to/parthenon/build-test/example/particle_leapfrog/particle-leapfrog[0x447c20]
[nvl-004:13883] [11] /path/to/parthenon/build-test/example/particle_leapfrog/particle-leapfrog[0x48126b]
[nvl-004:13883] [12] /path/to/parthenon/build-test/example/particle_leapfrog/particle-leapfrog[0x488bb8]
[nvl-004:13883] [13] /path/to/parthenon/build-test/example/particle_leapfrog/particle-leapfrog[0x48ea59]
[nvl-004:13883] [14] /path/to/parthenon/build-test/example/particle_leapfrog/particle-leapfrog[0x48f074]
[nvl-004:13883] [15] /path/to/parthenon/build-test/example/particle_leapfrog/particle-leapfrog[0x47709b]
[nvl-004:13883] [16] /path/to/parthenon/build-test/example/particle_leapfrog/particle-leapfrog[0x439c8d]
[nvl-004:13883] [17] /path/to/parthenon/build-test/example/particle_leapfrog/particle-leapfrog[0x429555]
[nvl-004:13883] [18] /path/to/parthenon/build-test/example/particle_leapfrog/particle-leapfrog[0x451f09]
[nvl-004:13883] [19] /path/to/parthenon/build-test/example/particle_leapfrog/particle-leapfrog[0x41cabd]
[nvl-004:13883] [20] /lib64/libc.so.6(__libc_start_main+0xf5)[0x2aecbcf5d555]
[nvl-004:13883] [21] /path/to/parthenon/build-test/example/particle_leapfrog/particle-leapfrog[0x422613]
[nvl-004:13883] *** End of error message ***
Aborted
```

Looks like it's related to the unload buffer part (using `cuda-memcheck`):
```
cycle=0 time=0.0000000000000000e+00 dt=1.0825317547305485e-02 zone-cycles/wsec_step=0.00e+00 wsec_total=4.61e-02 wsec_step=1.46e+00
========= Invalid __global__ read of size 8
=========     at 0x00000260 in _ZN6Kokkos4Impl34cuda_parallel_launch_global_memoryINS0_11ParallelForIZN9parthenon5Swarm14UnloadBuffers_EvEUliE_NS_11RangePolicyIJNS_12Experimental16WorkItemProperty20ImplWorkItemPropertyILm1EEEEEENS_4CudaEEEEEvPKT_
=========     by thread (0,26,0) in block (2,0,0)
=========     Address 0x000021b0 is out of bounds
=========     Device Frame:_ZN6Kokkos4Impl34cuda_parallel_launch_global_memoryINS0_11ParallelForIZN9parthenon5Swarm14UnloadBuffers_EvEUliE_NS_11RangePolicyIJNS_12Experimental16WorkItemProperty20ImplWorkItemPropertyILm1EEEEEENS_4CudaEEEEEvPKT_ (_ZN6Kokkos4Impl34cuda_parallel_launch_global_memoryINS0_11ParallelForIZN9parthenon5Swarm14UnloadBuffers_EvEUliE_NS_11RangePolicyIJNS_12Experimental16WorkItemProperty20ImplWorkItemPropertyILm1EEEEEENS_4CudaEEEEEvPKT_ : 0x260)
=========     Saved host backtrace up to driver entry point at kernel launch time
=========     Host Frame:/lib64/libcuda.so.1 (cuLaunchKernel + 0x34e) [0x2d725e]
=========     Host Frame:/opt/software/CUDAcore/11.0.2/lib64/libcudart.so.11.0 [0xf62b]
=========     Host Frame:/opt/software/CUDAcore/11.0.2/lib64/libcudart.so.11.0 (cudaLaunchKernel + 0x1c1) [0x4f5b1]
=========     Host Frame:/path/to/parthenon/build-test/example/particle_leapfrog/particle-leapfrog [0xc1621]
=========     Host Frame:/path/to/parthenon/build-test/example/particle_leapfrog/particle-leapfrog [0x8e8ad]
=========     Host Frame:/path/to/parthenon/build-test/example/particle_leapfrog/particle-leapfrog [0x8f074]
=========     Host Frame:/path/to/parthenon/build-test/example/particle_leapfrog/particle-leapfrog [0x7709b]
=========     Host Frame:/path/to/parthenon/build-test/example/particle_leapfrog/particle-leapfrog [0x39c8d]
=========     Host Frame:/path/to/parthenon/build-test/example/particle_leapfrog/particle-leapfrog [0x29555]
=========     Host Frame:/path/to/parthenon/build-test/example/particle_leapfrog/particle-leapfrog [0x51f09]
=========     Host Frame:/path/to/parthenon/build-test/example/particle_leapfrog/particle-leapfrog [0x1cabd]
=========     Host Frame:/lib64/libc.so.6 (__libc_start_main + 0xf5) [0x22555]
=========     Host Frame:/path/to/parthenon/build-test/example/particle_leapfrog/particle-leapfrog [0x22613]
=========
========= Invalid __global__ read of size 8
=========     at 0x00000260 in _ZN6Kokkos4Impl34cuda_parallel_launch_global_memoryINS0_11ParallelForIZN9parthenon5Swarm14UnloadBuffers_EvEUliE_NS_11RangePolicyIJNS_12Experimental16WorkItemProperty20ImplWorkItemPropertyILm1EEEEEENS_4CudaEEEEEvPKT_
=========     by thread (0,25,0) in block (2,0,0)
=========     Address 0x00002178 is out of bounds
=========     Device Frame:_ZN6Kokkos4Impl34cuda_parallel_launch_global_memoryINS0_11ParallelForIZN9parthenon5Swarm14UnloadBuffers_EvEUliE_NS_11RangePolicyIJNS_12Experimental16WorkItemProperty20ImplWorkItemPropertyILm1EEEEEENS_4CudaEEEEEvPKT_ (_ZN6Kokkos4Impl34cuda_parallel_launch_global_memoryINS0_11ParallelForIZN9parthenon5Swarm14UnloadBuffers_EvEUliE_NS_11RangePolicyIJNS_12Experimental16WorkItemProperty20ImplWorkItemPropertyILm1EEEEEENS_4CudaEEEEEvPKT_ : 0x260)
=========     Saved host backtrace up to driver entry point at kernel launch time
=========     Host Frame:/lib64/libcuda.so.1 (cuLaunchKernel + 0x34e) [0x2d725e]
=========     Host Frame:/opt/software/CUDAcore/11.0.2/lib64/libcudart.so.11.0 [0xf62b]
=========     Host Frame:/opt/software/CUDAcore/11.0.2/lib64/libcudart.so.11.0 (cudaLaunchKernel + 0x1c1) [0x4f5b1]
=========     Host Frame:/path/to/parthenon/build-test/example/particle_leapfrog/particle-leapfrog [0xc1621]
=========     Host Frame:/path/to/parthenon/build-test/example/particle_leapfrog/particle-leapfrog [0x8e8ad]
=========     Host Frame:/path/to/parthenon/build-test/example/particle_leapfrog/particle-leapfrog [0x8f074]
=========     Host Frame:/path/to/parthenon/build-test/example/particle_leapfrog/particle-leapfrog [0x7709b]
=========     Host Frame:/path/to/parthenon/build-test/example/particle_leapfrog/particle-leapfrog [0x39c8d]
=========     Host Frame:/path/to/parthenon/build-test/example/particle_leapfrog/particle-leapfrog [0x29555]
=========     Host Frame:/path/to/parthenon/build-test/example/particle_leapfrog/particle-leapfrog [0x51f09]
=========     Host Frame:/path/to/parthenon/build-test/example/particle_leapfrog/particle-leapfrog [0x1cabd]
=========     Host Frame:/lib64/libc.so.6 (__libc_start_main + 0xf5) [0x22555]
=========     Host Frame:/path/to/parthenon/build-test/example/particle_leapfrog/particle-leapfrog [0x22613]

```


As for the test *with MPI* I  noticed that the outflow test has a hardcoded number of ranks set to 1, so it doesn't really test communication across blocks on different ranks.
Was this set on purpose or can we update the test to use more ranks?

brryan:
@pgrete Thanks for double checking! Yes this is the error I was seeing yesterday, and the load/unload buffer logic was broken because I was mixing `n` with `pmb->pbval->neighbor[n].bufid` and `pmb->pbval->neighbor[n].targetid`. I believe I fixed this but I'm still cleaning up my branch, I'll try to get this wrapped up today

EDIT: Hm my guess is the ranks = 1 is a mistake I made setting up the test, I'll check on that

brryan:
@jlippuner @pgrete OK new round of fixes in and everything is working for me now. The darwin CI just passed but I had to rerun it after removing a leftover `printf` call, the checks here should update to complete pretty soon.

`ctest -R particle` works for me on a Darwin x86 partition (EDIT: on a Volta GPU) just now on `ofe29ce` (except for the "slurm not enabled" when ctest runs the 1-rank tests without `mpiexec` but I think this is just an environment issue on my end)

EDIT2: `regression_mpi_test:particle_leapfrog_outflow` now actually does use more than 1 MPI rank

src/interface/swarm.cpp
```diff
@@ -851,20 +860,22 @@ void Swarm::LoadBuffers_(const int max_indices_size) {
   auto &bdvar = vbswarm->bd_var_;
   auto num_particles_to_send = num_particles_to_send_;
   auto particle_indices_to_send = particle_indices_to_send_;
+  auto neighbor_buffer_index = neighbor_buffer_index_;
   pmb->par_for(
       "Pack Buffers", 0, max_indices_size,
```
pgrete: Just noticed this now: This is a `par_for`, i.e., the bounds are inclusive. Is it correct that the upper bound should be `max_indices_size`, i.e., there's data that can be accessed at `myvar(max_indices_size)` or should this be `max_indices_size - 1` as loop bound?

src/interface/swarm.cpp
```diff
@@ -941,12 +954,14 @@ void Swarm::CountReceivedParticles_() {
 
 void Swarm::UpdateNeighborBufferReceiveIndices_(ParArrayND<int> &neighbor_index,
                                                 ParArrayND<int> &buffer_index) {
-  const int max_neighbor = vbswarm->bd_var_.nbmax;
+  auto pmb = GetBlockPointer();
   auto neighbor_index_h = neighbor_index.GetHostMirror();
+  // buffer_index is the index of the particle in this swarm buffer, not the neighbor
+  // buffer index
```
pgrete: I assume you added this comment because this `buffer_index` is used in a non-typical way here.
I'm not sure I fully follow the comment. Am I reading it correct that there's a dual use of the buffer index array, e.g., is it used here for the particle index whereas somewhere else for the neighbor buffer index?
If that's the case is there a way to make this cleaner/easier to follow? For example, calling it a "work" array or sth so that it's clear that there is dual use?

tst/regression/test_suites/particle_leapfrog/particle_leapfrog.py
```diff
@@ -50,7 +50,9 @@ def Analyse(self, parameters):
...
         )
-        return (
-            structured_to_unstructured(final_data[["x", "y", "z", "vx", "vy", "vz"]])
-            == ref_data
-        ).all()
+        final_data = structured_to_unstructured(
+            final_data[["x", "y", "z", "vx", "vy", "vz"]]
+        )
+        if ref_data.shape != final_data.shape:
+            return False
```
pgrete: If we can fail because of different conditions, I suggest to expose that info:
```suggestion
        if ref_data.shape != final_data.shape:
            print("TEST FAIL: Mismatch between actual and reference data shape.")
            return False
```

tst/regression/test_suites/particle_leapfrog_outflow/particle_leapfrog_outflow.py
```diff
@@ -36,19 +36,22 @@ def Analyse(self, parameters):
...
         )
-        return (
-            structured_to_unstructured(final_data[["x", "y", "z", "vx", "vy", "vz"]])
-            == ref_data
-        ).all()
+        final_data = structured_to_unstructured(
+            final_data[["x", "y", "z", "vx", "vy", "vz"]]
+        )
+        if ref_data.shape != final_data.shape:
+            return False
```
pgrete: As above
```suggestion
        if ref_data.shape != final_data.shape:
            print("TEST FAIL: Mismatch between actual and reference data shape.")
            return False
```

tst/regression/test_suites/particle_leapfrog_outflow/particle_leapfrog_outflow.py
```diff
@@ -36,19 +36,22 @@ def Analyse(self, parameters):
...
         # here, outflow boundaries are used such that some of the particles escape
         # this tests code paths not otherwise present (when blocks have >0 neighbors)
         ref_data = np.array(
             [
-                [-0.1, 0.3, 0.3353164693, 0.0, 0.0, 0.5],
-                [0.3906329387, 0.4706329387, -0.0293670613, 1.0, 1.0, 1.0],
+                [0.4, 0.2, 0.3, 1.0, 0.0, 0.0],
+                [0.4, 0.4, 0.3, 0.0, 1.0, 0.0],
+                [-0.1, 0.3, 0.45, 0.0, 0.0, 0.5],
             ]
```
pgrete: I see why you changed the answer here in combination with the `tlim=0.5` in the input file.
Looks like the original test was not working correctly.
Do you think it's worth updating the test case (and, thus, also `particle_leapfrog.cpp`) to include more particles that pass through all boundaries? Just to make sure everything works as expected?

pgrete (approved):
LGTM and test also pass when I tested on x86/cuda (tried 10x)

src/interface/swarm.cpp
brryan: Yes you are right, thanks, I must've gotten confused between this and the `max_active_index` variable which doesn't need the `- 1` for inclusive loop bounds. Fixed.

src/interface/swarm.cpp
brryan: This comment may only cause extra confusion... What I wanted to highlight was that this is the index of each particle in its buffer array, not the `bufid` variable that identifies the particle buffer for MPI calls in the `bvals` machinery. But that is a "buffer ID", not a "buffer index", so technically there isn't dual use. I've changed this to just `// Index of each particle in its received buffer"` -- is that OK?

tst/regression/test_suites/particle_leapfrog/particle_leapfrog.py
brryan: Good idea, thanks

tst/regression/test_suites/particle_leapfrog_outflow/particle_leapfrog_outflow.py
brryan: Yes this is a good idea. I added particles that test all corner and face crossings (not edge crossings). Golds updated, results look good to me.

pgrete: Looks also good to me! I'll trigger the full suite now to to make sure.

src/interface/swarm.cpp
pgrete: yes, that is way clearer :+1:

pgrete:
@jlippuner feel free to merge once you got the chance to review and approve.

jlippuner (approved):
Sorry for the delay. I was trying to get MPI+CUDA working on my local workstation, but couldn't because we rely on GPUDirect (see #593). But this all looks great now and the leapfrog tests pass on my local workstation. I understand you and @pgrete have also tested it on other platforms. Great job!

brryan:
@jlippuner @pgrete Thanks for all your help! The particles implementation is in a much better place now. And sorry I broke the CI!

PR #597, opened 2021-09-30
Added tuning functionality for HDF5 output

brtnfld:
## PR Summary

Added new  HDF5 optimizations, and add a mechanism through the use of environment variables to tune HDF5 and MPI IO.
The results of these improvements were shown in a report.

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted


src/outputs/parthenon_hdf5.cpp
```diff
@@ -415,17 +488,26 @@ void PHDF5Output::WriteOutputFileImpl(Mesh *pm, ParameterInput *pin, SimTime *tm
     ~MPI_InfoDeleter() { MPI_Info_free(&info); }
   } delete_info{FILE_INFO_TEMPLATE};
 
-  PARTHENON_HDF5_CHECK(H5Pset_sieve_buf_size(acc_file, 262144));
```
brtnfld, line 418: I disabled this by default, I can change that if needed. I need to look into the effects of always enabling this,  possibly with a poor choice of the value specified.

src/outputs/parthenon_hdf5.cpp
```diff
@@ -401,9 +431,52 @@ void PHDF5Output::WriteOutputFileImpl(Mesh *pm, ParameterInput *pin, SimTime *tm
...
+  bool exists, exists2;
+
+  // Sets the maximum size of the data sieve buffer, in bytes.
+  // The sieve_buf_size should be equal to a multiple of the disk block size
+  size_t sieve_buf_size = HDF5::get<size_t>("H5_sieve_buf_size", 256*KB, exists);
+  if(exists)
+     PARTHENON_HDF5_CHECK(H5Pset_sieve_buf_size(acc_file, sieve_buf_size));
+
+  // Sets the minimum metadata block size, in bytes.
+  hsize_t meta_block_size = HDF5::get<hsize_t>("H5_meta_block_size", 16*MB, exists);
```
brtnfld: Not yet sure if this should be enabled by default.

src/outputs/parthenon_hdf5.cpp
```diff
@@ -415,17 +488,26 @@ void PHDF5Output::WriteOutputFileImpl(Mesh *pm, ParameterInput *pin, SimTime *tm
...
+    PARTHENON_MPI_CHECK(MPI_Info_set(FILE_INFO_TEMPLATE, "cb_block_size",  cb_block_size));
+    // Specifies the total buffer space that can be used for collective buffering on each target node, 
+    // usually a multiple of cb_block_size
+    char *cb_buffer_size = HDF5::get<char *>("cb_buffer_size", (char*)"4194304", exists);
+    PARTHENON_MPI_CHECK(MPI_Info_set(FILE_INFO_TEMPLATE, "cb_buffer_size", cb_buffer_size));
+  }
 
   /* tell the HDF5 library that we want to use MPI-IO to do the writing */
   PARTHENON_HDF5_CHECK(H5Pset_fapl_mpio(acc_file, MPI_COMM_WORLD, FILE_INFO_TEMPLATE));
-  PARTHENON_HDF5_CHECK(H5Pset_fapl_mpio(acc_file, MPI_COMM_WORLD, MPI_INFO_NULL));
```
brtnfld, line 428: It does not look like collective_buffering was ever enabled, the second call to the same function would override the MPI info with MPI_INFO_NULL.

brtnfld:
I would guess as known good tuning settings per machine or filesystem become known that some of these values could be set at build time with #define to avoid remembering to set env. variables.

src/outputs/parthenon_hdf5.cpp
```diff
@@ -401,9 +431,52 @@ void PHDF5Output::WriteOutputFileImpl(Mesh *pm, ParameterInput *pin, SimTime *tm
...
+
+  hsize_t threshold; // Threshold value. Setting to 0 forces everything to be aligned.
+  hsize_t alignment; // Alignment value.
+
+  threshold = HDF5::get<hsize_t>("H5_alignment_threshold", 1*MB, exists);
+  alignment = HDF5::get<hsize_t>("H5_alignment_alignment", 16*MB, exists2);
+  if(exists || exists2)
+     PARTHENON_HDF5_CHECK(H5Pset_alignment(acc_file, threshold, alignment));
+
+  // Defer metadata flush
```
brtnfld, line 447: I'm checking if defer_metadata_flush makes much of a difference.

docs/outputs.md
```diff
@@ -32,6 +32,22 @@ be used to set the compression level (between 1 and 9, default is 5). Compressio
...
+| H5_alignment_alignment | disabled | int | The alignment value, in bytes, of H5Pset_alignment. If a value is not set then the default is 16 MiB.  Setting the environment variable automatically enables alignment.  H5Pset_alignment sets the alignment properties of a file access property list. Choose an alignment  that is a multiple of the disk block size, enabling this usually shows better performance on GPFS.   Enabling may increase the file size significantly. |
+| H5_defer_metadata_flush | disabled | int | Value of 1 enables deferring metadata flush. Value of 0 disables. |
+| MPI_access_style | enabled | string | Specifies the manner in which the file will be accessed until the file is closed. Default is "write_once" |
+| MPI_collective_buffering | disabled | int | Value of 1 enables MPI collective buffering. Value of 0 disables. |
+| MPI_cb_block_size | N/A | int | Sets the block size, in bytes, to be used for collective buffering file access. Default is 1 MiB. |
+| MPI_cb_buffer_size | N/A | int | Sets the total buffer space, in bytes, that can be used for collective buffering on each target node,  usually a multiple of cb_block_size. Default is 4 MiB. |
```
Yurlungur: This added documentation is great! I notice there's no comment on how these options effect performance. I suppose it is simply problem and hardware dependent?

src/outputs/parthenon_hdf5.cpp
```diff
@@ -36,6 +36,10 @@
 #include "outputs/parthenon_hdf5.hpp"
 #include "utils/string_utils.hpp"
 
+#define GB            (1024 * 1024 * 1024)
+#define MB            (1024 * 1024)
+#define KB            (1024)
```
Yurlungur: Is there a reason to use macros here, instead of `constexpr int`?

In the latter case, perhaps we should move these into global `decs` so they're available everywhere?

src/outputs/parthenon_hdf5.cpp
```diff
@@ -77,6 +81,32 @@ void HDF5WriteAttribute(const std::string &name, const std::vector<bool> &values
...
+  } else {
+    // Environment variable is set and value is set
+    if (std::is_same<T, char*>::value)
+      res = (T)value;
+    else
+      std::istringstream(value) >> res;
+    return res;
+  }
+}
+
```
Yurlungur: This doesn't appear HDF5-specific to me. Perhaps we should move it into our utilities header or something?

Yurlungur (approved):
Thanks for this excellent contribution, @brtnfld ! It looks like you're still experimenting a bit, but to my mind this can go in when you're ready. (Although I do have a few non-blocking suggestions below.)

I wonder if we could include some of the results of your analysis on what these options do in the wiki or documentation? Would that be okay?

docs/outputs.md
brtnfld: > I wonder if we could include some of the results of your analysis on what these options do in the wiki or documentation? Would that be okay?

Yes, that would be okay.

Yurlungur: Great! :+1:

brtnfld: > This added documentation is great! I notice there's no comment on how these options effect performance. I suppose it is simply problem and hardware dependent?

It would be system and hardware-dependent, one thing I hope this PR does is to make exploring how the different settings would affect performance. But some general advice for setting them can be given.

I left some of the tuning parameters enabled in the original code as I'm not sure that maybe there is a use case that they are beneficial.

Yurlungur: Got it. That makes sense.

brtnfld: I've done some more test runs with them enabled and disabled for some of them, and I'll add them to the report.

src/outputs/parthenon_hdf5.cpp
brtnfld: No reason. If you think they would be helpful generally, it would be good to make them available.

src/outputs/parthenon_hdf5.cpp
brtnfld: Correct, nothing generally specific for HDF5. This could be moved.

src/outputs/parthenon_hdf5.cpp
Yurlungur: I think I like putting this [here](https://github.com/lanl/parthenon/blob/develop/src/defs.hpp) but it's also fine to leave it where it is.

src/outputs/parthenon_hdf5.cpp
Yurlungur: I like putting this [here](https://github.com/lanl/parthenon/blob/develop/src/utils/utils.hpp). But it's also fine to leave it where it is.

docs/outputs.md
brtnfld: I changed most of the options to be disabled to be on the safe side, although set_aligment is usually suitable for good performance, so if the file size is still reasonable after enabling, this can be enabled by default.

src/outputs/parthenon_hdf5.cpp
brtnfld: updated and moved.

src/outputs/parthenon_hdf5.cpp
brtnfld: moved.

Yurlungur (approved):
@brtnfld is this ready for merge from your perspective? If so can we get another pair of eyes? Maybe @pgrete or @forrestglines ?

I think to get the tests to run, I will need to make a new branch on the repo itself, rather than a fork. But that shouldn't be a problem.

docs/outputs.md
```diff
@@ -32,6 +32,22 @@ be used to set the compression level (between 1 and 9, default is 5). Compressio
...
+| H5_defer_metadata_flush | disabled | int | Value of 1 enables deferring metadata flush. Value of 0 disables. Experiment with before using. |
+| MPI_access_style | enabled | string | Specifies the manner in which the file will be accessed until the file is closed. Default is "write_once" |
+| MPI_collective_buffering | disabled | int | Value of 1 enables MPI collective buffering. Value of 0 disables. Experiment with before using. |
+| MPI_cb_block_size | N/A | int | Sets the block size, in bytes, to be used for collective buffering file access. Default is 1 MiB. |
+| MPI_cb_buffer_size | N/A | int | Sets the total buffer space, in bytes, that can be used for collective buffering on each target node,  usually a multiple of cb_block_size. Default is 4 MiB. |
```
pgrete: The entries in the "default" column don't seem to match what is written as default in the description for several parameters.

src/defs.hpp
```diff
@@ -153,6 +153,11 @@ using pMeshBlockApplicationData_t = std::unique_ptr<MeshBlockApplicationData>;
 // we still need to define this somewhere, though
 inline MeshBlockApplicationData::~MeshBlockApplicationData() {}
 
+// Convience definitions
+constexpr int GiB = 1024 * 1024 * 1024;
```
pgrete: We're getting really close to max `int` here -- especially when using these in math expression further down.
How about using `uint64_t` instead?

src/outputs/parthenon_hdf5.cpp
```diff
@@ -401,9 +401,56 @@ void PHDF5Output::WriteOutputFileImpl(Mesh *pm, ParameterInput *pin, SimTime *tm
...
+#endif
+
+  bool exists, exists2;
+
+  // Sets the maximum size of the data sieve buffer, in bytes.
+  // The sieve_buf_size should be equal to a multiple of the disk block size
+  // Default: Disabled
+  size_t sieve_buf_size = Env::get<size_t>("H5_sieve_buf_size", 256*KiB, exists);
+  if(exists)
+     PARTHENON_HDF5_CHECK(H5Pset_sieve_buf_size(acc_file, sieve_buf_size));
```
pgrete: Our format check will likely fail here, but this can be addressed automatically once we move the branch over.
Nevertheless, adding braces would be great (here and below).

src/utils/utils.hpp
```diff
@@ -50,6 +50,38 @@ void Report();
...
 } // namespace SignalHandler
 
+//----------------------------------------------------------------------------------------
+//! Env
+//  \brief static function to check and retrieve environment settings
+namespace Env {
+
+// template to get environment variables
+template<typename T>
+static T get(const char* name, T DefaultVal, bool exists) {
```
pgrete: Does this currently work as expected?
`exists` is neither a pointer nor a reference.

Also, for consistent style lowercase `DefaultVal` would be great.

src/utils/utils.hpp
```diff
@@ -50,6 +50,38 @@ void Report();
...
+  exists = true;
+  char* value = std::getenv(name);
+
+  // Environment variable is not set
+  if (!value) {
```
pgrete:
```suggestion
  if (value == nullptr) {
```

src/utils/utils.hpp
```diff
@@ -50,6 +50,38 @@ void Report();
...
+//----------------------------------------------------------------------------------------
+//! Env
+//  \brief static function to check and retrieve environment settings
+namespace Env {
+
+// template to get environment variables
+template<typename T>
+static T get(const char* name, T DefaultVal, bool exists) {
+  exists = true;
+  char* value = std::getenv(name);
```
pgrete:
```suggestion
  const char* value = std::getenv(name);
```

src/utils/utils.hpp
```diff
@@ -50,6 +50,38 @@ void Report();
...
+
+  // Environment variable is not set
+  if (!value) {
+    exists = false;
+    return DefaultVal;
+  }
+
+  T res;
+  // Environment variable is set but no value is set, use the default
+  if( !value[0]) {
```
pgrete:
```suggestion
  if( value[0] == '\0') {
```
Suggestion for increased readability.

src/outputs/parthenon_hdf5.cpp
```diff
@@ -415,17 +462,26 @@ void PHDF5Output::WriteOutputFileImpl(Mesh *pm, ParameterInput *pin, SimTime *tm
...
+  if(collective_buffering) {
+    PARTHENON_MPI_CHECK(MPI_Info_set(FILE_INFO_TEMPLATE, "collective_buffering", "true"));
+    // Specifies the block size to be used for collective buffering file acces
+     char *cb_block_size = Env::get<char *>("MPI_cb_block_size", (char*)"1048576", exists);
+    PARTHENON_MPI_CHECK(MPI_Info_set(FILE_INFO_TEMPLATE, "cb_block_size",  cb_block_size));
+    // Specifies the total buffer space that can be used for collective buffering on each target node, 
+    // usually a multiple of cb_block_size
+    char *cb_buffer_size = Env::get<char *>("MPI_cb_buffer_size", (char*)"4194304", exists);
+    PARTHENON_MPI_CHECK(MPI_Info_set(FILE_INFO_TEMPLATE, "cb_buffer_size", cb_buffer_size));
+  }
```
pgrete: Why did the logic on how to handle env vars change here compared to above (where `exits` would be use to distinguish between default and not set at all)?

pgrete (commented):
Thanks for taking care of this! I'm already looking forward to test this in the wild.

I left a couple of (mostly minor) comments that'd be great get addressed prior to moving the branch to this repo for testing.
Also you can try local testing to ensure things are working as expected on your workstation using `ctest -L regression -LE "mpi-no|perf-reg"`

docs/outputs.md
brtnfld: Changed "default" to "initial state" to avoid confusion.

src/outputs/parthenon_hdf5.cpp
brtnfld: changed.

src/utils/utils.hpp
brtnfld: changed.

src/utils/utils.hpp
brtnfld: changed.

src/utils/utils.hpp
brtnfld: changed

src/outputs/parthenon_hdf5.cpp
brtnfld: fixed to use exists

brtnfld:
> @brtnfld is this ready for merge from your perspective? If so can we get another pair of eyes? Maybe @pgrete or @forrestglines ?
> 
> I think to get the tests to run, I will need to make a new branch on the repo itself, rather than a fork. But that shouldn't be a problem.

I have seen an improvement forcing HDF5 to use the 1.8 file format specifications, so I will add a new optimization.  It does not look like you are using any 1.10 or 1.12 features.

pgrete:
@brtnfld thanks for the updates.
Regarding version 1.8, is there any feature that we may want to use from 1.8?
And are there any side-effect with respect to compression?

brtnfld:
> Regarding version 1.8, is there any feature that we may want to use from 1.8?

Do you mean features from newer versions 1.10,1.12? I list of features requiring new 1.10, 1.12 file formats are listed in  Table I, pg 7. https://github.com/HDFGroup/hdf5doc/blob/master/RFCs/HDF5_Library/Library_version_bounds/RFC-bounds.docx

Namely, the SWMR and virtual dataset features, which I don't think you will need to use. There are other upcoming features such as async, subfiling and multi-dataset I/O that are features we will want to explore.

> And are there any side-effect with respect to compression?

Not that I've encountered.

brtnfld:
FYI, this  PR is not necessarily for chunked datasets, but I did notice that Parthenon is not setting a chunk cache size, so we may be able to improve chunked dataset performance (and hence compression performance) by tuning this.

brtnfld:
Could you give me insight into why the 5D array is ordered as (block , nx , ny , nz , num vars)?  I would think a better ordering would be (block, nvars, nx, ny, nz), and a chunk size of (1, 1, nx, ny, nz). If the selections in the XDMF reader are not contiguous on disk then disk seeks could be excessive.

Also, the ordering of this loop could have a fair amount of cache misses.

```
          auto v_h = v->data.GetHostMirrorAndCopy();
          for (int k = out_kb.s; k <= out_kb.e; ++k) {
            for (int j = out_jb.s; j <= out_jb.e; ++j) {
              for (int i = out_ib.s; i <= out_ib.e; ++i) {
                for (int l = 0; l < vlen; ++l) {
                  tmpData[index++] = static_cast<OutT>(v_h(l, k, j, i));
```

Yurlungur:
> Could you give me insight into why the 5D array is ordered as (block , nx , ny , nz , num vars)? I would think a better ordering would be (block, nvars, nx, ny, nz), and a chunk size of (1, 1, nx, ny, nz). If the selections in the XDMF reader are not contiguous on disk then disk seeks could be excessive.

Hmm... actually I think the right ordering is probably (block, nvars, nz, ny, nz), as that reflects the ordering of the indices in memory.

> 
> Also, the ordering of this loop could have a fair amount of cache misses.
> 
> ```
>           auto v_h = v->data.GetHostMirrorAndCopy();
>           for (int k = out_kb.s; k <= out_kb.e; ++k) {
>             for (int j = out_jb.s; j <= out_jb.e; ++j) {
>               for (int i = out_ib.s; i <= out_ib.e; ++i) {
>                 for (int l = 0; l < vlen; ++l) {
>                   tmpData[index++] = static_cast<OutT>(v_h(l, k, j, i));
> ```

:+1: good catch

pgrete:
> Could you give me insight into why the 5D array is ordered as (block , nx , ny , nz , num vars)? I would think a better ordering would be (block, nvars, nx, ny, nz), and a chunk size of (1, 1, nx, ny, nz). If the selections in the XDMF reader are not contiguous on disk then disk seeks could be excessive.
> 
> Also, the ordering of this loop could have a fair amount of cache misses.
> 
> ```
>           auto v_h = v->data.GetHostMirrorAndCopy();
>           for (int k = out_kb.s; k <= out_kb.e; ++k) {
>             for (int j = out_jb.s; j <= out_jb.e; ++j) {
>               for (int i = out_ib.s; i <= out_ib.e; ++i) {
>                 for (int l = 0; l < vlen; ++l) {
>                   tmpData[index++] = static_cast<OutT>(v_h(l, k, j, i));
> ```

That may be a question for @jlippuner : Was this potentially related to assuming some correlation between variables so that compression is more efficient?

Also IIRC we already briefly talked about this during the call and identified this as a potential improvement.


> FYI, this PR is not necessarily for chunked datasets, but I did notice that Parthenon is not setting a chunk cache size, so we may be able to improve chunked dataset performance (and hence compression performance) by tuning this.

I thought that data compression requires chunking? Is that not the case?


Finally, with respect to the file format.
My main concern is compatibility (also with other analysis codes).
I'm strongly in favor of making changes once that improve performance, but it'd prefer if we only make at most one change to the data format itself so that we don't need to track various data formats in different Parthenon version for different analysis code.

Of course, the format spec/design is different from the original tuning here.
How about we proceed by not specifying a format and add tuning in this PR and then talk about format and how we write data in a separate PR?

Yurlungur:
> My main concern is compatibility (also with other analysis codes). I'm strongly in favor of making changes once that improve performance, but it'd prefer if we only make at most one change to the data format itself so that we don't need to track various data formats in different Parthenon version for different analysis code.
> 
> Of course, the format spec/design is different from the original tuning here. How about we proceed by not specifying a format and add tuning in this PR and then talk about format and how we write data in a separate PR?

I would be in favor of this approach. I do think we can change things now. It's not too late to make breaking changes to file formats.

brtnfld:
> I thought that data compression requires chunking? Is that not the case?

Yes, it needs to be a chunked dataset, chunk caching is different. But that is mainly an issue for partial I/O so it might not help in most situations. It is good that you are chunking with a size of (1,nx,ny,nz,1).

> 
> Finally, with respect to the file format. My main concern is compatibility (also with other analysis codes). I'm strongly in favor of making changes once that improve performance, but it'd prefer if we only make at most one change to the data format itself so that we don't need to track various data formats in different Parthenon version for different analysis code.
> 
> Of course, the format spec/design is different from the original tuning here. How about we proceed by not specifying a format and add tuning in this PR and then talk about format and how we write data in a separate PR?

I'm okay with that, and then we could test the format change on its own to see if it is worth it. We are in the process of studying parallel compression in HDF5, so this would be a good use case.

brtnfld:
Is there anything I need to further address with this PR?

pgrete:
> Is there anything I need to further address with this PR?

Not as far as I can tell at the moment.
I'll pull it over and trigger a full test run and apply formatting.

pgrete:
Merged via #602

PR #598, opened 2021-10-01
Add tensor field indexing functionality to PackIndexMap

lroberts36:
## PR Summary
Add an internal map to PackIndexMap which contains the shape of all variables in the associated VariablePack and also define a method in PackIndexMap which gets the flattened index of a tensor component given the components indices. Also defines a method to return the shape of any object included in the PackIndexMap. This is an attempt to simplify and reduce the likelihood of error when indexing into tensor fields in VariablePacks.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/src/interface/variable_pack.hpp
+++ b/src/interface/variable_pack.hpp
@@ -141,22 +141,34 @@ class PackIndexMap {
     }
 
     // Check that the correct dimensionality is being specified
-    assert(indices.size() == itr_shape->second.size());
+    if (indices.size() != itr_shape->second.size()) {
+      PARTHENON_THROW("Wrong number of indices for variable " + key + ".")
+    }
 
     // Find the flat index from the indices assuming fastest moving index is
     // the rightmost index
     int idx = 0;
     if (indices.size() > 0) {
-      idx = indices[0];
-      assert(indices[0] < itr_shape->second[0]);
-      for (int idim = 1; idim < indices.size(); ++idim) {
-        assert(indices[idim] < itr_shape->second[idim]);
-        idx = indices[idim] + idim * itr_shape->second[idim];
+      // for (int idim = 0; idim < indices.size(); ++idim) {
+      for (int idim = indices.size() - 1; idim >= 0; --idim) { 
+        if (indices[idim] >= itr_shape->second[idim]) {
+          PARTHENON_THROW("Index " + std::to_string(indices[idim]) +
+                          " too large for dimension " + std::to_string(idim) + " of " +
+                          key + ".")
+        }
+        idx = indices[idim] + idx * itr_shape->second[idim];
       }
     }
-    idx += itr->second.first;
+    idx +=
+        itr->second.first; // Offset for position relative to other variables in the pack
 
     return idx;
+  }
+
+  std::vector<int> GetShape(const std::string &key) {
+    auto itr_shape = shape_map_.find(key);
+    if (itr_shape == shape_map_.end()) return {-1};
+    return itr_shape->second;
   }
 
   bool Has(std::string const &base_name, int sparse_id = InvalidSparseID) const {
@@ -410,9 +422,9 @@ void FillVarView(const CellVariableVector<T> &vars, bool coarse,
     }
 
     std::vector<int> shape;
-    if (v->GetDim(6) > 0) shape.push_back(v->GetDim(6));
-    if (v->GetDim(5) > 0) shape.push_back(v->GetDim(5));
-    if (v->GetDim(4) > 0) shape.push_back(v->GetDim(4));
+    if (v->GetDim(4) > 1) shape.push_back(v->GetDim(4));
+    if (v->GetDim(5) > 1) shape.push_back(v->GetDim(5));
+    if (v->GetDim(6) > 1) shape.push_back(v->GetDim(6));
 
     if (pvmap != nullptr) {
       pvmap->insert(
@@ -480,9 +492,9 @@ void FillFluxViews(const CellVariableVector<T> &vars, const int ndim,
     }
 
     std::vector<int> shape;
-    if (v->GetDim(6) > 0) shape.push_back(v->GetDim(6));
-    if (v->GetDim(5) > 0) shape.push_back(v->GetDim(5));
-    if (v->GetDim(4) > 0) shape.push_back(v->GetDim(4));
+    if (v->GetDim(4) > 1) shape.push_back(v->GetDim(4));
+    if (v->GetDim(5) > 1) shape.push_back(v->GetDim(5));
+    if (v->GetDim(6) > 1) shape.push_back(v->GetDim(6));
 
     if (pvmap != nullptr) {
       pvmap->insert(
--- a/src/interface/variable_pack.hpp
+++ b/src/interface/variable_pack.hpp
@@ -79,31 +79,31 @@ class FlatIdx {
 
   KOKKOS_FORCEINLINE_FUNCTION
   int operator()() const {
-    if (ndim_ != 0) PARTHENON_DEBUG_FAIL("Wrong number of dimensions.");
+    PARTHENON_DEBUG_REQUIRE(ndim_ == 0, "Wrong number of dimensions.");
     return offset_;
   }
 
   KOKKOS_FORCEINLINE_FUNCTION
   int operator()(const int idx1) const {
-    if (ndim_ != 1) PARTHENON_DEBUG_FAIL("Wrong number of dimensions.");
-    if (idx1 >= shape_[0]) PARTHENON_DEBUG_FAIL("Idx1 too large.");
+    PARTHENON_DEBUG_REQUIRE(ndim_ == 1, "Wrong number of dimensions.");
+    PARTHENON_DEBUG_REQUIRE(idx1 < shape_[0], "Idx1 too large.");
     return offset_ + idx1;
   }
 
   KOKKOS_FORCEINLINE_FUNCTION
   int operator()(const int idx1, const int idx2) const {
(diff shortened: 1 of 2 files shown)


src/interface/variable_pack.hpp
```diff
@@ -49,6 +49,9 @@ using SwarmVarList = std::forward_list<std::shared_ptr<ParticleVariable<T>>>;
 // the pairs represent interval (inclusive) of those indices
 using IndexPair = std::pair<int, int>;
 
+// Used for storing the shapes of variable fields
+using Shape = std::vector<int>;
```
Yurlungur, line 53: If we wanted to be able to put things on device, this could be a class that owns a `ParArray1D<int>` and the `GetFlatIdx` method. And you could get that object for a given key.

Yurlungur (approved):
Looks good to me. So the way you've set this up, you have to compute the flat index on host before passing into a kernel, since the `PackIndexMap` can't be copied to device.

An alternative approach, if you wanted, would be to create  a device context like
```C++
class FlatIdx {
 public:
  template <typename...Ts>
  int operator()(Ts... idx_pack) { /* stuff */ }
 private:
  ParArray1D<int> shape_;
}
```
which would then be called like
```C++
PackIndexMap imap;
auto pack = rc->PackVariables(names, imap);
auto flat_idx = imap[name];
// somewhere in a kernel
pack(flat_idx(3,2), k, j, i);
```

I'm happy with either way you want to do things.

src/interface/variable_pack.hpp
lroberts36: I have no strong preference, although I think the flat indices should probably be pre-computed on the host. I guess this could be done during construction in your suggested `FlatIdx` class, although I guess the that would require a `ParArrayND` member of indices rather than just a `ParArray1D` containing the shape .

Yurlungur: If you like them being precomputed on the host, then I think the way you've written it is perfectly good.

Yurlungur:
@lroberts36 if this is ready for merge, we should remove the WIP and get one more pair of eyes on it.

lroberts36:
WIP removed

src/interface/variable_pack.hpp
lroberts36: hm, looking at one of the use cases I have a little bit more, there might be some advantage to calculating the FlatIdx on the device since the indexing array could be reasonably large. I am going to revert back to WIP and implement your suggestion since it is more flexible.

Yurlungur: Sounds good.

Yurlungur: Ping me when you're ready for re-review

lroberts36: It is ready to be looked at again. I ran it through tests on my machine and everything works as expected, but I have only tested purely on CPU.

src/interface/variable_pack.hpp
```diff
@@ -49,6 +49,53 @@ using SwarmVarList = std::forward_list<std::shared_ptr<ParticleVariable<T>>>;
...
+
+// Index arbitrary rank fields into flattened indices in a VariablePack
+class FlatIdx {
+ public:
+  FlatIdx(std::vector<int> shape, int offset)
+      : shape_("shape", shape.size()), offset_(offset) {
+    for (int i = 0; i < shape.size(); ++i) {
+      shape_(i) = shape[i];
+    }
```
Yurlungur: This isn't going to work on GPU as is. You'll need to copy the vector into a host mirror and then do a Kokkos deep copy. Alternatively, you could make your `shape` array above the Kokkos view object and do the deep copy outside the `FlatIdx` object.

There's a trade-off here. If the deep copy is performed when the `PackIndexMap` is filled, then it will happen for every variable every time a pack is constructed. (Fortunately packs are cached, so when you ask for a pack it's only reconstructed if something has changed.)

On the other hand, if the deep copy is performed when the `FlatIdx` object is requested, then there will be a small deep copy outside every relevant kernel where the `FlatIdx` object is needed. But it wont' be called when `PackVariables` is called. So that could be advantageous if there aren't many places where `FlatIdx` is needed.

I don't know. What do you think?

src/interface/variable_pack.hpp
```diff
@@ -49,6 +49,53 @@ using SwarmVarList = std::forward_list<std::shared_ptr<ParticleVariable<T>>>;
...
+      PARTHENON_THROW("Wrong number of indices for variable.")
+    }
+
+    // Find the flat index from the indices assuming fastest moving index is
+    // the rightmost index
+    int idx = 0;
+    if (indices.size() > 0) {
+      for (int idim = indices.size() - 1; idim >= 0; --idim) {
+        if (indices[idim] >= shape_(idim)) {
+          PARTHENON_THROW("Index " + std::to_string(indices[idim]) +
```
Yurlungur: On GPU this should be `PARTHENON_FAIL`. C++ error patterns don't work on device.

Yurlungur (changes requested):
I don't think this will work on GPU as written... which is probably an indication we should write a unit test too. See my comments below as to how to fix this for GPU.

src/interface/variable_pack.hpp
lroberts36: I was worried about that... 

My guess is that `FlatIdx` won't be too widely needed. It is just useful for radiation moments that have vector indices, species indices and maybe an energy index if we were to ever do multi-group (and maybe a pressure tensor). Therefore, I would say that the deep copy should be done in the FlatIdx constructor. That being said, I have limited experience with both Parthenon and Phoebus so I really don't know how many other use cases there are likely to be.

src/interface/variable_pack.hpp
lroberts36: I am not thinking that we want to have this error checking in there if it is going to be repeatedly called on the GPU. What do you think?

Yurlungur: In that case use `PARTHENON_DEBUG_FAIL`... that way it'll be hidden behind preprocessor macros and will be absent in production builds.

src/interface/variable_pack.hpp
Yurlungur: Doing it in the constructor works for me. We can modify it later if it's a performance issue.

src/interface/variable_pack.hpp
```diff
@@ -49,6 +49,54 @@ using SwarmVarList = std::forward_list<std::shared_ptr<ParticleVariable<T>>>;
...
+  int DimSize(int i) const { return shape_and_offset_(i); }
+
+  template <typename... Ts>
+  KOKKOS_INLINE_FUNCTION int operator()(Ts... idx_pack) const {
+    const int size = sizeof...(idx_pack);
+    int indices[size] = {idx_pack...};
+
+    // Check that the correct dimensionality is being specified
+    if (size != shape_and_offset_.size() - 1) {
+      PARTHENON_FAIL("Wrong number of indices for variable.");
```
brryan: This could be condensed as `PARTHENON_REQUIRE(size == shape_and_offset_.size() - 1, "Wrong number of indices for variable.");`

src/interface/variable_pack.hpp
```diff
@@ -49,6 +49,54 @@ using SwarmVarList = std::forward_list<std::shared_ptr<ParticleVariable<T>>>;
...
+        Kokkos::create_mirror_view(Kokkos::HostSpace(), shape_and_offset_);
+    for (int i = 0; i < shape.size(); ++i) {
+      host_shape_and_offset(i) = shape[i];
+    }
+    host_shape_and_offset(shape.size()) = offset;
+    Kokkos::deep_copy(shape_and_offset_, host_shape_and_offset);
+  }
+
+  KOKKOS_INLINE_FUNCTION
+  int DimSize(int i) const { return shape_and_offset_(i); }
```
brryan: What about adding a `PARTHENON_DEBUG_REQUIRE(i < shape_and_offset_.size() - 1)` here? I just think a bit of extra caution might be warranted for this pattern of an array storing an array plus a scalar. Could help with code readability too -- it took me a minute to realize what `host_shape_and_offset` was storing in each index

src/interface/variable_pack.hpp
lroberts36: Since the maximum rank we will have to deal with is 3, I decided to replace the `ParArray1D<int>` with a fixed length `C` array inside `FlatIdx` to keep track of the shape so that instances of `FlatIdx` can be captured by the Kokkos lambdas during parallel dispatch. This seems to work fine and be in line with how indexing is being done in other places in the code and keeps things simple.

src/interface/variable_pack.hpp
lroberts36: Switched.

brryan (approved):
Only two minor comments. This is a welcome feature and I approve.

src/interface/variable_pack.hpp
Yurlungur: Yep that should work. :+1:

src/interface/variable_pack.hpp
Yurlungur: Should be a `PARTHENON_DEBUG_REQUIRE` probably

lroberts36: Switched.

src/interface/variable_pack.hpp
lroberts36: This admittedly confusing pattern is now gone after switching away from ParArray1D for the internal storage. I still added the check on the requested dimension though.

lroberts36:
Ok, I think this should now be working and passing the unit test I added. Removing WIP. Please take a look @Yurlungur.

Yurlungur (approved):
Looks good to me. Thanks, @lroberts36 !

PR #602, opened 2021-10-11
Added tuning functionality for HDF5 output 

pgrete:
## PR Summary

PR for testing. Discussion/review see #597 

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted


pgrete:
@par-hermes format

Yurlungur (approved):

Yurlungur:
looks like style check failed. @pgrete should we handle this? Or should we ask @brtnfld to fix the style? Should be fast to fix I think

pgrete:
I went down the `type` rabbit hole as I suspected some potential issues with the lifetime of `const char*`.
Now separate string objects are used and passed with `c_str()` to the MPI function, which should guarantee they exist and don't contain garbage info when setting the parameters.
At the same time I also added some additional checks for boolean and `(h)size_t` environment parameters.
It'd be great if you,@Yurlungur and @brtnfld , could briefly review the last commit I made before we merge.

src/utils/utils.hpp
```diff
@@ -50,6 +53,80 @@ void Report();
...
+//! Env
+//  \brief static function to check and retrieve environment settings
+namespace Env {
+
+// TODO(the person bumping standard to C++17) Clean up this mess and use constexpr if
+template <typename T>
+T parse_value(std::string &strvalue);
+
+// Parse env. variable expected to hold a boolean value allowing for different conventions
+template <>
```
Yurlungur, line 68: Maybe add a comment that `parse_value` modifies the input string

Yurlungur (approved):
Your modifications seem fine to me, @pgrete . I approve. Maybe one thing is that the parse value methods in `Env` should be private? Or should be hidden in an `Impl` namespace?

src/utils/utils.hpp
pgrete: Both addressed.

pgrete:
Technically, we need a second approval here. I cannot approve because I opened the PR, ping @jlippuner @jdolence @forrestglines

brtnfld:
> I went down the `type` rabbit hole as I suspected some potential issues with the lifetime of `const char*`. Now separate string objects are used and passed with `c_str()` to the MPI function, which should guarantee they exist and don't contain garbage info when setting the parameters. At the same time I also added some additional checks for boolean and `(h)size_t` environment parameters. It'd be great if you,@Yurlungur and @brtnfld , could briefly review the last commit I made before we merge.

Looks good to me.

jlippuner (approved):
This looks good. Although I'm not sure why we're using environment variables here instead of the input file. Wouldn't it make more sense to have all these parameters specified in the input file like all the other Parthenon options? I guess downside is that we couldn't share input file between machines that need different tuning settings...

pgrete:
Thanks! I'll go ahead an squash and merge then.

> This looks good. Although I'm not sure why we're using environment variables here instead of the input file. Wouldn't it make more sense to have all these parameters specified in the input file like all the other Parthenon options? I guess downside is that we couldn't share input file between machines that need different tuning settings...

With respect to the parameters, I actually think that these type of parameters are better suited for environment parameter (potentially included in a machine file as info or similar).
My reasoning is that those are independent from the simulation content itself and are tied to the compute "environment". Similarly, many other options and parameters set by scheduling/batch systems are environment variables so this seems like a good fit.

PR #604, opened 2021-10-14
Allow modification of SimTime in PreStepUserWorkInLoop

### LINKED ISSUE #601: RFC: Allow modification of `dt` in `PreStepUserWorkInLoop`
`PreStepUserWorkInLoop` is currently a `std::function<void(Mesh *, ParameterInput *, SimTime const &)>`.
Any objection to remove the `const` from the `SimTime` parameter?

The rationale is that I'm currently working on a super timestepping algorithm for parabolic terms where there's a practical "correctness" limit on the super timestep (despite not having a stability limit for the super timestep).
In order to calculate the parabolic timestep I modified our `EstimateTimestep` function that calculate all timesteps but does not return the parabolic timestep as minimum but instead saves it in the package `Params`.
In that function only the rank local blocks are handled and the overall MPI reduction for `dt` occurs in Parthenon only after `Step()` in the driver. 
However, that only applies to the single `dt` returned from that function so I added another reduction for `dt_par` in `PreStepUserWorkInLoop`.
Given the practical constraint on `dt_par` I want to have the opportunity to limit the actual `dt`, which is currently not possible because `SimTime` is `const`.

I'm also open to other suggestions in terms of implementation, but on first sight it looks like the cleanest and generally most flexible solution to me without additional interfaces/machinery in Parthenon.

Thoughts from downstream developers? Ping @Yurlungur @jdolence @brryan

pgrete:
## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added (new features/APIs/variables/...)
 
 ### Changed (changing behavior/API/variables/...)
+- [[PR 604]](https://github.com/lanl/parthenon/pull/604) Allow modification of SimTime in PreStepUserWorkInLoop
 
 ### Fixed (not changing behavior/API/variables/...)
 - [[PR 595]](https://github.com/lanl/parthenon/pull/595) Fix build options so that non-MPI builds cannot be paired with an MPI HDF5 lib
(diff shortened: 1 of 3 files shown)


pgrete:
Should be a quick review for a +4/-3 total change.

Yurlungur (approved):
LGTM. I did a quick grep in riot and phoebus and I don't think this even breaks anything.

jlippuner (approved):
Looks good, although according to the Google style guide, `SimTime` should be a pointer instead of a non-const reference. I would vote for changing that, but I'll leave the decision up to you.

jdolence (approved):

PR #605, opened 2021-10-20
Add output triggering by signaling

### LINKED ISSUE #574: Missing restart output from "time" signal
Using `./bin -i input -t 1:0:0` result in the binary to exit after one hour, but there is no final restart output that could be used to continue the simulation nor is there a final message on the run statistics.

pgrete:
## PR Summary

From the updated docs:
```
In addition to time base outputs, two additional options to trigger outputs
(applies to HDF5 and restart outputs) exist.

- Signaling: If `Parthenon` catches a signal, e.g., `SIGALRM` which is often sent by
schedulers such as Slurm to signal a job of exceeding the job's allocated walltime,
`Parthenon` will gracefully terminate and write output files with a `final` id rather
than a number.
This also applies to the `Parthenon` internal walltime limit, e.g., when executing an
application with the `-t HH:MM:SS` parameter on the command line.
- File trigger: If a user places a file with the name `output_now` in the working
directory of a running application, `Parthenon` will write output files with a `now` id
rather than a number.
After the output is being written the `output_now` file is removed and the simulation
continues normally.
The user can repeat the process any time by creating a new `output_now` file.

Note, in both cases the original numbering of the output will be unaffected and the
`final` and `now` files will be overwritten each time without warning.
```

Closes #574

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Current develop
 
 ### Added (new features/APIs/variables/...)
+- [[PR 605]](https://github.com/lanl/parthenon/pull/605) Add output triggering by signaling.
 
 ### Changed (changing behavior/API/variables/...)
 
--- a/docs/outputs.md
+++ b/docs/outputs.md
@@ -4,6 +4,24 @@ Outputs from Parthenon are controled via `<parthenon/output*>` blocks, where `*`
 
 To disable an output block without removing it from the intput file set the block's `dt < 0.0`.
 
+In addition to time base outputs, two additional options to trigger outputs
+(applies to HDF5 and restart outputs) exist.
+
+- Signaling: If `Parthenon` catches a signal, e.g., `SIGALRM` which is often sent by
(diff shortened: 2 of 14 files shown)


Yurlungur (approved):
This is a great addition---and very useful for real life.

I like the `output_now` trick... although relying on a file to trigger it is pretty clunky. I don't see a better way though.

jlippuner (approved):
Looks great! This is a nice and very useful addition. I only looked at the code and didn't test the new output mechanisms myself. I'm assuming you've tested them already.

pgrete:
> Looks great! This is a nice and very useful addition. I only looked at the code and didn't test the new output mechanisms myself. I'm assuming you've tested them already.

Yes, already using both signaling and `output_now` in practice and it's also integrated in the test suite :smile: 


Enabling auto-merge (squash) to merge after the pipeline passed.

PR #608, opened 2021-10-21
Add Metadata::None variables to PHDF output

### LINKED ISSUE #606: Output does not support variables with reduced rank
@nmsriram and I recently tried to register a variable not tied to grid topology (e.g., with `Metadata::None`) and shape something like `5x1x1x1x1x1`. This *can* be registered. And for the types of things you might want this for, e.g., per-meshblock scratch space of some kind, it works fine. However, the I/O doesn't capture it. Instead, the output shape is always as if it were a cell variable.

Probably related to #600 . Pinging @jlippuner

nmsriram:
Parthenon currently does not handle variables with Metadata::None.  It puts them in the PHDF file with a cell based topology and trash for values.  This PR attempts to fix this problem.

Fixes #606

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


nmsriram:
@Yurlungur For some reason multi-dimensional arrays are not being referenced correctly in macro `LOADVARIABLERAW` in outputs/parthenon_hdf.hpp.  Could you please take a look?

Yurlungur:
> @Yurlungur For some reason multi-dimensional arrays are not being referenced correctly in macro `LOADVARIABLERAW` in outputs/parthenon_hdf.hpp. Could you please take a look?

Sure. What's the symptom of the bad reference? Also do you have an example you've been playing with? e.g., the advection example or something?

nmsriram:
> > @Yurlungur For some reason multi-dimensional arrays are not being referenced correctly in macro `LOADVARIABLERAW` in outputs/parthenon_hdf.hpp. Could you please take a look?
> 
> Sure. What's the symptom of the bad reference? Also do you have an example you've been playing with? e.g., the advection example or something?

Sorry, I got caught up in debugging.  I've generalized this to 6D block arrays and it works correctly with multi-D arrays with latest commit.

Yurlungur:
> > > @Yurlungur For some reason multi-dimensional arrays are not being referenced correctly in macro `LOADVARIABLERAW` in outputs/parthenon_hdf.hpp. Could you please take a look?
> > 
> > 
> > Sure. What's the symptom of the bad reference? Also do you have an example you've been playing with? e.g., the advection example or something?
> 
> Sorry, I got caught up in debugging. I've generalized this to 6D block arrays and it works correctly with multi-D arrays with latest commit.

Haha! Now I don't have to do anything. ;)

nmsriram:
@Yurlungur how do I trigger CI on Darwin?

Yurlungur:
> @Yurlungur how do I trigger CI on Darwin?

Click on the "darwin" hyperlink. Then you can either look for the pipelines that are greyed out and say "paused" and press the play arrow on the right. Or you can click "run pipeline" on the top right and select your branch.

You might have to wait for your most recent commit to show up in the gitlab mirror... sometimes it's kind a slow.

Yurlungur (approved):
Nice fix!

src/outputs/parthenon_hdf5.cpp
```diff
@@ -428,13 +430,12 @@ void PHDF5Output::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm) {
       tmpData[i] = -1.25;
 
     // Write mesh coordinates to file
-    hsize_t local_start[5], global_count[5], local_count[5];
+    hsize_t local_start[7], global_count[7], local_count[7];
```
pgrete, line 433: Why is this change required and how is it connected to block variables?
How does this affect dimension of the other variables?
Also, does this have any implications on the xdmf output and/or our python scripts to read data?

CHANGELOG.md
```diff
@@ -86,6 +86,7 @@ Date: 01/19/2021
 - [[PR 387]](https://github.com/lanl/parthenon/pull/387) Add missing const that was needed
 - [[PR 353]](https://github.com/lanl/parthenon/pull/353) Fixed small error in input\_parameter logic
 - [[PR 352]](https://github.com/lanl/parthenon/pull/352) Code compiles cleanly (no warnings) with nvcc_wrapper
+- [[PR 608]](https://github.com/lanl/parthenon/pull/608) Variables with `Metadata::None` can now be output to PHDF file
```
pgrete, line 89: This should probably go up to the top in the `current develop` section.

src/outputs/parthenon_hdf5.hpp
```diff
@@ -113,6 +113,24 @@ using H5T = H5Handle<&H5Tclose>;
...
 
+#define LOADVARIABLERAW(index, dst, var, i1, i2, i3, i4, i5, i6)                         \
+  do {                                                                                   \
+    for (int i = 0; i < i1; i++) {                                                       \
+      for (int j = 0; j < i2; j++) {                                                     \
+        for (int k = 0; k < i3; k++) {                                                   \
+          for (int l = 0; l < i4; l++) {                                                 \
+            for (int m = 0; m < i5; m++) {                                               \
+              for (int n = 0; n < i6; n++) {                                             \
+                dst[index] = var(i, j, k, l, m, n);                                      \
```
pgrete, line 124: Is this ordering intentional as it somewhat goes against the typical use of loops elsewhere in the code?
If yes, then some explaining comment would be great. If not, then I suggest to rearrange.

src/outputs/parthenon_hdf5.cpp
```diff
@@ -609,6 +613,55 @@ void PHDF5Output::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm) {
...
+    for (auto &vwrite : ciX.vars) { // for each variable we write
+      if (!vwrite->metadata().IsSet(Metadata::None)) {
+        continue;
+      }
+      const std::string vWriteName = vwrite->label();
+      hid_t vLocalSpace, vGlobalSpace;
+      H5S vLocalSpaceNew, vGlobalSpaceNew;
+
+      // Count up dimensions to output
+      int spaceCount, startDim;
```
pgrete: Is `startDim` used?

src/outputs/parthenon_hdf5.cpp
```diff
@@ -609,6 +613,55 @@ void PHDF5Output::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm) {
...
+      spaceCount = 1;
+      for (int idim = 1; idim <= 6; idim++) {
+        if (vwrite->GetDim(idim) > 1) {
```
pgrete, line 630: Are `local_count` and `global_count` properly inititalized for `vwrite->GetDim(idim) == 1`?

pgrete (commented):
Thanks for the update @nmsriram 
I have a couple of question/comments, see below.
On top of that I'd be great to somehow add a test (or probably extend an existing one), just to make sure those changes (including the python scripts etc) work as expected.

jlippuner:
This PR seems to be based on a very old version of `develop`. The HDF5 output has changed substantially since then.

pgrete:
Oh interesting. Looks like that develop dates back to April. I'm actually surprised GitHub didn't tell us (or I missed it).
So the question now is probably if the issue is still present in current develop.

Yurlungur:
> Oh interesting. Looks like that develop dates back to April. I'm actually surprised GitHub didn't tell us (or I missed it). So the question now is probably if the issue is still present in current develop.

I tried this on develop recently when @nmsriram first reported it, and experienced the same issue. Bit I didn't catch the age of the commit in this PR.

jlippuner:
> Oh interesting. Looks like that develop dates back to April. I'm actually surprised GitHub didn't tell us (or I missed it). So the question now is probably if the issue is still present in current develop.

Well, GitHub is showing the conflicts in the `parthenon_hdf5.*` files, which are really the only files touched in this PR.

pgrete:
> > Oh interesting. Looks like that develop dates back to April. I'm actually surprised GitHub didn't tell us (or I missed it). So the question now is probably if the issue is still present in current develop.
> 
> Well, GitHub is showing the conflicts in the `parthenon_hdf5.*` files, which are really the only files touched in this PR.

IIRC those conflicts first showed up after #602 which was merged only after this PR was opened.

jlippuner:
> > > Oh interesting. Looks like that develop dates back to April. I'm actually surprised GitHub didn't tell us (or I missed it). So the question now is probably if the issue is still present in current develop.
> > 
> > 
> > Well, GitHub is showing the conflicts in the `parthenon_hdf5.*` files, which are really the only files touched in this PR.
> 
> IIRC those conflicts first showed up after #602 which was merged only after this PR was opened.

Hmm, that's odd. In any case, this PR needs to be rebased on current develop. It's calling the macro `WRITEH5SLAB2` that was removed back in April.

src/outputs/parthenon_hdf5.cpp
nmsriram: @pgrete sorry for delay - my work on parthenon is sporadic these days.

**Why is this change required and how is it connected to block variables?**
On a downstream code I'm working on I need to store a per-block basis.  Parthenon already allows this to be done via `Metadata::None` flag.  However the HDF output does not allow me to write this variable to the `phdf` file so that I can pick it up post run.  This change is to allow me to output the per-block variable as per-block instead of per-cell which would expand it by 47k times for 32^3 blocks.

**How does this affect dimension of the other variables?**
It doesn't affect any of the other variables at all.  It just so happens that to declare it I have to declare it as the full 6-dimensional array.

**Also, does this have any implications on the xdmf output and/or our python scripts to read data?**
It gets put out in the `XDMF` as a `Grid` type variable (this PR).  In Python it will be available as usual using `h5py`

Please let me know if I've answered your questions.

src/outputs/parthenon_hdf5.cpp
nmsriram: This occurs on line 552 where `local_count[0]` is set to number of blocks.  `spaceCount` is only incremented for non-scalar grid arrays.

src/outputs/parthenon_hdf5.cpp
nmsriram: Good catch - Fixed.

nmsriram:
@jlippuner @Yurlungur @pgrete: Catching up to the bottom of this set of comments.  I'm going to decline this MR and redo the work based on latest develop since that's easier and more likely to be done right than fixing the conflicts.

PR #613, opened 2021-11-09
WIP: getHDF5Type Compiler workaround for g++-10 on OS X

nmsriram:
## PR Summary
Added an Apple specific getType for getHDF5Type with size_t as argument since it is not recognized as uint64_t by g++-10

Enables compiling on OSX with g++-10.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/src/outputs/parthenon_hdf5.hpp
+++ b/src/outputs/parthenon_hdf5.hpp
@@ -102,6 +102,9 @@ static hid_t getHDF5Type(const int32_t *) { return H5T_NATIVE_INT32; }
 static hid_t getHDF5Type(const int64_t *) { return H5T_NATIVE_INT64; }
 static hid_t getHDF5Type(const uint32_t *) { return H5T_NATIVE_UINT32; }
 static hid_t getHDF5Type(const uint64_t *) { return H5T_NATIVE_UINT64; }
+#ifdef __APPLE__
+static hid_t getHDF5Type(const size_t *) { return H5T_NATIVE_UINT64; }
+#endif
 static hid_t getHDF5Type(const float *) { return H5T_NATIVE_FLOAT; }
 static hid_t getHDF5Type(const double *) { return H5T_NATIVE_DOUBLE; }
 static H5T getHDF5Type(const char *const *) {


Yurlungur (approved):
No objections... but it got me thinking... why do we specify all these widths by hand in the first place? Why not just use `size_t` and `H5T_NATIVE_*` for everything?

nmsriram:
> No objections... but it got me thinking... why do we specify all these widths by hand in the first place? Why not just use `size_t` and `H5T_NATIVE_*` for everything?

This allows us to have a HDF read and write from our restart file (see HDF5WriteND for example a few lines down).  Ideally the H5 C++ interface would have provided us with a templated instance of this, but alas that is not to be...

pgrete:
Should we, for compatibility reasons, not just remove that ifdef so that all `size_t` are `unit64_t`?

Yurlungur:
Not sure I follow @pgrete What's the suggestion?

Also @nmsriram what's the status of this PR? Should we merge it?

src/outputs/parthenon_hdf5.hpp
```diff
@@ -102,6 +102,9 @@ static hid_t getHDF5Type(const int32_t *) { return H5T_NATIVE_INT32; }
 static hid_t getHDF5Type(const int64_t *) { return H5T_NATIVE_INT64; }
 static hid_t getHDF5Type(const uint32_t *) { return H5T_NATIVE_UINT32; }
 static hid_t getHDF5Type(const uint64_t *) { return H5T_NATIVE_UINT64; }
+#ifdef __APPLE__
+static hid_t getHDF5Type(const size_t *) { return H5T_NATIVE_UINT64; }
+#endif
```
pgrete, line 107:
```suggestion
static hid_t getHDF5Type(const size_t *) { return H5T_NATIVE_UINT64; }
```

This is my suggestion. Basically, making sure that our `size_t` is the same on all systems from an hdf5 point of view (so no extra switch for OSX).

Yurlungur: Ah I see. For HDF5 I think this is reasonable... Ideally the types on disk *should* be the same for all platforms

pgrete: Actually, this suggestion won't work as it results in two identical definitions on non-Apple arch so we may need to keep this.
However, I'm currently leaning towards using a template approach (based on checking if `size_t` is `uint64_t`) so that this is compiler and architecture independent (though I cannot test due to lack of access to OSX).

pgrete (commented):
This should probably also go into develop rather than sparse, doesn't it?

Yurlungur:
THIS PR points to the `jlippuner/true-sparse` branch which was long ago merged. Should we keep this?

pgrete:
We actually fixed this with template magic ;)
```

// On MacOS size_t is "unsigned long" and uint64_t is != "unsigned long".
// Thus, size_t is not captured by the overload above and needs to selectively enabled.
template <typename T,
          typename std::enable_if<std::is_same<T, unsigned long>::value && // NOLINT
                                      !std::is_same<T, uint64_t>::value,
                                  bool>::type = true>
static hid_t getHDF5Type(const T *) {
  return H5T_NATIVE_ULONG;
}
```

PR #614, opened 2021-11-12
A GetCoordinates dispatch function for kernels templated on pack type

Yurlungur:
## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/docs/interface/containers.md
+++ b/docs/interface/containers.md
@@ -239,8 +239,8 @@ or contain variables and fluxes.
 The coordinate system for a given `MeshBlock` object is contained in
 `MeshBlock.coords`, a public field. As discussed above,
 `MeshBlockPack` objects contain a 1D array of `coords` objects. The
-difference between these two interfaces... a Kokkos view in one case,
-and a single static object in the other, can cause difficulties when
+difference between these two interfaces---a Kokkos view in one case,
+and a single static object in the other---can cause difficulties when
 writing generic kernels. As such, we provide the free-floating
 dispatch function
 
--- a/docs/interface/containers.md
+++ b/docs/interface/containers.md
@@ -233,3 +233,25 @@ MeshBlockVarFluxPack<T>
 ```
 which correspond to packs over `MeshBlock`s that contain just variables
 or contain variables and fluxes.
+
+## Packs and Coordinates
+
+The coordinate system for a given `MeshBlock` object is contained in
+`MeshBlock.coords`, a public field. As discussed above,
+`MeshBlockPack` objects contain a 1D array of `coords` objects. The
+difference between these two interfaces... a Kokkos view in one case,
+and a single static object in the other, can cause difficulties when
+writing generic kernels. As such, we provide the free-floating
+dispatch function
+
+```C++
+template<typename Data, typename Pack>
+ParArray1D<Coordinates_t> parthenon::GetCoordinates(Data *rc, Pack &p);
+```
+
+where the first argument is intended to be a `MeshData` or
+`MeshBlockData` object, and the second a `VariablePack` or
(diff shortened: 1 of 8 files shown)


Yurlungur:
Oops forgot to write what this is above:

## PR description

I got tired of jumping through hoops to access `Coordinates_t` in generic templated functions that are intended to operate on either `MeshData` or `MeshBlockData`, so I wrote a function to jump through those hoops for me. I chose this approach over carrying the coordinates object into the `VariablePack` to reduce the amount of data we're copying around and to keep the `VariablePack`s lightweight.

An alternative would be to put `coords` into the `VariablePack` and redesign `MeshBlockPack` to expose it, rather than carry around its own array of the objects. I'd be fine either way, but I chose this method as a first pass, since it seems easy and lightweight.

Pinging folks who I think would be interested, as well as downstream developers who may have run into this. @jdolence @brryan @lroberts36 @pgrete @forrestglines

Yurlungur:
@par-hermes format

Yurlungur:
@par_hermes format

Yurlungur:
@par-hermes format

Yurlungur:
@par-hermes format

Yurlungur:
@pgrete @jdolence @lroberts36 @forrestglines any opinions on this?

docs/interface/containers.md
```diff
@@ -233,3 +233,25 @@ MeshBlockVarFluxPack<T>
...
 which correspond to packs over `MeshBlock`s that contain just variables
 or contain variables and fluxes.
+
+## Packs and Coordinates
+
+The coordinate system for a given `MeshBlock` object is contained in
+`MeshBlock.coords`, a public field. As discussed above,
+`MeshBlockPack` objects contain a 1D array of `coords` objects. The
+difference between these two interfaces... a Kokkos view in one case,
+and a single static object in the other, can cause difficulties when
```
lroberts36: Maybe replace `... a Kokkos view in one case, and a single static object in the other,` with `-- a Kokkos view in one case, and a single static object in the other --`. The meaning of the ellipsis was confusing to me when I first read it.

src/coordinates/get_coordinates.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+namespace parthenon {
+
+template <>
+ParArray1D<Coordinates_t> GetCoordinates(MeshBlockData<Real> *rc, VariablePack<Real> &p) {
+  ParArray1D<Coordinates_t> coords("Parthenon::Coordinates_t array", 1);
+  auto pmb = rc->GetParentPointer();
+  auto pmb_coords = pmb->coords;
+  pmb->par_for(
+      "fill coords view", 0, 0, KOKKOS_LAMBDA(const int i) { coords(0) = pmb_coords; });
```
lroberts36, line 28: Is it likely to be a performance issue that there is a kernel launch every time this function is called?

lroberts36 (commented):
This looks like a reasonable approach to me. Another approach might be to add `ParArray1D<Coordinates_t>& coord_array = coords` to `MeshBlockPack` and `ParArray1D<Coordinates_t> coord_array` to `MeshBlock`.

src/coordinates/get_coordinates.cpp
Yurlungur: Ah, good question... I'm not sure. The solution of adding a kokkos view to `VariablePack`s would fix that, but incur other overhead. I don't know which solution is more performant. Maybe worth trying them both out.

Yurlungur:
Obviously only one of this or #617 should be merged.

Yurlungur:
I am now in favor of #617 over this PR, so I am closing this one.

PR #617, opened 2021-11-20
Introduce coords object to variable packs

Yurlungur:
## PR Summary

This is an alternative to #614 . I am sick of trying to write kernels that look different for `MeshData` vs `MeshBlockData`. This helps unify the two by putting a single method `GetCoords` into both variable packs and meshblock packs. The results are the following:
- `VariablePack` has a coords object in it, which is a bit more register pressure.
- Constructing `MeshBlockPack` no longer requires a deep copy of coords data to device. It's simply already in the `VariablePack` objects.
- I also remove the `IndexShape` object from meshblock packs, again, to make things more similar. Alternatively, both classes should get an `IndexShape` accessor. I'm open to adding that.
- The result is the API changed slightly downstream for these methods. So this is a breaking change.

Pinging @jdolence @forrestglines @pgrete @lroberts36 @brryan and other interested downstream developers. What's the right approach here? Should we use this approach? Or the one in #614? In #614 @lroberts36 raised an interesting performance question. 

I think both methods are defensible. I like this one better, but it also implies deeper code surgery. What do you all think?

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ Date: 09/22/2021
 - [[PR 404]](https://github.com/lanl/parthenon/pull/404) Add capability to communicate particles across meshblocks/MPI processes
 
 ### Changed (changing behavior/API/variables/...)
+- [[PR 617]](https://github.com/lanl/parthenon/pull/617) Unify the API for MeshBlockPack and VariablePack
 - [[PR 558]](https://github.com/lanl/parthenon/pull/558) Boundary bugfix(es) incl. regression tests and exposing FluxDiv_ interface
 - [[PR 583]](https://github.com/lanl/parthenon/pull/583) Fix file numbering logic for writing outputs after restarting a simulation.
 - [[PR 581]](https://github.com/lanl/parthenon/pull/581) Change return status of `ArgParse` so that `complete` is returned when passing in the help flag.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - [[PR 602]](https://github.com/lanl/parthenon/pull/602) Added tuning functionality for HDF5 output
 
 ### Changed (changing behavior/API/variables/...)
+- [[PR 617]](https://github.com/lanl/parthenon/pull/617) Unify the coordinates API for MeshBlockPack and VariablePack
 
 ### Fixed (not changing behavior/API/variables/...)
 - [[PR 616]](https://github.com/lanl/parthenon/pull/609) Restore sparse base names in PackIndexMap
@@ -36,7 +37,6 @@ Date: 09/22/2021
 - [[PR 404]](https://github.com/lanl/parthenon/pull/404) Add capability to communicate particles across meshblocks/MPI processes
 
 ### Changed (changing behavior/API/variables/...)
-- [[PR 617]](https://github.com/lanl/parthenon/pull/617) Unify the coordinates API for MeshBlockPack and VariablePack
 - [[PR 558]](https://github.com/lanl/parthenon/pull/558) Boundary bugfix(es) incl. regression tests and exposing FluxDiv_ interface
 - [[PR 583]](https://github.com/lanl/parthenon/pull/583) Fix file numbering logic for writing outputs after restarting a simulation.
 - [[PR 581]](https://github.com/lanl/parthenon/pull/581) Change return status of `ArgParse` so that `complete` is returned when passing in the help flag.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,7 +36,7 @@ Date: 09/22/2021
 - [[PR 404]](https://github.com/lanl/parthenon/pull/404) Add capability to communicate particles across meshblocks/MPI processes
 
 ### Changed (changing behavior/API/variables/...)
-- [[PR 617]](https://github.com/lanl/parthenon/pull/617) Unify the API for MeshBlockPack and VariablePack
+- [[PR 617]](https://github.com/lanl/parthenon/pull/617) Unify the coordinates API for MeshBlockPack and VariablePack
 - [[PR 558]](https://github.com/lanl/parthenon/pull/558) Boundary bugfix(es) incl. regression tests and exposing FluxDiv_ interface
 - [[PR 583]](https://github.com/lanl/parthenon/pull/583) Fix file numbering logic for writing outputs after restarting a simulation.
 - [[PR 581]](https://github.com/lanl/parthenon/pull/581) Change return status of `ArgParse` so that `complete` is returned when passing in the help flag.
--- a/example/advection/advection_package.cpp
+++ b/example/advection/advection_package.cpp
@@ -381,13 +381,13 @@ template <typename T>
 Real AdvectionHst(MeshData<Real> *md) {
   auto pmb = md->GetBlockData(0)->GetBlockPointer();
 
+  const auto ib = pmb->cellbounds.GetBoundsI(IndexDomain::interior);
+  const auto jb = pmb->cellbounds.GetBoundsJ(IndexDomain::interior);
+  const auto kb = pmb->cellbounds.GetBoundsK(IndexDomain::interior);
+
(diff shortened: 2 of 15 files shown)


Yurlungur:
@par-hermes format

Yurlungur:
Obviously only one of this PR and #614 should be merged

c-prather:
I think I have a potential corner-case for any performance difference here.

In KHARMA, I subclass Coordinates to carry around a bunch of extra information for GR, both some heavy variant objects representing coordinate systems and a bunch of references to the geometry cache arrays.  Then I go and use it obsessively in every kernel.

I'm assuming that this branch would be faster for me since it avoids deep-copying Coordinates, but I'd be happy to try both and see, if you think there's potential for a substantive difference.

Yurlungur:
> I think I have a potential corner-case for any performance difference here.
> 
> In KHARMA, I subclass Coordinates to carry around a bunch of extra information for GR, both some heavy variant objects representing coordinate systems and a bunch of references to the geometry cache arrays. Then I go and use it obsessively in every kernel.
> 
> I'm assuming that this branch would be faster for me since it avoids deep-copying Coordinates, but I'd be happy to try both and see, if you think there's potential for a substantive difference.

@brryan made a suggestion that I want to pursue, where a device-side coordinate object is created per meshblock, and then the variable packs use that. Would that avoid the issue for you, @bprather ?

c-prather:
That would avoid all H->D copies of Coordinates objects entirely, right?  That would be ideal!

If this would involve running the Coordinates() constructor device-side I think I may have some porting work in front of me, but if I'm understanding correctly I'm happy with it as a way forward.

Yurlungur:
> That would avoid all H->D copies of Coordinates objects entirely, right? That would be ideal!
> 
> If this would involve running the Coordinates() constructor device-side I think I may have some porting work in front of me, but if I'm understanding correctly I'm happy with it as a way forward.

The way I was imagining it, it would involve one H->D copy per meshblock at meshblock creation. Call the constructor host-side and then deep-copy the coords object over to device. Would that cause problems?

c-prather:
Oh! Even easier.  I can't imagine how that would cause problems, no.

Yurlungur:
I have now implemented @brryan 's suggestion. Here's the way it works now:
- `MeshBlocks` own a `coords` object that lives on host and a `Kokkos::View<Coordinates_t>` object that lives on device
- When `VariablePack` is created, it gets a copy of the device-view, which is basically just a pointer to the device-side coords object
- Both `VariablePack` and `MeshBlockPack` support a `GetCoords(int i)` function. In the case of `VariablePack` the integer argument is discarded. In the case of `MeshBlockPack`, the integer is used to pass through to the `i`th `VariablePack` object and get its coords pointer.

I'm actually quite happy with how this came out. Thanks @brryan for the suggestion! @bprather @jdolence @lroberts36 @pgrete @forrestglines I'd like some eyes on this. What do you think?

Yurlungur:
> I also remove the IndexShape object from meshblock packs, again, to make things more similar. Alternatively, both classes should get an IndexShape accessor. I'm open to adding that.

I just noticed that `MeshData` and `MeshBlockData` expose `GetBounds*()` methods, so no reason to put that machinery in the pack.

CHANGELOG.md
```diff
@@ -36,6 +36,7 @@ Date: 09/22/2021
 - [[PR 404]](https://github.com/lanl/parthenon/pull/404) Add capability to communicate particles across meshblocks/MPI processes
 
 ### Changed (changing behavior/API/variables/...)
+- [[PR 617]](https://github.com/lanl/parthenon/pull/617) Unify the API for MeshBlockPack and VariablePack
```
```suggestion
- [[PR 617]](https://github.com/lanl/parthenon/pull/617) Unify the coordinates API for MeshBlockPack and VariablePack
```

pgrete: Also, this should be up under the `develop` heading.

example/advection/advection_package.cpp
```diff
@@ -381,14 +381,14 @@ template <typename T>
 Real AdvectionHst(MeshData<Real> *md) {
   auto pmb = md->GetBlockData(0)->GetBlockPointer();
 
+  const auto ib = pmb->cellbounds.GetBoundsI(IndexDomain::interior);
+  const auto jb = pmb->cellbounds.GetBoundsJ(IndexDomain::interior);
+  const auto kb = pmb->cellbounds.GetBoundsK(IndexDomain::interior);
```
pgrete, line 386: (Independent of this PR) this change in lines makes we wonder if `cellbounds` should not be at an even more central place (given that they are fixed "globally" for any given simulation).

example/calculate_pi/calculate_pi.cpp
```diff
@@ -168,6 +171,9 @@ TaskStatus ComputeArea(std::shared_ptr<MeshData<Real>> &md, ParArrayHost<Real> a
   bool const use_sparse =
       md->GetMeshPointer()->packages.Get("calculate_pi")->Param<bool>("use_sparse");
 
+  auto pmb = md->GetBlockData(0)->GetBlockPointer();
+  auto cellbounds = pmb->cellbounds;
```
pgrete:
```suggestion
  const auto &cellbounds = pmb->cellbounds;
```

pgrete (approved):
I like this design!
The performance implication @lroberts36 mentioned didn't make it to the current/final implementation (or did I miss it)?

CHANGELOG.md
Yurlungur: :+1: oops how'd I miss that

example/advection/advection_package.cpp
Yurlungur: Actually after this PR I noticed it already exists: `MeshData` and `MeshBlockData` both provide `GetBounds*`

Yurlungur:
> I like this design! The performance implication @lroberts36 mentioned didn't make it to the current/final implementation (or did I miss it)?

Thanks, @pgrete ! As soon as gitlab CI passes, I'll merge. The issue @lroberts36 found was referring to #614. I submitted two PRs, with two different designs to try and figure out the better design. I liked this one better so I closed #614.

Yurlungur:
Ah actually can't merge. Need one more pair of eyes. @bprather @brryan @lroberts36 since you've all looked at this, can one of you approve?

lroberts36 (approved):
Looks good to me

pgrete:
Looks like sth is wrong (full test suite):
```
-------------------------------------------------------------------------------
/builds/pgrete/parthenon/tst/performance/test_meshblock_data_iterator.cpp:241
...............................................................................
/builds/pgrete/parthenon/tst/performance/test_meshblock_data_iterator.cpp:241: FAILED:
due to unexpected exception with message:
  ### PARTHENON ERROR
    Message:     Invalid pointer to MeshBlock!
    File:        /builds/pgrete/parthenon/src/interface/meshblock_data.hpp
    Line number: 82
```
https://gitlab.com/pgrete/parthenon/-/pipelines/419557136

Yurlungur:
Oof. I'll fix it.

Yurlungur:
This is weird. It works fine on my desktop, and also the MPI version passes.

Yurlungur:
Also no function in `test_meshblock_data_iterator.cpp` calls `GetParentPointer()`.

pgrete:
I now retriggered the test suite. It also failed on the host with the identical error. Not sure what's going on.

pgrete:
PS: I can reproduce locally with `cmake -G Ninja -DCMAKE_BUILD_TYPE:STRING=Release -DKokkos_ARCH_SKX=ON -DPARTHENON_DISABLE_MPI=ON -DPARTHENON_DISABLE_HDF5=ON ..` and `ctest -R performance_tests` but will first get to debugging tomorrow if you cannot reproduce.

Yurlungur:
Got it to reproduce. I was stupidly running the unit instead of performance tests. I think I know what's going on. Will pursue it now.

Yurlungur:
@pgrete fixed in PR #618. Take a look.

PR #619, opened 2021-12-07
Sort particles by cell

brryan:
## PR Summary

We often want to be able to quickly loop through all the particles in a particular cell. By default, particles live in per-meshblock lists and these will not in general be ordered by zone. This PR introduces a method `SwarmContainer::SortParticlesByCell` that uses a 1D sort operation to populate a map from cell index `i,j,k` to a list of per-meshblock indices of all particles in that cell.

The `sort` routine is provided separately and uses separate implementations for CPU and GPU. For the GPU, it relies on `thrust` (Kokkos provides a `sort` routine but from some open issues on that repo it appears to be slow, especially on the GPU). 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Add API for accessing this sorted list from outside `Swarm`, and demonstrate usage in the `particles` example.

### DIFF
--- a/example/particles/particles.cpp
+++ b/example/particles/particles.cpp
@@ -20,71 +20,6 @@
 #include <utility>
 #include <vector>
 
-// TODO(BRR) remove!
-//#include <Kokkos_Sort.hpp>
-/*struct SwarmKey {
-  KOKKOS_INLINE_FUNCTION
-  SwarmKey() {}
-  KOKKOS_INLINE_FUNCTION
-  SwarmKey(const int cell_idx_1d, const int swarm_idx_1d) :
-    cell_idx_1d_(cell_idx_1d), swarm_idx_(swarm_idx_1d) {}
-
-  int cell_idx_1d_;
-  int swarm_idx_;
-};
-struct SwarmKeyCompare {
-  KOKKOS_INLINE_FUNCTION
-  bool operator() (const SwarmKey& s1, SwarmKey& s2) {
-    return s1.cell_idx_1d_ < s2.cell_idx_1d_;
-  }
-};
-
-KOKKOS_INLINE_FUNCTION
-bool operator< (const SwarmKey s1, const SwarmKey s2) {
-  return s1.cell_idx_1d_ < s2.cell_idx_1d_;
-}
-KOKKOS_INLINE_FUNCTION
-bool operator> (const SwarmKey s1, const SwarmKey s2) {
-  return s1.cell_idx_1d_ > s2.cell_idx_1d_;
-}
-KOKKOS_INLINE_FUNCTION
-bool operator== (const SwarmKey s1, const SwarmKey s2) {
-  return s1.cell_idx_1d_ == s2.cell_idx_1d_;
-}
-KOKKOS_INLINE_FUNCTION
-int operator- (const SwarmKey s1, const SwarmKey s2) {
-  return s1.cell_idx_1d_ - s2.cell_idx_1d_;
-}
-namespace Kokkos {
-template<>
-struct reduction_identity<SwarmKey> {
-  KOKKOS_FORCEINLINE_FUNCTION constexpr static int sum() {
-    return static_cast<int>(0);
-  }
-  KOKKOS_FORCEINLINE_FUNCTION constexpr static int prod() {
-    return static_cast<int>(1);
-  }
-  KOKKOS_FORCEINLINE_FUNCTION constexpr static int min() {
-    return INT_MIN;
-  }
-  KOKKOS_FORCEINLINE_FUNCTION constexpr static int max() {
-    return INT_MAX;
-  }
-  KOKKOS_FORCEINLINE_FUNCTION constexpr static int bor() {
-    return static_cast<int>(0x0);
-  }
-  KOKKOS_FORCEINLINE_FUNCTION constexpr static int band() {
-    return ~static_cast<int>(0x0);
-  }
-  KOKKOS_FORCEINLINE_FUNCTION constexpr static int lor() {
-    return static_cast<int>(0);
-  }
-  KOKKOS_FORCEINLINE_FUNCTION constexpr static int land() {
-    return static_cast<int>(1);
-  }
-};
-}*/
-
 // *************************************************//
 // redefine some internal parthenon functions      *//
 // *************************************************//
@@ -235,34 +170,6 @@ TaskStatus DepositParticles(MeshBlock *pmb) {
       });
 
   const int ndim = pmb->pmy_mesh->ndim;
-
-  /*// Practice a sort
-  using KeyType = SwarmKey;
-  //using KeyType = int;
-  using KeyViewType = Kokkos::View<KeyType*, DevExecSpace>;
-  int nkeys = 10;
-  KeyViewType keys("keys", nkeys);
-  pmb->par_for("Set Keys", 0, nkeys-1, KOKKOS_LAMBDA(const int i) {
-    int cell_idx_1d = nkeys - i;
-    if (i == 4) {
-      cell_idx_1d = 5;
-    }
-    keys(i) = SwarmKey(cell_idx_1d, i);
-    //keys(i) = nkeys - i;
-    //if (i == 4) keys(i) = 5;
-  });
-  pmb->par_for("Print Keys", 0, nkeys-1, KOKKOS_LAMBDA(const int i) {
-    printf("key(%i) = %i\n", keys(i).swarm_idx_, keys(i).cell_idx_1d_);
-  });
-  printf("\n");
-  std::sort(keys.data(), keys.data() + keys.extent(0), SwarmKeyCompare());
-  //Kokkos::sort(keys);
-  pmb->par_for("Print Keys", 0, nkeys-1, KOKKOS_LAMBDA(const int i) {
-    //printf("key(%i) = %i\n", i, keys(i));
-    printf("key(%i) = %i\n", keys(i).swarm_idx_, keys(i).cell_idx_1d_);
-  });
-  printf("\n");
-  exit(-1);*/
 
   pmb->par_for(
       "DepositParticles", 0, swarm->GetMaxActiveIndex(), KOKKOS_LAMBDA(const int n) {
@@ -688,11 +595,10 @@ TaskCollection ParticleDriver::MakeFinalizationTaskCollection() const {
 
     auto destroy_some_particles = tl.AddTask(none, DestroySomeParticles, pmb.get());
 
-    auto sort_particles =
-        tl.AddTask(destroy_some_particles, &SwarmContainer::SortParticlesByCell, sc.get());
+    auto sort_particles = tl.AddTask(destroy_some_particles,
+                                     &SwarmContainer::SortParticlesByCell, sc.get());
 
-    auto deposit_particles =
-        tl.AddTask(sort_particles, DepositParticles, pmb.get());
+    auto deposit_particles = tl.AddTask(sort_particles, DepositParticles, pmb.get());
 
     // Defragment if swarm memory pool occupancy is 90%
     auto defrag = tl.AddTask(deposit_particles, &SwarmContainer::Defrag, sc.get(), 0.9);
--- a/example/particles/particles.cpp
+++ b/example/particles/particles.cpp
@@ -61,7 +61,8 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
       destroy_particles_frac >= 0. && destroy_particles_frac <= 1.,
       "Fraction of particles to destroy each timestep must be between 0 and 1");
 
-  std::string deposition_method = pin->GetOrAddString("Particles", "deposition_method", "per_particle");
+  std::string deposition_method =
+      pin->GetOrAddString("Particles", "deposition_method", "per_particle");
   pkg->AddParam<>("deposition_method", deposition_method);
 
   bool orbiting_particles =
@@ -143,15 +144,11 @@ TaskStatus DestroySomeParticles(MeshBlock *pmb) {
 }
 
 TaskStatus SortParticlesIfNecessary(MeshBlock *pmb) {
-    printf("%s:%i\n", __FILE__, __LINE__);
   auto pkg = pmb->packages.Get("particles_package");
   const auto deposition_method = pkg->Param<std::string>("deposition_method");
   if (deposition_method == "per_cell") {
     auto swarm = pmb->swarm_data.Get()->Get("my particles");
-    printf("%s:%i\n", __FILE__, __LINE__);
     swarm->SortParticlesByCell();
-    printf("sorted!\n");
-    printf("%s:%i\n", __FILE__, __LINE__);
   }
 
   return TaskStatus::complete;
@@ -186,7 +183,6 @@ TaskStatus DepositParticles(MeshBlock *pmb) {
   const int ndim = pmb->pmy_mesh->ndim;
 
   if (deposition_method == "per_particle") {
-    printf("%s:%i\n", __FILE__, __LINE__);
     // Reset particle count
     pmb->par_for(
         "ZeroParticleDep", kb.s, kb.e, jb.s, jb.e, ib.s, ib.e,
@@ -214,20 +210,15 @@ TaskStatus DepositParticles(MeshBlock *pmb) {
           }
         });
   } else if (deposition_method == "per_cell") {
-    printf("%s:%i\n", __FILE__, __LINE__);
-    pmb->par_for("DepositParticlesByCell", kb.s, kb.e, jb.s, jb.e, ib.s, ib.e,
-      KOKKOS_LAMBDA(const int k, const int j, const int i) {
-        particle_dep(k, j, i) = 0.;
-        printf("k j i: %i %i %i\n", k, j, i);
-        printf(" count: %i\n", swarm_d.GetParticleCountPerCell(k, j, i));
-        for (int n = 0; n < swarm_d.GetParticleCountPerCell(k, j, i); n++) {
-          printf("  n: %i\n", n);
-          const int idx = swarm_d.GetFullIndex(k, j, i, n);
-          printf("   idx: %i\n", idx);
-          particle_dep(k, j, i) += weight(idx);
-        }
-      });
-    printf("%s:%i\n", __FILE__, __LINE__);
+    pmb->par_for(
+        "DepositParticlesByCell", kb.s, kb.e, jb.s, jb.e, ib.s, ib.e,
+        KOKKOS_LAMBDA(const int k, const int j, const int i) {
+          particle_dep(k, j, i) = 0.;
+          for (int n = 0; n < swarm_d.GetParticleCountPerCell(k, j, i); n++) {
+            const int idx = swarm_d.GetFullIndex(k, j, i, n);
+            particle_dep(k, j, i) += weight(idx);
+          }
+        });
   } else {
     PARTHENON_FAIL("deposition_method not recognized!");
   }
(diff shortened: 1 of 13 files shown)


src/interface/swarm.cpp
```diff
@@ -472,6 +476,111 @@ void Swarm::Defrag() {
...
+
+  // Update per-cell arrays for easier accessing later
+  const IndexRange &ib = pmb->cellbounds.GetBoundsI(IndexDomain::entire);
+  const IndexRange &jb = pmb->cellbounds.GetBoundsJ(IndexDomain::entire);
+  const IndexRange &kb = pmb->cellbounds.GetBoundsK(IndexDomain::entire);
+  pmb->par_for(
+      "Update per-cell arrays", kb.s, kb.e, jb.s, jb.e, ib.s, ib.e,
+      KOKKOS_LAMBDA(const int k, const int j, const int i) {
+        int cell_idx_1d = i + nx1 * (j + nx2 * k);
+        // Find starting index, first by guessing
```
brryan, line 536: This isn't necessarily a very efficient approach, of first guessing where each cell's particles will be in the 1D list, and then marching in O(n) fashion up or down the list until we find the starting particle in that cell, but I couldn't immediately think of a better way to do this and at least it's parallelized on device.

src/utils/sort.hpp
```diff
@@ -0,0 +1,73 @@
...
+void sort(ParArray1D<Key> data, KeyComparator comparator, size_t min_idx,
+          size_t max_idx) {
+  PARTHENON_DEBUG_REQUIRE(min_idx >= 0 && min_idx < data.extent(0),
+                          "Invalid minimum sort index!");
+  PARTHENON_DEBUG_REQUIRE(max_idx >= 0 && max_idx < data.extent(0),
+                          "Invalid maximum sort index!");
+#ifdef KOKKOS_ENABLE_CUDA
+  thrust::device_ptr<Key> first_d = thrust::device_pointer_cast(data.data()) + min_idx;
+  thrust::device_ptr<Key> last_d = thrust::device_pointer_cast(data.data()) + max_idx + 1;
+  thrust::sort(first_d, last_d, comparator);
```
brryan, line 44: As far as I know we don't use `thrust` anywhere else in parthenon, but my understanding is this should always be available when CUDA is available. Still, calling this out explicitly in case anyone has concerns.

example/particles/particles.cpp
```diff
@@ -61,6 +61,10 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
       destroy_particles_frac >= 0. && destroy_particles_frac <= 1.,
       "Fraction of particles to destroy each timestep must be between 0 and 1");
 
+  std::string deposition_method =
+      pin->GetOrAddString("Particles", "deposition_method", "per_particle");
+  pkg->AddParam<>("deposition_method", deposition_method);
+
```
pgrete: I suggest to use an enum class based rather than a string based internal handle?

example/particles/particles.cpp
```diff
@@ -162,34 +180,48 @@ TaskStatus DepositParticles(MeshBlock *pmb) {
...
+  } else if (deposition_method == "per_cell") {
+    pmb->par_for(
+        "DepositParticlesByCell", kb.s, kb.e, jb.s, jb.e, ib.s, ib.e,
+        KOKKOS_LAMBDA(const int k, const int j, const int i) {
+          particle_dep(k, j, i) = 0.;
+          for (int n = 0; n < swarm_d.GetParticleCountPerCell(k, j, i); n++) {
+            const int idx = swarm_d.GetFullIndex(k, j, i, n);
+            particle_dep(k, j, i) += weight(idx);
+          }
+        });
```
pgrete, line 229: Maybe add a comment that this is purely meant as an example of the machinery and that it's not recommended to use this approach to deposit particles in practice from a performance point of view?

example/particles/particles.cpp
```diff
@@ -595,8 +627,10 @@ TaskCollection ParticleDriver::MakeFinalizationTaskCollection() const {
 
     auto destroy_some_particles = tl.AddTask(none, DestroySomeParticles, pmb.get());
 
-    auto deposit_particles =
-        tl.AddTask(destroy_some_particles, DepositParticles, pmb.get());
+    auto sort_particles =
+        tl.AddTask(destroy_some_particles, SortParticlesIfNecessary, pmb.get());
```
pgrete: What defines "is necessary" here? I'm not sure I follow the intention of this naming (in the context of the example).

src/interface/swarm_device_context.hpp
```diff
@@ -14,9 +14,28 @@
...
 namespace parthenon {
 
+struct SwarmKey {
+  KOKKOS_INLINE_FUNCTION
+  SwarmKey() {}
+  KOKKOS_INLINE_FUNCTION
+  SwarmKey(const int cell_idx_1d, const int swarm_idx_1d)
+      : cell_idx_1d_(cell_idx_1d), swarm_idx_(swarm_idx_1d) {}
+
+  int cell_idx_1d_;
```
pgrete, line 28: Thinking long run here, I suggest to add either a safety count that nx*ny*nz < max int, or increase to int64_t.

src/interface/swarm_container.cpp
```diff
@@ -80,6 +80,7 @@ TaskStatus SwarmContainer::Defrag(double min_occupancy) {
                            "Max fractional occupancy of swarm must be >= 0 and <= 1");
 
   for (auto &s : swarmVector_) {
+    // TODO(BRR) Why is this called?
```
pgrete: Good question :smile:

src/utils/sort.hpp
pgrete: What happens on Intel and AMD GPUs? The `std::sort` probably won't work without additional magic, does it?
What's the issue with Kokkos sort? You mentioned performance but my quick search in the Kokkos wiki didn't bring up any helpful information.

src/interface/swarm.cpp
pgrete: What does the literature say about this kind of algorithm? I'd expect that there's previous work on how to solve this in a parallel way.
Also, given that this happens after the sorting step, my naive approach would be to go over all particles in the `par_for` and use an atomic operation to updated the start index for the per cell values.

pgrete (commented):
@brryan thanks for this PR. The overall functionality will definitely be useful downstream.
I left a couple of (high level) comments with respect to the approach(es) chosen.

src/utils/sort.hpp
Yurlungur: I think @brryan was scared off by this issue:
https://github.com/kokkos/kokkos/issues/2914

For intel and AMD GPUs I suggest we fall back to `Kokkos::sort`. But I like using `thrust::sort` and `std::sort` on appropriate hardware for now, until `Kokkos::sort` is more performant.

src/interface/swarm.cpp
Yurlungur: There is a fair amount on this kind of algorithm, but I'm in favor of getting something working now and making it fast later. I think ideally in the long term we would want to rely on `Kokkos::sort`, rather than roll our own solution.

example/particles/particles.cpp
```diff
@@ -139,11 +143,25 @@ TaskStatus DestroySomeParticles(MeshBlock *pmb) {
   return TaskStatus::complete;
 }
 
+TaskStatus SortParticlesIfNecessary(MeshBlock *pmb) {
+  auto pkg = pmb->packages.Get("particles_package");
+  const auto deposition_method = pkg->Param<std::string>("deposition_method");
+  if (deposition_method == "per_cell") {
```
Yurlungur: Is this the "Necessary" condition?

src/interface/swarm.cpp
```diff
@@ -472,6 +479,111 @@ void Swarm::Defrag() {
...
+  // Write an unsorted list
+  pmb->par_for(
+      "Write unsorted list", 0, max_active_index_, KOKKOS_LAMBDA(const int n) {
+        int i, j, k;
+        swarm_d.Xtoijk(x(n), y(n), z(n), i, j, k);
+        const int cell_idx_1d = i + nx1 * (j + nx2 * k);
+        cellSorted(n) = SwarmKey(cell_idx_1d, n);
+      });
+
+  sort(cellSorted, SwarmKeyComparator(), 0, max_active_index);
```
Yurlungur, line 526: This actually came out really nice, didn't it?

src/interface/swarm.cpp
Yurlungur: oh wait I misunderstood this when just looking at the comments before looking at the code. I deleted my above comment as not relevant.

src/utils/sort.hpp
```diff
@@ -0,0 +1,75 @@
...
+#else
```
Yurlungur, line 45: `#elif some check we're on CPU`

src/utils/sort.hpp
```diff
@@ -0,0 +1,75 @@
...
+  std::sort(data.data() + min_idx, data.data() + max_idx + 1, comparator);
+#endif // KOKKOS_ENABLE_CUDA
```
Yurlungur, line 51: `#else Kokkos::sort`

Yurlungur (approved):
Nice work @brryan I'm pleased how simple this came out. For the most part, I second @pgrete 's suggestions but overall am happy with this as it is now.

src/interface/swarm.cpp
brryan: So AMReX uses a counting sort algorithm to organize its particles. I need to read a bit more to understand exactly how that might work here but it might avoid the cost of this kernel, basically.

src/utils/sort.hpp
brryan: `std::is_same<DevExecSpace, HostExecSpace>::value`? I'm not sure I really trust that but I'm not sure how else to check for CPU

Yurlungur: oh... that's a good idea. That should work. Not sure you can do that inside a macro though...

example/particles/particles.cpp
brryan: I had thought that in at least some cases (many particles per cell, for example) this code path might be more performant than the use of atomics, although maybe I am too distrustful of atomic operations. This example can actually be used to answer that performance question for varying numbers of cells and particles. Is your expectation that the atomics approach will always win?

src/interface/swarm_device_context.hpp
brryan: Yeah this will probably be a problem eventually, I should've been using `int64_t` from the beginning 😬

src/interface/swarm.cpp
brryan: Looking a little more closely I'm not sure the AMReX approach is a clear win relative to what we're doing here, at least looking at https://github.com/AMReX-Codes/amrex/blob/development/Src/Particle/AMReX_DenseBins.H there appears to be a serial region more or less in place of this block of code here.

Similar to the question of atomics in the particles example, is your expectation @pgrete that replacing this with a loop over particles and atomics would be more performant than what I have here even for a large number of particles per cell?

example/particles/particles.cpp
pgrete: Good question. I guess the answer will strongly depend on whether the particles are actually sorted or not.
Probably only benchmarking will tell.

src/interface/swarm.cpp
pgrete: As above, only benchmark will (at least somewhat) answer the questions and I expect that the answer heavily depends on whether the particles are sorted (and, as you say, the average number of particles per cell).
If they are sorted, then I'm fairly optimistic that a parallel for region will be faster when written in a way that an individual thread handles N > 1 particles at a time (so that the atomic update of the index for a given individual cell is handles as few threads at possible simultaneously).

src/utils/sort.hpp
pgrete: How about asking the Kokkos team about the current status and doing some (limited) performance tests covering the intended (immediate) use case  ourselves?
My main motivation is to keep the amount of different code paths as low as possible (especially given that the other architecture will become more and more available next year).

example/particles/particles.cpp
brryan: Sorry, should've responded before, this sort is necessary if we're using the pre-sorted per-cell lists for the non-atomic branch of the subsequent particle deposition call (which assumes those sorted lists exist). `SortParticlesIfNecessaryForDeposition`?

example/particles/particles.cpp
brryan: Yeah, since these sorted arrays are used by `"per_cell"` deposition in the next task

src/interface/swarm.cpp
brryan: The benefit of letting another library launch kernels for you :)

Yurlungur:
@brryan what's the status of this?

example/particles/particles.cpp
brryan: Done

example/particles/particles.cpp
brryan: Changed to `SortParticlesIfUsingPerCellDeposition`

src/interface/swarm_device_context.hpp
brryan: Actually this particular one may always be fine since at most this should be `nx*ny*nz` of a meshblock times the number of meshblocks on a processor. We'd need about 1200^3 cells on a single compute processor to cause an issue here. I'll still add the debug check now just for future-proofing.

I'm actually adding this check where `cell_idx_1d` is evaluated, line 520 of `src/interface/swarm.cpp`, since it will already arrive here as an integer.

src/interface/swarm_container.cpp
brryan: Ok removed

src/utils/sort.hpp
brryan: I looked into this but because `Kokkos::sort` doesn't currently support key-value pairs, this would be a substantial rewrite of this feature. Since it sounds like it is on the Kokkos team's [radar](https://github.com/kokkos/kokkos/issues/4487) to improve sort including supporting key-value pairs like here, and since we don't need non-CUDA non-CPU sorting now, I propose that we table this and we reassess when `Kokkos::sort` is updated (maybe we can remove both `thrust::sort` and `std::sort` calls then).

src/utils/sort.hpp
brryan: As I just replied to Jonah, it [sounds like](https://github.com/kokkos/kokkos/issues/4487#issuecomment-963540068) the Kokkos team is looking to rewrite their sorting methods, which should both allow for key-value pair sorting like I'm using here and hopefully improve performance. OK if I leave this PR as-is and open an issue to plan to move to `Kokkos::sort` exclusively when it is ready?

src/interface/swarm.cpp
brryan: For this kernel the particles are already sorted (which we assume when guessing the start_index). I actually don't immediately see how to write a kernel that gives each thread multiple particles to work with (right now I'm only looping over all particles or all cells with parthenon kernels) -- are you envisioning some runtime parameter that sets the number of particles per thread in this kernel? That might be a bit esoteric for the user, and unwarranted if this kernel isn't very slow (my guess is it already probably is not a dominant cost)

If this kernel does turn out to be slow, a potentially simpler speedup approach would be to just iteratively guess the start index, by evaluating the expected number of particles the previous guess of `start_index` was off by based on the `cell_idx_1d` of the previously guessed `start_index`

brryan:
@Yurlungur Thanks for following up, I responded to the queries above and if you, @pgrete, and the CI are happy I think this is ready

src/utils/sort.hpp
brryan: To be clear this PR currently will not work for Intel or AMD GPUs as is, but is that a broad requirement now?

src/utils/sort.hpp
Yurlungur: works for me.

Yurlungur:
> @Yurlungur Thanks for following up, I responded to the queries above and if you, @pgrete, and the CI are happy I think this is ready

I'm happy. But I'll wait for @pgrete to approve as well.

src/interface/swarm.cpp
pgrete: > For this kernel the particles are already sorted (which we assume when guessing the start_index). I actually don't immediately see how to write a kernel that gives each thread multiple particles to work with (right now I'm only looping over all particles or all cells with parthenon kernels) -- are you envisioning some runtime parameter that sets the number of particles per thread in this kernel? That might be a bit esoteric for the user, and unwarranted if this kernel isn't very slow (my guess is it already probably is not a dominant cost)

What I had in mind was something like a  `TeamPolicy` where the `N` particles are automatically split across Team and that each thread in a team handles "warp_size" number of indices at a time.
Again, assuming the particles are already sorted should limit the clashes of atomic updates.
I'm happy to table this and leave this discussion for a later when this turns out to be a bottleneck.

My gut feeling is that doing nx1*nx2*nx3 searches (in parallel with potential branch divergence across threads) will be slower than `num_active` parallel atomic updates in many cases.

src/interface/swarm_device_context.hpp
pgrete: How about moving the `PARTHENON_DEBUG_REQUIRE` that currently lives inside a parallel region to somewhere outside (e.g., the constructor) and make it a check that is also triggered without debug mode?
My reasoning is that the MeshBlock size won't change for a simulation, so we need the check only once, and that it's rather unlikely to run a problem at scale (where the issue will show first) in debug mode.

PS: I don't think that 1k^3 is that far off. We can already fit multi-stage hydrodynamics at 512^3 on a single A100 with 40G.
So if I take hydro out and just do pure n-body (i.e., reducing the number of cell centered variables needed) and take the next gen GPU with >=128G memory, we may get there sooner than later.

src/utils/sort.hpp
pgrete: Following the discussion on the `Kokkos::sort` status, I'm fine with keeping thrust for Cuda for now.
From a functional point of view it'd still be great to call `Kokkos::sort()` if neither `std::sort` nor `thrust` is availabe, or throw a meaningful error message (at compile time).
AMD GPUs may be there sooner than later and I'm currently fixing a couple of things to make those work in #629 as AMD GPUs will become very relevant for us this year.

pgrete (commented):
I left two minor comments as replies to previous threads and have one additional question that I missed in the first review:
Is the new deposition method tested somewhere,i.e, that it results in the same distribution as the original method?

src/interface/swarm_device_context.hpp
brryan: Good point about lack of debug mode at scale -- I just moved the `PARTHENON_REQUIRE` up out of the `KOKKOS_LAMBDA` but in the same function so it's clear where the problem is -- I don't think that should hurt performance (and the constructor itself doesn't have a meshblock pointer so the check would have to be in `SetBlockPointer` and then might not be propagated correctly if that part of the code is refactored). 

PS The future is arriving sooner than I expected :)

src/utils/sort.hpp
brryan: Good point, yes, I should include an error message (done). This might still cause a compiler warning unless the `if(std::is_same<...` is a `constexpr if`, but I assume it will be if compiling in C++17 mode?

Searching real quick, it seems to me that AMD GPUs support their [Bolt library, which has its own sort function?](https://hsa-libraries.github.io/Bolt/html/) However, I can't immediately tell whether that is guaranteed to be included in an AMD compilation environment like thrust is with CUDA. However, if we need a performant sort on AMD in the near term, that is one possibility

brryan:
@pgrete How do you feel about approving this? Looking through the comments just now I think I closed the loop on everything, but I didn't mark anything resolved in case I'm mistaken

pgrete (approved):
Thanks for the updates. Looks good. I will merge now.

Issue #621, opened 2021-12-14
Const correctness in Params


Yurlungur:
I have a heretical suggestion: `params` should not return a const reference. It should just return a reference. We are already using `params` for non-const objects, namely containers. We're fooling the `const` correctness by just hiding things behind `Kokkos::Views`. Since we're not obeying the spirit of `const`, why are we obeying the letter?

This is also relevant for things like global reductions. I recently discovered that I can't hide a `Reduce<T>` object in `Params` without putting it inside a `Kokkos::View` because `const` correctness prevents me from calling the reducer methods, but I need the object to persist---it can't be a copy.

Interested in what people think.

brryan:
In the situation where a large number of packages maintained by different developers are all interacting through the same driver, I can see the advantages of being able to set parameters to be constant (it'd be nice if this could be enforced for data in `View`s as well). 

But it seems clear to me that a `params`-like bin for non-`const` data is needed. Like you say we already use it. In my mind the optimal solution is a genuinely-`const` `params` and a separate non-`const` entity called `PackageData` or something. Existing non-constant data in params could gradually be migrated (or not)

Yurlungur:
Hmm this gives me an idea, @brryan . I think we can easily do both with a single object and minimal code surgery. Check out PR #623.

PR #622, opened 2021-12-14
Generalize the Reductions to Support More Data Types

Yurlungur:
## PR Summary

This PR is motivated by a desire downstream to reduce more data types than primitives and `std::vector`. In particular, I wanted to reduce `Kokkos::View`s (e.g., `ParArray*D`s). This required several changes:
1. @jdolence 's reduction machinery has some hooks for new data types. I use some template black magic to generalize the hook for `std::vector` to support any container-like object that has `size` and `data` methods.
2. Because Kokkos uses the pointer semantics to do higher-dimensional arrays, I had to also add a little template magic to strip all pointers from a type. (The standard library can strip *one* pointer but not six.)
3. A managed `Kokkos::View` cannot live in the driver, as the driver lives outside the scope of the kokkos runtime. This necessitates one of several possible solutions, which I will discuss below.

### What to do about `Reduce<ParArray*D<T>>`

In the `Poisson` example, the `AllReduce` and `Reduce` objects for `int`, `Real`, and `std::vector` lived in the driver. Unfortunately, a managed `Kokkos::View` can't live in the driver, because this breaks `Kokkos`'s reference counting. (The code runs, but Kokkos yells at you.) To complicate the issue, the data for the reduction must obviously be persistent, so it can't live in the `TaskList`. And the same is true for the `Reduce` object itself, since it contains state, including things like MPI tags.

There are two solutions I thought of:
1. Stash the `Kokkos::View` in the `Params` of a package. Keep the `Reduce` object in the driver, but make it only take an unmanaged `Kokkos::View` so that Kokkos's garbage collection continues to function. This is what I implemented in the Poisson example.
2. Stash the `Reduce` object *and* the `View` in the params of a package. This actually is actually surprisingly subtle, due to the fact that `Params` always returns a `const` reference. Since `Reduce` has non-const methods that must be called (and captured) in the task list, one can't simply pull the `Reduce` out of params as a reference. And since `Reduce` is stateful, one can't copy it out of `Params` by value. One could redesign `Reduce` to hide its own state behind pointers, or one can hide `Reduce` inside a `Kokkos::View` or smart pointer in `Params`, fooling the `const` correctness. I found this very ugly, so I didn't end up pushing it. But it motivated issue #621 : if we remove `const` correctness from `Params`, we could implement this much more cleanly.

Pinging @jdolence since he wrote the reduction framework and @pgrete for thoughts from the Athena++ folks.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


Yurlungur:
@par-hermes format

pgrete:
I need some more time to think about this as I'm not sure what the implications are to effectively make all our basic data structures "unmanaged".

Yurlungur:
> I need some more time to think about this as I'm not sure what the implications are to effectively make all our basic data structures "unmanaged".

I'm not sure what you mean by this @pgrete . The only implication is that any Kokkos views that live in the driver must be unmanaged, which was the case before this PR as well.

That said, the issue will be resolved with PR #623 . If params can be made mutable, then we can stash the reduction objects in `params` and they can be reference counted as usual.

pgrete:
Oh, my bad. I missed that these were new typedefs. I thought the original ones were redefined as unmanaged.

pgrete:
@Yurlungur , now that #623 is merged, do you plan to make any modifications to this PR?

Also, could you please elaborate on "Keep the Reduce object in the driver, but make it only take an unmanaged Kokkos::View so that Kokkos's garbage collection continues to function."
If I didn't miss anything `Kokkos::Init` is called before the driver and also the packages are processes before the driver is started.
So I'd expect that the managed views should work in this case.

Yurlungur:
> @Yurlungur , now that #623 is merged, do you plan to make any modifications to this PR?

None are needed, but I'm willing to modify to move views out of the driver, if that's preferred.
 
> Also, could you please elaborate on "Keep the Reduce object in the driver, but make it only take an unmanaged Kokkos::View so that Kokkos's garbage collection continues to function." If I didn't miss anything `Kokkos::Init` is called before the driver and also the packages are processes before the driver is started. So I'd expect that the managed views should work in this case.

The driver object is a field of the parthenon manager. So it exists before `Kokkos::Init` is called.

Yurlungur:
Reminder. Can I get some reviews on this? @pgrete ? @brryan @jdolence @lroberts36 ?

src/utils/reductions.hpp
```diff
@@ -44,17 +46,21 @@ enum MPI_Op {
...
 void *GetPtr(U &v) {
   return &v;
 }
 
-template <typename U>
-int GetSize(std::vector<U> &v) {
+// Generic for containers with a size() method, including Kokkos::View and std::vector
+template <template <typename U, typename... Args> class Container, typename U,
+          typename... Args>
+int GetSize(Container<U, Args...> &v) {
```
lroberts36, line 63: This might be made more generic with 
```C++
template <class T> auto GetSizeImp(T& v, int) -> decltype(v.size()) {return v.size();}
template <class T> auto GetSizeImp(T& v, double) -> int {return 1;}
template <class T> auto GetSize(T& v) -> decltype(GetSizeImp(v, int())) { 
    return GetSizeImp(std::forward<T&>(v), int());
}
```
and something similar for `GetPtr` above

lroberts36 (approved):
This looks good to me. Only one small suggestion that you can take or leave below.

src/utils/reductions.hpp
Yurlungur: Hmm that's interesting. I agree that would work and is more generic.

Yurlungur: I tried this briefly and ran into a few inconsistencies with other parts of the reduce infrastructure. So I think I'll skip this for now. But I'll create an issue.

Yurlungur:
@par-hermes format

Yurlungur:
@pgrete I carved out 5 minutes to move the kokkos view that we reduce into params. It works beautifully. I removed the unmanaged view typedefs.

example/advection/#parthinput.advection#
```diff
@@ -0,0 +1,73 @@
+# ========================================================================================
+#  Athena++ astrophysical MHD code
```
pgrete: This file can probably go?

example/advection/advection_package.
```diff
@@ -0,0 +1 @@
+c
```
pgrete, line 1: This can probably go as well?

example/poisson/poisson_package.hpp
```diff
@@ -15,11 +15,14 @@
...
 #include <memory>
 
+#include <kokkos_abstraction.hpp>
 #include <parthenon/package.hpp>
 
 namespace poisson_package {
 using namespace parthenon::package::prelude;
 
+using HostArray1D = typename parthenon::ParArray1D<int>::HostMirror;
+
```
pgrete: If we are using this pattern more often downstream it may be worth putting those types for all dims in `kokkos_abstraction`.

example/poisson/poisson_package.cpp
```diff
@@ -86,6 +86,20 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
...
+  // We must use a mutable Param here so that
+  // the object can be modified by the driver.
+  parthenon::AllReduce<HostArray1D> view_reduce;
+  view_reduce.val = HostArray1D("Reduce me", 10);
+  for (int i = 0; i < view_reduce.val.size(); i++) {
+    view_reduce.val(i) = 0;
+  }
+  for (int i = 0; i < view_reduce.val.size(); i++) {
+    view_reduce.val(i) += i;
+  }
```
pgrete, line 105: Is the first `for` required or just there to illustrate "initialization"?

pgrete (approved):
Looks good now (pending the cleanup of the extra(?) files I'm already approving).

One question that I started to wonder is: what happens if I would use a device view for a reduction -- also with respect to Kokkos reduction becoming non blocking when the reducer is a view?

example/advection/#parthinput.advection#
Yurlungur: whoops! That should not have been committed. Yes I will delete. Also I will add emacs backup files to the giitignore.

example/advection/advection_package.
Yurlungur: embarassing this made it in here too... removed.

example/poisson/poisson_package.cpp
Yurlungur: Just to illustrate initialization. I will make a comment saying so.

example/poisson/poisson_package.hpp
Yurlungur: good idea.

Yurlungur:
> One question that I started to wonder is: what happens if I would use a device view for a reduction 

I haven't actually tried this. I *think* it will depend on whether or not MPI is CUDA aware. For CUDA aware MPI it should just work.

> also with respect to Kokkos reduction becoming non blocking when the reducer is a view?

The "in-task-list" reduction is handled by a separate task. So I think it's up to the user to put a kokkos barrier at an appropriate place. Maybe a barrier task before the MPI reduce task?

Yurlungur:
With all requested changes made and tests passing, I'm pullng the trigger.

PR #623, opened 2021-12-14
enable volatile params

### LINKED ISSUE #621: Const correctness in Params

Yurlungur:
## PR Summary

This is motivated by @brryan 's comment in issue #621 . I modify `Params` to enable the user to tag a param as "volatile." The effect of this is that the user can optionally extract the *non*-`const` pointer from the `Params` object with `GetVolatile` in the `Params` object or `VolatileParam` in the `StateDescriptor` object. If the user attempts to extract a non-volatile param with `GetVolatile`, a runtime error is raised.

I thought about returning a non-`const` reference, but the pointer felt safer and more explicit, without being any more cumbersome really. On the other hand, returning a pointer with `GetVolatile` means that the two `Get` methods return different types, which is a little weird. So I'd be open to returning a non-`const` reference instead, if people would like that better.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/src/interface/params.hpp
+++ b/src/interface/params.hpp
@@ -43,10 +43,11 @@ class Params {
   ///
   /// Throws an error if the key is already in use
   template <typename T>
-  void Add(const std::string &key, T value) {
+  void Add(const std::string &key, T value, bool make_volatile = false) {
     PARTHENON_REQUIRE_THROWS(!(hasKey(key)), "Key " + key + " already exists");
     myParams_[key] = std::unique_ptr<Params::base_t>(new object_t<T>(value));
     myTypes_.emplace(make_pair(key, std::type_index(typeid(value))));
+    myVolatile_[key] = make_volatile;
   }
 
   /// Updates existing object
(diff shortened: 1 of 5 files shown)


Yurlungur:
Pinging @brryan @pgrete @jdolence @nmsriram @forrestglines as people with stakes in this machinery and/or who wrote the original version

dholladay00:
small nomenclature suggestion: find-replace `volatile` with `mutable`.

Yurlungur:
> small nomenclature suggestion: find-replace `volatile` with `mutable`.

This is a good idea.

Yurlungur:
> small nomenclature suggestion: find-replace `volatile` with `mutable`.

done.

src/interface/params.hpp
```diff
@@ -43,10 +43,11 @@ class Params {
   ///
   /// Throws an error if the key is already in use
   template <typename T>
-  void Add(const std::string &key, T value) {
+  void Add(const std::string &key, T value, bool make_mutable = false) {
```
pgrete: Nitpicking: I suggest `is_mutable` (here and below) as I see this more like a property of the parameter rather than as an a posteriori action.

src/interface/params.hpp
```diff
@@ -43,10 +43,11 @@ class Params {
...
     PARTHENON_REQUIRE_THROWS(!(hasKey(key)), "Key " + key + " already exists");
     myParams_[key] = std::unique_ptr<Params::base_t>(new object_t<T>(value));
     myTypes_.emplace(make_pair(key, std::type_index(typeid(value))));
+    myMutable_[key] = make_mutable;
```
pgrete: Out of curiosity, is there a reason to use `emplace` in one case and the direct assignment in the other case?
Also, I wonder, if it wouldn't be cleaner the only use one map (e.g., with a struct) instead of three, but that's probably a separate discussion.

pgrete (changes requested):
:+1: I like this approach to be more specific.

The one additional change I'd like to see before approving is an update to the `Update` function.
From a consistency point of view, I think that if we differentiate between mutable and non-mutable parameter, then  `Update` should also check if the parameter is mutable and otherwise throw an error.

src/interface/params.hpp
Yurlungur: I used assignment because it was easier. I don't know why the original was `emplace`. I can copy that if it's preferred.

src/interface/params.hpp
Yurlungur: done.

Yurlungur:
@pgrete take a look now. I added a check for `mutable` in `Params::Update` and renamed `make_mutable` to `is_mutable`.

Yurlungur:
Pinging @gshipman to review and @pgrete to approve as discussed on the parthenon call.

pgrete (approved):
Looks good to go!

jdolence (approved):

PR #649, opened 2022-02-08
Ensure LB send buffers are filled and allow async recv LB

pgrete:
## PR Summary

First part of the PR:
As far as I can tell there might have been issues on devices because there was no `fence` before the `MPI_ISend`.
Within the functions called in that loop, we also used the triple argument `deep_copy`, which is non-blocking.
Now, all the MPI comm is started at the end of the loop and we first gather the info (tags, destination rank, data count) similar to other places in the code.

Second part of the PR:
Blocking `MPI_Wait` was used for receiving blocks during AMR LB.
This was probably never a performance issue with only "few" blocks per process, but on GPUs we now handle thousands of blocks, so I expect a performance gain by actually using non-blocking calls.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - [[PR 617]](https://github.com/lanl/parthenon/pull/617) Unify the coordinates API for MeshBlockPack and VariablePack
 
 ### Fixed (not changing behavior/API/variables/...)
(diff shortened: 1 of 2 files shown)


Yurlungur (approved):
LGTM assuming all tests pass.

Yurlungur:
Missing CHANGELOG at the very least.

forrestglines (approved):
As far as I can tell, looks good

PR #653, opened 2022-02-23
Allow for multi-D particle variables

### LINKED ISSUE #657: Particles don't work for one meshblock with outflow boundaries
In this case there are zero neighbors which parthenon particles currently cannot handle (`neighbor_buffer_index` is sized to `0`)

brryan:
## PR Summary

We want to allow particles to transport vector quantities as well as scalar quantities. E.g. one `ParticleVariable` `v[3]` instead of three `ParticleVariable`s `vx`, `vy`, and `vz`, as well as vectors whose size can vary at runtime.

~~In some places~~ the code supports up to 5D quantities for a `ParticleVariable`~~, but elsewhere the dimensionality is restricted (e.g. when packing/unpacking particles for MPI communication, only 2D `ParticleVariable`s are supported due to `FlatIdx` only supporting up to 3 indices currently). I included a check in the `PackAllVariables` function,~~

This PR also replaces existing `ParticleVariable<bool>`s with `ParArray1D<bool>`s. 

Fixes #657

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/docs/particles.md
+++ b/docs/particles.md
@@ -113,8 +113,9 @@ AMR is currently not supported, but support will be added in the future.
 Similarly to grid variables, particle swarms support `ParticleVariable` packing, by the function
 `Swarm::PackVariables`.
 
-Note that this 1D or 2D ParticleVariables are currently untested with this function; in particular,
-`FlatIdx` is not guaranteed to work correctly.
+Note that higher-dimensional (more than one element per particle) ParticleVariables are
+currently untested with this function; in particular, `FlatIdx` is not guaranteed to work
+correctly.
 
 ## Boundary conditions
 
--- a/docs/particles.md
+++ b/docs/particles.md
@@ -12,8 +12,8 @@ positions `x`, `y`, and `z` of the particles in a swarm are three separate
 `ParticleVariable`s. `ParticleVariable`s can be either `Real`- or `int`-valued, which is
 specified by the metadata values `Metadata::Real` and `Metadata::Integer`.
 `ParticleVariable`s should also contain the `Metadata::Particle` flag. By default,
-`ParticleVariable`s provide one scalar quantity per particle, but 1D data per particle is
-currently supported, by passing `std::vector<int>{N}` as the second argument to the
+`ParticleVariable`s provide one scalar quantity per particle, but up to 2D data per particle is
+currently supported, by passing `std::vector<int>{N1, N2}` as the second argument to the
 `ParticleVariable` `Metadata`. All `Swarm`s by default contain `x`, `y`, and `z`
 `ParticleVariable`s; additional fields can be added as:
 ```c++
@@ -107,6 +107,14 @@ further details. Note that this pattern is blocking, and may be replaced in the
 future.
 
 AMR is currently not supported, but support will be added in the future.
+
+## Variable Packing
+
+Similarly to grid variables, particle swarms support `ParticleVariable` packing, by the function
+`Swarm::PackVariables`.
+
+Note that this 1D or 2D ParticleVariables are currently untested with this function; in particular,
+`FlatIdx` is not guaranteed to work correctly.
 
 ## Boundary conditions
 
--- a/docs/particles.md
+++ b/docs/particles.md
@@ -110,12 +110,9 @@ AMR is currently not supported, but support will be added in the future.
 
 ## Variable Packing
 
-Similarly to grid variables, particle swarms support `ParticleVariable` packing, by the function
-`Swarm::PackVariables`.
-
-Note that higher-dimensional (more than one element per particle) ParticleVariables are
-currently untested with this function; in particular, `FlatIdx` is not guaranteed to work
-correctly.
+Similarly to grid variables, particle swarms support `ParticleVariable` packing, by the
+function `Swarm::PackVariables`. This also supports `FlatIdx` for indexing; see the
+`particle_leapfrog` example for usage.
 
 ## Boundary conditions
 
--- a/docs/particles.md
+++ b/docs/particles.md
@@ -10,8 +10,12 @@ A `Swarm` contains all the particle data for all particles of a given species. I
 set of `ParticleVariable`s, one for each value of each particle. For example, the spatial
 positions `x`, `y`, and `z` of the particles in a swarm are three separate
 `ParticleVariable`s. `ParticleVariable`s can be either `Real`- or `int`-valued, which is
-specified by the metadata values `Metadata::Real` and `Metadata::Integer`. All `Swarm`s by
-default contain `x`, `y`, and `z` `ParticleVariable`s; additional fields can be added as:
+specified by the metadata values `Metadata::Real` and `Metadata::Integer`.
+`ParticleVariable`s should also contain the `Metadata::Particle` flag. By default,
+`ParticleVariable`s provide one scalar quantity per particle, but 1D data per particle is
+currently supported, by passing `std::vector<int>{N}` as the second argument to the
+`ParticleVariable` `Metadata`. All `Swarm`s by default contain `x`, `y`, and `z`
+`ParticleVariable`s; additional fields can be added as:
 ```c++
 Swarm.Add(name, metadata)
 ```
@@ -26,10 +30,10 @@ Hereafter we refer to it as `swarm_d`.
 
 To add particles to a `Swarm`, one calls
 ```c++
-ParArrayND<bool> new_particles_mask = swarm->AddEmptyParticles(num_to_add, new_indices)
+ParArray1D<bool> new_particles_mask = swarm->AddEmptyParticles(num_to_add, new_indices)
 ```
 This call automatically resizes the memory pools as necessary and returns a
-`ParArrayND<bool>` mask indicating which indices in the `ParticleVariable`s are newly
+`ParArray1D<bool>` mask indicating which indices in the `ParticleVariable`s are newly
 available. `new_indices` is a reference to a `ParArrayND<int>` of size `num_to_add` which
 contains the indices of each newly added particle.
 
--- a/example/particle_leapfrog/particle_leapfrog.cpp
+++ b/example/particle_leapfrog/particle_leapfrog.cpp
@@ -185,11 +185,6 @@ TaskStatus WriteParticleLog(BlockList_t &blocks, int ncycle) {
           for (int j = 0; j < 3; j++) {
             if (std::fabs(vv(i, j, n)) > 1.e-10) {
               PARTHENON_REQUIRE(
-                  2. * std::fabs(vp(ivv(i, j), n) - vp(iv(i), n) * vp(iv(j), n)) <
-                      1.e-10 * (std::fabs(vp(ivv(i, j), n)) +
-                                std::fabs(vp(iv(i), n) * vp(iv(j), n))),
-                  "packed vv not consistent with packed v*v!");
-              PARTHENON_REQUIRE(
                   2. * std::fabs(vv(i, j, n) - v(i, n) * v(j, n)) <
                       1.e-10 * (std::fabs(vv(i, j, n)) + std::fabs(v(i, n) * v(j, n))),
                   "vv not consistent with v*v!");
@@ -198,6 +193,24 @@ TaskStatus WriteParticleLog(BlockList_t &blocks, int ncycle) {
         }
       }
     }
+  
+    const auto is_active_d = swarm->GetMask();
+    pmb->par_for(
+      "CheckParticlePacks", 0, swarm->GetMaxActiveIndex(), KOKKOS_LAMBDA(const int n) {
+      if (is_active_d(n)) {
+        for (int i = 0; i < 3; i++) {
+          for (int j = 0; j < 3; j++) {
+            if (std::fabs(vp(ivv(i, j), n)) > 1.e-10) {
+              PARTHENON_REQUIRE(
+                  2. * std::fabs(vp(ivv(i, j), n) - vp(iv(i), n) * vp(iv(j), n)) <
+                      1.e-10 * (std::fabs(vp(ivv(i, j), n)) +
+                                std::fabs(vp(iv(i), n) * vp(iv(j), n))),
+                  "packed vv not consistent with packed v*v!");
+            }
+          }
+        }
+      }
+    });
   }
 
   // Step 2b. Gather data on root process
--- a/example/particle_leapfrog/particle_leapfrog.cpp
+++ b/example/particle_leapfrog/particle_leapfrog.cpp
@@ -47,6 +47,25 @@ Packages_t ProcessPackages(std::unique_ptr<ParameterInput> &pin) {
   return packages;
 }
 
+// initial particle position: x,y,z,vx,vy,vz
+constexpr int num_test_particles = 14;
+const std::array<std::array<Real, 6>, num_test_particles> particles_ic = {{
+    {-0.1, 0.2, 0.3, 1.0, 0.0, 0.0},   // along x direction
+    {0.4, -0.1, 0.3, 0.0, 1.0, 0.0},   // along y direction
+    {-0.1, 0.3, 0.2, 0.0, 0.0, 0.5},   // along z direction
+    {0.0, 0.0, 0.0, -1.0, 0.0, 0.0},   // along -x direction
+    {0.0, 0.0, 0.0, 0.0, -1.0, 0.0},   // along -y direction
+    {0.0, 0.0, 0.0, 0.0, 0.0, -1.0},   // along -z direction
+    {0.0, 0.0, 0.0, 1.0, 1.0, 1.0},    // along xyz diagonal
+    {0.0, 0.0, 0.0, -1.0, 1.0, 1.0},   // along -xyz diagonal
+    {0.0, 0.0, 0.0, 1.0, -1.0, 1.0},   // along x-yz diagonal
+    {0.0, 0.0, 0.0, 1.0, 1.0, -1.0},   // along xy-z diagonal
+    {0.0, 0.0, 0.0, -1.0, -1.0, 1.0},  // along -x-yz diagonal
+    {0.0, 0.0, 0.0, 1.0, -1.0, -1.0},  // along x-y-z diagonal
+    {0.0, 0.0, 0.0, -1.0, 1.0, -1.0},  // along -xy-z diagonal
+    {0.0, 0.0, 0.0, -1.0, -1.0, -1.0}, // along -x-y-z diagonal
+}};
+
 // *************************************************//
 // define the "physics" package particles_package, *//
 // which includes defining various functions that  *//
@@ -88,6 +107,29 @@ Real EstimateTimestepBlock(MeshBlockData<Real> *rc) {
   int max_active_index = swarm->GetMaxActiveIndex();
 
   const auto &v = swarm->Get<Real>("v").Get();
+  // Test, check particles
+
+
+  {
+    const auto &id = swarm->Get<int>("id").Get();
+  auto swarm_d = swarm->GetDeviceContext();
+  pmb->par_for("TEST", 0, max_active_index, KOKKOS_LAMBDA(const int n) {
+    if (swarm_d.IsActive(n)) {
+      bool is_good = true;
+      if (particles_ic[id(n)][3] != v(0,n)) is_good = false;
+      if (particles_ic[id(n)][4] != v(1,n)) is_good = false;
+      if (particles_ic[id(n)][5] != v(2,n)) is_good = false;
+      if (!is_good) {
+        printf("%s:%i\n", __FILE__, __LINE__);
+        printf("Particle %i velocity is bad! gid: %i\n", id(n), pmb->gid);
+        printf("v: %e %e %e\n", v(0,n), v(1,n), v(2,n));
+        printf("v true: %e %e %e\n", particles_ic[id(n)][3],
+          particles_ic[id(n)][4], particles_ic[id(n)][5]);
+        exit(-1);
+      }
+    }
+  });
+  }
 
   // Assumes a grid with constant dx, dy, dz within a block
   const Real &dx_i = pmb->coords.dx1f(0);
@@ -159,6 +201,27 @@ TaskStatus WriteParticleLog(BlockList_t &blocks, int ncycle) {
     const auto &y = swarm->Get<Real>("y").Get().GetHostMirrorAndCopy();
     const auto &z = swarm->Get<Real>("z").Get().GetHostMirrorAndCopy();
     const auto &v = swarm->Get<Real>("v").Get().GetHostMirrorAndCopy();
+  // Test, check particles
+  {
+  auto swarm_d = swarm->GetDeviceContext();
+  int max_active_index = swarm->GetMaxActiveIndex();
+  pmb->par_for("TEST", 0, max_active_index, KOKKOS_LAMBDA(const int n) {
+    if (swarm_d.IsActive(n)) {
+      bool is_good = true;
+      if (particles_ic[id(n)][3] != v(0,n)) is_good = false;
+      if (particles_ic[id(n)][4] != v(1,n)) is_good = false;
+      if (particles_ic[id(n)][5] != v(2,n)) is_good = false;
+      if (!is_good) {
+        printf("%s:%i\n", __FILE__, __LINE__);
+        printf("Particle %i velocity is bad! gid: %i\n", id(n), pmb->gid);
+        printf("v: %e %e %e\n", v(0,n), v(1,n), v(2,n));
+        printf("v true: %e %e %e\n", particles_ic[id(n)][3],
+          particles_ic[id(n)][4], particles_ic[id(n)][5]);
+        exit(-1);
+      }
+    }
+  });
+  }
 
     const auto &is_active =
         Kokkos::create_mirror_view_and_copy(HostMemSpace(), swarm->GetMask());
@@ -257,25 +320,6 @@ TaskStatus WriteParticleLog(BlockList_t &blocks, int ncycle) {
   return TaskStatus::complete;
 }
 
-// initial particle position: x,y,z,vx,vy,vz
-constexpr int num_test_particles = 14;
-const std::array<std::array<Real, 6>, num_test_particles> particles_ic = {{
-    {-0.1, 0.2, 0.3, 1.0, 0.0, 0.0},   // along x direction
-    {0.4, -0.1, 0.3, 0.0, 1.0, 0.0},   // along y direction
-    {-0.1, 0.3, 0.2, 0.0, 0.0, 0.5},   // along z direction
-    {0.0, 0.0, 0.0, -1.0, 0.0, 0.0},   // along -x direction
-    {0.0, 0.0, 0.0, 0.0, -1.0, 0.0},   // along -y direction
-    {0.0, 0.0, 0.0, 0.0, 0.0, -1.0},   // along -z direction
-    {0.0, 0.0, 0.0, 1.0, 1.0, 1.0},    // along xyz diagonal
-    {0.0, 0.0, 0.0, -1.0, 1.0, 1.0},   // along -xyz diagonal
-    {0.0, 0.0, 0.0, 1.0, -1.0, 1.0},   // along x-yz diagonal
-    {0.0, 0.0, 0.0, 1.0, 1.0, -1.0},   // along xy-z diagonal
-    {0.0, 0.0, 0.0, -1.0, -1.0, 1.0},  // along -x-yz diagonal
-    {0.0, 0.0, 0.0, 1.0, -1.0, -1.0},  // along x-y-z diagonal
-    {0.0, 0.0, 0.0, -1.0, 1.0, -1.0},  // along -xy-z diagonal
-    {0.0, 0.0, 0.0, -1.0, -1.0, -1.0}, // along -x-y-z diagonal
-}};
-
 void ProblemGenerator(MeshBlock *pmb, ParameterInput *pin) {
   auto pkg = pmb->packages.Get("particles_package");
   auto swarm = pmb->swarm_data.Get()->Get("my particles");
@@ -356,6 +400,27 @@ TaskStatus TransportParticles(MeshBlock *pmb, const StagedIntegrator *integrator
   auto &z = swarm->Get<Real>("z").Get();
   auto &v = swarm->Get<Real>("v").Get();
 
+  // Test, check particles
+  {
+  auto swarm_d = swarm->GetDeviceContext();
+  pmb->par_for("TEST", 0, max_active_index, KOKKOS_LAMBDA(const int n) {
+    if (swarm_d.IsActive(n)) {
+      bool is_good = true;
+      if (particles_ic[id(n)][3] != v(0,n)) is_good = false;
+      if (particles_ic[id(n)][4] != v(1,n)) is_good = false;
+      if (particles_ic[id(n)][5] != v(2,n)) is_good = false;
+      if (!is_good) {
+        printf("%s:%i\n", __FILE__, __LINE__);
+        printf("Particle %i velocity is bad! gid: %i\n", id(n), pmb->gid);
+        printf("v: %e %e %e\n", v(0,n), v(1,n), v(2,n));
+        printf("v true: %e %e %e\n", particles_ic[id(n)][3],
+          particles_ic[id(n)][4], particles_ic[id(n)][5]);
+        exit(-1);
+      }
+    }
+  });
+  }
+
   auto swarm_d = swarm->GetDeviceContext();
   // keep particles on existing trajectory for now
   const Real ax = 0.0;
@@ -383,6 +448,27 @@ TaskStatus TransportParticles(MeshBlock *pmb, const StagedIntegrator *integrator
           swarm_d.GetNeighborBlockIndex(n, x(n), y(n), z(n), on_current_mesh_block);
         }
       });
+
+  // Test, check particles
+  {
+  auto swarm_d = swarm->GetDeviceContext();
+  pmb->par_for("TEST", 0, max_active_index, KOKKOS_LAMBDA(const int n) {
+    if (swarm_d.IsActive(n)) {
+      bool is_good = true;
+      if (particles_ic[id(n)][3] != v(0,n)) is_good = false;
+      if (particles_ic[id(n)][4] != v(1,n)) is_good = false;
+      if (particles_ic[id(n)][5] != v(2,n)) is_good = false;
+      if (!is_good) {
+        printf("%s:%i\n", __FILE__, __LINE__);
+        printf("Particle %i velocity is bad! gid: %i\n", id(n), pmb->gid);
+        printf("v: %e %e %e\n", v(0,n), v(1,n), v(2,n));
+        printf("v true: %e %e %e\n", particles_ic[id(n)][3],
+          particles_ic[id(n)][4], particles_ic[id(n)][5]);
+        exit(-1);
+      }
+    }
+  });
+  }
 
   return TaskStatus::complete;
 }
--- a/example/particle_leapfrog/particle_leapfrog.cpp
+++ b/example/particle_leapfrog/particle_leapfrog.cpp
@@ -1,9 +1,9 @@
 //========================================================================================
 // Parthenon performance portable AMR framework
-// Copyright(C) 2021 The Parthenon collaboration
+// Copyright(C) 2021-2022 The Parthenon collaboration
 // Licensed under the 3-clause BSD License, see LICENSE file for details
 //========================================================================================
-// (C) (or copyright) 2020-2021. Triad National Security, LLC. All rights reserved.
+// (C) (or copyright) 2020-2022. Triad National Security, LLC. All rights reserved.
 //
 // This program was produced under U.S. Government contract 89233218CNA000001 for Los
 // Alamos National Laboratory (LANL), which is operated by Triad National Security, LLC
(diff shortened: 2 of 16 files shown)


brryan:
One thing I noticed while working on this is that the `particle_leapfrog` tests don't seem to be getting run by the CI. It's not obvious to me how exactly to fix this -- they seem to be enrolled in `tst/regression/CMakeLists.txt`?

src/interface/metadata.cpp
```diff
@@ -141,6 +141,13 @@ std::array<int, 6> Metadata::GetArrayDims(std::weak_ptr<MeshBlock> wpmb,
       arrDims[i + 3] = shape[i];
     for (int i = N; i < 3; i++)
       arrDims[i + 3] = 1;
+  } else if (Where() == Particle) {
+    assert(N >= 1 && N <= 5);
+    arrDims[0] = 1; // To be updated by swarm based on pool size before allocation
+    for (int i = 0; i < N; i++)
+      arrDims[i + 1] = shape[i];
+    for (int i = N; i < 5; i++)
+      arrDims[i + 1] = 1;
```
Yurlungur: I don't fully understand this change. Why do you need to specify `Particle` in the metadata if you're adding it to a swarm? Is it because this logic lives in the metadata?

If so, maybe we could remove the need to `AddSwarmValue` and just `Add` but with particle metadata, which short-circuits to adding a swarm?

brryan: Somehow inside `Metadata` we need to know if we're dealing with a `Particle` variable because these variables have a specific `arrDims` (first index is the index inside the swarm pool, the other five indices are for up to a 5D dataset per particle). 

If we remove `AddSwarmValue` would that cause issues for having multiple swarms in one application? Because `AddSwarmValue` adds the `ParticleVariable` to a specific `Swarm`

Yurlungur (approved):
LGTM also appreciate the miscellaneous cleanups

example/particles/particles.cpp
```diff
@@ -86,11 +90,11 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
   std::string swarm_name = "my particles";
   Metadata swarm_metadata({Metadata::Provides, Metadata::None});
   pkg->AddSwarm(swarm_name, swarm_metadata);
-  Metadata real_swarmvalue_metadata({Metadata::Real});
+  Metadata real_swarmvalue_metadata({Metadata::Real, Metadata::Particle});
```
lroberts36: I am not sure I understand why `Metadata::Particle` needs to be included here. Shouldn't any swarm be `Metadata::Particle` by default?

lroberts36: nm, just saw Jonah's comment below and the response.

src/interface/swarm.cpp
```diff
@@ -268,80 +266,59 @@ void Swarm::setPoolMax(const int nmax_pool) {
...
-        KOKKOS_LAMBDA(const int m) { newvar_data(m) = oldvar_data(m); });
-
-    intVector_[n] = newvar;
-    intMap_[oldvar->label()] = newvar;
+    auto &data = intVector_[n]->data;
+    std::array<int, 6> dim;
+    for (int i = 0; i < 6; i++) {
+      dim[i] = data.GetDim(6 - i);
+    }
+    data.Resize(dim[0], dim[1], dim[2], dim[3], dim[4], nmax_pool);
```
lroberts36: Why not 
```c++
for (auto& data : intVector) 
     data.Resize(data.GetDim[6], data.GetDim[5], data.GetDim[4], 
                 data.GetDim[3], data.GetDim[2], nmax_pool);
```

lroberts36 (approved):
LGTM.

docs/particles.md
```diff
@@ -10,8 +10,12 @@ A `Swarm` contains all the particle data for all particles of a given species. I
 set of `ParticleVariable`s, one for each value of each particle. For example, the spatial
 positions `x`, `y`, and `z` of the particles in a swarm are three separate
 `ParticleVariable`s. `ParticleVariable`s can be either `Real`- or `int`-valued, which is
-specified by the metadata values `Metadata::Real` and `Metadata::Integer`. All `Swarm`s by
-default contain `x`, `y`, and `z` `ParticleVariable`s; additional fields can be added as:
+specified by the metadata values `Metadata::Real` and `Metadata::Integer`.
+`ParticleVariable`s should also contain the `Metadata::Particle` flag. By default,
+`ParticleVariable`s provide one scalar quantity per particle, but 1D data per particle is
```
pgrete: Should this be `ND` rather than `1D`?

src/interface/metadata.cpp
pgrete: How about removing the `Particle` MetaData  from the metadata that is passed to `pkg->AddSwarmValue` and add it inside that function call before further processing?
As far as I can tell `AddSwarmValue` will always and only take `Particle` type data, doesn't it?

src/interface/swarm.cpp
pgrete: I also think that this would be more illustrative (at least the last part with the single line resize rather than with the additional `for` loop)

src/interface/swarm.cpp
```diff
@@ -268,80 +266,59 @@ void Swarm::setPoolMax(const int nmax_pool) {
...
-        "setPoolMax_real", 0, nmax_pool_ - 1,
-        KOKKOS_LAMBDA(const int m) { newvar_data(m) = oldvar_data(m); });
-    realVector_[n] = newvar;
-    realMap_[oldvar->label()] = newvar;
+    auto &data = realVector_[n]->data;
+    std::array<int, 6> dim;
+    for (int i = 0; i < 6; i++) {
+      dim[i] = data.GetDim(6 - i);
+    }
+    data.Resize(dim[0], dim[1], dim[2], dim[3], dim[4], nmax_pool);
```
pgrete: Same here

src/interface/swarm.cpp
```diff
@@ -455,23 +432,35 @@ void Swarm::Defrag() {
...
+            }
           }
           for (int i = 0; i < int_vars_size; i++) {
-            vint(i, from_to_indices(n)) = vint(i, n);
+            for (int j = 0; j < ipack_indices_shapes(1, i); j++) {
+              for (int k = 0; k < ipack_indices_shapes(2, i); k++) {
+                vint(ipack_indices_shapes(0, i), k, j, from_to_indices(n)) =
+                    vint(ipack_indices_shapes(0, i), k, j, n);
+              }
+            }
```
pgrete: I had troubles mentally parsing these lines here and then realized that this is/was likely related to me not being familiar with the packing of SwarmVariables.
If I didn't miss anything this is not documented yet, isn't it?
It'd be great to add some documentation both inline as well as in the `particles.md` file.

src/interface/swarm.cpp
```diff
@@ -533,7 +522,8 @@ void Swarm::SortParticlesByCell() {
       KOKKOS_LAMBDA(const int k, const int j, const int i) {
         int cell_idx_1d = i + nx1 * (j + nx2 * k);
         // Find starting index, first by guessing
-        int start_index = static_cast<int>((cell_idx_1d * num_active / ncells));
+        int start_index =
+            static_cast<int>((cell_idx_1d * static_cast<Real>(num_active) / ncells));
```
pgrete, line 490: Why this change? Aren't these all `int`s?

src/interface/swarm.cpp
```diff
@@ -893,7 +890,8 @@ void Swarm::SetupPersistentMPI() {
 
 int Swarm::CountParticlesToSend_() {
   auto blockIndex_h = blockIndex_.GetHostMirrorAndCopy();
-  auto mask_h = mask_.data.GetHostMirrorAndCopy();
+  auto mask_h = Kokkos::create_mirror_view_and_copy(HostMemSpace(), mask_);
+  // auto mask_h = mask_.data.GetHostMirrorAndCopy();
```
```suggestion
```

src/interface/swarm.hpp
```diff
@@ -285,18 +295,43 @@ inline SwarmVariablePack<T> Swarm::PackVariables(const std::vector<std::string>
 }
 
 template <class T>
-inline SwarmVariablePack<T> Swarm::PackAllVariables(PackIndexMap &vmap) {
+inline SwarmVariablePack<T>
+Swarm::PackAllVariables(PackIndexMap &vmap, ParArrayND<int> &pack_indices_shapes) {
```
pgrete: As mentioned above, some doc on `pack_indices_shapes` would be helpful potentially even as function docstring

src/interface/swarm.hpp
```diff
@@ -285,18 +295,43 @@ inline SwarmVariablePack<T> Swarm::PackVariables(const std::vector<std::string>
...
     names.push_back(v->label());
   }
-  return PackVariables<T>(names, vmap);
+
+  auto ret = PackVariables<T>(names, vmap);
+  auto map = vmap.Map();
+
+  // Get shape of packed variables
+  pack_indices_shapes = ParArrayND<int>("Pack indices and shapes", 6, names.size());
+  auto pack_indices_shapes_h = pack_indices_shapes.GetHostMirrorAndCopy();
```
pgrete: Why copy?

src/interface/variable.hpp
```diff
@@ -220,12 +220,13 @@ class FaceVariable {
...
-  FaceArray<T> data;
-
  private:
-  std::array<int, 6> dims_;
   Metadata m_;
   std::string label_;
+  std::array<int, 6> dims_;
+
+ public:
+  FaceArray<T> data;
```
pgrete: Why this change?

src/interface/variable.hpp
```diff
@@ -282,7 +282,15 @@ class ParticleVariable {
   }
 
   KOKKOS_FORCEINLINE_FUNCTION
-  auto GetDim(const int i) const { return data.GetDim(i); }
+  auto GetDim(const int i) const {
+    PARTHENON_REQUIRE(0 < i && i <= 6, "ParArrayNDGenerics are max 6D");
```
pgrete: How about `PARTHENON_DEBUG_REQUIRE` (similar to the assert in for the ParArrayND itself)?

src/interface/variable.hpp
```diff
@@ -295,12 +303,13 @@ class ParticleVariable {
...
-  ParArrayND<T> data;
-
  private:
-  int npool_;
   Metadata m_;
   std::string label_;
+  std::array<int, 6> dims_;
+
+ public:
+  ParArrayND<T> data;
```
pgrete, line 311: As for the FaceFields, why is this change required?
I'm asking as this is a parthenon public interface.

src/interface/variable_pack.hpp
```diff
@@ -573,17 +579,20 @@ void FillSwarmVarView(const vpack_types::SwarmVarList<T> &vars, PackIndexMap *vm
...
     int vstart = vindex;
     // Reusing ViewOfParArrays which expects 3D slices
     host_view(vindex++) = v->data.Get(0, 0, 0);
+
+    std::vector<int> shape;
+    if (v->GetDim(2) > 1) shape.push_back(v->GetDim(2));
+    if (v->GetDim(3) > 1) shape.push_back(v->GetDim(3));
+    if (v->GetDim(4) > 1) shape.push_back(v->GetDim(4));
+    if (v->GetDim(5) > 1) shape.push_back(v->GetDim(5));
+    if (v->GetDim(6) > 1) shape.push_back(v->GetDim(6));
```
pgrete: Why is the shape flexible here? As far as I can in other places it's hardocded to 6 with unused dims = 1.

pgrete (commented):
Overall looks good. I just left a coupe of comments (mostly asking for clarification/documentation).

pgrete:
> One thing I noticed while working on this is that the `particle_leapfrog` tests don't seem to be getting run by the CI. It's not obvious to me how exactly to fix this -- they seem to be enrolled in `tst/regression/CMakeLists.txt`?

They are not part of the "short" tests, but the full test suite covers those, see, e.g., https://gitlab.com/pgrete/parthenon/-/jobs/2175054435#L328
```
      Start 42: regression_mpi_test:particle_leapfrog
 1/10 Test #42: regression_mpi_test:particle_leapfrog ...........   Passed    0.63 sec
      Start 44: regression_mpi_test:particle_leapfrog_outflow
 2/10 Test #44: regression_mpi_test:particle_leapfrog_outflow ...   Passed    0.64 sec
      Start 46: regression_mpi_test:restart
```

Edit: If you think it's worth to add those tests also to the short suite so that some particles integration test is always included, you got my vote as the tests are short.

src/interface/metadata.cpp
brryan: Yes adding `Metadata::Particle` inside `AddSwarmValue()` is a good idea, done

example/particles/particles.cpp
brryan: Anyway removed following @pgrete's comment above

src/interface/swarm.cpp
brryan: Yes this is much better, thanks (although I do keep curly brackets around even one-line `for` loops, as a personal eccentricity)

docs/particles.md
brryan: Actually only up to `2D` at the moment, due to flat index packing only working for 3D and one dimension being reserved for the swarm index (Still an error, which I changed)

src/interface/swarm.cpp
brryan: Done

src/interface/swarm.cpp
brryan: Good point, thanks. `PackAllVariables` actually has a unique syntax really tailored just to filling buffers with all particle data for communication, so I've added inline comments describing the layout but also made it a private function. The `PackVariables` function is a traditional pack of `ParArrayND`s so I'll add some documentation to `particles.md` at least to say that it exists (It's not used in any examples right now).

src/interface/swarm.cpp
brryan: I wanted floating point division for `num_active`/`ncells`, especially for the case when `num_active << ncells`.

src/interface/swarm.cpp
brryan: Thanks :+1:

src/interface/swarm.hpp
brryan: I added the inline description of `pack_indices_shapes` to this function, let me know if I could do a better job describing it

src/interface/swarm.hpp
brryan: No reason, fixed.

src/interface/variable.hpp
brryan: Sorry I had an issue with the ordering of the data in `ParticleVariable` for the initializer list and thought this would appear in `FaceVariable`, but looking again at the code this was correct in the first place. I reverted this change.

src/interface/variable.hpp
brryan: Agreed, fixed.

src/interface/variable.hpp
brryan: It's an initializer list problem, for `ParticleVariable`s it makes more sense to initialize `data` after `dims_`. Does that seem like an OK change?

src/interface/variable_pack.hpp
brryan: Is this not consistent with e.g. `FillVarView` (line ~550)? Maybe I'm misunderstanding

brryan:
@pgrete to add the particle leapfrog tests to the `short` suite do I just add
```
ctest -R regression_mpi_test:particle_leapfrog
```
below the similar `output_hdf5` tests in `parthenon/gitlab-ci.yml`?

Otherwise assuming the CI passes and you are happy with my answers above I think this is ready to go

src/interface/swarm.cpp
```diff
@@ -906,12 +890,32 @@ int Swarm::CountParticlesToSend_() {
...
+    for (int n = 0; n <= max_active_index_; n++) {
+      if (mask_h(n)) {
+        if (blockIndex_h(n) == no_block_) {
+          noblock_indices_h(counter) = n;
+          counter++;
+        }
+      }
+    }
+    noblock_indices.DeepCopy(noblock_indices_h);
+    ApplyBoundaries_(total_noblock_particles, noblock_indices);
```
brryan, line 907: I had broken the `particle_bounds_outflow` test here because the addition of the `no_block_` value when getting a neighbor block index (which is what got outflow boundaries working in the first place) meant that a particle wouldn't be sent across meshblocks which was the only code path triggering `ApplyBoundaries`. Unfortunately this adds *another* deep copy to the particles, which I will go ahead and clean up once `par_scan` is in parthenon (see #674 for the issue I just opened)

brryan:
@pgrete This is ready for re-review

docs/particles.md
```diff
@@ -104,6 +108,14 @@ future.
 
 AMR is currently not supported, but support will be added in the future.
 
+## Variable Packing
+
+Similarly to grid variables, particle swarms support `ParticleVariable` packing, by the function
+`Swarm::PackVariables`.
+
+Note that this 1D or 2D ParticleVariables are currently untested with this function; in particular,
+`FlatIdx` is not guaranteed to work correctly.
```
pgrete: Not sure I understand this comment.
Does this refer to particles in 1D or 2D simulations or to 1D/2D variables as per "multi-D" particles variables?
Also what does "dimensionality" refer to here? Is a vector with multiple components a 1D variable (similar to our cell centered scalars having no additional entry in their shape variable)?
Some clarification would be great and, independently, some error check (e.g., debug check) in the code in case someone to access this functionality.

example/particle_leapfrog/particle_leapfrog.cpp
```diff
@@ -69,11 +69,11 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
...
   pkg->AddSwarm(swarm_name, swarm_metadata);
-  Metadata real_swarmvalue_metadata({Metadata::Real});
   pkg->AddSwarmValue("id", swarm_name, Metadata({Metadata::Integer}));
-  pkg->AddSwarmValue("vx", swarm_name, real_swarmvalue_metadata);
-  pkg->AddSwarmValue("vy", swarm_name, real_swarmvalue_metadata);
-  pkg->AddSwarmValue("vz", swarm_name, real_swarmvalue_metadata);
+  Metadata vreal_swarmvalue_metadata({Metadata::Real}, std::vector<int>{3});
+  pkg->AddSwarmValue("v", swarm_name, vreal_swarmvalue_metadata);
+  Metadata vvreal_swarmvalue_metadata({Metadata::Real}, std::vector<int>{2, 2});
+  pkg->AddSwarmValue("vv", swarm_name, vvreal_swarmvalue_metadata);
```
pgrete: Are the `"vv"` actually used somewhere?

example/particles/particles.cpp
```diff
@@ -301,21 +298,21 @@ TaskStatus CreateSomeParticles(MeshBlock *pmb, const double t0) {
...
+            v(2, n) = v(2, n) - vdN / NdN * z(n);
 
             // Normalize
-            Real v_tmp = sqrt(vx(n) * vx(n) + vy(n) * vy(n) + vz(n) * vz(n));
-            vx(n) *= v / v_tmp;
-            vy(n) *= v / v_tmp;
-            vz(n) *= v / v_tmp;
+            Real v_tmp = sqrt(v(0, n) * v(0, n) + v(1, n) * v(1, n) + v(2, n) * v(2, n));
+            for (int ii = 0; ii < 3; ii++) {
+              v(ii, n) *= vel / v_tmp;
```
pgrete, line 315: Should probably check for `v_tmp != 0`

example/particles/particles.cpp
```diff
@@ -402,17 +397,17 @@ TaskStatus TransportParticles(MeshBlock *pmb, const StagedIntegrator *integrator
     pmb->par_for(
         "TransportOrbitingParticles", 0, max_active_index, KOKKOS_LAMBDA(const int n) {
           if (swarm_d.IsActive(n)) {
-            Real v = sqrt(vx(n) * vx(n) + vy(n) * vy(n) + vz(n) * vz(n));
+            Real vel = sqrt(v(0, n) * v(0, n) + v(1, n) * v(1, n) + v(2, n) * v(2, n));
             while (t(n) < t0 + dt) {
-              Real dt_cell = dx_push / v;
+              Real dt_cell = dx_push / vel;
```
pgrete, line 404: Check for `vel != 0`

example/particles/particles.cpp
```diff
@@ -422,17 +417,18 @@ TaskStatus TransportParticles(MeshBlock *pmb, const StagedIntegrator *integrator
...
               // Project v onto plane normal to sphere
-              Real vdN = vx(n) * x(n) + vy(n) * y(n) + vz(n) * z(n);
+              Real vdN = v(0, n) * x(n) + v(1, n) * y(n) + v(2, n) * z(n);
               Real NdN = r * r;
-              vx(n) = vx(n) - vdN / NdN * x(n);
-              vy(n) = vy(n) - vdN / NdN * y(n);
-              vz(n) = vz(n) - vdN / NdN * z(n);
+              v(0, n) = v(0, n) - vdN / NdN * x(n);
+              v(1, n) = v(1, n) - vdN / NdN * y(n);
+              v(2, n) = v(2, n) - vdN / NdN * z(n);
```
pgrete, line 428: check `NdN != 0`

example/particles/particles.cpp
```diff
@@ -422,17 +417,18 @@ TaskStatus TransportParticles(MeshBlock *pmb, const StagedIntegrator *integrator
...
 
               // Normalize
-              Real v_tmp = sqrt(vx(n) * vx(n) + vy(n) * vy(n) + vz(n) * vz(n));
-              vx(n) *= v / v_tmp;
-              vy(n) *= v / v_tmp;
-              vz(n) *= v / v_tmp;
+              Real v_tmp =
+                  sqrt(v(0, n) * v(0, n) + v(1, n) * v(1, n) + v(2, n) * v(2, n));
+              for (int ii = 0; ii < 3; ii++) {
+                v(ii, n) *= vel / v_tmp;
```
pgrete, line 435: check `v_tmp != 0`

example/particles/particles.cpp
```diff
@@ -450,15 +446,15 @@ TaskStatus TransportParticles(MeshBlock *pmb, const StagedIntegrator *integrator
     pmb->par_for(
         "TransportParticles", 0, max_active_index, KOKKOS_LAMBDA(const int n) {
           if (swarm_d.IsActive(n)) {
-            Real v = sqrt(vx(n) * vx(n) + vy(n) * vy(n) + vz(n) * vz(n));
+            Real vel = sqrt(v(0, n) * v(0, n) + v(1, n) * v(1, n) + v(2, n) * v(2, n));
             while (t(n) < t0 + dt) {
-              Real dt_cell = dx_push / v;
+              Real dt_cell = dx_push / vel;
```
pgrete, line 457: check `vel != 0`

example/particles/particles.cpp
```diff
@@ -422,17 +417,18 @@ TaskStatus TransportParticles(MeshBlock *pmb, const StagedIntegrator *integrator
               z(n) *= r / r_tmp;
 
               // Project v onto plane normal to sphere
-              Real vdN = vx(n) * x(n) + vy(n) * y(n) + vz(n) * z(n);
```
pgrete, line 425: Above check `r_tmp != 0`

src/interface/metadata.hpp
```diff
@@ -325,14 +327,16 @@ class Metadata {
       return Edge;
     } else if (IsSet(Node)) {
       return Node;
+    } else if (IsSet(Particle)) {
+      return Particle;
```
pgrete: This may be notation nitpicking, but I wouldn't consider "Particle" to be a "topological location".
And is it even required to introduce a "particle" topological location or can we just work with Particles also having "none" type location and use the `Metadata::IsParticle` in places where `Where() == Particle` is being used right now?
I'd also like to hear other people thought on this, e.g.,  @Yurlungur @lroberts36 or @jdolence

src/interface/metadata.cpp
```diff
@@ -141,6 +141,13 @@ std::array<int, 6> Metadata::GetArrayDims(std::weak_ptr<MeshBlock> wpmb,
       arrDims[i + 3] = shape[i];
     for (int i = N; i < 3; i++)
       arrDims[i + 3] = 1;
+  } else if (Where() == Particle) {
```
pgrete: How about `IsSet(Particle)` instead (see more extended comment below).

src/interface/swarm.cpp
pgrete: Just leaving a comment after I now realized why I got confused here. In my mind `k`, `j`, and `i` are kind of closely tied to "mesh" variables, so I got thrown off.
If more people have that "mental map" it may be worth discussing some notation (this also, for example, applies to the higher dimensional variable names used for mesh variables).

src/interface/swarm.cpp
pgrete: General question: Why are all these host functions and not on the device?

src/interface/variable_pack.hpp
pgrete: You just updated the corresponding lines with PR #678 last week and I'm probably more confused now.
Shape is not necessarily tied to the `dims_` of a variable, isn't it?
Following https://github.com/lanl/parthenon/blob/develop/docs/interface/containers.md#meshblockdata a cell centered vector variable could have `shape = std::vector<int>({3});` but the dims of the actual variable would be different (with the first three being tied to the mesh and only the forth one tied to the vector component).

pgrete (commented):
I'm really sorry but I got a few more comments, but I promise a swift re-review after we discussed/resolved those.

Regarding the "division by zero" comments, feel free to ignore those. I just pointed them out as they may lead to some unintended (and hard to debug) behavior (that we have not yet encountered given the current problem setup).

src/interface/swarm.cpp
brryan: It wasn't obvious to me earlier how to do this in parallel in device code, but the answer is basically `parallel_scan` so once #667 goes in I will go back and rewrite all these host functions

example/particle_leapfrog/particle_leapfrog.cpp
brryan: This was only being used to see that such a swarm could be added. I now populate this with `vv(i,j,n) = v(i,n) * v(j,n)` and check for consistency during the log writing step.

src/interface/metadata.hpp
brryan: OK I see your point about topological location. And since as you point out below we can just use `IsSet(Particle)` in `metadata.cpp` this is unnecessary -- I've removed this

src/interface/metadata.cpp
brryan: Yep agreed 🤝

src/interface/variable_pack.hpp
brryan: Ah yes this needs to be updated for consistency with #678 (done). Yes, `shape` is not tied to the `dims_` of a variable, but the relationship between `shape` and `dims_` should be fixed based on what kind of variable it is. For a cell-centered variable, the first three indices are always the `i,j,k` spatial indices, leaving three more indices for `shape`, whereas for particle variable the first index is always the swarm index, leaving five more indices for `shape`. This is I think correctly captured by the different functions `FillSwarmVarView` and `FillVarView`. Maybe I am missing something though -- do you see a case that the current logic in `variable_pack.hpp` will fail for?

src/interface/swarm.cpp
brryan: Yes I agree, looking at this again `i, j, k` are unfortunate choices. I switched this to something hopefully clearer, `i -> vidx`, `j -> l`, `k -> m` (and switched the `l` and `m` loop ordering to match the index ordering in the arrays)

brryan:
> I'm really sorry but I got a few more comments, but I promise a swift re-review after we discussed/resolved those.
> 
> Regarding the "division by zero" comments, feel free to ignore those. I just pointed them out as they may lead to some unintended (and hard to debug) behavior (that we have not yet encountered given the current problem setup).

No worries @pgrete these are good comments that I need to address! I have one more issue to clean up, swarm variable packing indeed does not work with `FlatIdx`, but I think it is easier to just fix this than to add a check that `FlatIdx` is not being used with `SwarmVariablePack`s. Working on this now, hopefully it'll be done today. These changes should actually bring `SwarmVariablePack` behavior much more in line with the existing variable packs.

docs/particles.md
brryan: @pgrete OK I reworked the packing machinery and `FlatIdx` is now supported. Also, "multi-D" (in the sense of greater-than-scalar data per particle (like if each particle holds a `v[3]`)) data works now. This should actually work with up to 5D data per particle, but the particle_leapfrog example now includes a demo with packing of 2D data using the `vv` quantity that you noticed was unused earlier.

example/particle_leapfrog/particle_leapfrog.cpp
```diff
@@ -158,21 +158,44 @@ TaskStatus WriteParticleLog(BlockList_t &blocks, int ncycle) {
...
+        particle_output_this_rank(offset++) = v(2, n);
+
+        // Check that vv is still consistent with v
+        for (int i = 0; i < 3; i++) {
+          for (int j = 0; j < 3; j++) {
+            if (std::fabs(vv(i, j, n)) > 1.e-10) {
+              PARTHENON_REQUIRE(
+                  2. * std::fabs(vp(ivv(i, j), n) - vp(iv(i), n) * vp(iv(j), n)) <
+                      1.e-10 * (std::fabs(vp(ivv(i, j), n)) +
+                                std::fabs(vp(iv(i), n) * vp(iv(j), n))),
```
brryan: Here's the `FlatIdx`'d packed particle data being checked. After I migrated the swarm variable packing to be more consistent with the existing packing, `FlatIdx` "just worked" which was nice.

src/interface/swarm.cpp
```diff
@@ -437,23 +406,25 @@ void Swarm::Defrag() {
...
+  const int realPackDim = vreal.GetDim(2);
+  const int intPackDim = vint.GetDim(2);
 
   pmb->par_for(
       "Swarm::DefragVariables", 0, max_active_index_, KOKKOS_LAMBDA(const int n) {
         if (from_to_indices(n) >= 0) {
-          for (int i = 0; i < real_vars_size; i++) {
-            vreal(i, from_to_indices(n)) = vreal(i, n);
+          for (int vidx = 0; vidx < realPackDim; vidx++) {
+            vreal(vidx, from_to_indices(n)) = vreal(vidx, n);
```
brryan, line 424: This no longer renders in the diff, but standardizing the swarm packing cleaned up a lot of weirdness here (extra `ParArrayND`s to store the dimensions of each packed variable, basically).

src/interface/variable_pack.hpp
```diff
@@ -574,30 +585,42 @@ void FillVarView(const CellVariableVector<T> &vars, bool coarse,
...
+    std::vector<int> shape;
+    auto mshape = v->metadata().Shape();
+    if (mshape.size() > 0) shape.push_back(v->GetDim(2));
+    if (mshape.size() > 1) shape.push_back(v->GetDim(3));
+    if (mshape.size() > 2) shape.push_back(v->GetDim(4));
+    if (mshape.size() > 3) shape.push_back(v->GetDim(5));
+    if (mshape.size() > 4) shape.push_back(v->GetDim(6));
+
+    if (pvmap != nullptr) {
+      pvmap->insert(v->label(), IndexPair(vstart, vindex - 1), shape);
```
brryan, line 619: Here's the updated swarm variable packing. Each of the elements in the five non-swarm index dimensions now gets its own pack index, like each of the elements in the three non-spatial index dimensions for cell variable packs.

example/particle_leapfrog/particle_leapfrog.cpp
brryan: (oops this was calling device memory from host, I moved it to a separate `par_for` to make the GPU CI happy)

brryan:
Looks like there's some issue with the CI -- the latest commit `ee3a07f`'s pipeline passed on gitlab: https://re-git.lanl.gov/eap-oss/parthenon/-/commit/ee3a07fe22d3b3ad18766629d0af9e67d5ffa3b3/pipelines

pgrete:
> Looks like there's some issue with the CI -- the latest commit `ee3a07f`'s pipeline passed on gitlab: https://re-git.lanl.gov/eap-oss/parthenon/-/commit/ee3a07fe22d3b3ad18766629d0af9e67d5ffa3b3/pipelines

Does the problem still persist and did it only happen on the LANL internal CI? I cannot see the error message and the "standard" CI has passed now.

src/interface/variable_pack.hpp
pgrete: :+1: I think this is more consistent with the other variable packing.

example/particle_leapfrog/particle_leapfrog.cpp
```diff
@@ -172,8 +179,38 @@ TaskStatus WriteParticleLog(BlockList_t &blocks, int ncycle) {
...
         particle_output_this_rank(offset++) = v(2, n);
+
+        // Check that vv is still consistent with v
+        for (int i = 0; i < 3; i++) {
+          for (int j = 0; j < 3; j++) {
+            if (std::fabs(vv(i, j, n)) > 1.e-10) {
+              PARTHENON_REQUIRE(
+                  2. * std::fabs(vv(i, j, n) - v(i, n) * v(j, n)) <
+                      1.e-10 * (std::fabs(vv(i, j, n)) + std::fabs(v(i, n) * v(j, n))),
+                  "vv not consistent with v*v!");
```
pgrete: I like using the `vv` both as sanity check for multi-d swarm variables as well as for index testing!

Regarding the check, does a simple `vv = v * v` fail? Given that not that much is happening I would be surprised if the results are not bitwise identical.

brryan:
> > Looks like there's some issue with the CI -- the latest commit `ee3a07f`'s pipeline passed on gitlab: https://re-git.lanl.gov/eap-oss/parthenon/-/commit/ee3a07fe22d3b3ad18766629d0af9e67d5ffa3b3/pipelines
> 
> Does the problem still persist and did it only happen on the LANL internal CI? I cannot see the error message and the "standard" CI has passed now.

Yes the standard CI passed now, I just made one tiny change to restart the CI and now everything is happy.

src/interface/metadata.hpp
```diff
@@ -332,7 +334,7 @@ class Metadata {
     PARTHENON_THROW("No topology flag set");
   }
 
-  bool IsMeshTied() const { return Where() != None; }
+  bool IsMeshTied() const { return (Where() != None && Where() != Particle); }
```
pgrete: This can probably also be reverted given that Particle is "topologically" also `None` now

example/particle_leapfrog/particle_leapfrog.cpp
brryan: I think that would work, at least mostly, I just never completely trust `==` for floating point.

Maybe it would be worth adding a `soft_equiv` function in `utils/error_checking.hpp` that does the above comparison

pgrete: For a given binary on a given system this `==` should always hold given that we are not doing any other operations or changing the order of operations in between.

src/interface/metadata.hpp
brryan: Yes, thanks for catching this, done

example/particle_leapfrog/particle_leapfrog.cpp
brryan: OK, the only potential subtlety here is that `vv` and `v` are separately copied into and out of buffers for communication across meshblocks, but maybe that isn't relevant. I switched to `==`, if the CI passes I guess we can assume we're good.

example/particle_leapfrog/particle_leapfrog.cpp
```diff
@@ -158,23 +158,54 @@ TaskStatus WriteParticleLog(BlockList_t &blocks, int ncycle) {
...
-        particle_output_this_rank(offset++) = vy(n);
-        particle_output_this_rank(offset++) = vz(n);
+        particle_output_this_rank(offset++) = v(0, n);
+        particle_output_this_rank(offset++) = v(1, n);
+        particle_output_this_rank(offset++) = v(2, n);
+
+        // Check that vv is still consistent with v
+        for (int i = 0; i < 3; i++) {
+          for (int j = 0; j < 3; j++) {
+            if (std::fabs(vv(i, j, n)) > 1.e-10) {
```
pgrete: Why the check for `>1e-10`?

example/particle_leapfrog/particle_leapfrog.cpp
```diff
@@ -158,23 +158,54 @@ TaskStatus WriteParticleLog(BlockList_t &blocks, int ncycle) {
...
       }
     }
+
+    const auto is_active_d = swarm->GetMask();
+    pmb->par_for(
+        "CheckParticlePacks", 0, swarm->GetMaxActiveIndex(), KOKKOS_LAMBDA(const int n) {
+          if (is_active_d(n)) {
+            for (int i = 0; i < 3; i++) {
+              for (int j = 0; j < 3; j++) {
+                if (std::fabs(vp(ivv(i, j), n)) > 1.e-10) {
```
pgrete: As above, why the check for `>1e-10`?

pgrete (approved):
Looks good now! Sorry for the back and forth.
Feel free to squash and merge whenever.

example/particle_leapfrog/particle_leapfrog.cpp
brryan: Oops, this is a leftover from using a soft equivalence (which can get upset when one of the arguments is exactly `0`. Removed.

brryan:
> Looks good now! Sorry for the back and forth. Feel free to squash and merge whenever.

Thanks for the careful review @pgrete! Merging when CI is finished

PR #655, opened 2022-03-02
Enable user boundary conditions for particles

brryan:
## PR Summary

Particle swarms are intended to support user-provided boundary conditions, but there was a bit of connectivity missing between the user-side application inputs and the Swarms themselves; specifically, we were checking for correct particle boundaries before user boundaries could be set.

On an earlier attempt at making this change I noticed that `InitUserMeshData` is (I think) not very useful right now because application-specified functions can't access the `Mesh` variables. In particular, I need to be able to set `Swarm` boundary conditions to application-specific `user` values at this point in the mesh setup. I don't use `InitUserMeshData` here but I assume we want this change?

All this PR does is add the `this` pointer to the `Mesh` calls to `InitUserMeshData`. 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/docs/particles.md
+++ b/docs/particles.md
@@ -112,4 +112,5 @@ when they are in boundary regions are allocated depending on the boundary flags 
 in the input file. Currently, outflow and periodic boundaries are supported natively.
 User-specified boundary conditions must be set by specifying the "user" flag in the input
 parameter file and then updating the appropriate Swarm::bounds array entries to separately
-allocated boundary condition objects.
+allocated boundary condition objects. An example is given in the `particles` example when
+ix1 and ox1 are set to `user` in the input parameter file.
--- a/docs/particles.md
+++ b/docs/particles.md
@@ -111,6 +111,6 @@ classes containing boundary condition functions for updating particles or removi
 when they are in boundary regions are allocated depending on the boundary flags specified
 in the input file. Currently, outflow and periodic boundaries are supported natively.
 User-specified boundary conditions must be set by specifying the "user" flag in the input
-parameter file and then updating the appropriate Swarm::bounds array entries to separately
-allocated boundary condition objects. An example is given in the `particles` example when
-ix1 and ox1 are set to `user` in the input parameter file.
+parameter file and then updating the appropriate Swarm::bounds array entries to factory
+functions that allocate device-side boundary condition objects. An example is given in the
+`particles` example when ix1 and ox1 are set to `user` in the input parameter file.
--- a/example/particles/main.cpp
+++ b/example/particles/main.cpp
@@ -38,10 +38,14 @@ int main(int argc, char *argv[]) {
   if (pman.pinput->GetString("parthenon/mesh", "ix1_bc") == "user") {
     pman.app_input->boundary_conditions[parthenon::BoundaryFace::inner_x1] =
         parthenon::BoundaryFunction::OutflowInnerX1;
+    pman.app_input->swarm_boundary_conditions[parthenon::BoundaryFace::inner_x1] =
+      particles_example::SetSwarmIx1UserBC;
   }
   if (pman.pinput->GetString("parthenon/mesh", "ox1_bc") == "user") {
     pman.app_input->boundary_conditions[parthenon::BoundaryFace::outer_x1] =
         parthenon::BoundaryFunction::OutflowOuterX1;
+    pman.app_input->swarm_boundary_conditions[parthenon::BoundaryFace::outer_x1] =
+      particles_example::SetSwarmOx1UserBC;
   }
   pman.ParthenonInitPackagesAndMesh();
 
--- a/example/particles/main.cpp
+++ b/example/particles/main.cpp
@@ -20,12 +20,8 @@ int main(int argc, char *argv[]) {
   using parthenon::ParthenonStatus;
   ParthenonManager pman;
 
-  // Redefine parthenon defaults
-  pman.app_input->ProcessPackages = particles_example::ProcessPackages;
-  pman.app_input->ProblemGenerator = particles_example::ProblemGenerator;
-
   // call ParthenonInit to initialize MPI and Kokkos, parse the input deck, and set up
-  auto manager_status = pman.ParthenonInit(argc, argv);
+  auto manager_status = pman.ParthenonInitEnv(argc, argv);
   if (manager_status == ParthenonStatus::complete) {
     pman.ParthenonFinalize();
     return 0;
@@ -34,8 +30,20 @@ int main(int argc, char *argv[]) {
     pman.ParthenonFinalize();
     return 1;
   }
-  // Now that ParthenonInit has been called and setup succeeded, the code can now
-  // make use of MPI and Kokkos
+
+  // Redefine parthenon defaults
+  pman.app_input->ProcessPackages = particles_example::ProcessPackages;
+  pman.app_input->ProblemGenerator = particles_example::ProblemGenerator;
+  pman.app_input->InitUserMeshData = particles_example::InitUserMeshData;
+  if (pman.pinput->GetString("parthenon/mesh", "ix1_bc") == "user") {
+    pman.app_input->boundary_conditions[parthenon::BoundaryFace::inner_x1] =
+        parthenon::BoundaryFunction::OutflowInnerX1;
+  }
+  if (pman.pinput->GetString("parthenon/mesh", "ox1_bc") == "user") {
+    pman.app_input->boundary_conditions[parthenon::BoundaryFace::outer_x1] =
+        parthenon::BoundaryFunction::OutflowOuterX1;
+  }
+  pman.ParthenonInitPackagesAndMesh();
 
   // Initialize the driver
   particles_example::ParticleDriver driver(pman.pinput.get(), pman.app_input.get(),
--- a/example/particles/main.cpp
+++ b/example/particles/main.cpp
@@ -1,5 +1,5 @@
 //========================================================================================
-// (C) (or copyright) 2020. Triad National Security, LLC. All rights reserved.
+// (C) (or copyright) 2020-2022. Triad National Security, LLC. All rights reserved.
 //
 // This program was produced under U.S. Government contract 89233218CNA000001 for Los
 // Alamos National Laboratory (LANL), which is operated by Triad National Security, LLC
--- a/example/particles/main.cpp
+++ b/example/particles/main.cpp
@@ -35,12 +35,20 @@ int main(int argc, char *argv[]) {
   pman.app_input->ProcessPackages = particles_example::ProcessPackages;
   pman.app_input->ProblemGenerator = particles_example::ProblemGenerator;
   if (pman.pinput->GetString("parthenon/mesh", "ix1_bc") == "user") {
+    // In this case, we are setting a custom swarm boundary condition while still using
+    // a default parthenon boundary condition for cell variables. In general, one can
+    // provide both custom cell variable and swarm boundary conditions. However, to use
+    // custom boundary conditions for either cell variables or swarms, the parthenon
+    // boundary must be set to "user" and both cell variable and swarm boundaries provided
+    // as here.
     pman.app_input->boundary_conditions[parthenon::BoundaryFace::inner_x1] =
         parthenon::BoundaryFunction::OutflowInnerX1;
     pman.app_input->swarm_boundary_conditions[parthenon::BoundaryFace::inner_x1] =
         particles_example::SetSwarmIx1UserBC;
   }
   if (pman.pinput->GetString("parthenon/mesh", "ox1_bc") == "user") {
+    // Again, we use a default parthenon boundary condition for cell variables but a
+    // custom swarm boundary condition.
     pman.app_input->boundary_conditions[parthenon::BoundaryFace::outer_x1] =
         parthenon::BoundaryFunction::OutflowOuterX1;
     pman.app_input->swarm_boundary_conditions[parthenon::BoundaryFace::outer_x1] =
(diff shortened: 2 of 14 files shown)


docs/particles.md
```diff
@@ -112,4 +112,5 @@ when they are in boundary regions are allocated depending on the boundary flags
 in the input file. Currently, outflow and periodic boundaries are supported natively.
 User-specified boundary conditions must be set by specifying the "user" flag in the input
 parameter file and then updating the appropriate Swarm::bounds array entries to separately
-allocated boundary condition objects.
+allocated boundary condition objects. An example is given in the `particles` example when
+ix1 and ox1 are set to `user` in the input parameter file.
```
Yurlungur: This isn't really accurate right? It might be better to say that we set a factory function which generates the appropriately allocated boundary condition objects. Could that be made a little more explicit in the documentation?

example/particles/particles.cpp
```diff
@@ -25,6 +25,16 @@
...
+std::unique_ptr<ParticleBound, DeviceDeleter<parthenon::DevMemSpace>>
+SetSwarmIx1UserBC() {
+  return DeviceAllocate<ParticleBoundIX1Outflow>();
+}
+
+std::unique_ptr<ParticleBound, DeviceDeleter<parthenon::DevMemSpace>>
+SetSwarmOx1UserBC() {
+  return DeviceAllocate<ParticleBoundOX1Outflow>();
+}
+
```
Yurlungur: It might be nice to have the outflow boundaries separately defined in the example so that the user can see what these objects need to look like.

Yurlungur (approved):
A few minor documentation nitpicks. Otherwise looks good.

docs/particles.md
brryan: Yeah good point, especially since the application-provided cell-centered and particle boundaries are somewhat different (necessary due to how we implement particle boundary condition switching for performance). I'll update.

example/particles/particles.cpp
brryan: Yep done, thanks

src/mesh/mesh.cpp
```diff
@@ -960,6 +960,21 @@ void Mesh::EnrollBndryFncts_(ApplicationInput *app_in) {
...
     }
+
+    switch (mesh_bcs[f]) {
+    case BoundaryFlag::user:
+      if (app_in->swarm_boundary_conditions[f] != nullptr) {
+        SwarmBndryFnctn[f] = app_in->swarm_boundary_conditions[f];
+      } else {
+        std::stringstream msg;
+        msg << "A user boundary condition for face " << f
+            << " was requested, but not swarm condition was enrolled." << std::endl;
```
lroberts36: `not swarm...` -> `no swarm...`

src/interface/swarm.cpp
```diff
@@ -102,7 +102,14 @@ void Swarm::AllocateBoundaries() {
...
+    if (pmb->pmy_mesh->SwarmBndryFnctn[0] != nullptr) {
+      bounds_uptrs[0] = pmb->pmy_mesh->SwarmBndryFnctn[0]();
+    } else {
+      msg << "ix1 user boundary requested but provided function is null!";
+      PARTHENON_THROW(msg);
+    }
+  } else {
     msg << "ix1 boundary flag " << static_cast<int>(bcs[0]) << " not supported!";
     PARTHENON_THROW(msg);
   }
```
lroberts36: Could this be templated as another function call to reduce code duplication? 
```c++
template<class BOutflow, class BPeriodic, int iFace> 
void AllocateBoundariesImpl(MeshBlock* pmb) { 
  std::stringstream msg;
  auto &bcs = pmb->pmy_mesh->mesh_bcs;
  if (bcs[iFace] == BoundaryFlag::outflow) {
    bounds_uptrs[iFace] = DeviceAllocate<BOutflow>();
  } else if (bcs[iFace] == BoundaryFlag::periodic) {
    bounds_uptrs[iFace] = DeviceAllocate<BPeriodic>();
  } else if (bcs[iFace] == BoundaryFlag::user) {
    if (pmb->pmy_mesh->SwarmBndryFnctn[iFace] != nullptr) {
      bounds_uptrs[iFace] = pmb->pmy_mesh->SwarmBndryFnctn[iFace]();
    } else {
      msg << "ix" << iFace+1 << " user boundary requested but provided function is null!";
      PARTHENON_THROW(msg);
    }
  } else {
    msg << "ix" << iface+1 << " boundary flag " << static_cast<int>(bcs[iface]) << " not supported!";
    PARTHENON_THROW(msg);
  }
} 
```

src/interface/swarm.cpp
```diff
@@ -556,6 +598,10 @@ void Swarm::SortParticlesByCell() {
 
           if (cellSorted(start_index).cell_idx_1d_ >= cell_idx_1d) {
             start_index--;
+            if (start_index < 0) {
+              start_index = -1;
+              break;
+            }
```
lroberts36, line 544: Are these changes related to setting the user boundary conditions or is this a separate fix?

src/mesh/mesh.cpp
```diff
@@ -960,6 +960,21 @@ void Mesh::EnrollBndryFncts_(ApplicationInput *app_in) {
     default: // periodic/block BCs handled elsewhere.
       break;
     }
+
+    switch (mesh_bcs[f]) {
```
lroberts36, line 964: I guess this implies that the swarm boundary conditions have to be the same as the boundary conditions for cell centered variables? (unless you specify `user` I guess)

lroberts36 (approved):
LGTM. None of my comments are blocking.

example/particles/main.cpp
```diff
@@ -34,8 +30,23 @@ int main(int argc, char *argv[]) {
...
   }
-  // Now that ParthenonInit has been called and setup succeeded, the code can now
-  // make use of MPI and Kokkos
+
+  // Redefine parthenon defaults
+  pman.app_input->ProcessPackages = particles_example::ProcessPackages;
+  pman.app_input->ProblemGenerator = particles_example::ProblemGenerator;
+  if (pman.pinput->GetString("parthenon/mesh", "ix1_bc") == "user") {
+    pman.app_input->boundary_conditions[parthenon::BoundaryFace::inner_x1] =
+        parthenon::BoundaryFunction::OutflowInnerX1;
```
pgrete: How about adding a comment here that setting the (normal)  boundary conditions like this is just done here to make the hardcoded example work and is not considered best practice (because this way the info in the parameter file and outputs does not reflect what's actually happening in the code and boundary conditions are not aligned, i.e., mixing standard and user).

example/particles/main.cpp
```diff
@@ -34,8 +30,23 @@ int main(int argc, char *argv[]) {
...
+    pman.app_input->swarm_boundary_conditions[parthenon::BoundaryFace::inner_x1] =
+        particles_example::SetSwarmIx1UserBC;
+  }
+  if (pman.pinput->GetString("parthenon/mesh", "ox1_bc") == "user") {
+    pman.app_input->boundary_conditions[parthenon::BoundaryFace::outer_x1] =
+        parthenon::BoundaryFunction::OutflowOuterX1;
```
pgrete, line 53: Same as above

example/particles/particles.cpp
```diff
@@ -25,6 +25,16 @@
 // *************************************************//
 namespace particles_example {
 
+std::unique_ptr<ParticleBound, DeviceDeleter<parthenon::DevMemSpace>>
+SetSwarmIx1UserBC() {
+  return DeviceAllocate<ParticleBoundIX1User>();
+}
```
pgrete: Nitpicking: why the mismatch between SetSwarmI**x**1User and ParticleBoundI**X**User?

src/application_input.hpp
```diff
@@ -33,7 +33,7 @@ struct ApplicationInput {
   std::function<Packages_t(std::unique_ptr<ParameterInput> &)> ProcessPackages = nullptr;
 
   // Mesh functions
-  std::function<void(ParameterInput *)> InitUserMeshData = nullptr;
+  std::function<void(Mesh *, ParameterInput *)> InitUserMeshData = nullptr;
```
pgrete, line 36: Is this change still required?

src/mesh/mesh.cpp
pgrete: I think that make sense.
Either all variables/swarms use some provided default option, or, user type boundaries conditions have to be implemented for all types.
This would prevent silent bugs where boundaries conditions are not aligned.

pgrete (changes requested):
In addition to the comments I left, how about modifying/adding a test case where this infrastructure is called?

src/interface/swarm.cpp
brryan: This is a separate fix that came up while I was testing this PR -- for some pools of particles (in particular when using very few particles) one could end up trying to index into an array with `start_index` where `start_index` was `-1` which would segfault.

src/mesh/mesh.cpp
brryan: Yes I think it is substantially more likely that this would lead to bugs than an application wanting to, say, have outflow cell-centered variable boundaries but periodic particle boundaries. And even if an application does want that, they can just have `user` parthenon BCs and lightweight wrappers around the respective outflow and periodic BC functions.

example/particles/main.cpp
brryan: I actually think this is a reasonable use case -- one could imagine wanting outflow boundary conditions for a simulation, but the default parthenon swarm boundary outflow conditions of just deleting the particles as they leave not being suitable. A better particle boundary condition is often having a particle enter the simulation to match each particle leaving the simulation, assuming some equilibrium distribution between particles and cell variables at the boundary. I did add a comment here though going into this in some more detail -- let me know if you don't think that's sufficient.

example/particles/main.cpp
brryan: Some comment added here as well :+1:

src/application_input.hpp
brryan: This isn't a required change, but from my earlier discarded attempt at this PR I was left with the impression that `InitUserMeshData` basically isn't useful unless we pass the `Mesh` pointer in so I figured I would leave this change in case this function is needed in the future -- is that reasonable?

example/particles/particles.cpp
brryan: Fixed (to `X` from `x`) :+1:

src/interface/swarm.cpp
brryan: Works great, thanks!

brryan:
> In addition to the comments I left, how about modifying/adding a test case where this infrastructure is called?

@pgrete I added an `AllocateBoundaries` call to the `Swarm` that includes a custom `IX1` boundary

example/particles/main.cpp
pgrete: I see. I looked at it from a point of view where normal variables and swarms were treated separately (and therefore overwriting the normal variables boundary conditions felt odd).
Now, I understand that the "`user`" part actually applies to variables and swarms and in the example the "`user`" condition just reused one of the default implementations by chance -- which makes sense.
Thanks for adding the comment.

src/application_input.hpp
pgrete: Agreed, this changes makes sense so let's just keep it here.

src/interface/swarm.cpp
```diff
@@ -92,65 +92,47 @@ Swarm::Swarm(const std::string &label, const Metadata &metadata, const int nmax_
...
+  AllocateBoundariesImpl_<ParticleBoundOX1Outflow, ParticleBoundOX1Periodic, 1>(
+      pmb.get());
+  AllocateBoundariesImpl_<ParticleBoundIX2Outflow, ParticleBoundIX2Periodic, 2>(
+      pmb.get());
+  AllocateBoundariesImpl_<ParticleBoundOX2Outflow, ParticleBoundOX2Periodic, 3>(
+      pmb.get());
+  AllocateBoundariesImpl_<ParticleBoundIX3Outflow, ParticleBoundIX3Periodic, 4>(
+      pmb.get());
+  AllocateBoundariesImpl_<ParticleBoundOX3Outflow, ParticleBoundOX3Periodic, 5>(
+      pmb.get());
```
pgrete, line 135: I like the new code deduplication!
And now I'm actually wondering if we could actually get away even  without passing the two "classes/typenames" by also templating those to the `int` (or eventually use C++17 `constexpr if` inside the `impl_`).
If you feel like this is a good idea now, then it'd great if you could make this interface change.
Otherwise, I suggest to open an issue with "good first issue" and "refactor" tag and defer the change to a later time.

tst/unit/test_swarm.cpp
```diff
@@ -61,10 +82,16 @@ TEST_CASE("Swarm memory management", "[Swarm]") {
...
+  mesh->SwarmBndryFnctn[0] = SetSwarmIX1UserBC;
+  for (int i = 1; i < 6; i++) {
+    mesh->mesh_bcs[i] = BoundaryFlag::outflow;
+  }
   meshblock->pmy_mesh = mesh.get();
   Metadata m;
   auto swarm = std::make_shared<Swarm>("test swarm", m, NUMINIT);
   swarm->SetBlockPointer(meshblock);
+  swarm->AllocateBoundaries();
   auto swarm_d = swarm->GetDeviceContext();
```
pgrete, line 95: This change here, just checks that the user boundaries can be allocated, correct?
Otherwise, I'm not seeing where the boundary function actually becomes active, i.e., the test case where a particles leaves the domain in this unit test.

pgrete (approved):
Thanks for the updates.
I'm already approving pending the potential changes pertaining to the boundary allocation.

tst/unit/test_swarm.cpp
brryan: Yes that's right. I was leveraging the `particles` example (which isn't even part of the CI) but that defaults to periodic BCs anyway. I just updated the swarm unit test to move a particle below the `IX1` boundary and use the custom boundary condition to mark it for removal and then remove it and then make sure it was removed.

src/interface/swarm.cpp
brryan: Kudos to @lroberts36 

I think this would be a great place for `constexpr if` in the future, I'll open the issue... and done (#655)

PR #586, opened 2021-09-03
Implement true sparse variables

jlippuner:
## PR Summary

Implements true sparse variables that are not allocated on every block. This is currently working with and without AMR on a single rank. It is not implemented for MPI yet. It's only implemented with the "pack-in-one" communication methods and probably won't work with the legacy communication. To this end, I've turned `PARTHENON_ENABLE_INIT_PACKING` to ON by default. We should probably remove this option and code path altogether as part of this PR.

The current true sparse implementation (without MPI) works as follows.

1. The `BoundaryVariable` abstract base class now has an additional field `std::array<bool, NMAX_NEIGHBORS> neighbor_allocated`, which tracks if this particular variable is allocated on a given neighbor block. This is used to determine if boundary buffers should be sent to local neighboring blocks on the same rank. For neighbors on another rank, this will probably not be used (since it would require additional MPI communication to keep it updated), but that's TBD (so maybe the field should be renamed to `local_neighbor_allocated`).
2. In `TaskStatus MeshBlockData<T>::StartReceiving(BoundaryCommSubset phase)`, which gets called each cycle at the beginning of the boundary exchange, we update `neighbor_allocated` for local neighbors.
3. In `void ResetSendBufferBoundaryInfo(MeshData<Real> *md, size_t buffers_used)` we typically set the send buffer for a variable directly to the receive buffer for this variable on the neighboring block if the neighbor is local. Now we only do this if the neighbor is local and has the variable allocated, otherwise we use the send buffer as if this neighbor was remote.
4. In `TaskStatus SendBoundaryBuffers(std::shared_ptr<MeshData<Real>> &md)` where the send buffers are being filled in with the data to be communicated, we check if any of the values in a particular buffer is non-zero and we record this in a flag for each buffer.
5. In `void SendAndNotify(MeshData<Real> *md)` we didn't do anything for local neighbors, because their associated send buffers were pointing directly at the receive buffers on the target block. Now this is not the case if the target block does not have the sparse variable allocated. So if we have non-zero values in the send buffer and the local target block does not have the variable allocated, we allocate this sparse variable on the target block, copy the send buffer to the newly allocated receive buffer, and set all the appropriate flags for the boundary communication. We also update the corresponding `neighbor_allocated` flag for the boundary variable that is being communicated and now newly allocated on that neighbor. Note that the sparse variable gets allocated on all the stages of the target block.
[...]

PR #656, opened 2022-03-03
Extend CC bvars to 5-, 6-D ParArrays

brryan:
## PR Summary

#586 required CC boundary variables to be 1D i.e. `nx5 = nx6 = 1`. This PR lifts that restriction, mainly by extending the array flattening/unflattening used in communication buffers (flux corrections being limited to 1D actually seems to predate #586).

I'm not an expert in this part of the code so criticism of my changes is welcome!

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Update `bvals_cc_in_one.cpp` for `GetDim(5)`, `GetDim(6)` awareness
- [x] Downstream code runs


Yurlungur (commented):
Wait... I'm pretty sure bvals and flux correction *do* support non-vector valued variables... Haven't we been using them? I thought the bvals in one stuff was all that would need to change?

brryan:
My understanding is the previous `CellCenteredBoundaryVariable` constructor, which didn't know what values `GetDim(5)` and `GetDim(6)` returned, meant that nothing higher than a vector was supported. What code are you thinking of that we're currently using? (The current `phoebus/develop` doesn't run at least since #586 went into `parthenon` for the above reason; it throws on 
```c++
   PARTHENON_REQUIRE_THROWS(
        GetDim(4) == NumComponents(),
        "CellCenteredBoundaryVariable currently only supports rank-1 variables");
    vbvar = std::make_shared<CellCenteredBoundaryVariable>(pmb, IsSparse(), label(),
                                                           GetDim(4));
```

Yurlungur:
Huh. Ok. I was thinking of `bvals/cc_in_one.cpp`, which I thought made some assumptions about variable shape when computing boundary conditions. Not sure of the lines now.

brryan:
In `bvals/cc_in_one.cpp` it looks like we're currently limited to vector data (e.g. only `GetDim(4)` is called). I can update that as well in this PR.

src/bvals/cc/bvals_cc.cpp
```diff
@@ -46,16 +46,14 @@ namespace parthenon {
...
                                                            bool is_sparse,
                                                            const std::string &label,
-                                                           int dim4)
-    : BoundaryVariable(pmb, is_sparse, label), nl_(0), nu_(dim4 - 1) {
-  // CellCenteredBoundaryVariable should only be used w/ 4D or 3D (nx4=1) ParArrayND
-  // For now, assume that full span of 4th dim of input ParArrayND should be used:
-  // ---> get the index limits directly from the input ParArrayND
-  // <=nu_ (inclusive), <nx4 (exclusive)
-  if (nu_ < 0) {
+                                                           int dim4, int dim5, int dim6)
```
lroberts36, line 49: Should `dim5` and `dim6` have default arguments of 1 for backward compatibility? Or is that likely to introduce other issues?

src/bvals/cc/bvals_cc.hpp
```diff
@@ -37,7 +37,7 @@ namespace parthenon {
 class CellCenteredBoundaryVariable : public BoundaryVariable {
  public:
   CellCenteredBoundaryVariable(std::weak_ptr<MeshBlock> pmb, bool is_sparse,
-                               const std::string &label, int dim4);
+                               const std::string &label, int dim4, int dim5, int dim6);
```
lroberts36, line 40: Once again, should these have default arguments?

src/bvals/cc/bvals_cc.hpp
```diff
@@ -75,7 +75,7 @@ class CellCenteredBoundaryVariable : public BoundaryVariable {
   bool ReceiveFluxCorrection(bool is_allocated) final;
 
  protected:
-  int nl_, nu_;
+  int ll_, lu_, ml_, mu_, nl_, nu_;
```
lroberts36, line 78: It looks like above the size of each dimension is assumed to be `nu_ + 1`, etc. Is `nl_`, etc. supposed to note the lower bound of each dimension? Should the sizes be `nu_ - nl_ + 1` or is something else going on? I assume things are correct as they are now, but it would be good to have a comment here about what these variables denote.

lroberts36 (commented):
If I understand correctly, `bvals_in_one` is currently limited to 4-D ParArrays (see [bvals_in_one.cpp](https://github.com/lanl/parthenon/blob/6b8d3fab8f1b333280614a08dbeb14680fc21de1/src/bvals/cc/bvals_cc_in_one.cpp#L333). I think fixing this should probably be included in this pull request as well. (nm, I didn't see @brryan comment above while going through the review)

src/bvals/cc/bvals_cc.cpp
brryan: As far as I'm aware this call isn't part of the API so there shouldn't be any need for backward compatibility... pinging @Yurlungur  and @pgrete in case I'm mistaken though

src/bvals/cc/bvals_cc_in_one.cpp
```diff
@@ -555,18 +563,28 @@ TaskStatus SendBoundaryBuffers(std::shared_ptr<MeshData<Real>> &md) {
 
         Kokkos::parallel_for(
             Kokkos::TeamThreadRange<>(team_member, NvNkNj), [&](const int idx) {
```
brryan: `NtNuNvNkNj`

src/bvals/cc/bvals_cc.hpp
brryan: Good point -- `nsize = nu_ - nl_` wasn't originally used in `flux_correction` but now it is, and it should be `nu_ - nl_ + 1` like you say.

src/mesh/refinement_cc_in_one.cpp
```diff
@@ -23,7 +23,75 @@
...
+                                  Kokkos::ALL(), Kokkos::ALL(), Kokkos::ALL());
+              b++;
+            }
+          }
+          bb++;
+        }
+      }
+    }
+  }
+  Kokkos::deep_copy(refine_info, refine_info_h);
```
brryan: I'm assuming this (and line 31) is the extra host-device copy that the testing caught. We're basically rewriting this `comm_info` array of buffers into a similar struct (`refine_info`; see `bvals/cc/bvals_cc_in_one.hpp` for the struct definitions) but with 4D rather than 6D arrays (each `(t, u, :, :, :, :)` is its own buffer in the `Restrict` function at line `94`). An alternative would be to just use one `BndInfo` struct with 6D arrays but loop over `t` and `u` explicitly in the subsequent `Restrict` function (or just loop over the buffers where each one is the full `(:,:,:,:,:,:)` array). Would something like that be better? @Yurlungur @pgrete 

EDIT: Note `nbuffers = fine.GetDim(6) * fine.GetDim(5)` in `mesh/mesh_refinement.cpp:77`

Yurlungur: I'm not sure I understand what you're suggesting. Maybe you can walk me through it tomorrow?

brryan: From our discussion, `nbuffers = 1` here now and I reverted to one `BndInfo` struct with 6D data

brryan: Actually now the `par_for_outer`/`par_for_inner` logic in `mesh/refinement_cc_in_one.cpp` is trivial -- do you anticipate `nbuffers` somehow not being `1` in the future here, or should I just make this a regular `par_for` (or still have an outer loop over `N6/N5`)? @Yurlungur

brryan:
OK no longer copying boundary info between two different structs solved the extra copies issue the CI was upset about before. Apart from the above question about a trivial `par_for_outer/par_for_inner` I think this is ready to go if @lroberts36 you are happy about the no-default-arguments and `nu - nl + 1` changes.

src/bvals/cc/bvals_cc.hpp
lroberts36: It looks like it is still `nsize = nu_ - nl_`, etc. in flux_correction_cc.cpp.

src/bvals/cc/bvals_cc.cpp
lroberts36: ok, sounds good.

src/bvals/cc/bvals_cc.hpp
brryan: In `flux_correction_cc.cpp` right now (on github and locally) I'm seeing 
```
 95       int msize = mu_ - ml_ + 1;                                                                     
 96       int nsize = nu_ - nl_ + 1;
 ```
 and 
 ```
357       int msize = mu_ - ml_ + 1;                                                                     
358       int nsize = nu_ - nl_ + 1; 
```
-- is github having an issue with rendering the latest changeset, or am I missing another `nu_ - nl_`?

lroberts36: Looks good now, not sure what was going on before.

src/mesh/refinement_cc_in_one.cpp
Yurlungur: I think `nbuffers` might be not 1 in the future... but let's just leave it for now. Maybe make a comment to remind ourselves?

Yurlungur (approved):
LGTM now.

lroberts36 (approved):
LGTM

Issue #659, opened 2022-03-24
Regression test fail on Spock/Crusher/Frontier/HIP


pgrete:
```
$ ctest -L regression -LE "mpi-no|perf-reg" --output-on-failure
Test project /gpfs/alpine/ast146/world-shared/parth/build-hip-mpi
      Start 42: regression_mpi_test:particle_leapfrog
 1/10 Test #42: regression_mpi_test:particle_leapfrog ...........   Passed    1.30 sec
      Start 44: regression_mpi_test:particle_leapfrog_outflow
 2/10 Test #44: regression_mpi_test:particle_leapfrog_outflow ...   Passed    0.97 sec
      Start 46: regression_mpi_test:restart
 3/10 Test #46: regression_mpi_test:restart .....................   Passed   32.23 sec
      Start 48: regression_mpi_test:calculate_pi
 4/10 Test #48: regression_mpi_test:calculate_pi ................   Passed    0.94 sec
      Start 50: regression_mpi_test:advection_convergence
 5/10 Test #50: regression_mpi_test:advection_convergence .......   Passed   49.83 sec
      Start 52: regression_mpi_test:output_hdf5
 6/10 Test #52: regression_mpi_test:output_hdf5 .................***Failed   22.07 sec


test_dir=['/ccs/home/pgrete/src/parthenon/tst/regression/test_suites/output_hdf5']
output_dir='/gpfs/alpine/ast146/world-shared/parth/build-hip-mpi/tst/regression/outputs/output_hdf5_mpi'
driver=['/gpfs/alpine/ast146/world-shared/parth/build-hip-mpi/example/advection/advection-example']
driver_input=['/ccs/home/pgrete/src/parthenon/tst/regression/test_suites/output_hdf5/parthinput.advection']
kokkos_args=['--kokkos-num-devices=4']
num_steps=4
mpirun=['srun']
mpirun_ranks_flag='-n'
mpirun_ranks_num=4
mpirun_opts=[]
coverage=False
analyze=False
sparse_disabled=False
*****************************************************************
Beginning Python regression testing script
*****************************************************************

Initializing Test Case
Using:
driver at:       /gpfs/alpine/ast146/world-shared/parth/build-hip-mpi/example/advection/advection-example
driver input at: /ccs/home/pgrete/src/parthenon/tst/regression/test_suites/output_hdf5/parthinput.advection
[...]

PR #661, opened 2022-03-25
Fix MPI tag creation

### LINKED ISSUE #660: Are we properly incrementing `phys_id`?
I dug deeper into the issue that was raised during yesterday's call.

Here's what I found so far:
In the original Athena++ implementation, the physical ids for "standard" variables (e.g., hydro or b-fields) were implicitly updated in the constructor of various "physics", e.g., in `meshblock.cpp`
```c++
    phydro = new Hydro(this, pin);
    pbval->AdvanceCounterPhysID(HydroBoundaryVariable::max_phys_id);
  if (MAGNETIC_FIELDS_ENABLED) {
    pfield = new Field(this, pin);
    pbval->AdvanceCounterPhysID(FaceCenteredBoundaryVariable::max_phys_id);
  }
```
the `Hydro` and `Field` constructors would use the *current* `next_phys_id` (private var of the `MeshBlock`) for setting up the boundary variables.
The ids are *not* updated when they are assigned to the bvars, but afterwards manually incremented (by some const. max value).

*We don't call `AdvanceCounterPhysID` anywhere...

At the `Mesh` level Athena++ also has (and calls) `ReserveTagPhysIDs` in order to set a mesh level id (mesh private `next_phys_id`) so that other, extra  physics that need boundary comm start with a proper offset for the tags.

Given that from a Parthenon point of view everything is "extra" (as in separate packages), we don't need two separate functions, but need to ensure that we always advance the id for all kind of `Variables`.

Some dirty debugging in Parthenon with 
```diff
diff --git a/src/bvals/cc/bvals_cc.cpp b/src/bvals/cc/bvals_cc.cpp
index a22b0ed0..04bfc526 100644
--- a/src/bvals/cc/bvals_cc.cpp
+++ b/src/bvals/cc/bvals_cc.cpp
@@ -66,6 +66,7 @@ CellCenteredBoundaryVariable::CellCenteredBoundaryVariable(std::weak_ptr<MeshBlo
   // KGF: dead code, leaving for now:
   // cc_phys_id_ = pmb->pbval->ReserveTagVariableIDs(1);
   cc_phys_id_ = pmb.lock()->pbval->bvars_next_phys_id_;
+  std::cout << label << " uses id " << cc_phys_id_ << std::endl;
 #endif
   if (pmy_mesh_->multilevel) { // SMR or AMR
     InitBoundaryData(bd_var_flcor_, BoundaryQuantity::cc_flcor);
```
with the advection example and `Advection/num_vars=2`:
```
advected uses id 1
advected_1 uses id 1
advected uses id 1
advected_1 uses id 1
advected uses id 1
advected_1 uses id 1
...
```
not great.

I assume that this hasn't caused any bugs because the packing machinery has mostly been used with a single pack per rank and everything was well ordered (so even if two messages were send with the same tag for two different variables the receiving function were also in the same order).
This would also be connected to the intermittent MPI errors that Jonas saw when he implemented sparse variables.

A first fix 
```diff
diff --git a/src/bvals/cc/bvals_cc.cpp b/src/bvals/cc/bvals_cc.cpp
index a22b0ed0..491e6e79 100644
--- a/src/bvals/cc/bvals_cc.cpp
+++ b/src/bvals/cc/bvals_cc.cpp
@@ -63,14 +63,14 @@ CellCenteredBoundaryVariable::CellCenteredBoundaryVariable(std::weak_ptr<MeshBlo
   InitBoundaryData(bd_var_, BoundaryQuantity::cc);
 
 #ifdef MPI_PARALLEL
-  // KGF: dead code, leaving for now:
-  // cc_phys_id_ = pmb->pbval->ReserveTagVariableIDs(1);
-  cc_phys_id_ = pmb.lock()->pbval->bvars_next_phys_id_;
+  cc_phys_id_ = pmb.lock()->pbval->AdvanceCounterPhysID(1);
+  std::cout << label << " uses id " << cc_phys_id_ << std::endl;
 #endif
   if (pmy_mesh_->multilevel) { // SMR or AMR
     InitBoundaryData(bd_var_flcor_, BoundaryQuantity::cc_flcor);
 #ifdef MPI_PARALLEL
-    cc_flx_phys_id_ = cc_phys_id_ + 1;
+    cc_flx_phys_id_ = pmb.lock()->pbval->AdvanceCounterPhysID(1);
+    std::cout << label << " flx uses id " << cc_flx_phys_id_ << std::endl;
 #endif
   }
 }
```
resulting in some interesting behavior:
```
advected_1 uses id 1
advected_1 flx uses id 2
advected uses id 3
advected flx uses id 4
advected_1 uses id 1
advected_1 flx uses id 2
advected uses id 3
advected flx uses id 4
cycle=3 time=5.2734375000000003e-03 dt=1.7578125000000000e-03 zone-cycles/wsec_step=5.02e+06 wsec_total=2.20e-02 wsec_step=4.49e-03 zone-cycles/wsec=3.50e+06 wsec_AMR=1.96e-03
Root grid = 4 x 4 x 1 MeshBlocks
Total number of MeshBlocks = 94
Number of physical refinement levels = 2
Number of logical  refinement levels = 4
  Physical level = 0 (logical level = 2): 3 MeshBlocks, cost = 3
  Physical level = 1 (logical level = 3): 39 MeshBlocks, cost = 39
  Physical level = 2 (logical level = 4): 52 MeshBlocks, cost = 52
advected_1 uses id 5
advected_1 flx uses id 6
advected uses id 7
advected flx uses id 8
advected_1 uses id 9
advected_1 flx uses id 10
advected uses id 11
advected flx uses id 12
```
I don't understand yet where the ids >4 are coming from.

pgrete:
## PR Summary

Fixes #660 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -95,7 +95,11 @@ parthenon-cuda-short:
       ../
     - make -j${J} advection-example
     - export OMPI_MCA_mpi_common_cuda_event_max=1000
+    # Run one example with variables
     - ctest -R regression_mpi_test:output_hdf5
+    # and one that uses swarms
+    - make -j${J} particle-leapfrog
+    - ctest -R regression_mpi_test:particle_leapfrog
     # Now testing if there are no hidden memcopies between host and device.
     # Using a static grid (i.e., not AMR) as additional transfers are expected
     # during loadbalance and refinement, but not for a static grid. We also need to
@@ -126,8 +130,12 @@ parthenon-cpu-short:
     - cmake -DCMAKE_BUILD_TYPE=Release
       -DMACHINE_VARIANT=mpi
       ../
+    # Run one example with variables
     - make -j${J} advection-example
     - ctest -R regression_mpi_test:output_hdf5
+    # and one that uses swarms
+    - make -j${J} particle-leapfrog
+    - ctest -R regression_mpi_test:particle_leapfrog
   artifacts:
     when: always
     expire_in: 3 days
--- a/docs/interface/state.md
+++ b/docs/interface/state.md
@@ -140,5 +140,8 @@ The overload `Add(const std::string &label, MeshBlockData<T> &src, const std::ve
 
 Two simple examples of usage of these new containers are 1) to provide storage for multistage integration schemes and 2) to provide a mechanism to allocate storage for right hand sides, deltas, etc.  Both of these usages are demonstrated in the advection example that ships with Parthenon.
 
-Note that in multistage integrator the fluxes and `bvars` of a variable
+Note that in multistage integrator the fluxes and `bvars` (and their MPI communicator) of a variable
 are shared by default across all stages.
+This means that any kind of communication (most prominently flux correction and ghost zone
+exchange) of a given variable at a given stage should not be interleaved with any other
+modifications/communication of said variable as it may result in undefined behavior.
--- a/src/bvals/bvals_base.cpp
+++ b/src/bvals/bvals_base.cpp
@@ -31,6 +31,7 @@
 #include "globals.hpp"
 #include "mesh/mesh.hpp"
 #include "utils/buffer_utils.hpp"
+#include "utils/error_checking.hpp"
 
 namespace parthenon {
 
@@ -272,22 +273,23 @@ int BoundaryBase::FindBufferID(int ox1, int ox2, int ox3, int fi1, int fi2) {
 }
 
 //----------------------------------------------------------------------------------------
-//! \fn int BoundaryBase::CreateBvalsMPITag(int lid, int bufid, int phys)
+//! \fn int BoundaryBase::CreateBvalsMPITag(int lid, int bufid)
 //  \brief calculate an MPI tag for Bval communications
-//  MPI tag = local id of destination (remaining bits) + bufid(6 bits) + physics(5 bits)
+//  MPI tag = local id of destination (remaining bits) + bufid(6 bits)
 
 // WARN: The below procedure of generating unsigned integer bitfields from signed integer
 // types and converting output to signed integer tags (required by MPI) is tricky and may
 // lead to unsafe conversions (and overflows from built-in types and MPI_TAG_UB).  Note,
 // the MPI standard requires signed int tag, with MPI_TAG_UB>= 2^15-1 = 32,767 (inclusive)
 
-// TODO(felker) Consider adding safety check: if (tag > MPI_TAG_UB) PARTHENON_FAIL();
-// TODO(felker) Consider adding safety check: signed int inputs & outputs are positive
-// TODO(felker) Store # of bits for each bitfield component in preprocessor macros
-//              TAG_BITS_PHYS=5, MAX_NUM_PHYS=31
-
-int BoundaryBase::CreateBvalsMPITag(int lid, int bufid, int phys) {
-  return (lid << 11) | (bufid << 5) | phys;
+int BoundaryBase::CreateBvalsMPITag(int lid, int bufid) {
+  PARTHENON_REQUIRE_THROWS(lid >= 0 && bufid >= 0, "Ids are expected to be positive.")
+  int tag = (lid << 6) | bufid;
+  // Technically we could query MPI_TAG_UB here. However, using a fixed (lower) bound
+  // guaranteed by the standard for now to reduce the additional functions calls or
+  // adding a new internal upper lim variable.
+  PARTHENON_REQUIRE_THROWS(tag >= 0 && tag < 32768, "MPI tag out of bounds.")
+  return tag;
 }
 
 //----------------------------------------------------------------------------------------
--- a/src/bvals/bvals_swarm.cpp
+++ b/src/bvals/bvals_swarm.cpp
@@ -41,7 +41,7 @@ namespace parthenon {
 BoundarySwarm::BoundarySwarm(std::weak_ptr<MeshBlock> pmb)
     : bswarm_index(), pmy_block(pmb), pmy_mesh_(pmb.lock()->pmy_mesh) {
 #ifdef MPI_PARALLEL
-  swarm_id_ = pmb.lock()->pbval->bvars_next_phys_id_;
+    swarm_id_ = pmb.lock()->pbval->AdvanceCounterPhysID(1);
 #endif
 
   InitBoundaryData(bd_var_);
--- a/src/bvals/bvals_swarm.cpp
+++ b/src/bvals/bvals_swarm.cpp
@@ -40,11 +40,6 @@ namespace parthenon {
 
 BoundarySwarm::BoundarySwarm(std::weak_ptr<MeshBlock> pmb)
     : bswarm_index(), pmy_block(pmb), pmy_mesh_(pmb.lock()->pmy_mesh) {
-#ifdef MPI_PARALLEL
-  // Create unique communicator for this swarm
-  PARTHENON_MPI_CHECK(MPI_Comm_dup(MPI_COMM_WORLD, &swarm_comm));
-#endif
-
   InitBoundaryData(bd_var_);
 }
 
@@ -62,6 +57,8 @@ void BoundarySwarm::InitBoundaryData(BoundaryData<> &bd) {
 #endif
   }
 }
+
+void BoundarySwarm::SetComm(MPI_Comm mpi_comm) { swarm_comm = mpi_comm; }
 
 void BoundarySwarm::SetupPersistentMPI() {
 #ifdef MPI_PARALLEL
--- a/src/bvals/bvals_swarm.cpp
+++ b/src/bvals/bvals_swarm.cpp
@@ -39,8 +39,10 @@
 namespace parthenon {
 
 BoundarySwarm::BoundarySwarm(std::weak_ptr<MeshBlock> pmb, const std::string &label)
-    : bswarm_index(), pmy_block(pmb), pmy_mesh_(pmb.lock()->pmy_mesh),
-      swarm_comm(pmy_mesh_->GetMPIComm(label)) {
+    : bswarm_index(), pmy_block(pmb), pmy_mesh_(pmb.lock()->pmy_mesh) {
+#ifdef MPI_PARALLEL
+  swarm_comm = pmy_mesh_->GetMPIComm(label);
+#endif
   InitBoundaryData(bd_var_);
 }
 
--- a/src/bvals/bvals_swarm.cpp
+++ b/src/bvals/bvals_swarm.cpp
@@ -41,7 +41,7 @@ namespace parthenon {
 BoundarySwarm::BoundarySwarm(std::weak_ptr<MeshBlock> pmb)
     : bswarm_index(), pmy_block(pmb), pmy_mesh_(pmb.lock()->pmy_mesh) {
 #ifdef MPI_PARALLEL
-    swarm_id_ = pmb.lock()->pbval->AdvanceCounterPhysID(1);
+  swarm_id_ = pmb.lock()->pbval->AdvanceCounterPhysID(1);
 #endif
 
   InitBoundaryData(bd_var_);
--- a/src/bvals/bvals_swarm.cpp
+++ b/src/bvals/bvals_swarm.cpp
@@ -41,11 +41,8 @@ namespace parthenon {
 BoundarySwarm::BoundarySwarm(std::weak_ptr<MeshBlock> pmb)
     : bswarm_index(), pmy_block(pmb), pmy_mesh_(pmb.lock()->pmy_mesh) {
 #ifdef MPI_PARALLEL
-  // Technically, we don't need both (a separate communicator and a tag)
-  // anymore if we use a single communicator per swarm.
-  // TODO(someone) cleanup this and also COMM logic for other vars.
-  PARTHENON_MPI_CHECK(MPI_Comm_dup(MPI_COMM_WORLD, &MPI_COMM_SWARM));
-  swarm_id_ = pmb.lock()->pbswarm->AdvanceCounterPhysID();
+  // Create unique communicator for this swarm
+  PARTHENON_MPI_CHECK(MPI_Comm_dup(MPI_COMM_WORLD, &swarm_comm));
 #endif
 
   InitBoundaryData(bd_var_);
@@ -76,9 +73,8 @@ void BoundarySwarm::SetupPersistentMPI() {
 
     // Neighbor on different MPI process
     if (nb.snb.rank != Globals::my_rank) {
-      send_tag[nb.bufid] =
-          pmb->pbswarm->CreateBvalsMPITag(nb.snb.lid, nb.targetid, swarm_id_);
-      recv_tag[nb.bufid] = pmb->pbswarm->CreateBvalsMPITag(pmb->lid, nb.bufid, swarm_id_);
+      send_tag[nb.bufid] = pmb->pbswarm->CreateBvalsMPITag(nb.snb.lid, nb.targetid);
+      recv_tag[nb.bufid] = pmb->pbswarm->CreateBvalsMPITag(pmb->lid, nb.bufid);
       if (bd_var_.req_send[nb.bufid] != MPI_REQUEST_NULL) {
         MPI_Request_free(&bd_var_.req_send[nb.bufid]);
       }
@@ -104,7 +100,7 @@ void BoundarySwarm::Send(BoundaryCommSubset phase) {
                         "Trying to create a new send before previous send completes!");
       PARTHENON_MPI_CHECK(MPI_Isend(bd_var_.send[nb.bufid].data(), send_size[nb.bufid],
                                     MPI_PARTHENON_REAL, nb.snb.rank, send_tag[nb.bufid],
-                                    MPI_COMM_SWARM, &(bd_var_.req_send[nb.bufid])));
+                                    swarm_comm, &(bd_var_.req_send[nb.bufid])));
 #endif // MPI_PARALLEL
     } else {
       MeshBlock &target_block = *pmy_mesh_->FindMeshBlock(nb.snb.gid);
@@ -142,8 +138,8 @@ void BoundarySwarm::Receive(BoundaryCommSubset phase) {
       MPI_Status status;
 
       if (bd_var_.flag[nb.bufid] != BoundaryStatus::completed) {
-        PARTHENON_MPI_CHECK(MPI_Iprobe(MPI_ANY_SOURCE, recv_tag[nb.bufid], MPI_COMM_SWARM,
-                                       &test, &status));
+        PARTHENON_MPI_CHECK(
+            MPI_Iprobe(MPI_ANY_SOURCE, recv_tag[nb.bufid], swarm_comm, &test, &status));
         if (!static_cast<bool>(test)) {
           bd_var_.flag[nb.bufid] = BoundaryStatus::waiting;
         } else {
@@ -157,7 +153,7 @@ void BoundarySwarm::Receive(BoundaryCommSubset phase) {
           }
           PARTHENON_MPI_CHECK(MPI_Recv(bd_var_.recv[nb.bufid].data(), recv_size[nb.bufid],
                                        MPI_PARTHENON_REAL, nb.snb.rank,
-                                       recv_tag[nb.bufid], MPI_COMM_SWARM, &status));
+                                       recv_tag[nb.bufid], swarm_comm, &status));
         }
       }
     }
--- a/src/bvals/bvals_swarm.cpp
+++ b/src/bvals/bvals_swarm.cpp
@@ -38,8 +38,9 @@
 
 namespace parthenon {
 
-BoundarySwarm::BoundarySwarm(std::weak_ptr<MeshBlock> pmb)
-    : bswarm_index(), pmy_block(pmb), pmy_mesh_(pmb.lock()->pmy_mesh) {
+BoundarySwarm::BoundarySwarm(std::weak_ptr<MeshBlock> pmb, const std::string &label)
+    : bswarm_index(), pmy_block(pmb), pmy_mesh_(pmb.lock()->pmy_mesh),
+      swarm_comm(pmy_mesh_->GetMPIComm(label)) {
   InitBoundaryData(bd_var_);
 }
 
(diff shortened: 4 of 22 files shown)


Yurlungur (commented):
Overall this looks right to me. At some point, I do thing we should rethink how we do the boundary variables more deeply than this. But this at least gets us to a "correct" implementation.

Thanks for digging in to this!

lroberts36:
Doesn't this still leave open the possibility that the `phyd_id` overruns the five bits reserved for it in `BoundaryBase::CreateBvalsMPITag`? A check for this should at least be added as noted in the TODO at https://github.com/lanl/parthenon/blob/1d72227179ce2e34dcba9736587c0f167326d32b/src/bvals/bvals.cpp#L136.

pgrete:
> Doesn't this still leave open the possibility that the `phyd_id` overruns the five bits reserved for it in `BoundaryBase::CreateBvalsMPITag`? A check for this should at least be added as noted in the TODO at
> 
> https://github.com/lanl/parthenon/blob/1d72227179ce2e34dcba9736587c0f167326d32b/src/bvals/bvals.cpp#L136
> .

Good point. I added an appropriate check with error message.

pgrete:
While going through this PR again,I think this needs additional input from @brryan 

For `BoundarySwarms`, the id is always 0 (which is technically reserved for AMR)
https://github.com/lanl/parthenon/blob/97f181b79f145946372a64b9ba2e0382601df632/src/bvals/bvals.hpp#L125
as also set here (for `BoundarySwarms`)
https://github.com/lanl/parthenon/blob/97f181b79f145946372a64b9ba2e0382601df632/src/bvals/bvals.cpp#L197
but
https://github.com/lanl/parthenon/pull/661/files#diff-05bb8e353c4c92e6744e925fad3f395d17ea4e7a98f61978e183873d09e5b11c
actually uses the `pbval` pointer and not the `pbswarm` pointer.

So it looks to me that some things are not properly aligned.
Any comments (also from others)?

Yurlungur:
Is the swarm thing not automatically fixed by just setting a bit for swarms? Or alternatively, making a new communicator for swarms?

pgrete:
> Is the swarm thing not automatically fixed by just setting a bit for swarms? Or alternatively, making a new communicator for swarms?

Yes, there are various straightforward fixes. I think it'd make sense to go with the one that @brryan considers most appropriate given the other open PRs -- especially thinking ahead in terms of supporting multiple Swarms.
So I could imagine that a separate communicator would make the most sense.

brryan:
Thanks for pinging me on this -- yes I would imagine this PR would potentially cause issues for swarms. I think @Yurlungur / @pgrete 's solution of one communicator per swarm is a good path forward. Right now all particle variables are always communicated simultaneously so there shouldn't be any additional subtleties to worry about at the moment. I am out of the office this week but I can think through this in detail on Monday. Probably the tracers example with MPI is the best place to test these changes, since that has both cc vars and a swarm, but I don't think that's included in our automated testing currently.

src/bvals/bvals_swarm.cpp
```diff
@@ -41,7 +41,7 @@ namespace parthenon {
 BoundarySwarm::BoundarySwarm(std::weak_ptr<MeshBlock> pmb)
     : bswarm_index(), pmy_block(pmb), pmy_mesh_(pmb.lock()->pmy_mesh) {
 #ifdef MPI_PARALLEL
-  swarm_id_ = pmb.lock()->pbval->bvars_next_phys_id_;
+  swarm_id_ = pmb.lock()->pbval->AdvanceCounterPhysID(1);
```
brryan:
```suggestion
  swarm_id_ = pmb.lock()->pbswarm->AdvanceCounterPhysID(1);
```
Yes this is a bug, thanks for noticing this @pgrete!

brryan:
@pgrete Looking at this now, my thinking for moving forward with swarm phys ids is to:

1) change to `pbswarm` in `src/bvals/bvals_swarm.cpp`
2) Use the same logic in `AdvanceCounterPhysId` for swarms as for variables, except have `num_phys` always be `1` i.e. the counter is incremented once per swarm (void argument):
```c++
int BoundarySwarm::AdvanceCounterPhysID() {
#ifdef MPI_PARALLEL
  int start_id = bvars_next_phys_id_;
  bvars_next_phys_id_ += 1;
  PARTHENON_REQUIRE_THROWS(
      bvars_next_phys_id_ < 32,
      "Next phys_id would be >= 32, which is currently not supported as Parthenon only "
      "reserves 5 bits for phys_id. Please open an issue on GitHub if you see this "
      "message to discuss options.")
  return start_id;
#else
  return 0;
#endif
}
```

In this pattern each swarm has a specific `phys_id`. I don't anticipating needing large numbers of swarms so a small byte limit might be fine here. To generalize each swarm could have its own communicator later on. 

One thing this prevents is different communication patterns for different subsets of a swarm's `ParticleVariable`s. We always communicate all `ParticleVariable`s currently and I think this is mostly reasonable but probably could lead to unnecessary communication in some cases. I think, though, that that is an issue best addressed later if necessary (dealing with that would probably necessitate a one-communicator-per-swarm model).

Does that seem reasonable to you? I certainly could be missing something.

Yurlungur:
> @pgrete Looking at this now, my thinking for moving forward with swarm phys ids is to:
> 
>     1. change to `pbswarm` in `src/bvals/bvals_swarm.cpp`
> 
>     2. Use the same logic in `AdvanceCounterPhysId` for swarms as for variables, except have `num_phys` always be `1` i.e. the counter is incremented once per swarm (void argument):
> 
> 
> ```c++
> int BoundarySwarm::AdvanceCounterPhysID() {
> #ifdef MPI_PARALLEL
>   int start_id = bvars_next_phys_id_;
>   bvars_next_phys_id_ += 1;
>   PARTHENON_REQUIRE_THROWS(
>       bvars_next_phys_id_ < 32,
>       "Next phys_id would be >= 32, which is currently not supported as Parthenon only "
>       "reserves 5 bits for phys_id. Please open an issue on GitHub if you see this "
>       "message to discuss options.")
>   return start_id;
> #else
>   return 0;
> #endif
> }
> ```
> 
> In this pattern each swarm has a specific `phys_id`. I don't anticipating needing large numbers of swarms so a small byte limit might be fine here. To generalize each swarm could have its own communicator later on.
> 
> One thing this prevents is different communication patterns for different subsets of a swarm's `ParticleVariable`s. We always communicate all `ParticleVariable`s currently and I think this is mostly reasonable but probably could lead to unnecessary communication in some cases. I think, though, that that is an issue best addressed later if necessary (dealing with that would probably necessitate a one-communicator-per-swarm model).
> 
> Does that seem reasonable to you? I certainly could be missing something.

Won't this fail unless there's a separate communicator for swarms vs grid variables?

brryan:
> Won't this fail unless there's a separate communicator for swarms vs grid variables?

Oh yes you're right... I suppose we now need either (1) a separate communicator for swarms (2) a bit reserved in the tag for swarm vs cc variable

Yurlungur:
I am in favor of a separate communicator for swarms, which frees up more bits for variables in their communicator.

pgrete:
I now added a separate communicator (in addition to the tag fix, which is technically not required [any more]).
However, sth goes wrong and I get a deadlock in Receive. Just replacing the three `MPI_COMM_SWARM` with `MPI_COMM_WORLD` int the `BoundarySwarm::Send` and `BoundarySwarm::Receive` methods makes everything work again.
If anyone has a quick idea about what I'm missing, I'm happy for input.

brryan:
@pgrete Thanks for adding the swarm communicator! Unrelated (or not) I just ran into an issue today with particles hanging with MPI (on my heavily modified branches), I'm guessing due to `Swarm::Receive`. I'll take a look at your branch now and see if I notice anything.

pgrete:
FYI this is now ready for re-review. Everything seems to work as expected now as far as I can tell.

pgrete:
FYI this is now ready for re-review. Everything seems to work as expected now as far as I can tell.

pgrete:
Okay, new problem:
When running with true sparse variables, parthenon segfaults at
```
  /// Repoint vbvar's var_cc array at the current variable
  inline void resetBoundary() { vbvar->var_cc = data; }
```
(or more specifically at the)
```
  operator=(const ParArrayNDGeneric<Data> &t) = default;
```
call inside.

I'm not sure why this is happening (maybe because there's no `data` for true sparse variables to begin with.

Before digging into that, I'm actually wondering, do we still need the `resetBoundary` functionality or is this effectively also dead code? Also a question for @lroberts36

lroberts36:
I was discussing this with @Yurlungur on Monday. It is not clear that the call to `resetBoundary()` in `ResetSendBuffers(...)` does anything. `BndInfo.var` is set directly from the variable (i.e. by `v->data` rather than by `v->vbvar->var_cc`), and I don't see anywhere else that uses `vbvar->var_cc` in the `bvals_in_one` code. Maybe the call to `resetBoundary()` can just be commented out? 

I don't think I will use this functionality in my update to the boundary communication routines.

Yurlungur:
Yeah, I suspect that `resetBoundary()` call is not doing anything. I suggest we remove it and see if that breaks anything.

lroberts36:
Yeah, running the `sparse_advection` regression test with that line commented out succeeds.

Yurlungur:
Sounds like we should just remove that line then.

pgrete:
Interesting, for me this still fails because the `vbvar` is not properly setup anymore (so a call to the block pointer fails).
I think this is related to the updated allocation in this PR (that the bvar allocation now actually takes place in `AllocateFluxesAndBdryVar` and not in the variable constructor).
I'm still investigating.

pgrete:
With respect to whether we need `resetBoundary` or not depends on a basic design decision.
For variables across multiple stages the fluxes and boundary vars are getting reused.
So depending on whether we want to allow for sending data and fluxes from other than the base container we need to repoint the data pointer or not.
AthenaPK currently works under the assumption that only the `base` stage contains data/fluxes to be sent.

Yurlungur:
> AthenaPK currently works under the assumption that only the base stage contains data/fluxes to be sent.

I don't think Phoebus or rage make this assumption. But I think in `pack_in_one`, it should not matter, because we use `restrict_in_one` to do the restriction, right? In which case, the metadata in `vbvar` doesn't participate. Or do we still use per-meshblock restriction?

lroberts36:
Yes, anywhere we are working with `MeshData` instead of `MeshBlock`s I think it is unnecessary.

.gitlab-ci.yml
```diff
@@ -95,7 +95,11 @@ parthenon-cuda-short:
       ../
     - make -j${J} advection-example
     - export OMPI_MCA_mpi_common_cuda_event_max=1000
+    # Run one example with variables
     - ctest -R regression_mpi_test:output_hdf5
+    # and one that uses swarms
+    - make -j${J} particle-leapfrog
+    - ctest -R regression_mpi_test:particle_leapfrog
```
brryan, line 102: Thanks!

brryan (approved):
LGTM! A long-overdue fix.

pgrete:
Quick update: the fix I tried didn't work after reintroducing the `resetBoundary` calls as now there's a clash between (reused) `bvars` across stages.
I need to further investigate...

pgrete:
@brryan @Yurlungur this is ready for complete re-review now.
I think I got everything (including an index bug that is currently in `develop` and resulted in the pipeline to fail there [1]).
This PR actually fixes an additional bugs around MPI comm (apart from the tags), i.e., the `bval` should now be properly shared among all shallow copies of each `Variable` (originally there were multiple extra copies as the `PARTHENON_REQUIRE` conditional was off).
I can now see how this could have messed up comms in some circumstances.

[1] I think it may be worth to make sure we always trigger the full suite before merging any open PR in the future
[edit] This reminds me that I need to fix/remove the coverage stage as the original sparse PR resulted in the coverage stage taking hours and then timing out [/edit]

pgrete:
Also ping @lroberts36 so that this is ready to be tested against the MPI issue you encountered for the sparse comm.

Yurlungur:
Thanks @pgrete I'll try to look at this today.

? [1] I think it may be worth to make sure we always trigger the full suite before merging any open PR in the future
[edit] This reminds me that I need to fix/remove the coverage stage as the original sparse PR resulted in the coverage stage taking hours and then timing out [/edit]

I think it depends. For infrastructure PRs I think that's for sure. For smaller changes probably not necessary. We can use our judgement.

docs/interface/state.md
```diff
@@ -140,5 +140,8 @@ The overload `Add(const std::string &label, MeshBlockData<T> &src, const std::ve
 
 Two simple examples of usage of these new containers are 1) to provide storage for multistage integration schemes and 2) to provide a mechanism to allocate storage for right hand sides, deltas, etc.  Both of these usages are demonstrated in the advection example that ships with Parthenon.
 
-Note that in multistage integrator the fluxes and `bvars` of a variable
+Note that in multistage integrator the fluxes and `bvars` (and their MPI communicator) of a variable
 are shared by default across all stages.
+This means that any kind of communication (most prominently flux correction and ghost zone
+exchange) of a given variable at a given stage should not be interleaved with any other
+modifications/communication of said variable as it may result in undefined behavior.
```
Yurlungur, line 147: We can send and then modify before receiving right? Although obviously you probably don't want to do that.

src/bvals/bvals_base.cpp
```diff
@@ -272,22 +273,23 @@ int BoundaryBase::FindBufferID(int ox1, int ox2, int ox3, int fi1, int fi2) {
...
-//              TAG_BITS_PHYS=5, MAX_NUM_PHYS=31
-
-int BoundaryBase::CreateBvalsMPITag(int lid, int bufid, int phys) {
-  return (lid << 11) | (bufid << 5) | phys;
+int BoundaryBase::CreateBvalsMPITag(int lid, int bufid) {
+  PARTHENON_REQUIRE_THROWS(lid >= 0 && bufid >= 0, "Ids are expected to be positive.")
+  int tag = (lid << 6) | bufid;
+  // Technically we could query MPI_TAG_UB here. However, using a fixed (lower) bound
+  // guaranteed by the standard for now to reduce the additional functions calls or
+  // adding a new internal upper lim variable.
```
Yurlungur, line 290: I am in favor of just using the standard-guaranteed value forever. This is better for portability. Otherwise I can imagine the code dying on some machines and not others...

src/interface/variable.cpp
```diff
@@ -180,18 +187,19 @@ void CellVariable<T>::AllocateFluxesAndBdryVar(std::weak_ptr<MeshBlock> wpmb) {
     }
 
     if (IsSet(Metadata::FillGhost)) {
+      // Ensure data of boundary variable points to the newly allocated data.
       vbvar->Reset(data, coarse_s, flux);
 
+      // TODO(someone): double check if following todo is still applicable with the
+      // "restrict-in-one" machinery. The original code added `vbvar` to
+      // `pmb->pbval->bvars`, but we already do this in the constructor. Also, we
+      // potentially don't iterate over `bvars` any more in the "restrict-in-one" case.
```
Yurlungur, line 196: `RestrictBoundaries` is called per meshblock because the physical boundary condition machinery is still per-meshblock. We need to move the boundary stuff to be "in-one" for this to TODO to become non-relevant.

src/mesh/mesh.cpp
```diff
@@ -485,6 +478,11 @@ Mesh::Mesh(ParameterInput *pin, ApplicationInput *app_in, Packages_t &packages,
 
   mesh_data.SetMeshPointer(this);
 
+  resolved_packages = ResolvePackages(packages);
```
Yurlungur, line 481: oops good catch

pgrete:
One more comment/conundrum regarding the failing regression test on Spock #659

The `regression_mpi_test:output_hdf5` currently passes with `develop` but it still fails with this PR.
I just confirmed this in a single session (same node) with fresh binaries.
Even the "conserved" quantity is off, see `advection_3d.hst` of the regression test (third column)
```
  #  History data
  # [1]=time     [2]=dt       [3]=total_advected[4]=max_advected[5]=min_advected
    0.00000e+00  4.68750e-03  1.39160e-02  1.00000e+00  0.00000e+00
    2.53125e-01  4.68750e-03  1.39160e-02  9.38903e-01  0.00000e+00
    5.01563e-01  4.68750e-03  1.39160e-02  7.44601e-01  6.35003e-16
    7.54687e-01  4.68750e-03  1.39160e-02  5.89906e-01  1.27520e-09
    1.00000e+00  3.12500e-03  1.39275e-02  4.80914e-01  1.45889e-07
```
So I'm now wondering if there's sth else we're missing (here).

pgrete:
Update: `develop` is also broken (just in a slightly different way).
Running the test code for longer results in 
```
cycle=288 time=1.3499999999999928e+00 dt=4.6874999999999998e-03 zone-cycles/wsec_step=4.06e+06 wsec_total=1.14e+01 wsec_step=2.31e-02 zone-cycles/wsec=3.84e+06 wsec_AMR=1.36e-03
  what():  ### PARTHENON ERROR
  Condition:   !pvar_cc->IsAllocated()
  Message:     FinishRecvSameLevel: Received variable that was not allocated on sending block but it is allocated on receiving block
  File:        /ccs/home/pgrete/src/parthenon/src/mesh/amr_loadbalance.cpp
  Line number: 1313
```
in both cases. So sth is definitely off with MPI comm (and may or may not be system/env dependent).

Yurlungur:
Oof. I guess we should figure out what's going on here before merge. I also still need to run it through Phoebus to see if that breaks anything... though if `develop` is broken, it probably will.

src/mesh/mesh.cpp
```diff
@@ -1050,8 +1053,10 @@ void Mesh::Initialize(bool init_problem, ParameterInput *pin, ApplicationInput *
     // Create send/recv MPI_Requests for all BoundaryData objects
     for (int i = 0; i < nmb; ++i) {
       auto &pmb = block_list[i];
+      // TODO(mpi people) do we still need the pbval part? Discuss also in the context of
+      // other than cellvariables, see comment above on communicators.
```
Yurlungur, line 1057: Is there any non-"in one" communication we still do?

Yurlungur (commented):
Overall looks reasonable to me, but I want to run it through phoebus and understand what is causing the tests to fail before merge.

pgrete:
> Oof. I guess we should figure out what's going on here before merge. I also still need to run it through Phoebus to see if that breaks anything... though if develop is broken, it probably will.

> Overall looks reasonable to me, but I want to run it through phoebus and understand what is causing the tests to fail before merge.

Are you referring to the failed tests on Spock? I'm still not sure what's going on there but I was not able to reproduce the issue anywhere else so far.
Given that current `develop` has clear bugs (tags, `vbars`, and index) that are all (hopefully) fixed here, my preference is to merge this first and then tackle the other issues #659 #668

docs/interface/state.md
pgrete: It all depends on the driver design (and how the tasks are ordered in one or more task regions).
My intention with this comment was that people should be really mindful with the async nature of tasks in combination with a shared data (flux, coarse, ghost zone buffers) across stages/containers of a single variable.

src/bvals/bvals_base.cpp
pgrete: Technically `MPI_TAG_UB` should provide the right info for the given MPI lib at compile time.
Practically, I'm also in favor of keeping the hard coded limit for now until this actually becomes an issue.

Yurlungur:
> Given that current develop has clear bugs (tags, vbars, and index) that are all (hopefully) fixed here, my preference is to merge this first and then tackle the other issues https://github.com/lanl/parthenon/issues/659 https://github.com/lanl/parthenon/issues/668

That works for me. I'll try running a test with Phoebus today.

src/mesh/mesh.cpp
pgrete: If I understood your comment above correctly then `RestrictBoundaries` may be the last place.
However, as far as I can tell the following line can independently be removed because the `vbar` ctor already enlists the `vbars` in the `pbval` list.

Yurlungur: That seems reasonable.

Yurlungur:
I ran a 2d blast wave in Phoebus with the updated machinery, and 24 MPI ranks. Things seemed to work just fine. Might be good to run a more strenuous test such as a radiation test though, if @brryan or @lroberts36 feel up to it.

Yurlungur (approved):
Since the updated machinery seems to not break Phoebus's integration, I'm calling this good.

pgrete:
@brryan or @lroberts36 do you also want to give this PR another look before merging?

src/bvals/bvals_base.cpp
```diff
@@ -272,22 +273,23 @@ int BoundaryBase::FindBufferID(int ox1, int ox2, int ox3, int fi1, int fi2) {
...
-// TODO(felker) Consider adding safety check: if (tag > MPI_TAG_UB) PARTHENON_FAIL();
-// TODO(felker) Consider adding safety check: signed int inputs & outputs are positive
-// TODO(felker) Store # of bits for each bitfield component in preprocessor macros
-//              TAG_BITS_PHYS=5, MAX_NUM_PHYS=31
-
-int BoundaryBase::CreateBvalsMPITag(int lid, int bufid, int phys) {
-  return (lid << 11) | (bufid << 5) | phys;
+int BoundaryBase::CreateBvalsMPITag(int lid, int bufid) {
+  PARTHENON_REQUIRE_THROWS(lid >= 0 && bufid >= 0, "Ids are expected to be positive.")
+  int tag = (lid << 6) | bufid;
```
lroberts36, line 287: Should we be saving more bits for the `bufid` now that there are no physics bits?

src/mesh/mesh.hpp
```diff
@@ -212,6 +212,11 @@ class Mesh {
   // size of default MeshBlockPacks
   int default_pack_size_;
 
+#ifdef MPI_PARALLEL
+  // Global map of MPI comms for separate variables
+  std::map<std::string, MPI_Comm> mpi_comm_map_;
```
lroberts36: Maybe should be `unordered_map`, although I doubt it matters.

lroberts36 (approved):
LGTM. I also have this merged into #663.

src/bvals/bvals_base.cpp
pgrete: I don't think this is required at the moment as we have a max. of 57 buffers globally.

src/mesh/mesh.hpp
pgrete: I had a misplaced `const` that did nothing and I am about to fix. Will update this at the same time.

pgrete:
Thanks for double checking with Phoebus. Pulling the trigger now so that `develop` works again (pending sparse variables).

PR #662, opened 2022-04-14
Remove SetPrecise

brryan:
## PR Summary

`ParameterInput::SetReal` only stores a floating point representation of the `Real` type; while we do provide a `SetPrecise` function which stores the full `Real`, given that I just got tripped up by this and I don't right now see a use case for storing `Real`s as `float`s, I just removed `SetPrecise`. 

If we want to keep the original behavior, I would advocate that we rename `SetReal` to the more descriptive `SetRealAsFloat` and `SetPrecise` to `SetReal`.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - [[PR 586]](https://github.com/lanl/parthenon/pull/586) Implement true sparse capability with automatic allocation and deallocation of sparse
 
 ### Changed (changing behavior/API/variables/...)
+- [[PR 662]](https://github.com/lanl/parthenon/pull/662) Remove SetPrecise
 - [[PR 623]](https://github.com/lanl/parthenon/pull/623) Enable Params to optionally return non-const pointers
 - [[PR 604]](https://github.com/lanl/parthenon/pull/604) Allow modification of SimTime in PreStepUserWorkInLoop
 - [[PR 617]](https://github.com/lanl/parthenon/pull/617) Unify the coordinates API for MeshBlockPack and VariablePack
--- a/src/driver/driver.hpp
+++ b/src/driver/driver.hpp
@@ -1,5 +1,5 @@
 //========================================================================================
-// (C) (or copyright) 2020-2021. Triad National Security, LLC. All rights reserved.
+// (C) (or copyright) 2020-2022. Triad National Security, LLC. All rights reserved.
 //
 // This program was produced under U.S. Government contract 89233218CNA000001 for Los
 // Alamos National Laboratory (LANL), which is operated by Triad National Security, LLC
--- a/src/driver/driver.hpp
+++ b/src/driver/driver.hpp
@@ -64,11 +64,11 @@ class EvolutionDriver : public Driver {
  public:
   EvolutionDriver(ParameterInput *pin, ApplicationInput *app_in, Mesh *pm)
       : Driver(pin, app_in, pm) {
-    Real start_time = pinput->GetOrAddPrecise("parthenon/time", "start_time", 0.0);
-    Real tstop = pinput->GetOrAddPrecise("parthenon/time", "tlim",
-                                         std::numeric_limits<Real>::infinity());
+    Real start_time = pinput->GetOrAddReal("parthenon/time", "start_time", 0.0);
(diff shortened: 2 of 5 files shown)


Yurlungur:
> If we want to keep the original behavior, I would advocate that we rename SetReal to the more descriptive SetRealAsFloat and SetPrecise to SetReal.

This would be my preferred solution, depending on how much of the code base has to change. Why not rename `SetReal` to `SetFloat` and `SetPrecise` to `SetReal`?

Yurlungur:
> > If we want to keep the original behavior, I would advocate that we rename SetReal to the more descriptive SetRealAsFloat and SetPrecise to SetReal.
> 
> This would be my preferred solution, depending on how much of the code base has to change. Why not rename `SetReal` to `SetFloat` and `SetPrecise` to `SetReal`?

oh duh. That's what you did. Nice. :+1:

Yurlungur (approved):

brryan:
> > > If we want to keep the original behavior, I would advocate that we rename SetReal to the more descriptive SetRealAsFloat and SetPrecise to SetReal.
> > 
> > 
> > This would be my preferred solution, depending on how much of the code base has to change. Why not rename `SetReal` to `SetFloat` and `SetPrecise` to `SetReal`?
> 
> oh duh. That's what you did. Nice. 👍

Well I didn't add a `SetFloat` function but since there's no distinction (I think) between `Real` and other floating point types in parthenon do we even need a `SetFloat`?

pgrete:
I think this topic requires a broader discussion.
The main motivation (as far as I can tell) between the original distinction is that some parameter (like `dt`) have to be in double precision in order to allow for (ideally) bitwise identical simulations on the same system with the same binary, e.g., for restarts.
Our `Real` can technically be a `float` rather than a `double` (though we haven't used/tested this in details as fas as I'm aware of), but there are application scenarios where `float` has advantages over `double` (e.g., in terms of memory usage and performance).

In what ways did you get tripped up by this? I'm currently leaning towards keeping the original implementation and just add more specific documentation so that's why I'm curious to hear how this came up so that we can discuss it with that context in mind.

Yurlungur:
> Our Real can technically be a float rather than a double (though we haven't used/tested this in details as fas as I'm aware of), but there are application scenarios where float has advantages over double (e.g., in terms of memory usage and performance).

While this is true, I can't see any scenario where it hurts to read it in from the input deck at double precision. Is there any scenario where that extra precision would hurt? If we're running with `using Real = float;`, presumably it will just get downcast when actually assigned, which is fine.

brryan:
> In what ways did you get tripped up by this? I'm currently leaning towards keeping the original implementation and just add more specific documentation so that's why I'm curious to hear how this came up so that we can discuss it with that context in mind.

The issue I ran into was setting a parameter at runtime with `SetReal`, and then later reading it back out with `GetReal` and finding that I had a different number (differed at the level of `float` precision). I expected `SetReal` to be associated with the `Real` type rather than `float`.

I see now the point you make about always having `SetPrecise` regardless of whether `Real` is `double` or `float`.  What about if I reintroduce `SetPrecise` and change `SetReal` to `ss_value.precision(std::numeric_limits<Real>::max_digits10);` rather than defaulting to floating point accuracy?

brryan:
OK I updated this according to our conversation on Thursday @Yurlungur @pgrete:
- [x] Default to `double` precision for `SetReal`
- [x] Remove `GetOrAddPrecise`
- [x] `static_assert` that `Real` is less than or equal to `double` in size in `SetReal`
Stop me if I'm misremembering what we agreed on!

Yurlungur (approved):
LGTM

pgrete (changes requested):
I can't comment on the line as it has not been touched in this diff, but I think the `GetOrAddReal` function also needs to be modified so that the default value that is set is also written with all significant digits.

pgrete:
@brryan any reason why the Kokkos and Catch2 modules are updated here?

brryan:
@pgrete Oops thanks for noticing, those modules should have not have been updated here, fixed, `git` must have picked up the change while I was working on multiple branches

pgrete (approved):

PR #663, opened 2022-04-19
Add sparse communication buffers

lroberts36:
## PR Summary

This PR includes a new set of in-one boundary communication routines that are built to work with sparse variables and avoid the `CellCenteredBoundaryVariable` machinery. The boundary buffers, which inherit from `BufArray1D`, are reference counted and associated with a shared memory pool from which they are pulled and to which they are released when they are destroyed to minimize the number of allocations on device (see `utils/memory_pool.hpp`). Buffers are only allocated on sending blocks where sparse variables are allocated. The `CommBuffer` class wraps the memory pool buffers and provides a mechanism for signaling whether or not non-zero data is being sent from the sending block (both for same and differing rank communications). 

## PR Checklist

- [x] Passes `sparse_advection` regression test 
- [x] MPI implementation tested within Parthenon 
- [x] Employs new MPI communicators in (see #661) 
- [x] Code passes cpplint
- [x] New features are documented.
- [x] Code is formatted
- [x] Changes are summarized in CHANGELOG.md
- [x] CI has been triggered on [Darwin](https://re-git.lanl.gov/eap-oss/parthenon/-/pipelines) for performance regression tests.
- [x] (@lanl.gov employees) Update copyright on changed files

PR #669, opened 2022-04-28
Bump clang-format >=11.0

### LINKED ISSUE #637: clang format version
We are now many many versions behind on clang format. `clang-format-13` is the default on my machine now. I suggest that we at some point update clang-format and commit to a new pinned version.

pgrete:
## PR Summary

Quick clang-format update. Fixes #637 

## PR Checklist

- [ ] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/cmake/Format.cmake
+++ b/cmake/Format.cmake
@@ -14,8 +14,16 @@
 find_program(
     CLANG_FORMAT
     NAMES
-        clang-format-8 # Debian package manager, among others, provide this name
-        clang-format-mp-8.0 # MacPorts
+        clang-format-13 # Debian package manager, among others, provide this name
+        clang-format-mp-13.0 # MacPorts
+        clang-format-12 # Debian package manager, among others, provide this name
+        clang-format-mp-12.0 # MacPorts
+        clang-format-11 # Debian package manager, among others, provide this name
+        clang-format-mp-11.0 # MacPorts
+        clang-format-10 # Debian package manager, among others, provide this name
+        clang-format-mp-10.0 # MacPorts
+        clang-format-9 # Debian package manager, among others, provide this name
+        clang-format-mp-9.0 # MacPorts
         clang-format # Default name
     )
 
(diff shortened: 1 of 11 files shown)


Yurlungur:
Why clang-format 9? Perhaps we should jump to 10 or more recent? The most recent version of clang is 13. (That one is probably too new.)

pgrete:
We actually need to jump to >=11 because the stye conflict stems from a bugfix (https://reviews.llvm.org/D83564) first included in that release.
However, I need to bump par-hermes for that, see https://github.com/JoshuaSBrown/cpp-py-formatter/pull/1

[edit] originally I tried 9 as I thought the fix was already included in 9 so all later version would be consistent with this oldest release [/edit]

Yurlungur:
:+1: `clang-format-11` makes sense. For what it's worth, on `singularity-eos` we are on `clang-format-12`.

pgrete:
@par-hermes format

pgrete:
Looks like we cannot test par-hermes without this being merged into `develop` as the current trigger action still uses the old version (Josh's repo).
Anyway, was far as I can tell this works now and I have no reason to believe that the `format` command does not work as the `check` command works as expected.
So this is ready for review now.

cmake/Format.cmake
```diff
@@ -14,8 +14,12 @@
...
     CLANG_FORMAT
     NAMES
-        clang-format-8 # Debian package manager, among others, provide this name
-        clang-format-mp-8.0 # MacPorts
+        clang-format-13 # Debian package manager, among others, provide this name
+        clang-format-mp-13.0 # MacPorts
+        clang-format-12 # Debian package manager, among others, provide this name
+        clang-format-mp-12.0 # MacPorts
+        clang-format-11 # Debian package manager, among others, provide this name
+        clang-format-mp-11.0 # MacPorts
```
Yurlungur, line 22: Do we want all of these available? Might provide weird issues where the local format and the CI format disagree.

Yurlungur (approved):
Looks good. My only concern is clang-format 11 through 13 might have unstable results if mixed.

cmake/Format.cmake
pgrete: As far as I can tell, the results of all these version are consistent and only disagree with older version because this bug (https://reviews.llvm.org/D83564) in the older versions got fixed.
Therefore, my preference is to allow for more options (so that fewer users see a warning about an "untested" version, which is also why I removed the `<=` check and left it at a single `>` check) and revisit/update first, when we see a future breaking version (which is likely to be discovered by us first as I assume that our build/dev environments are on average more update to date than the one of typical users/supercomputer environments).

Yurlungur: Yeah ok that works for me.

lroberts36 (approved):
LGTM. I am running into issues with inconsistency between the clang version on the machine I am working on and the CI being triggered #663, so would be good to get this merged in.

PR #676, opened 2022-05-16
Remove broken swarm user boundary check

### LINKED ISSUE #675: User bounds force swarm bounds
When using user boundary function, swarm boundary function are currently enforced in Parthenon at the mesh level, e.g., 
```c++
    switch (mesh_bcs[f]) {
    case BoundaryFlag::user:
      if (app_in->swarm_boundary_conditions[f] != nullptr) {
        SwarmBndryFnctn[f] = app_in->swarm_boundary_conditions[f];
      } else {
        std::stringstream msg;
        msg << "A user boundary condition for face " << f
            << " was requested, but no swarm condition was enrolled." << std::endl;
        PARTHENON_THROW(msg);
      }
      break;
    default: // Default BCs handled elsewhere
      break;
    }
```

If a downstream app/package does not make use of particles, this results in a hard fail nevertheless.

I wonder what the best solution would be, i.e., where the error handling/sanity should happen?
At the downstream code? In Parthenon (e.g., when a package enrolls particles)? Should we only have one "user function/struct" that contains both so that we implicit enforce "users" to be conscious about both types of functions?
Any thoughts? @Yurlungur @brryan

brryan:
## PR Summary

@pgrete noticed that I recently introduced a bug where, if `user` boundary conditions were requested, `parthenon` required that a `Swarm` user boundary condition be provided even for applications that contain no particles. This small PR removes that broken check, which was actually unnecessary since the boundary condition existence is checked by the `Swarm` itself during setup in `Swarm::AllocateBoundaries`. 

Fixes #675 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - [[PR 586]](https://github.com/lanl/parthenon/pull/586) Implement true sparse capability with automatic allocation and deallocation of sparse
 
 ### Changed (changing behavior/API/variables/...)
+- [[PR 676]](https://github.com/lanl/parthenon/pull/662) Remove broken swarm user boundary check
 - [[PR 662]](https://github.com/lanl/parthenon/pull/662) Remove SetPrecise
 - [[PR 673]](https://github.com/lanl/parthenon/pull/673) Remove smallest meshblock case from advection_performance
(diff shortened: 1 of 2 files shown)


Yurlungur (approved):

pgrete (approved):
Thanks for the quick fix.
I think it makes sense to have the test in the swarm machinery.

PR #677, opened 2022-05-17
Fix restart without `SparseInfo` object. Fixes #648

### LINKED ISSUE #648: Cannot restart without sparse data (SparseInfo) in restart file
```
  what():  ### PARTHENON ERROR
  Message:     HDF5 failure: `H5Oexists_by_name(fh_, "SparseInfo", H5P_DEFAULT)`, Code: 0xffffffff
  File:        /ccs/home/pgrete/parthenon-hydro/external/parthenon/src/outputs/restart.cpp
  Line number: 62
```

https://github.com/lanl/parthenon/blob/3c620221ed9a693c19a36ec4dcb9df07dd159264/src/outputs/restart.cpp#L62
https://github.com/lanl/parthenon/blob/3c620221ed9a693c19a36ec4dcb9df07dd159264/src/outputs/restart.cpp#L63

As a first guess I think that we should remove the `PARTHENON_HDF5_CHECK` here because we actually process the status afterwards.

pgrete:
## PR Summary

Background:

`H5Oexists_by_name`" verifies only that the target object exists. If name includes either a relative path or an absolute path to the target link, intermediate steps along the path must be verified before the existence of the target link can be safely checked", see https://docs.hdfgroup.org/hdf5/develop/group___h5_o.html#title16
With [H5Lexists()](https://docs.hdfgroup.org/hdf5/develop/group___h5_l.html#ga171be6e41dc1a464edc402df0ebdf801) we can check if the link exits (independent of whether it's an object, etc.).
We're only interested in the latter ("do we have a `SparseInfo` entry or not), so I decided to just to this check and not continue with additional checks on what's behind the "link"

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 - [[PR 617]](https://github.com/lanl/parthenon/pull/617) Unify the coordinates API for MeshBlockPack and VariablePack
 
 ### Fixed (not changing behavior/API/variables/...)
+- [[PR 677]](https://github.com/lanl/partheon/pull/677) Fix restart without `SparseInfo` object
 - [[PR 670]](https://github.com/lanl/partheon/pull/670) Fix typo in `parse_value` for non-hdf5 builds
 - [[PR 656]](https://github.com/lanl/partheon/pull/656) Extend CC bvars to 5-, 6-D ParArrays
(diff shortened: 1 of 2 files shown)


brryan (approved):
Looks good to me

Yurlungur (approved):

PR #681, opened 2022-06-08
Refactor ParArrayNDGeneric to work with arbitrary rank Kokkos arrays

lroberts36:
## PR Summary
This pull request refactors `ParArrayNDGeneric` so that it can wrap Kokkos arrays of arbitrary rank. For a 6D array, its behavior should be unchanged from the previous version. Additionally, it adds the possibility of including some state information in a `ParArrayNDGeneric` by inheriting from a state class, which is envisioned to generally be a POD struct (which could contain things like a per variable sparse threshold, for instance). The state is available on both host and device, so this should also allow us to remove things like `VariablePack::sparse_ids_`, `VariablePack::vector_component_`, etc. to simplify the packing machinery as well. 

Any thoughts on the general design are welcome. I tried to add state to the ParArrays in a way that was least invasive and didn't end up putting state data on device where it didn't need to be. 

~~Currently, I am only wrapping arrays down to 3D with `ParArrayNDGeneric`. It seems like the 1D arrays are assumed to be pure `Kokkos::views` enough places in the code that it is hard to make a drop in solution without writing a lot more Kokkos overloads.~~ I wrote the overloads, `ParArrayGeneric` now wraps all dimensions of `ParArray`s.  

The next step will be to thread this through `CellVariable` and `VariablePack`. 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


src/kokkos_abstraction.hpp
```diff
@@ -62,16 +63,21 @@ template <typename T>
 using ParArray1D = Kokkos::View<T *, LayoutWrapper, DevMemSpace>;
 template <typename T>
 using ParArray2D = Kokkos::View<T **, LayoutWrapper, DevMemSpace>;
```
Yurlungur: Why are ParArray1D and ParArray2D different? That seems strange.

src/interface/variable.hpp
```diff
@@ -80,14 +94,14 @@ class CellVariable {
   KOKKOS_FORCEINLINE_FUNCTION
   auto GetDim(const int i) const {
     // we can't query data.GetDim() here because data may be unallocated
-    assert(0 < i && i <= 6 && "ParArrayNDGenerics are max 6D");
+    assert(0 < i && i <= 6 && "ParArrayNDs are max 6D");
```
Yurlungur, line 83: Just realized this is not true anymore. We support up to 7D?

src/parthenon_array_generic.hpp
```diff
@@ -0,0 +1,339 @@
...
+  // Allow a ParArrayGeneric to be cast to any compatible Kokkos view
+  template <class... Ts>
+  operator Kokkos::View<Ts...>() const {
+    return data_;
+  }
+
+  template <class... Args, class = typename enable_if_all_integral<Args...>::type>
+  ParArrayGeneric(const std::string &label, Args... args)
+      : ParArrayGeneric(label, State(),
+                        std::make_index_sequence<Data::rank - sizeof...(Args)>{},
```
Yurlungur, line 132: I did not know about `make_index_sequence`. Nice.

src/interface/variable.hpp
```diff
@@ -54,6 +54,20 @@ inline std::string MakeVarLabel(const std::string &base_name, int sparse_id) {
...
+  VariableState(double alloc, double dealloc)
+      : sparse_allocation_threshold(alloc), sparse_deallocation_threshold(dealloc) {}
+
+  KOKKOS_DEFAULTED_FUNCTION
+  VariableState() = default;
+  KOKKOS_INLINE_FUNCTION
+  explicit VariableState(const empty_state_t &) {}
+
+  double sparse_allocation_threshold;
+  double sparse_deallocation_threshold;
```
brryan: Should these `threshold`s be `const`?

brryan (approved):
LGTM, although I don't have great intuition for all the C++ going on here.

As you say, `ParArray1D`s are widely assumed to actually be `Kokkos::View`s. Personally I think the code would benefit from moving towards a unified `ParArray` approach (so we don't have e.g. `Kokkos::resize()` in some places and `.Resize()` in others). I know not everyone agrees with me on that though. Maybe this isn't the time to re-litigate this issue, but just bringing it up again now that we have this new capability.

tst/unit/test_pararrays.cpp
```diff
@@ -487,8 +553,10 @@ TEST_CASE("Check registry pressure", "[ParArrayND][performance]") {
 
   // view of views. See:
   // https://github.com/kokkos/kokkos/wiki/View#6232-whats-the-problem-with-a-view-of-views
+  using view_3d_t =
+      Kokkos::View<Real ***, parthenon::LayoutWrapper, parthenon::DevMemSpace>;
```
Yurlungur, line 568: should these types be stored somewhere globally?

Yurlungur (approved):
Overall this seems like a reasonable approach. The ability to template on additional information seems like a nice touch and very useful. Some comments below.

src/interface/variable.hpp
lroberts36: When I made them `const` there were some issues with copying the ParArrays around.

Actually, this code probably should not be in this pull request since I am not actually changing how the arrays in `CellVariable` behave. I am going to remove it for now.

src/kokkos_abstraction.hpp
lroberts36: It works fine if I wrap `ParArray2D` as well with `ParArrayGeneric` but wrapping `ParArray1D` causes a number of compilation errors. I think this is because `ParArray1D`s are being used as if they are raw `Kokkos::views` in many places throughout the code and I have not added every possible overload for `ParArrayGeneric` in the overloads I put in the `Kokkos` namespace.

There is no fundamental reason why they shouldn't all be specializations of `ParArrayGeneric`. I was mostly just being lazy. Next week when I am not burned out on looking at compiler error explosions I will try to make `ParArray1D` work as well.

src/interface/variable.hpp
lroberts36: I think this still only supports 6D since `ParArrayND` is still 6D underneath the hood. Of course, it would only be a couple of lines to change this.

Yurlungur: Ah, I thought I saw that `ParArrayND` was 7D under the hood now. Maybe not.

tst/unit/test_pararrays.cpp
lroberts36: I would lean towards no. I just put this there so I didn't have to type as much, but I don't think it would be used many places in the code and the type name is probably not descriptive enough if it is used far away from where it is defined.

Yurlungur: :+1:

src/kokkos_abstraction.hpp
lroberts36: done.

src/parthenon_array_generic.hpp
lroberts36: Yeah, it turns out to have a pretty wide range of uses.

lroberts36:
@pgrete: It would be good to get an external to LANL set of eyes on this before merging. Can you do a quick review? It is ready to go by me at this point. This can go in before #663.

src/parthenon_array_generic.hpp
```diff
@@ -0,0 +1,400 @@
+//========================================================================================
+// Athena++ astrophysical MHD code
+// Copyright(C) 2014 James M. Stone <jmstone@princeton.edu> and other code contributors
```
pgrete:
```suggestion
// Parthenon performance portable AMR framework
// Copyright(C) 2022 The Parthenon collaboration
```

src/parthenon_array_generic.hpp
```diff
@@ -0,0 +1,400 @@
...
+      const ParArrayGeneric<Data2, State2> &arr_in)
+      : State(static_cast<State2>(arr_in)), data_(arr_in.data_) {}
+
+  template <class State2, class... Ts>
+  operator ParArrayGeneric<Kokkos::View<Ts...>, State2>() const {
+    return ParArrayGeneric<Kokkos::View<Ts...>, State2>(
+        data_, State2(static_cast<State>(*this)));
+  }
+
+  // Allow a ParArrayGeneric to be cast to any compatible Kokkos view
```
pgrete, line 86: Very helpful!

src/parthenon_array_generic.hpp
pgrete: Doesn't `make_index_sequence` produce `0 1 2 ... N - 1` rather than `1 1 1 ...` which I'd expect based on the comment above?

src/parthenon_array_generic.hpp
```diff
@@ -0,0 +1,400 @@
...
+  }
+  auto GetHostMirrorAndCopy() { return GetMirrorAndCopy(Kokkos::HostSpace()); }
+
+  template <typename... Args>
+  KOKKOS_INLINE_FUNCTION auto Slice(Args... args) const {
+    auto v = Kokkos::subview(data_, std::forward<Args>(args)...);
+    return ParArrayGeneric<decltype(v), State>(v, *this);
+  }
+
+  // AthenaArray.InitWithShallowSlice(src,dim,indx,nvar)
```
pgrete:
```suggestion
```

src/parthenon_array_generic.hpp
```diff
@@ -0,0 +1,400 @@
...
+  // translates into auto dest = src.SliceD<dim>(std::make_pair(indx,indx+nvar))
+  template <std::size_t N = Data::rank>
+  auto SliceD(index_pair_t slc) const {
+    static_assert(N <= Data::rank);
+    static_assert(N > 0);
+    return SliceD(std::make_index_sequence<Data::rank - N>{},
+                  std::make_index_sequence<N - 1>{}, slc);
+  }
+
+  // AthenaArray.InitWithShallowSlice(src,dim,indx,nvar)
```
pgrete:
```suggestion
```

pgrete (approved):
I think the general design is a good idea.
Admittedly, I don't fully follow all the `make_index_sequence` logic/magic used but I trust that the unit tests do their job.
Also I triggered the full pipeline and the Poisson test seems to fail. This should be fixed prior to merging (but I'm already approving assuming it's a straightforward fix).

src/parthenon_array_generic.hpp
lroberts36: Yes, it produces an `index_sequence` that gives the template parameter pack `0 1 2 ... N-1`, but you can use the comma operator and parameter pack unpacking to create a list of the same size but just containing one value repeated, i.e. 
```c++ 
template <std::size_t SIZE>
std::vector<int> GetVectorOfOnes() {
      return GetVectorOfOnesImpl(std::make_index_sequence<SIZE>{});
}
template <std::size_t... I>
std::vector<int> GetVectorOfOnesImpl(std::index_sequence<I...>) {
      return {(void) I, 1)...};
}
```
returns a vector of `1`s with a size equal to the size of the index sequence. This is essentially the main trick used to make everything work for generic rank arrays. The `(void)` cast is just to suppress compiler warnings about unused variables.

lroberts36:
@pgrete: Poisson test is now fixed. Required reworking the template magic for `GetType` in the reduction machinery.

src/utils/reductions.hpp
```diff
@@ -70,16 +70,26 @@ int GetSize(U &v) {
 
 #ifdef MPI_PARALLEL
 // Generic for containers including Kokkos::View and std::vector
-// strips the pointers off of the type
-template <template <typename U, typename... Args> class Container, typename U,
-          typename... Args>
-MPI_Datatype GetType(Container<U, Args...> &v) {
-  return MPITypeMap<typename cleantypes::remove_all_pointers<U>::type>::type();
```
pgrete, line 77: If `remove_all_pointers` this is gone, can we then also remove `cleantypes.hpp` entirely? As far as I can tell this was the only place it was used.

src/utils/reductions.hpp
```diff
@@ -70,16 +70,26 @@ int GetSize(U &v) {
...
-}
-template <typename U>
-MPI_Datatype GetType(U &v) {
-  return MPITypeMap<U>::type();
+template <class T, class = typename T::value_type>
+struct value_type_test {
+  using void_t = void;
+};
```
pgrete: Just for clarification, does this now "strip the pointer" and makes it `void` type?
I'm trying to match the functionality of the code to the name. Maybe a brief comment could help.

src/parthenon_array_generic.hpp
pgrete: I see. Would you mind adding a brief comment to the first use of `make_index_sequence` and first use of ` ((void)I, 1)` to explain this?
I think this could be helpful for other developers looking at this code in the future as this might not be obvious to everyone (including me :smile: )

lroberts36: Sure, I will add a comment.

src/utils/reductions.hpp
lroberts36: The "`void_t` trick" is SFINAE that can be used for checking if a type contains specific other  types or methods. A more clean implementation than what I have in the code is:  
```c++
template <class... Ts> 
using void_t = void; 

template<class T, class = void> 
struct has_foo_method : std::false_type{}; 

template<class T> 
struct has_foo_method<T, void_t<decltype(std::declval<T>().foo())>> : std::true_type {}; 

template<class T, class = void> 
struct has_value_type : std::false_type{}; 

template<class T> 
struct has_value_type<T, void_t<typename T::value_type>> : std::true_type {}; 

struct A {};
struct B {using value_type = int; void foo(){}};

int main() {
    std::cout << has_value_type<A>::value << std::endl; // prints 0
    std::cout << has_value_type<B>::value << std::endl; // prints 1
    std::cout << has_foo_method<A>::value << std::endl; // prints 0
    std::cout << has_foo_method<B>::value << std::endl; // prints 1
}
``` 
The type doesn't necessarily need to be `void` (for instance there are actually some reasons why it might be a little better to use `bool`). What is happening is `void_t<...> = void` for any well formed `...`, so the compiler chooses the specialization (which inherits from `std::true_type`) as the best match since the second template parameter defaults to `void`. If `...` is not well formed (i.e. the class does not have the requested method) the specialization fails to compile and the general template (which inherits from `std::false_type`) is chosen.  You can then choose between functions just using `std::enable_if<has_*<T>::value>` or `std::enable_if<!has_*<T>::value>`:
```c++
template<class T, typename std::enable_if<has_value_type<T>::value, bool>::type = true>
void test_func(const T&) {
  std::cout << "Has value type." << std::endl;
}

template<class T, typename std::enable_if<!has_value_type<T>::value, bool>::type = true>
void test_func(const T&) {
  std::cout << "Doesn't value type." << std::endl;
}
```
Essentially, I think all of this should go away when C++ concepts become available.

I noticed last night that there are a few other functions above this one that need to be fixed as well, I will try to go through later this weekend and clean everything up. I will also add a little explanation because I agree this is a bit inscrutable. It may make sense to actually implement something like `is_container` that checks if a type has `data()` and `size()` and `value_type` all at the same time.

lroberts36: After thinking about this and looking around a little yesterday, I think I figured out how to implement this in a way that is a little bit more general and doesn't look so obscure. The suggested implementation is in #694.

src/parthenon_array_generic.hpp
lroberts36: done.

src/utils/reductions.hpp
lroberts36: It looks like it is the only place it is used, but I don't see any reason to get rid of it though.

src/utils/reductions.hpp
lroberts36: done.

lroberts36:
I have switched to using concepts in the `reduction` machinery and using concepts to more carefully check the input to `ParArrayGeneric`. This is passing all tests on the Darwin `Power9` nodes. I am going to set this to auto-merge once the gitlab CI runs.

PR #682, opened 2022-06-17
Prolongate in one

Yurlungur:
## PR Summary

This PR implements prolongation-in-one routines, analogous to the restriction-in-one routines. As with restriction-in-one, the old per meshblock/per face routines now point to the in-one routines to (a) test them and ensure they work and (b) reduce code duplicaiton. 

I made this change for the following reasons:
1. It paves the way to cleanup the boundary value machinery and move entirely to in-one routines as we move forward.
2. It's a useful place to hook in for other prolongation/restriction routines, such as face-centered fields or higher-order methods.
3. (Most immediately) I plan to start modifying physical boundaries to support an in-one kind of code-path.

As of now, no coalescence is actually taking place, so this should have no correctness or performance impact. It's just a forward-looking code cleanup.

## PR Checklist

- [x] Code passes cpplint
- [ ] Code is formatted


Yurlungur:
@par-hermes format

Yurlungur:
@pgrete @forrestglines @brryan @jdolence @lroberts36 any thoughts on this?

lroberts36:
> @pgrete @forrestglines @brryan @jdolence @lroberts36 any thoughts on this?

I will take a look today.

src/mesh/refinement_cc_in_one.cpp
```diff
@@ -218,5 +223,218 @@ void ComputePhysicalRestrictBounds(MeshData<Real> *md) {
...
+  auto cib = c_cellbounds.GetBoundsI(interior);
+  auto kb = cellbounds.GetBoundsK(interior);
+  auto jb = cellbounds.GetBoundsJ(interior);
+  auto ib = cellbounds.GetBoundsI(interior);
+
+  const int nbuffers = info.extent_int(0);
+  const int scratch_level = 1; // 0 is actual scratch (tiny); 1 is HBM
+  size_t scratch_size_in_bytes = 1;
+
+  if (cellbounds.ncellsk(entire) > 1) { // 3D
```
lroberts36: Can this be re-written so that there is not a separate piece of code for each dimensionality? I have not looked in detail, but it appears that there are just extra terms that show up as you go up in dimensions. One possibility is to make this a separate template function with the dimension as a template parameter and then just have `if` statements in a single loop. I can imagine having one piece of code to maintain will make implementing other prolongation methods easier.

src/mesh/mesh_refinement.cpp
```diff
@@ -303,164 +303,33 @@ void MeshRefinement::ProlongateCellCenteredValues(const ParArrayND<Real> &coarse
...
-          Real gx1c =
-              0.5 * (SIGN(gx1m) + SIGN(gx1p)) * std::min(std::abs(gx1m), std::abs(gx1p));
-
-          // interpolate on to the finer grid
-          fine(n, fk, fj, fi) = ccval - gx1c * dx1fm;
-          fine(n, fk, fj, fi + 1) = ccval + gx1c * dx1fp;
-        });
+  int b = 0;
+  int nbuffers = coarse.GetDim(6) * coarse.GetDim(5);
+  int nvars = coarse.GetDim(4);
```
lroberts36: I don't understand why the different dimensions are distinguished here and only dimensions 5 and 6 are looped over below even though `Get()` appears to return the entire `coarse` and `fine` arrays.

src/basic_types.hpp
```diff
@@ -36,6 +36,7 @@ using Real = double;
 
 enum class TaskStatus { fail, complete, incomplete, iterate, skip };
 enum class AmrTag : int { derefine = -1, same = 0, refine = 1 };
+enum class RefinementOp_t { Prolongation, Restriction, None };
```
lroberts36, line 39: This is definitely better than carrying around boolean flags.

src/mesh/mesh_refinement.cpp
```diff
@@ -303,164 +303,33 @@ void MeshRefinement::ProlongateCellCenteredValues(const ParArrayND<Real> &coarse
...
+      info_h(b).Nt = coarse.GetDim(6);
+      info_h(b).Nu = coarse.GetDim(5);
+      info_h(b).Nv = nvars;
+      info_h(b).refinement_op = RefinementOp_t::Prolongation;
+      info_h(b).coords = pmb->coords;
+      info_h(b).coarse_coords = this->coarse_coords;
+      info_h(b).fine = fine.Get();
+      info_h(b).coarse = coarse.Get();
+      ++b;
+    }
```
lroberts36: Does this duplicate what is done in `bvals_in_one`? Could that be reused here to save some `deep_copy`s?

src/mesh/mesh_refinement.cpp
```diff
@@ -303,164 +303,33 @@ void MeshRefinement::ProlongateCellCenteredValues(const ParArrayND<Real> &coarse
...
   }
-  return;
+  Kokkos::deep_copy(info, info_h);
```
lroberts36: Does this `deep_copy` occur at every step that prolongation occurs on? Is that necessary?

lroberts36 (commented):
This looks like the right direction to me.

src/mesh/mesh_refinement.cpp
Yurlungur: I think this is a bug actually. In the original implementation I wrote, there was one `BufferInfo` per dim4 of the pack. But since then, it's been modified to take higher-rank objects. I'll clean this up.

src/mesh/mesh_refinement.cpp
Yurlungur: I *think* `bvals_in_one` only treats restriction. My suggestion is that with `prolongate_in_one` in place, we can later remove the per-meshblock code path entirely, and then this function can go away.

src/mesh/mesh_refinement.cpp
Yurlungur: Yes. And no respectively. However, I would argue we just leave it for now, assuming it doesn't impact performance too much, as this machinery is per-meshblock, and I would like to remove it eventually.

src/mesh/refinement_cc_in_one.cpp
Yurlungur: Hmm---it's not entirely obvious to me how to do this a priori, because the loop bounds change. Let me think about it.

lroberts36: Can't you just add 
```c++
int sk = info(buf).sk;
int ek = info(buf).ek;
if (ndim < 3) {
  sk = ckb.s
  ek = ckb.s
}
int sj = info(buf).sj;
int ej = info(buf).ej;
if (ndim < 2) {
  sj = cjb.s
  ej = cjb.s
}
```
just before the `par_for_inner` and then loop over 
```c++
par_for_inner(inner_loop_pattern_ttr_tag, team_member, 0, info(buf).Nt - 1, 0,
                          info(buf).Nu - 1, 0, info(buf).Nv - 1, sk, ek, sj, ej, 
                          info(buf).si, info(buf).ei,
                          [&](const int l, const int m, const int n, const int k,
                              const int j, const int i) {
```
which will give `int fk = (k - ckb.s) * 2 + kb.s = kb.s`, which is what is set by hand now at lower dimensionality.

forrestglines:
> @pgrete @forrestglines @brryan @jdolence @lroberts36 any thoughts on this?

I'm currently on a house hunting trip in New Mexico but I want to look at this one. I'll try to go through it this week but I can't guarantee when I can.

src/mesh/refinement_cc_in_one.cpp
Yurlungur: @lroberts36 take a look now.

src/mesh/mesh_refinement.cpp
Yurlungur: Cleaned up.

Yurlungur:
@par-hermes format

Yurlungur:
@lroberts36 after unifying the restriction code, something broke and the tests now diff. I'm going to have to do some debugging to see if I can fix my cleaned up code

Yurlungur:
> @lroberts36 after unifying the restriction code, something broke and the tests now diff. I'm going to have to do some debugging to see if I can fix my cleaned up code

This is now fixed and should be ready for review again. Interested in your thoughts @lroberts36 . Also would appreciate a look from either @forrestglines or @pgrete on the AthenaPK side.

src/mesh/mesh_refinement.cpp
lroberts36: Yeah, that approach makes sense.

src/mesh/refinement_cc_in_one.cpp
```diff
@@ -218,5 +224,83 @@ void ComputePhysicalRestrictBounds(MeshData<Real> *md) {
...
+                info(buf).Nu - 1, 0, info(buf).Nv - 1, info(buf).sk, info(buf).ek,
+                info(buf).sj, info(buf).ej, info(buf).si, info(buf).ei,
+                [&](const int l, const int m, const int n, const int k, const int j,
+                    const int i) {
+                  impl::ProlongateCellMinMod<3>(
+                      l, m, n, k, j, i, ckb, cjb, cib, kb, jb, ib, info(buf).coords,
+                      info(buf).coarse_coords, info(buf).coarse, info(buf).fine);
+                });
+          }
+        });
```
lroberts36: Can't you just have one of these `par_for_inner`s and set the bounds for `k` depending on dimensionality outside (or maybe it just *works* because of how the bounds of `k` are set originally)? I think this is mostly a matter of taste, but it would remove quite a few lines I think.

src/mesh/refinement_cc_in_one.cpp
```diff
@@ -218,5 +224,83 @@ void ComputePhysicalRestrictBounds(MeshData<Real> *md) {
...
+              info(buf).refinement_op == RefinementOp_t::Prolongation) {
+            par_for_inner(
+                inner_loop_pattern_ttr_tag, team_member, 0, info(buf).Nt - 1, 0,
+                info(buf).Nu - 1, 0, info(buf).Nv - 1, info(buf).sk, info(buf).ek,
+                info(buf).sj, info(buf).ej, info(buf).si, info(buf).ei,
+                [&](const int l, const int m, const int n, const int k, const int j,
+                    const int i) {
+                  impl::ProlongateCellMinMod<3>(
+                      l, m, n, k, j, i, ckb, cjb, cib, kb, jb, ib, info(buf).coords,
+                      info(buf).coarse_coords, info(buf).coarse, info(buf).fine);
```
lroberts36: Good choice to move this to a separate function

src/mesh/refinement_cc_in_one.hpp
```diff
@@ -36,6 +36,112 @@ TaskStatus RestrictPhysicalBounds(MeshData<Real> *md);
...
+  dxfm = xc - fxm;
+  dxfp = fxp - xc;
+}
+
+// TODO(JMM): This function signature is real gross. Might be worth
+// coalescing some of this stuff into structs, rather than unpacking
+// it and then passing it in.
+
+// TODO(JMM): Compared to the previous version of the code, this one
+// multiplies by zero sometimes.
```
lroberts36: I think you could get the exact same set of operations in the same order by including the `ccval` contribution at the end and enclosing some more stuff in the if statements. Not sure if it is worth the time though

lroberts36 (approved):
LGTM once CI passes, my comments are non-blocking

src/mesh/refinement_cc_in_one.hpp
Yurlungur: I thought about that, and decided this version was cleaner. But if we want exactly the same operations, agreed, we could do that.

src/mesh/refinement_cc_in_one.cpp
Yurlungur: Yeah---let me think about that. Agreed that would work and could reduce quite a few lines. Not sure what the best way is stylistically.

Yurlungur: I think I'm going to leave this for now. But when I tackle registering alternative, e.g., higher-order prolongation/restriction, I think that will be the right time to tackle that issue.

Yurlungur:
I also did the same simplification to the `Restrict` calls, so now there's one function for the inner loop of prolongation and one for restriction.

I didn't unify the 3 branched loops. I think that should be a project for a later PR, when I (or somebody) adds user-registerable prolongation/restriction stencils.

Yurlungur:
@par-hermes format

Yurlungur:
@pgrete @forrestglines gentle ping.

src/mesh/mesh_refinement.cpp
```diff
@@ -75,29 +75,23 @@ void MeshRefinement::RestrictCellCenteredValues(const ParArrayND<Real> &fine,
 
   int b = 0;
   int nbuffers = 1;
-  int nvars = fine.GetDim(4);
   cell_centered_bvars::BufferCache_t info("refinement info", nbuffers);
   auto info_h = Kokkos::create_mirror_view(info);
-  for (int l = 0; l < fine.GetDim(6); ++l) {
-    for (int m = 0; m < fine.GetDim(5); ++m) {
```
pgrete, line 82: Are these two indices (for separate `info` structs) not required any more as those components are handled within the restriction routines?

src/mesh/refinement_cc_in_one.cpp
```diff
@@ -37,6 +49,10 @@ void Restrict(cell_centered_bvars::BufferCache_t &info, IndexShape &cellbounds,
 
   // TODO(BRR) nbuffers is currently always 1. In the future nbuffers could be > 1 to
   // improve performance.
+  // JMM: Pretty sure this is only true for the per-meshblock calls,
+  // but for bvals_in_one, this is not the case. I think
+  // nbuffers/variable is 1, but there is one buffer per variable per
+  // meshblock/neighbor pair when doing boundary comms in-one.
```
pgrete: IIRC this statement is correct

src/mesh/refinement_cc_in_one.cpp
```diff
@@ -46,77 +62,35 @@ void Restrict(cell_centered_bvars::BufferCache_t &info, IndexShape &cellbounds,
...
         });
   } else if (cellbounds.ncellsj(entire) > 1) { // 2D
     par_for_outer(
         DEFAULT_OUTER_LOOP_PATTERN, "RestrictCellCenteredValues2d", DevExecSpace(),
         scratch_size_in_bytes, scratch_level, 0, nbuffers - 1,
         KOKKOS_LAMBDA(team_mbr_t team_member, const int buf) {
-          if (info(buf).allocated && info(buf).restriction) {
+          if (impl::DoRefinementOp(info(buf), RefinementOp_t::Restriction)) {
+            const int ck = ckb.s;
             const int k = kb.s;
```
pgrete:
```suggestion
```

src/mesh/refinement_cc_in_one.hpp
```diff
@@ -36,6 +37,170 @@ TaskStatus RestrictPhysicalBounds(MeshData<Real> *md);
...
+  // access cells if dimensionality is correct.
+  const Real vol000 = coords.Volume(k, j, i);
+  const Real vol001 = coords.Volume(k, j, i + 1);
+  const Real fine000 = fine(l, m, n, k, j, i);
+  const Real fine001 = fine(l, m, n, k, j, i + 1);
+  Real vol010, vol011, vol100, vol101, vol110, vol111;
+  Real fine010, fine011, fine100, fine101, fine110, fine111;
+  vol010 = vol011 = vol100 = vol101 = vol110 = vol111 = 0;
+  fine010 = fine011 = fine100 = fine101 = fine110 = fine111 = 0;
+  if (DIM > 1) {
```
pgrete: How about adding a `// TODO(cpp17) make this constexpr` ?

src/mesh/refinement_cc_in_one.hpp
```diff
@@ -36,6 +37,170 @@ TaskStatus RestrictPhysicalBounds(MeshData<Real> *md);
...
+    vol010 = coords.Volume(k, j + 1, i);
+    vol011 = coords.Volume(k, j + 1, i + 1);
+    fine010 = fine(l, m, n, k, j + 1, i);
+    fine011 = fine(l, m, n, k, j + 1, i + 1);
+  }
+  if (DIM > 2) {
```
pgrete: How about adding a `// TODO(cpp17) make this constexpr` ?

src/mesh/refinement_cc_in_one.cpp
```diff
@@ -125,25 +99,19 @@ void Restrict(cell_centered_bvars::BufferCache_t &info, IndexShape &cellbounds,
         DEFAULT_OUTER_LOOP_PATTERN, "RestrictCellCenteredValues1d", DevExecSpace(),
         scratch_size_in_bytes, scratch_level, 0, nbuffers - 1,
         KOKKOS_LAMBDA(team_mbr_t team_member, const int buf) {
-          if (info(buf).allocated && info(buf).restriction) {
+          if (impl::DoRefinementOp(info(buf), RefinementOp_t::Restriction)) {
             const int ck = ckb.s;
             const int cj = cjb.s;
             const int k = kb.s;
             const int j = jb.s;
```
pgrete:
```suggestion
```

src/mesh/refinement_cc_in_one.cpp
pgrete: In principle, we could also move the template up (to `Prolongate<DIM>`) and then work with `constexpr if` inside a single kernel.

src/mesh/refinement_cc_in_one.hpp
```diff
@@ -36,6 +37,170 @@ TaskStatus RestrictPhysicalBounds(MeshData<Real> *md);
...
+}
+
+// TODO(JMM): this could be simplified if grid spacing was always uniform
+template <int DIM>
+KOKKOS_FORCEINLINE_FUNCTION Real GetX(const Coordinates_t &coords, int i) {
+  if (DIM == 1) return coords.x1v(i);
+  if (DIM == 2) return coords.x2v(i);
+  if (DIM == 3) return coords.x3v(i);
+  PARTHENON_FAIL("Unknown dimension");
+}
```
pgrete: How about adding a `// TODO(cpp17) make this constexpr` ?
Or even implement the different types explicitly by hand (shouldn't be many more lines).

src/mesh/refinement_cc_in_one.hpp
```diff
@@ -36,6 +37,170 @@ TaskStatus RestrictPhysicalBounds(MeshData<Real> *md);
...
+
+// TODO(JMM): This function signature is real gross. Might be worth
+// coalescing some of this stuff into structs, rather than unpacking
+// it and then passing it in.
+
+// TODO(JMM): Compared to the previous version of the code, this one
+// multiplies by zero sometimes.
+
+// In the absence of constexpr if, use template DIM to resolve ifs at
+// compile time and avoid branching.
```
pgrete: Not sure I understand the comment. Is this a todo? I still see `if (DIM > 1)` etc in the function that could eventually be  `constexpr`

src/mesh/refinement_cc_in_one.cpp
```diff
@@ -218,5 +186,86 @@ void ComputePhysicalRestrictBounds(MeshData<Real> *md) {
...
+  const int scratch_level = 1; // 0 is actual scratch (tiny); 1 is HBM
+  size_t scratch_size_in_bytes = 1;
+
+  if (cellbounds.ncellsk(entire) > 1) { // 3D
+    par_for_outer(
+        DEFAULT_OUTER_LOOP_PATTERN, "ProlongateCellCenteredValues3d", DevExecSpace(),
+        scratch_size_in_bytes, scratch_level, 0, nbuffers - 1,
+        KOKKOS_LAMBDA(team_mbr_t team_member, const int buf) {
+          if (impl::DoRefinementOp(info(buf), RefinementOp_t::Prolongation)) {
+            par_for_inner(
```
pgrete: Not sure if this is the right place but I think this is the best bet (should be confirmed with some profiling first) for a significant performance regression.
I ran a quick test (3d blast wave with 2 levels of AMR in AthenaPK) and this branch is reaches 53% total performance compared to the previous version, which, if confirmed, I would consider blocking.
My first guess is the different loop abstraction, which is why I made the comment here, because for prolongation hierarchical parallelism is now used.
If there's only one buffer per call (hard-coded above in `mesh_refinement.cpp` prior to the `Prolongate` call, then there's only one team handling this kernel -- translating to using only one CU/SM on a device.

Have you done some profiling/benchmarking with Phoebu and this branch?

pgrete (commented):
Overall, I really like this approach and it should make a transition to a full in-one approach fairly straightforward.
My only (and big) concern at the moment is performance, see the more detailed separate comment, which I think we should investigate/discuss first.

src/mesh/mesh_refinement.cpp
Yurlungur: yes that's right. This was a change @brryan made to support `bvals_in_one` for tensors.

Yurlungur: These loops were left over, but I'm pretty confident they weren't doing anything.

src/mesh/refinement_cc_in_one.hpp
Yurlungur: Yeah this is essentially the TODO, make the ifs constexpr in the future.

src/mesh/refinement_cc_in_one.cpp
Yurlungur: Ah hmm. No I didn't look at performance in Pheobus with this branch. I agree, the problem is probably that the per meshblock calls are under-utilizing the GPU. I think I have a solution---let me try something.

src/mesh/refinement_cc_in_one.hpp
Yurlungur: But you can check that if statements on template parameters do get compiled out.

src/mesh/refinement_cc_in_one.cpp
pgrete: As long as this is hardcoded to a single buffer we can probably revert back to the original `par_for` (just with the new function call inside).

Yurlungur: yes that was my thinking too.

Yurlungur:
@par-hermes format

Yurlungur:
@pgrete @lroberts36 resolving the performance regression @pgrete found turns out to be a bit more involved than I thought. I have come up with two possible paths, which I'd like your feedback on:

1. As @pgrete suggested, one can implement a switch such that the prolongation operation is a single `par_for` per buffer if the number of buffers is smaller than some critical number. This is actually more complex to implement than I initially thought, because the buffer info struct lives on device, and thus the loop bounds required to launch the kernel and the views that are operated on live on device. Copying to host introduces extra latency and also causes problems because copying a view from device to host breaks the reference counting in `Kokkos`. The solution is to pass around a host mirror of the buffer info view of views. This implies threading the host mirror through the `MeshData` and pack in one machinery.
2. @lroberts36 suggested an alternate form of hierarchical parallelism which assigns multiple thread teams to each buffer requiring prolongation or restriction. This involves more complex logic but doesn't require threading things through the infrastructure.

To experiment with both versions, I refactored the the prolongation/restriction code a bit further, abstracting the loop construct out so you only need to define one loop, and it takes in the inner stencil operation as a template parameter. This seems to work nicely, and I will discuss it further in a bit.

What I would like your feedback on is which path to go down as far as the switch vs. the hierarchical parallelism. The switch is, in my opinion, slightly more complicated code-wise, but is exactly as fast as develop, since in the limit of one buffer, it essentially reduces to the method in develop. The hierarchical parallelism, on the other hand, is a nasty `Kokkos` loop that doesn't quite look like our standard `par_for_inner`/`par_for_outer` pattern, but implies a pretty significant performance regression. Some timings on GPU for the advection test I ran below:
| Code Version             | hardware          | zone-cycles/wallsecond |
| ------------------------- | ----------------- | ---------------------- |
| develop                     | 1 A100            | 1.33e7                 |
| prolongate_nbuffs       | 1 A100            | 1.34e7                 |
| prolongate_workers     | 1 A100            | 8.99e6                 |

I don't know whether to try and tweak the worker hierarchical parallelism machinery at this point, or give up and commit to the larger code surgery. Thoughts?

pgrete:
Thanks for the detailed update.

I wonder if this is really an issue thinking in the long run, i.e., the performance regression only shows up because the buffer kernels are different while at the same time the other machinery has not changed.
Once prolong/restrict are actually "in-one" this should (in theory *fingerscrossed*) go away because for practical simulations with AMR the number of block per device will always be larger than one so that there are enough buffers (on average) in that kernel.
Having that said, I now realize that this may also not necessarily be true in case there's just one block on a device that require prolongation/restriction.

Is there potentially some middle ground, e.g., not launching teams per block, but teams per block and variable.
This should (again for practical application) increase the number of teams to a large enough number.
If I don't miss anything everything could stay as is (in terms of how we store boundary info), we'd just need to keep track of
$\sum_i b_i * Nv_i$

Yurlungur:
> Is there potentially some middle ground, e.g., not launching teams per block, but teams per block and variable.

I think if we were to go down this path, the best option would be to return to the buffer being per variable component---e.g., rather than having the inner loop over l, m, n, k, j, i, make it over k, j, i, and have an additional buffer per l, m, n. I can stare at that maybe. (This was was originally how I designed it, but at some point this was changed to support tensor variables in pack_in_one.)

pgrete:
That sounds like a solution, too.
Depending on how much time you want to invest, there's even a further optimization that reduces the boundary info size.
Given that the indices are fixed for a given buffer location (right face, top left corner, etc), we could just store that info in a single variable of the struct, then calculate indices once, and then fetch on the fly in the kernel.

Note, I'm basically happy with any solution that does not degrade the "default" performance significantly.
If it turns out to be too complicated in "this" PR, we could stash it while working on a more complete (like actually doing prolong/restrict-in-one) solution in a branch derived from here.

src/mesh/refinement_cc_in_one.hpp
```diff
@@ -29,13 +30,186 @@
...
+    ProlongationRestrictionLoop<3, Stencil>(std::forward<Args>(args)...);
+  } else if (cellbnds.ncellsj(entire) > 1) { // 2D
+    ProlongationRestrictionLoop<2, Stencil>(std::forward<Args>(args)...);
+  } else if (cellbnds.ncellsi(entire) > 1) { // 1D
+    ProlongationRestrictionLoop<1, Stencil>(std::forward<Args>(args)...);
+  }
+}
+
+} // namespace impl
+
```
Yurlungur: These are the new unified loop constructs. So you can call, for example,
```C++
DoProlongationRestrictionOp<MyProlongationStencil>(cellbnds, info, info_h, cellbounds, c_cellbounds, RefinementOp_t::Prolongation);
```
and that will automatically do prolongation on the set of buffers/variables, appropriately switching to a coalesced loop or hierarchical parallelism as appropriate, and doing the appropriate dimensionality of the loop.

I know it's a little distasteful to have to include cellbounds twice, but there's a win as far as code size.

src/mesh/mesh_refinement_ops.hpp
```diff
@@ -0,0 +1,245 @@
...
+    // symmetry
+    const Real tvol =
+        ((vol000 + vol010) + (vol001 + vol011)) + ((vol100 + vol110) + (vol101 + vol111));
+    coarse(l, m, n, ck, cj, ci) =
+        (((fine000 * vol000 + fine010 * vol010) + (fine001 * vol001 + fine011 * vol011)) +
+         ((fine100 * vol100 + fine110 * vol110) +
+          (fine101 * vol101 + fine111 * vol111))) /
+        tvol;
+  }
+};
```
Yurlungur: This is an example of a functor. You need one per kind of operation you want to do. The struct is templated on `int DIM` and has a `KOKKOS_INLINE_FUNCTION` `Do` which takes indices, index ranges, coords, and the `ParArray`s to operate on. The functor class is passed into the loop construct, but it doesn't have to be specialized to a particular `DIM`. The loop construct can do that for you.

Yurlungur (commented):
@forrestglines here's the locations in the PR I mentioned. Note I still need to change the loops so that each `BufferInfo_t` struct is over elements of a variable, which should increase opportunities for parallelism.

Yurlungur:
@par-hermes format

Yurlungur:
> I think if we were to go down this path, the best option would be to return to the buffer being per variable component---e.g., rather than having the inner loop over l, m, n, k, j, i, make it over k, j, i, and have an additional buffer per l, m, n. I can stare at that maybe. (This was was originally how I designed it, but at some point this was changed to support tensor variables in pack_in_one.)

Ok after staring at this for a little while, I'd like to punt on this until the sparse stuff that @lroberts36 has been working on is in. This requires changing the sparse communication logic, and I'd rather wait until the new sparse machinery lands to avoid wasted effort.

This means I believe this PR is ready for re-review. The changes I have made are now the following:
- Refactor loop structure so there's only one loop function (with some overloads) for both prolongation and restriction of any dimensionality.
- The loop supports both a hierarchical parallelism loop and a coalesced loop
- To support this looping machinery, and looking forward to registering custom prolongation/restriction operations, I implement a *functor* system, where the user defines a struct, templated on dimension. The struct then has a `Do` method that defines the prolongation/restriction stencil. The templating on dimension avoids branching in the stencil itself (if you use `constexpr if`) and allows for the same function to be used "automagically" in the above looping construct, regardless of dimensionality. In the future, I plan to have users register these functors.
- To support the switch between hierarchical parallelism and coalesced looping, without extraneous host-device copies (or requiring Kokkos host pinned memory or unmanaged views), I thread the host mirror of the buffer info struct through `MeshData`. This is a little gross, and we may want to remove this switching mechanism eventually. But it gives us extra flexibility in getting good performance on a piece of hardware and, in the end, doesn't cost much added complexity.
- I enable a run-time parameter that lets you set the minimum number of thread teams required before hierarchical parallelism is switched on. This means it can be set so that in, e.g., re-meshing when prolongating or restricting a whole mesh block, a coalesced loop is used, while in buffer packing with a large meshblock pack, hierarchical parallelism is used. The precise run-time value probably is hardware dependent, which is why I made it run-time set-able.

At this point, very interested in your thoughts, @pgrete and @lroberts36 . @forrestglines I think you might want to take a look too, since your face-centered fields will probably build off of this.

Yurlungur:
@par-hermes format

Yurlungur:
Also worth noting that the switching mechanism means a performance regression is impossible---the old per-meshblock machinery reduces to the coalesced loop, while in-one calls use the hierarchical loop.

Yurlungur:
Looks like CI is still failing?

pgrete:
> > I think if we were to go down this path, the best option would be to return to the buffer being per variable component---e.g., rather than having the inner loop over l, m, n, k, j, i, make it over k, j, i, and have an additional buffer per l, m, n. I can stare at that maybe. (This was was originally how I designed it, but at some point this was changed to support tensor variables in pack_in_one.)
> 
> Ok after staring at this for a little while, I'd like to punt on this until the sparse stuff that @lroberts36 has been working on is in. This requires changing the sparse communication logic, and I'd rather wait until the new sparse machinery lands to avoid wasted effort.
> 
> This means I believe this PR is ready for re-review. The changes I have made are now the following:
> 
>     * Refactor loop structure so there's only one loop function (with some overloads) for both prolongation and restriction of any dimensionality.
> 
>     * The loop supports both a hierarchical parallelism loop and a coalesced loop
> 
>     * To support this looping machinery, and looking forward to registering custom prolongation/restriction operations, I implement a _functor_ system, where the user defines a struct, templated on dimension. The struct then has a `Do` method that defines the prolongation/restriction stencil. The templating on dimension avoids branching in the stencil itself (if you use `constexpr if`) and allows for the same function to be used "automagically" in the above looping construct, regardless of dimensionality. In the future, I plan to have users register these functors.
> 
>     * To support the switch between hierarchical parallelism and coalesced looping, without extraneous host-device copies (or requiring Kokkos host pinned memory or unmanaged views), I thread the host mirror of the buffer info struct through `MeshData`. This is a little gross, and we may want to remove this switching mechanism eventually. But it gives us extra flexibility in getting good performance on a piece of hardware and, in the end, doesn't cost much added complexity.
> 
>     * I enable a run-time parameter that lets you set the minimum number of thread teams required before hierarchical parallelism is switched on. This means it can be set so that in, e.g., re-meshing when prolongating or restricting a whole mesh block, a coalesced loop is used, while in buffer packing with a large meshblock pack, hierarchical parallelism is used. The precise run-time value probably is hardware dependent, which is why I made it run-time set-able.
> 
> 
> At this point, very interested in your thoughts, @pgrete and @lroberts36 . @forrestglines I think you might want to take a look too, since your face-centered fields will probably build off of this.

The overall approach sounds reasonable to me. I'll re-review later this week but don't expect any major comments.

pgrete:
> Looks like CI is still failing?

Yes, and this is likely for real. I get the same error for a local host build.

Yurlungur:
Ah yes this is a typo. I'll fix it.

Yurlungur:
@par-hermes format

Yurlungur:
CI looks like it's passing now. In addition to @pgrete it would be good to have a second pair of eyes from LANL on the PR too. @jdolence @brryan @lroberts36

src/bvals/cc/bvals_cc_in_one.cpp
```diff
@@ -366,7 +365,7 @@ void ResetSendBufferBoundaryInfo(MeshData<Real> *md, std::vector<bool> alloc_sta
 
               auto &coarse_buf = v->vbvar->coarse_buf;
               boundary_info_h(b).var = coarse_buf.Get();
-              boundary_info_h(b).restriction = true;
+              boundary_info_h(b).refinement_op = RefinementOp_t::Restriction;
```
lroberts36, line 369: Changes in `bvals_cc_in_one.cpp` will have to be brought into the sparse bvals stuff (or vice versa depending on which gets merged first).  It doesn't look like it will be too bad but may require some thought. Also, I think I like this refinement op enum.

src/mesh/mesh_refinement_ops.hpp
```diff
@@ -0,0 +1,245 @@
...
+ * multiplies by zero sometimes.
+ */
+
+namespace parthenon {
+namespace refinement_ops {
+
+namespace util {
+// TODO(JMM): this could be simplified if grid spacing was always uniform
+template <int DIM>
+KOKKOS_INLINE_FUNCTION Real GetX(const Coordinates_t &coords, int i);
```
lroberts36: Maybe this should be `GetXCC` or something to denote it is the value at the cell center.

src/mesh/mesh_refinement_ops.hpp
```diff
@@ -0,0 +1,245 @@
...
+  const Real xm = GetX<DIM>(coarse_coords, i - 1);
+  const Real xc = GetX<DIM>(coarse_coords, i);
+  const Real xp = GetX<DIM>(coarse_coords, i + 1);
+  dxm = xc - xm;
+  dxp = xp - xc;
+  const Real fxm = GetX<DIM>(coords, fi);
+  const Real fxp = GetX<DIM>(coords, fi + 1);
+  dxfm = xc - fxm;
+  dxfp = fxp - xc;
+}
```
lroberts36: `GetSlopes` seems like a confusing name for this function since it appears to be getting the distances between the cell centers on the fine and the coarse grid. Maybe it should be `GetGridSpacings`? 

I also don't love passing output variables by reference since it is not clear at the call site that these variables may be changed within the function.

src/mesh/mesh_refinement_ops.hpp
```diff
@@ -0,0 +1,245 @@
...
+KOKKOS_FORCEINLINE_FUNCTION
+Real GradMinMod(const Real ccval, const Real fm, const Real fp, const Real dxm,
+                const Real dxp) {
+  const Real gxm = (ccval - fm) / dxm;
+  const Real gxp = (fp - ccval) / dxp;
+  return 0.5 * (SIGN(gxm) + SIGN(gxp)) * std::min(std::abs(gxm), std::abs(gxp));
+}
+} // namespace util
```
lroberts36: `ccval` -> `fc` for consistency

src/mesh/mesh_refinement_ops.hpp
```diff
@@ -0,0 +1,245 @@
...
+    if (DIM > 2) { // TODO(c++17) make constexpr
+      vol100 = coords.Volume(k + 1, j, i);
+      vol101 = coords.Volume(k + 1, j, i + 1);
+      vol110 = coords.Volume(k + 1, j + 1, i);
+      vol111 = coords.Volume(k + 1, j + 1, i + 1);
+      fine100 = fine(l, m, n, k + 1, j, i);
+      fine101 = fine(l, m, n, k + 1, j, i + 1);
+      fine110 = fine(l, m, n, k + 1, j + 1, i);
+      fine111 = fine(l, m, n, k + 1, j + 1, i + 1);
+    }
```
lroberts36:
```suggestion
    Real vol[2][2][2], fine[2][2][2];
    const int xoff = 1; 
    const int yoff = DIM > 1 ? 1 : 0;
    const int zoff = DIM > 2 ? 1 : 0; 
    for (int iz = 0; iz < 2; ++iz) 
      for (int iy = 0; iy < 2; ++iy)
        for (int ix = 0; ix < 2; ++ix) {
        vol[iz][iy][ix] = coords.Volume(k + iz * zoff, j + iy * yoff, i + ix * xoff);
        fine[iz][iy][ix] = fine(k + iz * zoff, j + iy * yoff, i + ix * xoff);
    }    
```

src/mesh/mesh_refinement_ops.hpp
```diff
@@ -0,0 +1,243 @@
...
+};
+
+template <int DIM>
+struct ProlongateCellMinMod {
+  KOKKOS_FORCEINLINE_FUNCTION static void
+  Do(const int l, const int m, const int n, const int k, const int j, const int i,
+     const IndexRange &ckb, const IndexRange &cjb, const IndexRange &cib,
+     const IndexRange &kb, const IndexRange &jb, const IndexRange &ib,
+     const Coordinates_t &coords, const Coordinates_t &coarse_coords,
+     const ParArray6D<Real> &coarse, const ParArray6D<Real> &fine) {
```
lroberts36: I would pass these as pointers to make it clear they can be altered (also in other `Do` instances).

src/globals.cpp
```diff
@@ -43,5 +43,22 @@ Real receive_boundary_buffer_timeout;
...
+namespace cell_centered_refinement {
+// If the info object has more buffers than this, do
+// hierarchical parallelism. If it does not, loop over buffers on the
+// host and launch kernels manually.
+//
+// min_num_bufs = 1 implies that the old per-buffer machinery doesn't
+// use hierarchical parallelism. This also means that for
+// prolongation/restriction over a whole meshblock, hierarchical
+// parallelism is not used, which is probably important for
+// re-meshing.
```
lroberts36: Isn't hierarchical parallelism used if `num_bufs > min_num_bufs` and otherwise multiple kernels are launched?

lroberts36 (approved):
General approach LTGM, although I left a few comments relating to code clarity. 

Also, I find it confusing that the prolongation and restriction operations live in `refinement_cc_in_one.*` and that they rely on `BufferCache_t` from `cell_centered_bvars`. I think it might be worth it to reorganize this as part of a future pull request.

src/mesh/mesh_refinement_ops.hpp
Yurlungur: :+1: Done. Might be worth revisiting when we bring in face-centered fields, i.e., maybe we want to expose an enum param or another template parameter or something. But for now let's make things explicit.

src/mesh/mesh_refinement_ops.hpp
Yurlungur: I guess it's my old-school C programming habits, but I prefer passing by reference over returning a struct or tuple. Agreed on the naming, and fixed.

src/mesh/mesh_refinement_ops.hpp
Yurlungur: :+1:

src/mesh/mesh_refinement_ops.hpp
Yurlungur: this is technically more cycles, isn't it? Since the looping will happen regardless of `DIM`? I'd rather avoid the loops, or at least not coalesce them. But I think replacing the names with arrays is a good improvement.

src/mesh/mesh_refinement_ops.hpp
Yurlungur: what about just removing `const`?

src/mesh/mesh_refinement_ops.hpp
lroberts36: yeah, it is more cycles for 1D and 2D but I am skeptical that it makes a difference compared to memory access.

src/mesh/mesh_refinement_ops.hpp
lroberts36: I think it is more of how it looks on the calling end. For a function `f(double a_in, double b_in, double& c_out)` it is unclear at first glance at the call site what might be being updated
```c++
double a, b, c; 
func(a, b, c); 
```
Switching to `f(double a_in, double b_in, double* c_out)` and calling with 
```c++
double a, b, c; 
func(a, b, &c); 
```
makes it instantly clear that `c` may be changed by `func`. This is also what I meant for the `GetSlopes`/`GetGridSpacing` function above.

Yurlungur: yeah ok I can see that. I don't think it really matters here, but I don't feel strongly about it, so I'll change it.

Yurlungur: One thing to just put a pin in: when capturing `Kokkos::View`s, it's actually pretty important the capture be by value. But this is already all device code, so I think it won't matter here.

src/mesh/mesh_refinement_ops.hpp
Yurlungur: Actually just realized this won't work because I need `vol` to be set to zero for the elements that are unused. Fortunately, I think I found a better solution that uses your loop idea:
```C++
    Real vol[2][2][2], terms[2][2][2];
    std::memset(&vol[0][0][0], 0., 8*sizeof(Real));
    std::memset(&terms[0][0][0], 0., 8*sizeof(Real));
    constexpr int xoff = 1;
    constexpr int yoff = (DIM > 1);
    constexpr int zoff = (DIM > 2);
    for (int iz = 0; iz < 1 + zoff; ++iz) {
      for (int iy = 0; iy < 1 + yoff; ++iy) {
        for (int ix = 0; ix < 1 + xoff; ++ix) {
          vol[iz][iy][ix] = coords.Volume(k + iz, j + iy, i + ix);
          terms[iz][iy][ix] = vol[iz][iy][ix]*fine(l, m, n, k + iz, j + iy, i + ix);
        }
      }
    }
```
this has the advantage of no extra cycles too, since those loops are exaclty the size we need. And better yet, the bounds are evaluated at compile time, since the offsets can be `constexpr` so we don't need `constexpr if` here to get all the advantages of the templating.

src/globals.cpp
Yurlungur: yes that's right. I think that's what I wrote here. Maybe this verbose comment is too much. I'll simplify the statement to be a bit more clear.

src/mesh/mesh_refinement_ops.hpp
Yurlungur: Actually---I can't change the `ParArray6D`s to pointers, because the `[=]` capture in `KOKKOS_LAMBDA` automatically makes everything thats captured into a `const`.

src/mesh/mesh_refinement_ops.hpp
lroberts36: I think the original solution will work since you are just adding the volume twice for every term you add twice (when `zoff` and/or `yoff` is zero), so the average is the same, but I agree this is probably a little faster.

src/mesh/mesh_refinement_ops.hpp
Yurlungur: Well---maybe const pointers. I'll try that and see if the compiler accepts it.

Yurlungur:
Assuming tests pass, I think I've addressed all your feedback, @lroberts36 .

lroberts36:
Thanks, LGTM!

src/mesh/mesh_refinement_ops.hpp
```diff
@@ -0,0 +1,239 @@
...
+  *fi = (i - cib.s) * 2 + ib.s;
+  const Real xm = GetXCC<DIM>(coarse_coords, i - 1);
+  const Real xc = GetXCC<DIM>(coarse_coords, i);
+  const Real xp = GetXCC<DIM>(coarse_coords, i + 1);
+  *dxm = xc - xm;
+  *dxp = xp - xc;
+  const Real fxm = GetXCC<DIM>(coords, *fi);
+  const Real fxp = GetXCC<DIM>(coords, *fi + 1);
+  *dxfm = xc - fxm;
+  *dxfp = fxp - xc;
```
pgrete, line 117: Is this still the original Athena++ naming convention meaning that *f* would refer to the face centered positions?
For uniform Cartesian this does not matter otherwise we would need to further specialize GetXCC to also access `coords.x#f(i)`.

If `fxm` etc are sth different then I suggest to add a comment to the function as I got confused here.

src/mesh/mesh_refinement_ops.hpp
```diff
@@ -0,0 +1,239 @@
...
+    constexpr int yoff = (DIM > 1);
+    constexpr int zoff = (DIM > 2);
+    for (int iz = 0; iz < 1 + zoff; ++iz) {
+      for (int iy = 0; iy < 1 + yoff; ++iy) {
+        for (int ix = 0; ix < 1 + xoff; ++ix) {
+          vol[iz][iy][ix] = coords.Volume(k + iz, j + iy, i + ix);
+          terms[iz][iy][ix] = vol[iz][iy][ix] * fine(l, m, n, k + iz, j + iy, i + ix);
+        }
+      }
+    }
```
pgrete: In case you are going to make a change to this file prior to merging, I suggest to use ijk instead of xyz for overall consistency.

src/mesh/refinement_cc_in_one.hpp
```diff
@@ -20,22 +20,179 @@
...
+      DevExecSpace(), scratch_size_in_bytes, scratch_level, 0, nbuffers - 1,
+      KOKKOS_LAMBDA(team_mbr_t team_member, const int buf) {
+        if (DoRefinementOp(info(buf), op)) {
+          int sk, ek, sj, ej, si, ei;
+          GetLoopBoundsFromBndInfo<DIM>(info(buf), ckb.s, cjb.s, sk, ek, sj, ej, si, ei);
+          par_for_inner(inner_loop_pattern_ttr_tag, team_member, 0, info(buf).Nt - 1, 0,
+                        info(buf).Nu - 1, 0, info(buf).Nv - 1, sk, ek, sj, ej, si, ei,
+                        [&](const int l, const int m, const int n, const int k,
+                            const int j, const int i) {
+                          Stencil<DIM>::Do(l, m, n, k, j, i, ckb, cjb, cib, kb, jb, ib,
```
pgrete: Switching between u, v, t, and l, m, n here.
At some point we should try to make the entire codebase consistent.

src/parthenon_manager.cpp
```diff
@@ -131,6 +131,10 @@ ParthenonStatus ParthenonManager::ParthenonInitEnv(int argc, char *argv[]) {
   Globals::receive_boundary_buffer_timeout =
       pinput->GetOrAddReal("parthenon/time", "recv_bdry_buf_timeout_sec", -1.0);
 
+  // set boundary comms buffer switch trigger
+  Globals::cell_centered_refinement::min_num_bufs =
+      pinput->GetOrAddReal("parthenon/mesh", "refinement_in_one_min_nbufs", 6);
+
```
pgrete: This number should probably be higher. An A100 has up to 128 SMs but I don't have a feeling for where the tradeoff between launching kernels and runtime is for small block sizes.
For small blocks this number is easily in the thousands on a device and for large blocks the latency will be less important.
My gut tells me that sth like 64 might be a good start but as this is a runtime parameter we can leave as is and remind people that this may impact performance.
Having said that I'm wondering it'd make sense to start a doc page to highlight the most important performance knobs for a given simulation, e.g., pack_size, number of ranks per GPU, this parameter, etc.

pgrete (approved):
LGTM!
Approving pending a reply to the question on potential grid spacing between faces.

src/mesh/mesh_refinement_ops.hpp
Yurlungur: These are the original `Athena++` names but they're not faces. They're the fine grid. I'll add a comment.

src/mesh/mesh_refinement_ops.hpp
Yurlungur: can't use `i,j,k` directly because they're already in use. But I can come up with something closer to the standard names

src/mesh/refinement_cc_in_one.hpp
Yurlungur: agreed this is a holdover from previous code. I can change it here.

src/parthenon_manager.cpp
Yurlungur: 6 was chosen arbitrarily. I can change it to 64 to start. Agreed on the docs page:
https://github.com/lanl/parthenon/issues/706

Yurlungur:
Thanks @pgrete I think I've addressed all your comments. So I'll go ahead and merge when tests pass.

src/mesh/mesh_refinement_ops.hpp
Yurlungur: :point_up: That said, we will need to specialize for face-centered fields likely. But lets cross that bridge when we come to it. The intent here is not that this utility is used for all refinement operations. I just wrote it to clean up the cell-centered ones.

PR #688, opened 2022-06-21
Restore component labels for multicomponent non-vector field

### LINKED ISSUE #684: Output component naming
After the merge #679 reading AthenaPK data is broken because we have another edge case.

Our main variables are stored in a single field, but it's not a vector in the mathematical/physical sense.
Therefore the change introduced by the PR
```c++
    if (vlen == 1 || is_vector) {
      component_labels = component_labels_.size() > 0 ? component_labels_
                                                      : std::vector<std::string>({label});
    } else {
      for (int i = 0; i < vlen; i++) {
        component_labels.push_back(
            label + "_" +
            (component_labels_.empty() ? std::to_string(i) : component_labels_[i]));
      }
    }
``` 
results in the components being labelled as
```hdf5
         DATASPACE  SIMPLE { ( 5 ) / ( 5 ) }
         DATA {
         (0): "prim_Density", "prim_Velocity1", "prim_Velocity2",
         (3): "prim_Velocity3", "prim_Pressure"
         }
```

One easy fix would be to adjust the analysis tools (e.g., `yt`) to use the new field names.
However, I'm not sure if the current behavior is the desired behavior.
From a downstream developer point of view, I'd naively expect that if I provide an array/vecor with labels for all components then these are the desired labels.
Therefore, I'm currently in favor of not making compound names in case a developer has specified names for all components.

Any thoughts? @brryan @Yurlungur @forrestglines

pgrete:
## PR Summary

Fixes #684 

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 - [[PR 617]](https://github.com/lanl/parthenon/pull/617) Unify the coordinates API for MeshBlockPack and VariablePack
 
 ### Fixed (not changing behavior/API/variables/...)
+- [[PR 688]](https://github.com/lanl/parthenon/pull/688) Restore component labels for multicomponent non-vector field
 - [[PR 679]](https://github.com/lanl/parthenon/pull/679) Handle case of multidim var labeling for output
 - [[PR 680]](https://github.com/lanl/partheon/pull/680) Fix hanging compilation for sort unit test
 - [[PR 678]](https://github.com/lanl/partheon/pull/678) Fix FlatIdx packing for size-1 dimensions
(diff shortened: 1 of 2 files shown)


brryan (approved):
@pgrete Thanks for cleaning up my mistake, this works with our downstream code, so assuming the gitlab CI passes (the `output_hdf5` test is a good sanity check for changes here) I think this can just be merged in ASAP

Yurlungur (approved):
LGTM. This is fine for Phoebus and I think it's the behavior we want.

PR #689, opened 2022-06-21
Add `Mesh::ProblemGenerator`

pgrete:
## PR Summary

We talked about capabilities to do reduction during initialization and this seemed to me the least intrusive method.
No API change (the MeshBlock one is still the default one) so should not break downstream codes.
I also modified the poisson example to use the Mesh version and the test suite passes.

If people think this is not a good idea, I'm also open to alternatives.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Current develop
 
 ### Added (new features/APIs/variables/...)
+- [[PR 689]](https://github.com/lanl/parthenon/pull/689) Add `Mesh::ProblemGenerator` (allows reductions during init)
 - [[PR 667]](https://github.com/lanl/parthenon/pull/667) Add parallel scan
 - [[PR 654]](https://github.com/lanl/parthenon/pull/654) Add option for returning FlatIdx when requested variable doesn't exist
 - [[PR 653]](https://github.com/lanl/parthenon/pull/653) Allow for multi-D particle variables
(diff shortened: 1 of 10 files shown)


Yurlungur (commented):
I'm not opposed in principle. How does this interact with AMR? I guess you just run the problem generator on the whole mesh after each refinement at initialization?

In Phoebus, we have a post-initialization modifier, where we walk through the mesh and make changes. This is certainly cleaner than that.

pgrete:
> I'm not opposed in principle. How does this interact with AMR? I guess you just run the problem generator on the whole mesh after each refinement at initialization?

This should work just the same way the MeshBlock one work as there's only place place where these functions are called.
More specifically, they are called in the `do ...while` loop that only ends if the number of blocks is unchanged at the end (after checking for refinement and then potentially refine and redistribute).
Given that `ProblemGenerator` is called early on in that loop, everything inside (including potential normalization or other "global" work) will be done before any of the refinement machinery is called.

Yurlungur (approved):
Then seems fine to me. I think this simplifies some logic in Phoebus as well, which is nice. @brryan we should rewrite to avoid the post-initialization modifier stuff.

Also I think this allows us to run, e.g., the TOV solver in the mesh pgen as opposed to before pgen gets called.

brryan (approved):
Nice enhancement! Yes as @Yurlungur says this will clean up our initialization routines substantially.

PR #692, opened 2022-07-13
Machinery for packing sparse variables

lroberts36:
## PR Summary
This pull request is for a new class `SparsePack` that packs sparse data across blocks in a way that eliminates the need to check the allocation status of each variable. This duplicates some of the functionality of the other types of packs already in Parthenon and in one of @jdolence branches, but has the following nice features:

- Allows for regexes to generically select groupings of variables. 
- Only includes allocated variables in the pack, so that allocation status does not need to be checked in kernels.
- Stores the range of allocated variables of each group on each block on device, so that the range of indices for a particular variable group can easily be accessed in an outer loop.
- Provides both an `unordered_map<string, SparseVarIdx>` and a type based method for accessing variables within a kernel (see the `test_sparse_pack.cpp` unit test for examples). 
- Templated so that `SparsePack`s can be created from `MeshData` or `MeshBlockData`.
- Implements a `SparsePackCache` type to internally handle cacheing. 
- Aside from including a `SparsePackCache` member variable in `MeshData` and `MeshBlockData` (and an associated `GetSparsePackCache()` method), methods relating to `SparsePack` are not included in `MeshData` and `MeshBlockData`. 
- For better or worse, the View of Views for the pack is currently implemented as a `ParArray3D<ParArray3D<Real>>` with the first index distinguishing between the variable and its fluxes, the second the block index, and the third the variable index on the block. Since the number of variables differs between blocks when things are sparsely allocated, the size of the last dimension is the maximum number of variables allocated on any block. This means there is sometimes space in the array that is unused, but I doubt this will be a real major issue unless it creates too large of a `deep_copy`. 

Currently, I have changed `FluxDivergence(MeshData<Real>*, MeshData<Real>*)` to use `SparsePack`s just to test if this breaks anything. Not surprisingly, the exact choice of loop pattern substantially impacts performance. The chosen loop pattern seems to do as well as the old one on the `regression_test:advection_performance` test, but maybe there is a better way. Aside from this, this PR should not impact any aspect of the code and the change to `FluxDivergence` could be reverted. 

I also removed the `<regex>` header from the list of unapproved headers in `cpplint.py`. As far as I am aware, there isn't a really good reason for having it there (see https://github.com/google/styleguide/issues/194). That being said, a cursory glance around the internet suggested that `boost::regex` might be superior performance wise.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


Yurlungur:
> I also removed the <regex> header from the list of unapproved headers in cpplint.py. As far as I am aware, there isn't a really good reason for having it there (see https://github.com/google/styleguide/issues/194). That being said, a cursory glance around the internet suggested that boost::regex might be superior performance **wise.**

I remember there being a bug in `ccplint.py` that caused problems with some headers. Maybe `<regex>` was one of them. But if the linter is behaving itself, I don't object to removing the blacklist.

Assuming the `<regex>` isn't a performance bottleneck, I strongly prefer the `stdlib` header over `boost`, simply to keep the number of dependencies down.

lroberts36:
I think the linter is doing what it is supposed to. Rather, I think that google has an in-house regex library that they are supposed to use, so they flag `std::regex`.

src/interface/meshblock_data.hpp
```diff
@@ -473,8 +476,8 @@ class MeshBlockData {
                                  "', but no such sparse variable exists");
 
     auto var = GetCellVarPtr(label);
-    PARTHENON_REQUIRE_THROWS(var->IsSparse(),
-                             "Tried to deallocate non-sparse variable " + label);
+    // PARTHENON_REQUIRE_THROWS(var->IsSparse(),
+    //                         "Tried to deallocate non-sparse variable " + label);
```
Yurlungur, line 480: I can't tell at a glance whether or not removing this `REQUIRE` is correct. But I will trust that it is. If so, should the commented out code be deleted?

src/interface/sparse_pack.hpp
```diff
@@ -0,0 +1,282 @@
...
+  KOKKOS_INLINE_FUNCTION
+  base_t() : idx(0) {}
+
+  KOKKOS_INLINE_FUNCTION
+  explicit base_t(int idx1) : idx(idx1) {}
+
+  virtual ~base_t() = default;
+
+  // All of these are just static methods so that there is no
+  // extra storage in the struct
```
Yurlungur, line 91: Should these be marked `virtual` too to avoid shadowing warnings? Or does that just happen automatically with `static`?

src/interface/sparse_pack_base.cpp
```diff
@@ -0,0 +1,265 @@
...
+
+namespace {
+// SFINAE for block iteration so that sparse packs can work for MeshBlockData and MeshData
+template <class T, class F>
+inline auto IterateBlocks(T *pmd, F func) -> decltype(T().GetBlockData(0), void()) {
+  for (int b = 0; b < pmd->NumBlocks(); ++b) {
+    auto &pmbd = pmd->GetBlockData(b);
+    func(b, pmbd.get());
+  }
+}
```
Yurlungur: This is nice. Might want to pull this out of `sparse_pack_base` and put it somewhere more general.

src/interface/sparse_pack_base.hpp
```diff
@@ -0,0 +1,199 @@
...
+#include "utils/utils.hpp"
+
+namespace parthenon {
+namespace impl {
+struct PackDescriptor {
+  std::vector<std::string> vars;
+  std::vector<bool> use_regex;
+  std::vector<MetadataFlag> flags;
+  bool with_fluxes;
+  bool coarse;
```
Yurlungur: This is a nitpick, but I definitely prefer these at the bottom of the struct.

src/interface/sparse_pack_base.hpp
```diff
@@ -0,0 +1,199 @@
...
+#include <utility>
+#include <vector>
+
+#include "coordinates/coordinates.hpp"
+#include "interface/variable.hpp"
+#include "utils/utils.hpp"
+
+namespace parthenon {
+namespace impl {
+struct PackDescriptor {
```
Yurlungur, line 40: This is nice. Maybe we should use something like this for the whole non-sparse pack machinery. Or is the idea that `SparsePack` eventually replaces dense packs?

src/interface/sparse_pack_base.hpp
```diff
@@ -0,0 +1,199 @@
...
+                      "Probably shouldn't be making a coarse pack with fine fluxes.");
+  }
+};
+} // namespace impl
+
+using namespace impl;
+
+class PackIdx {
+  std::size_t vidx;
+  int off;
```
Yurlungur: I like these at the bottom of the struct

src/interface/sparse_pack_base.hpp
```diff
@@ -0,0 +1,199 @@
...
+                 const std::vector<MetadataFlag> &flags, bool with_fluxes, bool coarse)
+      : vars(vars_in), use_regex(vars_in.size(), false), flags(flags),
+        with_fluxes(with_fluxes), coarse(coarse) {
+    PARTHENON_REQUIRE(!(with_fluxes && coarse),
+                      "Probably shouldn't be making a coarse pack with fine fluxes.");
+  }
+};
+} // namespace impl
+
+using namespace impl;
```
Yurlungur: Does this import `impl` into every header that imports `sparse_pack_base.hpp`? If so, I'd rather we move this `using` statement into a scope.

src/interface/sparse_pack_base.hpp
```diff
@@ -0,0 +1,199 @@
...
+    vidx = var_idx;
+    off = 0;
+    return *this;
+  }
+  KOKKOS_INLINE_FUNCTION
+  PackIdx(std::size_t var_idx, int off) : vidx(var_idx), off(off) {}
+  KOKKOS_INLINE_FUNCTION
+  std::size_t Vidx() { return vidx; }
+  KOKKOS_INLINE_FUNCTION
+  int Off() { return off; }
```
Yurlungur: These are ordered weirdly. Why are the two constructors not next to each other?

src/interface/sparse_pack_base.hpp
```diff
@@ -0,0 +1,199 @@
...
+  KOKKOS_INLINE_FUNCTION
+  PackIdx &operator=(std::size_t var_idx) {
+    vidx = var_idx;
+    off = 0;
+    return *this;
+  }
+  KOKKOS_INLINE_FUNCTION
+  PackIdx(std::size_t var_idx, int off) : vidx(var_idx), off(off) {}
+  KOKKOS_INLINE_FUNCTION
+  std::size_t Vidx() { return vidx; }
```
Yurlungur: Why not use `operator()` here instead of `Vidx()`?

src/interface/sparse_pack_base.hpp
```diff
@@ -0,0 +1,199 @@
...
+template <class T, class = typename std::enable_if<std::is_integral<T>::value>::type>
+KOKKOS_INLINE_FUNCTION PackIdx operator+(T off, PackIdx idx) {
+  return idx + off;
+}
+
+using SparsePackIdxMap = std::unordered_map<std::string, std::size_t>;
+
+class SparsePackBase {
+ protected:
+  friend class SparsePackCache;
```
Yurlungur, line 111: Does `SparsePackCache` not need a forward declaration?

src/interface/sparse_pack_base.hpp
```diff
@@ -0,0 +1,199 @@
...
+  KOKKOS_INLINE_FUNCTION
+  std::size_t Vidx() { return vidx; }
+  KOKKOS_INLINE_FUNCTION
+  int Off() { return off; }
+};
+
+template <class T, class = typename std::enable_if<std::is_integral<T>::value>::type>
+KOKKOS_INLINE_FUNCTION PackIdx operator+(PackIdx idx, T off) {
+  return PackIdx(idx.Vidx(), idx.Off() + off);
+}
```
Yurlungur: Nice :+1:

src/interface/sparse_pack_base.hpp
```diff
@@ -0,0 +1,199 @@
...
+  std::size_t vidx;
+  int off;
+
+ public:
+  KOKKOS_INLINE_FUNCTION
+  explicit PackIdx(std::size_t var_idx) : vidx(var_idx), off(0) {}
+  KOKKOS_INLINE_FUNCTION
+  PackIdx &operator=(std::size_t var_idx) {
+    vidx = var_idx;
+    off = 0;
```
Yurlungur: it wasn't immediately obvious to me that `off` was an `offset`. It might be better to rename it `offset` at the cost of 3 more characters.

src/interface/sparse_pack_base.hpp
```diff
@@ -0,0 +1,199 @@
...
+class SparsePackBase {
+ protected:
+  friend class SparsePackCache;
+
+  using test_func_t =
+      std::function<bool(int, const std::shared_ptr<CellVariable<Real>> &)>;
+  using pack_t = ParArray3D<ParArray3D<Real>>;
+  using bounds_t = ParArray3D<int>;
+  using alloc_t = std::vector<bool>;
+  using coords_t = ParArray1D<ParArray0D<Coordinates_t>>;
```
Yurlungur: Appreciate the typedefs.

src/interface/sparse_pack_base.hpp
```diff
@@ -0,0 +1,199 @@
...
+
+  pack_t pack_;
+  bounds_t bounds_;
+  coords_t coords_;
+
+  bool with_fluxes_;
```
Yurlungur: I don't know how I feel about the fluxpack being the same type as the variable pack. I *think* I like it.

src/interface/sparse_pack_base.cpp
```diff
@@ -0,0 +1,265 @@
...
+      if (alloc_status[i] != alloc_status_in[i]) goto make_new_pack;
+    }
+    return pack_map[ident].first;
+  }
+
+make_new_pack:
+  pack_map[ident] = {SparsePackBase::Build(pmd, desc),
+                     SparsePackBase::GetAllocStatus(pmd, desc)};
+  return pack_map[ident].first;
+}
```
Yurlungur: Is there a way we could avoid the `goto`s here?

src/interface/sparse_pack.hpp
Yurlungur, line 91: Is a virtual tag needed to avoid shadowing warnings? Or is that automatic for static?

src/interface/sparse_pack.hpp
```diff
@@ -0,0 +1,282 @@
...
+
+template <class... Ts>
+class SparsePack : public SparsePackBase {
+ public:
+  SparsePack() = default;
+
+  explicit SparsePack(const SparsePackBase &spb) : SparsePackBase(spb) {}
+
+  template <class T>
+  static SparsePack Make(T *pmd, const std::vector<MetadataFlag> &flags = {},
```
Yurlungur: It feels vaguely distasteful that it's `Make` here and `Build` in the base class. I assume this done to avoid shadowing though? Maybe there's a way around that?

src/interface/sparse_pack.hpp
```diff
@@ -0,0 +1,282 @@
...
+
+  KOKKOS_FORCEINLINE_FUNCTION
+  int GetNDim() const { return ndim_; }
+
+  KOKKOS_FORCEINLINE_FUNCTION
+  int GetDim(const int i) const {
+    assert(i > 0 && i < 6);
+    PARTHENON_REQUIRE(
+        i != 2, "Should not ask for the second dimension since it is logically ragged");
+    return dims_[i];
```
Yurlungur: Thinking ahead, if faces and cells and edges are all mixed in a pack, perhaps we should only allow one to ask for `GetDim(5)` and `GetDim(4)`.

My thinking here is that we always know the dim in the k, j, i dimensions. All we need to know are the numbers of blocks and the numbers of variables.

src/interface/sparse_pack.hpp
```diff
@@ -0,0 +1,282 @@
...
+
+  KOKKOS_INLINE_FUNCTION
+  const Coordinates_t &GetCoordinates(const int b) const { return coords_(b)(); }
+
+  // Bound overloads
+  KOKKOS_INLINE_FUNCTION int GetLowerBound(const int b) const { return bounds_(0, b, 0); }
+
+  KOKKOS_INLINE_FUNCTION int GetUpperBound(const int b) const {
+    return bounds_(1, b, nvar_ - 1);
+  }
```
Yurlungur, line 201: Should these have a default `b = 0`? That way when on a single meshblock, the user can neglect the block index

tst/unit/test_sparse_pack.cpp
```diff
@@ -0,0 +1,231 @@
...
+  BlockList_t block_list;
+  block_list.reserve(NBLOCKS);
+  for (int i = 0; i < NBLOCKS; ++i) {
+    auto pmb = std::make_shared<MeshBlock>(NSIDE, NDIM);
+    auto &pmbd = pmb->meshblock_data.Get();
+    pmbd->Initialize(pkg, pmb);
+    block_list.push_back(pmb);
+  }
+  return block_list;
+}
```
Yurlungur, line 56: I wonder if this should be in a `test_utils` file or something, as you're not the first to implement it.

tst/unit/test_sparse_pack.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+struct v1 : public parthenon::variables::base_t<false> {
```
Yurlungur: This is cool. But I kind of wonder if `base_t` is the right name for this class. I don't have a better one though. `name_t` maybe?

src/interface/update.cpp
```diff
@@ -60,24 +61,54 @@ template <>
...
+                if (ndim > 1)
+                  du += (coords.Area(X2DIR, k, j + 1, i) *
+                             vin.flux(b, X2DIR, l, k, j + 1, i) -
+                         coords.Area(X2DIR, k, j, i) * vin.flux(b, X2DIR, l, k, j, i));
+                if (ndim > 2)
+                  du += (coords.Area(X3DIR, k + 1, j, i) *
+                             vin.flux(b, X3DIR, l, k + 1, j, i) -
+                         coords.Area(X3DIR, k, j, i) * vin.flux(b, X3DIR, l, k, j, i));
+                du = -du / coords.Volume(k, j, i);
+              });
```
Yurlungur: I would move this into one of the examples, rather than putting it in `update`, that way we can test this in a self-contained way, without changing `update` permanently.

Yurlungur (commented):
Overall, I like this a lot. If this turns out to be the cleanest way to do things, I think it has implications as far as the sparse pool machinery (we can generalize it substantially and remove a lot of code) and it also has implications as far as the other variable/meshblock pack machinery.

I don't see why the sparse pack couldn't replace all the other pack machinery completely. The one concern I would have about that is the caching of per-meshblock data and whether that's worthwhile. We should maybe instrument that before deciding. Certainly your templated type/name machinery could make its way out of the sparse pack stuff to all packing.

One other thought: In the rest of the code, we have `PackVariables` as a member function of `MeshBlockData` and `MeshData`. I don't see why we couldn't do that here too. Just have the member functions call `PackType::Make` We could pass through all the relevant type parameters, in addition to `decltype(*this)` for the pack type. What do you think?

Yurlungur (approved):
I'm going to approve, as I think all my comments are non-blocking, but I'd like at least a few of them, especially the `goto`s addressed.

src/interface/sparse_pack.hpp
forrestglines: Agreed, this was a design obstacle throughout the existing variable packs that I was struggling to design around - each different face and edge will have a different dimension from cell or node centered variables.

src/interface/sparse_pack_base.cpp
lroberts36: https://imgflip.com/i/6mtqtc
To be fair it would be pretty easy to remove the `goto`s by defining a new function 
```c++
template <class T>
SparsePackBase &SparsePackCache::MakeNewPack(T *pmd, const PackDescriptor &desc) { 
  pack_map[ident] = {SparsePackBase::Build(pmd, desc),
                     SparsePackBase::GetAllocStatus(pmd, desc)};
  return pack_map[ident].first;
}
```
and replacing all of the `goto`s with `return MakeNewPack(pad, desc);`

src/interface/sparse_pack_base.cpp
```diff
@@ -0,0 +1,265 @@
...
+#include "interface/mesh_data.hpp"
+#include "interface/meshblock_data.hpp"
+#include "interface/sparse_pack_base.hpp"
+#include "interface/variable.hpp"
+#include "utils/utils.hpp"
+
+namespace {
+// SFINAE for block iteration so that sparse packs can work for MeshBlockData and MeshData
+template <class T, class F>
+inline auto IterateBlocks(T *pmd, F func) -> decltype(T().GetBlockData(0), void()) {
```
forrestglines: This function applies `F func` to every block in the MeshBlock/MeshData. Could this be renamed to `ForEachBlock` since this functions like `std::for_each`? Could you also add a comment about what this is doing?

```suggestion
//Apply a function to each block in a MeshBlockData or MeshData
template <class T, class F>
inline auto IterateBlocks(T *pmd, F func) -> decltype(T().GetBlockData(0), void()) {
```

src/interface/sparse_pack_base.hpp
```diff
@@ -0,0 +1,199 @@
...
+
+  // VAR_VEC can be:
+  //   1) std::vector<std::string> of variable names (in which case they are all assumed
+  //   not to be regexs)
+  //   2) std::vector<std::pair<std::string, bool>> of (variable name, treat name as
+  //   regex) pairs
+  template <class T, class VAR_VEC>
+  static std::tuple<SparsePackBase, SparsePackIdxMap>
+  Make(T *pmd, const VAR_VEC &vars, const std::vector<MetadataFlag> &flags = {},
+       bool fluxes = false, bool coarse = false) {
```
forrestglines:
```suggestion
  //Make a `SparsePackBase` with a corresponding `SparsePackIdxMap` from the provided `vars` and `flags`, creating the pack in `pmd->SparsePackCache` if it doesn't already exist. 
  //Note that the `SparsePackBack` will not yet be built and accessible on the device.
  // VAR_VEC can be:
  //   1) std::vector<std::string> of variable names (in which case they are all assumed
  //   not to be regexs)
  //   2) std::vector<std::pair<std::string, bool>> of (variable name, treat name as
  //   regex) pairs
  template <class T, class VAR_VEC>
  static std::tuple<SparsePackBase, SparsePackIdxMap>
  Make(T *pmd, const VAR_VEC &vars, const std::vector<MetadataFlag> &flags = {},
       bool fluxes = false, bool coarse = false) {
```

forrestglines: Since `vars`, `flags`, `fluxes`, and `coarse` are just passed into a `PackDescriptor`, I think it would be preferable to just accept a `PackDescriptor` as an argument, not these individual arguments.

src/interface/sparse_pack_base.hpp
```diff
@@ -0,0 +1,199 @@
...
+  static std::tuple<SparsePackBase, SparsePackIdxMap>
+  MakeWithCoarse(T *pmd, const VAR_VEC &vars,
+                 const std::vector<MetadataFlag> &flags = {}) {
+    const bool fluxes = false;
+    const bool coarse = true;
+    return Make(pmd, vars, flags, fluxes, coarse);
+  }
+};
+
+class SparsePackCache {
```
forrestglines:
```suggestion
//Map of `PackDescriptor` to `SparsePackBase`
class SparsePackCache {
```

src/interface/sparse_pack_base.hpp
```diff
@@ -0,0 +1,199 @@
...
+
+  bool with_fluxes_;
+  bool coarse_;
+  int nblocks_;
+  int ndim_;
+  int dims_[6];
+  int nvar_;
+
+  template <class T>
+  static alloc_t GetAllocStatus(T *pmd, const PackDescriptor &desc);
```
forrestglines:
```suggestion
//Get a list of booleans of the allocation status of every variable in pmd matching the PackDescriptor desc
  template <class T>
  static alloc_t GetAllocStatus(T *pmd, const PackDescriptor &desc);
```

src/interface/sparse_pack_base.hpp
```diff
@@ -0,0 +1,199 @@
...
+
+  template <class T>
+  static SparsePackBase Build(T *pmd, const PackDescriptor &desc);
```
forrestglines:
```suggestion
  //Build a usable `SparsePackBase` from the variables specified in desc contained in the blocks  in a MeshBlockData/MeshBlock with a variable pack allocated on the device.
  template <class T>
  static SparsePackBase Build(T *pmd, const PackDescriptor &desc);
```

src/interface/sparse_pack_base.hpp
```diff
@@ -0,0 +1,199 @@
...
+                 const std::vector<MetadataFlag> &flags = {}) {
+    const bool fluxes = true;
+    const bool coarse = false;
+    return Make(pmd, vars, flags, fluxes, coarse);
+  }
+
+  template <class T, class VAR_VEC>
+  static std::tuple<SparsePackBase, SparsePackIdxMap>
+  MakeWithCoarse(T *pmd, const VAR_VEC &vars,
+                 const std::vector<MetadataFlag> &flags = {}) {
```
forrestglines: Is there a reason to make these separate functions instead of just having `Make`? If the boolean flags do the job, would fewer functions would lead to a simpler/narrower interface? Or are explicitly named `Make` functions more useful?

src/interface/sparse_pack.hpp
```diff
@@ -0,0 +1,282 @@
...
+                         bool fluxes = false, bool coarse = false) {
+    auto &cache = pmd->GetSparsePackCache();
+    return SparsePack(cache.Get(
+        pmd, PackDescriptor(std::vector<std::string>{Ts::name()...},
+                            std::vector<bool>{Ts::regex()...}, flags, fluxes, coarse)));
+  }
+
+  template <class T, class VAR_VEC>
+  static std::tuple<SparsePack, SparsePackIdxMap>
+  Make(T *pmd, const VAR_VEC &vars, const std::vector<MetadataFlag> &flags = {},
```
forrestglines:
```suggestion
  //Make a `SparsePack` with a corresponding `SparsePackIdxMap` from the provided `vars` and `flags`, creating the pack in `pmd->SparsePackCache` if it doesn't already exist. 
  //The pack will be created and accessible on the device
  // VAR_VEC can be:
  //   1) std::vector<std::string> of variable names (in which case they are all assumed
  //   not to be regexs)
  //   2) std::vector<std::pair<std::string, bool>> of (variable name, treat name as
  //   regex) pairs
  template <class T, class VAR_VEC>
  static std::tuple<SparsePack, SparsePackIdxMap>
  Make(T *pmd, const VAR_VEC &vars, const std::vector<MetadataFlag> &flags = {},
```

src/interface/sparse_pack_base.cpp
```diff
@@ -0,0 +1,265 @@
...
+    for (int i = 0; i < alloc_status_in.size(); ++i) {
+      if (alloc_status[i] != alloc_status_in[i]) goto make_new_pack;
+    }
+    return pack_map[ident].first;
+  }
+
+make_new_pack:
+  pack_map[ident] = {SparsePackBase::Build(pmd, desc),
+                     SparsePackBase::GetAllocStatus(pmd, desc)};
+  return pack_map[ident].first;
```
forrestglines: I found it convoluted that the pack is first "made" but not accessible on the device, then implicitly "built" in order to function on the device via a call to `SparsePackCache::Get`. Maybe document this for developers somewhere?

src/interface/sparse_pack.hpp
```diff
@@ -0,0 +1,282 @@
...
+       bool fluxes = false, bool coarse = false) {
+    static_assert(sizeof...(Ts) == 0);
+    auto out = SparsePackBase::Make(pmd, vars, flags, fluxes, coarse);
+    return {SparsePack(std::get<0>(out)), std::get<1>(out)};
+  }
+
+  template <class T>
+  static SparsePack MakeWithFluxes(T *pmd, const std::vector<MetadataFlag> &flags = {}) {
```
forrestglines: I'm conflicted about whether these additional `MakeWithFluxes` and associated functions make the interface more complicated and harder to modify if needed since they just pass booleans to `Make`, or whether the explicitly purpose make functions are handy.

src/interface/update.cpp
forrestglines: Does this affect performance in anyway of existing examples or downstream codes? I'd be hesitant to make this change here if there are any performance degradations.

forrestglines (commented):
Looks good! There are a few changes that I would *like* to have included, the only blocking matter being whether using this pack in `FluxDivergence` degrades performance in downstream codes.

Another more general problem that I see with the larger code base, not just this commit, is that we're not doing a great job of commenting what code is doing for the sake of developers, not just downstream users. It's sometimes hard to figure out how pieces fit together.

src/interface/sparse_pack_base.cpp
```diff
@@ -0,0 +1,265 @@
...
+}
+
+// Specialize for the only two types this should work for
+template SparsePackBase
+SparsePackBase::Build<MeshBlockData<Real>>(MeshBlockData<Real> *, const PackDescriptor &);
+template SparsePackBase SparsePackBase::Build<MeshData<Real>>(MeshData<Real> *,
+                                                              const PackDescriptor &);
+
+inline SparsePackBase::test_func_t
+SparsePackBase::GetTestFunction(const PackDescriptor &desc) {
```
forrestglines: Is there a good reason to create a test function rather than just having a test function? To me, it would make more sense for the `PackDescriptor` to have such a test function built in. The regexs could be built in the `PackDescriptor` constructor, which shouldn't be a problem unless that descriptor needs to be pushed to the device.

Also, "GetTestFunction" is an ambiguous name, it doesn't suggest that it returns a function that tests whether a `CellVariable` is included by the provided `PackDescriptor`.

forrestglines:
Generally, I like this design for the variable pack a lot more than the current design to have different design packs for just variables and variables with fluxes.  If this is as performant (or nearly performant) as the currently used variable packs, I think these should be condensed into this one variable pack.

For now, I'm making edge and face variables work off this new sparse variable pack.

src/interface/sparse_pack.hpp
```diff
@@ -0,0 +1,282 @@
...
+struct IncludesType<T, U, Ts...> : IncludesType<T, Ts...> {};
+
+} // namespace impl
+
+using namespace impl;
+
+namespace variables {
+// Struct that all variables types should inherit from
+template <bool REGEX, int... NCOMP>
+struct base_t {
```
forrestglines: On second read through, I don't understand what is going on here. I see that `variables::any` is used in `update.cpp`, but I don't understand the point of these types. This comment suggests to me that `CellVariable` inherit from `base_t` which is not the case. What is "variable types" referring to here?

src/interface/update.cpp
lroberts36: I don't think it does, but I agree we should leave it out of this PR. I had added this to test that the packs worked for a number of different problems. I will revert.

src/interface/meshblock_data.hpp
lroberts36: This is just required for the test to work, since you can't create a sparse variable without creating it in a sparse pool apparently. I don't think this restriction on creating sparse variables makes sense and should be removed (since it is easy to think of places where you might want sparse variables that are not part of a pool), but I think that should be for another pull request. I think we should leave it commented out for now so it isn't forgotten, but I don't think anyone is really stressing the sparse stuff right now so I think it is ok to disable the check for now.

src/interface/sparse_pack.hpp
lroberts36: This is essentially a trick for indexing into the pack inside of a kernel on gpu. Rather than pass variable names to the pack, we create a pack that takes each variable struct as a template parameter. Each of these variable types ahas a method that returns the name of the variable, so you can create a vector of variable names and create a pack like normal. The advantage is that on device you can just pass an object of a particular variable type as one of the pack indices and it will internally find that index of that variable in the pack for you. This reduces a lot of boilerplate and is particularly helpful for sparse variables. It should be totally independent of `CellVariable`s 

I agree the naming could be clearer. If anyone has ideas, I am certainly open to them.

src/interface/sparse_pack.hpp
lroberts36: I think it just happens automatically and I don't think there is any virtual look up table for static functions. I have not seen any shadowing warnings.

lroberts36: see above

src/interface/sparse_pack.hpp
lroberts36: Mostly it was just which word jumped to my mind when I wrote these things at separate times, although `Build` in the base class is truly building a pack while `Make` is potentially just getting a pre-existing pack from the cache.

src/interface/sparse_pack.hpp
lroberts36: Personally, I think it makes for a cleaner interface since I think it can be confusing when a function takes multiple booleans  (since reading `func(true, false, true)` is often unclear). Additionally, I think we want to prevent people from makeing packs with fluxes and coarse variables unless they really want to.

src/interface/sparse_pack.hpp
lroberts36: I think that makes sense, I was just trying to partially adhere to the pre-existing pack interface. I will remove this.

src/interface/sparse_pack.hpp
lroberts36: Yeah, I think that makes sense although maybe we want to add an assert in that case the `nblock==0`

src/interface/sparse_pack_base.cpp
lroberts36: Yeah, I think renaming it `ForEachBlock` is a good idea for consistency with the stdlib and for consistency with the boundary iteration machinery.

src/interface/sparse_pack_base.cpp
lroberts36: Agreed that it might be good to have somewhere else. Any ideas on where it should go?

src/interface/sparse_pack_base.hpp
lroberts36: Oh, good point. I will move that.

src/interface/sparse_pack_base.hpp
lroberts36: I think this serves as a forward declaration (?)

src/interface/sparse_pack_base.hpp
lroberts36: Unless there is a performance hit I think I like it, it certainly reduces code complexity in my mind.

src/interface/sparse_pack_base.cpp
lroberts36: I have to think about it a little bit more, but I think you are right that this could just be a member function of `PackDescriptor`. `GetTestFunction` predates `PackDescriptor` in the development process, so there may be no good reason for it anymore. It was just there to easily create a reusable functor and reduce code duplication.

src/interface/sparse_pack_base.cpp
lroberts36: `Make` should always return a pack that is accessible on device. Without cacheing, `Make` would just call `Build` for `SparsePackBase` directly. `SparsePackCache::Get` is just a method that checks if there is a cached version of the pack and returns a reference to that, or if there isn't one builds it, adds it to the cache, and returns a reference.

src/interface/sparse_pack_base.hpp
lroberts36: I am not sure about the best way forward. I think `SparsePack`s have essentially all of the same functionality of the pre-existing packing machinery, but having them be the only packs would mean that a lot of other code needs to be changed.

src/interface/sparse_pack_base.hpp
lroberts36: Yeah, I will change this for consistency with the rest of the code.

src/interface/sparse_pack_base.hpp
lroberts36: will move

src/interface/sparse_pack_base.hpp
lroberts36: good point, will change

src/interface/sparse_pack_base.hpp
lroberts36: leftover from debugging, will reorder

tst/unit/test_sparse_pack.cpp
lroberts36: Yeah, actually I just copied and pasted this from another test.

tst/unit/test_sparse_pack.cpp
lroberts36: Maybe the namespace should have a different name? `variable_name_types`?

src/interface/meshblock_data.hpp
Yurlungur: :+1:

src/interface/sparse_pack_base.cpp
Yurlungur: Didn't you make a set of looping utilities for `bvals_cc_in_one` in your sparse branch? Maybe unify this with those and put it all in a header file in utils called something like `block_utils` or similar?

src/interface/sparse_pack_base.hpp
Yurlungur: I lean towards replacing the old packs completely. But agreed htat that's a surgery. Once this is all merged in I think we should have a separate PR that does this, which probably also implies a code version bump. Thoughts, @jdolence @forrestglines @pgrete ?

src/interface/sparse_pack_base.hpp
Yurlungur: Oh interesting. I didn't realize that. I've always added the forward declaration separately.

src/interface/sparse_pack_base.cpp
Yurlungur: Haha. It's a matter of taste, but I would prefer a new function and no gotos.

src/interface/sparse_pack.hpp
Yurlungur: Maybe the public version should be called `Get` then? Or `PackVariables` for consistency with the rest of the code?

src/interface/sparse_pack.hpp
Yurlungur: :+1: I like perhaps just a method for `GetNumBlocks` and your per-variable dimension methods that you already have.

tst/unit/test_sparse_pack.cpp
Yurlungur: that would work too, but also feels kind of verbose. I like `parthenon::variables::name_t`.

Yurlungur: Could also do `variable_names::base_t`?

src/interface/sparse_pack_base.cpp
lroberts36: Yeah, I think that is a good idea. Maybe we should do that in a separate pull request after everything is merged together though. I think this could benefit from the concepts stuff as well.

Yurlungur:
> Generally, I like this design for the variable pack a lot more than the current design to have different design packs for just variables and variables with fluxes. If this is as performant (or nearly performant) as the currently used variable packs, I think these should be condensed into this one variable pack.
> 
> For now, I'm making edge and face variables work off this new sparse variable pack.

I already mentioned in my own review but I'm strongly in favor of throwing away the old packs once this is in and known to be working performantly.

src/interface/sparse_pack_base.hpp
lroberts36: Yeah, we also need to do performance testing as well before making such a big change I think

Yurlungur: Yep absolutely. :+1:

src/interface/sparse_pack.hpp
lroberts36: I think I like `Get` since it is always called as `SparsePack<...>::Get(...)`.

tst/unit/test_sparse_pack.cpp
lroberts36: I think `variable_names` is a good compromise between descriptiveness and verboseness. @forrestglines do you have an opinion?

src/interface/sparse_pack_base.cpp
lroberts36: I created issue #696 for this

src/interface/sparse_pack_base.hpp
lroberts36: To me, it is not obvious exactly what `operator()` should return for a `PackIdx`. It could be interpreted to be the variable index as you suggest, but I think that opens up the possibility of confusion. Since it should not be called anywhere in downstream code (so a longer name won't make things much more verbose), I think it is better to have a descriptive name. I have changed it to `VariableIdx()` for better clarity.

src/interface/sparse_pack_base.hpp
lroberts36: done

src/interface/sparse_pack.hpp
lroberts36: Switched to `Get`

src/interface/sparse_pack.hpp
lroberts36: On second though, I think allowing for default `b=0` basically double the number of operator overloads in `SparsePack` and make things much less readable. For now, lets force the user to explicitly write the block index. If it becomes clear that we need these overloads, we can add them in a future PR.

tst/unit/test_sparse_pack.cpp
lroberts36: For now I am going with `variable_names::base_t`, but not married to that choice

src/interface/sparse_pack_base.cpp
lroberts36: done

src/interface/sparse_pack_base.hpp
lroberts36: I reworked the `SparsePackBase` interface a bit if you want to take a look again. I think it should address you concern here.

src/interface/sparse_pack_base.hpp
lroberts36: I think the explicitly named `Make` (now `Get`) functions in `SparsePack` make downstream code clearer. Since `SparsePackBase` shouldn't be used explicitly in downstream code, I have removed all of the public `Make` static methods here

src/interface/sparse_pack_base.cpp
lroberts36: Yep, your suggestion is definitely cleaner. I have made the change.

src/interface/sparse_pack_base.cpp
lroberts36: switched :'(

src/interface/sparse_pack_base.cpp
lroberts36: I changed this around a little bit. Take a look and see if the cacheing machinery makes more sense now.

src/interface/update.cpp
lroberts36: reverted.

lroberts36:
@forrestglines: I think this is ready for your re-review at this point. If you are ok with it, go ahead and merge.  

Also added an issue #697 to add documentation for sparse packs

src/interface/sparse_pack_base.hpp
Yurlungur: :+1:

src/interface/sparse_pack.hpp
Yurlungur: Is this actually true? Why not just add a default argument, rather than adding an overload?

lroberts36: I think for these three you can add just the default argument, but I think the interface should be consistent so that if you can elide the block number here you should be able to do it in all of the `operator()` methods.

Yurlungur: Ah, I see. Yes that adds some additional complexity. I think that would be nice. But maybe not necessary for now.

src/interface/sparse_pack.hpp
forrestglines: Ok, this makes more sense. I can see how that might be faster on the device. (How does that index end up on the device after compiling though? Does it end up in the instruction code for the kernel? I'm curious but I don't think this is important to know).

I'm not sure about a different name. `variable_names` for a namespace seems ok.

forrestglines (approved):

src/interface/sparse_pack.hpp
lroberts36: The mechanism for getting an index in the pack from a type in `variable_names` is probably easiest to see in `GetLowerBound`: 
```C++
  template <class TIn, REQUIRES(IncludesType<TIn, Ts...>::value)>
  KOKKOS_INLINE_FUNCTION int GetLowerBound(const int b, const TIn &) const {
    const int vidx = GetTypeIdx<TIn, Ts...>::value;
    return bounds_(0, b, vidx);
  }
  ```
`GetTypeIdx<...>` just searches the list of types provided as template parameters to the `sparse_pack`, `Ts...`, for the type `TIn` and returns its integet index in the parameter pack `Ts...` if it is present in the parameter pack (compilation should fail if the type is not in the parameter pack). This integer index can then be used like normal for indexing into the pack (except for sparse variables the upper and lower index bounds for a particular variable may change with different blocks). 

I don't think there is any speed advantage to this approach, but it makes code significantly less verbose than creating a parameter pack, pulling out indices on host, capturing them to device in the kernel, and then using those indices in the kernel. All you have to do with the type based pack is 
```c++
auto my_pack = sparse_pack<var1_t, var2_t, ...>::Make(md); 
par_for(...,[](int b, int k, int j, int i) {
  my_pack(b, var1_t(), k, j, i) = my_pack(b, var2_t(), k, j, i);
});
```

forrestglines (approved):

forrestglines:
Is this ready to merge? I think it is. Do we need to do something to get that "ci/gitlab/gitlab.com" check to go through?

lroberts36:
Yeah, I think it should automatically merge once the gitlab CI runs, but that CI is broken right now for everything. There is a way to force a merge I think, but I am not sure how that interacts github.

forrestglines:
As I've been testing out using this pack for face and edge centered fields, I've realized that the index order for `SparsePack:operator()` is a bit unintuitive for me. Currently we have
```
v( dir, meshblock, var, k, j, i)
```
In my head, the following organization seems more intuitive
```
v( var, dir, meshblock, k, j, i)
```
So that the first two indicies are to index the variable and the last four are to index the location in space.

lroberts36:
I am not sure what the best way to order this is. I chose the current ordering for regular variables to be the same as `MeshBlockPack`s. I don't think there is a way to directly access fluxes in `MeshBlockPack`s, so there it appears there are no `flux` calls that include direction, block, variable, and kji currently (@Yurlungur correct me if I am wrong). I think your ordering makes logical sense, but I would be concerned that it requires a lot of index reordering if someone where to update their code from the old packing machinery.

forrestglines:
> I don't think there is a way to directly access fluxes in `MeshBlockPack`s, so there it appears there are no `flux` calls that include direction, block, variable, and kji currently

I agree, that sounds correct. However, that will change with face and edge centered fields. I considered adding another `comp` for "component" to complement `flux` but I found it a bit redundant

> but I would be concerned that it requires a lot of index reordering if someone where to update their code from the old packing machinery.

I suspect that too. I think the `v( meshblock, var, k, j, i)` is likewise awkward although it reflects the order of things in memory. However, it's probably too much bug-prone work to change now. If we could automate that refactor in downstream codes with a script, I would advocate for that change later on.

PR #693, opened 2022-07-15
WIP: Face and Edge Centered Fields

forrestglines:
Adds the necessary machinery for face and edge centered fields (i.e. variable packs, boundary communication, and edge-centered AMR corrections)

## PR Summary

Current plan is to use the new `SparsePack` from #692 to handle the different directions for face and edge centered fields. 

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted


example/loop_advection/loop_advection_package.cpp
```diff
@@ -0,0 +1,368 @@
...
+
+  auto pkg = pmb->packages.Get("loop_advection_package");
+
+  const auto &v1 = pkg->Param<Real>("v1");
+  const auto &v2 = pkg->Param<Real>("v2");
+  const Real v3 = 0;
+
+  auto &v =  SparsePack<B_face,B_cell,E_cell,E_edge>::Make(pmd->meshblock_data.Get());
+
+  using dBdt_face = B_face;
```
lroberts36, line 237: I don't see where `dBdt_face` is used

example/loop_advection/loop_advection_package.cpp
```diff
@@ -0,0 +1,368 @@
...
+
+namespace loop_advection_example {
+
+
+struct B_face : public parthenon::variable_names::base_t<false> {
+  template <class... Ts>
+  KOKKOS_INLINE_FUNCTION B_face(Ts &&...args)
+      : parthenon::variable_names::base_t<false>(std::forward<Ts>(args)...) {}
+  static std::string name() { return "B_face"; }
+};
```
lroberts36: This looks like the right approach to creating `variable_name` types, although I think we may want to create a macro to reduce boiler plate at some point.

example/loop_advection/loop_advection_package.cpp
```diff
@@ -0,0 +1,368 @@
...
+// Compute v x B to compute E, then take the curl to compute dBdt
+parthenon::TaskStatus calc_dBdt(parthenon::MeshBlock *pmb){
+
+  auto pkg = pmb->packages.Get("loop_advection_package");
+
+  const auto &v1 = pkg->Param<Real>("v1");
+  const auto &v2 = pkg->Param<Real>("v2");
+  const Real v3 = 0;
+
+  auto &v =  SparsePack<B_face,B_cell,E_cell,E_edge>::Make(pmd->meshblock_data.Get());
```
lroberts36, line 235: This looks like the right way to create a `SparsePack`

example/loop_advection/loop_advection_package.cpp
```diff
@@ -0,0 +1,368 @@
...
+  {
+
+    pmb->par_for(
+      "LoopAdvection::ProblemGenerator:calc_E_cell", 0, vf.dim(5)-1, kb.s, kb.e, jb.s, jb.e, ib.s, ib.e,
+      KOKKOS_LAMBDA(const int k, const int j, const int i) {
+
+        //Average B from the faces to the cell centers
+        v(b, B_cell(0), k, j, i) = 0.5 * (v( 1, b, B_face, k, j, i) + v( 1, b, B_face,  k, j, i+1));
+        v(b, B_cell(1), k, j, i) = 0.5 * (v( 2, b, B_face, k, j, i) + v( 2, b, B_face,  k, j+1, i));
+        v(b, B_cell(2), k, j, i) = 0.5 * (v( 3, b, B_face, k, j, i) + v( 3, b, B_face,  k+1, j, i));
```
lroberts36: When accessing a pack using a `variable_name` type, you need to instantiate the `variable_name` object by calling its constructor, so for instance this would need to change to 
```suggestion
        v(b, B_cell(0), k, j, i) = 0.5 * (v( 1, b, B_face(), k, j, i) + v( 1, b, B_face(),  k, j, i+1));
        v(b, B_cell(1), k, j, i) = 0.5 * (v( 2, b, B_face(), k, j, i) + v( 2, b, B_face(),  k, j+1, i));
        v(b, B_cell(2), k, j, i) = 0.5 * (v( 3, b, B_face(), k, j, i) + v( 3, b, B_face(),  k+1, j, i));
```
It looks like there are a number of other places a similar change is required.

example/loop_advection/loop_advection_package.cpp
```diff
@@ -0,0 +1,368 @@
...
+  //(Do it in one loop for demonstrative purposes)
+  {
+
+    pmb->par_for(
+      "LoopAdvection::ProblemGenerator:calc_E_edge", 0, vf.dim(5)-1, kb.s, kb.e, jb.s, jb.e, ib.s, ib.e,
+      KOKKOS_LAMBDA(const int k, const int j, const int i) {
+        //Average E from the cell centers to cell edges
+        if( k+1 < kb.e && j+1 < jb.e) 
+          v( 1, b, E_edge, k, j, i) = v( 1, b, E_face,   k,   j,   i) + v( 1, b, E_face,   k, j+1,   i);
+                                    + v( 1, b, E_face, k+1,   j,   i) + v( 1, b, E_face, k+1, j+1,   i);
```
lroberts36: I am not clear about how the indexing works for faces, edges, and nodes (i.e. for the `i` face does `i -> i-1/2`, `j->j`, and `k->k`? And for the `i` edge does `i -> i`, `j -> j-1/2`, and `k -> k-1/2`? If so, shouldn't the offsets here be `j-1` instead of `j+1` and similarly for `k`?). It may be useful to start a doc describing how this all works and also describing what other changes need to be made to the code to include face, edge, and vertex variables (is everything going to be unified in `CellVariable`? What needs to be changed about boundary communication? What changes need to be made to prolongation and restriction? etc.)

example/loop_advection/loop_advection_package.cpp
forrestglines: I thought that too as I wrote these. Fortunately, I'm guessing defining `variable_name`s in downstream codes will be fairly constrained so there won't be too much boilerplate.

example/loop_advection/loop_advection_package.cpp
forrestglines: That indexing is correct, to keep the different arrays for faces/edges all starting at 0.

I'm currently designing around a `FaceVariable` and `EdgeVariable` which keeps separate arrays for each dimension. I first tried working with `CellVariable` but it was awkward.

There are still a lot of changes to prolongation and restriction to be made. There are a few hooks for those in the `loop_advection` example but nothing implemented yet.

lroberts36: image

This is how I thought the indexing would work (for a 2D slice), where `c` are the cell centered values of some scalar, `f_x` is the x-face centered value of the scalar, `e_z` is the z edge (which I guess could also be called the xy edge) value of the field.

example/loop_advection/loop_advection_package.cpp
```diff
@@ -0,0 +1,368 @@
...
+
+
+  //Initialize magnetic field from curl of vector potential
+  //Computing each direction individually (not necessarily the fastest method)
+  for( int face_dir=1; face_dir <= 3; face_dir++){
+    CartDir face_dir_enum = static_cast<CartDir>(face_dir);
+
+    IndexRange ib = pmb->cellbounds.GetFaceBoundsI(IndexDomain::entire,face_dir_enum);
+    IndexRange jb = pmb->cellbounds.GetFaceBoundsJ(IndexDomain::entire,face_dir_enum);
+    IndexRange kb = pmb->cellbounds.GetFaceBoundsK(IndexDomain::entire,face_dir_enum);
```
Yurlungur: I feel torn on adding `GetFaceBounds` is it not always GetBounds* + (DIR == FACEDIR)? On the other hand, maybe the generality of the infrastructure is worth supporting the "right" way for things like boundary conditions?

Similar question for `GetEdgeBounds` and `GetNodeBounds` I suppose.

example/loop_advection/loop_advection_package.cpp
```diff
@@ -0,0 +1,368 @@
...
+                                                           j - (cart_dir_p1 == 2),
+                                                           i - (cart_dir_p1 == 1)))/dp1;
+
+        //Second term: dEydz, dEzdx, or dExdy
+        const Real dEp1dp2 =(v( cart_dir_p1, b, A_face(), k, j, i)
+                            -v( cart_dir_p1, b, A_face(), k - (cart_dir_p2 == 3),
+                                                          j - (cart_dir_p2 == 2),
+                                                          i - (cart_dir_p2 == 1)))/dp2;
+
+        v( cart_dir, b, B_face(), k, j, i) = -(dEp2dp1 - dEp1dp2);
```
Yurlungur: wonder if we should have some machinery for computing divergences and curls automatically assuming second-order

example/loop_advection/loop_advection_package.cpp
```diff
@@ -0,0 +1,368 @@
...
+                                           coords.dx3(k);
+        const Real dp2 = (cart_dir == 1) ? coords.dx1(i) :
+                        +(cart_dir == 2) ? coords.dx2(j) :
+                                           coords.dx3(k);
+
+        //First term: dEzdy, dExdz, or dEydx
+        const Real dEp2dp1 =(v( cart_dir_p2, b, A_face(),  k, j, i)
+                            -v( cart_dir_p2, b, A_face(),  k - (cart_dir_p1 == 3),
+                                                           j - (cart_dir_p1 == 2),
+                                                           i - (cart_dir_p1 == 1)))/dp1;
```
Yurlungur: It feels very weird to me that the direction is the outermost index, rather than the meshblock. I haven't looked at the packing machinery yet, so maybe I'll change my mind after looking at that.

I also wonder if we could generalize Luke's template name stuff so that we could have
`v(b, A_face(cart_dir_p2), k, j, i)` so it looks identical to cell variables?

example/loop_advection/loop_advection_package.cpp
```diff
@@ -0,0 +1,368 @@
...
+          v( 2, b, dBdt_face(), k, j, i) = (v( 1, b, E_edge(), k+1, j, i) - v( 1, b, E_edge(), k, j, i))/coords.dx3(k)
+                                         - (v( 3, b, E_edge(), k, j, i+1) - v( 3, b, E_edge(), k, j, i))/coords.dx1(i);
+        if( j+1 < jb.e && k+1 < kb.e) 
+          v( 3, b, dBdt_face(), k, j, i) = (v( 2, b, E_edge(), k, j, i+1) - v( 2, b, E_edge(), k, j, i))/coords.dx3(k)
+                                         - (v( 1, b, E_edge(), k, j+1, i) - v( 1, b, E_edge(), k, j, i))/coords.dx2(j);
+      });
+  }
+
+  //After averaging from faces to edges, reduce bounds by 1 - so dBdt doesn't include ghosts
+  //FIXME(forrestglines): How to shrink bounds?
```
Yurlungur, line 311: Not sure this is something we'd want to thread through the `mesh/domain` stuff. Maybe this is a time when a `-1` hardcoded into the loop bounds would be reasonable?

That said, telescoping ghost zones would be a useful capability so maybe worth thinking more carefully about it

src/parthenon_arrays.hpp
```diff
@@ -99,7 +99,7 @@ using FaceField = FaceArray<Real>;
 
 template <typename T>
 struct EdgeArray {
-  ParArrayND<Real> x1e, x2e, x3e;
+  ParArrayND<T> x1e, x2e, x3e;
```
Yurlungur, line 102: looks like you're using the struct of vars, rather than a single big var? or tying them together somehow in metadata?

src/mesh/domain.hpp
```diff
@@ -155,6 +162,61 @@ class IndexShape {
     return (domain == IndexDomain::interior) ? x_[2] : IndexRange{ks(domain), ke(domain)};
   }
 
+  //FIXME(forrestglines) How necessary are these additional functions?
```
Yurlungur: I'm torn on this, but I think we can probably do without them.

src/interface/variable.hpp
```diff
@@ -159,6 +159,7 @@ class CellVariable {
 /// and label so that we can refer to variables by name.  Since Athena
 /// currently only has scalar Face fields, we also only allow scalar
 /// face fields
+//  FIXME(forrestglines): What is it meant by only "scalar" Face fields? Athena++ ONLY has vector face fields (x,y,z for each face)
```
Yurlungur: I believe the author of this comment (Sriram maybe?) is referring to the fact that edge-centered variables are split into separate arrays per direction?

It may also be a (incorrect?) assumption that tensor-valued fields are not supported--i.e., you can't have a face-centered field where there's 5 indices for, say, chemical abundances, as well as the 3 directional indices. I think something may have been lost in translation between Athena++, where all variables are packed together, and Parthenon, where they are registered separately.

src/interface/variable.hpp
```diff
@@ -258,6 +261,33 @@ class EdgeVariable {
...
+  }
+  template <typename... Args>
+  KOKKOS_FORCEINLINE_FUNCTION T &operator()(int dir, Args... args) const {
+    assert(1 <= dir && dir <= 3);
+    if (dir == 1) return data.x1e(std::forward<Args>(args)...);
+    if (dir == 2)
+      return data.x2e(std::forward<Args>(args)...);
+    else // dir == 3
+      return data.x3e(std::forward<Args>(args)...);
+  }
```
Yurlungur: This is a longer discussion but: if we always interact with variable packs or `ParArray`s, I wonder if the `operator()` methods should just be removed from variables.

src/interface/sparse_pack_base.cpp
```diff
@@ -167,6 +167,64 @@ SparsePackBase SparsePackBase::Build(T *pmd, const PackDescriptor &desc) {
...
+                  PARTHENON_REQUIRE(
+                      pack_h(0, b, idx).size() > 0,
+                      "Seems like this variable might not actually be allocated.");
+                  idx++;
+                }
+              }
+            }
+          }
+        }
+      }
```
Yurlungur: I don't love having to add all this essentially identical boilerplate. If we moved to all variables being the same type, i.e., make them not structs anymore, this could all go away.

Alternatively, if you're married to the struct machinery (not my preference), we could probably make some generic templated functions for doing this looping.

Yurlungur (commented):
So it looks like you went down the path of structs for the face and edge centered variables, rather than tying them together somehow in metadata or simply wasting space and making `NDIMx(NX1 + 1)x(NX2 + 1)x(NX3 + 1)` vars? Was there a reason for sticking with the struct design?

Other than that, the packing machinery looks reasonable to me. I have  a few comments below.

Yurlungur:
> There are still a lot of changes to prolongation and restriction to be made. There are a few hooks for those in the loop_advection example but nothing implemented yet.

You may want to wait until my `prolongation-in-one` stuff is in place (hopefully soon, it's dragged on a bit). I have something in mind here. We can chat about it in the call tomorrow, or in a separate call if you like.

example/loop_advection/loop_advection_package.cpp
forrestglines: Yes, that's the indexing I'm going for. In memory, each edge and face variable would start at 0 for each dimension. Do you see an off-by-one error here that I'm not seeing? I wouldn't be surprised.

forrestglines:
> So it looks like you went down the path of structs for the face and edge centered variables, rather than tying them together somehow in metadata or simply wasting space and making `NDIMx(NX1 + 1)x(NX2 + 1)x(NX3 + 1)` vars? Was there a reason for sticking with the struct design?

I've only partially-implemented that design since I've encountered the least amount of resistance going down the `FaceVariable`, `EdgeVariable`, `NodeVariable` path rather than everything into `CellVariable`. I'm open to changing direction, although I had difficulty figuring out exactly what needed to change to fit faces and edges into `CellVariable`. There's a lot going on in `CellVariable` and the functions that use it that I don't understand, especially regarding sparsity, allocation, and fluxes. I can't tell how implemented face or edge centered fields directly into the `CellVariable` - where the first dimension is a direction, the following dimensions having ragged indexing - would affect the code I don't understand well. Would creating `face_data_` and `edge_data_` inside of `CellVariable` much like there currently exists a `flux_data_` make sense?

example/loop_advection/loop_advection_package.cpp
forrestglines: Yes, that is the correct math that will always get called. It's a question of whether we want that `+ (dir == face_dir)` inside these bounds functions or outside when the loops get called. I have a slight preference for putting that logic in one place rather than everywhere it's needed.

example/loop_advection/loop_advection_package.cpp
forrestglines: Outside of this demo, those functions would probably be useful.

example/loop_advection/loop_advection_package.cpp
forrestglines: I mentioned this in #692 - I also find the ordering with mesh block awkward as well, but I think it should go after the variable idx since the mesh block refers to a location in space much like `k`, `j`, `i`. If I could rewrite things, I would use `v( A_face(cart_dir_p2), b, k, j, i)` to group the variable indices and spatial indices together.

I like the idea of using the template name machinery, although I don't understand it well enough to implement it like that.

example/loop_advection/loop_advection_package.cpp
forrestglines: This was a note mostly to myself to see if Parthenon already had something in place to telescope ghost zones as the stencil shrinks.

src/interface/variable.hpp
forrestglines: I'm not entirely sure how face-centered tensor-valued fields would be used since I'm not entirely sure how cell-centered tensor-valued fields are used in downstream methods.

The only example I can imagine is an two-fluid MHD method I like to implement where magnetic and electric fields are co-located at the faces and evolved together.

src/interface/variable.hpp
forrestglines: I think I agree - we almost always and even preferably interface with variable packs. I wonder if the variable structs entirely could be removed.

src/interface/sparse_pack_base.cpp
forrestglines: I don't like this repeated looping either, nor would these be the only loops repeated. On the other hand, if we only keep the cell-centered loop, we'd need to make sure the loop applies equally well to face-centered and edge-centers variables.  Would one of the indices in the variable be dedicated then to 3-directions so that face-centered and edge-centered variables have one less dimension available? Would we also have to make sure that the coarse variable and flux variable are never allocated? Or only pack the dimensions as needed, like we do for fluxes?

example/loop_advection/loop_advection_package.cpp
Yurlungur: I strongly prefer the meshblock index `b` to be the outermost index, as it should be slowest-moving, even slower than the face location IMO. I think we're in agreement consistency would be best, though. ;)

I bet it wouldn't be too hard to generalize the face name stuff, right @lroberts36 ? But we should probably wait on that until the data structures for faces/nodes/edges have stabilized.

example/loop_advection/loop_advection_package.cpp
Yurlungur: There's nothing automatic I don't think, but it might be a useful feature.

src/interface/variable.hpp
Yurlungur: Any tensor where one index is divergence free---for example a stress tensor---would be appropriate as a face-centered value. That said, I don't know if we have such a thing at the moment and I wouldn't worry about it too much.

src/interface/variable.hpp
Yurlungur: The nice thing about the variable object is it owns the underlying pararrays as well as fluxes, coarse buffers, and metadata, so I don't know if we want to remove the concept entirely, but I think it could use some cleanup.

src/interface/sparse_pack_base.cpp
Yurlungur: I'm not sure I fully understand the question. In principle, the loop is just flattening over dims 6,5,4, which should be generic over grid location. The reason you have special logic for faces and edges is because they live in structs. I think that could be resolved in various ways though.

example/loop_advection/loop_advection_package.cpp
lroberts36: Based on my diagram, I think I would get 
```c++
v(b, edge_z, E_edge(), k, j, i) = 0.25 * (v(b, face_y, E_face(), k, j, i)  
                                        + v(b, face_x, E_face(), k, j, i) 
                                        + v(b, face_y, E_face(), k, j, i - 1) 
                                        + v(b, face_x, E_face(), k, j - 1, i))  
```

example/loop_advection/loop_advection_package.cpp
lroberts36: Another option is to change to `GetBounds*(const IndexDomain &idx_domain, const CellElement& element = CellElement::Center);` with an enum
```c++
enum class CellElement {Center, Face_x, Face_y, Face_z, Edge_xy, Edge_xz, Edge_yz, Vertex};    
```

example/loop_advection/loop_advection_package.cpp
lroberts36: Remembering to when I wrote this, there is no way to be consistent with the previous flux implementation since there is no direct access to fluxes from a `MeshBlockPack`.  For a `MeshBlockPack`s you have to pull out a `VariableFluxPack` and access the fluxes from it. So I don't think there are any consistency issues (contrary to anything I may or may not have said previously). I tend to agree with @Yurlungur that the block index probably is the most logical thing to have out front.

src/interface/variable.hpp
lroberts36: I think it is probably useful to have tensor fields definable at faces, etc. For instance, one might want the three-metric on a face. I also don't think there is anything limiting us from doing this.

src/interface/variable.hpp
lroberts36: Yeah, I think the `Variable` object still needs to exist but the `operator()` stuff does not need to. This would just end up being a second place to maintain when the underlying structure of the data in `Variable` gets changed. I would vote to remove it unless there is somewhere it gets used that is hard to switch to packs.

src/interface/sparse_pack_base.cpp
lroberts36: I agree that we want to avoid code repetition like this if at all possible. 

As I look at things, I am more and more thinking that the relatively simple, fairly flexible way to go is in `CellVariable` to switch to
```c++
ParArrayND<T, VariableState> data; 
ParArrayND<T, VariableState> coarse_s;
ParArrayND<T, VariableState> data_all[8]; 
ParArrayND<T, VariableState> coarse_s_all[8]; 
// After allocation in AllocateData()
data = data_all[0];
coarse_s = coarse_s_all[0]; 
``` 
with `0->cell center`, `1->x face`, etc. and then only allocating the pieces which are necessary based on the metadata in `CellVariable::AllocateData()`. I think by pointing `data` at the cell centered value, it wouldn't break anything that is already in the code. This would leave open the option of saving memory by allocating each member of `data_all` separately so that its size could account for the `+1` that shows up in some dimensions. Alternatively, it would be easy to shift to including the `+1` for all elements, allocate everything necessary as a single large array, and then take subviews (like is done for fluxes currently).  

We would also need to add an extra index to the sparse pack `operator()` specifying the cell element that is being requested (and maybe keep the old version and have it default to cell centered). 

We would then increase the size of the first dimension of `SparsePackBase::pack_` similarly to what we do with fluxes already to account for variables being defined on other geometric elements. As we discussed in the call, it would allow variables to be defined on both cell centers and faces for instance (which seems cleaner to me generally, but I can see some cases where you wouldn't always want that), but would also allow for defining fields that exist on only one kind of element.

Yurlungur: I spoke to @jdolence about this this afternoon and he pointed out that different parts of the data array may be one copy or not. For example cell-centered data for B fields may not need to have a separate copy per stage. To me this still suggests the treatment of face-centered fields being variables as variables are currently written, just with a different shape

lroberts36: I think that fits fine within this model though, since it is flexible enough that you can either have the field on cell centers, faces, or on both. If you want different metadata for the cell centers and the faces, you just create two separate fields and have one be only defined on cell centers and the other only on faces.

The other possible advantage of this model is that to during development, we can force the cell centered values to be always allocated. Then all of the current machinery should at least run and things like boundary communication, prolongation, restriction, for non-cell centered variables can be implemented one at a time.

Yurlungur: But if that's the case, is there a reason to add the extra complexity of all these different arrays? I think having different metadata for different locations will be common.

lroberts36: Of course it is subjective, but I think it introduces limited extra complexity. It is very similar to the amount of complexity required for possibly carrying around fluxes at the end of the day. I can also think of the example of GR quantities for a static space time, which we would want at the cell centers and cell faces, and would be one copy everywhere.

example/loop_advection/loop_advection_package.cpp
forrestglines: I like this idea the most -- I'll implement this

PR #694, opened 2022-07-16
Add concepts lite (small)

lroberts36:
## PR Summary
Adds a relatively clean implementation of concepts to Parthenon as well as a test that the implementation is working. It is probably easiest to look at the comments in the code for the implementation and the test code for how to use it, but the general idea is that one would be able to write a function with different implementations depending on whether or not the input type implements a particular concept. For example:
```c++
template<class T, REQUIRES(implements<my_concept(T)>::value)>
void foo(T& in){ /*implementation when T conforms to my_concept*/ }

template<class T, REQUIRES(!implements<my_concept(T)>::value && implements<my_other_concept(T)>::value)>
void foo(T& in){ /*implementation when T conforms to my_other_concept but not my concept*/ }

template<class T, REQUIRES(!implements<my_concept(T)>::value && !implements<my_other_concept(T)>::value)>
void foo(T& in){ /*default implementation*/ }
```
where the correct implementation is chosen via SFINAE. 

The main impetus for this addition is that in a few places in the code there are templates that take a class that must conform to a particular concept. Most notably, the type passed to `parthenon::ReductionBase` must logically be a contiguous container to be used in MPI calls. A contiguous container essentially has to have methods `size()` and `data()`, and internal type `value_type` (`std::array`, `std::vector`, and `Kokkos::view` all conform to this concept and `ParthenonArray*` should conform to this) or be a type that just stores a single value (so that its size is always one and its `data()` pointer is just a pointer to itself). This was enforced in a way in `reductions.hpp` that broke when `ParArray*` was refactored in #681. Similarly, the `CommunicationBuffer` defined in #663 has to be templated on a type that conforms to the contiguous container concept. In addition to the general framework for concepts, this PR introduces a `contiguous_container` concept and some helper functions for generically getting the size of an object and data pointer of an object, whether it is a contiguous container or a single object (and which also fail to compile if they are passed a non-contiguous container).  

Concepts are of course more general than this, so I am guessing there will be other concepts that we want to define and use in other places in the code. For instance, there may be a way to define a concept that distinguishes between objects that correspond to a single block and objects that correspond to a list of blocks which may allow writing more generic code for dealing with `MeshData` vs `MeshBlockData`. 

This PR does not actually make any changes to the main code. I will make the necessary changes `reductions.hpp` and `communication_buffer.hpp` in #663 and #681 respectively after this PR is merged and before those two PRs are merged. 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/tst/unit/test_concepts_lite.cpp
+++ b/tst/unit/test_concepts_lite.cpp
@@ -38,10 +38,13 @@ bool SatisfiesContainerRequirements(T &&in, VAL_TYPE val, int size_in) {
 
 TEST_CASE("Check that the contiguous container concept works", "") {
   GIVEN("Some containers and some data") {
-    int SIZE = 10;
+    constexpr const int SIZE = 10;
     int val = 2;
 
     int my_int = val;
+    int my_c_array[SIZE];
+    for (int i = 0; i < SIZE; ++i)
+      my_c_array[i] = val;
     std::vector<int> my_vec(SIZE, val);
     // This should also work fine for objects defined on device, but
     // it is more work to write a generic function for checking values
@@ -58,6 +61,7 @@ TEST_CASE("Check that the contiguous container concept works", "") {
     }
 
     REQUIRE(SatisfiesContainerRequirements(my_int, val, 1));
+    REQUIRE(SatisfiesContainerRequirements(my_c_array, val, SIZE));
     REQUIRE(SatisfiesContainerRequirements(my_vec, val, SIZE));
     REQUIRE(SatisfiesContainerRequirements(my_view, val, SIZE));
     REQUIRE(SatisfiesContainerRequirements(my_pararr, val, SIZE));
--- a/tst/unit/test_concepts_lite.cpp
+++ b/tst/unit/test_concepts_lite.cpp
@@ -29,7 +29,7 @@ bool SatisfiesContainerRequirements(T &&in, VAL_TYPE val, size_t size_in) {
   test = test && (size == size_in);
 
   // Check that we can access the data and they all have value val
-  int *pin = contiguous_container::data(in);
+  VAL_TYPE *pin = contiguous_container::data(in);
   for (auto i = 0; i < size; ++i)
     test = test && (val == pin[i]);
 
--- a/tst/unit/test_concepts_lite.cpp
+++ b/tst/unit/test_concepts_lite.cpp
@@ -55,7 +55,11 @@ TEST_CASE("Check that the contiguous container concept works", "") {
     // interface
     parthenon::ParArray1D<int>::host_mirror_type my_pararr("Test", SIZE);
 
+    // We do not expect standard map to conform to contiguous_container
+    std::map<int, int> my_map;
+
     for (int i = 0; i < SIZE; ++i) {
+      my_map[i] = val;
       my_pararr(i) = val;
       my_view(i) = val;
     }
@@ -65,5 +69,9 @@ TEST_CASE("Check that the contiguous container concept works", "") {
     REQUIRE(SatisfiesContainerRequirements(my_vec, val, SIZE));
     REQUIRE(SatisfiesContainerRequirements(my_view, val, SIZE));
     REQUIRE(SatisfiesContainerRequirements(my_pararr, val, SIZE));
+    // Uncommenting the line below should cause compilation to fail
+    // since my_map should not work with the contiguous_container
+    // helper functions. When I have tested, it does cause a compilation failure (LFR).
+    // SatisfiesContainerRequirements(my_map, val, SIZE);
   }
 }
--- a/tst/unit/test_concepts_lite.cpp
+++ b/tst/unit/test_concepts_lite.cpp
@@ -30,7 +30,7 @@ bool SatisfiesContainerRequirements(T &&in, VAL_TYPE val, int size_in) {
 
   // Check that we can access the data and they all have value val
   int *pin = contiguous_container::data(in);
-  for (int i = 0; i < size; ++i)
+  for (auto i = 0; i < size; ++i)
     test = test && (val == pin[i]);
 
   return test;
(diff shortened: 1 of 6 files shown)


Yurlungur (changes requested):
I think this is a useful improvement (though it's painful that we're backporting a C++20 feature). That said, I think it's important that this is documented in the markdown, not just the comments. Could perhaps add some usage/implementation examples to the markdown docs?

lroberts36:
@Yurlungur: Agreed that it would be nice to have markdown documentation, but I would like to get this in quickly and I am a little bandwidth limited for writing the documentation. Would you be amenable to creating an issue to add documentation and doing that in a separate pull request in the future? I think the comments are descriptive enough that there isn't a big danger that I will forget what is going on between now and when I write the doc.

Yurlungur:
> @Yurlungur: Agreed that it would be nice to have markdown documentation, but I would like to get this in quickly and I am a little bandwidth limited for writing the documentation. Would you be amenable to creating an issue to add documentation and doing that in a separate pull request in the future? I think the comments are descriptive enough that there isn't a big danger that I will forget what is going on between now and when I write the doc.

Sure that's fine. How about in the meantime, we add a stub markdown file that points to the implementation/test files?

Yurlungur (approved):
Approved pending github issue + markdown stub

lroberts36:
Stub added and issue #695 created.

tst/unit/test_concepts_lite.cpp
```diff
@@ -0,0 +1,69 @@
...
+#include "kokkos_abstraction.hpp"
+#include "utils/concepts_lite.hpp"
+
+template <class T, class VAL_TYPE>
+bool SatisfiesContainerRequirements(T &&in, VAL_TYPE val, int size_in) {
+  // Check that the value_type of the container is what we would expect
+  using cont_val_type = decltype(contiguous_container::value_type(in));
+  bool test = std::is_same<cont_val_type, VAL_TYPE>::value;
+
+  int size = contiguous_container::size(in);
```
pgrete:
```suggestion
  size_t size = contiguous_container::size(in);
```

tst/unit/test_concepts_lite.cpp
```diff
@@ -0,0 +1,69 @@
...
+  test = test && (size == size_in);
+
+  // Check that we can access the data and they all have value val
+  int *pin = contiguous_container::data(in);
+  for (int i = 0; i < size; ++i)
```
pgrete:
```suggestion
  for (auto i = 0; i < size; ++i)
```

tst/unit/test_concepts_lite.cpp
```diff
@@ -0,0 +1,69 @@
...
+#include <type_traits>
+#include <vector>
+
+#include <catch2/catch.hpp>
+
+#include "kokkos_abstraction.hpp"
+#include "utils/concepts_lite.hpp"
+
+template <class T, class VAL_TYPE>
+bool SatisfiesContainerRequirements(T &&in, VAL_TYPE val, int size_in) {
```
pgrete:
```suggestion
bool SatisfiesContainerRequirements(T &&in, VAL_TYPE val, size_t size_in) {
```

tst/unit/test_concepts_lite.cpp
```diff
@@ -0,0 +1,69 @@
...
+  // Check that the value_type of the container is what we would expect
+  using cont_val_type = decltype(contiguous_container::value_type(in));
+  bool test = std::is_same<cont_val_type, VAL_TYPE>::value;
+
+  int size = contiguous_container::size(in);
+  test = test && (size == size_in);
+
+  // Check that we can access the data and they all have value val
+  int *pin = contiguous_container::data(in);
```
pgrete:
```suggestion
  VAL_TYPE *pin = contiguous_container::data(in);
```

tst/unit/test_concepts_lite.cpp
```diff
@@ -0,0 +1,69 @@
...
+  for (int i = 0; i < size; ++i)
+    test = test && (val == pin[i]);
+
+  return test;
+}
+
+TEST_CASE("Check that the contiguous container concept works", "") {
+  GIVEN("Some containers and some data") {
+    constexpr const int SIZE = 10;
```
pgrete:
```suggestion
    constexpr const size_t SIZE = 10;
```

tst/unit/test_concepts_lite.cpp
```diff
@@ -0,0 +1,69 @@
...
+    for (int i = 0; i < SIZE; ++i) {
+      my_pararr(i) = val;
+      my_view(i) = val;
+    }
+
+    REQUIRE(SatisfiesContainerRequirements(my_int, val, 1));
+    REQUIRE(SatisfiesContainerRequirements(my_c_array, val, SIZE));
+    REQUIRE(SatisfiesContainerRequirements(my_vec, val, SIZE));
+    REQUIRE(SatisfiesContainerRequirements(my_view, val, SIZE));
+    REQUIRE(SatisfiesContainerRequirements(my_pararr, val, SIZE));
```
pgrete, line 71: Can we also have a "negative" example or would that not work because the compiler wouldn't compile the code?

pgrete (approved):
Some nitpicking suggestion (feel free to ignore) and a question on the negative example.
Otherwise LGTM!

tst/unit/test_concepts_lite.cpp
lroberts36: Yeah, negative examples should fail at compile time. I can think of two ways to include this:
- Just include a negative example commented out and people can test by hand 
- Do some template magic to check if a function doesn't exist and then add that as a check for a negative example (it seems like there is a template solution to this, but I would need to spend a little time thinking about it)

pgrete: > Just include a negative example commented out and people can test by hand

Sounds good and like minimal effort.
Feel free to merge whenever (so that you can get this into the other PRs, too)

lroberts36: Done.

PR #699, opened 2022-07-27
Add Independent Sparse Thresholds and Sparse Allocation Control Fields

### LINKED ISSUE #642: Sparse "buffer send" threshold is applied to dense variables
Ping @Yurlungur and @dholladay00 as you reviewed before:

We should also check for `src_allocated` here, don't we?
Otherwise, a dense field that happens to have `0` in all ghost zones, would not be sent or does this not matter because data will always be sent and just not be processes because the the last index in the buffer will contain a 0?
In the latter case, I may be concerned that the receiving buffer is not "clean". IIRC we never reset the receiving buffers because the original (Athena++) designed assumed that all values will always be communicated (with fresh data).

_Originally posted by @pgrete in https://github.com/lanl/parthenon/pull/586#r760305244_

### LINKED ISSUE #643: Need per-variable sparse (de)alloc thresholds
Currently, there's only a single threshold applied to all sparse variables.
For different kinds of sparse variables, different thresholds may be useful.

lroberts36:
## PR Summary

This PR adds a few different sparse variable features and fixes some bugs: 
- Ability to specify the sparse allocation and deallocation thresholds for individual fields. 
- Functionality to allocate/deallocate sparse fields based on the allocation/deallocation status of another sparse field (i.e. controlled and controlling fields).
- Allow for default values of unallocated or newly allocated fields to have values other than zero. 
- NaN payload tags for flagging unitialized, newly allocated sparse fields. 
- Sparse advection test that uses these new features. 
- There are also a couple fixes of small bugs in the sparse machinery.
- Include better checks on whether or not packs containing sparse variables are stale. The old machinery could result in packs being used that were stale.

The first two features are being used downstream in Riot.  

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


lroberts36:
@par-hermes format

lroberts36:
@par-hermes format

lroberts36:
Left to do before removing WIP:
- [x] Flesh out documentation of post-allocation initialization capabilities. 
- [x] Fix `SparsePack` cache IDs to use new `num_allocs_` in `CellVariable` rather than a `bool`. 
- [x] Remove commented out code from deallocation task 
- ~~[ ] Clean up new sparse advection test and add convergence test (maybe split into new PR?)~~
- [x] Remove new sparse advection test (saving for later PR)

src/bvals/cc/bvals_cc_in_one.hpp
```diff
@@ -95,10 +96,12 @@ struct BndInfo {
   RefinementOp_t refinement_op = RefinementOp_t::None;
   Coordinates_t coords, coarse_coords; // coords
 
-  buf_pool_t<Real>::weak_t buf; // comm buffer from pool
-  ParArray6D<Real> var;         // data variable used for comms
-  ParArray6D<Real> fine;        // fine data variable for prolongation/restriction
-  ParArray6D<Real> coarse;      // coarse data variable for prolongation/restriction
+  buf_pool_t<Real>::weak_t buf;                   // comm buffer from pool
+  parthenon::ParArray6D<Real, VariableState> var; // data variable used for comms
+  parthenon::ParArray6D<Real, VariableState>
```
Yurlungur: does this imply a type-casting somewhere? Or is `VariableState` implicit in all the `ParArray*D`s?

src/bvals/cc/sparse_bvals_cc_in_one.cpp
```diff
@@ -458,9 +458,11 @@ TaskStatus SetBounds(std::shared_ptr<MeshData<Real>> &md) {
...
           cache.recv_bnd_info_h(ibuf).buf = buf.buffer();
           if (buf.GetState() == BufferState::received) {
-            cache.recv_bnd_info_h(ibuf).allocated = true;
-            PARTHENON_DEBUG_REQUIRE(v->IsAllocated(),
-                                    "Variable must be allocated to receive");
+            // With control variables, we can end up in a state where a
+            // variable that is not receiving null data is unallocated.
+            // for allocated to be set, the buffer must be sending non-null
+            // data and the receiving variable must be allocated
+            cache.recv_bnd_info_h(ibuf).allocated = v->IsAllocated();
```
Yurlungur: I don't fully understand this comment.

src/interface/meshblock_data.hpp
```diff
@@ -426,6 +426,19 @@ class MeshBlockData {
     return true;
   }
 
+  void SetAllVariablesToInitialized() {
+    std::for_each(varVector_.begin(), varVector_.end(),
+                  [](auto &sp_var) { sp_var->data.initialized = true; });
+  }
```
Yurlungur, line 392: so this doesn't actually initialize the data, just set the metadata. Is that safe?

src/interface/variable.hpp
```diff
@@ -306,6 +314,39 @@ class ParticleVariable {
...
+    base_regexs.push_back(std::regex(base_name + ".*"));
+
+  for (auto &pvar : cv_in) {
+    if (std::any_of(base_regexs.begin(), base_regexs.end(), [&](const auto &in) {
+          return std::regex_match(pvar->label(), in);
+        })) {
+      out.push_back(pvar);
+    }
+  }
+  return out;
```
Yurlungur, line 235: These are nice additions.

src/interface/variable_state.hpp
```diff
@@ -0,0 +1,60 @@
...
+#include "basic_types.hpp"
+#include "globals.hpp"
+#include "interface/metadata.hpp"
+#include "parthenon_arrays.hpp"
+
+namespace parthenon {
+
+static constexpr int InvalidSparseID = std::numeric_limits<int>::min();
+
+struct VariableState : public empty_state_t {
```
Yurlungur, line 27: will variable state always look the same? Should we name this something like `SparseState` in case we eventually have other kinds of state? I'm not sure this matters much and I'm fine leaving it as is.

src/outputs/parthenon_hdf5.cpp
```diff
@@ -890,8 +888,8 @@ void PHDF5Output::WriteOutputFileImpl(Mesh *pm, ParameterInput *pin, SimTime *tm
       if (!is_allocated) {
         if (vinfo.is_sparse) {
           hsize_t N = varSize * vlen;
-          memset(tmpData.data() + index, 0, N * sizeof(OutT));
-          index += N;
+          for (int i = 0; i < N; ++i)
+            tmpData[index++] = std::numeric_limits<Real>::quiet_NaN();
```
Yurlungur: is there a reason this couldn't be `memset` to `quiet_NaN`? It probably doesn't matter.

src/utils/nan_payload_tag.hpp
```diff
@@ -0,0 +1,123 @@
...
+  assert(std::isnan(flag_nan));
+  assert(!BitwiseCompare(std::numeric_limits<T>::signaling_NaN(), flag_nan));
+  assert(!BitwiseCompare(std::numeric_limits<T>::quiet_NaN(), flag_nan));
+
+  return flag_nan;
+}
+
+template <class T, REQUIRES(!std::numeric_limits<T>::is_iec559)>
+T GetNaNWithPayloadTag(uint8_t tag = 1) {
+  // TODO(LFR): Probably need to warn here that we can't tag without IEEE
```
Yurlungur: a debug warn is probably enough

Yurlungur (approved):
This all seems reasonable to me. A few minor comments below.

src/interface/meshblock_data.hpp
lroberts36: The idea of the `initialized` flag is that a user can check it in a kernel to see if a variable has been freshly allocated with all zeros. In some use cases (none that are actually implemented right now), it would make sense to initialize with something other than zero. Users can register `package->InitNewlyAllocatedVars`, which will be called in the task `InitNewlyAllocatedVariables` before calls to `SetAllVariablesToInitialized` are made.

Yurlungur: Ok that makes sense. I guess my concern is that `SetAllVariablesToInitialized` could get called out of sync with the variable actually getting initialized, which could cause some subtle bug. I'm not sure what the solution is though.

src/bvals/cc/sparse_bvals_cc_in_one.cpp
lroberts36: If a variable does not control its own allocation status, it can be being passed supposedly above threshold data but be unallocated because its controlling variable is not allocated. This needs to be checked for so we don't try to write data to an unallocated array.

Yurlungur: Ah I see. :+1:

src/bvals/cc/bvals_cc_in_one.hpp
lroberts36: I think that there shouldn't be type-casting going on because the `CellVariable`s now contain `ParArray6D<Real, VariableState>` as well. Nevertheless, type-casting from `ParArray*<Real>` to `ParArray*<Real, VariableState>` should work.

src/interface/variable.hpp
lroberts36: There was a collision that occurred in Riot using the old method.

src/interface/variable_state.hpp
lroberts36: I think that we would probably add other non-sparse state to this object. This is the state that is carried around by all the data held by `CellVariable`s, whether or not they are sparse.

src/outputs/parthenon_hdf5.cpp
lroberts36: yeah, `memset` might be better but I don't think it matters. This should probably be switched back to zero, since previously the non-allocated sparse fields were zero.

src/interface/meshblock_data.hpp
lroberts36: Yeah, I don't think there is a completely safe way to do this because `initialized` can't be changed on device.

src/utils/nan_payload_tag.hpp
lroberts36: added.

lroberts36:
@pgrete Can you give this a review when you get the chance? Although there might be some more change to the sparse framework as we work on things in Riot, I think at this point it would be good to get this into `develop`. It had `develop` had diverged from it significantly over the last month, so I think my original plan of waiting until things were completely stable in Riot may not be a good one (in particular I am thinking of changes coming as we include face, edge, and vertex variables).

docs/interface/sparse.md
```diff
@@ -1,38 +1,139 @@
...
+  below some threshold). This is an operation that needs to be defined in a task by the downstream 
+  code, since it is problem specific. **Currently, Parthenon does not have this capability. It will 
+  probably be necessary to define some hooks for checking blocks for the necessity of allocation and 
+  having some internal Parthenon functions for actually performing the allocation.**
+- *Allocation due to neighbors:* If a sparse field on a neighbor block passes ghost data that is 
+  anywhere above the `allocation_threshold`, the receiving block will allocate the sparse field if 
+  it is unallocated before loading the ghost data. `allocation_threshold` is set in the `Metadata` 
+  of a field.  A detailed description of sparse ghost zone communication is 
+  given [here](../sparse_boundary_communication.md). 
+- *Allocation due to other fields:* In some instances, we may want to allocate or deallocate a
```
pgrete, line 55: It may be good to document the logical relationship between those two mechanism, i.e., are they `exclusive or`, `or`, `and` and, if both are defined if there's an error or which one takes precedence.

docs/interface/sparse.md
```diff
@@ -1,38 +1,139 @@
...
+  allocated). Looping over fields in these type of packs generally requires hierarchichal parallelism. 
+  Currently, `VariablePack` and `MeshBlockPack` currently employ a "sparse sparse packing" 
+  strategy, where all fields are included in the index space of the pack but the allocation status of 
+  `(block, field)` must be checked before accessing `(block, field, k, j, i)` since this is not 
+  guaranteed to point to valid memory. *There is "dense sparse pack" implementation of `VariablePack` 
+  and `MeshBlockPack` in the branch `lroberts36/merge-sparse-with-jdolence-sparse` that is being used 
+  in `Riot`. This should probably be brought into `develop`, since the "sparse sparse pack" access 
+  pattern is probably not desirable.  
+
+In comparison, a dense field only requires the operation *Access*.
```
pgrete: I'm not quite sure about the context/meaning of this sentence here.

docs/interface/sparse.md
```diff
@@ -132,162 +235,37 @@ that block, see [Boundary exchange](#boundary-exchange) for details. The infrast
...
+There is a new task called `SparseDealloc` in `src/interface/update.cpp` taking a `MeshData` pointer. 
+It is meant to be run after the update task for the last stage (of course, it does not have to be run 
+every time step). On every block, it checks the values of all sparse variables. If the maximum 
+absolute value is below the user-defined deallocation threshold, the variable is flagged for 
+deallocation on that block. The variable is only actually deallocated if it has been flagged for 
+deallocation a certain number of times in a row (if any of the values exceeds the deallocation 
+threshold, the counter is reset to 0). That number is the deallocation count, which is also settable 
+by the user in the input file. Note that currently the deallocation threshold is the same for all 
+variables, a future improvement is to add the capability of setting a different threshold per sparse 
+pool.
```
pgrete: Is this true? The PR description also mentions individual dealloc thresholds.

docs/interface/sparse.md
```diff
@@ -132,162 +235,37 @@ that block, see [Boundary exchange](#boundary-exchange) for details. The infrast
...
-the update task for the last stage (of course, it does not have to be run every time step). On every
-block, it checks the values of all sparse variables. If the maximum absolute value is below the
-user-defined deallocation threshold, the variable is flagged for deallocation on that block. The
-variable is only actually deallocated if it has been flagged for deallocation a certain number of
-times in a row (if any of the values exceeds the deallocation threshold, the counter is reset to 0).
-That number is the deallocation count, which is also settable by the user in the input file. Note
-that currently the deallocation threshold is the same for all variables, a future improvement is to
-add the capability of setting a different threshold per sparse pool.
+There is a new task called `SparseDealloc` in `src/interface/update.cpp` taking a `MeshData` pointer. 
+It is meant to be run after the update task for the last stage (of course, it does not have to be run
```
pgrete, line 242: So this mean, a downstream codes needs to add this Task in the driver to fully leverage sparse, doesn't it?

example/sparse_advection/sparse_advection_driver.cpp
```diff
@@ -130,13 +130,16 @@ TaskCollection SparseAdvectionDriver::MakeTaskCollection(BlockList_t &blocks,
         tl.AddTask(update, parthenon::cell_centered_bvars::SendBoundaryBuffers, mc1);
     auto recv = tl.AddTask(update | start_bound,
                            parthenon::cell_centered_bvars::ReceiveBoundaryBuffers, mc1);
-
     auto set = tl.AddTask(recv, parthenon::cell_centered_bvars::SetBoundaries, mc1);
 
+    auto init_allocated =
+        tl.AddTask(set, InitNewlyAllocatedVars<MeshData<Real>>, mc1.get());
```
pgrete, line 136: At some point we should decide if we pass pointers or shared pointers around.

example/sparse_advection/sparse_advection_driver.cpp
```diff
@@ -130,13 +130,16 @@ TaskCollection SparseAdvectionDriver::MakeTaskCollection(BlockList_t &blocks,
...
+
     auto restrict = set;
```
pgrete:
```suggestion
    auto restrict = init_allocated;
```

Potentially a bug

src/utils/nan_payload_tag.hpp
```diff
@@ -0,0 +1,123 @@
...
+// for the U.S. Department of Energy/National Nuclear Security Administration. All rights
+// in the program are reserved by Triad National Security, LLC, and the U.S. Department
+// of Energy/National Nuclear Security Administration. The Government is granted for
+// itself and others acting on its behalf a nonexclusive, paid-up, irrevocable worldwide
+// license in this material to reproduce, prepare derivative works, distribute copies to
+// the public, perform publicly and display publicly, and to permit others to do so.
+//========================================================================================
+
+#ifndef UTILS_NAN_PAYLOAD_TAG_HPP_
+#define UTILS_NAN_PAYLOAD_TAG_HPP_
```
pgrete, line 15: Is there any user doc (or even mention) of this feature?

src/utils/nan_payload_tag.hpp
```diff
@@ -0,0 +1,123 @@
...
+  auto &a_bits = reinterpret_cast<const impl::contiguous_bitset<sizeof(a)> &>(a);
+  auto &b_bits = reinterpret_cast<const impl::contiguous_bitset<sizeof(b)> &>(b);
+  bool val = true;
+  for (int i = 0; i < sizeof(U); ++i) {
+    val = val & (b_bits.bytes[i] == a_bits.bytes[i]);
+  }
+  return val;
+  // memcmp returns zero if the memory is the same
+  // which is a little confusing, it is also not available on device
+  // return !std::memcmp(&a, &b, sizeof(a));
```
pgrete: leftover/dead code?

tst/unit/test_nan_tags.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+using real_t = double;
+using policy2d = Kokkos::MDRangePolicy<Kokkos::Rank<2>>;
+using arr2d_t = parthenon::ParArray2D<real_t>;
+
+TEST_CASE("NaN payload tagging", "[NaN payload]") {
+  real_t qnan = std::numeric_limits<real_t>::quiet_NaN();
+  real_t flag1 = parthenon::GetNaNWithPayloadTag<real_t>(1);
+  real_t flag2 = parthenon::GetNaNWithPayloadTag<real_t>(2);
+  GIVEN("An two NaNs with tags") {
```
pgrete:
```suggestion
  GIVEN("Two NaNs with tags") {
```

tst/unit/test_nan_tags.cpp
```diff
@@ -0,0 +1,85 @@
...
+        policy2d({0, 0}, {N, N}),
+        KOKKOS_LAMBDA(const int j, const int i) { arr2d(j, i) = flag1; });
+
+    int num_flag1 = 0;
+    int num_flag2 = 0;
+    Kokkos::parallel_reduce(
+        policy2d({0, 0}, {N, N}),
+        KOKKOS_LAMBDA(const int j, const int i, int &l_num_flag1, int &l_num_flag2) {
+          if (parthenon::BitwiseCompare(flag1, arr2d(i, j))) ++l_num_flag1;
+          if (parthenon::BitwiseCompare(flag2, arr2d(i, j))) ++l_num_flag2;
```
pgrete:
```suggestion
          if (parthenon::BitwiseCompare(flag1, arr2d(j, i))) ++l_num_flag1;
          if (parthenon::BitwiseCompare(flag2, arr2d(j, i))) ++l_num_flag2;
```
Not an issue as `N==N`, but for consistency.

pgrete (commented):
First bunch of comments, more tomorrow.

tst/regression/test_suites/advection_outflow/advection_outflow.py
```diff
@@ -46,7 +46,8 @@ def Analyse(self, parameters):
                 "outflow.out0.final.phdf",
                 parameters.parthenon_path
                 + "/tst/regression/gold_standard/outflow.out0.final.phdf",
-            ]
+            ],
+            check_metadata=False,
```
pgrete, line 50: Leftover from debugging or why this change?

tst/regression/test_suites/poisson/poisson.py
```diff
@@ -45,7 +45,8 @@ def Analyse(self, parameters):
                 "poisson.out0.00001.phdf",
                 parameters.parthenon_path
                 + "/tst/regression/gold_standard/poisson.out0.00001.phdf",
-            ]
+            ],
+            check_metadata=False,
```
pgrete, line 49: Leftover from debugging or why this change?

tst/regression/test_suites/output_hdf5/output_hdf5.py
```diff
@@ -126,7 +126,8 @@ def Analyse(self, parameters):
             ["max", 9.43685e-01, 4.80914e-01],
             [
                 "min",
-                1.67171e-10,
+                # 1.67180e-10 if parameters.sparse_disabled else 1.67171e-10,
+                1.67180e-10,
```
pgrete, line 130: Does the commented line not work as expected or are the results now in agreement (in which case the commented line could go).

src/bvals/cc/bnd_info.cpp
```diff
@@ -338,6 +338,18 @@ BndInfo BndInfo::GetSetBndInfo(std::shared_ptr<MeshBlock> pmb, const NeighborBlo
...
     out.var = v->data.Get();
   }
+
+  if (buf->GetState() == BufferState::received) {
+    // With control variables, we can end up in a state where a
+    // variable that is not receiving null data is unallocated.
+    // for allocated to be set, the buffer must be sending non-null
+    // data and the receiving variable must be allocated
+    out.allocated = v->IsAllocated();
+  } else if (buf->GetState() == BufferState::received_null) {
```
pgrete: Is there any way the buffer stay may change asynchronously between the two calls to `GetState()`?
If so it may be safer to pull the function call outside of the `if` to a temp variable.

src/bvals/cc/boundary_communication.cpp
```diff
@@ -195,9 +195,9 @@ TaskStatus ReceiveBoundBufs(std::shared_ptr<MeshData<Real>> &md) {
           // (the state could also be BufferState::received_null, which corresponds to no
           // data)
           if (buf.GetState() == BufferState::received && !v->IsAllocated()) {
-            pmb->AllocateSparse(v->label());
-            // TODO(lfroberts): Need to flag this so that the array gets filled with
-            //                  something sensible, currently just defaulted to zero.
+            constexpr bool flag_uninitialized = true;
+            constexpr bool only_control = true;
+            pmb->AllocateSparse(v->label(), only_control, flag_uninitialized);
```
pgrete, line 200: Is this for readability or do you expect additional changes (in which case another `todo` may be useful).

src/driver/multistage.cpp
```diff
@@ -90,6 +90,10 @@ TaskListStatus MultiStageDriver::Step() {
   TaskListStatus status;
   integrator->dt = tm.dt;
   for (int stage = 1; stage <= integrator->nstages; stage++) {
+    // Clear any initialization info. We should be relying
+    // on only the immediately preceding stage to contain
+    // reasonable data
+    pmesh->SetAllVariablesToInitialized();
```
pgrete, line 96: I saw that @Yurlungur also had a comment on this function/logic and even after reading it I'm not sure I fully understand the logic.
Why is this "clear" required here? Can we get to a point where the "clear" actually clears sth or may that be just a hint that sth went wrong (in which case this could also be a safety check rather than a hard reset).

src/interface/sparse_pack_base.cpp
```diff
@@ -114,8 +114,8 @@ SparsePackBase SparsePackBase::Build(T *pmd, const PackDescriptor &desc) {
   pack.pack_ = pack_t("data_ptr", leading_dim, nblocks, max_size);
   auto pack_h = Kokkos::create_mirror_view(pack.pack_);
 
-  pack.bounds_ = bounds_t("bounds", 2, nblocks, nvar);
-  auto bounds_h = Kokkos::create_mirror_view(pack.bounds_);
+  pack.bounds_ = bounds_t("bounds", 2, nblocks, nvar + 1);
```
pgrete, line 118: "`+1`" because we use inclusive upper loop bounds or because it "Record the maximum for easy access".
May be good to add a comment (or point to the one I missed).

src/interface/state_descriptor.cpp
```diff
@@ -172,7 +174,14 @@ class FieldProvider : public VariableProvider {
...
     } else {
-      added = state_->AddField(base_name, metadata);
+      for (auto &pair : packages_.AllPackages()) {
+        auto &package = pair.second;
+        if (package->FieldPresent(base_name)) {
+          auto controller = package->GetFieldController(base_name);
+          added = state_->AddField(base_name, metadata, controller);
+          break;
+        }
+      }
```
pgrete, line 184: This code is in the `else` branch of the `if (metadata.IsSet(Metadata::Sparse))` conditional.
Does this mean "dense" fields also get a "controller" and, if so, what does this actually mean?

src/interface/update.cpp
```diff
@@ -154,79 +154,73 @@ TaskStatus SparseDealloc(MeshData<Real> *md) {
...
 
+  const auto tup = SparsePack<>::Get(md, control_vars, {Metadata::Sparse});
+  auto pack2 = std::get<0>(tup);
+  auto pack2Idx = std::get<1>(tup);
+  for (int b = 0; b < num_blocks; ++b) {
+    for (auto &pair : pack2Idx) {
+      int lo = pack2.GetLowerBoundHost(b, PackIdx(pack2Idx[pair.first]));
+      int hi = pack2.GetUpperBoundHost(b, PackIdx(pack2Idx[pair.first]));
+    }
+  }
```
pgrete: What does this code do?
`lo` and `hi` don't exists outside of the loop and a `Get` call should not modify.

src/interface/update.hpp
```diff
@@ -150,6 +151,73 @@ TaskStatus FillDerived(T *rc) {
   return TaskStatus::complete;
 }
 
+template <typename T>
+TaskStatus InitNewlyAllocatedVars(T *rc) {
+  if (!rc->AllVariablesInitialized()) {
+    std::cout << "Initializing variable interiors after allocation" << std::endl;
```
pgrete: Leftover debug statement? This printed by all ranks, isn't it?

src/interface/update.hpp
```diff
@@ -150,6 +151,73 @@ TaskStatus FillDerived(T *rc) {
...
+        Kokkos::TeamPolicy<>(parthenon::DevExecSpace(), v.GetNBlocks(), Kokkos::AUTO),
+        KOKKOS_LAMBDA(parthenon::team_mbr_t team_member) {
+          const int b = team_member.league_rank();
+          int lo = v.GetLowerBound(b, variable_names::any());
+          int hi = v.GetUpperBound(b, variable_names::any());
+
+          for (int vidx = lo; vidx <= hi; ++vidx) {
+            if (!v(b, vidx).initialized) {
+              Real val = v(b, vidx).sparse_default_val;
+              // printf("Initializing block % i variable %i to %e \n", b, vidx, val);
```
pgrete: debug statement

src/interface/update.hpp
```diff
@@ -150,6 +151,73 @@ TaskStatus FillDerived(T *rc) {
...
 
+template <typename T>
+TaskStatus InitNewlyAllocatedVars(T *rc) {
+  if (!rc->AllVariablesInitialized()) {
+    std::cout << "Initializing variable interiors after allocation" << std::endl;
+
+    const IndexDomain interior = IndexDomain::interior;
+    const IndexRange ib = rc->GetBoundsI(interior);
+    const IndexRange jb = rc->GetBoundsJ(interior);
+    const IndexRange kb = rc->GetBoundsK(interior);
```
pgrete, line 160: Why is only the interior initialized?

src/interface/variable.hpp
```diff
@@ -118,18 +118,24 @@ class CellVariable {
...
+  int GetAllocationStatus() {
+    if (!is_allocated_) return 0;
+    return num_alloc_;
+  }
+
  private:
   // allocate data, fluxes, and boundary variable
-  void Allocate(std::weak_ptr<MeshBlock> wpmb);
+  void Allocate(std::weak_ptr<MeshBlock> wpmb, bool flag_uninitialized = false);
+  int num_alloc_ = 0;
```
pgrete, line 135: Can this var been anything other than `0` or `1`?

pgrete (commented):
Alright. I'm done with the review. I left a couple of comments (nothing too big) that I'd like to see/hear an answer first before doing the final (swift) approve.

docs/interface/sparse.md
lroberts36: I think that it should be `or` between `allocation due to neighbors` and `allocation due to other fields`. We should always default to allocating a field if anything requires allocation, since that won't degrade the accuracy of the calculation.

Right now, there is no mechanism for `allocation due to other fields` in the code.

docs/interface/sparse.md
lroberts36: It is in relation to the sentence leading into the bullet list of spare operations. I have changed it slightly to highlight that the comparison is to sparse fields.

docs/interface/sparse.md
lroberts36: Good catch, I missed updating that part of the documentation. Fixed.

docs/interface/sparse.md
lroberts36: yes, but this is not new to this PR.

example/sparse_advection/sparse_advection_driver.cpp
lroberts36: Yes, I always default to raw pointers if the function does not take ownership. I only left the `SendBoundaryBuffers`, etc. as smart pointers for backward compatibility.

example/sparse_advection/sparse_advection_driver.cpp
lroberts36: fixed

src/utils/nan_payload_tag.hpp
lroberts36: no, there isn't. I thought it might be useful in some cases where sparse variables are allowed to take any floating point value (so you need some sentinel other than e.g. zero to flag that the data hasn't been initialized) and figured out how to do it, but it isn't used anywhere right now. We could take it out of the PR and introduce it later if someone starts to use it.

src/utils/nan_payload_tag.hpp
lroberts36: I changed the comment just to explain why we don't use `std::memcmp` since it naively seems like the obvious solution.

tst/regression/test_suites/advection_outflow/advection_outflow.py
lroberts36: I was getting errors about the output versions not matching, probably better to update the gold files I guess.

tst/regression/test_suites/poisson/poisson.py
lroberts36: see above

tst/regression/test_suites/output_hdf5/output_hdf5.py
lroberts36: The results are now in agreement and I *think* the commented line can go, but I don't fully understand why things changed so I am hesitant to remove the old value.

src/bvals/cc/bnd_info.cpp
lroberts36: I don't think that the buffer can change state in between those two calls since everything is serial on a particular rank. I switched though to possibly future proof (i.e. if there are multiple threads), although I think there may be other race conditions if the buffer states are being updated by multiple threads.

src/bvals/cc/boundary_communication.cpp
lroberts36: This is only for readability. I think having `f(true, false)` is always confusing.

src/driver/multistage.cpp
lroberts36: All that `SetAllVariablesToInitialized` does is to set `VariableState::initialized = true` for all variables. It doesn't change anything about the values in the field. 

The idea is that newly allocated fields on a block are set to zero but in some cases, users may want to set a different default value on allocation. This could be just a number other than zero or something more complicated that depends on the local state. If they call the task `InitNewlyAllocatedVars`, the variables with `VariableState::Initialized = false` get set to `VariableState::sparse_default_val` initially, and then the user methods for initialization the user has registered in various packages can search for where the fields are set to `sparse_default_val` and make whatever changes. (One possibility is that `VariableState::sparse_default_val` would be set to a nan_tag.)

But if the user doesn't add the task `InitNewlyAllocatedVars` and just assumes fields are initialized to zero, the initialized flags still need to be set to `true` at some point (although they are probably never being used).

src/interface/sparse_pack_base.cpp
lroberts36: `+1` because we record the maximum for easy access. I added a comment to this effect.

src/interface/state_descriptor.cpp
lroberts36: The dense fields get themselves as controllers. It shouldn't really get used, but something needed to be in the controller field to keep all variables on the same footing.

src/interface/update.cpp
lroberts36: It is there because I was testing the host functions. I have removed it, as well as some other code that was left over from using the old style packs.

src/interface/update.hpp
lroberts36: Yeah, debug statement. Removed.

src/interface/update.hpp
lroberts36: removed.

src/interface/update.hpp
lroberts36: You potentially have received data from other blocks in your boundaries that is not unitialized.

src/interface/variable.hpp
lroberts36: It should increase by one every time the field is allocated, so after allocation->deallocation->allocation it should be larger by two. This gives a way to check if the data that is pointed to in a pack is stale or not (since just checking the allocation status of fields can resulting in thinking your pack is not stale when it is).

lroberts36:
Thanks for the review @pgrete! I think I have responded to all your comments.

docs/interface/sparse.md
pgrete: > Right now, there is no mechanism for `allocation due to other fields` in the code.

Now I'm confused. I thought this PR is about adding this functionality (as described also here in the doc). What am I missing?

src/interface/update.hpp
pgrete: Is there a case where this may lead to problems (e.g., because reconstructed, but uninitialized boundary data is being used to update some active zones)?

src/utils/nan_payload_tag.hpp
pgrete: Let's keep it in for now (but I volunteered you to add the doc later, see https://github.com/parthenon-hpc-lab/parthenon/issues/792)

src/driver/multistage.cpp
pgrete: I see. To be honest, I did not get this (initially) from reading the docs. If you feel like it may be worth adding some info around that I won't object.

tst/regression/test_suites/output_hdf5/output_hdf5.py
pgrete: If the sparse results are now in agreement with the original "dense" results, I'd say this is an improvement and it may have fixed a hidden bug.
I only would have been concerned if the dense results would now yield the original sparse value.

tst/regression/test_suites/advection_outflow/advection_outflow.py
pgrete: Is this still the case after current develop has been merged it? Otherwise that should also fail.
I created an new issue to double check.

pgrete (approved):
Thanks for addressing my comments.
I'm approving now (but would still be happy for final replies to the unresolved conversations).

Also one final note on sparse in general. The complexity/logic of the implementation is now beyond what I was able to follow based on the docs.
I know there are more pressing concerns in terms of getting things work, but at some (hopefully not too distant future), I'd be very happy to see a more detailed (developer) doc for the internals of the parse machinery.
I'm probably also biased as I again spent hours this week trying to track down #659 with little progress,

docs/interface/sparse.md
lroberts36: Sorry, there are two ideas here. The first is that you would like to always allocate one field whenever some other field is allocated. This is what this PR introduced. The second, which is not implemented, is allocating a field based upon some more complicated criterion in the interior of a block (e.g. the temperature being above/below some threshold).

src/interface/update.hpp
lroberts36: The boundaries that are receiving data are filled and the boundaries that are not receiving data are filled with the default value in `ReceiveBoundBufs`, so everything should be initialized.

src/driver/multistage.cpp
lroberts36: At some point I will go through and update the developer sparse documentation to more clearly describe this.

Issue #644, opened 2022-01-26
Cleanup HDF5 ifdef logic for restart interfaces


pgrete:
Currently, the restart interfaces are fully exposed even when HDF5 is not enabled.
Given that we only support restarts with HDF5 files, the code in all interfaces is guarded by `#ifdefs`.
However, this also "leaks" outside the interfaces itself as some structs defined within the restart sources are also used in other parts requiring more `#ifdefs` to handle the unknown HDF5 data types.

I think a cleanup may be in order to reduce the number of `#ifdefs` and ensure that the logic is self-contained to the restart files.

PR #710, opened 2022-08-05
Remove data transpose for hdf5 output

### LINKED ISSUE #709: C++17 code in current release
Clang tells me:
```
[build] /home/pgrete/src/parthenon/src/mesh/refinement_cc_in_one.hpp:100:35: warning: template template parameter using 'typename' is a C++17 extension [-Wc++17-extensions]
[build] template <int DIM, template <int> typename Stencil>
[build]                                   ^~~~~~~~
[build]                                   class
[build] /home/pgrete/src/parthenon/src/mesh/refinement_cc_in_one.hpp:133:35: warning: template template parameter using 'typename' is a C++17 extension [-Wc++17-extensions]
[build] template <int DIM, template <int> typename Stencil>
[build]                                   ^~~~~~~~
[build]                                   class
[build] /home/pgrete/src/parthenon/src/mesh/refinement_cc_in_one.hpp:168:35: warning: template template parameter using 'typename' is a C++17 extension [-Wc++17-extensions]
[build] template <int DIM, template <int> typename Stencil>
[build]                                   ^~~~~~~~
[build]                                   class
[build] /home/pgrete/src/parthenon/src/mesh/refinement_cc_in_one.hpp:182:26: warning: template template parameter using 'typename' is a C++17 extension [-Wc++17-extensions]
[build] template <template <int> typename Stencil, typename... Args>
```

Has anyone tried with an older compiler if this actually results in an error when the compiler cannot handle this?

pgrete:
## PR Summary

Remove the data transpose for HDF outputs.
To this end, I introduced a new internal variable specifying the output format in anticipation that other updates may happen in the future.
Moreover, this will allow analysis tools (like yt) to process both old and new data.

I also tried to maintain backwards compatibility both for
- the outputs, i.e., we can still read original restart files, but will always write in the new format. A warning is shown for the user that this happens.
- our python readers so that output from version "1" and 2 can be compared.

Finally, xdmf files are now always written. I don't see why this should not be the case as it made analyzing restart outputs with visit or Paraview unnecessarily complicated.

NB: I also fixed quite a few compiler warnings that clang was throwing at me for using C++17 features while defining C++14 standard.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


pgrete:
PR is now ready for review.

I encourage a thorough internal downstream testing of this change as your code have way more "variables" than our.

scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
```diff
@@ -404,7 +408,11 @@ def Get(self, variable, flatten=True):
             if self.varData[variable] is None:
                 self.varData[variable] = self.fid[variable][:]
                 vShape = self.varData[variable].shape
-                vLen = vShape[-1]
+                if self.OutputFormatVersion == -1:
```
Yurlungur, line 412: I would add a TODO to remove the legacy mode eventually.

src/mesh/refinement_cc_in_one.hpp
```diff
@@ -97,7 +97,7 @@ GetLoopBoundsFromBndInfo(const Info_t &info, const int ckbs, const int cjbs, int
 // a device version, which requires the buffer cache device only,
 // and a version that automatically swaps between them depending on
 // the size of the buffer cache.
-template <int DIM, template <int> typename Stencil>
+template <int DIM, template <int> class Stencil>
```
Yurlungur, line 100: Why this change?

src/outputs/io_wrapper.hpp
```diff
@@ -50,7 +50,7 @@ class IOWrapper {
   enum class FileMode { read, write };
 
   // wrapper functions for basic I/O tasks
-  int Open(const char *fname, FileMode rw);
+  void Open(const char *fname, FileMode rw);
```
Yurlungur, line 53: No error code?

src/outputs/parthenon_hdf5.cpp
```diff
@@ -1,6 +1,6 @@
 //========================================================================================
-// Athena++ astrophysical MHD code
-// Copyright(C) 2014 James M. Stone <jmstone@princeton.edu> and other code contributors
+// Parthenon performance portable AMR framework
+// Copyright(C) 2020-2022 The Parthenon collaboration
 // Licensed under the 3-clause BSD License, see LICENSE file for details
```
Yurlungur, line 4: By the terms of the original Athena++ BSD license, do we not need to include the original CC as well as the newer ones?

src/outputs/parthenon_hdf5.cpp
```diff
@@ -185,24 +185,27 @@ static void writeXdmfSlabVariableRef(std::ofstream &fid, const std::string &name
...
+    // START: iblock variable(_component)  0   0   0
+    // STRIDE: 1               1           1   1   1
+    // COUNT:  1           vector_size    nx3 nx2 nx1
     fid << prefix << "    "
-        << R"(<DataItem Dimensions="3 5" NumberType="Int" Format="XML">)" << iblock
-        << " 0 0 0 " << i << " 1 1 1 1 1 1 " << dims321 << " " << vector_size
-        << "</DataItem>" << std::endl;
+        << R"(<DataItem Dimensions="3 5" NumberType="Int" Format="XML">)" << iblock << " "
+        << i << " 0 0 0 "
+        << " 1 1 1 1 1 1 " << vector_size << " " << dims321 << "</DataItem>" << std::endl;
```
Yurlungur, line 202: I do not remember my XDMF well enough to know if this works, but I assume so. Did you try reading a file in visit or paraview?

src/outputs/parthenon_hdf5.cpp
```diff
@@ -841,49 +844,48 @@ void PHDF5Output::WriteOutputFileImpl(Mesh *pm, ParameterInput *pin, SimTime *tm
...
-                for (int l = 0; l < nx6; ++l) {
-                  for (int m = 0; m < nx5; ++m) {
-                    for (int n = 0; n < nx4; ++n) {
-                      tmpData[index++] = static_cast<OutT>(v_h(l, m, n, k, j, i));
+          for (int t = 0; t < nx6; ++t) {
+            for (int u = 0; u < nx5; ++u) {
+              for (int v = 0; v < nx4; ++v) {
+                for (int k = out_kb.s; k <= out_kb.e; ++k) {
+                  for (int j = out_jb.s; j <= out_jb.e; ++j) {
+                    for (int i = out_ib.s; i <= out_ib.e; ++i) {
```
Yurlungur, line 869: Can we make this `nx3`, `nx2`, `nx1` to support variables that are not the same shape as a cell-centered var on the meshblock? I.e., that way we can automatically support faces, nodes, etc? I know that's not the point of this PR, but I think it's a small enough change we should just make it "while we're here."

src/outputs/restart.hpp
```diff
@@ -146,8 +155,23 @@ class RestartReader {
 
     /** Select hyperslab in dataset **/
     hsize_t offset[5] = {static_cast<hsize_t>(range.s), 0, 0, 0, 0};
-    hsize_t count[5] = {static_cast<hsize_t>(range.e - range.s + 1), bsize[2], bsize[1],
-                        bsize[0], vlen};
+    hsize_t count[5];
+    if (file_output_format_version == -1) {
+      count[0] = static_cast<hsize_t>(range.e - range.s + 1);
```
Yurlungur, line 160: See above regarding decoupling variable size from meshblock size

src/parthenon_manager.cpp
```diff
@@ -304,17 +326,34 @@ void ParthenonManager::RestartPackages(Mesh &rm, RestartReader &resfile) {
...
+            }
+          }
+        }
+      } else if (file_output_format_ver == HDF5::OUTPUT_VERSION_FORMAT) {
+        for (int t = 0; t < Nt; ++t) {
+          for (int u = 0; u < Nu; ++u) {
+            for (int v = 0; v < Nv; ++v) {
+              for (int k = out_kb.s; k <= out_kb.e; ++k) {
+                for (int j = out_jb.s; j <= out_jb.e; ++j) {
+                  for (int i = out_ib.s; i <= out_ib.e; ++i) {
```
Yurlungur, line 351: See above regarding decoupling meshblock size from variable size

Yurlungur (commented):
@pgrete sorry it took me so long to get to this. I will check that it works with Phoebus today. In the meantime, I have a few comments below. My biggest comment is I'd like to decouple variable size from meshblock size for the purposes of outputs. I think this should be a small enough change we should just do it here. But if that's not the case, we can proceed without that fix.

scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
pgrete: Will do (when I make final changes)

src/mesh/refinement_cc_in_one.hpp
pgrete: clang yelled at my that "template typename" is a   C++17 thing (though it worked but clobbered the compiler output) versus "template class" (which is okay for C++14)

src/outputs/io_wrapper.hpp
pgrete: We never returned a value (because if the open fails we call `PARTHENON_FAIL` inside that function) so I considered it a cleanup.

src/outputs/parthenon_hdf5.cpp
pgrete: IIRC JL rewrote the entire machinery (Athena++ used a binary format for restarts) so I thought the copyright was a copy & paste artifact from some skeleton file, but I'll double check.
If the code has just been modified I'll reintroduce the Athena++ copyright.

src/outputs/parthenon_hdf5.cpp
pgrete: yes, I tried paraview and it looked fine. Nevertheless, might be good if you double check when you try this downstream anyway.

src/outputs/parthenon_hdf5.cpp
pgrete: Given that I originally thought this PR was also "just a small change" but in the end there were quite a few not that obvious places, I decided to keep this separate and easier to review (especially as I also want to introduce a test case for the non cell centered output).
If you think it's better to combine I can also add the change here. Otherwise, I'll open a separate PR soonish (which should not affect downstream codes in terms of consistency as it just adds/fixes things).

pgrete:
One additional question: Apparently there's a CUDA compiler issue for some Cuda version, see failed pipeline:
```
/home/gitlab-runner/builds/TmLLi8kU/0/theias/hpc/jmstone/athena-parthenon/parthenon-ci-mirror/src/outputs/restart.hpp(146): error: name followed by "::" must be a class or namespace name
1 error detected in the compilation of "/home/gitlab-runner/builds/TmLLi8kU/0/theias/hpc/jmstone/athena-parthenon/parthenon-ci-mirror/src/mesh/mesh.cpp".
make[3]: *** [src/CMakeFiles/parthenon.dir/build.make:454: src/CMakeFiles/parthenon.dir/mesh/mesh.cpp.o] Error 1
make[3]: *** Waiting for unfinished jobs....
make[2]: *** [CMakeFiles/Makefile2:1665: src/CMakeFiles/parthenon.dir/all] Error 2
make[1]: *** [CMakeFiles/Makefile2:1780: example/advection/CMakeFiles/advection-example.dir/rule] Error 2
```
IIRC I didn't come across that issue in more recent cuda versions.
Is this worth fixing or can we live with this? (I am in the process of migrating that pipeline to a different machine where more recent cuda version are available).

src/mesh/refinement_cc_in_one.hpp
Yurlungur: ...huh ok. That's good to know for the future.

src/outputs/io_wrapper.hpp
Yurlungur: :+1:

src/outputs/parthenon_hdf5.cpp
Yurlungur: Sounds good.

src/outputs/parthenon_hdf5.cpp
Yurlungur: :+1: will do.

src/outputs/parthenon_hdf5.cpp
Yurlungur: Ah ok. Two small PRs works for me. You're right that's probably the better way to go. Just wanted to make sure that the meshblock size thing didn't get lost in the noise.

Yurlungur:
> 

Do we know what version of cuda this is a problem for? Depending on the version, we can probably live with it.

src/interface/sparse_pack.hpp
```diff
@@ -144,7 +144,7 @@ class SparsePack : public SparsePackBase {
   static std::tuple<SparsePack, SparsePackIdxMap>
   Get(T *pmd, const VAR_VEC &vars, const std::vector<MetadataFlag> &flags = {},
       bool fluxes = false, bool coarse = false) {
-    static_assert(sizeof...(Ts) == 0);
+    static_assert(sizeof...(Ts) == 0, "not sure");
```
lroberts36:
```suggestion
    static_assert(sizeof...(Ts) == 0, "Cannot create a string/type hybrid pack");
```

src/interface/sparse_pack.hpp
```diff
@@ -201,12 +201,12 @@ class SparsePack : public SparsePackBase {
   }
 
   KOKKOS_INLINE_FUNCTION int GetLowerBound(const int b, PackIdx idx) const {
-    static_assert(sizeof...(Ts) == 0);
+    static_assert(sizeof...(Ts) == 0, "not sure");
```
lroberts36:
```suggestion
    static_assert(sizeof...(Ts) == 0, "Cannot create a string/type hybrid pack");
```

src/interface/sparse_pack.hpp
```diff
@@ -201,12 +201,12 @@ class SparsePack : public SparsePackBase {
...
     return bounds_(0, b, idx.VariableIdx());
   }
 
   KOKKOS_INLINE_FUNCTION int GetUpperBound(const int b, PackIdx idx) const {
-    static_assert(sizeof...(Ts) == 0);
+    static_assert(sizeof...(Ts) == 0, "not sure");
```
lroberts36:
```suggestion
    static_assert(sizeof...(Ts) == 0, "Cannot create a string/type hybrid pack");
```

src/interface/sparse_pack.hpp
```diff
@@ -235,7 +235,7 @@ class SparsePack : public SparsePackBase {
   KOKKOS_INLINE_FUNCTION
   Real &operator()(const int b, PackIdx idx, const int k, const int j,
                    const int i) const {
-    static_assert(sizeof...(Ts) == 0);
+    static_assert(sizeof...(Ts) == 0, "not sure");
```
lroberts36:
```suggestion
    static_assert(sizeof...(Ts) == 0, "Cannot create a string/type hybrid pack");
```

src/interface/sparse_pack.hpp
```diff
@@ -264,7 +264,7 @@ class SparsePack : public SparsePackBase {
   KOKKOS_INLINE_FUNCTION
   Real &flux(const int b, const int dir, PackIdx idx, const int k, const int j,
              const int i) const {
-    static_assert(sizeof...(Ts) == 0);
+    static_assert(sizeof...(Ts) == 0, "Not sure");
```
lroberts36:
```suggestion
    static_assert(sizeof...(Ts) == 0, "Cannot create a string/type hybrid pack");
```

src/outputs/parthenon_hdf5.cpp
lroberts36: Agreed that two small PRs is better.

lroberts36 (approved):
LGTM

Yurlungur (approved):
Just tested this in Phoebus and generated some output in VISIT. Looks good. We have some downstream analysis tools that will need to change, but that's on us.

src/outputs/parthenon_hdf5.cpp
pgrete: I just compared the `athena_hdf5.cpp` to our `parthenon_hdf5.cpp` and they look very different. So I'll stick to the updated copyright.

pgrete:
> Just tested this in Phoebus and generated some output in VISIT. Looks good. We have some downstream analysis tools that will need to change, but that's on us.

Great. I also have to update the `yt` frontend, but that's also downstream.


> Do we know what version of cuda this is a problem for? Depending on the version, we can probably live with it.

It's Cuda 11.1, so already ~2 years old. I now removed those (simple) regression tests in favor of reintroducing more advanced ones in the next few weeks. Our admin is currently preparing the new machine here so that we can run containers with A100 support.

pgrete:
Turns out the build issue was legit as it captured the "non-hdf5" buil, which was broken.
I added a couple more `#ifdef ENABLE_HDF5` (which I don't like) and everything should work now.
With these additional `ifdefs` #644 seems now even higher on the cleanup list.
Also, we should add build configs without the optional dependencies to the overall build matrix, see #722

PR #729, opened 2022-09-27
Optional modifications to output format

brryan:
## PR Summary

Currently parthenon hard-codes a 5-digit integer ID for output file names. This can be restrictive for very long-running applications. Also, the output file produced at simulation completion is always named `*.final.*`, which can confuse some visualization software packages. This PR adds optional runtime input parameters to choose the number of digits in the output file ID and to opt out of the `*.final.*` naming convention *only* for output, not restart, files. 

## PR Checklist

- [ ] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/src/outputs/outputs.cpp
+++ b/src/outputs/outputs.cpp
@@ -150,6 +150,10 @@ Outputs::Outputs(Mesh *pm, ParameterInput *pin, SimTime *tm) {
       // set file number, basename, id, and format
       op.file_number = pin->GetOrAddInteger(op.block_name, "file_number", 0);
       op.file_basename = pin->GetOrAddString("parthenon/job", "problem_id", "parthenon");
+      op.file_number_width =
+          pin->GetOrAddInteger("parthenon/job", "file_number_width", 5);
+      op.file_label_final =
+          pin->GetOrAddBoolean("parthenon/job", "use_final_label", true);
       char define_id[10];
       std::snprintf(define_id, sizeof(define_id), "out%d",
                     op.block_number); // default id="outN"
--- a/src/outputs/outputs.cpp
+++ b/src/outputs/outputs.cpp
@@ -150,10 +150,8 @@ Outputs::Outputs(Mesh *pm, ParameterInput *pin, SimTime *tm) {
       // set file number, basename, id, and format
       op.file_number = pin->GetOrAddInteger(op.block_name, "file_number", 0);
       op.file_basename = pin->GetOrAddString("parthenon/job", "problem_id", "parthenon");
-      op.file_number_width =
-          pin->GetOrAddInteger("parthenon/job", "file_number_width", 5);
-      op.file_label_final =
-          pin->GetOrAddBoolean("parthenon/job", "use_final_label", true);
+      op.file_number_width = pin->GetOrAddInteger(op.block_name, "file_number_width", 5);
+      op.file_label_final = pin->GetOrAddBoolean(op.block_name, "use_final_label", true);
       char define_id[10];
       std::snprintf(define_id, sizeof(define_id), "out%d",
                     op.block_number); // default id="outN"
--- a/src/outputs/outputs.cpp
+++ b/src/outputs/outputs.cpp
@@ -3,7 +3,7 @@
 // Copyright(C) 2014 James M. Stone <jmstone@princeton.edu> and other code contributors
 // Licensed under the 3-clause BSD License, see LICENSE file for details
 //========================================================================================
-// (C) (or copyright) 2020-2021. Triad National Security, LLC. All rights reserved.
+// (C) (or copyright) 2020-2022. Triad National Security, LLC. All rights reserved.
 //
 // This program was produced under U.S. Government contract 89233218CNA000001 for Los
 // Alamos National Laboratory (LANL), which is operated by Triad National Security, LLC
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Current develop
 
 ### Added (new features/APIs/variables/...)
+- [[PR 729]](https://github.com/lanl/parthenon/pull/729) Optional modifications to output format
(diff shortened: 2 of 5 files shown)


src/outputs/outputs.cpp
```diff
@@ -150,6 +150,10 @@ Outputs::Outputs(Mesh *pm, ParameterInput *pin, SimTime *tm) {
       // set file number, basename, id, and format
       op.file_number = pin->GetOrAddInteger(op.block_name, "file_number", 0);
       op.file_basename = pin->GetOrAddString("parthenon/job", "problem_id", "parthenon");
+      op.file_number_width =
+          pin->GetOrAddInteger("parthenon/job", "file_number_width", 5);
+      op.file_label_final =
+          pin->GetOrAddBoolean("parthenon/job", "use_final_label", true);
```
Yurlungur: Maybe default to 8 and false respectively? Ah, maybe this would break tests though.

Yurlungur (approved):
Is there a reason this is WIP? Seems a simple enough change we can just go ahead?

src/outputs/outputs.cpp
brryan: Yeah current behavior has been standard for long enough that I think it's better to leave it as the default to avoid breaking others' workflows

pdmullen: Should this belong to `parthenon/job` or `op.block_name`?

brryan:
I switched the logic around a bit so the output formatting options are now per-output block (these parameters didn't quite make sense in the `parthenon/job` input block I don't think) and opting out of the `final` labeling convention for restarts *is* possible now.

src/outputs/outputs.cpp
brryan: @pdmullen Yep just changed that around

brryan:
Also I checked that this works locally with the `advection` example but I didn't add any tests for this small feature.

pdmullen (approved):

pgrete:
Have you tested the behavior when the simulation reaches the walltime limit (or another interrupt) but not the final time yet?
The main motivation for the `final` label was that if the sim was interrupted, the file to restart would always be the "final" one (which is convenient for chained job scripts).
This also applied for other types of variables outputs so that the numbering is not messed up as otherwise an walltime interrupted output may yield a new number in the variable output so that simulation times and output numbers are not synced any more.

We may want to revisit this next week and confirm that the behavior with interrupted outputs is still what we want.

brryan:
@pgrete Thanks for noticing this, I didn't appreciate that use of `final` for restarts and yes this changeset should break that. My changes are off by default at least so this shouldn't affect anyone's existing workflows until I fix this.

I'm not sure I understand the concern about walltime interrupted dumps, is the issue that an interrupted simulation would write e.g. `*.00006.phdf` instead of `*.final.phdf`, and then the restarted simulation would write another `*.00006.phdf`? This seems OK to me because the restarted simulation will write a dump file at later time.

That being said maybe it's best/easy to just add another check that ignores `use_final_label=false` if a simulation is walltime-interrupted? Happy to discuss next week in person!

Yurlungur:
>  I didn't appreciate that use of final for restarts and yes this changeset should break that. My changes are off by default at least so this shouldn't affect anyone's existing workflows until I fix this.

Wait I thought that for restarts in particular, the `final` was still used?

> I'm not sure I understand the concern about walltime interrupted dumps, is the issue that an interrupted simulation would write e.g. *.00006.phdf instead of *.final.phdf, and then the restarted simulation would write another *.00006.phdf? This seems OK to me because the restarted simulation will write a dump file at later time.

I think the issue is that upon restart, the next output will be `*.00007.phdf` which means that if the output cadence was, e.g., 1 in physical time, then `*.00007.phdf` corresponds to 6 units of physical time, not 7. 

I think just ignoring the `use_final_label=false` for walltim-interrupted outputs is a good solution personally.

pdmullen:
> I think the issue is that upon restart, the next output will be *.00007.phdf which means that if the output cadence was, e.g., 1 in physical time, then *.00007.phdf corresponds to 6 units of physical time, not 7.

This isn't the case though, right? With
```
<parthenon/output0>
file_type = hdf5
dt = 0.05
variables = prim
id = prim
use_final_label = false

<parthenon/output1>
file_type = rst
dt = 0.1
```

`mpiexec -n 4 ./athena -i sod.in -d r0 -t 00:01:00` may, for example, produce a walltime interrupted `r0/sod.prim.00000006.phdf*` and `r0/sod.out1.final.rhdf`.  

then 

`mpiexec -n 4 ./athena -r r0/sod.out1.final.rhdf -d r1 -t 00:01:00` will produce a `r1/sod.prim.00000006.phdf*` as its first primitive dump.  

Wherein `r0/sod.prim.00000006.phdf*` and `r1/sod.prim.00000006.phdf*` correspond to primitive dumps at different times.  The former at the walltime interrupted simulation time; the latter at the expected dt cadence.  

> I think just ignoring the use_final_label=false for walltim-interrupted outputs is a good solution personally.

Currently, `use_final_label` is an optional input parameter for each output block, whose default value is `true`.  Any reason we want to force the user's hand at having the `final` id for walltime-interrupted outputs?  I suspect that `use_final_label=false` for restarts would still work perfectly fine, just that it may not be as handy for recursive jobs.

Yurlungur:
> Wherein r0/sod.prim.00000006.phdf* and r1/sod.prim.00000006.phdf* correspond to primitive dumps at different times. The former at the walltime interrupted simulation time; the latter at the expected dt cadence.

Wait is this actually true? Does the dump number not get iterated internally and then stored to the restart? I.e., the first output from the restart will be `*7.phdf`? Maybe one of us should just run the relevant code and see...

> Currently, use_final_label is an optional input parameter for each output block, whose default value is true. Any reason we want to force the user's hand at having the final id for walltime-interrupted outputs? I suspect that use_final_label=false for restarts would still work perfectly fine, just that it may not be as handy for recursive jobs.

Ah good point it's per block. So the expected behavior is retained, and users can decide what they want to do.

Yurlungur:
> Wait is this actually true? 

Yep you're right. So I think the indexing for output files works fine.

PR #732, opened 2022-09-27
Add `Metadata::ForceRemeshComm` Flag

lroberts36:
## PR Summary

Add a metadata flag `Metadata::RemeshComm` that guarantees that the values of a field will be communicated between ranks during remeshing. Fields with `Metadata::Independent` and `Metadata::FillGhost` are also automatically communicated, as was previously done, so this should not change the behavior of existing downstream codes. 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


Yurlungur (approved):
LGTM

Yurlungur:
@jdolence @pgrete could we get this reviewed and in? I need it for the prolongation/restriction stuff

Yurlungur:
Ping @forrestglines @jdolence @pgrete this is a small change and needs approval

lroberts36:
Ping @forrestglines @jdolence @pgrete for a review

docs/interface/Metadata.md
```diff
@@ -132,6 +132,14 @@ classes may be allocated. The behaviours are the following:
   `Metadata::WithFluxes` and `Metadata::FillGhosts` to send flux
   corrections across meshblocks.
 
+- If `Metadata::RemeshComm` is set, the variable is communicated between 
+  ranks during remeshing. Variables with `Metadata::Independent` and/or 
+  `Metadata::FillGhost` are also automatically communicated when a block 
+  is communicated from one process to another. Other variables **are not** 
+  communicated across ranks, which has the possibility to cause downstream 
+  codes to produce different results when the same problem is run on 
+  different numbers of ranks.
```
pgrete: I'm not sure I follow the comment.
How can running on a different number of ranks produce different results?
I could image this happening when a variable is neither `independent` nor `dependent`, but think we should not allow for that case.

src/mesh/meshblock.cpp
```diff
@@ -171,20 +171,33 @@ void MeshBlock::Initialize(int igid, int ilid, LogicalLocation iloc,
...
-  for (int n = 0; n < vars.size(); n++) {
+  for (int n = 0; n < vars.size(); ++n) {
     RegisterMeshBlockData(vars[n]);
   }
 
   if (pm->multilevel) {
+    const auto refine_vars =
+        real_container
+            ->GetVariablesByFlag({Metadata::Independent, Metadata::FillGhost}, false)
+            .vars();
```
pgrete, line 195: Likely related to my question above: In which case you want to communicate a variable but not refine it?
I'm probably missing the big picture.

docs/interface/Metadata.md
lroberts36: When a block is moved in the block list within a rank, all of the fields are copied over. When a block is moved between ranks, only independent data is transferred on the assumption that the dependent data can be reconstituted on the new rank from the dependent data. This is normally the case, but sometimes it may be the case that it is too expensive or introduces numerical noise to rebuild some `dependent` fields. Then, it is useful to have a flag that indicates the data in particular dependent variables should be communicated during load balancing.

src/mesh/meshblock.cpp
lroberts36: We were running into issues in Riot where results differed when run on different numbers of ranks, which that seemed to be related to certain `dependent` variables not being initialized after blocks were moved between ranks. Presumably, by very carefully reconstituting the dependent variables, we could get bitwise agreement independent of number of ranks, but a simpler solution was to just communicate some dependent data. 

When refining, we still only want to refine `independent` variables.

docs/interface/Metadata.md
pgrete: > When a block is moved in the block list within a rank, all of the fields are copied over. 

I thought we'd should move around the pointer to the block and update the data within the block and not copy the data itself.

> When a block is moved between ranks, only independent data is transferred on the assumption that the dependent data can be reconstituted on the new rank from the dependent data.

Yes, this is/was also my working assumption when I wrote the comment.

> This is normally the case, but sometimes it may be the case that it is too expensive or introduces numerical noise to rebuild some dependent fields. Then, it is useful to have a flag that indicates the data in particular dependent variables should be communicated during load balancing.

I'm probably a little hesitant as I'm afraid that the proposed functionality/fix hides a deeper problem and/or invites for some "unfavorable" design choices downstream.
While I understand that there's some extra cost involved in order to rebuild dependent data, I'm currently weighing the "strict" design decision that "everything derived is handled by `FillDerived` at clear points in the dataflow" as a stronger argument than performance (but probably your performance use case is vastly different from the ones I have in mind).

How about we discuss this tomorrow during the call? I'd like to hear feedback from more people on this as I consider this a breaking change (with respect to the underlying design decision).

lroberts36: Sure, I am happy to discuss it during the call tomorrow.

lroberts36:
@pgrete: I updated the documentation to reflect our discussion and renamed the flag to `ForceRemeshComm`. Can you please approve when you get the change?

pgrete (approved):

PR #734, opened 2022-10-03
[WIP]: Cell Count Example

Yurlungur:
## PR Summary

I recently wanted to scope out a simulation cost, and thought "why not let Parthenon do the work for me?" That led me to developing this cute little example, which I figured I'd push upstream. The idea here is to have Parthenon generate a mesh tree/hierarchy based on some estimated needed resolution (say 100m resolution for a neutron star) in some region (say a sphere of some radius) without actually allocating any variables. Then have Parthenon report how big the mesh is.

The way I implement this is I create a package with no variables, but a custom mesh refinement check that checks to see if the desired resolution has been reached in the desired region. (Just for simplicity, this is hardcoded to a sphere. I'm not trying to create a full-featured capability here. If you want to spec something out other than a single isolated star, I think you could use this example as a jumping off point pretty easily.)

The example has no driver. I just init packages and mesh, run a loop through the mesh to count total number of cells heuristically, and print the results. Results can also be saved to a text file.

To visualize the mesh structure, I also wrote a little script that can read a `phdf` file and plot the meshblocks as rectangles, similar to the `movie2d` script. I initialize outputs in my little example (no driver so it must be done manually) and spit out an hdf5 file if requested. I found that by far the slowest part the calculation was outputting the mesh---in particular, the coordinates grids. So I added a new hdf5 output type, which I call `gridh5`. This is the standard `phdf` format, but instead of outputting full cell and face coordinates, it just outputs the minimum and maximum face coordinate values for each meshblocks. To do so I had to make a few one-line changes (and some indentation) to the `parthenon_hdf5` output.

Outputs visualized with my little script, which I called `plot_mesh.py` look like this:
image

Those boxes are meshblocks, not cells. This particular grid had 32^3 meshblocks. This is an X3=0 slice. Grids with cell counts of 100s of millions of cells can be generated and visualized in less than a minute.

I hooked this machinery into regression tests, but because I didn't want to deal with a new gold file release, I didn't hook the `gridh5` into the regression tests, just a cell count.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


scripts/python/packages/parthenon_tools/setup.py
```diff
@@ -22,8 +22,8 @@
     description="Python helper scripts to be used with Parthenon",
     long_description=long_description,
     url="https://github.com/lanl/parthenon/tree/develop/scripts/python/parthenon_tools",
-    author="Josh Dolence, Philipp Grete, Andrew Gaspar, Joshua S. Brown",
-    author_email="joshbro42867@yahoo.com",
+    author="The Parthenon Team",
+    author_email="",
```
Yurlungur, line 26: Josh Brown has moved on so his email should probably not be here. That said, should we have an official Parthenon email?

Yurlungur:
@pgrete just a ping on this. I think there's a higher-level discussion to be had about this capability.

PR #735, opened 2022-10-04
clean up var naming and hdf5 access props

Yurlungur:
## PR Summary

As a base for I/O work does some common, uncontroversial cleanup for IO:
- Move Access prop into sub function
- Move filename generation into sub function

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


brryan (approved):
LGTM

jdolence (approved):
LGTM

src/outputs/parthenon_hdf5.cpp
```diff
@@ -343,145 +456,10 @@ void PHDF5Output::WriteOutputFileImpl(Mesh *pm, ParameterInput *pin, SimTime *tm
...
-    PARTHENON_MPI_CHECK(
-        MPI_Info_set(FILE_INFO_TEMPLATE, "cb_buffer_size", cb_buffer_size.c_str()));
-  }
-
-  /* tell the HDF5 library that we want to use MPI-IO to do the writing */
-  PARTHENON_HDF5_CHECK(H5Pset_fapl_mpio(acc_file, MPI_COMM_WORLD, FILE_INFO_TEMPLATE));
-#else
-  hid_t const acc_file = H5P_DEFAULT;
-#endif // ifdef MPI_PARALLEL
+  H5P const acc_file = H5P::FromHIDCheck(HDF5::GenerateFileAccessProps());
```
pgrete, line 462: I know this was already in the original code, but why is one `acc_file` `H5P` (for MPI) and the other `hid_t` (for serial)?

Also, now it seems that the function returns `hid_t` and here defined as `H5P`

Yurlungur: So `H5P` is an automatically garbage-collected wrapper around `hid_t` for property lists. I don't want it garbage collected in `GenerateFileAccessProps`, because I want to return the handle at the end of the function. Think of `GenerateFileAccessProps` as a factory. But it should be garbage collected at the end of `WriteOutputs` so  that we clean up after ourselves. So I take the `hid_t` that I now return, and wrap it in the garbage collected class.

pgrete (approved):

Issue #46, opened 2020-03-12
Metadata flag conversion not working with nvcc


pgrete:
In the process of integrating Kokkos I stumbled across a bug when compiling for Cuda (i.e., using the `nvcc_wrapper`) where a conversion from a list to a vector for  the `Metadata::flags` seems to fail.

```
[ 32%] Building CXX object src/CMakeFiles/parthenon.dir/better_boundaries/boundary_conditions.cpp.o
/home/pgrete/src/parthenon/src/better_boundaries/boundary_conditions.cpp: In function ‘void parthenon::ApplyBoundaryConditions(parthenon::Container<double>&)’:
/home/pgrete/src/parthenon/src/better_boundaries/boundary_conditions.cpp:29:60: error: expected ‘;’ before ‘}’ token
     ContainerIterator<Real> citer(rc, {m.independent});
                                                            ^
                                                            ;
/home/pgrete/src/parthenon/src/better_boundaries/boundary_conditions.cpp:29:62: error: no matching function for call to ‘parthenon::ContainerIterator<double>::ContainerIterator(parthenon::Container<double>&, parthenon::Metadata::flags)’
     ContainerIterator<Real> citer(rc, {m.independent});
                                                              ^
/home/pgrete/src/parthenon/src/interface/ContainerIterator.hpp:41:1: note: candidate: ‘parthenon::ContainerIterator<T>::ContainerIterator(parthenon::Container<T>&, const std::vector<parthenon::Metadata::flags>&) [with T = double]’
   ContainerIterator<T>(Container<T>& c, const std::vector<Metadata::flags> &flagVector) {
 ^ ~~~~~~~~~~~~~~~
/home/pgrete/src/parthenon/src/interface/ContainerIterator.hpp:41:1: note:   no known conversion for argument 2 from ‘parthenon::Metadata::flags’ to ‘const std::vector<parthenon::Metadata::flags>&’
/home/pgrete/src/parthenon/src/interface/ContainerIterator.hpp:31:7: note: candidate: ‘parthenon::ContainerIterator<double>::ContainerIterator(const parthenon::ContainerIterator<double>&)’
 class ContainerIterator {
       ^~~~~~~~~~~~~~~~~
/home/pgrete/src/parthenon/src/interface/ContainerIterator.hpp:31:7: note:   candidate expects 1 argument, 2 provided
[...]

PR #741, opened 2022-10-05
Draft: Try to fix build issue with hdf5 on mac

jdolence:
## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted

### DIFF
--- a/src/outputs/parthenon_hdf5.hpp
+++ b/src/outputs/parthenon_hdf5.hpp
@@ -103,14 +103,18 @@ using H5P = H5Handle<&H5Pclose>;
 using H5T = H5Handle<&H5Tclose>;
 using H5S = H5Handle<&H5Sclose>;
 
-// Static functions to return HDF type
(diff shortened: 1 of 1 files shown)


tagardiner:
The code compiles on a mac, but it does so with a warning.

/Users/tagardi/src/parthenon.git/src/outputs/parthenon_hdf5.hpp:117:1: warning: non-void function does not return a value [-Wreturn-type]
}

This is pointing out that the function does not have a default return.  One option is to declare a variable of type hid_t before the if block and initialize it in the if block and then return the result.  Does HDF5 have a semaphore for "no type" to catch a failure to match?

tagardiner:
Running the unit tests we also have a failure

      Start 46: regression_test:restart
46/53 Test #46: regression_test:restart .................................***Failed   12.11 sec

Rerunning with

$ ctest --rerun-failed --output-on-failure

shows that this test fails with sig-abort.

PR #742, opened 2022-10-05
Draft: Multiple types/packs for multiple types

brryan:
## PR Summary

We would like:
- Packing for particle variables (which can be both `Real` and `int` types)
- ~~Support for arbitrary types for parthenon cell variables.~~

This PR attempts to implement this by
- [X] Templating `SparsePack`s on a datatype
- [ ] ~~Replacing `varVector_` and `varMap_` of `CellVariable` in `Mesh*Data` with a `Params`-like type-erased container that can return a vector of all `CellVariables` of a particular type~~
- ~ ] ~~Replace `Add` for `CellVariables` with a function templated on datatype (maybe defaulting to `Real`)~~
- [ ] Modify `SparsePack` to also support packing `ParticleVariable`s of a particular `Swarm`
- [ ] ~~Resolve buffer packing of different-sized types~~

Allowing arbitrary types in `CellVariable`s has some potential complications that we may want to just deliberately avoid at least for now. For example how does one prolongate/restrict a cell-centered `enum`? If non-`Real` `CellVariable`s are enrolled do we just require for now that they not have `FillGhost`? It might also be possible to allow/require developers to enroll their own prolongation/restriction operators for non-`Real` (or also `Real`?) types.

There also needs to be some rule for packing non-`Real`-sized data into buffers. The easiest thing (that particles already do in a similar context) is to assume other types are the same size as or smaller than `Real` and then just use casts going in and out of the buffer.

If the custom types framework were sufficiently flexible one could use vector types to have a fastest moving index other than `i` as @forrestglines pointed out the other day.

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted


tst/unit/test_sparse_pack.cpp
```diff
@@ -128,10 +128,10 @@ TEST_CASE("Test behavior of sparse packs", "[SparsePack]") {
            "blocks") {
         // Create a pack use type variables
         auto sparse_pack =
-            parthenon::SparsePack<v5, v3>::Get(&mesh_data, {Metadata::WithFluxes});
+            parthenon::SparsePack<v5, v3>::Get<MeshData<Real>, Real>(&mesh_data, {Metadata::WithFluxes});
```
brryan: It would be great to *not* have to be explicit about `MeshData<Real>` in the template parameter list here but I couldn't immediately figure out how to do it.

pdmullen:
@brryan Can we close this one?

PR #753, opened 2022-10-06
Cleanup UniformCartesian

forrestglines:
## PR Summary

Cleanup the current UniformCartesian coordinate class to have fewer functions, utilizing templates on coordinate direction to reduce the number of functions. Done in preparation to implement other coordinate system classes.

Here are the new functions and any differences highlighted

Position of elements (previously`x1v`, `x1f`, and `x1s2`*)
```
Xc<int dir>(const int idx); //Cell Centers
Xf<int face, int dir>(const int idx); //Face Centers
Xe<int edge, int dir>(const int idx); //Edge Centers
Xn< int dir>(const int idx); //Cell Nodes
```
where `1<=dir<=3` is which component of the direction being queried (i.e. x,y,z), `face` and `edge` specify which face/edge is being queried, and `idx` is the index along `dir`. The addition of `face` and `edge` along with the functions `Xe` and `Xn` are new. These allow for consistently getting the positions of faces, edges, and nodes with one function without mixing calls to `Xc` and `Xf` to get the position of off-cell-center locations.

*In particular with `Xf`, when `face==dir` it returns the same as `x1f` etc. did in the old version. Now when `face!=dir`, it returns the face-averaged position of the other directions, replacing `x1s2`, `x2s3`, etc.

I've also dropped versions using `i,j,k` since the uniform cartesian, spherical, and cylindrical coordinates do not need them. If we want non-uniform coordinate systems later, we'll have to go back and change them.

Distance between elements (previously `dx1v`, `dx1f` etc. )
```
Dxc<int dir>(const int idx); //Distance between cell centers along dir
Dxf<int dir, int face>(const int idx); //Distance between face centers along dir
Dxe<int dir, int edge>(const int idx); //Distance between edge centers along dir
Dxn< int dir>(const int idx); //Distance between cell Nodes along dir
```
Likewise, `face`, `edge`, `Dxe`, and `Dxn` are new to get distances between different elements along different dimensions.

Integration elements (previously `Dx`, `Area`, `Edgelength`, and `Volume`)
```
CellWidth<int dir>( const int k, const int j, const int i); //Cell width at cell center
FaceArea<int face>( const int k, const int j, const int i); //Area of Face
EdgeLength<int edge>( const int k, const int j, const int i); //Length of edge
CellVolume( const int k, const int j, const int i); //Cell Volume
```
These functions take all three indices to accommodate extension to spherical and cylindrical. They need `k`,`j`, and `i` for the sake of spherical and cylindrical coordinates but for now these arguments are unenforced in Parthenon.

~~Area-averaged positions on face-centers (previously `x1s2`, `x3s1` etc. )~~
```
Xs<int dir, int face>( int idx);
```
~~These are used in the AMR machinery for face-centered fields, which is currently missing from Parthenon. However, I believe these are identical to `Xf<int dir, int face>` for the uniform coordinates proposed. `Xs` might not be necessary anymore.~~
*I removed `Xs` since the new `Xf` can give all the same information unambiguously.

For each of these functions templated on `dir`, `face`, `edge`, etc, we can introduce undercase version with runtime parameters, i.e.
```
xc( const int dir, const int idx); //Cell Centers
faceArea( const int dir, const int k, const int j, const int i); //Face Areas
```
This convention avoids naming conflicts between simplified functions and runtime functions. We can implement these later as needed.

We'll probably need to implement more coordinate functions as we need them for cylindrical and spherical coordinates.

Also, here is a script to do renaming of coordinates in downstream codes:
```
#!/bin/bash

#Rename x\1v( -> Xc<\1>(
find ./ -name "*.*pp" -exec sed -i "s/\.x\([123]\)v(/.Xc<\1>(/g" {} \;

#Rename x\1f( -> Xf<\1>(
find ./ -name "*.*pp" -exec sed -i "s/\.x\([123]\)f(/.Xf<\1>(/g" {} \;

#Rename x\1e( -> Xe<\1>(
find ./ -name "*.*pp" -exec sed -i "s/\.x\([123]\)e(/.Xe<\1>(/g" {} \;

#Rename x\1s\2( -> Xs<\1,\2>(
find ./ -name "*.*pp" -exec sed -i "s/\.x\([123]\)s\([123]\)(/.Xs<\1,\2>(/g" {} \;

#Rename dx\1v -> Dxc<\1>)
find ./ -name "*.*pp" -exec sed -i "s/\.dx\([123]\)v(/.Dxc<\1>(/g" {} \;

#Rename dx\1f -> Dxf<\1>
find ./ -name "*.*pp" -exec sed -i "s/\.dx\([123]\)f(/.Dxf<\1>(/g" {} \;

#Rename Dx(, -> CellWidthFA(,
find ./ -name "*.*pp" -exec sed -i "s/\.Dx(/.CellWidthFA(/g" {} \;

#Rename Area(, -> FaceAreaFA(,
find ./ -name "*.*pp" -exec sed -i "s/\.Area(/.FaceAreaFA(/g" {} \;

#Rename Volume(, -> CellVolmue(,
find ./ -name "*.*pp" -exec sed -i "s/\.Volume(/.CellVolume(/g" {} \;
```

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted


forrestglines:
@par-hermes format

forrestglines:
Here's the interface that I'd like to propose for `UniformCartesian` and future coordinate systems moving forward:

Position of elements (previously`x1v`, `x1f` etc)
```
Xc<int dir>( k, j, i); //Cell Centers
Xf<int face, int dir>( k, j, i); //Face Centers
Xe<int edge, int dir>( k, j, i); //Edge Centers
Xn< int dir>( k, j, i); //Cell Nodes
```
Distance between elements (previously `dx1v`, `dx1f` etc. )
```
Dxc<int dir>( k, j, i); //Distance between cell centers along dir
Dxf<int face, int dir>( k, j, i); //Distance between face centers along dir
Dxe<int edge, int dir>( k, j, i); //Distance between edge centers along dir
Dxn< int dir>( k, j, i); //Distance between cell Nodes along dir
```
Integration elements
```
Dx<int dir>( k, j, i); //Distance between cell centers along dir
Da<int face>( k, j, i); //Area of Face
Ds<int edge>( k, j, i); //Length of edge
Dv( k, j, i); //Volume
```
Area-averaged (face?) positions for AMR on face-centers (previously `x1s2` etc. )
```
Xs<int dir, int face>( k, j, i)
```
We'll probably need to implement more coordinate functions as we need them for cylindrical and spherical coordinates.

Note that each of these functions accept `i,j,k` and not just a single index -- I think we should provide and use these as a default in Parthenon because they provide flexibility down the road and they avoid bugs where the wrong index is passed into the coordinate functions (see the error in Athena++ pointed out by Shengtai). However, we can still provide the "minimally indexed" functions for simple coordinate systems. For example for `UniformCartesian`, we have
```
Xc<int dir>(int i);
Dx<int dir>();
Area<int dir>();
etc...
```
Alternatively, we could only provide the bare minimum for a limited number of coordinate systems.

I'd liked to adopt the convention for uppercase first letter to be a templatized function discerned at compile time and undercase first letter to be determined at runtime. I.e. we'd implement these as well:
```
xc( int dir, k, j, i); //Cell Centers
xc( int dir); //Cell Centers
dx<int dir>( k, j, i);
dx<int dir>();
```
This convention avoids naming conflicts between simplified functions and runtime functions. 

Also note that this would lead to  4 different implementations for each function. I'm not sure how much templates with variable arguments would help.

Yurlungur:
I'd be ok with this set. If we don't allow for non-uniform mesh spacing (in coordinate space) we might be able to reduce the number of functions here a little bit, which might be nice.

forrestglines:
I ended up implementing a slightly different interface for `UniformCartesian` between my last comment and what's currently implemented/inherited from Athena++. Here are the new functions and any differences highlighted

Position of elements (previously`x1v`, `x1f`, and `x1s2`*)
```
Xc<int dir>(const int idx); //Cell Centers
Xf<int face, int dir>(const int idx); //Face Centers
Xe<int edge, int dir>(const int idx); //Edge Centers
Xn< int dir>(const int idx); //Cell Nodes
```
where `1<=dir<=3` is which component of the direction being queried (i.e. x,y,z), `face` and `edge` specify which face/edge is being queried, and `idx` is the index along `dir`. The addition of `face` and `edge` along with the functions `Xe` and `Xn` are new. These allow for consistently getting the positions of faces, edges, and nodes with one function without mixing calls to `Xc` and `Xf` to get the position of off-cell-center locations.

*In particular with `Xf`, when `face==dir` it returns the same as `x1f` etc. did in the old version. Now when `face!=dir`, it returns the face-averaged position of the other directions, replacing `x1s2`, `x2s3`, etc.

I've also dropped versions using `i,j,k` since the uniform cartesian, spherical, and cylindrical coordinates do not need them. If we want non-uniform coordinate systems later, we'll have to go back and change them.

Distance between elements (previously `dx1v`, `dx1f` etc. )
```
Dxc<int dir>(const int idx); //Distance between cell centers along dir
Dxf<int dir, int face>(const int idx); //Distance between face centers along dir
Dxe<int dir, int edge>(const int idx); //Distance between edge centers along dir
Dxn< int dir>(const int idx); //Distance between cell Nodes along dir
```
Likewise, `face`, `edge`, `Dxe`, and `Dxn` are new to get distances between different elements along different dimensions.

Integration elements (previously `Dx`, `Area`, `Edgelength`, and `Volume`)
```
CellWidth<int dir>( const int k, const int j, const int i); //Cell width at cell center
FaceArea<int face>( const int k, const int j, const int i); //Area of Face
EdgeLength<int edge>( const int k, const int j, const int i); //Length of edge
Cell Volume( const int k, const int j, const int i); //Cell Volume
```
These functions take all three indices to accommodate extension to spherical and cylindrical. They need `k`,`j`, and `i` for the sake of spherical and cylindrical coordinates but for now these arguments are unenforced in Parthenon.

~~Area-averaged positions on face-centers (previously `x1s2`, `x3s1` etc. )~~
```
Xs<int dir, int face>( int idx);
```
~~These are used in the AMR machinery for face-centered fields, which is currently missing from Parthenon. However, I believe these are identical to `Xf<int dir, int face>` for the uniform coordinates proposed. `Xs` might not be necessary anymore.~~
*I removed `Xs` since the new `Xf` can give all the same information unambiguously.

For each of these functions templated on `dir`, `face`, `edge`, etc, we can introduce undercase version with runtime parameters, i.e.
```
xc( const int dir, const int idx); //Cell Centers
faceArea( const int dir, const int k, const int j, const int i); //Face Areas
```
This convention avoids naming conflicts between simplified functions and runtime functions. We can implement these later as needed.

We'll probably need to implement more coordinate functions as we need them for cylindrical and spherical coordinates.

forrestglines:
@par-hermes format

Yurlungur:
> I ended up implementing a slightly different interface for UniformCartesian between my last comment and what's currently implemented/inherited from Athena++. Here are the new functions and any differences highlighted

I think these choices make sense. My one concern is regarding `Da` and `Ds` but I can't think of better names

forrestglines:
I encountered an error that I'm a little confused about as I tried out the new coordinates in AthenaPK. It occurred in this templated member function that was kind-of pointlessly templated on a coordinate class.
```
template <typename View1D, typename Coords>
PRhoProfile<View1D>
generate_P_rho_profile(parthenon::IndexRange ib, parthenon::IndexRange jb,
                     parthenon::IndexRange kb, Coords coords) const;
```
And when I call this function in the class
```
template <int dir, class ... args>
  KOKKOS_FORCEINLINE_FUNCTION Real Dxc(Args... args) const {
    assert(dir > 0 && dir < 4);
    return dx_[dir - 1];
  }
```
It causes this error with clang:
```
/vast/home/forrestglines/code/athenapk-project/athenapk/src/pgen/cluster/hydrostatic_equilibrium_sphere.cpp:176:28: error: reference to non-static member function must be called
  PARTHENON_REQUIRE(coords.Dxc<1>(0) == coords.Dxc<1>(1), "No equidistant grid in x1dir");
                    ~~~~~~~^~~
/vast/home/forrestglines/code/athenapk-project/athenapk/external/parthenon/src/utils/error_checking.hpp:36:9: note: expanded from macro 'PARTHENON_REQUIRE'
  if (!(condition)) {                                                                    \
        ^~~~~~~~~
```
and this error with CUDA/11.7.0:
```
/vast/home/forrestglines/code/athenapk-project/athenapk/src/pgen/cluster/hydrostatic_equilibrium_sphere.cpp:176:22: error: invalid operands of types ‘<unresolved overloaded function type>’ and ‘int’ to binary ‘operator<’
   PARTHENON_REQUIRE(coords.Dxc<1>(0) == coords.Dxc<1>(1), "No equidistant grid in x1dir");
           ~~~~~~~~~~~~~~^~~~
```
I don't think it's related to the `PARTHENON_REQUIRE` macro since commenting out those lines lead to the same error. I also tried removing the varidic template on `Dxc` but that had no effect. I got around the issue by removing the templating on `Coords` in AthenaPK so this is non-blocking for this PR. Does anyone have an idea what might have caused this error though?

Problematic AthenaPK code here:
[hydrostatic_equilibrium_sphere.hpp](https://github.com/parthenon-hpc-lab/athenapk/blob/97837c127da2040041369bd01ea5be569d131dee/src/pgen/cluster/hydrostatic_equilibrium_sphere.hpp#L156)

[hydrostatic_equilibrium_sphere.cpp](https://github.com/parthenon-hpc-lab/athenapk/blob/97837c127da2040041369bd01ea5be569d131dee/src/pgen/cluster/hydrostatic_equilibrium_sphere.cpp#L160)

Work-around AthenaPK code here:
[hydrostatic_equilibrium_sphere.hpp](https://github.com/parthenon-hpc-lab/athenapk/blob/01e363c3c704a31d0430f1b3a1fa33564131276c/src/pgen/cluster/hydrostatic_equilibrium_sphere.hpp#L156)

[hydrostatic_equilibrium_sphere.cpp](https://github.com/parthenon-hpc-lab/athenapk/blob/01e363c3c704a31d0430f1b3a1fa33564131276c/src/pgen/cluster/hydrostatic_equilibrium_sphere.cpp#L160)

forrestglines:
I used this script to update AthenaPK to use the new coordinate names. It should work for any other derivative Parthenon code (or any Parthenon PR as well).
```
#!/bin/bash

#Rename x\1v( -> Xc<\1>(
find ./ -name "*.*pp" -exec sed -i "s/\.x\([123]\)v(/.Xc<\1>(/g" {} \;

#Rename x\1f( -> Xf<\1>(
find ./ -name "*.*pp" -exec sed -i "s/\.x\([123]\)f(/.Xf<\1>(/g" {} \;

#Rename x\1e( -> Xe<\1>(
find ./ -name "*.*pp" -exec sed -i "s/\.x\([123]\)e(/.Xe<\1>(/g" {} \;

#Rename x\1s\2( -> Xs<\1,\2>(
find ./ -name "*.*pp" -exec sed -i "s/\.x\([123]\)s\([123]\)(/.Xs<\1,\2>(/g" {} \;

#Rename dx\1v -> Dxc<\1>)
find ./ -name "*.*pp" -exec sed -i "s/\.dx\([123]\)v(/.Dxc<\1>(/g" {} \;

#Rename dx\1f -> Dxf<\1>
find ./ -name "*.*pp" -exec sed -i "s/\.dx\([123]\)f(/.Dxf<\1>(/g" {} \;

#Rename Dx(, -> CellWidthFA(,
find ./ -name "*.*pp" -exec sed -i "s/\.Dx(/.CellWidthFA(/g" {} \;

#Rename Area(, -> FaceAreaFA(,
find ./ -name "*.*pp" -exec sed -i "s/\.Area(/.FaceAreaFA(/g" {} \;

#Rename Volume(, -> CellVolmue(,
find ./ -name "*.*pp" -exec sed -i "s/\.Volume(/.CellVolume(/g" {} \;
```

Edited once more for changes to API

Yurlungur:
> I encountered an error that I'm a little confused about as I tried out the new coordinates in AthenaPK.

That's bizarre. I have no idea.

forrestglines:
> I think these choices make sense. My one concern is regarding `Da` and `Ds` but I can't think of better names

I renamed these to `Dx`, `Area`, `EdgeLength`, and `Volume`. I'm thinking about being verbose in the final version to use `CellWidth`, `FaceArea`, `EdgeLength`, and `CellVolume`. I'd like to hear people's input on this but I'm leaning towards those names for the final edit since they're closer to Athena++'s names.

Yurlungur:
I like being verbose in this case. :+1:

forrestglines:
> I like being verbose in this case. 👍

I went along and made that change, updating the comments and bash script to reflect the more verbose integration element names

forrestglines:
@Yurlungur @lroberts36 @jdolence @pgrete @bprather

The coordinate rename is ready for discussion and review. Not everyone needs to review this PR, but I would like the downstream code maintainers to be on board with the coordinate rename.  I've provided a script to do the rename as well.

Are these renames useful? I think template parameterization on directions instead of a different function for each dimension  makes this worthwhile on its on.

c-prather:
I think since I subclass geometry, these changes affect me the least of any downstream; I'll be surprised if the updates touch more than two or three files.

I'm happy with the new names & especially the templating & especially especially the additional documentation.  I'm happy to take a closer look if other people are busy, but overall I'm happy if other downstreams are.

pgrete:
:+1:  for the new naming scheme. I'm also in favor of the more verbose description and cleaned up interface!

src/mesh/mesh_refinement.cpp
```diff
@@ -117,10 +117,10 @@ void MeshRefinement::RestrictFieldX1(const ParArrayND<Real> &fine,
...
           int i =
               (ci - pmb->c_cellbounds.is(interior)) * 2 + pmb->cellbounds.is(interior);
-          const Real area00 = coords.Area(X1DIR, k, j, i);
-          const Real area01 = coords.Area(X1DIR, k, j + 1, i);
-          const Real area10 = coords.Area(X1DIR, k + 1, j, i);
-          const Real area11 = coords.Area(X1DIR, k + 1, j + 1, i);
+          const Real area00 = coords.faceArea(X1DIR, k, j, i);
+          const Real area01 = coords.faceArea(X1DIR, k, j + 1, i);
+          const Real area10 = coords.faceArea(X1DIR, k + 1, j, i);
+          const Real area11 = coords.faceArea(X1DIR, k + 1, j + 1, i);
```
c-prather: Can this & below be `FaceArea<XNDIR>()`?  If not, is that something we can fix with constexpr?

c-prather (approved):
LGTM but confused about template vs runtime call use in one spot, see comment

src/mesh/mesh_refinement.cpp
c-prather: Similarly, is it worth standardizing all N->XNDIR using this PR, maybe with additional XNFACE variables?

forrestglines: > Can this & below be `FaceArea<XNDIR>()`? If not, is that something we can fix with constexpr?

These can be replaced, I just pushed that change along with a few other files.

> Similarly, is it worth standardizing all N->XNDIR using this PR, maybe with additional XNFACE variables?

Maybe? It would mean a lot more function name bloat (i.e. `x1f`->`Xf<X1DIR,X1DIR>`) when it should be clear from context that the integers refer to directions and it wouldn't add any extra checks/code safety.

c-prather: Agreed it's verbose and unwieldy to use the macros. I just kept forgetting what Xf<1, 1> was referring to.  Especially in terms of the argument order, since the convention switches by variable.  Xf<X1FACE, X1DIR> or Dxf<X1DIR, X1FACE> is at least totally explicit. (I absolutely mixed up both those orders the first time).

It also wouldn't add any type checks, true.  Which would make it hard to enforce, since using the macros would be purely a matter of convention and require much more typing.

...but that's given me a better (worse) idea: maybe we *should* type-check face locations vs edge locations vs directions?  I mean,  Xf<FaceLocation f, Direction d> seems like even more overkill, but at least then we get the compile-time check benefits and it's easy to enforce a readable standard.

Maybe the answer for now is just to use the `Xf<int>` shortcut everywhere?  That's unambiguous, resembles the old Athena convention, and it's all we need for UniformCartesian, right?  And, it and leaves room for a future typed solution, since changing the full macro only involves changing the shortcut and not a bunch of calls scattered around the code.

src/coordinates/uniform_cartesian.hpp
```diff
@@ -58,98 +58,77 @@ class UniformCartesian {
...
+    assert(dir > 0 && dir < 4 && face > 0 && face < 4);
+    return dx_[face - 1];
+  }
+
+  template <int dir>
+  KOKKOS_FORCEINLINE_FUNCTION Real Xc(const int idx) const {
     assert(dir > 0 && dir < 4);
-    return area_[dir - 1];
+    return xmin_[dir - 1] + (idx + 0.5) * dx_[dir - 1];
   }
```
Yurlungur: The choice to only accept one index here precludes coordinate systems where the coordinate directions are non-orthogonal, e.g., where `Xc` depends on `k`, `j`, and `i`. 

I think we talked about this at the meeting but I don't remember the conclusion. Did we decide that this was an acceptable limitation? If not, I suggest we include a set of 3d overloads.

As an aside, the extra overloads are convenient in that they're a little less error prone. For example, consider the following code snippet from Pheobus, which I had to change:
```C++
  r = coords.Xc<1>(k, j, i);
  th = coords.Xc<2>(k, j, i);
  ph = coords.Xc<3>(k, j, i);
```
in the original version, I didn't have to think about which index to pass into each function. With the new version, I had to write:
```C++
  r = coords.Xc<1>(i);
  th = coords.Xc<2>(j);
  ph = coords.Xc<3>(k);
```
which isn't a big deal. But I thought I'd throw it out there.

I can also imagine the extra overloads being useful for metaprogramming even if you don't need them. For example, consider the following code snippet, again from Phoebus, which needed to change:
```C++
  template <typename CoordinateSystem, class... Args>
  KOKKOS_INLINE_FUNCTION ConToPrimStatus operator()(const CoordinateSystem &geom,
                                                    const singularity::EOS &eos,
                                                    const Coordinates_t &coords,
                                                    Args &&...args) const {
    VarAccessor<T> v(var, std::forward<Args>(args)...);
    CellGeom g(geom, std::forward<Args>(args)...);
    Real x1 = coords.Xc<1>(std::forward<Args>(args)...);
    Real x2 = coords.Xc<2>(std::forward<Args>(args)...);
    Real x3 = coords.Xc<3>(std::forward<Args>(args)...);
    return solve(v, g, eos, x1, x2, x3);
  }
}
``` 
Now that only the scalar index is passed in, this needs specializations.

src/coordinates/uniform_cartesian.hpp
```diff
@@ -58,98 +58,77 @@ class UniformCartesian {
...
+    assert(dir > 0 && dir < 4);
+    return dx_[dir - 1];
   }
   template <class... Args>
-  KOKKOS_FORCEINLINE_FUNCTION Real dx1v(Args... args) const {
-    return dx_[0];
+  KOKKOS_FORCEINLINE_FUNCTION Real cellWidth(const int dir, Args... args) const {
+    assert(dir > 0 && dir < 4);
+    return dx_[dir - 1];
   }
```
Yurlungur: I found myself getting caught by "why isn't this template function working?" only to see that I missed the capitalization. I propose a slightly different name for the lower-case, takes the direction as a function argument, functions:
```C++
CellWidthNT
```
or 
```C++
CellWidthFA
```
to specify no-template, or function-arg respectively. That way it's more distinct to my case-insensitive eyes.

src/coordinates/uniform_cartesian.hpp
```diff
@@ -58,98 +58,77 @@ class UniformCartesian {
...
     assert(dir > 0 && dir < 4);
-    return area_[dir - 1];
+    return xmin_[dir - 1] + (idx + 0.5) * dx_[dir - 1];
   }
 
-  template <class... Args>
-  KOKKOS_FORCEINLINE_FUNCTION Real dx1f(Args... args) const {
-    return dx_[0];
+  template <int face>
+  KOKKOS_FORCEINLINE_FUNCTION Real Xf(const int idx) const {
```
Yurlungur: See above regarding overloads.

src/coordinates/uniform_cartesian.hpp
```diff
@@ -58,98 +58,77 @@ class UniformCartesian {
...
     return dx_[dir - 1];
   }
 
-  template <class... Args>
-  KOKKOS_FORCEINLINE_FUNCTION Real Area(const int dir, Args... args) const {
+  template <int dir, int face, class... Args>
+  KOKKOS_FORCEINLINE_FUNCTION Real Dxf(Args... args) const {
+    assert(dir > 0 && dir < 4 && face > 0 && face < 4);
+    return dx_[face - 1];
+  }
```
Yurlungur, line 87: I'm not sure why... but this function in particular seemed to throw C++'s type deduction through a loop. A number of places, where I had
```C++
const auto &coords = pmb->coords;
```
I had to change to
```C++
const parthenon::Coordinates_t &coords = pmb->coords;
```
for template resolution of, e.g.,
```C++
coords.DXf<1,1>(i)
```
to resolve properly.

Yurlungur (changes requested):
Overall I like the new naming scheme.

I was able to go through Phoebus and build and run with the new uniform-cartesian setup but it was not as trivial as I'd hoped. A number of locations in the code required manual intervention. I list a few questions/comments related to my manual interventions below.

IMO none of these are blocking, but I'd like to chat about them before we merge, so I'm requesting changes for now, and if you push back about the changes, I'll just click approve. :)

Yurlungur:
I also notice a number of tests seem to be failing... possibly related to the template type deduction issue I noticed above.

src/mesh/mesh_refinement.cpp
forrestglines: The `Xf<int,int>` template is so that we can use `Xf` for all references to locations, replacing the `x*f` and `x*s*` functions in Athena++.

That said, we're currently only using `Xf<x,x>` in Parthenon (and Athena++) although that will change soonish when we add the edge-corrections for face centered fields on AMR blocks. We can add an overloaded template `Xf<int>` when both the face and coordinate direction are the same.

For enforcing typing, we would have to change `CoordinateDirection` from an `enum` to `enum class` I believe. Hopefully that wouldn't break anything. Anywhere we really need to covert from an int to this enum would be slightly more awkward. I would be fine with that change as long as people are onboard with the code bloat (in terms of characters added) that would create.

src/coordinates/uniform_cartesian.hpp
forrestglines: Let's add in those overloaded operators with all three coordinate directions. You shouldn't have to change Phoebus for lack of those operators.

Originally we thought about only using the full `(k,j,i)` operators, perhaps enforcing those so that those non-orthogonal coordinates would work in the future. I agree that those operators are also safer and less error-prone. However, there was push back so I decided to hold off doing that for now. Since Parthenon (and AthenaPK) only uses the `(i)` convention, I didn't bother adding `(k,j,i)` operators. Since you're asking, I'll add them.

Are `(i)` and `(k,j,i)` versions sufficient? I found the `<class... Args>` versions obtuse to work with but I can try.

src/coordinates/uniform_cartesian.hpp
forrestglines: That's reasonable, I can make that change.

src/coordinates/uniform_cartesian.hpp
forrestglines: Hmm, is this `class... Args` causing the type deduction causing issues? Did it happen on multiple compilers?

forrestglines:
> I also notice a number of tests seem to be failing... possibly related to the template type deduction issue I noticed above.

What tests where are failing? In Parthenon or in Phoebus? The last merge introduced code using the old naming conventions so I need to fix those before it will compile.

Yurlungur:
I just meant tests in parthenon. Phoebus tests passed after I fixed things.

src/coordinates/uniform_cartesian.hpp
Yurlungur: `i` and `(k,j,i)` should be sufficient. :+1:

src/coordinates/uniform_cartesian.hpp
Yurlungur: I'm not sure... that's the likeliest culprit but I didn't see what was wrong. Only tried on gnu.

src/mesh/mesh_refinement.cpp
Yurlungur: I dunno if I want to move `CoordinateDirection` to an `enum class`... that's going to change everything everywhere.

c-prather: I definitely get why the <int, int> template is necessary.  I was proposing something like:
```cpp
template <int dirface, class... Args>
KOKKOS_FORCEINLINE_FUNCTION Real Dxf(Args... args) const {
  return Dxf<dirface, dirface>(args);
}
```

i.e., just overloading and calling through.  I misread the diff and thought that a similar overload was already implemented.  I *think* this would be simple enough to do in a general way, even if the specific code above doesn't work, but the difference is obviously minimal.  If this isn't uniformly agreed to be more readable it's not worth it.

Re: making directions (and faces) an enum class, I am abstractly very in favor of this.  Probably not in this PR though; it was just an idea for someday, and as Jonah says it will mean a lot of lines changed, even if there's always a clear way to translate the code.

src/coordinates/uniform_cartesian.hpp
Yurlungur: I don't think this is urgent to fix, by the way.

src/mesh/mesh_refinement.cpp
forrestglines: > ```c++
> template <int dirface, class... Args>
> KOKKOS_FORCEINLINE_FUNCTION Real Dxf(Args... args) const {
>   return Dxf<dirface, dirface>(args);
> }
> ```
I see, that should be implemented, I've now added it in. 


> I dunno if I want to move `CoordinateDirection` to an `enum class`... that's going to change everything everywhere.

That's mostly why I'm hesitant to make that change.  But with just the enum, we don't have any type checking, so using the enum everywhere at the moment won't add much. In the few places where `Xf<int,int>` will be used, it could be advantageous to use different enums to differentiate between coordinate direction and face direction.

> Re: making directions (and faces) an enum class, I am abstractly very in favor of this. Probably not in this PR though; it was just an idea for someday, and as Jonah says it will mean a lot of lines changed, even if there's always a clear way to translate the code.

Yep, let's save it for another PR

src/coordinates/uniform_cartesian.hpp
forrestglines: That change has been made

src/coordinates/uniform_cartesian.hpp
forrestglines: I've added that overload

forrestglines:
@jonahm-LANL This is ready for review again

src/interface/swarm_device_context.hpp
```diff
@@ -87,16 +87,18 @@ class SwarmDeviceContext {
...
+    j = (ndim_ > 1)
+            ? static_cast<int>(std::floor(
+                  (y - y_min_) / coords_.CellWidthFA(CoordinateDirection::X2DIR))) +
+                  jb_s_
+            : jb_s_;
+    k = (ndim_ > 2)
+            ? static_cast<int>(std::floor(
+                  (z - z_min_) / coords_.CellWidthFA(CoordinateDirection::X3DIR))) +
+                  kb_s_
+            : kb_s_;
```
Yurlungur: I've noticed lots of places where you've made this change: Dx -> CellWidthFA. I think the better replacement rule would be `Dx -> Dxc<dir>` as that's probably the intent in all these places and it'll be the faster option down the road. For examples it also offers more concise and parthenonic code.

src/outputs/vtk.cpp
```diff
@@ -146,10 +146,10 @@ void VTKOutput::WriteContainer(SimTime &tm, Mesh *pm, ParameterInput *pin, bool
     // write x1-coordinates as binary float in big endian order
     std::fprintf(pfile, "X_COORDINATES %d float\n", ncoord1);
     if (ncells1 == 1) {
-      data[0] = static_cast<float>(pmb->coords.x1v(out_ib.s));
+      data[0] = static_cast<float>(pmb->coords.Xc<1>(out_ib.s));
```
Yurlungur, line 149: Does the vtk output even work? Not for this MR but maybe we should either revive it for real or kill it.

Yurlungur (approved):
Sorry for the delay---been under the weather. LGTM. Two minor nitpicks below. Neither blocking.

src/interface/swarm_device_context.hpp
forrestglines: If I recall correctly, `Dx` was a run time determined function inherited from Athena++ and thus became `CellWidthFA` from the Athena++ comments on its behavior vs. `dx1v`. I think it might make a difference for some coordinate systems. Everywhere `Dx` was used in Parthenon, however, would be better served by `Dxc<dir>` so I'll make that change

src/coordinates/uniform_cartesian.hpp
forrestglines: This issue came up when I switched from function argument functions to template arguments in the Poisson example. 
```
/vast/home/forrestglines/code/parthenon-project/parthenon/example/poisson/poisson_package.cpp(233): error: expected an expression

/vast/home/forrestglines/code/parthenon-project/parthenon/example/poisson/poisson_package.cpp(162): error: expected an expression

2 errors detected in the compilation of "/vast/home/forrestglines/code/parthenon-project/parthenon/example/poisson/poisson_package.cpp".
```

Changing out `auto` fixed the compilation issue

Yurlungur:
@forrestglines can you share the most recent version of your script to find/replace all the instances?

pgrete:
The updated script in the comment above did the job for `parthenon-hydro`

forrestglines:
> 

The script in the comment should work. I forgot that I had an old script in the description of the PR. I've now updated that to match the new interface

Yurlungur:
Cool thanks, @forrestglines @pgrete

PR #740, opened 2022-10-05
Make python file loading faster

c-prather:
Opening this to trigger tests/float quickly. Very small change but useful for downstream codes using phdf.py.

## PR Summary

Replaces a python nested loop with vectorized/numpy operations to reduce phdf.py's file open time from ~2s to trivial.  Necessary in some downstream code for treating Parthenon output.

## PR Checklist

- [ ] Code passes cpplint
- [ ] New features are documented.
- [ ] Adds a test for any bugs fixed. Adds tests for new features.
- [ ] Code is formatted
- [ ] Changes are summarized in CHANGELOG.md
- [ ] CI has been triggered on [Darwin](https://re-git.lanl.gov/eap-oss/parthenon/-/pipelines) for performance regression tests.
- [ ] (@lanl.gov employees) Update copyright on changed files

PR #755, opened 2022-10-07
Text file of all parameters

c-prather:
## PR Summary

This PR adds a function to EvolutionDriver, DumpInputParameters, which writes a text file parthinput.archive (optionally with a timestamp on the end) containing the current state of the InputParameters object.

It also adds a call to this function just before driver execution, if the option parthenon/job/archive_parameters is specified.  Optionally the filename can include a timestamp, with parthenon/job/archive_parameters_timestamp.

Note this is only in EvolutionDriver: this shouldn't affect downstream codes but means that adding it to e.g. calculate_pi example will not work.

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted


Yurlungur:
Great thanks for this! Looks formatting is failing---can you run "make format" on it? Also we should add the changelog entry.

c-prather:
@par-hermes format

c-prather:
Looks like this is stuck on the formatter, with "src/driver/driver.cpp:15:  <chrono> is an unapproved C++11 header.  [build/c++11] [5]"

Not sure if I should switch to something else for timestamping, or if this is a matter of loosening/updating the style check's expectations?

tagardiner:
My guess is that this is failing because some of the functionality in std::chrono is C++17 and some is C++20.

Yurlungur:
The `std::chrono` issue is a [google-ism](https://stackoverflow.com/questions/33653326/google-style-guide-chrono-is-an-unapproved-c11-header) and a quirk of `cpplint`. I thought we had found a way to disable it in `cpplint` but apparently not. I'll see if I can find a workaround.

Yurlungur:
I just changed `cpplint` to allow it. We distribute our own copy of `cpplint` and this is not the first time we've had to remove a google-forbidden C++ header from the blacklist. It's basically always because the header doesn't interact well with chrome, and Google rolls their own.

scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
```diff
@@ -305,22 +305,18 @@ def GenAuxData(self):
...
+            (self.MeshBlockSize[0] * self.MeshBlockSize[1] * self.MeshBlockSize[2], 3),
+        )
+        self.isGhost = (
+            (zo[0] > self.BlockIdx[:, 0])
+            | (self.BlockIdx[:, 0] >= zo[1])
+            | (yo[0] > self.BlockIdx[:, 1])
+            | (self.BlockIdx[:, 1] >= yo[1])
+            | (xo[0] > self.BlockIdx[:, 2])
+            | (self.BlockIdx[:, 2] >= xo[1])
+        )
```
Yurlungur, line 319: Make sure the relevant PR for this this gets merged in before this PR.

Yurlungur (approved):
LGTM assuming tests pass. I don't think we need to trigger this on Darwin, it won't have any performance impact.

Yurlungur:
We need two approvals for a merge. Pinging @tagardiner @jdolence @brryan @forrestglines @pgrete to offer a second review. @tagardiner does this meet your needs?

tagardiner:
I just checked out the branch to build and test.  A quick look at the code and it looks good.  I'll get back to you after my meeting.

tagardiner:
I just built the code, ran the tests and manually ran the advection problem to generate the input file.
I turned on the time-stamping as well and it generated a file called

parthinput.archive.2022-10-10T17:31:37Z

So it has the date and GM time in the name.
What are the T and Z supposed to indicate in this name?

In general, I think this is great - thank you!

I wonder, though, if we can't get a bit more functionality out of this by either reducing this to a single parameter like

<parthenon/job>
archive_parameters = {true, timestamp, false}

(and anything not in this set is also considered false)

or, you could add the ability for the user to control the name of the archive input file. e.g.
archive_parameters = filename
archive_parameters_timestamp = {true, false}

Thoughts?

c-prather:
The T and Z are from the ISO formatter to [indicate UTC](https://en.wikipedia.org/wiki/ISO_8601#Coordinated_Universal_Time_(UTC)).

I'm in favor of the more adaptable single parameter (I'd love to use filenames in my own downstream convention, for example), but I'm not sure how Parthenon handles this.  I'd love to add a few more complex parameters like that myself, so if there's a quick example of adding/fetching a parameter like that from the pinput object I can mimic that.

tagardiner:
Ahh, so the wiki page gives this example

"14:45:15 UTC" would be "14:45:15Z" or "T144515Z".

I'm guessing you're keeping the T as a. separator between the date and time?

Re. more complex parameters, I think what you're looking for is to either use logical conditions like

  int DoesParameterExist(const std::string &block, const std::string &name);

which has the downside that if nothing happens when the parameter does not exist, then its existence is not recorded in the output parameter file, or the function

  std::string GetOrAddString(const std::string &block, const std::string &name,
                             const std::string &value);

that can be used to default the filename to something like "parthinput.archive" and optionally set to whatever string is in the parameter, e.g.

archive_parameters = Karma_input

Yurlungur:
I have an idea to make this more automated. Give me a moment and I'll push.

Yurlungur:
@tagardiner @bprather I added a new overload of `GetOrAddString` which can take allowed parameters. Looks something like this:
```C++
auto param = pin->GetOrAddString("parthenon/job", "archive_parameters", "false", std::vector<std::string>{"true", "false", "timestamp"});
```
and the input parser checks whether or not the value in the input deck is one of the allowed parameters.

The framework is more general---I added a correctness checker to the `ParameterInput` class that can check any type, but I only added the overload for strings for now.

tagardiner:
Ok, so assuming that I'm understanding the code properly, the code that Jonah just added simplifies this case

<parthenon/job>
archive_parameters = {true, timestamp, false}

with a default of false.

Ben, do you want the option to easily set the filename or anything else that we can get in this commit?  Do you want something like

archive_filename = filename

or what have you.  It seems to me that now's the time to massage this code to get everything you want out of it in the cleanest possible interface.  This interface currently meets my minimum requirements.  Adding the ability to set the filename as well would be great.

c-prather:
Hmm, I was misinterpreting: I was reading the set notation {} as a literal vector, instead of enumerating alternatives.  So, my comments about flexibility are with an eye toward better Parthenon-level handling of lists or vectors of values, which I need for other things.  That said, I will love being able to enumerate alternatives for an option: many fewer manual string checks or option typos.

I don't need a customizable filename to merge this, I'd probably only use it over my downstream version in the context of an effort to be more "Parthenonic" anyway.  Since no other filenames in Parthenon are customizable I don't think it's a bug not to offer the option here, though it's worth thinking about offering more filename customization in the context of the IO work.  It seems like an option here could fit into a later IO rework PR or similar.

tagardiner (approved):

tagardiner (approved):
I'm happy with these changes.

forrestglines (approved):
Looks good, I like this addition.

Yurlungur:
I'm not sure what the blocker is on this---I think CI was broken so tests were failing. Merged in develop and re-triggered tests. Hopefully that will get this to pull through.

pgrete:
Did you force merge or how did the merge go through with tests failing?

Yurlungur:
I did not force merge. Auto merge was on. Assumed tests needed to pass first before it would merge. 

Perhaps we need to disable auto merge if it doesn't wait for tests. Could also double check to see if there's a restriction on merges that got lost in the transfer to the new user group.

c-prather:
Oh no, this was a mess, my bad.  My apologies for the mixups, the dependency was an accident so I should at least have explicitly mentioned #740 in this commit.  Noted for future PRs.

pgrete:
I think I know what happened.
GitHub Action "status checks" are not automatically "mandatory" to pass, so once I removed the old GitLab ones, no tests were enforced again.
I now added the unit and short regressions tests as mandatory again, so that should prevent PRs with failed tests to automerge.

PR #757, opened 2022-10-07
Move to Flux Correction in One

lroberts36:
## PR Summary

Switch from per buffer loads and unloads of flux corrections to batched loads and unloads. This is in anticipation of generalizing the boundary values machinery to work with face, edge, and vertex centered fields. Also refactors bvals code since there was a lot of code duplication between the boundary communication routines and the flux correction in one routines.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Changes are summarized in CHANGELOG.md
- [ ] CI has been triggered on [Darwin](https://re-git.lanl.gov/eap-oss/parthenon/-/pipelines) for performance regression tests.
- [x] (@lanl.gov employees) Update copyright on changed files

Issue #736, opened 2022-10-04
Plan for face-centered fields


Yurlungur:
From Hackathon:
- Cell vars generalized to become differently shaped fields
- Generalize "WithFluxes" to be more verbose about what kind of fluxes
- "EMF correction" implemented as "line integral correction"
- Keep in mind concern about order of operations

TODOS:
- [x] Flux correction in one (Luke)
- [x] Remove all the old face-centered stuff leftover from athena++ (Forrest) #738 
- [x] Rename CC and Cell to Var (Josh) #867 
- [x] Implement newly shaped vars. (Josh) #868 
- [x] Add metadata. (Josh) #868 
- [x] Pull newly shaped vars through packs. Remove `GetDim` from packs (Luke, Deferred to pack discussion) #868 
- [x] Update/Fix single-level comms and fluxes (Philipp) #872 
- [x] Physical boundary conditions -> `MeshData` so prolongation/restriction can be done in-one
- [x] Pull Prolongation/Restriction hooks to userspace (Jonah) #769 
- [x] Pull prolongation/restriction through new vars (Pull from AthenaK) #872 
- [ ] Include face-centered fields, etc. in output
- [ ] Post-implementation cleanup
- [ ] Documentation

Yurlungur:
Additional note: AMR load balance is missing some logic for tensor-valued quantities to communicate tensor values accross mesh boundaries. We need to repeat what @brryan did for buffer packing.

lroberts36:
One of the TODOs is to "Add metadata", does anyone know what this refers to? It looks like metadata already exists for faces, edges, and nodes. Is is just referring to update metadata to change the shapes of variables based on their topological type?

pgrete:
Good question. Your first guess would have also been mine.

Yurlungur:
I'm not sure either. Let's just check it off lol.

PR #769, opened 2022-10-26
userspace prolongation

Yurlungur:
## PR Summary

This PR finally pulls the custom prolongation-restriction operators through the infrastructure. I'm still testing, but I think it's in a place now where it can at least be looked at for review. In particular I would really like @lroberts36 's eyes on it since it mucks about quite a bit with the boundary stuff.

As described in #682 , I moved to a functor-based approach for prolongation/restriction operations. In this PR, I add machinery to for a user register these functors (using closures based on `std::function`) on a per-variable basis. The registration is through the metadata, and looks something like this:
```C++
Metadata m;
m.RegisterRefinementOps<MyProlongationOp, MyRestrictionOp>();
```
Since the metadata is carried with the variable, the relevant ops can be pulled out on a per-variable basis as needed. For example:
```C++
  const auto &metadata = coarse->metadata();
  std::shared_ptr<MeshBlock> pmb = GetBlockPointer();
  const auto &refinement_funcs = metadata.GetRefinementFunctions();
  const auto &prolongator = refinement_funcs.prolongator_host;
  int b = 0;
  int nbuffers = 1;
  cell_centered_bvars::BufferCacheHost_t info_h("refinement info", nbuffers);
  refinement::loops::IdxHost_t idxs_h("host data", nbuffers);
  idxs_h(b) = b;
  // fill the relevant quantities in the info_h object
  prolongator(info_h, idxs_h, pmb->cellbounds, pmb->c_cellbounds, nbuffers);
```

Then if a variable is registered with this metadata, the relevant `std::function`s that close the prolongation/restriction ops are registered in a cache in the `StateDescriptor`, the authoritative cache being in `resolved_packages`. This cache can map prolongation/restriction operators to a unique id per op.

That registration part was the easy part. I also had to replace all the places in the code where prolongation/restriction are performed with calls to the registered functions. For the old per-meshblock calls, this is easy---see code above. But for the in-one methods, especially with sparse, this was challenging. 

The approach I came up with (with @lroberts36 's help) was to add a 2D matrix to each subcache of `BndInfo_t` arrays (as introduced in #757 ). The rows of this matrix are matched to prolongation/restriction operators, and the columns to the subsets of the `BndInfo_t` arrays which need to be operated on by a particular operator. Essentially this matrix acts as a list of sebsets of the total sets of `BndInfo_t` objects. Then the infrastructure can loop over each type of operation and perform the relevant prolongation/restriction ops. And the total array is also retained for things like in-one communication in the optimal number of kernel calls.

In the process, I also:
- unified the old `RestrictPhysicalBounds` functionality with @lroberts36 's new `bvals` looping and caching machinery. 
- removed the `Slice` stuff from `ParArrayNDGeneric` as it's unused, and replaced it with a `Kokkos::subview` overload consistent with the other overloads.

This PR depends on #757 and #732 so those need to be merged first.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


Yurlungur:
No new regression tests are necessary as userspace features are in unit tests, and the new machinery is threaded through the infrastructure so automatically tested in regression tests.

Yurlungur:
I didn't add documentation yet. I will.

Yurlungur:
Docs now added.

Yurlungur:
This is a significant change, so at least on the API side, it would be good for others to have eyes as well. Especially those looking towards face-centered fields.

Yurlungur:
> This is a significant change, so at least on the API side, it would be good for others to have eyes as well. Especially those looking towards face-centered fields.

@pgrete @jdolence @forrestglines @bprather @brryan

src/bvals/cc/bnd_info.cpp
```diff
@@ -296,7 +349,8 @@ BndInfo BndInfo::GetSendBndInfo(std::shared_ptr<MeshBlock> pmb, const NeighborBl
 
 BndInfo BndInfo::GetSetBndInfo(std::shared_ptr<MeshBlock> pmb, const NeighborBlock &nb,
                                std::shared_ptr<CellVariable<Real>> v,
-                               CommBuffer<buf_pool_t<Real>::owner_t> *buf) {
+                               CommBuffer<buf_pool_t<Real>::owner_t> *buf,
+                               const OffsetIndices &of) {
```
lroberts36: Is `of` used anywhere? Should it be an unnamed argument? See also `GetSendCCFluxCor` and `GetSendBndInfo`. Is this only included so that `GetSetBndInfo` has the same call signature as `GetCCRestrictInfo`?

src/bvals/cc/bnd_info.cpp
```diff
@@ -236,6 +236,58 @@ void CalcIndicesLoadToFiner(int &si, int &ei, int &sj, int &ej, int &sk, int &ek
...
+      rs = b.s - 1;
+      re = b.s - 1;
+    }
+  };
+
+  CalcIndices(ris, rie, ni, nb.ni.ox1, cib);
+  CalcIndices(rjs, rje, nj, nb.ni.ox2, cjb);
+  CalcIndices(rks, rke, nk, nb.ni.ox3, ckb);
+}
+
```
lroberts36: It is not quickly obvious to me exactly what is going on in this index calculation. I know it goes against the code style guidelines ;), but it would be nice to have a brief comment about what the various quantities are (i.e. `rie`).

src/bvals/cc/bnd_info.hpp
```diff
@@ -61,18 +72,32 @@ struct BndInfo {
...
+  // These are are used to generate the BndInfo struct for various
+  // kinds of boundary types and operations.
+  // TODO(JMM): As we move to non CC-centered, this may need a
+  // refactor or this section of the struct might explode in
+  // complexity.
   static BndInfo GetSendBndInfo(std::shared_ptr<MeshBlock> pmb, const NeighborBlock &nb,
                                 std::shared_ptr<CellVariable<Real>> v,
-                                CommBuffer<buf_pool_t<Real>::owner_t> *buf);
+                                CommBuffer<buf_pool_t<Real>::owner_t> *buf,
+                                const OffsetIndices &);
```
lroberts36: I think the boundary communication interface shouldn't have to change when we extend to different types of topological elements. `v` knows what type of variable it is, so we can check inside the factory method about how the index ranges need to be changed.

src/bvals/cc/bvals_utils.hpp
```diff
@@ -96,8 +97,33 @@ inline void ForEachBoundary(std::shared_ptr<MeshData<Real>> &md, F func) {
...
+                  if (ntype == 0 ||
+                      pmb->pbval->nblevel[nk + 1][nj + 1][ni + 1] != pmb->loc.level)
+                    continue;
+                  OffsetIndices offsets(nk, nj, ni);
+                  if (func_caller(func, pmb, rc, nb, v, offsets) ==
+                      LoopControl::break_out)
+                    return;
+                }
+              }
+            }
```
lroberts36, line 122: I am not sure if this makes sense to have in the `ForEachBoundary` logic. It appears that this can introduce up to nine visits to the function for a single boundary element, which means that the function is not really being called for each boundary but possibly multiple times for each boundary. I would advocate for having this internal looping be done explicitly inside the kernel or create a new `ForEachBoundaryAndOffset` method. 

I am also not a fan of having to pass around the `offsets` everywhere even though it isn't used often.

src/bvals/cc/bvals_utils.hpp
```diff
@@ -244,25 +270,92 @@ inline auto CheckReceiveBufferCacheForRebuild(std::shared_ptr<MeshData<Real>> md
...
+      }
+    } else {
+      rebuild = true;
+    }
+    ++nbound;
+  });
+  rebuild = rebuild || (nbound == 0) || (nbound != cache.idx_vec.size());
+  return std::make_tuple(rebuild, nbound);
+}
+
```
lroberts36, line 298: As an example, doesn't this possibly create multiple `BndInfo` objects for a a single boundary when `BOUND_TYPE = restricted`?

src/interface/state_descriptor.hpp
```diff
@@ -83,6 +70,41 @@ struct VarIDHasher {
...
+
+  std::size_t size() const noexcept { return next_refinement_id_; }
+  // A unique enumeration of refinement functions starting from zero.
+  // This is used for caching which prolongation/restriction operator
+  // matches which BndInfo struct in the buffer packing caches.
+  // the other relevant information is in metadata, so this is all we
+  // need.
+  std::unordered_map<refinement::RefinementFunctions_t, std::size_t,
+                     refinement::RefinementFunctionsHasher>
+      funcs_to_ids;
```
lroberts36, line 102: Does this imply that a registered refinement operation may show up more than once in the `funcs_to_ids` list (since it is the prologation/restriction pair that is used for the hash)?

Yurlungur:
Depending on how the discussion for #732 shakes out, I can remove it from this PR. The relevant changes are minor.

Yurlungur:
This machinery is required for face-centered fields. See #736

src/mesh/refinement_in_one.hpp
```diff
@@ -0,0 +1,155 @@
...
+    RefinementFunctions_t funcs(label);
+    funcs.restrictor = [](const cell_centered_bvars::BufferCache_t &info,
+                          const cell_centered_bvars::BufferCacheHost_t &info_h,
+                          const loops::Idx_t &idxs, const loops::IdxHost_t &idxs_h,
+                          const IndexShape &cellbnds, const IndexShape &c_cellbnds,
+                          const std::size_t nbuffers) {
+      loops::DoProlongationRestrictionOp<RestrictionOp>(
+          cellbnds, info, info_h, idxs, idxs_h, cellbnds, c_cellbnds,
+          RefinementOp_t::Restriction, nbuffers);
+    };
```
lroberts36: This is a nice way to store the functors.

src/utils/communication_buffer.hpp
```diff
@@ -39,7 +39,15 @@ enum class BufferState { stale, sending, sending_null, received, received_null }
 
 enum class BuffCommType { sender, receiver, both, sparse_receiver };
 
-enum class BoundaryType : int { local, nonlocal, any, flxcor_send, flxcor_recv };
+constexpr int NUM_BNDRY_TYPES = 6;
```
lroberts36, line 42: Probably needed to be done

lroberts36 (commented):
This looks reasonable to me, but I think I need to understand the changes to `ForEachBoundary` a little better before approval.

src/bvals/cc/bnd_info.cpp
Yurlungur: >  Should it be an unnamed argument?

Yes it can be unnamed.

> Is this only included so that GetSetBndInfo has the same call signature as GetCCRestrictInfo?

Yep

src/bvals/cc/bnd_info.cpp
Yurlungur: I copy pasted this out of the old restriction code. I don't fully understand it either, but I will try to disentangle it and add a comment.

src/bvals/cc/bnd_info.hpp
Yurlungur: Hmm, not sure I follow. My thought was we'll need a `GetSendBndInfo` function for each of the topological elements. Is that not the case? I can remove the comment as it's not relevant to the PR anway.

src/bvals/cc/bvals_utils.hpp
Yurlungur: Yes but I think that's by design. See above.

src/bvals/cc/bvals_utils.hpp
Yurlungur: > It appears that this can introduce up to nine visits to the function for a single boundary element, which means that the function is not really being called for each boundary but possibly multiple times for each boundary. 

It introduces up to eight visits... the 0,0,0 val is never called. But I *think* this is by design. The sub-pieces of each boundary halo are accessed. This is different than tying explicitly to buffers. That said, I have to admit, this was copied out of the original Athena++ code and I don't fully understand it.

> I would advocate for having this internal looping be done explicitly inside the kernel

I first wanted to put this loop in the kernel, but I can't actually because of the `continue`.

> or create a new ForEachBoundaryAndOffset method.

A new `ForEachBoundaryAndOffset` function would work, but it'll end up resulting in more repeated code. There would need to be a new `RebuildBufferCache` function for example. If that's the path we want to go down, I can be convinced, but we'd basically be copy-pasting two identical functions. I agree including the `offsets` in all the other functions is gross.

I guess in the longer term, rewriting the indexing logic to be less opaque might also resolve the issue, but I'd rather punt on that.

src/interface/state_descriptor.hpp
Yurlungur: Yes---the same restriction op might show up multiple times if it's coupled with a different prolongation function. We could resolve that by splitting prolongation/restriction pairs into separate objects. But I'm not sure how much sense that makes physically. IMO your prolongation and restriction should both be coupled to some assumption about discretization. Maybe I'm wrong about that though.

I'm also not sure it matters too much if that ends up happening anyway.

src/bvals/cc/bnd_info.cpp
Yurlungur: Ok I stared at this some and I think I have a better understanding, and I added some comments. The reason to restrict the ghost halos independent of communication is to ensure the coarse buffer has sufficient information for *prolongation* to the fine buffer. This is needed for physical boundaries at the very least, and possibly other things as well.

So this method finds the exact pieces of the coarse buffer, either on the `interior` or `boundary` of the meshblock that enable prolongation to the fine buffer as needed. That's why the weird logic and also why the offsets. Essentially the extra loop over offsets in `ForEachBoundary` is looking at the neighbor the buffer corresponds to and the regions *around* it to see what needs to be restricted to enable prolongation.

This is relevant to the discussion below on the `ForEachBoundary` loop as well, since it is a slightly different motivation for the logic. It pushes me more in the direction of a new loop, but I still don't love the additional code that that introduces.

src/bvals/cc/bnd_info.cpp
Yurlungur: Made it an unnamed argument

src/bvals/cc/bvals_utils.hpp
Yurlungur: > It introduces up to eight visits... the 0,0,0 val is never called. But I think this is by design. The sub-pieces of each boundary halo are accessed. This is different than tying explicitly to buffers. That said, I have to admit, this was copied out of the original Athena++ code and I don't fully understand it.

Confirmed this is by design. What's happening here is the restriction isn't over the same shape as the boundaries for ghost halo exchange. It's over whatever indices in the coarse buffer are required to provide enough information for prolongation, which may contain pieces of the interior and also contains things like edges and corners of the meshblock.

I don't know what that implies about what we should choose for the `ForEachBoundary` function. I guess we need to decide: Is `ForEachBoundary` only for comms, or is this infrastructure more general?

src/bvals/cc/bnd_info.hpp
Yurlungur: Removed comment.

src/bvals/cc/bnd_info.cpp
Yurlungur: > So this method finds the exact pieces of the coarse buffer, either on the interior or boundary of the meshblock that enable prolongation to the fine buffer as needed.

I suppose a solution would be to just waste cycles and restrict the whole meshblock. It wouldn't be *that* much more expensive lol.

src/bvals/cc/bnd_info.hpp
lroberts36: There will be different bounds for each topological element, but those changes can just be dealt with internally in `GetSetBndInfo` I think (unless I am missing some dependency).

Yurlungur: Ah I see. Yes that's possibly true. (Though then that function might start to get gross...) In any case we can see what this needs to look like when we actually get there.

docs/boundary_communication.md
```diff
@@ -128,4 +165,14 @@ The flux correction routines mirror the boundary routines, except that they do n
 - **`StartReceiveFluxCorrections(std::shared_ptr<MeshData<Real>>&)`**
 - **`LoadAndSendFluxCorrections(std::shared_ptr<MeshData<Real>>&)`**
 - **`ReceiveFluxCorrections(std::shared_ptr<MeshData<Real>>&)`**
-- **`SetFluxCorrections(std::shared_ptr<MeshData<Real>>&)`**
\ No newline at end of file
+- **`SetFluxCorrections(std::shared_ptr<MeshData<Real>>&)`**
+
+### Non-communication tasks
+
+These tasks use the same infrastructure as the boundary communication machinery but do not communicate
```
c-prather:
```suggestion
These tasks use the same infrastructure as the boundary communication machinery but do not communicate.
```

docs/interface/RefinementOperations.md
```diff
@@ -0,0 +1,104 @@
...
+refinement regions must be communicated via prolongation and
+restriction. The default operations used for cell-centered variables
+are averaging weighted by cell volume for restriction and linear
+interpolation with minmod limiting for prolongation.
+
+## User-Defined Operations
+
+A user may define their own prolongation and restriction
+operations. To do so, you must define a struct, templated on
+dimension, with a void method `Do` with tthe following signature:
```
c-prather:
```suggestion
dimension, containing only a void function `Do` with the following signature:
```

docs/interface/RefinementOperations.md
```diff
@@ -0,0 +1,104 @@
...
+KOKKOS_FORCEINLINE_FUNCTION static void
+Do(const int l, const int m, const int n, const int ck, const int cj, const int ci,
+   const IndexRange &ckb, const IndexRange &cjb, const IndexRange &cib,
+   const IndexRange &kb, const IndexRange &jb, const IndexRange &ib,
+   const Coordinates_t &coords, const Coordinates_t &coarse_coords,
+   const ParArray6D<Real> *pcoarse, const ParArray6D<Real> *pfine)
+```
+
+where `l`, `m`, `n` are the indices of a variable object not tied to
+mesh (for example the tensor indices of a rank 3 tensor). `ck`, `cj`,
```
c-prather:
```suggestion
mesh (for example the tensor indices of a rank 3 tensor, or 0,0,1 for a scalar). `ck`, `cj`,
```

docs/interface/RefinementOperations.md
```diff
@@ -0,0 +1,104 @@
...
+and `ci` are `k`, `j`, and `i` indices of the cell on the coarse
+buffer. `ckb`, `cjb`, `cib` are the `k`, `j`, and `i` indexrange
+bounds for the coarse buffer. `kb`, `jb`, and `ib` are the same but on
+teh fine buffer. `coords` and `coarse_coords` are the coordinates
```
c-prather:
```suggestion
the fine buffer. `coords` and `coarse_coords` are the coordinates
```

c-prather (approved):
I think we should have a long issue/discussion about the many places we can/should re-use this pattern.  The ability to have user-registered device-side functions as callbacks attached to individual variables is huge, I can see this being applied (often more simply!) to:
- reconstruction. Parthenon should have an "apply X reconstruction" pattern, so it can enforce that the last index run fastest regardless of reconstruction direction.
- floors, which can often be registered per-variable safely
- boundaries, which should be a per-variable host-side callback and use the same registration convention
- reductions, which generally need to apply a local function over a domain and apply some operation over the results, leading to a ton of boilerplate in the current host-side registration paradigm
- etc, etc.  I can see adding a similar pattern in places to my own code too.

Parthenon is a framework, and we now have a way to apply that framework to saving duplicated code both on the host side and device side.  We should take advantage.

I think Parthenon passes around a silly number of bare indices, but this is not the place to fix that gripe.

src/bvals/cc/bvals_utils.hpp
lroberts36: After our discussion yesterday, I think we should leave this as is for this PR and think about simplifying things in a future PR. I am fairly convinced that this is more complicated than it needs to be (and that the ghost restriction step just needs to be done over some subset of the topological elements of a block). Approving the PR.

lroberts36 (approved):
LGTM

src/bvals/cc/bvals_utils.hpp
Yurlungur: :+1: Sounds good. I agree and I am looking at simplifying this in a future PR.

docs/interface/RefinementOperations.md
Yurlungur: Should be
```suggestion
mesh (for example the tensor indices of a rank 3 tensor, or 0,0,0 for a scalar). `ck`, `cj`,
```

Yurlungur:
Thanks, @bprather ! I agree we should look to doing this elsewhere in the code. For boundary conditions I've been meaning to refactor things for a while.

I want to wait until someone from AthenaPK has a chance to review before merge.

Yurlungur:
@bprather points out that the prolongation/restriction operators do not have information about what variables they are, or have access to other variables that have been registered. For things like "primitive prolongation" or "upwinding" this may need re-visiting. 

That said that would be a deeper change, either requiring boundary packing to work on meshblockpacks explicitly or requiring `ParArrayNDGeneric` to carry around additional information. I suggest that this change be revisited as needed in a separate PR.

c-prather:
I'm actually realizing I probably don't need that info just for prolongation and restriction: see my comment on #779.  Definitely happy merging this as-is and dealing with that problem (if indeed it is a problem) when I have an implementation to point to.

That said, some work addressing variable (and potentially package) dependency & grouping tooling would be great, and something I might look into.  I'm thinking especially of boundaries and floors, which often need predictable groups of variables.

Yurlungur:
> I'm actually realizing I probably don't need that info just for prolongation and restriction: see my comment on #779. Definitely happy merging this as-is and dealing with that problem (if indeed it is a problem) when I have an implementation to point to.
> 
> That said, some work addressing variable (and potentially package) dependency & grouping tooling would be great, and something I might look into. I'm thinking especially of boundaries and floors, which often need predictable groups of variables.

I have some stuff in mind for retooling boundaries. Let me know if you want to chat about that at some point. It would be great to have your help on that.

pgrete:
I now had a first look at the PR and tried it downstream, see https://github.com/parthenon-hpc-lab/athenapk/compare/e79f7e7...pgrete/prim-prolong

I really like the design and how enrolling functors works.

The big/main question I have right now (before giving it a more detailed review) is: How would I enroll prolong/restrict ops that depend on the state of multiple vars (in an elegant way)?
Specifically, if I'd want to do primitive reconstruction I need the info of more that one variable for the conversion.
According to my current understanding of the code,  depending on the impl. of variables (large single vector or independent fields or a mix), I can easily access other variables (through `n`) for a large state vector, but would recalculate the coarse values multiple times (for different `k,j,i`  which have overlapping origin regions) or need to get very creative to access other "fields".

c-prather:
Heh, this is exactly what Jonah and I were talking about above, and would take some extra work/thinking.  To be clear, primitive prolongation/restriction would necessarily involve U->P->prolongate->P->U all in the prolongation/restriction kernel?

Jonah had an idea for this I didn't fully grok, but my instinct would be "hand the user the whole enchilada, along with a way to map variables->indices on the device side."  Basically, don't pretend each variable is its own universe, just pass the pointer to the world and trust that the user will be responsible with it.  The same idea could apply to a bunch of kernels -- boundaries, etc, and avoids registering a bunch of dependencies.  There's still some bookkeeping though, as you might want to call 5 kernels to update 5 fields, or one kernel which has to update all of them together to be remotely efficient.  Registering "which fields does this kernel update" is still a problem, even if we don't explicitly require a registration for "which fields does this kernel need."

Any good solution would I think need to provide semantic info on the device, since the user does not directly control which fields are in each prolongation/restriction bundle, but still needs to know which is which.  And, conveying that sounds like it's predicated on forcing users to adopt the "variables are types" work from @lroberts36 (which I also don't fully grok) -- I'm not against that at all, but it's another consideration.

Yurlungur:
> How would I enroll prolong/restrict ops that depend on the state of multiple vars (in an elegant way)?

So as of now, there are only two ways to do this:
1. "break model" and make primitives 1st-class variables and communicate them..
2. Wrap all variables needed for con2prim in a single vector.

I may be missing something, but I don't think either of the above approaches recalculates things unnecessarily in the ghost zones but I could be wrong about that.

The current prolongation/restriction infrastructure, which we inherited from Athena++ treats each variable as a universe, as @bprather said. I have a few ideas for how to extend this, but they all imply development work, and are maybe kind of gross:
1. Rewrite all of our infrastructure to prolongate/restrict and communicate *variable packs* instead of individual `ParArrayND`s. This is pretty deep surgery, but perhaps the most elegant solution.
2. Pass in a meshblock or variable pack along with the relevant arrays per buffer. Gives the array *world* access. Shallowest surgery, but IMO the grossest solution.
3. Register "relationships" between variables, which are somehow stored in the extra metadata now carried around by `ParArrayNDGeneric` (as implemented by @lroberts36 ). There's kind of something in the public-facing API for this already, through the `associated` metadata thing. I can imagine this being somewhere between the above two.
4. Live with what we have.

I feel like this deserves a longer discussion as it's a pretty deep design decision and introduces additional complexity into the infrastructure. I also would strongly argue it should be a separate PR from what we do above. That said, the path we go down may imply small changes in this PR. Thoughts?

lroberts36:
Was just writing this up when @Yurlungur commented. Essentially, I would suggest something half way along the lines of his suggestion 1.

I think we would eventually like to move to doing boundary communication/prolongation/restriction all "in-one". Therefore, if we want to be able to perform restriction operations that rely on multiple variables, those extra fields probably need to be accessible from the `BndInfo` object. One possibility would be to have a way of registering fields that the prolongation/restriction operator depends on in a fields `Metadata`. Then, when the prolongation and restriction `BndInfo` is created it would include a variable pack within the `BndInfo` object that contains these other fields. This pack could then be passed to the prolongation and restriction operators as well as the variable that is actually getting filled. There is of course the issue of how to the downstream user accesses different fields in the pack, as @bprather points out, but I think this could be overcome with the type based indexing mechanism (but this would require jumping through some hoops to store a type list in the `Metadata`). The upside to this approach is that it is not too invasive to the code but is fairly flexible, but the downside is that there would still be per field prolongation/restriction which might introduce repeated calculations.

Yurlungur:
I like the idea of adding the dependencies to the `BndInfo` struct. (Though I would still say that's a big change and should be a separate PR.)

lroberts36:
Yes, I think this is something to be done once we have completely integrated the current changes to prolongation and restriction (as well as the cleanup). Probably also after we have set up prolongation in-one.

lroberts36:
Maybe it actually makes more sense to have a set of auxiliary fields associated with the prolongation operator itself. If we wrote our own type erasure object like `std::function` but that also had a method to return a list of field names to hold the prolongation and restriction operators, then I think this would be pretty straightforward to build packs in `BndInfo` and then access them on device in the prolongation/restriction kernels via types.

lroberts36:
I just wanted to test how this might work, it actually seems pretty straightforward. 
```c++
// Like prolongation_restriction_op
struct functor {
  using my_pack_t = SparsePack<v3, v5>; 
  KOKKOS_INLINE_FUNCTION
  static auto GetPack(const SparsePackBase& base_pack) {return my_pack_t(base_pack);}
  template <class T>
  static auto GetPack(T *pmd) {return my_pack_t::Get(pmd);}
  
  KOKKOS_INLINE_FUNCTION
  void kernel(my_pack_t& pack, int b, int k, int j, int i) { 
    pack(b, v3(), k, j, i) = pack(b, v5(), k, j, i);
  }
};

//Like prolongation_restriction_loop
template <class F> 
void Loop(const SparsePackBase& base_pack, int NBlocks, IndexRange kb, IndexRange jb, IndexRange ib) { 
  const int scratch_level = 1;
  size_t scratch_size_in_bytes = 1;
  par_for_outer(
      DEFAULT_OUTER_LOOP_PATTERN, "Test",
      DevExecSpace(), scratch_size_in_bytes, scratch_level, 0, NBlocks - 1,
      KOKKOS_LAMBDA(parthenon::team_mbr_t team_member, const int b) {
        // Create a type based pack from the base pack on device
        auto dev_pack = F::GetPack(base_pack);
        F f;  
        par_for_inner(parthenon::inner_loop_pattern_ttr_tag, team_member, 
                      kb.s, kb.e, jb.s, jb.e, ib.s, ib.e,
                        [&](const int k, const int j, const int i) { 
            f.kernel(dev_pack, b, k, j, i); 
          });  
      });
}

...

// Pull out a pack that can be stored (for instance in BndInfo)
SparsePackBase sparse_pack_base = functor::GetPack(&mesh_data);
...

// Call loop function
Looper<functor>(sparse_pack_base, kb, jb, ib);
```

Yurlungur:
@pgrete ping. I'd like to get this in.

Yurlungur:
Changes to develop by #780 broke this PR. Now fixed.

Yurlungur:
I triggered the long CI, but it doesn't seem to be registering here. Is there some special way I need to trigger?

pgrete:
> I triggered the long CI, but it doesn't seem to be registering here. Is there some special way I need to trigger?

Yes, it'll automatically trigger once "auto-merge" is enabled. Only then it's reported back.
I haven't found workaround to report pack for manual triggered pipelines, see https://github.com/parthenon-hpc-lab/parthenon/pull/786#issuecomment-1316698266

Yurlungur:
> > I triggered the long CI, but it doesn't seem to be registering here. Is there some special way I need to trigger?
> 
> Yes, it'll automatically trigger once "auto-merge" is enabled. Only then it's reported back. I haven't found workaround to report pack for manual triggered pipelines, see [#786 (comment)](https://github.com/parthenon-hpc-lab/parthenon/pull/786#issuecomment-1316698266)

Ah ok. Works for me. :+1:

example/stochastic_subgrid/stochastic_subgrid_driver.cpp
```diff
@@ -161,8 +161,7 @@ TaskCollection StochasticSubgridDriver::MakeTaskCollection(BlockList_t &blocks,
           tl.AddTask(update, parthenon::cell_centered_bvars::ReceiveBoundBufs<any>, mc1);
       auto set = tl.AddTask(recv, parthenon::cell_centered_bvars::SetBounds<any>, mc1);
       if (pmesh->multilevel) {
-        tl.AddTask(set, parthenon::cell_centered_refinement::RestrictPhysicalBounds,
-                   mc1.get());
+        tl.AddTask(set, parthenon::cell_centered_bvars::RestrictGhostHalos, mc1, false);
```
pgrete, line 164: Is my understanding correct, that we now restrict all ghost halos whereas before, we only restricted the ghost cells with (physical/external) boundary conditions (or was the original naming somewhat misleading)?

src/bvals/bvals_refine.cpp
```diff
@@ -140,10 +113,9 @@ void BoundaryValues::ProlongateGhostCells_(const NeighborBlock &nb, int si, int
 
   for (auto cc_var : pmr->pvars_cc_) {
     if (!cc_var->IsAllocated()) continue;
-    ParArrayND<Real> var_cc = cc_var->data;
-    ParArrayND<Real> coarse_cc = cc_var->coarse_s;
-    int nu = var_cc.GetDim(4) - 1;
-    pmr->ProlongateCellCenteredValues(coarse_cc, var_cc, 0, nu, si, ei, sj, ej, sk, ek);
+    int nu = cc_var->GetDim(4) - 1;
+    pmr->ProlongateCellCenteredValues(cc_var.get(), cc_var.get(), 0, nu, si, ei, sj, ej,
+                                      sk, ek);
```
pgrete: If I haven't missed anything then `ProlongateCellCenteredValues` is not always called with the same `cc_var` for the coarse and fine variable.
Is there still a reason to pass the variable in twice or can we just parse it once (as the access to `.data()` and `.coarse_s` is already handled inside that function now.

src/bvals/cc/bnd_info.cpp
```diff
@@ -236,6 +236,75 @@ void CalcIndicesLoadToFiner(int &si, int &ei, int &sj, int &ej, int &sk, int &ek
...
+  } else {
+    getbounds(nb.ni.ox2, nj);
+  }
+
+  if (pmb->block_size.nx3 == 1) {
+    nk.s = nk.e = 0;
+  } else {
+    getbounds(nb.ni.ox3, nk);
+  }
+}
```
pgrete, line 258: Duplicated code already present in `src/bvals/bvals_refine.cpp`

src/bvals/cc/bnd_info.cpp
```diff
@@ -236,6 +236,75 @@ void CalcIndicesLoadToFiner(int &si, int &ei, int &sj, int &ej, int &sk, int &ek
...
+        rs = b.e;
+      } else if (ox == -1) {
+        re = b.s;
+      }
+    } else if (n == 1) { // need to fill "edges" or "corners" on this axis
+      rs = b.e + 1;      // TODO(JMM): Is this always true?
+      re = b.e + 1;      // should this end at b.e + NG - 1?
+    } else {             //(n ==  - 1)
+      rs = b.s - 1;      // TODO(JMM): should this start at b.s - NG + 1?
+      re = b.s - 1;      // or something similar?
```
pgrete, line 299: Are those "todos" resolved or is there still a question on those indices?

src/bvals/cc/bnd_info.cpp
```diff
@@ -490,5 +563,29 @@ BndInfo BndInfo::GetSetCCFluxCor(std::shared_ptr<MeshBlock> pmb, const NeighborB
...
+                                   CommBuffer<buf_pool_t<Real>::owner_t> *buf,
+                                   const OffsetIndices &no) {
+  BndInfo out;
+  if (!v->IsAllocated()) {
+    out.allocated = false;
+    return out;
+  }
+  out.allocated = true;
+  CalcIndicesRestrict(no.nk, no.nj, no.ni, out.si, out.ei, out.sj, out.ej, out.sk, out.ek,
+                      nb, pmb);
```
pgrete, line 579: at some point I'd really like to switch to a fixed "index register" that holds all (precomputed) indices for all combinations so that we'd just need to pass in a single index in our BndInfo object.

src/bvals/cc/bnd_info.hpp
```diff
@@ -95,20 +117,31 @@ struct BvarsSubCache_t {
...
+
+  // Can be used to inform the infrastructure to loop over only a
+  // subset of the bvars cache. Used for prolongation/restriction.
+  std::vector<std::size_t> buffer_subset_sizes;
+  ParArray2D<std::size_t> buffer_subsets{};
+  ParArray2D<std::size_t>::host_mirror_type buffer_subsets_h{};
 };
 
 struct BvarsCache_t {
   // The five here corresponds to the current size of the BoundaryType enum
```
pgrete:
```suggestion
```

src/bvals/cc/bvals_utils.hpp
```diff
@@ -96,8 +97,33 @@ inline void ForEachBoundary(std::shared_ptr<MeshData<Real>> &md, F func) {
...
+            ComputeRestrictionBounds(bni, bnj, bnk, nb, pmb);
+            // This loop is only over {-1, 0, 1}^3 at most
+            for (int nk = bnk.s; nk <= bnk.e; ++nk) {
+              for (int nj = bnj.s; nj <= bnj.e; ++nj) {
+                for (int ni = bni.s; ni <= bni.e; ++ni) {
+                  int ntype = std::abs(ni) + std::abs(nj) + std::abs(nk);
+                  // skip myself or coarse levels; only the same level must be restricted
+                  if (ntype == 0 ||
+                      pmb->pbval->nblevel[nk + 1][nj + 1][ni + 1] != pmb->loc.level)
+                    continue;
```
pgrete, line 115: This was probably done before, but conceptually why is "same level" restriction required but not coarse/fine?

src/bvals/cc/bvals_utils.hpp
```diff
@@ -145,17 +171,17 @@ void InitializeBufferCache(std::shared_ptr<MeshData<Real>> &md, COMM_MAP *comm_m
...
-        ++boundary_idx;
-      });
+  ForEachBoundary<bound_type>(md, [&](sp_mb_t pmb, sp_mbd_t rc, nb_t &nb, const sp_cv_t v,
+                                      const OffsetIndices &) {
+    auto key = KeyFunc(pmb, nb, v);
+    PARTHENON_DEBUG_REQUIRE(comm_map->count(key) > 0,
+                            "Boundary communicator does not exist");
+    // Create a unique index by combining receiver gid (second element of the key
+    // tuple) and geometric element index (fourth element of the key tuple)
+    int recvr_idx = 27 * std::get<1>(key) + std::get<3>(key);
```
pgrete, line 181: Just double checking (not related to this PR), did we update the entire machinery to only work with 27 buffer indices (compared to the max  of 56 in Athena++)?

src/bvals/cc/bvals_utils.hpp
```diff
@@ -244,25 +270,92 @@ inline auto CheckReceiveBufferCacheForRebuild(std::shared_ptr<MeshData<Real>> md
...
+  Mesh *pmesh = md->GetParentPointer();
+  StateDescriptor *pkg = (pmesh->resolved_packages).get();
+  if constexpr (!((BOUND_TYPE == BoundaryType::flxcor_send) ||
+                  (BOUND_TYPE == BoundaryType::flxcor_recv))) {
+    int nref_funcs = pkg->NumRefinementFuncs();
+    // Note that assignment of Kokkos views resets them, but
+    // buffer_subset_sizes is a std::vector. It must be cleared, then
+    // re-filled.
+    cache.buffer_subset_sizes.clear();
+    cache.buffer_subset_sizes.resize(nref_funcs, 0);
```
pgrete, line 324: If we know `nbounds` here, why is it init to `0`?

src/parthenon_array_generic.hpp
```diff
@@ -238,27 +238,6 @@ class ParArrayGeneric : public State {
...
-    return SliceD(std::make_index_sequence<Data::rank - N>{},
-                  std::make_index_sequence<N - 1>{}, slc);
-  }
-
-  // translates into auto dest = src.SliceD<dim>(indx,nvar)
-  template <std::size_t N = Data::rank>
-  auto SliceD(const int indx, const int nvar) {
-    return SliceD<N>(std::make_pair(indx, indx + nvar));
-  }
-
```
pgrete, line 261: I think that removal is not reflected in the `parthenon_arrays.md` docs yet.

src/parthenon_array_generic.hpp
```diff
@@ -339,6 +312,14 @@ inline bool UseSameResource(const PA &pa1, const PA &pa2) {
...
 namespace Kokkos {
 
+// JMM: for some reason this works better than Slice. And it seems
+// like we're doing evil Kokkos namespace overloads anyway so...
+template <class U, class SU, typename... Args>
+inline auto subview(const parthenon::ParArrayGeneric<U, SU> &arr, Args... args) {
+  auto v = Kokkos::subview(static_cast<U>(arr), std::forward<Args>(args)...);
+  return parthenon::ParArrayGeneric<decltype(v), SU>(v, arr);
+}
+
```
pgrete, line 322: Is this actually used anywhere right now?
Otherwise, we could also just skip this and encourage using `Kokkos::subview` directly (as already done in various places of the code).

tst/unit/test_state_descriptor.cpp
```diff
@@ -318,6 +351,31 @@ TEST_CASE("Test dependency resolution in StateDescriptor", "[StateDescriptor]")
         }
       }
     }
+
+    WHEN("We register a dense variable withuot custom") {
```
pgrete:
```suggestion
    WHEN("We register a dense variable custom prolongation/restriction") {
```

pgrete (commented):
Overall looks great and thanks for the extra doc!
I only have some minor questions/comments and then we can merge.

example/stochastic_subgrid/stochastic_subgrid_driver.cpp
Yurlungur: I believe the original naming was misleading---and it was my fault---I discovered the need to do the restriction because of physical boundaries. However, we need to restrict all ghost halos for the ghosts at coarse-fine boundaries to be correct.

src/bvals/cc/bnd_info.cpp
Yurlungur: I would too. I think this machinery still needs a re-write but I would say that should be "future work"

lroberts36: I agree that the index calculation and `BndInfo` needs to be cleaned up and streamlined, but I am not sure that an "index register" is the way to go. We would have to enumerate all types of index sets in one place, which might tie pieces of code together more than we would like. Of course I could be convinced otherwise if the repeated index calculations matter to performance.

src/bvals/cc/bvals_utils.hpp
lroberts36: At least the send and receive keys only use 27 indices for denoting the connection between two blocks. The other information is unnecessary, I think.

src/parthenon_array_generic.hpp
Yurlungur: It is used---in the updated prolongation/restriction logic, as well as in `refinement-in-one`. You just can't tell what is what because of the evil overloading

src/parthenon_array_generic.hpp
Yurlungur: Good catch. :+1:

src/bvals/bvals_refine.cpp
Yurlungur: Yeah I think we can pass it in only once. I considered leaving it split because I was thinking maybe we'd want coarse and fine in different stages or something. But for now that's probably overkill.

src/bvals/cc/bnd_info.cpp
Yurlungur: Right you are. Meant to move this and forgot to delete the older code in `bvals_refine.cpp` will fix.

src/bvals/cc/bvals_utils.hpp
Yurlungur: So this is the issue related to prolongating in ghost zones. Basically, at coarse fine boundaries, the block needs data in the ghost zones of the coarse buffer not only coming from the coarse neighbor, but from neighboring blocks on the same level.

There are cleanups of this discussed in the call and in my work queue... I didn't expect review of this PR to take a month or I would have started working on them earlier.

src/bvals/cc/bvals_utils.hpp
Yurlungur: Pretty sure we still use the 56 total buffers don't we?

src/bvals/cc/bvals_utils.hpp
Yurlungur: We know `nbound` but we do not know the size of each "subset" of buffers. The subsets are the buffers/variables associated with different prolongation/restriction operators. So we use this for indexing in the reconstruction algorithm a few lines down.

Yurlungur:
Thanks all for the review! I think I've addressed all concerns so I'm going to click "auto merge" now.

Yurlungur:
I'm seeing some bizarre errors on the CI that don't look like code issues but CI configuration:
https://github.com/parthenon-hpc-lab/parthenon/actions/runs/3609553944/jobs/6083261172

src/bvals/cc/bvals_utils.hpp
lroberts36: yes, a single block can have up to 56 total buffers it writes to and up to 56 other buffers it reads from. But the maximum number of geometric elements that two blocks can share is 27 (which would correspond to a single block, unigrid setup, with periodic boundary conditions in every direction). As a result, `(gid1, gid2, geometric_element_index)` uniquely defines which one of the 56 a particular buffer is (with `geometric_element_index` running between zero and 27).

PR #771, opened 2022-10-27
WIP: Fix single precision configuration

pgrete:
## PR Summary
In theory we already supported single precision.
In practice, it didn't work (not even compilation).

Todos:
- [ ] Add test coverage for single prec
- [ ] Check that all time/timing related parameters are `Real64`
- Check that all reductions, reduce to `Real64`
  - [ ] kernels
  - [ ] MPI
- [ ] check use of `Real` override in unit tests
- Discuss
  - [ ] (fixed) type for particles positions
  - [ ] (fixed) type for coordinates

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted


pgrete:
@brryan I encountered some interesting behavior for particles when I tried to compile in single precision.
Most of the position ("x", "y", "z") related function were hard-coded to `double`, which I understand.
However, the `View`s themselves were `Real` resulting in a mismatch when trying to get the data because the map would return `Real` where `double`s are expected in further processing.
Hardcoding the positions to `double`/`Real64` could be an option, but it would require to add just another map next to `int` and `Real`.
What's your take on this?

Yurlungur:
@pgrete @brryan what's the status here?

pgrete:
Probably still not working and needs fixing.

PR #772, opened 2022-10-28
Trigger short CI only for PRs and remove old SpaceInstances test

### LINKED ISSUE #770: Badges need an update?
Despite the CI on develop currently passing, the badges on the README indicate failing CI. Did something go wrong with them when we migrated to the new project space?

pgrete:
## PR Summary

Implemented changes as discussed during the call yesterday.
Fixes #770 

Also 
- fixed a hostcomm buffer logic on HIP that resulted in that feature silently falling back to device memory.
- started to use more abstract `KOKKOS_IF_ON_HOST` macros rather than separate checks for Cuda, Hip, ...

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/src/kokkos_abstraction.hpp
+++ b/src/kokkos_abstraction.hpp
@@ -48,8 +48,14 @@ using HostExecSpace = Kokkos::DefaultHostExecutionSpace;
 using LayoutWrapper = Kokkos::LayoutRight;
 using MemUnmanaged = Kokkos::MemoryTraits<Kokkos::Unmanaged>;
 
-#if defined(KOKKOS_ENABLE_CUDA) && defined(PARTHENON_ENABLE_HOST_COMM_BUFFERS)
+#if defined(PARTHENON_ENABLE_HOST_COMM_BUFFERS)
+#if defined(KOKKOS_ENABLE_CUDA)
 using BufMemSpace = Kokkos::CudaHostPinnedSpace::memory_space;
+#elif defined(KOKKOS_ENABLE_HIP)
+using BufMemSpace = Kokkos::Experimental::HipHostPinnedSpace::memory_space;
+#else
+#error "Unknow comm buffer space for chose execution space."
+#endif
 #else
 using BufMemSpace = Kokkos::DefaultExecutionSpace::memory_space;
 #endif
--- a/.github/workflows/ci-extended.yml
+++ b/.github/workflows/ci-extended.yml
@@ -25,7 +25,7 @@ jobs:
       matrix:
         device: ['cuda', 'host']
         parallel: ['serial', 'mpi']
-    runs-on: self-hosted
+    runs-on: [self-hosted, A100]
     container:
       image: registry.gitlab.com/pgrete/parthenon/cuda11.6-mpi-hdf5
       # map to local user id on CI  machine to allow writing to build cache
--- a/.github/workflows/ci-short.yml
+++ b/.github/workflows/ci-short.yml
@@ -1,6 +1,6 @@
 name: CI short
 
-on: [push, pull_request]
+on: [pull_request]
 
 # Cancel "duplicated" workflows triggered by pushes to internal
 # branches with associated PRs.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@
 - [[PR 663]](https://github.com/lanl/parthenon/pull/663) Change bvals_in_one to use sparse boundary buffers and add flux_correction in one.
 
 ### Fixed (not changing behavior/API/variables/...)
-- [[PR 743]](https://github.com/lanl/parthenon/pull/740) Faster PHDF file load times in phdf.py
+- [[PR 740]](https://github.com/lanl/parthenon/pull/740) Faster PHDF file load times in phdf.py
 - [[PR 751]](https://github.com/lanl/parthenon/pull/751) Delete useless file in advection example
 - [[PR 765]](https://github.com/lanl/parthenon/pull/765) Fix incorrect BC labeling in swarm
 - [[PR 759]](https://github.com/lanl/parthenon/pull/759) Add metadata so Visit treats outputs as time series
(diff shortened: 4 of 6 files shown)


Yurlungur (approved):
LGTM

Yurlungur:
Looks like one style issue:
```bash
 tst/unit/kokkos_abstraction.cpp:27:  Include "kokkos_abstraction.hpp" not in alphabetical order  [build/include_alpha] [4]
```

Yurlungur:
that should fix it.

Yurlungur:
uh actually we have a problem. format and lint disagree about the ordering of `kokkos_abstraction` vs `kokkos_macros`. Not sure what to do about that.

src/kokkos_abstraction.hpp
```diff
@@ -48,8 +48,14 @@ using HostExecSpace = Kokkos::DefaultHostExecutionSpace;
...
 using MemUnmanaged = Kokkos::MemoryTraits<Kokkos::Unmanaged>;
 
-#if defined(KOKKOS_ENABLE_CUDA) && defined(PARTHENON_ENABLE_HOST_COMM_BUFFERS)
+#if defined(PARTHENON_ENABLE_HOST_COMM_BUFFERS)
+#if defined(KOKKOS_ENABLE_CUDA)
 using BufMemSpace = Kokkos::CudaHostPinnedSpace::memory_space;
+#elif defined(KOKKOS_ENABLE_HIP)
+using BufMemSpace = Kokkos::Experimental::HipHostPinnedSpace::memory_space;
+#else
+#error "Unknow comm buffer space for chose execution space."
```
c-prather, line 57: The way I read this, it throws an error compiling for OpenMP with HOST_COMM_BUFFERS enabled.  Is there a reason to bother erroring rather than just setting BufMemSpace to the default?

In particular, when I tested it this seems to stop compilation of the header but not always dependencies, so it actually cascades into a bunch of unrelated errors.  If we really want to disallow OpenMP with HOST_COMM we should do that in the CMakeLists maybe.

c-prather (changes requested):
The actual changes look great to me, just the one comment re: errors.

src/kokkos_abstraction.hpp
pgrete: I tried to keep the logic simple and make it crash hard for unintended/unsupported cases.
For example, I'm not sure about the status of the memory space equivalent for the SYCL backend.
And for OpenMP it would depend on whether this is OpenMP or OpenMPTarget with the former not having any meaning and the latter I'm not sure about (again if it's supported etc).
As far as my testing goes "hostpinnedmemory" should only be used as an exception or for testing because in practice it was always slower.
That's why I didn't set a default but rather make it crash.
You prefer a cmake solution?

pgrete:
> uh actually we have a problem. format and lint disagree about the ordering of `kokkos_abstraction` vs `kokkos_macros`. Not sure what to do about that.

The macros should be available through the core header anyway so I just removed the extra header include.

src/kokkos_abstraction.hpp
Yurlungur: > Is there a reason to bother erroring rather than just setting BufMemSpace to the default?

I am not in favor of `HOST_COMM_BUFFERS` quietly doing nothing if you set it with an openmp backend. On top of that, is it always true that openmp backend will do nothing? If we ever try to use openmp target offload, is there a host-pinned equivalent?

> You prefer a cmake solution?

Even if we go for a cmake solution, I would not remove an error message from `kokkos_abstraction`. We could make the error more informative for cases where incompatible macros have been chosen though? That could be done with the preprocessor or cmake... or with `static_assert(false, "message")`

pgrete:
Hmm, I don't understand why the code is called even though it's in a ON_HOST macro

src/kokkos_abstraction.hpp
c-prather: I'm not that concerned about how exactly things fail when trying to compile or SYCL or another unsupported case, I think anyone attempting those things knows what they're doing enough that specific errors don't matter.

Instead, I'm imagining the case where someone is looking down the Parthenon options wishing to compile for OpenMP/CPU, finds a "HOST_BUFFERS" option without any other warnings or prefixes, and turns it on assuming that buffers, like all other computations in this case, should be on the host.  Or, someone repurposes a file for compiling against their desktop GPU one day to compile for CPU, and assumes that they can just change the Kokkos_ENABLE_X options.  Flooding the scrollback with errors seems hostile in these cases, vs. just putting the buffers on the host.  I'm also just generally not a fan of producing errors at compile time when they could have been caught on configure.

If it's important to keep the error in unsupported contexts (can we just check for those in the CMakeLists too?), we should document that the option is for use only in limited contexts:
```cmake
option(PARTHENON_ENABLE_HOST_COMM_BUFFERS "CUDA/HIP Only: Allocate communication buffers on host (slow)" OFF)
```

We can even just check it against the relevant flags:
```cmake
if (PARTHENON_ENABLE_HOST_COMM_BUFFERS)
  if (NOT Kokkos_ENABLE_CUDA AND NOT Kokkos_ENABLE_HIP)
    message(FATAL_ERROR "Host pinned buffers for MPI communication are supported only for CUDA and HIP backends.)
  endif()
endif()
```

Yurlungur: Yeah I'm fine with an informative error message in `cmake`, if that reduces the C++ error overload. My point is just that we *should* have an error in the preprocessor for any cases we missed. Just to be a bit more defensive.

pgrete: I like both suggested solutions and will update accordingly.

PS: As Jonah mentioned, so far we've typically chosen the "fail loud" approach for anything that we could reasonable catch as unsupported and untested. Too many got burned in the past by a code that silently ran in that regime.

c-prather: I agree with leaving the #error (with a couple more 'n's), if it has a guard in the CMake file; my issue was that without a clear error message it could easily become a sharp edge for new users, but I think a configure-time error makes it pretty foolproof.

Separately, though, it seems obvious to me that in the context of compiling for OpenMP on the host, asking for HOST_COMM_BUFFERS should be a no-op, because all the buffers are on the host anyway.  If others don't agree then we can keep it totally explicit, but "are we compiling for CPU" seems like something simple enough to determine during the configuration step (or which we should be determining anyway...)

Yurlungur: To me at least it's not obvious that `HOST_COMM_BUFFERS` should be a no-op with `OpenMP` on host. I'd rather the `cmake` explain what's going on rather than let that quietly do nothing.

c-prather: Huh, okay, thanks.  Will update my prior then.  Resolving this, to expedite the merge

c-prather (approved):

pgrete:
Okay, I think I fixed it. As @ctrott was pointing out in the Kokkos channel, `IF_ON_HOST` is more like `CUDA_ARCH` rather than checking if sth if device backends are disabled.
That's why I updated the checks to `if constexpr` on the default execution spaces.
If there are not further comments, I'll merge tomorrow.

Yurlungur (approved):
LGTM :+1:

PR #774, opened 2022-10-31
Add second order refinement criteria

### LINKED ISSUE #763: Second Order Refinement Error Criterion
From what I can see, the default Error Criterion is order 1 difference.

In my experience, a normalized second order error criterion has very good properties for picking out sharp features, so I'd like to implement it as an optional criterion.  I'll add details on the proposed enhancement in a subsequent update.

tagardiner:
The is related to issue #763 

## PR Summary

The code contains a bit of boilerplate, but it should work.  I implemented a second order refinement criterion.  The next step is to test it out with some parthenon example problem.  I think including it as an optional run for the advection problem would be a good place to start.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


tagardiner:
@par-hermes format

src/refinement/amr_criteria.cpp
```diff
@@ -65,4 +67,39 @@ AmrTag AMRFirstDerivative::operator()(const MeshBlockData<Real> *rc) const {
   return Refinement::FirstDerivative(pmb.get(), q, refine_criteria, derefine_criteria);
 }
 
+
+AMRSecondDerivative::AMRSecondDerivative(ParameterInput *pin, std::string &block_name) {
+  field = pin->GetOrAddString(block_name, "field", "NO FIELD WAS SET");
+  if (field == "NO FIELD WAS SET") {
+    std::cerr << "Error in " << block_name << ": no field set" << std::endl;
+    exit(1);
+  }
```
Yurlungur: Ideally we should do this with our `PARTHENON_ERROR` machinery. (It's possible the old AMRFirstDerivative was written before that machinery). We need
```C++
#include "utils/error_checking"
```
and then we can do
```C++
PARTHENON_REQUIRE_THROWS(pin->DoesParameterExist(block_name, "field"), "Error in " + block_name + ": no field set.");
field = pin->GetString(block_name, "field");
```

src/refinement/amr_criteria.cpp
```diff
@@ -65,4 +67,39 @@ AmrTag AMRFirstDerivative::operator()(const MeshBlockData<Real> *rc) const {
...
+  refine_criteria = pin->GetOrAddReal(block_name, "refine_tol", 0.5);
+  derefine_criteria = pin->GetOrAddReal(block_name, "derefine_tol", 0.05);
+  int global_max_level = pin->GetOrAddInteger("parthenon/mesh", "numlevel", 1);
+  max_level = pin->GetOrAddInteger(block_name, "max_level", global_max_level);
+  if (max_level > global_max_level) {
```
Yurlungur: Not sure if we do this in the first derivative code, but might be good to hide this in a check for MPI rank zero
```suggestion
  if ((max_level > global_max_level) && (Globals::my_rank == 0)) {
```
assumes that you have
```C++
#include "globals.hpp"
```
at the top of the file.

Yurlungur (changes requested):
This looks like a great addition! Thanks for adding this. I just have a few minor nitpicks below. Also:
- Looks like we're missing a changelog: Just add the PR to `CHANGELOG.md`.
- And looks like formatting isn't passing at least when I looked. 

For getting formatting working, the cmake should let you just type `make format` and then `make lint`. If that doesn't work, I can also give you a script (which I should just commit to the repo) which can format the code when cmake is being finicky.

Yurlungur:
> And looks like formatting isn't passing at least when I looked.

Ah you asked the bot to do it. Very good. :)

src/refinement/amr_criteria.cpp
tagardiner: Ok, this is easy enough - done.

src/refinement/amr_criteria.cpp
tagardiner: This too is easy enough, but honestly I'm not clear why this code has its own idea of max_level that could be <= global_max_level.

Is the idea here that you can ask for AMR refine / derefine calculations at levels less than global_max_level?  What happens then?  That is, if one does this, how do we ever end up with a level > max_level?

tagardiner:
cpp-py-formatter

src/refinement/amr_criteria.cpp
Yurlungur: I'm actually not sure---this may simply be extraneous code which should be removed.

Yurlungur: ^which should be removed at some point. No reason you have to do it now.

Yurlungur (approved):

src/refinement/amr_criteria.cpp
tagardiner: Looking at this a bit more closely, it sounds like you're suggesting that the error message written to stderr should be only done by rank 0.  But, constraining max_level <= global_max_level should be done by all ranks, right?  I'm assuming here that there will NOT be an additional synchronization step later.

tagardiner:
@par-hermes format

src/refinement/amr_criteria.cpp
Yurlungur: Ah, yeah that's right. I just want one error message, but the constraint should be global.

Yurlungur:
Looks like one of the CI tests is failing. I restarted it since I think it might be a transient error.

tagardiner:
@pgrete @jdolence I think this second order refinement criterion is ready to go.
Will you take a look and suggest changes or approve?  Thanks

Yurlungur (approved):
:+1: Thanks for fixing the first derivative code while you were at it.

src/refinement/amr_criteria.cpp
pgrete: There are use cases for which it may be desired that some refinement crit only trigger down to a certain level, which is why there's an extra check.
For example (not yet relevant for any downstream code right now as far as I'm aware of but who knows), in Enzo the gravity refinement based on particles is typically limited to several levels above the maximum level (for very deep hierarchies) so that the potential is effectively smoothed (and particles don't start acting as "black holes" on the finest levels).

src/refinement/refinement.cpp
```diff
@@ -130,6 +130,50 @@ AmrTag FirstDerivative(MeshBlock *pmb, const ParArrayND<Real> &q,
...
+    ku = dim3 - 2;
+  }
+  if (dim2 > 1) {
+    jl = 1;
+    ju = dim2 - 2;
+  }
+  if (dim1 > 1) {
+    il = 1;
+    iu = dim1 - 2;
+  }
```
pgrete: I know we're doing this for first order deriv, too, but what was actually the use case to check for refinement in the ghost zones?

docs/amr.md
```diff
@@ -30,7 +30,10 @@ In each refinement block, you are required to provide a ``method`` which is a st
 ### Predefined Criteria
 | Method | Description |
 |--------|-------------|
-| derivative_order_1 | ![formula](https://render.githubusercontent.com/render/math?math=\|dlnq\/dlnx\|), where q is the user selected variable |
+| derivative_order_1 | ![formula](https://render.githubusercontent.com/render/math?math=\|dlnq\/dlnx\|) |
+| derivative_order_2 | $$\frac{(\delta x_i)^2}{\|q\|} \left\| \frac{\partial^2 q}{\partial x_i^2} \right\|$$ Note that, as programmed, this quantity is bounded by [0,2). |
```
pgrete: I think there are some `| |` missing around the second deriv term.
Also, it'd be great if you could add a pointer to some doc (or document directly), why the the implementation differs from the definition here.
There are probably good reasons for it (and done in other codes) so this may help user to pick the right values based on the reference (e.g., the implementation in Pluto of the "second deriv" criterion is quite different from the one used here).

tagardiner: @pgrete The absolute value signs around the second derivative are present in the doc.  If you look at the file diff. page you can select a button to show the rich text preview of the doc.

The coordinate index subscript is not intended to implicitly mean to sum on the repeated index, rather each coordinate direction is tested independently.  Admittedly, this code assumes Cartesian coordinates, though that assumption is also present in the first derivative refinement criterion.

Honestly, I don't know what's done in Pluto - I've never looked at the code.  The denominator is a measure of the conserved variable, it's

$$  \frac{|q_i| + 0.5 |q_{i-1} + q_{i+1}| }{2} $$

The reason I use this for normalizing the refinement criterion can be seen by considering limiting cases where say $q_i \le 0$ and $(q_{i-1} + q_{i+1}) > 0$.  This often happens for say the momentum or magnetic field components.  Normalizing the second derivative in this way results in a mathematically bounded expression between $[0, 2]$.  In the implementation, I followed the convention used in the first derivative refinement criterion to include a tiny floating point value which formally changes the bounding to $[0,2)$.  This is noted in the documentation.

Are you asking for something more formal as to why this expression is used, like a description of these details in the doc file?

src/refinement/refinement.cpp
tagardiner: @pgrete I have no idea.  I don't know how many ghost zones are used in Parthenon or how to know if I'm looking at zone, face, edge, ... centered data.  I simply followed the pattern used in the first derivative refinement criterion assuming that it tested the appropriate domain.

Please feel free to update both order 1 and 2 to use the appropriate domain.

docs/amr.md
tagardiner: @pgrete I looked over the code again and it looks like to be consistent with the documentation I need an additional factor of 1/2.  Would you like me to add the factor of 1/2 to the documentation or write out explicitly the case for say the x coordinate so that it's clear.  In the end, I'm recommending second derivative error measure that's proportional to this

$$  \frac{ \| q_{i-1} + q_{i+1} - 2 q_i \| }{ \| q_{i-1} + q_{i+1} \| + 2 \| q_i \|} $$

In this form it's bounded by $[0,1]$.

tagardiner: I revised the code and documentation to match this error criterion.

src/refinement/refinement.cpp
pgrete: The number of ghost zones is an input parameter.
I'll put the question on the agenda for the call tomorrow so that we either fix this or add a comment why this is done.
I'm thinking about downstream development here, where a new developer tries to implement a new criterion and is wondering the same thing.

docs/amr.md
pgrete: > The absolute value signs around the second derivative are present in the doc. If you look at the file diff. page you can select a button to show the rich text preview of the doc.

I used the rich text version and it doesn't seem to properly render on my end:
image
So that's probably the way it is.

Regarding the doc, it'd be great if you could add that you use a "weighted q" in the denominator (for reasons) just to make it transparent on how to get from the analytic to the discrete formula.
However, as this is just nitpicking I'm already approving so don't consider this blocking my approval.

pgrete (approved):

Yurlungur:
@pgrete @tagardiner should we merge or is this waiting on anything?

docs/amr.md
tagardiner: How odd - here's what I see
doc

src/refinement/refinement.cpp
tagardiner: Sounds good - I won't be there tomorrow, so just post the decision or feel free to make the changes.

docs/amr.md
tagardiner: On the subject of the docs, it's probably also worth fixing the first order refinement criterion description.  From what I see, the code is not actually calculating

$$ \left \| \frac{d \ln q}{d \ln x} \right \|$$

tagardiner:
@Yurlungur I think the final remaining question pertains to the loop bounds and whether the order 1 and order 2 refinement criteria ought to be checking ghost zones or just computational zones.  I believe @pgrete said that he'll bring it up for discussion tomorrow.

I won't be able to join the discussion, so consider my input to be to go with the group consensus.

jdolence:
We didn't get to the discussion of tagging in ghost zones on today's call, but here's my 2 cents in preparation for the discussion next week.  I _think_ what we want is usually to tag on all the interior zones plus one layer of ghost zones.  My argument is basically that this helps to refine a block just before a feature propagates into it, as opposed to having something propagate across a refinement boundary and then afterward refine the block.  This is really driven by thinking about hyperbolic type problems.

With a tiny refactor I've put into #780, it would actually be really easy to have the number of ghost zones used in tagging set at runtime, with perhaps a default of one being reasonable.  Perhaps something like that is the right answer here.

pgrete:
Does anyone require this feature until next week? Otherwise, I suggest to wait for the discussion next week, then update docs and code and then merge (rather than merging now and "fixing" things next week).

pgrete:
I was just about to add the comment we discussed during the last sync and then merge, only to realize that the place for the comment is gone since in #780 @jdolence :
> AMR criteria are always evaluated on interior cells and one layer of ghost cells, independent of the number of ghost cells available. The previous behavior tagged on nghost - 1 ghost cells, which I do not believe is the desired behavior. Discussed in Add second order refinement criteria #774 also.

So this needs to be "unfixed" as discussed during the sync (i.e., we always want to check the maximum number of cells allowed by the stencil inside the kernel).
@tagardiner do you want to take a shot at this or should I volunteer myself?

tagardiner:
@pgrete I need to rewrite the second order refinement indicator due to the extensive changes made there.  Now that #791 is merged, I can move forward and rewrite it.  If you want to add a comment and change the inclusion of ghost cells, why don't you start a new PR and just get it done and I'll add a second order criterion after that.

pgrete:
Didn't you already rewrite it? It looks like it's already adapted to the new interface?

tagardiner:
@pgrete Look at that - I did rewrite it...  I should confess however that I haven't tested it in parthenon yet.  I just merged develop into this branch, so it should be good to go from that standpoint.

tagardiner:
@jdolence I'm trying to run the burgers test problem with first and second order refinement criteria to see that it works properly.  To interpret the results, I'm trying to visualize the output in VisIt and it's crashing whenever I try to make a pseudocolor plot.  What variable should I look at?  Thanks!

I should add, the error message is:

HDF5-DIAG: Error detected in HDF5 (1.8.14) thread 0:
  #000: ../../src/H5Shyper.c line 6606 in H5Sselect_hyperslab(): invalid stride==0 value
    major: Invalid arguments to routine
    minor: Bad value

tagardiner:
@pgrete @jdolence I'm unable to visualize any data from the advection problem as well.  Opening the xdmf files for the advection problem also results in this error when trying to generate a pseudocolor plot of the "advected" quantity.

HDF5-DIAG: Error detected in HDF5 (1.8.14) thread 0:
  #000: ../../src/H5Shyper.c line 6606 in H5Sselect_hyperslab(): invalid stride==0 value
    major: Invalid arguments to routine
    minor: Bad value

I suspect that this has something to do with the recent changes to HDF5 output.  Thoughts?

brryan:
@tagardiner I bet you are right about the (my) recent output change breaking XDMF. Sorry about that, I'm taking a look at this now.

tagardiner:
@brryan Thanks!

tagardiner:
@pgrete I revised the advection problem to initialize the "sphere" to have a value of 1 and the material outside of that to have a value of 0.5, turned off the existing refinement (by setting the refinement tolerance high) and added a second order refinement criterion to the input file.  The odd thing is that I'm seeing the code refines reasonably, but it doesn't derefine.  I should mention that I set derefine_count=2 so it should be happening every other time-step.  Any thoughts?

The thing that would really help is if we could output the refinement error to see what the values are.  Is this possible in parthenon?

tagardiner:
As far as I can tell, this refinement condition is working, so I'm pushing the auto-merge button.

Yurlungur:
> The thing that would really help is if we could output the refinement error to see what the values are. Is this possible in parthenon?

This isn't currently possible but it wouldn't be too hard to add I think

PR #780, opened 2022-11-02
Adds a new solver for Burgers' equation for benchmarking

jdolence:
## PR Summary

This PR introduces a new example-like code for solving Burgers' equation.  This is motivated by a desire to have a physics-free, Parthenon-based benchmark that reflects some of the features of more sophisticated physics codes.  Because of the desire for this to serve as a benchmark instead of an example, more effort was put into having the implementation reflect real usage instead of providing demonstrations of a wide array of Parthenon features.

The benchmark includes a README.md that provides documentation on the equations solved and the available options.

Along the way to developing this, a few new features were developed that are worth highlighting:
- `ParameterInput` objects now have a generic `GetVector<T>` call that returns a `std::vector<T>` built from a comma separated list in the input file.  We already had this kind of functionality to parse the list of variables for output, but this makes it generally available for all types and for more broad use.
- `AMRCriteria` can now specify the tensor/vector indices of a variable to be used as a tagging criteria.
- AMR criteria are always evaluated on interior cells and one layer of ghost cells, independent of the number of ghost cells available.  The previous behavior tagged on nghost - 1 ghost cells, which I do not believe is the desired behavior.  Discussed in #774 also.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


src/refinement/amr_criteria.cpp
```diff
@@ -37,6 +37,9 @@ AMRFirstDerivative::AMRFirstDerivative(ParameterInput *pin, std::string &block_n
     std::cerr << "Error in " << block_name << ": no field set" << std::endl;
     exit(1);
   }
+  comp6 = pin->GetOrAddInteger(block_name, "comp6", 0);
+  comp5 = pin->GetOrAddInteger(block_name, "comp5", 0);
+  comp4 = pin->GetOrAddInteger(block_name, "comp4", 0);
```
Yurlungur: I think these names are a little opaque. What about `tensor_component_{6,5,4}`?

src/refinement/amr_criteria.cpp
```diff
@@ -55,14 +58,14 @@ AMRFirstDerivative::AMRFirstDerivative(ParameterInput *pin, std::string &block_n
...
   if (!rc->HasCellVariable(field) || !rc->IsAllocated(field)) {
     return AmrTag::same;
   }
 
-  q = rc->Get(field).data;
+  auto q = rc->Get(field).data.Slice(comp6, comp5, comp4, Kokkos::ALL(), Kokkos::ALL(),
+                                     Kokkos::ALL());
   std::shared_ptr<MeshBlock> pmb = rc->GetBlockPointer();
-  return Refinement::FirstDerivative(pmb.get(), q, refine_criteria, derefine_criteria);
+  return Refinement::FirstDerivative(q, refine_criteria, derefine_criteria);
```
Yurlungur: Clever! And also supports how, e.g., `AthenaPK` does things I assume

Yurlungur (commented):
I like this! We may also want to point people to it, along with `parthenon-hydro` as a place to look for a modern "correct style" for a parthenon-based code.

src/refinement/amr_criteria.cpp
jdolence: Yeah, that's right.  This easily supports the state vector approach, which makes the parthenon provided criteria a lot more usable.

src/refinement/amr_criteria.cpp
jdolence: updated this -- take a look at it now and see what you think.

Yurlungur: I like it. :+1:

c-prather (approved):
This is great on a number of levels; as an example of modern Parthenon usage, the fast nested parallelism construct, and fast WENO5 reconstruction, and as an "example downstream" that will need to be updated if/as we change Parthenon's interface.  I think it's well worth any maintenance burden to keep in the repo proper, and presumably can be used as a realistic integrated performance regression test or similar in the future, too.

If this will be a common first package/problem, it might be good to reduce the default size slightly to fit on a laptop better, and to produce diagnostic outputs every step rather than every 10.  Both make it less useful as a completely standard benchmark though...
Ideally there should maybe also be a way to quell the "More computing power than you expected..." warning to prevent people thinking any small-meshblock problem is cluster-sized, but I think people will understand what's going on there soon enough anyway.

src/refinement/amr_criteria.cpp
```diff
@@ -22,21 +22,32 @@
...
+    PARTHENON_REQUIRE_THROWS(index.size() == 2,
+                             "tensor_ij requires two values, e.g. tensor_ij = 2, 1");
+    comp5 = index[0];
+    comp4 = index[1];
+  } else if (pin->DoesParameterExist(block_name, "vector_i")) {
+    auto index = pin->GetVector<int>(block_name, "vector_i");
+    PARTHENON_REQUIRE_THROWS(index.size() == 1,
+                             "vector_i requires one value, e.g. vector_i = 2");
+    comp4 = index[0];
+  }
```
Yurlungur, line 50: This is a nice improvement.

benchmarks/burgers/recon.hpp
```diff
@@ -0,0 +1,100 @@
...
+
+  Real dq = q3 - q2;
+  dq = 0.5 * mc(q2 - q1, dq);
+
+  const Real alpha_lin = 2.0 * alpha_l * alpha_r / (alpha_l + alpha_r);
+  ql = alpha_lin * ql + (1.0 - alpha_lin) * (q2 + dq);
+  qr = alpha_lin * qr + (1.0 - alpha_lin) * (q2 - dq);
+}
+
+} // namespace recon
```
Yurlungur, line 98: Doesn't have to happen now... but should these move into the main parthenon recon file?

Yurlungur (approved):

PR #781, opened 2022-11-02
Output variables with same shape as in parthenon

brryan:
## PR Summary

Currently all parthenon variables are output in HDF5 as `[nblock, nvar, nk, nj, ni]`. This PR changes this for dumps ~~(not restarts)~~ so that variables are output in the same shape as they exist in parthenon variables. For example, in the HDF5 file a cell-centered tensor will have shape `[nblock, nu, nv, nk, nj, ni]`, a cell-centered scalar will have shape `[nblock, nk, nj, ni]`, and a non-mesh-tied variable with metadata `shape = {a, b, c}` will have shape `[nblock, c, b, a]`. 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Get sign-off from a quorum of the various downstream code representatives
- [x] Fix `phdf.py` for the case `flatten = True`
- [x] Ensure that v2 restarts with scalar fields work with v3 parthenon

### DIFF
--- a/scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
+++ b/scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
@@ -489,23 +489,73 @@ class phdf:
         vShape = self.varData[variable].shape
         if flatten:
             # if variable is not a scalar flatten cells component-wise
-            if np.prod(vShape) > self.TotalCells:
-                # TODO(tbd) remove legacy mode in next major rel.
-                if self.OutputFormatVersion == -1:
+            print('sdifj')
+            print(vShape)
+            print(self.TotalCells)
+            # TODO(tbd) remove legacy mode in next major rel.
+            if self.OutputFormatVersion == -1:
+                if np.prod(vShape) > self.TotalCells:
                     return self.varData[variable][:].reshape(
                         self.TotalCells, vShape[-1]
                     )
+                else:
+                    return self.varData[variable][:].reshape(self.TotalCells)
 
-                ret = np.empty(
-                    (vShape[1], self.TotalCells), dtype=self.varData[variable].dtype
-                )
-                ret[:] = np.nan
-                for i in range(vShape[1]):
-                    ret[i] = self.varData[variable][:, i, :, :, :].ravel()
-                assert (ret != np.nan).all()
-                return ret
-            else:
-                return self.varData[variable][:].reshape(self.TotalCells)
+            elif self.OutputFormatVersion == 2:
+                if np.prod(vShape) > self.TotalCells:
+                      ret = np.empty(
+                          (vShape[1], self.TotalCells), dtype=self.varData[variable].dtype
+                      )
+                      ret[:] = np.nan
+                      for i in range(vShape[1]):
+                          ret[i] = self.varData[variable][:, i, :, :, :].ravel()
+                      assert (ret != np.nan).all()
+                      return ret
+                else:
+                    return self.varData[variable][:].reshape(self.TotalCells)
+
+            elif self.OutputFormatVersion == 3:
+                # TODO(BRR) This isn't quite right -- may need to output additional metadata
+                # Check for cell-centered data
+                if vShape[-1] * vShape[-2] * vShape[-3] == self.TotalCells:
+                    fieldShape = vShape[1:-3]
+                    totalFieldEntries = np.prod(fieldShape)
+                    ndim = len(fieldShape)
+                    if ndim == 0:
+                        return self.varData[variable].ravel()
+                    else:
+                        if ndim == 1:
+                            ret = np.empty(
+                              (vShape[1], self.TotalCells), dtype=self.varData[variable].dtype)
+                            ret[:] = np.nan
+                            for i in range(vShape[1]):
+                                ret[i] = self.varData[variable][:,i,:,:,:].ravel()
+                            assert (ret != np.nan).all()
+                            return ret
+                        elif ndim == 2:
+                            ret = np.empty(
+                              (vShape[1]*vShape[2], self.TotalCells), dtype=self.varData[variable].dtype)
+                            ret[:] = np.nan
+                            for i in range(vShape[1]):
+                                for j in range(vShape[2]):
+                                    ret[i + vShape[1]*j] = self.varData[variable][:,i,j,:,:,:].ravel()
+                            assert (ret != np.nan).all()
+                            return ret
+                        else:
+                            ret = np.empty(
+                              (vShape[1]*vShape[2]*vShape[3], self.TotalCells), dtype=self.varData[variable].dtype)
+                            ret[:] = np.nan
+                            for i in range(vShape[1]):
+                                for j in range(vShape[2]):
+                                    for k in range(vShape[3]):
+                                        ret[i + vShape[1]*(j + vShape[2]*k)] = self.varData[variable][:,i,j,k,:,:,:].ravel()
+                            assert (ret != np.nan).all()
+                            return ret
+                else:
+                  # Not cell-based variable
+                  raise Exception(
+                      f"Flattening only supported for cell-based variables but requested for {variable}"
+                  )
 
         if self.IncludesGhost and interior:
             nghost = self.NGhost
--- a/scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
+++ b/scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
@@ -499,21 +499,24 @@ class phdf:
 
             elif self.OutputFormatVersion == 2:
                 if np.prod(vShape) > self.TotalCells:
-                      ret = np.empty(
-                          (vShape[1], self.TotalCells), dtype=self.varData[variable].dtype
-                      )
-                      ret[:] = np.nan
-                      for i in range(vShape[1]):
-                          ret[i] = self.varData[variable][:, i, :, :, :].ravel()
-                      assert (ret != np.nan).all()
-                      return ret
+                    ret = np.empty(
+                        (vShape[1], self.TotalCells), dtype=self.varData[variable].dtype
+                    )
+                    ret[:] = np.nan
+                    for i in range(vShape[1]):
+                        ret[i] = self.varData[variable][:, i, :, :, :].ravel()
+                    assert (ret != np.nan).all()
+                    return ret
                 else:
                     return self.varData[variable][:].reshape(self.TotalCells)
 
             elif self.OutputFormatVersion == 3:
                 # Check for cell-centered data
-                if (vShape[-1] == self.MeshBlockSize[0] and vShape[-2] == self.MeshBlockSize[1] and
-                   vShape[-3] == self.MeshBlockSize[2]):
+                if (
+                    vShape[-1] == self.MeshBlockSize[0]
+                    and vShape[-2] == self.MeshBlockSize[1]
+                    and vShape[-3] == self.MeshBlockSize[2]
+                ):
                     fieldShape = vShape[1:-3]
                     totalFieldEntries = np.prod(fieldShape)
                     ndim = len(fieldShape)
@@ -522,36 +525,48 @@ class phdf:
                     else:
                         if ndim == 1:
                             ret = np.empty(
-                              (vShape[1], self.TotalCells), dtype=self.varData[variable].dtype)
+                                (vShape[1], self.TotalCells),
+                                dtype=self.varData[variable].dtype,
+                            )
                             ret[:] = np.nan
                             for i in range(vShape[1]):
-                                ret[i] = self.varData[variable][:,i,:,:,:].ravel()
+                                ret[i] = self.varData[variable][:, i, :, :, :].ravel()
                             assert (ret != np.nan).all()
                             return ret
                         elif ndim == 2:
                             ret = np.empty(
-                              (vShape[1]*vShape[2], self.TotalCells), dtype=self.varData[variable].dtype)
+                                (vShape[1] * vShape[2], self.TotalCells),
+                                dtype=self.varData[variable].dtype,
+                            )
                             ret[:] = np.nan
                             for i in range(vShape[1]):
                                 for j in range(vShape[2]):
-                                    ret[i + vShape[1]*j] = self.varData[variable][:,i,j,:,:,:].ravel()
+                                    ret[i + vShape[1] * j] = self.varData[variable][
+                                        :, i, j, :, :, :
+                                    ].ravel()
                             assert (ret != np.nan).all()
                             return ret
                         else:
                             ret = np.empty(
-                              (vShape[1]*vShape[2]*vShape[3], self.TotalCells), dtype=self.varData[variable].dtype)
+                                (vShape[1] * vShape[2] * vShape[3], self.TotalCells),
+                                dtype=self.varData[variable].dtype,
+                            )
                             ret[:] = np.nan
                             for i in range(vShape[1]):
                                 for j in range(vShape[2]):
                                     for k in range(vShape[3]):
-                                        ret[i + vShape[1]*(j + vShape[2]*k)] = self.varData[variable][:,i,j,k,:,:,:].ravel()
+                                        ret[
+                                            i + vShape[1] * (j + vShape[2] * k)
+                                        ] = self.varData[variable][
+                                            :, i, j, k, :, :, :
+                                        ].ravel()
                             assert (ret != np.nan).all()
                             return ret
                 else:
-                  # Not cell-based variable
-                  raise Exception(
-                      f"Flattening only supported for cell-based variables but requested for {variable}"
-                  )
+                    # Not cell-based variable
+                    raise Exception(
+                        f"Flattening only supported for cell-based variables but requested for {variable}"
+                    )
 
         if self.IncludesGhost and interior:
             nghost = self.NGhost
--- a/scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
+++ b/scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
@@ -488,10 +488,6 @@ class phdf:
 
         vShape = self.varData[variable].shape
         if flatten:
-            # if variable is not a scalar flatten cells component-wise
-            print('sdifj')
-            print(vShape)
-            print(self.TotalCells)
             # TODO(tbd) remove legacy mode in next major rel.
             if self.OutputFormatVersion == -1:
                 if np.prod(vShape) > self.TotalCells:
@@ -515,9 +511,9 @@ class phdf:
                     return self.varData[variable][:].reshape(self.TotalCells)
 
             elif self.OutputFormatVersion == 3:
-                # TODO(BRR) This isn't quite right -- may need to output additional metadata
                 # Check for cell-centered data
-                if vShape[-1] * vShape[-2] * vShape[-3] == self.TotalCells:
+                if (vShape[-1] == self.MeshBlockSize[0] and vShape[-2] == self.MeshBlockSize[1] and
+                   vShape[-3] == self.MeshBlockSize[2]):
                     fieldShape = vShape[1:-3]
                     totalFieldEntries = np.prod(fieldShape)
                     ndim = len(fieldShape)
--- a/src/interface/metadata.cpp
+++ b/src/interface/metadata.cpp
@@ -142,12 +142,16 @@ std::array<int, 6> Metadata::GetArrayDims(std::weak_ptr<MeshBlock> wpmb,
     for (int i = N; i < 3; i++)
       arrDims[i + 3] = 1;
   } else if (IsSet(Particle)) {
-    assert(N >= 1 && N <= 5);
+    assert(N >= 0 && N <= 5);
     arrDims[0] = 1; // To be updated by swarm based on pool size before allocation
     for (int i = 0; i < N; i++)
       arrDims[i + 1] = shape[i];
     for (int i = N; i < 5; i++)
       arrDims[i + 1] = 1;
+  } else if (IsSet(Swarm)) {
+    // No dimensions
+    // TODO(BRR) This will be replaced in the swarm packing PR, but is currently required
+    // since swarms carry metadata but do not have an array size.
   } else {
     // This variable is not necessarily tied to any specific
     // mesh element, so dims will be used as the actual array
--- a/src/interface/metadata.cpp
+++ b/src/interface/metadata.cpp
@@ -128,7 +128,7 @@ std::array<int, 6> Metadata::GetArrayDims(std::weak_ptr<MeshBlock> wpmb,
     // classes add the +1's where needed.  They all expect
     // these dimensions to be the number of cells in each
     // direction, NOT the size of the arrays
-    assert(N >= 1 && N <= 3);
+    assert(N >= 0 && N <= 3);
     PARTHENON_REQUIRE_THROWS(!wpmb.expired(),
                              "Cannot determine array dimensions for mesh-tied entity "
                              "without a valid meshblock");
--- a/src/interface/metadata.hpp
+++ b/src/interface/metadata.hpp
@@ -272,14 +272,6 @@ class Metadata {
   bool IsValid(bool throw_on_fail = false) const {
     bool valid = true;
 
-    // No empty shapes for variables not tied to mesh
-    if (shape_.size() == 0 && CountSet({None}) == 1) {
-      valid = false;
-      if (throw_on_fail) {
-        PARTHENON_THROW("Must specify non-empty Shape if variable is not tied to mesh");
-      }
-    }
-
     // Topology
     if (CountSet({None, Node, Edge, Face, Cell}) != 1) {
       valid = false;
--- a/src/interface/metadata.hpp
+++ b/src/interface/metadata.hpp
@@ -194,6 +194,8 @@ class Metadata {
   // There are 3 optional arguments: shape, component_labels, and associated, so we'll
   // need 8 constructors to provide all possible variants
 
+  // By default shape is empty; this corresponds to scalar data on the mesh
+
   // 4 constructors, this is the general constructor called by all other constructors, so
   // we do some sanity checks here
   Metadata(const std::vector<MetadataFlag> &bits, const std::vector<int> &shape = {},
@@ -270,6 +272,14 @@ class Metadata {
   // is true, throw a descriptive exception when invalid
   bool IsValid(bool throw_on_fail = false) const {
     bool valid = true;
+
+    // No empty shapes for variables not tied to mesh
+    if (shape.size() == 0 && CountSet({None}) == 1) {
+      valid = false;
+      if (throw_on_fail) {
+        PARTHENON_THROW("Must specify non-empty Shape if variable is not tied to mesh");
+      }
+    }
 
     // Topology
     if (CountSet({None, Node, Edge, Face, Cell}) != 1) {
--- a/src/interface/metadata.hpp
+++ b/src/interface/metadata.hpp
@@ -225,7 +225,7 @@ class Metadata {
     // check shape is valid
     // TODO(JL) Should we be extra pedantic and check that shape matches Vector/Tensor
     // flags?
-    PARTHENON_REQUIRE_THROWS(shape_.size() > 0, "Shape must have at least rank 1");
+    PARTHENON_REQUIRE_THROWS(shape_.size() >= 0, "Shape must have at least rank 0");
     if (IsMeshTied()) {
       PARTHENON_REQUIRE_THROWS(
           shape_.size() <= 3,
--- a/src/interface/metadata.hpp
+++ b/src/interface/metadata.hpp
@@ -227,7 +227,6 @@ class Metadata {
     // check shape is valid
     // TODO(JL) Should we be extra pedantic and check that shape matches Vector/Tensor
     // flags?
-    PARTHENON_REQUIRE_THROWS(shape_.size() >= 0, "Shape must have at least rank 0");
     if (IsMeshTied()) {
       PARTHENON_REQUIRE_THROWS(
           shape_.size() <= 3,
--- a/src/interface/metadata.hpp
+++ b/src/interface/metadata.hpp
@@ -196,7 +196,7 @@ class Metadata {
 
   // 4 constructors, this is the general constructor called by all other constructors, so
   // we do some sanity checks here
-  Metadata(const std::vector<MetadataFlag> &bits, const std::vector<int> &shape = {1},
+  Metadata(const std::vector<MetadataFlag> &bits, const std::vector<int> &shape = {},
            const std::vector<std::string> &component_labels = {},
            const std::string &associated = "")
       : shape_(shape), component_labels_(component_labels), associated_(associated) {
--- a/src/interface/metadata.hpp
+++ b/src/interface/metadata.hpp
@@ -51,6 +51,8 @@
   PARTHENON_INTERNAL_FOR_FLAG(Node)                                                      \
   /** particle variable */                                                               \
   PARTHENON_INTERNAL_FOR_FLAG(Particle)                                                  \
+  /** swarm */                                                                           \
+  PARTHENON_INTERNAL_FOR_FLAG(Swarm)                                                     \
   /************************************************/                                     \
   /** ROLE: Exactly one must be specified (default is Provides) */                       \
   /** Private to a package */                                                            \
--- a/src/interface/metadata.hpp
+++ b/src/interface/metadata.hpp
@@ -274,7 +274,7 @@ class Metadata {
     bool valid = true;
 
     // No empty shapes for variables not tied to mesh
-    if (shape.size() == 0 && CountSet({None}) == 1) {
+    if (shape_.size() == 0 && CountSet({None}) == 1) {
       valid = false;
       if (throw_on_fail) {
         PARTHENON_THROW("Must specify non-empty Shape if variable is not tied to mesh");
--- a/src/outputs/parthenon_hdf5.cpp
+++ b/src/outputs/parthenon_hdf5.cpp
@@ -842,16 +842,31 @@ void PHDF5Output::WriteOutputFileImpl(Mesh *pm, ParameterInput *pin, SimTime *tm
 
     int ndim = -1;
     if (vinfo.where == MetadataFlag(Metadata::Cell)) {
-      ndim = 3 + vinfo.ndim;
+      printf("CELL VARIABLE!\n");
+      ndim = 3 + vinfo.ndim + 1;
       for (int i = 0; i < vinfo.ndim; i++) {
-        local_count[1 + i] = alldims[3 - ndim + i];
+        local_count[1 + i] = global_count[1 + i] = alldims[3 - vinfo.ndim + i];
+        printf("extradim: alldims[%i] = %i\n", 3-vinfo.ndim + 1, alldims[3-vinfo.ndim+i]);
       }
-      local_count[ndim + 1] = nx3;
-      local_count[ndim + 2] = nx2;
-      local_count[ndim + 3] = nx1;
+      local_count[vinfo.ndim + 1] = global_count[vinfo.ndim + 1] = nx3;
+      local_count[vinfo.ndim + 2] = global_count[vinfo.ndim + 2] = nx2;
+      local_count[vinfo.ndim + 3] = global_count[vinfo.ndim + 3] = nx1;
     } else {
-      ndim = vinfo.ndim;
+      printf("NONE VARIABLE!\n");
+      ndim = vinfo.ndim + 1;
+      for (int i = 0; i < vinfo.ndim; i++) {
+        local_count[1 + i] = global_count[1 + i] = alldims[6 - vinfo.ndim + i];
+      }
     }
+
+    printf("label: %s\n", vinfo.label.c_str());
+    printf("  aldims: %i %i %i %i %i %i\n", alldims[0], alldims[1], alldims[2], alldims[3], alldims[4], alldims[5]);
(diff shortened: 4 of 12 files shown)


src/interface/metadata.hpp
```diff
@@ -196,7 +196,7 @@ class Metadata {
 
   // 4 constructors, this is the general constructor called by all other constructors, so
   // we do some sanity checks here
-  Metadata(const std::vector<MetadataFlag> &bits, const std::vector<int> &shape = {1},
+  Metadata(const std::vector<MetadataFlag> &bits, const std::vector<int> &shape = {},
```
brryan, line 203: The default value of `shape` being empty distinguishes a scalar field from a vector field with only one entry. Hopefully this isn't too much of a breaking change anywhere downstream, but I think it is necessary for variable output shape to match parthenon variable shape without having e.g. `[nblock, 1, 1, 1, nk, nj, ni]` for all scalar variables.

src/interface/metadata.cpp
```diff
@@ -128,7 +128,7 @@ std::array<int, 6> Metadata::GetArrayDims(std::weak_ptr<MeshBlock> wpmb,
     // classes add the +1's where needed.  They all expect
     // these dimensions to be the number of cells in each
     // direction, NOT the size of the arrays
-    assert(N >= 1 && N <= 3);
+    assert(N >= 0 && N <= 3);
```
Yurlungur, line 131: Not sure I follow this change

src/interface/metadata.hpp
Yurlungur: This should probably be commented somewhere in the metadata file, because it's pretty non-intuitive without a comment.

src/outputs/parthenon_hdf5.cpp
```diff
@@ -202,16 +202,24 @@ struct VarInfo {
   int nx6;
   int nx5;
   int nx4;
+  int ndim; // 1-, 2-, or 3-D
+  MetadataFlag where;
```
Yurlungur: why not just also report `nx3`, `nx2`, `nx1`? One could imagine these might not match meshblock shape. (In fact, I explicitly want us to support where they sometimes don't. For example geometry in Phoebus.)

Yurlungur (commented):
Thanks for taking this on! A few nitpicks below. (I know you're not done yet.)

src/interface/metadata.hpp
brryan: I actually think this change is more intuitive, in the sense that why do scalar fields have `shape = {1}` instead of `shape = {1, 1,}` or `shape = {1, 1, 1}`? That being said, this is a change to the way things have been for quite a while so I agree I should add a comment here

src/interface/metadata.cpp
brryan: This is to accommodate scalar fields having `shape = {}` as above, which means `N = 0` here.

src/outputs/parthenon_hdf5.cpp
brryan: Here I was using `where = Metadata::none` to account for non-meshblock-shaped variables (in fact I was specifically thinking of the Phoebus geometry :) ). The one issue with just getting all six `nx*` is that meshblock-tied and non-meshblock-tied variables have different `shape`s. For example the Phoebus geometry has some 4D `shape` for a 4D output grid whereas a cell-centered tensor in Phoebus will have a 2D shape and a 3D meshblock size for a 5D output grid. So at some point here we need to use `where` to make this distinction regardless.

src/interface/metadata.hpp
Yurlungur: I don't disagree regarding shape from a physics perspective, but from an array perspective I think the original way makes as much intuitive sense to me as anything.

To be clear, I don't object to your fix. Just want to make sure it's documented.

src/outputs/parthenon_hdf5.cpp
Yurlungur: hmm ok not sure I fully understand what you mean. But I guess it'll make sense when you're finished tweaking.

src/outputs/parthenon_hdf5.cpp
```diff
@@ -825,7 +829,30 @@ void PHDF5Output::WriteOutputFileImpl(Mesh *pm, ParameterInput *pin, SimTime *tm
...
+        local_count[1 + i] = global_count[1 + i] = alldims[3 - vinfo.ndim + i];
+      }
+      local_count[vinfo.ndim + 1] = global_count[vinfo.ndim + 1] = nx3;
+      local_count[vinfo.ndim + 2] = global_count[vinfo.ndim + 2] = nx2;
+      local_count[vinfo.ndim + 3] = global_count[vinfo.ndim + 3] = nx1;
+    } else if (vinfo.where == MetadataFlag(Metadata::None)) {
+      ndim = vinfo.ndim + 1;
+      for (int i = 0; i < vinfo.ndim; i++) {
+        local_count[1 + i] = global_count[1 + i] = alldims[6 - vinfo.ndim + i];
+      }
```
brryan, line 858: @Yurlungur here is where we need to know both `shape` and whether the variable is e.g. tied to cell centers (and thus whether its last three entries in `*_count` will be `nx3`, `nx2`, `nx1` or not)

brryan:
OK this works for me for both dumps and restarts downstream in phoebus with higher-dimensional data (e.g. a `[species, direction, nx3, nx2, nx1]` variable). I think the new output format, where HDF5 datasets match the parthenon datasets with unused `shape` indices suppressed (i.e. a scalar field has HDF5 shape `[nblocks, nx3, nx2, nx1]` instead of `[nblocks, 1, 1, 1, nx3, nx2, nx1]`), is what we agreed upon at the workshop. 

However, this is a breaking change. I incremented the output version number but I wonder to what degree we want to maintain support for old output versions? For example, we already require the latest `OUTPUT_VERSION_FORMAT` for restarts in `src/parthenon_manager.cpp`. 

Anyway I want to call downstream developers' attention to this so we can make sure this works for everyone. @bprather @forrestglines @pgrete

Yurlungur:
@brryan did you try metric information that's axisymmetric?

src/outputs/parthenon_hdf5.cpp
Yurlungur: I still don't understand this but I'll look more carefully.

Yurlungur:
> However, this is a breaking change. I incremented the output version number but I wonder to what degree we want to maintain support for old output versions? For example, we already require the latest OUTPUT_VERSION_FORMAT for restarts in src/parthenon_manager.cpp.

Presumably we need to update *all* the gold files with this change? Either that or support backwards portability

brryan:
> > However, this is a breaking change. I incremented the output version number but I wonder to what degree we want to maintain support for old output versions? For example, we already require the latest OUTPUT_VERSION_FORMAT for restarts in src/parthenon_manager.cpp.
> 
> Presumably we need to update _all_ the gold files with this change? Either that or support backwards portability

I think we will need to update all the gold files regardless of backward compatibility, because we want the tests to check against the current output format. If we support backwards portability we should be able to leave gold files used for restarts (if any) untouched, I think?

brryan:
> @brryan did you try metric information that's axisymmetric?

Good point (I just tried and it doesn't work) -- I'll fix this

forrestglines:
> Anyway I want to call downstream developers' attention to this so we can make sure this works for everyone. 

Since I'm supposedly in charge of maintaining the Parthenon frontend for yt, I'll take a look at this PR and make changes to yt and AthenaPK as I go. The format is not likely to further change within this PR, right? So it would be worth changing yt now?

> However, this is a breaking change. I incremented the output version number but I wonder to what degree we want to maintain support for old output versions? For example, we already require the latest `OUTPUT_VERSION_FORMAT` for restarts in `src/parthenon_manager.cpp`.

If we don't maintain support for older output versions, which I think is too much effort, we should instead provide a script to convert older versions to newer versions, especially if anyone currently has simulation data that they'd like to keep around. That script doesn't need to live in the repo but we should link it in this PR.

brryan:
> > @brryan did you try metric information that's axisymmetric?
> 
> Good point (I just tried and it doesn't work) -- I'll fix this

Fixed. The one issue is that because the axisymmetric variables are custom shapes, parthenon doesn't know how to not output the ghost zones. I think it's asking to much of parthenon to work this out for every case, and so I think it's reasonable to leave pruning the ghost zones to the downstream code developers. We can easily deal with this in phoebus in our `phoedf.py` class derived from `phdf`.

src/outputs/parthenon_hdf5.cpp
forrestglines: ~~Not fully understanding the context yet, but it seems like it would be more straightforward to just contain nx3, nx2, nx1, even if they usually match the cell-centered dims. The non-conventional shapes would be handled the same way. Or would constructing nx3, nx2, nx1 at the construction of `VarInfo` be inconvenient?~~ I think this comment is outdated since I made it

Yurlungur:
> we should instead provide a script to convert older versions to newer versions, especially if anyone currently has simulation data that they'd like to keep around. That script doesn't need to live in the repo but we should link it in this PR.

Big fan of this idea. I would in fact be in favor of sticking said script in the repo.

> Fixed. The one issue is that because the axisymmetric variables are custom shapes, parthenon doesn't know how to not output the ghost zones. I think it's asking to much of parthenon to work this out for every case, and so I think it's reasonable to leave pruning the ghost zones to the downstream code developers.

AAAAH. Yes I think that's fine.

Yurlungur:
> We can easily deal with this in phoebus in our phoedf.py class derived from phdf.

We should probably document this use case in the docs though.

brryan:
> > Anyway I want to call downstream developers' attention to this so we can make sure this works for everyone.
> 
> Since I'm supposedly in charge of maintaining the Parthenon frontend for yt, I'll take a look at this PR and make changes to yt and AthenaPK as I go. The format is not likely to further change within this PR, right? So it would be worth changing yt now?

Thanks for taking a look. I personally am happy with this format and don't have plans to change it, but if others don't like this as is there could still be changes. Sorry for the maybe not useful answer. For what it's worth, I'm already modifying phoebus assuming this output format going forward. 

> 
> > However, this is a breaking change. I incremented the output version number but I wonder to what degree we want to maintain support for old output versions? For example, we already require the latest `OUTPUT_VERSION_FORMAT` for restarts in `src/parthenon_manager.cpp`.
> 
> If we don't maintain support for older output versions, which I think is too much effort, we should instead provide a script to convert older versions to newer versions, especially if anyone currently has simulation data that they'd like to keep around. That script doesn't need to live in the repo but we should link it in this PR.

I think a conversion script is a good idea, and I'm happy to add that -- that way I could clean up some more code to do with older output versions inside parthenon. I think such a conversion script could reasonably live in our `parthenon_tools` python package. The only concern is that such a script might not be able to convert dump files too large to live on a single node's RAM.

Yurlungur:
> The only concern is that such a script might not be able to convert dump files too large to live on a single node's RAM.

Fixing that with `mpi4py` left as an exercise for the reader. ;)

src/outputs/parthenon_hdf5.cpp
```diff
@@ -825,7 +817,56 @@ void PHDF5Output::WriteOutputFileImpl(Mesh *pm, ParameterInput *pin, SimTime *tm
...
+      if (output_params.hdf5_compression_level > 0) {
+        // we need chunks to enable compression
+        std::array<hsize_t, H5_NDIM> chunk_size({1, 1, 1, 1, 1, 1, 1});
+        for (int i = 0; i < std::min<int>(vinfo.ndim, 3); i++) {
+          chunk_size[7 - i] = alldims[i];
+        }
+
+        PARTHENON_HDF5_CHECK(H5Pset_chunk(pl_dcreate, ndim, chunk_size.data()));
+        PARTHENON_HDF5_CHECK(H5Pset_deflate(
+            pl_dcreate, std::min(9, output_params.hdf5_compression_level)));
```
forrestglines: I think is is repeated code? It could be moved out of these conditionals along with the declaration for `chunk_size`

src/outputs/parthenon_hdf5.cpp
```diff
@@ -202,15 +202,19 @@ struct VarInfo {
   int nx6;
   int nx5;
   int nx4;
+  int ndim; // 1-, 2-, or 3-D
```
forrestglines: `ndim` refers to the dimensionality of the grid, not the tensor rank of the variable, right?

src/outputs/parthenon_hdf5.cpp
```diff
@@ -825,7 +817,56 @@ void PHDF5Output::WriteOutputFileImpl(Mesh *pm, ParameterInput *pin, SimTime *tm
...
+    std::vector<hsize_t> alldims({nx6, nx5, nx4, static_cast<hsize_t>(nx3),
+                                  static_cast<hsize_t>(nx2), static_cast<hsize_t>(nx1)});
+
+    int ndim = -1;
+    if (vinfo.where == MetadataFlag(Metadata::Cell)) {
+      ndim = 3 + vinfo.ndim + 1;
+      for (int i = 0; i < vinfo.ndim; i++) {
+        local_count[1 + i] = global_count[1 + i] = alldims[3 - vinfo.ndim + i];
+      }
+      local_count[vinfo.ndim + 1] = global_count[vinfo.ndim + 1] = nx3;
```
forrestglines: I'm a little bit confused how these lines match up given the indices offset by `vinfo.ndim` here and the same `local_count` offset by 3. If `vinfo.ndim=2` for a 2D simulation, wouldn't `nx4` over lap with `nx3` here?

src/outputs/parthenon_hdf5.cpp
brryan: So it actually turned out to be necessary to store `nx1` `nx2` and `nx3` here to resolve the issue that showed up with axisymmetric geometry variables in phoebus :)

src/outputs/parthenon_hdf5.cpp
brryan: I could be missing something but I think the issue here basically shows up because we want to suppress the unused dimensions in the ParArray e.g. `nx4`, `nx5`, `nx6` for a scalar field when generating the HDF5 dataset. But since these are always >= 1, we can't just look at the `nx*` to determine the actual shape of the desired HDF5 dataset.

src/interface/metadata.cpp
```diff
@@ -142,12 +142,16 @@ std::array<int, 6> Metadata::GetArrayDims(std::weak_ptr<MeshBlock> wpmb,
...
+    assert(N >= 0 && N <= 5);
     arrDims[0] = 1; // To be updated by swarm based on pool size before allocation
     for (int i = 0; i < N; i++)
       arrDims[i + 1] = shape[i];
     for (int i = N; i < 5; i++)
       arrDims[i + 1] = 1;
+  } else if (IsSet(Swarm)) {
+    // No dimensions
+    // TODO(BRR) This will be replaced in the swarm packing PR, but is currently required
+    // since swarms carry metadata but do not have an array size.
```
Yurlungur, line 154: Does this mean you're adding particle IO? :O :D

pgrete:
Given that @forrestglines already volunteered do handle the review and yt update, I removed myself as reviewer.
The only high level comment that I have already now that (to me) it'd be very important to maintain some backward compatibility (preferable transparent to the user, but if that's too complex, then via a conversion script) because I already have science data with the old format (and having one yt install per simulation because of different formats, would be rather unpractical -- especially for collaborators with less technical experience).

forrestglines (commented):
Here's my initial thoughts on the code. I still haven't made the changes to yt necessary to read these files due to other projects going on. I'm working on it today.

Since @pgrete has scientific data that uses the old format, I'd strongly recommend creating a script to convert the old data into new data before merging this commit. Has that effort been started?

forrestglines:
It might be preliminary but I might not need to make any changes to the yt frontend. I tried loading up Output-v2 and Output-v3 data in yt and it worked fine on 1D, 2D, and 3D data.

However, one of the tests in AthenaPK is failing while trying to write HDF5 data (it's test case 3 in "regression_test:aniso_therm_cond_ring_multid"). I'm still investigating that

Yurlungur (approved):

forrestglines:
The output change in Parthenon causes a test failure in AthenaPK while writing out to HDF5. It's unclear why they're an error, but here's the output while running test case 3 from the "regression_test:aniso_therm_cond_ring_multid" test
```
$ /sw/spack/delta-2022-03/apps/openmpi/4.1.4-gcc-11.2.0-7ooytax/bin/mpiexec -n 1 /u/glines/code/athenapk-project/builds/cuda-RelWithDebInfo-ampere80/bin/athenaPK -i /u/glines/code/athenapk-project/athenapk/inputs/diffusion.in parthenon/mesh/nx1=8 parthenon/meshblock/nx1=8 parthenon/mesh/nx2=64 parthenon/meshblock/nx2=32 parthenon/mesh/nx3=64 parthenon/meshbloc
k/nx3=32 problem/diffusion/iprob=21 parthenon/time/tlim=200.0 parthenon/output0/dt=200.0 parthenon/output0/id=3
#Variables in use:
# Variables:
# Name  Metadata flags
prim                      Cell,Provides,Real,Derived
cons                      Cell,Provides,Real,Independent,FillGhost,WithFluxes
# Sparse Variables:
# Name  sparse id       Metadata flags
# Swarms:
# Swarm Value   metadata

Starting up hydro driver

HDF5-DIAG: Error detected in HDF5 (1.12.2) MPI-process 0:
  #000: /tmp/svcdeltaswmgt/spack-stage/spack-stage-hdf5-1.12.2-7cnh43swqfluahwjmgiz4cp5rwajxdyk/spack-src/src/H5D.c line 141 in H5Dcreate2(): unable to create dataset
    major: Dataset
    minor: Unable to create file
  #001: /tmp/svcdeltaswmgt/spack-stage/spack-stage-hdf5-1.12.2-7cnh43swqfluahwjmgiz4cp5rwajxdyk/spack-src/src/H5VLcallback.c line 1806 in H5VL_dataset_create(): dataset create failed
    major: Virtual Object Layer
    minor: Unable to create file
  #002: /tmp/svcdeltaswmgt/spack-stage/spack-stage-hdf5-1.12.2-7cnh43swqfluahwjmgiz4cp5rwajxdyk/spack-src/src/H5VLcallback.c line 1771 in H5VL__dataset_create(): dataset create failed
    major: Virtual Object Layer
    minor: Unable to create file
  #003: /tmp/svcdeltaswmgt/spack-stage/spack-stage-hdf5-1.12.2-7cnh43swqfluahwjmgiz4cp5rwajxdyk/spack-src/src/H5VLnative_dataset.c line 73 in H5VL__native_dataset_create(): unable to create dataset
    major: Dataset
    minor: Unable to initialize object
  #004: /tmp/svcdeltaswmgt/spack-stage/spack-stage-hdf5-1.12.2-7cnh43swqfluahwjmgiz4cp5rwajxdyk/spack-src/src/H5Dint.c line 396 in H5D__create_named(): unable to create and link to dataset
    major: Dataset
    minor: Unable to initialize object
  #005: /tmp/svcdeltaswmgt/spack-stage/spack-stage-hdf5-1.12.2-7cnh43swqfluahwjmgiz4cp5rwajxdyk/spack-src/src/H5L.c line 1767 in H5L_link_object(): unable to create new link to object
    major: Links
    minor: Unable to initialize object
  #006: /tmp/svcdeltaswmgt/spack-stage/spack-stage-hdf5-1.12.2-7cnh43swqfluahwjmgiz4cp5rwajxdyk/spack-src/src/H5L.c line 2009 in H5L__create_real(): can't insert link
    major: Links
    minor: Unable to insert object
  #007: /tmp/svcdeltaswmgt/spack-stage/spack-stage-hdf5-1.12.2-7cnh43swqfluahwjmgiz4cp5rwajxdyk/spack-src/src/H5Gtraverse.c line 837 in H5G_traverse(): internal path traversal failed
    major: Symbol table
    minor: Object not found
  #008: /tmp/svcdeltaswmgt/spack-stage/spack-stage-hdf5-1.12.2-7cnh43swqfluahwjmgiz4cp5rwajxdyk/spack-src/src/H5Gtraverse.c line 613 in H5G__traverse_real(): traversal operator failed
    major: Symbol table
    minor: Callback failed
  #009: /tmp/svcdeltaswmgt/spack-stage/spack-stage-hdf5-1.12.2-7cnh43swqfluahwjmgiz4cp5rwajxdyk/spack-src/src/H5L.c line 1815 in H5L__link_cb(): unable to create object
    major: Links
    minor: Unable to initialize object
  #010: /tmp/svcdeltaswmgt/spack-stage/spack-stage-hdf5-1.12.2-7cnh43swqfluahwjmgiz4cp5rwajxdyk/spack-src/src/H5Oint.c line 2468 in H5O_obj_create(): unable to open object
    major: Object header
    minor: Can't open object
  #011: /tmp/svcdeltaswmgt/spack-stage/spack-stage-hdf5-1.12.2-7cnh43swqfluahwjmgiz4cp5rwajxdyk/spack-src/src/H5Doh.c line 289 in H5O__dset_create(): unable to create dataset
    major: Dataset
    minor: Unable to initialize object
  #012: /tmp/svcdeltaswmgt/spack-stage/spack-stage-hdf5-1.12.2-7cnh43swqfluahwjmgiz4cp5rwajxdyk/spack-src/src/H5Dint.c line 1348 in H5D__create(): unable to construct layout information
    major: Dataset
    minor: Unable to initialize object
  #013: /tmp/svcdeltaswmgt/spack-stage/spack-stage-hdf5-1.12.2-7cnh43swqfluahwjmgiz4cp5rwajxdyk/spack-src/src/H5Dchunk.c line 881 in H5D__chunk_construct(): chunk size must be <= maximum dimension size for fixed-sized dimensions
    major: Dataset
    minor: Unable to initialize object
  what():  ### PARTHENON ERROR
  Condition:   hid >= 0
  Message:     H5 FromHIDCheck failed
  File:        /u/glines/code/athenapk-project/athenapk/external/parthenon/src/outputs/parthenon_hdf5.hpp
  Line number: 67
  
  [gpua099:3233017] *** Process received signal ***
[gpua099:3233017] Signal: Aborted (6)
[gpua099:3233017] Signal code:  (-6)
[gpua099:3233017] [ 0] /usr/lib64/libpthread.so.0(+0x12b20)[0x7ff81d06db20]
[gpua099:3233017] [ 1] /usr/lib64/libc.so.6(gsignal+0x10f)[0x7ff81c31c37f]
[gpua099:3233017] [ 2] /usr/lib64/libc.so.6(abort+0x127)[0x7ff81c306db5]
[gpua099:3233017] [ 3] /sw/spack/delta-2022-03/apps/gcc/11.2.0-gcc-8.4.1-fxgnsyr/lib64/libstdc++.so.6(+0xa966a)[0x7ff81ccee66a]
[gpua099:3233017] [ 4] /sw/spack/delta-2022-03/apps/gcc/11.2.0-gcc-8.4.1-fxgnsyr/lib64/libstdc++.so.6(+0xb4eca)[0x7ff81ccf9eca]
[gpua099:3233017] [ 5] /sw/spack/delta-2022-03/apps/gcc/11.2.0-gcc-8.4.1-fxgnsyr/lib64/libstdc++.so.6(+0xb4f35)[0x7ff81ccf9f35]
[gpua099:3233017] [ 6] /sw/spack/delta-2022-03/apps/gcc/11.2.0-gcc-8.4.1-fxgnsyr/lib64/libstdc++.so.6(+0xb51c9)[0x7ff81ccfa1c9]
[gpua099:3233017] [ 7] /u/glines/code/athenapk-project/builds/cuda-RelWithDebInfo-ampere80/bin/athenaPK[0x43178c]
[gpua099:3233017] [ 8] /u/glines/code/athenapk-project/builds/cuda-RelWithDebInfo-ampere80/bin/athenaPK[0x69cfbd]
[gpua099:3233017] [ 9] /u/glines/code/athenapk-project/builds/cuda-RelWithDebInfo-ampere80/bin/athenaPK[0x6a460b]
[gpua099:3233017] [10] /u/glines/code/athenapk-project/builds/cuda-RelWithDebInfo-ampere80/bin/athenaPK[0x6b15f0]
[gpua099:3233017] [11] /u/glines/code/athenapk-project/builds/cuda-RelWithDebInfo-ampere80/bin/athenaPK[0x695dc2]
[gpua099:3233017] [12] /u/glines/code/athenapk-project/builds/cuda-RelWithDebInfo-ampere80/bin/athenaPK[0x60bcaa]
[gpua099:3233017] [13] /u/glines/code/athenapk-project/builds/cuda-RelWithDebInfo-ampere80/bin/athenaPK[0x41f428]
[gpua099:3233017] [14] /usr/lib64/libc.so.6(__libc_start_main+0xf3)[0x7ff81c308493]
[gpua099:3233017] [15] /u/glines/code/athenapk-project/builds/cuda-RelWithDebInfo-ampere80/bin/athenaPK[0x42c7ae]
[gpua099:3233017] *** End of error message ***
--------------------------------------------------------------------------
Primary job  terminated normally, but 1 process returned
a non-zero exit code. Per user-direction, the job has been aborted.
--------------------------------------------------------------------------
--------------------------------------------------------------------------
mpiexec noticed that process rank 0 with PID 3233017 on node gpua099 exited on signal 6 (Aborted).
--------------------------------------------------------------------------
```
Where the chunk_size is constructed on parthenon_hdf5.cpp:848, `chunk_size={1,1,1,1,32,32,8}` with `ndim=5` for our 3D sim.  Wouldn't this use only the first `ndim` dimensions for chunk sizes, the first four dimensions being 1's?

src/outputs/parthenon_hdf5.cpp
```diff
@@ -798,34 +794,84 @@ void PHDF5Output::WriteOutputFileImpl(Mesh *pm, ParameterInput *pin, SimTime *tm
...
+      local_count[vinfo.ndim + 2] = global_count[vinfo.ndim + 2] = nx2;
+      local_count[vinfo.ndim + 3] = global_count[vinfo.ndim + 3] = nx1;
+
+#ifndef PARTHENON_DISABLE_HDF5_COMPRESSION
+      if (output_params.hdf5_compression_level > 0) {
+        // we need chunks to enable compression
+        const std::array<hsize_t, H5_NDIM> chunk_size(
+            {1, 1, 1, 1, static_cast<hsize_t>(nx3), static_cast<hsize_t>(nx2),
+             static_cast<hsize_t>(nx1)});
+        PARTHENON_HDF5_CHECK(H5Pset_chunk(pl_dcreate, ndim, chunk_size.data()));
```
forrestglines:
```suggestion
        PARTHENON_HDF5_CHECK(H5Pset_chunk(pl_dcreate, ndim, chunk_size.data()+(7-ndim)));
```
Very quick hack but this worked to fix the failing test in AthenaPK. This would have to be fixed on the line below too

pgrete: Why does this fix work?
This looks to me like there's a discrepancy between been the chuck size definition and the slab definition.
IIRC we always want chunks that cover the data of a variable (component) of the entire block.

Yurlungur: It looks to me like the fix is either @forrestglines fix or the array size should instead always be 7 in the set chunk call?

forrestglines: From what I now understand about this commit, this bug, and why it's occurred here:

With both this commit and previous versions (output version 3 vs output version 2) AthenaPK will always output 5D data -- `block,var,k,j,i`. Other than the version increase from 2 to 3, the data format for AthenaPK does not change with this commit, which I've verified with `h5diff` for one case.

The chunk size here has been incorrectly constructed in the new commit, however. It now constructs the array `{1,1,1,nvar,nz,ny,nx}` for AthenaPK then tells HDF5 to use the first 5 dimensions for chunks, in other words, `{1,1,1,nvar,nz}` for chunks.

In this particular example we use a 32x32x8 grid with 9 variables, leading to a chunk size of `{1,1,1,9,32}`. We then write `{nblocks,9,32,32,8}` of data, which HDF5 apparently dislikes because that last dimension is smaller than the chunk dimension. Maybe this is the only test where `ndim<7` and `nz>nx`, which lead to this chunk error?

My fix is a dirty pointer arithmetic fix to push HDF5 to actually use the last 5 dimensions for a chunk size, or `{1,nvar,nz,ny,nx}`. A better fix would be to construct the array with `ndim` dimensions in the first place.

@Yurlungur  I first tried to increase the chunk dimensions to 7 to use all the array for the chunk size, which also caused HDF5 to fail.

Yurlungur: I think your fix makes sense.

pgrete: Your analysis makes sense @forrestglines 

Why does the higher dim chuncking not work?
Sounds like a bug in HDF5?

I think in terms of chunks we should force chunks of `{1,.., 1,nz,ny,nx}`  (or more specifically `{other potential dims set to 1} + {nz, ny, ,nx}`) because compression will work over individual chunks.
This way we'd compress always over a single component (scalar, vector or tensor component).

forrestglines: I don't think it's a bug, it's reasonable for HDF5 to require the chunk dimension to match or at least not exceed the data dimension.

For our compression purposes in AthenaPK, I'd imagine there's a lot of compression to be made within the variable vector since velocities in different dimensions and magnetic fields in different dimensions should be similar

brryan: @forrestglines I agree with what you say above, thanks for identifying and solving this bug. I've pushed a fix where we construct `chunk_size` to have the correct entries in the first `ndim` array elements. It is still sized to `H5_NDIM` at compile time. Could I trouble you to re-run the AthenaPK test to see that it still works (or let me know which test failed previously on chunking and I can try)?

I'm not sure it's an HDF5 bug if the always-7-dimensional chunk size was failing, since the actual output datasets are not always 7-dimensional and I assume chunk and dataset need to have the same dimension. I could be wrong though.

src/parthenon_manager.cpp
```diff
@@ -235,7 +235,7 @@ void ParthenonManager::RestartPackages(Mesh &rm, RestartReader &resfile) {
     // Being extra stringent here so that we don't forget to update the machinery when
     // another change happens.
     PARTHENON_REQUIRE_THROWS(
-        HDF5::OUTPUT_VERSION_FORMAT == 2,
+        HDF5::OUTPUT_VERSION_FORMAT == 3,
```
brryan: Given @forrestglines's demonstration that for AthenaPK the HDF5 format is unchanged, I think we can think of this PR more as a bugfix for higher-than-vector-dimensional data. In that case, does it still make sense to provide a conversion script to go from file format `2` to `3`? The file would either be unchanged, or higher-than-vector data would be would be rewritten ambiguously since that is basically the issue this PR is fixing. Instead of such a conversion script is it ok to simply have this check be 
```c++
PARTHENON_REQUIRE_THROWS(HDF5::OUTPUT_VERSION_FORMAT == 2 || HDF5::OUTPUT_VERSION_FORMAT == 3, "Auto conversion from old to current format not implemented yet.");
if (HDF5::OUTPUT_VERSION_FORMAT == 2) {
PARTHENON_WARN("HDF5 output version format 2 may be erroneous for data other than cell-centered scalar and vector fields");
}
```
where I've added a warning about higher-than-vector-dimensional data for the old file format?

c-prather: Does AthenaPK use scalar variables?  Won't those be 3D now instead of 4D, or am I misremembering?  If that's still a difference, I think an equivalent of squeeze() for scalars would be simple enough to write into Parthenon itself or especially a script.  Even if we don't provide a supported version we could list a short example in the changelog or docs like:

```python
file = h5py.File(fname, 'r')
for key in file:
    file[key] = np.squeeze(file[key][()])
```

ought to cover it.  Maybe with like, good code, and error checks or something idk.

Also: if no one was using the old (buggy) tensors, it just seems confusing to spit an error for everyone with V2 files.  I'd be in favor of removing the error entirely, but if fail loud is the law of the land maybe a check for at least one >1D variable after all the packages are loaded?

src/outputs/parthenon_hdf5.cpp
forrestglines: The test suite just passed - this does fix the issue.

src/parthenon_manager.cpp
forrestglines: If @pgrete is the only user with science data in the old format, which actually matches the new format, then I don't see a need anymore for the fix. Do others agree?

forrestglines (approved):

src/parthenon_manager.cpp
c-prather: My intent was to point out that I believe the format is only the same if you don't use scalar variables, which KHARMA does.  We also have a lot of data in the old format (and actually quite a bit in the format prior to that one, too...)

However, the standard tool for plotting and converting KHARMA output just supports both formats.  So, it's no bother to me or (afaik) KHARMA users if the format changes without any printed warnings or Parthenon-provided migration tools.

brryan:
@par-hermes format

brryan:
@bprather I edited the code to allow for restarts from v2 restart files assuming only scalar/vector data, and checked that this works for me with phoebus in GRMHD (all scalar data). I removed the warning about restarting from v2, there is already anyway another warning when restarting from older file formats that new outputs will be in the latest format.

I also edited `phdf.py` to change the way variables are flattened. This should not affect how vector fields are flattened, but scalar fields will have the `var` index squashed. (The changes are mainly to *fix* flattening for non-cell-based custom shape and higher-than-vector field data).

src/outputs/parthenon_hdf5.cpp
brryan: Thanks @forrestglines

brryan:
OK I think this is good now, the extended CI just succeeded (https://github.com/parthenon-hpc-lab/parthenon/actions/runs/3482314300) though the status hasn't been reported here. 

Any objections to me merging this maybe this afternoon @bprather @lroberts36 ?

src/interface/metadata.cpp
brryan: This isn't really related to particle IO (though that must be added...), just for working around some array size-based checks that don't apply to Swarms since they dont have a size but still carry metadata.

src/parthenon_manager.cpp
brryan: Currently this PR can restart from either its own dump format `3`, or the previous format `2` assuming only scalar and vector data. I personally don't see a need for additional fixes. @bprather if you think we could make additional changes here to improve your workflow let me know, but my impression is this PR as is should work pretty well for you.

scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
```diff
@@ -488,24 +488,85 @@ def Get(self, variable, flatten=True, interior=False):
...
+                                        ] = self.varData[variable][
+                                            :, i, j, k, :, :, :
+                                        ].ravel()
+                            assert (ret != np.nan).all()
+                            return ret
+                else:
+                    # Not cell-based variable
+                    raise Exception(
+                        f"Flattening only supported for cell-based variables but requested for {variable}"
+                    )
```
lroberts36, line 569: Does phdf still work for exploded views with ghost zones? I can try to check this.

c-prather:
I'm happy with merging this as-is, at this point.  Thanks for working out restarts.

CI should run & report when you hit auto-merge now.

src/outputs/parthenon_hdf5.cpp
```diff
@@ -868,9 +926,16 @@ void PHDF5Output::WriteOutputFileImpl(Mesh *pm, ParameterInput *pin, SimTime *tm
 
       if (!is_allocated) {
         if (vinfo.is_sparse) {
-          hsize_t N = varSize * vlen;
-          memset(tmpData.data() + index, 0, N * sizeof(OutT));
-          index += N;
+          hsize_t varSize{};
+          if (vinfo.where == MetadataFlag(Metadata::Cell)) {
+            varSize = vinfo.nx6 * vinfo.nx5 * vinfo.nx4 * (out_kb.e - out_kb.s + 1) *
+                      (out_jb.e - out_jb.s + 1) * (out_ib.e - out_ib.s + 1);
```
lroberts36, line 932: Why not just use the sizes contained in `vinfo` and get rid of the `if`?

lroberts36 (approved):
LGTM, comments are non-blocking

src/outputs/parthenon_hdf5.cpp
brryan: Using the sizes in `vinfo` was actually a bug that broke the `sparse_advection` restart test that I just fixed earlier here -- the `vinfo` dimensions contain ghost zones but to remove those we need e.g. `out_kb.e - out_kb.s + 1`. Whereas for non-`Metadata::Cell` variables, we don't in general know that e.g. `out_kb.e - out_kb.s + 1` is the appropriate subset of `nx3`.

src/outputs/parthenon_hdf5.cpp
brryan: Right here `ndim` refers to the tensor rank of the variable -- this is confusing with existing naming so I just changed this to `tensor_rank`

src/outputs/parthenon_hdf5.cpp
brryan: Yeah I shouldn't've used `ndim` for my  `vinfo.ndim`. Now `vinfo.tensor_rank` which I think answers your question. This is a cell-centered variable so the `k,j,i` indices are always present (this PR's output format, like previous formats, doesn't squash spatial indices based on dimensionality of the problem)

Also sorry I missed your comments before, for some reason I wasn't seeing them on the github PR page.

src/outputs/parthenon_hdf5.cpp
brryan: Yes you're right, this is repeated code, I've merged & moved the HDF5 calls out of the conditional

PR #777, opened 2022-11-01
New action: check PR dependencies & warn until requirements merged

c-prather:
Add a Github action which checks dependencies between PRs.  Should fail until all required PRs are merged.

Action repository is here: https://github.com/gregsdennis/dependencies-action

The required access token need cover only reading pull requests and issues, which are public anyway.  It keys on any dependencies mentioned in the opening comment, e.g.

Depends on #772

PR #782, opened 2022-11-03
WIP: Signal Handling

c-prather:
Currently, Parthenon has custom handlers for SIGINT, SIGTERM, and SIGALRM, which all set a flag, wait until the end of the next step, output any final files, and then terminate the program (a 'clean exit', as opposed to immediate termination). 

This PR as-is would exit immediately if more than three SIGINTs are received.  This was intended to restore an easy way of exiting the program without the extra stepping and file operations, which may not be necessary and could take a long time.

However, in researching and testing behavior under MPI, I've discovered that triggering a clean exit (that is, sending a single SIGINT or SIGTERM) is MPI implementation-dependent.  So much so that there is not a single key combination which will send a single SIGINT to Parthenon in all circumstances: some frontends (srun) eat the first SIGINT and do nothing, whereas others (OpenMPI) effectively do nothing on 2 or more SIGINTs as they exit immediately themselves upon receiving the second.  In fact, OpenMPI will never actually send a SIGINT specifically -- when it receives one, it upgrades it to a SIGTERM before sending it on (and then sets a timer before it sends a SIGKILL in addition).

As far as I know, there does not exist any signal which will be passed cleanly to a user process by all major MPI implementations, with the very possible exception of USR2, which seems like a documentation nightmare.  Certainly, signal forwarding is nowhere in the standard.

Thus I propose that Parthenon drop support for providing any special behavior upon receiving signals, and revert to using the default handlers/behavior (immediate exit on SIGINT and up).  Instead, we should cleanly exit upon seeing a file in the working directory (e.g., "die_now", or "exit_now" or something).  This is already what we use for triggering output ("output_now").

To be clear, this is a shame to me -- I'm very surprised the MPI standard isn't stricter about this, and that implementations have picked such a wide variety of useless extra behavior to tack to a simple ctrl-C.

## PR Summary

This would make the "signalflag" array in SignalHandler into a counter of signals received since the last reset.  It then adds intelligence to SetSignalFlag to throw an error if some number (currently 3) SIGINTs have been received since the flag was last cleared.

The number/behavior is not currently runtime-configurable, as at that point it's probably better to make SignalHandler a class, and I wasn't sure this was worth that effort yet.

Also, I updated the FS check to use std::filesystem::exists now that we're C++17, as @pgrete left in a comment a while ago.

## PR Checklist

- [ ] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/src/utils/utils.hpp
+++ b/src/utils/utils.hpp
@@ -50,7 +50,7 @@ constexpr int nsignal = 3;
 static volatile int signalflag[nsignal + 1];
 
 // Immediately throw upon receiving 3+ SIGINT without clearing them
-static int SIGINTS_BEFORE_THROW = 3;
+const int SIGINTS_BEFORE_THROW = 3;
 
 static sigset_t mask;
 void SignalHandlerInit();
--- a/src/utils/utils.hpp
+++ b/src/utils/utils.hpp
@@ -40,11 +40,18 @@ void ShowConfig();
 //  \brief static data and functions that implement a simple signal handling system
 namespace SignalHandler {
 
+// Enum of desired signal results
 enum class OutputSignal { none, now, final };
-constexpr int nsignal = 3;
-// using the +1 for signaling based on "output_now" trigger
-static volatile int signalflag[nsignal + 1];
+// Signals handled: SIGTERM, SIGINT, SIGALRM
 const int ITERM = 0, IINT = 1, IALRM = 2;
+constexpr int nsignal = 3;
+// Flag/counter of signals received of each type
+// using the +1 for signaling based on a trigger file
+static volatile int signalflag[nsignal + 1];
+
+// Immediately throw upon receiving 3+ SIGINT without clearing them
+static int SIGINTS_BEFORE_THROW = 3;
+
 static sigset_t mask;
 void SignalHandlerInit();
 OutputSignal CheckSignalFlags();
--- a/src/utils/utils.hpp
+++ b/src/utils/utils.hpp
@@ -43,14 +43,14 @@ namespace SignalHandler {
 // Enum of desired signal results
 enum class OutputSignal { none, now, final };
 // Signals handled: SIGTERM, SIGINT, SIGALRM
-const int ITERM = 0, IINT = 1, IALRM = 2;
+constexpr int ITERM = 0, IINT = 1, IALRM = 2;
 constexpr int nsignal = 3;
 // Flag/counter of signals received of each type
 // using the +1 for signaling based on a trigger file
 static volatile int signalflag[nsignal + 1];
 
 // Immediately throw upon receiving 3+ SIGINT without clearing them
-const int SIGINTS_BEFORE_THROW = 3;
+constexpr int SIGINTS_BEFORE_THROW = 3;
 
 static sigset_t mask;
 void SignalHandlerInit();
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - [[PR 712]](https://github.com/lanl/parthenon/pull/712) Allow to add params from cmdline
 
 ### Changed (changing behavior/API/variables/...)
+- [[PR 782]](https://github.com/parthenon-hpc-lab/parthenon/pull/782) Die on repeatedly receiving SIGINT
 - [[PR 758]](https://github.com/lanl/parthenon/pull/758) Bump required C++ standard to C++17
 - [[PR 710]](https://github.com/lanl/parthenon/pull/710) Remove data transpose in hdf5 and restart outputs
 - [[PR 713]](https://github.com/lanl/parthenon/pull/713) Remove Coordinates stub in favor of Coordinates_t
--- a/src/utils/signal_handler.cpp
+++ b/src/utils/signal_handler.cpp
@@ -24,6 +24,7 @@
 // first 2x macros and signal() are the only ISO C features; rest are POSIX C extensions
 #include <csignal>
 #include <iostream>
+#include <filesystem>
 
 #include "parthenon_mpi.hpp"
 
@@ -61,10 +62,8 @@ void SignalHandlerInit() {
 
 OutputSignal CheckSignalFlags() {
   if (Globals::my_rank == 0) {
-    // TODO(the person bumping std to C++17): use std::filesystem::exists
-    struct stat buffer;
     // if file "output_now" exists
-    if (stat("output_now", &buffer) == 0) {
+    if (std::filesystem::exists("output_now")) {
       signalflag[nsignal] = 1;
     }
   }
@@ -131,15 +130,17 @@ void SetSignalFlag(int s) {
   // Signal handler functions must have C linkage; C++ linkage is implemantation-defined
   switch (s) {
   case SIGTERM:
-    signalflag[ITERM] = 1;
+    signalflag[ITERM] += 1;
     signal(s, SetSignalFlag);
     break;
   case SIGINT:
-    signalflag[IINT] = 1;
+    signalflag[IINT] += 1;
+    if (signalflag[IINT] >= SIGINTS_BEFORE_THROW)
+      PARTHENON_THROW("Terminating immediately on repeated Terminate signal");
     signal(s, SetSignalFlag);
     break;
   case SIGALRM:
-    signalflag[IALRM] = 1;
+    signalflag[IALRM] += 1;
     signal(s, SetSignalFlag);
     break;
   default:
--- a/src/utils/signal_handler.cpp
+++ b/src/utils/signal_handler.cpp
@@ -23,8 +23,8 @@
 
 // first 2x macros and signal() are the only ISO C features; rest are POSIX C extensions
 #include <csignal>
-#include <iostream>
 #include <filesystem>
+#include <iostream>
 
 #include "parthenon_mpi.hpp"
 


c-prather:
@par-hermes format

src/utils/utils.hpp
```diff
@@ -40,11 +40,18 @@ void ShowConfig();
 //  \brief static data and functions that implement a simple signal handling system
 namespace SignalHandler {
 
+// Enum of desired signal results
 enum class OutputSignal { none, now, final };
+// Signals handled: SIGTERM, SIGINT, SIGALRM
+const int ITERM = 0, IINT = 1, IALRM = 2;
```
Yurlungur: Can we make these `constexpr`?

src/utils/utils.hpp
```diff
@@ -40,11 +40,18 @@ void ShowConfig();
...
 constexpr int nsignal = 3;
-// using the +1 for signaling based on "output_now" trigger
+// Flag/counter of signals received of each type
+// using the +1 for signaling based on a trigger file
 static volatile int signalflag[nsignal + 1];
-const int ITERM = 0, IINT = 1, IALRM = 2;
+
+// Immediately throw upon receiving 3+ SIGINT without clearing them
+const int SIGINTS_BEFORE_THROW = 3;
```
Yurlungur: `constexpr`

Yurlungur (approved):
Assuming this doesn't break anything---and I don't think it should but we should check---I'm a fan. :+1:

You do need a changelog and tests need to pass. Approval contingent on those being satisfied.

src/utils/utils.hpp
c-prather: There's a comment in SignalHandler: "signal handling functions must have C linkage."  I'd assumed the 'const' was related to that so I left it, but I'll be the first to admit knowing nothing about how signal callbacks work, so I was acting out of superstition.
If it was a mistake and constexpr is fine I'm happy to change it.

c-prather: ...but I guess PARTHENON_THROW is pretty solidly C++, so

c-prather:
Should all be done; it doesn't seem like constexpr breaks compiling or operation (for me).

Generally, I'm not sure small PRs like this  & #777 deserve changelog entries.  There's always the full list of closed PRs, for the curious.  Policy change to bring up at a meeting, maybe.

src/utils/utils.hpp
Yurlungur: Uh I have no idea. I defer to you and/or whoever originally wrote this part of the code.

Yurlungur:
> Generally, I'm not sure small PRs like this & https://github.com/parthenon-hpc-lab/parthenon/pull/777 deserve changelog entries. There's always the full list of closed PRs, for the curious. Policy change to bring up at a meeting, maybe.

There's a discussion there to be had for sure. (Personally I prefer changelogs.)

pdmullen:
> takes a very long time before the first step, between steps, or never steps at all -- so, waiting until the next step to process a flag sometimes means minutes of useless crunching and file writes when all I wanted was to stop execution.....I think generally when someone sends ^C^C^C^C^C^C^C to a program, they expect it to die.

This PR is certainly one solution...but why not just `control` + `z` and then `kill -9 [pid]`?

pdmullen (commented):
> takes a very long time before the first step, between steps, or never steps at all -- so, waiting until the next step to process a flag sometimes means minutes of useless crunching and file writes when all I wanted was to stop execution.....I think generally when someone sends ^C^C^C^C^C^C^C to a program, they expect it to die.

This PR is certainly one solution...but why not just `control` + `z` and then `kill -9 [pid]`?

c-prather:
There are definitely always ways to just manually kill the process, both locally and in an MPI/scheduler context (even without losing the reservation, if you're clever about 'scancel --signal' use).  So this definitely isn't missing functionality, I agree.

The issue I'm trying to address here is more about expectations & convenience though: since Parthenon is trying to be clever about SIGINT and SIGTERM rather than leaving the default behavior, our replacement behavior should at least be able to do the same things.  Stopping execution should also be convenient enough that I can think mostly about getting back to correcting the typo in my call, rather than remembering how to use 'ps -ef | grep'.

The idea of repeated signals is supposed to emulate what MPICH (& Intel MPI and so on) already does; it forwards the first SIGINT as-is, but then sends SIGKILL under the hood if another SIGINT is received before the processes have exited.

I should mention, though, that Parthenon should not support (and generally, people should be cautious of) trying to produce consistent response to signals sent via interaction with mpirun/mpiexec/srun (or even manually, via scancel or similar).  This is because every MPI implementation treats signals differently, and every workload manager (looking at you, srun) has the potential to complicate things further.

For reference, behavior of each MPI on SIGINT received:
1. MPICH & derivatives: 1st: SIGINT to processes. 2nd: kill processes and then self.
2. OpenMPI & derivatives: 1: SIGTERM (no, not SIGINT, I checked) to processes, timer to SIGKILL. 2: kill MPI host process but *not* spawned application processes
3. IBM MPI & LSF: Just passes through all signals unmodified.  You know, like you'd expect.
4. srun: 1: prints job status (if enabled) 2: sends SIGINT to individual processes (I think?  It might be to the MPI host process) 3: kills srun, but *not* MPI nor child processes (at least I don't *think* it kills MPI -- again, the PMIx interface is obscure and complex)

It's a *mess*.  Never use signals.

tagardiner:
FWIW - the recommended way to terminate a job on a cluster at Sandia is to send the signal SIGUSR1, e.g.

$ scancel -s USR1 <job_id>

The default action of this signal is to terminate the process, but it can be caught and this signal (together with SIGUSR2) are intended to have a user defined behavior.

c-prather:
Interesting!

It looks like USR1 is used internally in MPICH, and that hydra doesn't pass it through for that reason.  I believe OpenMPI passes through just USR1 and USR2 and the suspend/resume signals, and only IBM MPI passes through anything else (e.g., SIGALRM).

To me, all this inconsistency is actually a very good reason to do as little with signals in Parthenon as possible, generally.  *Three* layers of custom behavior on SIGINT just seems too unpredictable to make it the supported way of doing *anything* except immediately exiting, including performing any output or clean exit operations.

As an example of the problem: someone is using Intel MPI with a simulation that steps slowly, maybe it's on Stampede2.  They actually *want* the documented behavior, which is to output a file, then exit cleanly, printing statistics etc.  They press ctrl-C.  Except, since they usually run under 'srun', they know that the first ctrl-C only prints some statistics, so they have to press it twice.  Except oops, they're running under Intel MPI directly this time!  Rather than printing or outputting anything, they've just sent SIGKILL.

We can't fix the MPI implementations, but we can avoid putting users in a position where they *ever* have to send a signal and expect a particular result other than immediate program death.  To that end, I'm changing my position: Parthenon should not trap signals at all.  Instead, we should exit cleanly only on:
1. Wallclock timer
2. Presence of a file, call it "die_now"

I think flag files are as inelegant as the next person, but signals just can't promise consistent treatment and state across a large MPI run, or especially, consistent treatment between machines.

pgrete:
> 'm changing my position: Parthenon should not trap signals at all. Instead, we should exit cleanly only on:
> 
>     1. Wallclock timer
> 
>     2. Presence of a file, call it "die_now"

:+1: on this from my point of view. I also tripped over different signal handling on different machines, so let's remove this "undefined behavior".
With respect to naming I vote for `exit_now` over `die_now` as `die` may imply an unclean exit.

c-prather:
Agreed re: naming, implemented.

One of the cases files don't directly cover is backfill jobs, which can be killed by the scheduler after some time with only a signal as warning.  We should (I will) also provide an example job submission script which schedules a signal before job termination, traps it with the submit script, and creates the exit_now file(s) for running MPI jobs.  Setting up something like this is always scheduler dependent, but I think the scheduler is the right place to handle this and gives more flexibility, e.g. using the USR2 signal or whatever plays nice with the machine's particular stack.

Yurlungur:
@bprather what is the status of this PR? Was the discussion ever resolved?

nbowden045:
I understand the point about needing a clean way to **most certainly** kill a program even when it is not working right. But there were two use cases I was considering where a graceful exit on a signal might be nice:

1. Preemptible queues: on NCSA Delta and maybe other machines, using a preemptible partition will guarantee 10 min of execution at least before sending a SIGTERM followed 5 min later by SIGKILL (but maybe its silly to try to run anything but test problems on such a queue)
2. As an adaptive way to end based on wallclock: you can ask slurm to send a chosen signal to the process sufficiently before the wall clock time, e.g. `#SBATCH --signal=SIGTERM@600`, that way whatever time I configure via `#SBATCH` or dynamically update from the command line of sbatch, I would not have to worry that the `-t` argument to Parthenon-supported code does not match the new wall time. I realized this sigterm does not work (results in no final file, and a normal output could be incomplete/malformed)

I found a way to get what I want for (2) by doing the following calculation in my batch script to get input for `-t` from the slurm time limit:
```bash
# (Written by Google AI search, then tweaked)
# Convert D-HH:MM:SS to seconds (handles missing days gracefully)
orig_time_limit=$(squeue -j $SLURM_JOB_ID -h --Format TimeLimit)
if [[ $orig_time_limit == *-* ]]; then
    days=${orig_time_limit%%-*}
    hms=${orig_time_limit#*-}
else
    days=0
    hms=$orig_time_limit
fi

IFS=: read -r hours minutes seconds <<< "$hms"
total_seconds=$(( days * 86400 + hours * 3600 + minutes * 60 + seconds ))

# Subtract 10 minutes (600 seconds)
new_total_seconds=$((total_seconds - 600))

# Ensure time does not go below zero
if [[ $new_total_seconds -lt 0 ]]; then
    new_total_seconds=0
fi

# Convert to HH:MM:SS format for Parthenon
new_hours=$((new_total_seconds / 3600))
rem=$((new_total_seconds % 3600))
new_minutes=$((rem / 60))
new_seconds=$((rem % 60))
time_limit_lower=$(printf "%02d:%02d:%02d" "$new_hours" "$new_minutes" "$new_seconds")

echo
echo Time Limit: $orig_time_limit
echo Time Limit minus ten min: $time_limit_lower
echo
# $time_limit_lower is now suitable for:
# srun path/to/parthenon/code -t $time_limit_lower
```

Maybe there are ways to trap these signals in the batch script and write the flag file as discussed here? Maybe `#SBATCH --signal=B:USR1@600` and a suitable `trap` command.

Yurlungur:
@nbowden045 yes there should definitely be a way to do this, and this is the intent of what we currently support. I think @c-prather was finding that unfortunately the behavior wasn't consistent.

c-prather:
Opening an updated version.

PR #784, opened 2022-11-08
Fix an output bug

c-prather:
While looking at making output naming/production a little more flexible, I noticed an output bug: if the code would stop because the current step number is >=nlim (part of the KeepGoing() check), then an output may be produced twice on the last step.

This was likely only ever triggered when outputs were produced every single step during debugging, so I doubt anyone hit it in a way that mattered, and in any case the "workaround" to produce normal behavior would just be to delete the last numbered file before "final."

Anyway, I fixed it.  A wider reimagining of outputs will will take a while longer, so if it misses the next release we should at least not carry around a known bug with a one-line fix.

Ulterior motive: how small a PR is too small?  I can add the usual stuff to this PR, but I'd be happy if another pending output-related PR picked it up too -- at some point, writing too much about one line is a waste of time.  Maybe some protocol around "miscellaneous fixes" PRs to try to bundle things like this?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - [[PR 663]](https://github.com/lanl/parthenon/pull/663) Change bvals_in_one to use sparse boundary buffers and add flux_correction in one.
 
 ### Fixed (not changing behavior/API/variables/...)
+- [[PR 777]](https://github.com/parthenon-hpc-lab/parthenon/pull/784) Fix double-output of last file in rare cases 
 - [[PR 740]](https://github.com/lanl/parthenon/pull/740) Faster PHDF file load times in phdf.py
 - [[PR 751]](https://github.com/lanl/parthenon/pull/751) Delete useless file in advection example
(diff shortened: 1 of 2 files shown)


Yurlungur (approved):
Changelog?

Yurlungur:
> Ulterior motive: how small a PR is too small? I can add the usual stuff to this PR, but I'd be happy if another pending output-related PR picked it up too -- at some point, writing too much about one line is a waste of time. Maybe some protocol around "miscellaneous fixes" PRs to try to bundle things like this?

IMO it's fine either way. If you want to submit a tiny PR that's fine and if you want to submit a miscellaneous fixes PR that's fine too.

brryan (approved):
Nice fix, thanks for catching this. Someday I'm sure this would show up in a real calculation and it could be incredibly confusing.

c-prather:
I believe I've seen it. Definitely only in debug stuff, though, it (and regex being hard, and the fact they were sometimes stale...) led to me always ignoring the 'final' files when a sim ran to the end.

Re: PRs: alright, if this is just a way to do things, screw it.  Changelog updated.

PR #795, opened 2022-12-01
Fix length-1 vectors in output format version >= 3

brryan:
## PR Summary

With parthenon output format version 3 introduced in #781, scalars are now output as scalars instead of length 1 vectors, drawing a distinction between scalars and length 1 vectors. Previously, `phdf` would identify length-1 vectors as scalars and reshape such arrays in the `Get` call. This breaks length-1 vectors that we sometimes have downstream; I missed fixing this when working on #781. This small PR fixes this issue.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


brryan:
@par-hermes format

brryan:
This was actually a bug because in output format version 3 `vLen` was being set to `vShape[1]`, which would be either a length of a component index or `Nx3`.  And then if `vLen` was 1 the array would be reshaped. BUT if `vLen = Nx3 = 1`, the array would just be reshaped to `nblocks, nx3, nx2, nx1` i.e. it would be unchanged. Since this bug would only occur for length-1 vector data, I don't *think* this change should break either athenapk or kharma output

brryan:
@par-hermes format

c-prather:
The changes themselves look fine, though I don't output any length-1 vectors to test.

Is there a reason to use two `reshape()` calls here though, rather than `squeeze()`?

brryan:
> The changes themselves look fine, though I don't output any length-1 vectors to test.
> 
> Is there a reason to use two `reshape()` calls here though, rather than `squeeze()`?

Fortunately `output_hdf5` includes length-1 vectors (hence the need for an upgold). Also, I use length-1 vectors in phoebus and this works there.

The `reshape` code is from at least a while ago; in the interest of time I'd rather just leave it as is since I don't think there are performance or correctness issues with `reshape` vs `squeeze`.

c-prather (approved):

c-prather:
Fair enough. When someone gets around to it, `squeeze` will definitely use less memory and likely be faster -- maybe something for a Python cleanup PR later on.

Yurlungur (approved):
Seems fine to me but an AthenaPK developer should look at it

pgrete:
The pipeline failed: https://github.com/parthenon-hpc-lab/parthenon/actions/runs/3938681656/jobs/6737654886

@brryan looks like not all gold standard content is up to date

brryan:
@pgrete OK this should be working now, didn't require an upgold (only `phdf.py` changes anyway)

PR #796, opened 2022-12-01
At least partially fix XDMF output

brryan:
## PR Summary

@tagardiner noticed that #781 unfortunately broke XDMF output on e.g. the `advection` problem. It looks like this is due to the change in output format 2 -> 3 where scalars are no longer length-1 vectors in the HDF5 data, but rather have dimension `[nblock, nx3, nx2, nx1]`. This change in the HDF5 was not accompanied by any change in the XDMF.

Fixing the scalar case is straightforward I think, but it would be nice to also provide support for rank-2 and rank-3 parthenon output variables as well.

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted

### DIFF
--- a/src/outputs/parthenon_hdf5.cpp
+++ b/src/outputs/parthenon_hdf5.cpp
@@ -957,11 +957,6 @@ void PHDF5Output::WriteOutputFileImpl(Mesh *pm, ParameterInput *pin, SimTime *tm
     }
 
     // write data to file
-    printf("var: %s ndim: %i dims: ", var_name.c_str(), ndim);
-    for (int i = 0; i < ndim; i++) {
-      printf("%i ", local_count[i]);
-    }
-    printf("\n");
     HDF5WriteND(file, var_name, tmpData.data(), ndim, p_loc_offset, p_loc_cnt, p_glob_cnt,
                 pl_xfer, pl_dcreate);
   }
--- a/src/outputs/parthenon_hdf5.cpp
+++ b/src/outputs/parthenon_hdf5.cpp
@@ -316,7 +316,7 @@ static void writeXdmfSlabVariableRef(std::ofstream &fid, const std::string &name
     fid << ">" << std::endl;
     fid << prefix << "  "
         << R"(<DataItem ItemType="HyperSlab" Dimensions=")";
-    for (int i = 0; i < ndims; i++) {
+    for (int i = 1; i < ndims; i++) {
       fid << dims[i] << " ";
     }
     fid << R"(">)" << std::endl;
@@ -957,6 +957,11 @@ void PHDF5Output::WriteOutputFileImpl(Mesh *pm, ParameterInput *pin, SimTime *tm
     }
 
     // write data to file
+    printf("var: %s ndim: %i dims: ", var_name.c_str(), ndim);
+    for (int i = 0; i < ndim; i++) {
+      printf("%i ", local_count[i]);
+    }
+    printf("\n");
     HDF5WriteND(file, var_name, tmpData.data(), ndim, p_loc_offset, p_loc_cnt, p_glob_cnt,
                 pl_xfer, pl_dcreate);
   }
--- a/src/outputs/parthenon_hdf5.cpp
+++ b/src/outputs/parthenon_hdf5.cpp
@@ -316,7 +316,8 @@ static void writeXdmfSlabVariableRef(std::ofstream &fid, const std::string &name
     fid << ">" << std::endl;
     fid << prefix << "  "
         << R"(<DataItem ItemType="HyperSlab" Dimensions=")";
-    for (int i = 1; i < ndims; i++) {
+    fid << "1 ";
+    for (int i = 2; i < ndims; i++) {
       fid << dims[i] << " ";
     }
     fid << R"(">)" << std::endl;
@@ -957,6 +958,10 @@ void PHDF5Output::WriteOutputFileImpl(Mesh *pm, ParameterInput *pin, SimTime *tm
     }
 
     // write data to file
+    printf("var_name: %s ndim: %i vector? %i\n", var_name.c_str(), ndim, static_cast<int>(vinfo.is_vector));
+    for (int i = 0; i < ndim; i++) {
+      printf("  [%i]: %i\n", i, local_count[i]);
+    }
     HDF5WriteND(file, var_name, tmpData.data(), ndim, p_loc_offset, p_loc_cnt, p_glob_cnt,
                 pl_xfer, pl_dcreate);
   }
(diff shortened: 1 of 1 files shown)


tagardiner:
@brryan I cloned this branch and tried to open the advection file in visit.  In case it helps, I'm still finding an error from visit that looks like this:

HDF5-DIAG: Error detected in HDF5 (1.8.14) thread 0:
  #000: ../../src/H5Dio.c line 161 in H5Dread(): selection+offset not within extent
    major: Dataspace
    minor: Out of range
XDMF Error in /Users/miller86/visit/visit/33rc/release/build-mb-3.3.0-darwin-10.14-x86_64-release/thirdparty_shared/Xdmf/libsrc/XdmfValuesHDF.cxx line 109 (Can't Read Dataset advection.out0.00000.phdf:/Locations/z)
XDMF Error in /Users/miller86/visit/visit/33rc/release/build-mb-3.3.0-darwin-10.14-x86_64-release/thirdparty_shared/Xdmf/libsrc/XdmfDataItem.cxx line 531 (Reading Values Failed)
XDMF Error in /Users/miller86/visit/visit/33rc/release/build-mb-3.3.0-darwin-10.14-x86_64-release/thirdparty_shared/Xdmf/libsrc/XdmfValuesHDF.cxx line 89 (Return Array No Large Enough to Hold Selected Data)
 Alloc :: Cannot allocate memory

brryan:
@tagardiner Yes I don't think this was working yet (I should have marked this `Draft`). I'm unfamiliar with XDMF and was still flailing last night, but I think I just now figured out what the correct solution is, at least for the advection example. 

This current commit `d307e50` allows me to plot `advected` from the advection example in visit if you wouldn't mind trying this. It's possible other tensor ranks (`advected` is a one-element vector) still don't work correctly, I am still testing.

tagardiner:
@brryan Ok, so just in case it might matter based on types, etc. I'm on a mac.  Here's the version I'm building / running

commit d307e5034d13d86327f6f42d10337080792cd565 (HEAD -> brryan/xdmf_fix, origin/brryan/xdmf_fix)

I should also probably mention that I'm running the input file in

parthenon/example/advection/parthinput.advection

I'm finding that when I load the xdmf files in visit, there are 88 advected variables at the initial time, advected-1 through advected-88.  As I increment the time-step the variables change and the number of them change, for example advected-94 shows up. 

The mesh looks ok, but the data variable itself looks wacky and sometimes contains nans.  For example:

Advected

brryan:
@tagardiner Yes I am also seeing this problem (many `advected` variables) with the default advection example. For me actually (on Windows) the data doesn't plot at all, I just get the error I've been seeing
```
The compute engine running on fortran95.lanl.gov has exited abnormally.

Shortly thereafter, the following occured...

The Pseudocolor plot of "advected- 86" for the file "C:\Users\305862\Documents\visit_test\default\advection.out0.final.phdf.xdmf" could not be generated by the compute engine on host "localhost".
```

My guess (due to changing number of `advected` variables) is that I've mistakenly told visit to create a new variable for `advected` on each meshblock. I'm not quite sure how to fix this but I'm taking a look now.

brryan:
@tagardiner OK that was actually an easy fix, on `66c5018` here's what I get when I plot the one `advected` variable from the final output with the `parthenon/example/advection/parthinput.advection` input file, which looks right to me:
image
Sorry for the thrashing, please do let me know if you continue to encounter issues with `xdmf`. I'll look through this a bit more and then request reviewers and try to get it merged in.

tagardiner:
@brryan I think that did it - Thanks!

tagardiner:
@brryan I think we still have a bit of a problem.  Try running the burgers benchmark problem and it does the same thing.  That is, I'm seeing U_#-#, Ulx_#-#, Uly_#-#, etc.  FWIW I am running the input file burgers.pin in parthenon/benchmarks/burgers and setting the output time step to 1e-3.

brryan:
@tagardiner OK this time XDMF was creating one variable for each vector element inside of each vector element. I cleaned that up and now I get a set of visit variables that looks sensible to me (and they plot):
image
I only tested this with two small meshblocks because the original input parameter file produces a 1.6 GB output and my code execution and visit are separated by a VM but I think 2 meshblocks should be equivalent to any number of AMR meshblocks as far as `xdmf` is concerned.

A lot of the variables seem to be all zero (at least initially) when I plot them. I'm not familiar with `burgers` so I don't know if that is correct.

tagardiner:
@brryan From what I can tell the code seems to indicate that U is a velocity vector and some number of advection scalars.  In the default test case there are 8 advected scalars and 3 components to a velocity vector, so perhaps that is what the variables enumerated as (U_0, U_10) represent.  According to burgers_package.cpp the Ulx, Urx, ... are the reconstructed state on faces and there are 11 of each of these as well.  So it appears to me as well that you've got this corrected as well.

@jdolence is this interpretation of the burgers code correct?  Also, does this problem evolve in time - 'cause I'm not seeing much change over the first 10 time-steps?

src/outputs/parthenon_hdf5.cpp
```diff
@@ -406,15 +406,30 @@ void genXDMF(std::string hdfFile, Mesh *pm, SimTime *tm, int nx1, int nx2, int n
...
-    dims[2] = nx3;
-    dims[3] = nx2;
-    dims[4] = nx1;
+    int ndim;
     for (const auto &vinfo : var_list) {
+      // TODO(BRR) just let vinfo provide this
+      std::vector<hsize_t> alldims(
+          {static_cast<hsize_t>(vinfo.nx6), static_cast<hsize_t>(vinfo.nx5),
+           static_cast<hsize_t>(vinfo.nx4), static_cast<hsize_t>(vinfo.nx3),
+           static_cast<hsize_t>(vinfo.nx2), static_cast<hsize_t>(vinfo.nx1)});
```
Yurlungur: A better solution

Yurlungur (commented):
Overall this seems like a reasonable fix. Needs changelog.

pgrete:
ping @jdolence would you mind checking the results and review?

tagardiner:
LGTM

tagardiner (approved):
LGTM

BenWibking:
@brryan I see this warning from Paraview when reading the advection outputs with this PR. It shows it for every variable in every block:

```
Warning: In vtkDataSet.cxx, line 692
vtkRectilinearGrid (0x13e587440): Cell array Density with 1 components, has 1048576 tuples but there are only 32768 cells

Warning: In vtkDataSet.cxx, line 692
vtkRectilinearGrid (0x13e587440): Cell array Velocity1 with 1 components, has 1048576 tuples but there are only 32768 cells
```

The code referenced in the warning is here: https://github.com/Kitware/VTK/blob/master/Common/DataModel/vtkDataSet.cxx#L689.

I'm not sure where VTK is getting the number of tuples vs number of cells. The number of tuples is 32^4, whereas the number of cells is 32^3. Is there an extra multiplication happening somewhere while writing the metadata?

BenWibking:
Ok, the warning goes away if I change (for 32^3 Meshblocks):
```
        <DataItem ItemType="HyperSlab" Dimensions="1 32 32 32 ">
```
to
```
        <DataItem ItemType="HyperSlab" Dimensions="32 32 32 ">
```

Issue #808, opened 2023-01-06
Aggressive MPI hang detection?


c-prather:
Randomly when running KHARMA, but particularly for small block sizes relative to MPI latency, I often seem to trigger the following Parthenon error:

```
### PARTHENON ERROR
  Condition:   *nrecv_tries_ < 1e6
  Message:     MPI probably hanging after 1e6 receive tries.
  File:        /builds/AFD-Illinois/kharma/external/parthenon/src/utils/communication_buffer.hpp
  Line number: 326
MPI probably hanging after 1e6 receive tries.
Backtrace etc etc blah blah blah
```

When I comment the line that checks this, the code always seems to run fine.  However, as the error is not deterministic, I can't be absolutely sure it's *not* a hang in those cases.  Though fwiw, I have not encountered a true hang ever after disabling the check.

I definitely abuse the scatter/gather tooling from Parthenon (writing e.g. `while(my_reduction.CheckReduce() == TaskStatus::incomplete);`) but I don't believe that this receive routine is used in that tooling at all.  From what I can tell, this check is only for exchanging ghost zones, and the frequency of those checks is entirely controlled by Parthenon.

I'll continue to look at more reliably reproducing the error, and whether any of the situations causing this to fire are in fact hangs.  I'm posting this now in case anyone else has run into this, and to solicit any justification for the 1e6 number vs, e.g., 1e7 or 1e8, which seems likely still to translate to a few seconds/minutes of idle hang time in the worst case.  I understand that letting an MPI hang go unaddressed is probably inadvisable, as on large machines it would waste time, but it's looking like the detection should be slightly more lenient.

pgrete:
Same here. I can confirm that I've also experience this and increasing the counter fixed it without any other errors.
Given that the code only proceeds once all messages are successfully exchanged, seeing no errors is a strong indicator that this is not a bug but just a too stringent test.
So I'm in favor of increasing the counter to, say, 1e8. (as 1e6 typically happens effectively immediately so 1e7 may still be barely working and 1e8 should be safe to assume that comms failed).

lroberts36:
Yes, 1e6 was mostly a guess when I put it in. It seemed to work for the test suite problems, but I never tested it more than that. I think going to 1e8 is reasonable, but maybe it should be a user configurable parameter.

c-prather:
Merged #809 technically closes. User-specified and/or walltime-based hang criterion should be a separate issue

PR #809, opened 2023-01-09
Looser MPI hang detection: wait a few seconds to error

c-prather:
Addresses crash described in #808 in the crudest possible way.

## PR Summary

Some downstreams were seeing a Parthenon error due to some hang detection code in communication_buffer.hpp.  This is likely due to fast MPI checks vs. the network latency: even in normal circumstances, without an MPI hang, a rank could conceivably check 1e6 times for a message before receiving one.

This raises the hard-coded check limit to 1e8 tries, which still should not take very long to hit (judging by the rate to reach 1e6 checks, maybe a minute at the very most).

One alternative mentioned in the issue would be to make the limit user-configurable -- however, it's not immediately obvious to me how to implement this without an extra parameter in the CommBuffer constructor, which would require widespread changes and a couple of documentation updates.  Would love any pointers on how to plumb through the option if there's a better way to do so.

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted

### DIFF
--- a/src/utils/communication_buffer.hpp
+++ b/src/utils/communication_buffer.hpp
@@ -320,8 +320,8 @@ bool CommBuffer<T>::TryReceive() noexcept {
       *comm_type_ == BuffCommType::sparse_receiver) {
 #ifdef MPI_PARALLEL
     (*nrecv_tries_)++;
-    PARTHENON_REQUIRE(*nrecv_tries_ < 1e6,
-                      "MPI probably hanging after 1e6 receive tries.");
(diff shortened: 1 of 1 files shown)


lroberts36:
This looks fine to me. 

I *think* I would just add the user configurable limit as a global variable. A little bit hacky, but it gets around the problem @bprather points out. 

It may also be worth having a discussion if we want to stop on the number of calls or add a time to `CommBuffer`.

lroberts36 (approved):
LGTM

Yurlungur (approved):

pgrete (approved):

PR #810, opened 2023-01-11
Ascent support

### LINKED ISSUE #728: plans for Ascent support?
Hi,

Are there any plans to add support for using Ascent for in-situ visualization?

Thanks,
Ben

BenWibking:
## PR Summary

This adds basic support for the Ascent in-situ visualization framework (https://github.com/alpine-dav/ascent).

This version does not handle sparse variables.

Note: there is an unresolved issue with volume rendering of AMR datasets in Ascent (https://github.com/Alpine-DAV/ascent/issues/955). I haven't yet tested whether this affects octree AMR, or just patch-based/overlapping AMR.

Closes https://github.com/parthenon-hpc-lab/parthenon/issues/728.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


CMakeLists.txt
```diff
@@ -353,6 +353,8 @@ if (TEST_INTEL_OPTIMIZATION)
   add_compile_options(-fp-model fast=2 -qopt_report5 -vec-threshold0 -qopt_report_phase=vec)
 endif()
 
+find_package(Ascent REQUIRED NO_DEFAULT_PATH)
```
Yurlungur: Should this be an optional dependency?

pgrete:
@BenWibking thanks for the proof of concept.
We talked about this during the call last week.
Bottom line:
- we should get this in sooner than later (i.e., make it work and expand on flexibility later when and where required)
- Ascent should be an optional dependency
- No need to handle sparse variables for now (cf., "sooner than later")
- Ideally, this would be hooked into the existing output machinery (as it could inherit the triggering mechanisms) 

Are you interested in follow up on the making the required changes (when provided the right pointers) or should someone else take this over (I'm happy to take care of it early Feb when I'm back).

BenWibking:
> Are you interested in follow up on the making the required changes (when provided the right pointers) or should someone else take this over (I'm happy to take care of it early Feb when I'm back).

I can follow up on this. Can you provide pointers on how to i) loop over all of the variables and ii) where it should hook into the output machinery?

pgrete:
> > Are you interested in follow up on the making the required changes (when provided the right pointers) or should someone else take this over (I'm happy to take care of it early Feb when I'm back).
> 
> I can follow up on this. 

Great!

> Can you provide pointers on how to i) loop over all of the variables and 

You could use also obtain a `PackIndexMax` when packing all variables, cf.. https://github.com/parthenon-hpc-lab/parthenon/blob/8c387e04e8903a9205e125ec4c17625707850090/example/advection/parthenon_app_inputs.cpp#L59
which should provide you with all variable names and their index in the pack.

> ii) where it should hook into the output machinery?

You could add a new `AscentOutput` class derived from `OutputType`, cf. the files in the `src/output` directory (all output types are defined in the common `output.hpp` header) and then implement the Ascent logic in the `WriteOutputFile` function.
Then to trigger the output, you'd register a new "file_type", say "ascent" and create the `AscentOutput` object here https://github.com/parthenon-hpc-lab/parthenon/blob/develop/src/outputs/outputs.cpp#L286

I probably have missed sth but this should be the main steps.
In the long run I can also image to process additional parameter in the input file, e.g., the filename to an Ascent action yaml so that the action can be controlled on a "per output" basis (which could then be triggered by different condition). Or we don't need that machinery and just trigger the action via the Ascent yaml file.

BenWibking:
@pgrete has agreed to take this PR from here. thanks!

pgrete:
@BenWibking looks like you just updated.
Are you working on this? (just to sync up as I'm currently working on it)

BenWibking:
> @BenWibking looks like you just updated. Are you working on this? (just to sync up as I'm currently working on it)

Ah, sorry, no. I just saw it was out of date. I'll avoid doing this and let you take over :)

src/outputs/ascent.cpp
```diff
@@ -0,0 +1,199 @@
...
+  // reference:
+  // https://llnl-conduit.readthedocs.io/en/latest/blueprint_mesh.html#complete-uniform-example
+
+  using conduit::Node;
+
+  // Ascent needs the MPI communicator we are using
+  ascent::Ascent a;
+  Node ascent_opts;
+  ascent_opts["mpi_comm"] = MPI_Comm_c2f(MPI_COMM_WORLD);
+  a.open(ascent_opts);
```
BenWibking: This is probably not the main cause of the performance slowness, but we could move this to a higher-level scope and initialize at the start of a simulation, rather than re-creating the `ascent::Ascent` object and passing the MPI communicator each time. Alternatively, since it's only used in this function, we could make this a static variable and initialize it only upon first use.

src/outputs/ascent.cpp
```diff
@@ -0,0 +1,199 @@
...
+  scenes["s1/plots/p1/field"] = "advected:Advected_0_0";
+
+  // Set the output file name (ascent will add ".png")
+  scenes["s1/image_prefix"] = "ascent_render";
+
+  // execute the actions
+  a.execute(actions);
+
+  // close ascent
+  a.close();
```
BenWibking: If we initialize at simulation start, this would also have to be called only at the end of a simulation.

src/outputs/ascent.cpp
pgrete: Ah, great! Thanks for the input.
I was actually wondering about that but didn't find an info in their doc on whether it's fine to have a persistent object or if the model is "sim pause -> ascent start from scratch -> ascent data -> ascent done and cleanup -> sim resume".
Do you know if we could also reuse the nodes, i.e., caching them as long as the mesh structure didn't change?

BenWibking: Yes, the `ascent` object itself can definitely be re-used. I think the nodes *should* be able to be re-used as long as the mesh doesn't change, since the actual data is passed via `set_external`, but I haven't tested it.

pgrete:
Alright, this is ready for review.

The total amount of changed code is not that big (as half of the lines are for the docker image setup and adapted from the Ascent repo).

Our updated pipeline now uses the advection example to plot this:

ascent_render_57

Also, the container for testing (with a Ascent install with Cuda support) is also publicly available here: https://github.com/orgs/parthenon-hpc-lab/packages/container/package/cuda11.6-mpi-hdf5-ascent

src/outputs/ascent.cpp
```diff
@@ -0,0 +1,203 @@
...
+//----------------------------------------------------------------------------------------
+//! \fn void AscentOutput:::WriteOutputFile(Mesh *pm)
+//  \brief  Expose mesh and all Cell variables for processing with Ascent
+void AscentOutput::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm,
+                                   const SignalHandler::OutputSignal signal) {
+#ifndef PARTHENON_ENABLE_ASCENT
+  if (Globals::my_rank == 0) {
+    PARTHENON_WARN("Ascent output requested by input file, but Ascent support not "
+                   "compiled in. Skipping this output type.");
+  }
```
Yurlungur, line 55: Do we want this printed every ascent output? I guess that's not unreasonable... But I think I might prefer a quieter error message.

src/outputs/ascent.cpp
```diff
@@ -0,0 +1,203 @@
...
+  // reference:
+  // https://llnl-conduit.readthedocs.io/en/latest/blueprint_mesh.html#complete-uniform-example
+
+  using conduit::Node;
+
+  // Ascent needs the MPI communicator we are using
+  ascent::Ascent ascent;
+  Node ascent_opts;
+  ascent_opts["mpi_comm"] = MPI_Comm_c2f(MPI_COMM_WORLD);
+  ascent_opts["actions_file"] = pin->GetString(output_params.block_name, "actions_file");
```
Yurlungur, line 66: Should this just fail? Or should we print an error message explaining what to do if this file isn't present?

src/outputs/ascent.cpp
```diff
@@ -0,0 +1,203 @@
...
+          if ((i < ib_int.s) || (ib_int.e < i) || (j < jb_int.s) || (jb_int.e < j) ||
+              ((nz > 1) && ((k < kb_int.s) || (kb_int.e < k)))) {
+            vals_array[idx] = 1;
+          } else {
+            vals_array[idx] = 0;
+          }
+          idx++;
+        }
+      }
+    }
```
Yurlungur: For large meshblocks doing this might be slow... I wonder if we should support doing this on device in the limit of large meshblocks, like NX1 ~ NX2 ~ NX3 ~ 100?

(No need to do this now. I'm just thinking out loud.)

src/outputs/ascent.cpp
```diff
@@ -0,0 +1,203 @@
...
+      auto const &labels = var->metadata().getComponentLabels();
+
+      for (int icomp = 0; icomp < labels.size(); ++icomp) {
+        auto const data = Kokkos::subview(var->data, 0, 0, icomp, Kokkos::ALL(),
+                                          Kokkos::ALL(), Kokkos::ALL());
+        const std::string varname = packname + ":" + labels.at(icomp);
+        mesh["fields/" + varname + "/association"] = "element";
+        mesh["fields/" + varname + "/topology"] = "topo";
+        mesh["fields/" + varname + "/values"].set_external(data.data(), ncells);
+      }
```
Yurlungur: So if I understand this correctly, it also means that for vector fields, we need to assign labels?

src/outputs/ascent.cpp
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    // create a field for each component of each variable pack
+    auto &mbd = pmb->meshblock_data.Get();
+
+    for (const auto &var : mbd->GetCellVariableVector()) {
+      PARTHENON_REQUIRE(var->GetDim(6) * var->GetDim(5) == 1,
+                        "Only scalar and vector fields are currently exported in Ascent. "
+                        "Please open an issue on GitHub if you have a use case.");
```
Yurlungur: Is there a deep reason for this? I haven't tried to ascent, but I don't see a reason not to at least simply flatten tensor fields and treat each component as a scalar? That would make it at least *possible* to visualize tonsorial quantities in a clunky way?

src/outputs/ascent.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+      const std::string packname = var->label();
+      auto const &labels = var->metadata().getComponentLabels();
+
+      for (int icomp = 0; icomp < labels.size(); ++icomp) {
+        auto const data = Kokkos::subview(var->data, 0, 0, icomp, Kokkos::ALL(),
+                                          Kokkos::ALL(), Kokkos::ALL());
+        const std::string varname = packname + ":" + labels.at(icomp);
+        mesh["fields/" + varname + "/association"] = "element";
```
Yurlungur, line 175: Assume this will have to change for face-centered fields?

src/outputs/ascent.cpp
```diff
@@ -0,0 +1,203 @@
...
+    auto kb_int = bounds.GetBoundsK(IndexDomain::interior);
+
+    auto &coords = pmb->coords;
+    Real dx1 = coords.CellWidth<X1DIR>(ib.s, jb.s, kb.s);
+    Real dx2 = coords.CellWidth<X2DIR>(ib.s, jb.s, kb.s);
+    Real dx3 = coords.CellWidth<X3DIR>(ib.s, jb.s, kb.s);
+    std::array<Real, 3> corner = coords.GetXmin();
+
+    // create the coordinate set
+    mesh["coordsets/coords/type"] = "uniform";
```
Yurlungur, line 110: Assume this will have to change for non-Cartesian coords?

src/outputs/ascent.cpp
```diff
@@ -0,0 +1,203 @@
...
+          }
+          idx++;
+        }
+      }
+    }
+
+    // create a field for each component of each variable pack
+    auto &mbd = pmb->meshblock_data.Get();
+
+    for (const auto &var : mbd->GetCellVariableVector()) {
```
Yurlungur, line 168: It looks like we loop over all vars here, but I guess we just assign metadata so this is probably not a big cost?

Yurlungur (approved):
I haven't tried to use it as I don't have easy access to ascent---but this is cool. I'm pleasantly surprised how easy this is on the parthenon side. I have a few comments below, but they're all non-blocking... just me thinking out loud.

CMakeLists.txt
pgrete: Was a previous version of the branch shown when you reviewed?
That line should be encapsulated in an 
```
if (PARTHENON_ENABLE_ASCENT)
  find_package(Ascent REQUIRED NO_DEFAULT_PATH)
endif()
```
with `PARTHENON_ENABLE_ASCENT` being off by default.

src/outputs/ascent.cpp
pgrete: On second thought, how about a loud fail?
I'm coming from the point of view that if you expect an output it should be there,not that you run a sim and wonder where the files are after burning X hours of computing time.

src/outputs/ascent.cpp
pgrete: Good point.
Now that you mention it I think there's an even better solution.
Given that our meshblock size is fixed, I can just create one object tied to the class and then reuse every time the function is called.
I'll update this.

src/outputs/ascent.cpp
pgrete: Our default error message looks like
```
### PARTHENON ERROR
  Message:     ### FATAL ERROR in function [ParameterInput::GetString]
Parameter name 'actions_file' not found in block 'parthenon/output4'
  File:        /home/pgrete/src/parthenon/src/parameter_input.cpp
  Line number: 629
```
From my point of view that contains enough info to figure out what is wrong.
What do you think?

src/outputs/ascent.cpp
pgrete: Good point. I'll replicate/reuse the logic we use in the hdf5 outputs, i.e.,
```
    component_labels = {};
    if (vlen == 1 || is_vector) {
      component_labels = component_labels_.size() > 0 ? component_labels_
                                                      : std::vector<std::string>({label});
    } else if (component_labels_.size() == vlen) {
      component_labels = component_labels_;
    } else {
      for (int i = 0; i < vlen; i++) {
        component_labels.push_back(label + "_" + std::to_string(i));
      }
```

src/outputs/ascent.cpp
pgrete: The main reason is that (similar to prolong/restrict or outputs) that it'd be an implementation that "should" be correct without a proper test case.
Therefore, I opted for a loud fail so that once we have a use (and test case), this can be easily spotted an easily fixed. Otherwise, it could be another silent place where the code doesn't fail but also doesn't work correctly (because it's not properly tested).
Does this sound reasonable?

src/outputs/ascent.cpp
pgrete: Probably more would have to change for face fields.
We could like still use element and Cartesian geometry,but would need to adjust  the locations and element indices, i.e., create a new `topology` and then assign the face fields with that updated topology.

Nevertheless, if you agree with the "loud fail" option I suggested above for tensor variables, I'd add a loud fail here so that we remember to fix this once we have face fields.

src/outputs/ascent.cpp
pgrete: As above, if you agree I'd add a loud fail here, too.

src/outputs/ascent.cpp
pgrete: Correct, I'd be surprised if this is expensive as we just forward "external" pointers here.

CMakeLists.txt
Yurlungur: Ah weird. Very good.

src/outputs/ascent.cpp
Yurlungur: I find loud failure for outputs in particular annoying since I often copy around input decks and it's irritating to have to comment out blocks to get the code to run. But I don't feel too strongly either way.

src/outputs/ascent.cpp
Yurlungur: Yeah I think that's fine. I couldn't remember how verbose that particular error was.

src/outputs/ascent.cpp
Yurlungur: :+1: Nice solution.

src/outputs/ascent.cpp
Yurlungur: :+1:

src/outputs/ascent.cpp
Yurlungur: I think the prolong/restrict and outputs with tensors has actually bitten us, as we've added many iterative ad hoc fixes before settling. So I don't see a reason to wait.

The use-case I had in mind was, e.g., metric or radiation quantities in Phoebus. I don't think it would be too cumbersome to throw in a derived field in advection or something with tensorial shape that gets flattened by ascent. But seeing as I haven't tried ascent with Phoebus, it doesn't need to be prioritized right now.

Actually I just realized something: This will cause phoebus to crash if compiled with ascent support because it checks EVERY variable, not just the ones registered with ascent. So at the *very* least this should *skip* tensorial fields, rather than dying.

src/outputs/ascent.cpp
Yurlungur: See above comment: Rather than loud failing we should warn and skip. Otherwise the code will be inoperable with ascent and some variables.

You should also probably check that the variable shape is the same as the meshblock, as it might not be and that would also cause problems. (For example axisymmetric variables.)

src/outputs/ascent.cpp
Yurlungur: This one can can be a loud fail.

Yurlungur:
> Also, the container for testing (with a Ascent install with Cuda support) is also publicly available here: https://github.com/orgs/parthenon-hpc-lab/packages/container/package/cuda11.6-mpi-hdf5-ascent

It just occurred to me. Should we advertise containers with dependencies installed in the docs somewhere? Maybe also on the README since it's the "front page" of the repo?

src/outputs/ascent.cpp
Yurlungur: :+1:

src/outputs/ascent.cpp
BenWibking: I don't think we can pass face-centered fields to Ascent directly. Need to check, but I think `element` is always cell-centered.

Yurlungur: Oh interesting. Well for now the solution is the same. Skip them. :)

src/outputs/ascent.cpp
```diff
@@ -0,0 +1,203 @@
...
+  // Ascent needs the MPI communicator we are using
+  ascent::Ascent ascent;
+  Node ascent_opts;
+  ascent_opts["mpi_comm"] = MPI_Comm_c2f(MPI_COMM_WORLD);
+  ascent_opts["actions_file"] = pin->GetString(output_params.block_name, "actions_file");
+  // Only publish fields that are used within actions to reduce memory footprint.
+  // A user might need to override this, e.g., in a runtime ascent_options.yaml, if
+  // the required fields cannot be resolved by Ascent.
+  // See https://ascent.readthedocs.io/en/latest/AscentAPI.html#field-filtering
+  // ascent_opts["field_filtering"] = "true";
```
BenWibking, line 74: This is supposed to work with the latest ascent develop. I plan to test this out tomorrow.

example/advection/parthinput.advection
```diff
@@ -80,3 +80,8 @@ variables = advected, advected_1, & # comments are ok
 <parthenon/output3>
 file_type = hst
 dt = 0.05
+
+<parthenon/output4>
+file_type = ascent
+dt = -0.05   # soft disabled by default, as Ascent is an optional dependency
+actions_file = custom_ascent_actions.yaml
```
BenWibking, line 87: This is currently relative to the working directory of the executable, but we might want to make this path relative to the input file?

BenWibking:
~~@pgrete I've tried this with AthenaPK, but on GPU runs (only) I am getting:~~
```
s1/p1 pseudocolor plot yielded no data, i.e., no cells remain
```
~~This can be reproduced with `precipitator.in` from https://github.com/parthenon-hpc-lab/athenapk/pull/13.~~

Update: this may be an unrelated problem-specific issue, since the variable data in the HDF5 outputs is also different on GPU runs.

src/outputs/ascent.cpp
pgrete: I see your point.
I left it as it for now (as it doesn't add  new logic/idefs at other places).

Yurlungur: :+1: yeah that's fine

example/advection/parthinput.advection
pgrete: The path currently follows the same pattern `pin->GetString(output_params.block_name, "actions_file");` we use for other external files, too.
I'm not sure I want to touch this here. Might be sth to discuss as a more general case.
Anyway, full paths should always work as is.

pgrete:
> > Also, the container for testing (with a Ascent install with Cuda support) is also publicly available here: https://github.com/orgs/parthenon-hpc-lab/packages/container/package/cuda11.6-mpi-hdf5-ascent
> 
> It just occurred to me. Should we advertise containers with dependencies installed in the docs somewhere? Maybe also on the README since it's the "front page" of the repo?

Good point. Added to #836

src/outputs/ascent.cpp
pgrete: I now centralized the `VarInfo` struct to `outputs.#pp` so that it could be reused here for the labels.

src/outputs/ascent.cpp
pgrete: This restriction is now removed (by using component labels for a flat index).

src/outputs/ascent.cpp
pgrete: I'll leave this as is for now (no warning etc) assuming that the output with Ascent will be obviously wrong if someone is going to use it for non-cell-centered fields.

src/outputs/ascent.cpp
pgrete: done

pgrete:
Alright, I think I got all comments addressed now.

src/outputs/ascent.cpp
Yurlungur: Just so long as ascent doesn't crash or something. We should maybe check that that's the case?

Yurlungur (approved):
LGTM

src/outputs/ascent.cpp
pgrete: I'm at the moment not too concerned about this, given that I'm fairly sure that we'll be the first users of those features once we've implemented them.
And so far if Ascent crashed the backtrace always pointed out to an Ascent function, i.e., straightforward to track down.

pgrete:
@forrestglines any chances for a review soonish (or are you busy and we should ask someone else)?

doc/sphinx/src/outputs.rst
```diff
@@ -140,6 +140,44 @@ This will produce a text file (``.hst``) output file every 1 units of
...
+A ``<parthenon/output*>`` block might look like::
+
+  <parthenon/output9>
+  file_type = ascent
+  dt = 1.0
+  actions_files = my_actions.yaml
+
+see also the advection example
+`input file <https://github.com/parthenon-hpc-lab/parthenon/blob/ascent-support/example/advection/parthinput.advection>`__ and
+`actions file <https://github.com/parthenon-hpc-lab/parthenon/blob/ascent-support/example/advection/custom_ascent_actions.yaml>`__.
```
forrestglines: These links will work once the PR is merged in, right?

src/outputs/ascent.cpp
forrestglines: Did you get to this test?

If it would take too long, you could add a `// TODO(BenWibking)` to save this for a future PR

src/outputs/ascent.cpp
forrestglines: That will be unfortunate, although not the end of the world if we can't output face-cenered fields, looking at face-centered magnetic fields (and edge centered EMFs, vertex E-fields). Interpolating those things to cell-centers is fine for analysis, the face-centered values seem to be only necessary for debugging problems.

src/outputs/outputs.hpp
```diff
@@ -35,6 +36,36 @@ namespace parthenon {
 class Mesh;
 class ParameterInput;
 
+// Helper struct containing some information about a variable
+struct VarInfo {
+  std::string label;
+  int num_components;
```
forrestglines, line 42: So `vlen` has been renamed to `num_components`, correct? Have the non-AthenaPK developers double checked that `num_components` is being used consistently since AthenaPK uses components differently?

I don't see a functional difference, just that `num_components` is a better name in my opinion. From the AthenaPK perspective, this looks to work as I'd expect.

.github/workflows/ci-extended.yml
```diff
@@ -66,12 +66,31 @@ jobs:
...
+            -DPARTHENON_ENABLE_ASCENT=ON \
+            -DAscent_DIR=/usr/local/ascent-develop/lib/cmake/ascent
+          cmake --build build-ascent
+          cd example/advection/
+          mpirun -np 2 ../../build-ascent/example/advection/advection-example \
+            -i parthinput.advection \
+            parthenon/output4/dt=0.05 \
+            parthenon/time/tlim=0.1
+          # check if file exists
+          test -f ascent_render_57.png
```
forrestglines, line 85: So we just check that Ascent works? Probably not wise to do a bitwise check that it matches?

Can Ascent produce other data products that can be compared bitwise, would that be a more robust test? Don't hold up the PR to implement that check though.

forrestglines (approved):
Looks good to me, please merge soon!

.github/workflows/ci-extended.yml
BenWibking: We could add something like idiff (https://openimageio.readthedocs.io/en/latest/idiff.html) or perceptualdiff (https://pdiff.sourceforge.net/) to compare against a reference image output?

doc/sphinx/src/outputs.rst
pgrete: Good point, they should point to `develop` rather than `ascent-support`. I'll update.

src/outputs/ascent.cpp
pgrete: I tested it, but decided to not enable it (for now) until the version with the fix are more broadly available (and I didn't want to introduce specific checks and logic for various Ascent versions)

src/outputs/ascent.cpp
BenWibking: Ascent natively supports vertex-centered fields: https://ascent.readthedocs.io/en/latest/Tutorial_Intro_Conduit_Blueprint.html?highlight=vertex#creating-a-uniform-grid-with-a-single-field. Just not face-centered ones.

src/outputs/outputs.hpp
pgrete: Yes, `vlen` has been a leftover from times where we logically handled everything in the output as a 5-D array (bock id, flattened components, Z, Y, X).
Now that non-vectored variables have custom shapes, `vlen` does not make sense anymore so I decided to rename it given that I was already touching the code.

.github/workflows/ci-extended.yml
pgrete: Yes, I'd be happy to have sth like this (also for other tests), but I thought in the first round, this is beyond this PR (as it might require additional [unrelated] changes in the test infrastructure as well)

BenWibking: Ok, fine with me.

src/outputs/ascent.cpp
Yurlungur: face-centered fields are just funny vertex centered fields really ;)

pgrete:
I had to revert to a HostArray for the ghost field as I ran into segfaults on device otherwise.
Already asked the Ascent devs if this is desired/expected behavior.

BenWibking:
> I had to revert to a HostArray for the ghost field as I ran into segfaults on device otherwise. Already asked the Ascent devs if this is desired/expected behavior.

The AMReX implementation also allocates this on the host, so I suspect this may not have been tested.

pgrete:
All tests passed. I'm force merging now as the formatting check doesn't trigger/work on forks but I checked locally:
```
$ ninja format
[0/2] Re-checking globbed directories...
[1/1] cd /home/pgrete/src/parthenon/build-vscode && /usr/bin/clang-format -i /home/pgrete/src...c/parthenon/tst/regression/utils/test_case.py /home/pgrete/src/parthenon/tst/style/cpplint.py
All done! ✨ 🍰 ✨
41 files left unchanged.
```

PR #818, opened 2023-01-25
Bug Fix: Boundary prolongation only occurring on base stage

lroberts36:
## PR Summary

This PR fixes a bug in the `TaskStatus ProlongateBoundaries(std::shared_ptr<MeshBlockData<Real>> &rc)` logic that caused prolongation to occur on the `base` stage `MeshBlockData` container of a given block, independent of which `MeshBlockData` stage was passed to `ProlongateBoundaries`. Because `BoundaryValues::ProlongateBoundaries()` was eventually called from the `BoundaryValues` object owned by the parent `MeshBlock` of the `MeshBlockData`, no information was passed through about which `MeshBlockData` associated with a given `MeshBlock` was asking for the boundary prolongation to occur. This PR also refactors the boundary prolongation code a bit so that there is not a hard to follow chain of calls (which is probably why this bug has persisted). 

This fix of course causes regression tests that include mesh refinement to fail. If I switch the line 
```c++
boundary_cond_impl::ProlongateGhostCells_(rc, nb, bi.s, bi.e, bj.s, bj.e, bk.s, bk.e);
``` 
to 
```c++
boundary_cond_impl::ProlongateGhostCells_(pmb->meshblock_data.Get(), 
    nb, bi.s, bi.e, bj.s, bj.e, bk.s, bk.e);
``` 
in `parthenon::ProlongateBoundaries(std::shared_ptr<MeshBlockData<Real>> &rc)` all of the regression tests pass on v13. I have added a new set of updated gold files as v15 with the bug fixed. 

Thanks to @pdmullen for finding that some ghost zone values were unset during intermediate RK stages which pointed to this issue. 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


pdmullen (approved):
Looks good to me! I can confirm that @lroberts36's proposed fix solves issues I was encountering with ghost cell filling with mesh refinement.  

The downstream codes I am working with point to here: https://github.com/parthenon-hpc-lab/parthenon/commit/f7f4e17c36e8125b7d3f31fcdc6e20de4ba75d30 , but as I hope to be evidenced by regression testing, I trust this iteration of the fix equally addresses the issue.

lroberts36:
@pdmullen: I will change the fix on the Riot parthenon branch to be the same as this just to double check.

pdmullen:
> @pdmullen: I will change the fix on the Riot parthenon branch to be the same as this just to double check.

Sounds great; I'll keep an eye out for it.  I'll rerun my tests and report back here.

pdmullen:
@lroberts36 mimicked the changes in this PR in the [lroberts36/merge-sparse-with-jdolence-sparse](https://github.com/parthenon-hpc-lab/parthenon/tree/lroberts36/merge-sparse-with-jdolence-sparse) branch which our downstream code points to.  I can confirm that

```cpp
boundary_cond_impl::ProlongateGhostCells_(rc, nb, bi.s, bi.e, bj.s, bj.e, bk.s, bk.e);
```

vs

```cpp
boundary_cond_impl::ProlongateGhostCells_(pmb->meshblock_data.Get(), 
    nb, bi.s, bi.e, bj.s, bj.e, bk.s, bk.e);
```
    
makes the difference between a passing and failing test, respectively.  From my perspective, this PR is good to go.

c-prather (approved):
If Patrick vouches *that* it fixes the issue, I for one am happy with *how* it does.  Always love a net-negative line count.

Yurlungur (approved):
Nice catch! I wish I'd caught this when I was mucking about in prolongation/restriction. 

To update the gold files, run the tests locally, then zip up the files and make a new release. Add the zip files to the release, and update the version and hash in the cmake.

Yurlungur:
> To update the gold files, run the tests locally, then zip up the files and make a new release. Add the zip files to the release, and update the version and hash in the cmake.

Maybe @pgrete has some magic to automate this.

lroberts36:
I made the release that updates the gold files for this bug fix a pre-release. Once this is merged, I guess it should be switched to being the latest release.

pgrete:
> > To update the gold files, run the tests locally, then zip up the files and make a new release. Add the zip files to the release, and update the version and hash in the cmake.
> 
> Maybe @pgrete has some magic to automate this.

No magic here but that reminds me that we should probably document those steps (and also how to make a release) -- opened an issue for this https://github.com/parthenon-hpc-lab/parthenon/issues/823 .

tst/regression/test_suites/sparse_advection/sparse_advection.py
```diff
@@ -57,7 +57,7 @@ def Analyse(self, parameters):
                 + "/tst/regression/gold_standard/sparse_fake.out0.final.phdf",
             ],
             one=True,
-            tol=1e-6,
+            tol=2e-6,
```
pgrete, line 60: Why is an increase in tolerance required?

src/bvals/boundary_conditions.cpp
```diff
@@ -198,6 +214,56 @@ bool DoPhysicalBoundary_(const BoundaryFlag flag, const BoundaryFace face,
...
+  auto &pmr = pmb->pmr;
+
+  for (auto cc_var : rc->GetCellVariableVector()) {
+    if (!cc_var->IsAllocated()) continue;
+    if (!(cc_var->IsSet(Metadata::Independent) || cc_var->IsSet(Metadata::FillGhost)))
+      continue;
+
+    // TODO(LFR): Is this indexing correct for 5 and 6 dimensional fields?
+    int nu = cc_var->GetDim(4) - 1;
+    pmr->ProlongateCellCenteredValues(cc_var.get(), 0, nu, si, ei, sj, ej, sk, ek);
```
pgrete: I think we are, because (if I'm not missing sth) `ProlongateCellCenteredValues` does calculate the other dims:
```
  info_h(b).Nt = var->GetDim(6);
  info_h(b).Nu = var->GetDim(5);
  info_h(b).Nv = var->GetDim(4);
```
It actually looks like ``0, nu` are not even used any more.
How about removing those to clean this up?

pgrete (approved):
LGTM I just have one cleanup suggestion and one clarification request (both non-blocking)

tst/regression/test_suites/sparse_advection/sparse_advection.py
lroberts36: Because it fails with `1.e-6` ;). 

This test is comparing the sparse solution to the same problem run with fields allocated everywhere. I think there is no expectation that you will agree at a level better than the sparse threshold (1.e-6), just of order the sparse threshold. I didn't see how close to 1.e-6 I could get, it may be that the tolerance could be made to be 1.1e6 or something.

src/bvals/boundary_conditions.cpp
lroberts36: Good point, I have removed those extra calls. @Yurlungur can you confirm that this will not cause any issues?

Yurlungur: Correct. When I updated those calls, I moved to pulling this information directly from the variable. The `0` and `nu` were left in to keep the API fixed, but we could remove them from `ProlongateCellCenteredValues` and probably should.

PR #829, opened 2023-02-05
More IO cleanup

Yurlungur:
## PR Summary

In preparation from doing Particle I/O I decided to remove some dead code from the I/O machinery:
 
- I removed the "tab" format entirely, which represented ascii tables. I also removed the old history/table logic such as "sum" and "slice." We may want this functionality at some point, but the data model has changed so much that I think we probably just want to start from scratch.
- One piece of machinery that may need to be re-introduced, but probably in a different place in the code base, is the transformation operation from curvilinear coordinates to Cartesian. For vis, downstream codes will need to know how to do that. But I don't think the "weakly linked" function stub that was present here was the right place for it.
- I left the VTK stub alone as we may want that and the old logic is probably still a useful reference.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


src/outputs/outputs.cpp
```diff
@@ -27,21 +27,6 @@
...
-//
-// EXAMPLE of an <output[n]> block for a VTK dump:
-//   <output3>
-//   file_type   = tab       # Tabular data dump
-//   variable    = prim      # variables to be output
-//   data_format = %12.5e    # Optional data format string
-//   dt          = 0.01      # time increment between outputs
-//   x2_slice    = 0.0       # slice in x2
-//   x3_slice    = 0.0       # slice in x3
-//
```
Yurlungur, line 44: This documentation is not currently correct. We also now document this kind of information in the markdown, etc.

src/outputs/outputs.cpp
```diff
@@ -157,78 +133,14 @@ Outputs::Outputs(Mesh *pm, ParameterInput *pin, SimTime *tm) {
...
       // read ghost cell option
       op.include_ghost_zones = pin->GetOrAddBoolean(op.block_name, "ghost_zones", false);
 
       // read cartesian mapping option
       op.cartesian_vector = false;
 
       // read single precision output option
-      const bool is_hdf5_output =
-          (op.file_type == "rst") || (op.file_type == "ath5") || (op.file_type == "hdf5");
+      const bool is_hdf5_output = (op.file_type == "rst") || (op.file_type == "hdf5");
```
Yurlungur, line 143: `ath5` doesn't exist anywhere else in the code. We should either transform it to `hdf5` or just remove it.

src/outputs/outputs.cpp
```diff
@@ -465,228 +375,4 @@ void Outputs::MakeOutputs(Mesh *pm, ParameterInput *pin, SimTime *tm,
   Kokkos::Profiling::popRegion(); // MakeOutputs
 }
 
-//----------------------------------------------------------------------------------------
-//! \fn void OutputType::TransformOutputData(MeshBlock *pmb)
-//  \brief Calls sum and slice functions on each direction in turn, in order to allow
-//  mulitple operations performed on the same data set
-
-bool OutputType::TransformOutputData(MeshBlock *pmb) {
```
Yurlungur, line 473: This is machinery we may want to re-enable at some point, but I'm not convinced here is the place and this is the way. Some food for thought:
- In an AMR hierarchy, is slicing or summing a meaningful operation?
- What about interpolating to a reduced-dimensionality separate grid?
- Should this be machinery the downstream code can register in some way to specify, e.g., specific integration weights?
- Should we be able to do this for all formats? Only HDF5? Only ASCII?

src/outputs/outputs.hpp
```diff
@@ -46,26 +46,19 @@ struct OutputParameters {
   int file_number_width;
   bool file_label_final;
   std::string file_id;
-  std::string variable;
```
Yurlungur, line 49: As far as I can tel this is completely unused now. It was used in the VTK output but I think it shouldn't be as that's the wrong scope with our data model.

Yurlungur:
I will be building particle I/O on top of this PR.

pdmullen:
Why are we eliminating `*.tab` files? These are very useful for 1D test problems.

src/outputs/outputs.cpp
pgrete: I'm in favor of removing. I've never seen it used anywhere.

src/outputs/outputs.cpp
pgrete: > This is machinery we may want to re-enable at some point, but I'm not convinced here is the place and this is the way. Some food for thought:

It may be worth to revive that discussion after people got the chance to experiment with Ascent for a bit.
As far as I can tell all that machinery will be automatically in place (though I don't know how "fast" it'll be and also how where the emphasis should be between duplicated, but internal code versus an external potentially not easy to install external dependency).
 
>     * In an AMR hierarchy, is slicing or summing a meaningful operation?

Definitely. (I know we'll need it)

>     * What about interpolating to a reduced-dimensionality separate grid?

Likely useful, but not sure it's actually required internally as the plotting/postprocessing (like interpolation) can easily be handled by external tools
 
>     * Should this be machinery the downstream code can register in some way to specify, e.g., specific integration weights?

Might also be useful. Another solution could be to allow for full function callsbacks (similar to the hst outputs).

>     * Should we be able to do this for all formats? Only HDF5? Only ASCII?

I'm happy with plain hdf5 and providing easy python script to read/process/plot/convert the data based on the assumption that this will cover the use cases of a big majority of the users.

pgrete (approved):

pgrete:
> Why are we eliminating `*.tab` files? These are very useful for 1D test problems.

They never worked in Parthenon so this was literally dead code.
Similar to the vtk output (which is also not working), we discussed that we're happy to reintroduce it (and then properly adapted to the current code base and features) when there's a need. So far there' haven't been any loud voices for these kind of outputs.

pdmullen:
> > Why are we eliminating `*.tab` files? These are very useful for 1D test problems.
> 
> They never worked in Parthenon so this was literally dead code. Similar to the vtk output (which is also not working), we discussed that we're happy to reintroduce it (and then properly adapted to the current code base and features) when there's a need. So far there' haven't been any loud voices for these kind of outputs.

Understood.  I agree then with the decision to remove the dead code. 

I personally do like `*.tab` files.  They are the only way of visually inspecting the data (directly via the output file) without passing them through a binary/hdf5 (or similar) reader.  I am happy to assign this functionality to myself for a future PR (although given your comments above, maybe you don't even want the "alive" code in Parthenon :smile:)

src/outputs/outputs.cpp
Yurlungur: > It may be worth to revive that discussion after people got the chance to experiment with Ascent for a bit.
As far as I can tell all that machinery will be automatically in place (though I don't know how "fast" it'll be and also how where the emphasis should be between duplicated, but internal code versus an external potentially not easy to install external dependency).

Oh interesting. I think that probably won't work for Phoebus, since it does things with metrics. Yeah I don't think we need to tackle this now. I just had it at the mind as I was culling this code. :)

>  I'm happy with plain hdf5 and providing easy python script to read/process/plot/convert the data based on the assumption that this will cover the use cases of a big majority of the users.

I think we may want to provide in-situ reduced dimensionality output---it can be very convenient and save time- and memory-consuming steps. But I also don't know if I've entirely thought through what the right thing here is.

> > In an AMR hierarchy, is slicing or summing a meaningful operation?

> definitely. (I know we'll need it)

Can you elaborate? I can see how slice/sum operations over the whole grid are meaningful, but slicing/summing on a per-meshblock basis doesn't seem to provided what we need unless one assumes, e.g., symmetry in the sliced direction. Maybe I'm missing something.

Yurlungur:
> Why are we eliminating `*.tab` files? These are very useful for 1D test problems.

I think we'd be happy to have the `*.tab` files re-implemented if you want to do that, @pdmullen ! I eliminated `*.tab` in particular because this was the constructor:

```C++
void FormattedTableOutput::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm,
                                           const SignalHandler::OutputSignal signal) {
  throw std::runtime_error(std::string(__func__) + " is not implemented");
}
```

I.e., the code wasn't just dead---it was a stub.

Yurlungur:
@pdmullen if you're ok with this can you click approve? Alternatively perhaps someone else could take a quick look? @bprather @jdolence @brryan @lroberts36 ?

pdmullen:
> @pdmullen if you're ok with this can you click approve? Alternatively perhaps someone else could take a quick look? @bprather @jdolence @brryan @lroberts36 ?

Reviewing now.

src/outputs/outputs.cpp
pdmullen: Slicing with mesh refinement is a must.  Of course interpreting the slice is a bit weird, e.g., a midplane slice through a disk probes different heights depending on the spatially varying `dx3`.

pdmullen (approved):
If the slicing functionality didn't actually work, I am okay with removing it (albeit it looks like the infrastructure is nearly there?).  This is an important feature, as @pgrete also argues.

Yurlungur:
@pdmullen The slicing is nearly there I think... But the organization is a little bit wrong, and the hooks won't actually work. If you feel strongly I can leave the non-working code in place until we can revive it. But I figure it's still there in version control so we can reference it when it's time to re-enable. Does that work?

src/outputs/outputs.cpp
Yurlungur: Right---so why is that useful if the interpretation is so difficult? What exactly does slicing like this buy you? Why not instead compute a *real* slice by interpolating to a 1D or 2D uniform grid?

pdmullen:
> @pdmullen The slicing is nearly there I think... But the organization is a little bit wrong, and the hooks won't actually work. If you feel strongly I can leave the non-working code in place until we can revive it. But I figure it's still there in version control so we can reference it when it's time to re-enable. Does that work?

works for me 😸

src/outputs/outputs.cpp
pdmullen: it buys me easy, high cadence, cheap outputs for movies :). In the future, if Parthenon could support on the fly, high-cadence outputs that interpolate to a uniform grid, that would be _awesome_.

pdmullen: maybe ascent support is the answer to everything here...

Yurlungur: That's what I would like to be able to do. :) I don't think it would be too hard.

PR #830, opened 2023-02-05
particle output

Yurlungur:
## PR Summary

This is a work in progress to add particle output. I will split particle output and reading particle data via restarts into two spearate PRs. This work depends on PR #829 to clean up the I/O first.

The strategy is as follows:
- Per swarmvariable, output all particles to a single 1D flat array
- Output two additional metadata arrays: `particle_offsets` and `particle_counts` which map meshblock global index to the index where particles for that meshblock start and end in the big flat array

I update the docs to reflect these changes, and migrate to sphinx where appropriate.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


Yurlungur:
Ok this turned out to be more of an effort than I would have liked. My strategy is as above: 
- Per swarmvariable, output all particles to a single 1D flat array. This avoids having to write a ragged array.
- Output two additional metadata arrays: particle_offsets and particle_counts which map meshblock global index to the index where particles for that meshblock start and end in the big flat array

In the hdf5 file, a swarm is a group, containing two datasets: `counts` and `offsets` which are arrays containing the number of particles per meshblock, and what index in the global array of particles a meshblock starts. Each swarm also has a `SwarmVars` group containing the actual variables output. Full tensor variable shapes are supported. The output from the `particle_leapfrog` example looks something like this:
```
jonahm@ubuntu:$ h5ls -r particles.out0.final.phdf
/                        Group
/Blocks                  Group
/Blocks/loc.level-gid-lid-cnghost-gflag Dataset {8, 5}
/Blocks/loc.lx123        Dataset {8, 3}
/Blocks/xmin             Dataset {8, 3}
/Info                    Group
/Input                   Group
/Levels                  Dataset {8}
/Locations               Group
/Locations/x             Dataset {8, 9}
/Locations/y             Dataset {8, 9}
/Locations/z             Dataset {8, 9}
/LogicalLocations        Dataset {8, 3}
/Params                  Group
/VolumeLocations         Group
/VolumeLocations/x       Dataset {8, 8}
/VolumeLocations/y       Dataset {8, 8}
/VolumeLocations/z       Dataset {8, 8}
/my\ particles           Group
/my\ particles/SwarmVars Group
/my\ particles/SwarmVars/id Dataset {14}
/my\ particles/SwarmVars/v Dataset {3, 14}
/my\ particles/SwarmVars/vv Dataset {3, 3, 14}
/my\ particles/SwarmVars/x Dataset {14}
/my\ particles/SwarmVars/y Dataset {14}
/my\ particles/SwarmVars/z Dataset {14}
/my\ particles/counts    Dataset {8}
/my\ particles/offsets   Dataset {8}
```
I also added python code to read swarms. It looks something like this:
```C++
data = phdf(filename)
swarm = phdf.GetSwarm(swarmname)
plt.scatter(swarm.x, swarm.y, c=swarm['some field'])
```
which would make an x-y scatter plot with the scalar swarm field as the color.

Getting this all to work was a bigger ordeal than I had hoped, as the particles have to be gathered accross meshblocks in a fairly painful way, with a lot of book keeping. However, this machinery all works. The `particle_leapfrog` regression tests now use this hdf5 format for output. And on my machine they pass, though it looks like we're still having problems on the CI machine.

## What's misssing:

- I haven't gotten to restarts yet.
- I have not modified `movie2d.py` to plot swarms. It would be nice to add, but probably not immediately essential. I'm also waiting for #853 to merge.
- I wrote code to do xdmf output for swarms (and in fact I cleaned up and made the current xdmf more human-readable). Vis readers can read the new xdmf and plot grid data. But for the life of me I can't get them to actually PLOT the particle data. I've reached out for help but haven't gotten anywhere yet.

## So where are we at?

I think this MR is in a place where the featureset is 75% of the way there, and I'm being pulled towards other deliverables. I'd like to get this reviewed at the point where it's at and merged in, and then fix the xdmf and get restarts working in subsequent PRs. I recall that @BenWibking expressed interest in getting this through to enable tracer particles. If you're interested in helping pull this through I think it's now at a good point to collaborate on it. We could merge what's in here in and then we could separately tackle xdmf and restarts, either by splitting the tasks, or handling them together. What do you think?

I would like to get @brryan 's eyes on this in particular, as he's the one who wrote the swarm machinery. From the AthenaK crowd, @BenWibking would be an obvious choice if he wants to build on this. But @pgrete and @forrestglines your eyes would certainly be welcome. And any help anyone can offer on the XDMF is very welcome.

Yurlungur:
Ah I just realized #838 hasn't been merged. I depend on that for this MR. Can someone from AthenaK please review it? @pgrete @forrestglines ?

Yurlungur:
## Update

OK I sat down and did restarts and plotting.
- Restarts now work. I modified the `particle_leapfrog` regression test to use a restart. It generates a restart, then runs from the restart to completion.  Particle log csv file is now gone. Instead the phdf datasets are used.
- I updated the `movie2d.py` script (from the updated version in  #853 ) to optionally overplot particle datasets. In particular, I add three optional arguments:
  - `--swarm`, which is the swarm to plot. x and y positions are overplotted as a scatterplot.
  - `--swarmcolor`, which sets the color of the scatterplot points. By default the color is black. You can specify a matplotlib colorstring or a scalar swarmvar.
  - `--maxparticles` is motivated by the fact that any reasonable simulation will have more particles than you can reasonably scatterplot. `maxparticles` subsamples (by just skipping particles with even stride) the particle dataset down to some maximum number. If the original dataset had less than that maximum, no action is taken. The default number for maxparticles is 100.

This results in a plot like this one (which I got from `particle_tracers`):


So now I think the *only* outstanding issue is xdmf. I don't think I can move forward with xdmf until I get some help. So I believe we should merge this in. Then if anyone wants to tackle xdmf you're welcome to. I will continue to try and get some support from the paraview people at LANL but no promises.

example/particle_tracers/particle_tracers.cpp
```diff
@@ -326,7 +326,7 @@ void ProblemGenerator(MeshBlock *pmb, ParameterInput *pin) {
   auto rng_pool = tr_pkg->Param<RNGPool>("rng_pool");
 
   const int ndim = pmb->pmy_mesh->ndim;
-  PARTHENON_REQUIRE(ndim == 1, "Tracer particles example only supports 1D!");
+  PARTHENON_REQUIRE(ndim <= 2, "Tracer particles example only supports <= 2D!");
```
Yurlungur, line 329: This doesn't appear to be a hard limit. The tracers cluster around the "advected" quantity in 1d, but the sim runs fine in 2d.

example/particle_tracers/particle_tracers.cpp
```diff
@@ -397,6 +400,7 @@ void ProblemGenerator(MeshBlock *pmb, ParameterInput *pin) {
 
         y(n) = y_min + rng_gen.drand() * (y_max - y_min);
         z(n) = z_min + rng_gen.drand() * (z_max - z_min);
+        id(n) = static_cast<int>(std::round(number_mesh / nbtotal)) * gid + n;
```
Yurlungur: Previously tracer Ids were declared but not initialized

example/sparse_advection/sparse_advection_package.cpp
```diff
@@ -216,7 +216,7 @@ TaskStatus CalculateFluxes(std::shared_ptr<MeshBlockData<Real>> &rc) {
 
         for (int n = 0; n < nvar; n++) {
           if (!v.IsAllocated(n)) continue;
-          const auto this_v = vx[n % NUM_FIELDS];
+          const auto this_v = vx[v(n).sparse_id % NUM_FIELDS];
```
Yurlungur: These are changes from #838

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
```diff
@@ -152,7 +191,7 @@ def plot_dump(
     fig = plt.figure()
     p = fig.add_subplot(111, aspect=1)
     if time_title is not None:
-        p.set_title(f"t = {time_title} seconds")
+        p.set_title(f"t = {time_title}")
```
Yurlungur, line 226: This was just wrong. Units are not necessarily seconds.

scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
```diff
@@ -21,6 +21,52 @@
 import numpy as np
 
 
+class Swarm:
```
Yurlungur, line 24: We may need more machinery for the swarm class down the line. I have a bunch of Python code for particle data structures like this one [here](https://github.com/lanl/nubhlight/blob/master/script/analysis/hdf5_to_dict.py#L760) but I think porting it over into Parthenon now is premature optimization. I'll port the relevant stuff over as needed, or as requested. Some features that we might want to think of in the future:
- Generating a time-series from a collection of data files
- Filtering the particle dataset on certain conditions

scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
```diff
@@ -243,17 +289,21 @@ def load_ghost_coords(coord_i):
...
-        self.DatasetNumComponents = dict(
-            zip(
-                np.array(info.attrs["OutputDatasetNames"]).ravel().astype(str),
-                np.array(info.attrs["NumComponents"]).ravel(),
+        try:
+            self.DatasetNumComponents = dict(
+                zip(
+                    np.array(info.attrs["OutputDatasetNames"]).ravel().astype(str),
+                    np.array(info.attrs["NumComponents"]).ravel(),
+                )
```
Yurlungur, line 309: This change is motivated by the need for `phdf` to work even if there is no grid data present.

scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
```diff
@@ -458,7 +508,7 @@ def Get(self, variable, flatten=True, interior=False):
         xg and xng, etc.
         """
         try:
-            if self.varData[variable] is None:
+            if self.varData.get(variable) is None:
```
Yurlungur, line 517: `dictionary.get` returns `None` if the key is not present, so there's no reason to initialize this dictionary with `None` values.

scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
```diff
@@ -477,10 +527,10 @@ def Get(self, variable, flatten=True, interior=False):
                         )
                         self.varData[variable] = tmp.reshape((newShape))
 
-        except:
+        except KeyError:
```
Yurlungur, line 536: safer

src/CMakeLists.txt
```diff
@@ -247,6 +252,9 @@ set_target_properties(parthenon PROPERTIES SOVERSION ${parthenon_VERSION})
 
 target_compile_features(parthenon PUBLIC cxx_std_17)
 
+target_link_options(parthenon PUBLIC "$<$<AND:$<CXX_COMPILER_ID:GNU>,$<VERSION_LESS:$<CXX_COMPILER_VERSION>,9.1>>:-lstdc++fs>" )
+target_link_options(parthenon PUBLIC "$<$<AND:$<CXX_COMPILER_ID:Clang>,$<VERSION_LESS:$<CXX_COMPILER_VERSION>,9.0>>:-lc++fs>" )
```
Yurlungur: Coming from #838 @bprather is going to provide a better fix

src/interface/variable.hpp
```diff
@@ -190,6 +192,13 @@ class ParticleVariable {
     return data(std::forward<Args>(args)...);
   }
 
+  auto GetHostMirrorAndCopy() { return data.GetHostMirrorAndCopy(); }
+
+  auto GetHostMirrorAndCopy(int n6, int n5, int n4, int n3, int n2) {
+    auto data_slice = Kokkos::subview(data, n6, n5, n4, n3, n2, Kokkos::ALL());
+    return data_slice.GetHostMirrorAndCopy();
+  }
```
Yurlungur, line 200: Convenience thing for outputs

src/outputs/output_utils.hpp
```diff
@@ -0,0 +1,212 @@
...
+
+// Parthenon
+#include "interface/metadata.hpp"
+#include "interface/variable.hpp"
+#include "mesh/mesh.hpp"
+#include "mesh/meshblock.hpp"
+#include "utils/error_checking.hpp"
+
+namespace parthenon {
+namespace OutputUtils {
```
Yurlungur, line 39: `OutputUtils` namespace and `output_utils.{cpp,hpp}` are both new from some reorg I did of the hdf5 output. I put the `VarInfo` struct here, as well as my own new swarm logic. Other than the xdmf, this was the biggest lift of the PR. It was somewhat complicated to construct the data necessary for output of particle data in the shape I planned out. I think this is worth it, though, as a single flat array of particles is easiest to interpret in post-processing and is less fraught than a ragged array.

src/outputs/outputs.hpp
```diff
@@ -230,6 +202,13 @@ class PHDF5Output : public OutputType {
   std::string GenerateFilename_(ParameterInput *pin, SimTime *tm,
                                 const SignalHandler::OutputSignal signal);
   const bool restart_; // true if we write a restart file, false for regular output files
+  // TODO(JMM): these methods might want to live in the base class or in output_utils.hpp
+  void ComputeXminBlocks_(Mesh *pm, std::vector<Real> &data);
+  void ComputeLocs_(Mesh *pm, std::vector<int64_t> &locs);
+  void ComputeIDsAndFlags_(Mesh *pm, std::vector<int> &data);
+  void ComputeCoords_(Mesh *pm, bool face, const IndexRange &ib, const IndexRange &jb,
+                      const IndexRange &kb, std::vector<Real> &x, std::vector<Real> &y,
+                      std::vector<Real> &z);
```
Yurlungur, line 213: These are utility functions I added to pull this logic out of the main bit of `parthenon_hdf5` output. I did this because that file was just very difficult to read.

src/outputs/parthenon_hdf5.cpp
```diff
@@ -1003,13 +671,220 @@ std::string PHDF5Output::GenerateFilename_(ParameterInput *pin, SimTime *tm,
...
+}
 
 // explicit template instantiation
 template void PHDF5Output::WriteOutputFileImpl<false>(Mesh *, ParameterInput *, SimTime *,
                                                       SignalHandler::OutputSignal);
 template void PHDF5Output::WriteOutputFileImpl<true>(Mesh *, ParameterInput *, SimTime *,
                                                      SignalHandler::OutputSignal);
 
+// Utility functions implemented
+namespace HDF5 {
```
Yurlungur, line 735: Moved from the top of the file to the bottom for easier parsing.

src/outputs/parthenon_xdmf.cpp
```diff
@@ -0,0 +1,413 @@
...
+#include "mesh/mesh.hpp"
+#include "outputs/output_utils.hpp"
+#include "outputs/parthenon_hdf5.hpp"
+#include "outputs/parthenon_xdmf.hpp"
+#include "utils/utils.hpp"
+
+namespace parthenon {
+using namespace OutputUtils;
+
+namespace XDMF {
```
Yurlungur, line 46: I moved all the xdmf machinery into a separate `parthenon_xdmf` file, as I was having trouble reading the `parthenon_hdf5` file and navigating it. It's not much of a cleanup but I think it helps.

src/outputs/parthenon_xdmf.cpp
```diff
@@ -0,0 +1,413 @@
...
+}
+static std::string ParticleDatasetRef(const std::string &prefix,
+                                      const std::string &swmname,
+                                      const std::string &varname,
+                                      const std::string &hdffile,
+                                      const std::string &datatype,
+                                      const std::string &extradims, int particle_count) {
+  std::string precision_string = (datatype == "Float") ? " Precision=\"8\"" : "";
+  auto part =
+      StringPrintf("%s<DataItem Format=\"HDF\" Dimensions=\"%s%d\" Name=\"%s\" "
```
Yurlungur, line 321: I found the stream operator impossible to parse for writing xdmf, so I wrote a function that generates a `std::string` from a format `char*`. It's in `utils/utils.hpp`.  But this lets me write much more human-readable xdmf in this file.

src/utils/utils.hpp
```diff
@@ -35,6 +38,33 @@ namespace parthenon {
...
+  std::vector<char> buffer(size);
+  char *pbuffer = buffer.data();
+  std::snprintf(pbuffer, size, fmt, std::forward<Args>(args)...);
+  // Return cast to string
+  return std::string(pbuffer);
+}
+template <typename... Args>
+std::string StringPrintf(const std::string &fmt, Args... args) {
+  return StringPrintf(fmt.c_str(), std::forward<Args>(args)...);
+}
```
Yurlungur, line 66: Here's the `StringPrintf` function I mentioned. @brryan suggests that maybe we should provide a `PARTHENON_FAIL` macro that takes a fmt string and list of args. Not a bad idea, IMO.

src/interface/swarm.cpp
```diff
@@ -372,6 +372,8 @@ void Swarm::Defrag() {
 
   std::list<int> new_free_indices;
 
+  free_indices_.sort();
```
Yurlungur, line 375: This was a bug in the defrag code. Thanks @brryan for the find!

src/outputs/output_utils.cpp
```diff
@@ -0,0 +1,122 @@
...
+    out += buffer[i];
+  }
+  for (int i = 0; i < buffer.size(); ++i) {
+    tot_count += buffer[i];
+  }
+#else
+  tot_count = local;
+#endif // MPI_PARALLEL
+  return out;
+}
```
Yurlungur, line 120: This may need to be generalized. Also this assumes that `unsigned long long int = size_t` on all platforms we care about because I use `MPI_UNSIGNED_LONG_LONG` as the MPI type.

tst/regression/test_suites/particle_leapfrog/particle_leapfrog.py
```diff
@@ -30,16 +30,33 @@
...
+        parameters.coverage_status = "both"
 
+        if step == 1:
+            parameters.driver_cmd_line_args = ["parthenon/job/problem_id=gold"]
+        else: # step == 2
+            parameters.driver_cmd_line_args = [
+                "-r",
+                "gold.out1.00001.rhdf",
+                "parthenon/job/problem_id=particles",
+            ]
```
Yurlungur: restart from previous run to ensure restarts writing/reading work

tst/regression/test_suites/particle_leapfrog/particle_leapfrog.py
```diff
@@ -30,16 +30,33 @@
...
-        data = np.genfromtxt("particles.csv", delimiter=",", names=True)
-
-        # pick last cycle (given current parameter file)
-        final_data = data[data["ncycle"] == 184]
-        final_data.sort(order="particles_id")
+        data = phdf("particles.out0.final.phdf")
+        swarm = data.GetSwarm("my particles")
+        inds = np.argsort(swarm["id"])
+        final_data = np.vstack((swarm.x, swarm.y, swarm.z, swarm["v"]))
+        final_data = final_data.transpose()[inds]
```
Yurlungur: use the phdf data to compare to the known correct solution

src/parthenon_manager.hpp
```diff
@@ -53,6 +57,59 @@ class ParthenonManager {
 
  private:
   ArgParse arg;
+
+  template <typename T>
+  void ReadSwarmVars_(const SP_Swarm &pswarm, const BlockList_t &block_list,
```
Yurlungur, line 62: templated and in header file because there are int and real valued swarmvars

Yurlungur:
OK with #838 merged, the diffs for this PR are a lot more legible. This is ready for review. @BenWibking and @brryan this is probably of most interest to you. That said, CC @pgrete @forrestglines @bprather @pdmullen @jdolence @lroberts36 in case you're interested.

BenWibking:
Thanks for reminder, will look in detail tomorrow.

brryan:
> ## Update
> 
> OK I sat down and did restarts and plotting.
> 
>     * Restarts now work. I modified the `particle_leapfrog` regression test to use a restart. It generates a restart, then runs from the restart to completion.  Particle log csv file is now gone. Instead the phdf datasets are used.
> 
>     * I updated the `movie2d.py` script (from the updated version in  [Improve the performance of movie2d.py script #853](https://github.com/parthenon-hpc-lab/parthenon/pull/853) ) to optionally overplot particle datasets. In particular, I add three optional arguments:
>       
>       * `--swarm`, which is the swarm to plot. x and y positions are overplotted as a scatterplot.
>       * `--swarmcolor`, which sets the color of the scatterplot points. By default the color is black. You can specify a matplotlib colorstring or a scalar swarmvar.
>       * `--maxparticles` is motivated by the fact that any reasonable simulation will have more particles than you can reasonably scatterplot. `maxparticles` subsamples (by just skipping particles with even stride) the particle dataset down to some maximum number. If the original dataset had less than that maximum, no action is taken. The default number for maxparticles is 100.
> 
> 
> This results in a plot like this one (which I got from `particle_tracers`): 
> 
> So now I think the _only_ outstanding issue is xdmf. I don't think I can move forward with xdmf until I get some help. So I believe we should merge this in. Then if anyone wants to tackle xdmf you're welcome to. I will continue to try and get some support from the paraview people at LANL but no promises.

Not part of this PR but an interesting issue: The particles shown in your plot every four blocks are the same. I bet this means RNG seeds are being duplicated in a per-block way. I'll look into it.

doc/sphinx/src/particles.rst
```diff
@@ -161,3 +161,17 @@ Swarm::bounds array entries to factory functions that allocate
...
+Outputs for swarms can be set in an output block, just like any other
+variable. The user must specify two comma separated lists:
+
+* ``swarms = swarm1, swarm2, ...`` etc. This specifies that these swarms
+  are marked for output.
+
+* ``swarm_variables = var1, var2, ...`` etc. This specifies that for
+  each swarm the listed variables should be output. Note that if a
+  swarm is missing a variable in the ``swarm_variables`` list, no
+  error is raised. All available variables are output.
```
brryan: This should always produce the requested data in output files but for multiple swarms this could easily produce too much data, I think. 

One alternative would be to allow e.g. `<parthenon/output0/swarm0>` blocks that internally contained the swarm name and the vars to output e.g.
```
<parthenon/output0/swarm0>
swarm = my particles
variables = id, v, vv
```

This is something I could do subsequently, except that since it would be a breaking change maybe we should nail down the "right" way to do things here?

example/particle_leapfrog/parthinput.particle_leapfrog
```diff
@@ -45,10 +45,19 @@ nx2 = 8
 nx3 = 8
 
 <parthenon/time>
-tlim = 1.0
+tlim = 2.0
```
brryan: Necessary change? `dt = 0.5` below instead?

example/particle_leapfrog/particle_leapfrog.cpp
```diff
@@ -117,174 +118,6 @@ Real EstimateTimestepBlock(MeshBlockData<Real> *rc) {
 
 } // namespace Particles
 
-// Simple log function to ease regression tests and debugging until there's a proper
-// particles output method. Do NOT use in practice.
-TaskStatus WriteParticleLog(BlockList_t &blocks, int ncycle) {
```
brryan, line 122: Oh right this is deprecated :) It performed a good service for a long time

example/particle_tracers/particle_tracers.cpp
```diff
@@ -397,6 +400,7 @@ void ProblemGenerator(MeshBlock *pmb, ParameterInput *pin) {
 
         y(n) = y_min + rng_gen.drand() * (y_max - y_min);
         z(n) = z_min + rng_gen.drand() * (z_max - z_min);
+        id(n) = (num_tracers / nbtotal) * gid + n;
```
brryan: should this actually be `id(n) = num_tracers_meshblock * gid + n;`?

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
```diff
@@ -119,6 +139,22 @@
...
+    logger.error(f"{key} not found in {search_location}. Further processing stopped.")
+    logger.info(f"Available fields: {available}")
+    return
+
+
+def subsample(array, maxlen):
+    "Subsample array to have a maximum length of maxlen by even sampling"
+    ratio = len(array) / maxlen
+    if ratio >= 1:
+        aout = array[:: int(ratio)]
```
brryan, line 173: I'm unfamiliar with this syntax but I checked that it works. Could the above comment be made a little more specific, something like, `"Subsample array with fixed stride to have a maximum length of maxlen"`?

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
```diff
@@ -51,6 +52,25 @@
...
+    default=None,
+    help="Optional color of overplotted particle positions. Default is black. You may specify a scalar swarm variable as the color or a matplotlib color string.",
+)
+parser.add_argument(
+    "--maxparticles",
+    metavar="N",
+    type=int,
+    default=100,
+    help="Limit plot to only N particles at most. Default is 100.",
+)
```
brryan, line 94: Not blocking for this PR but `markersize` would be a nice option to have. The default is pretty large for anything more than a very small set of particles. This would set the `s` argument to `scatter()`

scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
brryan: Yeah this is worth thinking about in detail (not all your analysis tools may make sense at the `phdf` level, but they probably at least indicate what data needs to be exposed by `phdf`). Happy to leave this for now and return to it later.

src/interface/swarm.cpp
brryan: Yep, sorry about that!

src/interface/swarm.hpp
```diff
@@ -66,6 +66,7 @@ class Swarm {
 
  public:
   Swarm(const std::string &label, const Metadata &metadata, const int nmax_pool_in = 3);
+  Swarm() : Swarm("default", Metadata()) {}
```
brryan: Where is this needed?

src/interface/swarm_container.hpp
```diff
@@ -131,6 +124,7 @@ class SwarmContainer {
   /// Remove a variable from the container or throw exception if not
   /// found.
   /// @param label the name of the variable to be deleted
+  /// TODO(JMM): Should we support this operation?
```
brryan, line 127: Probably not...

src/outputs/output_utils.cpp
```diff
@@ -0,0 +1,122 @@
+//========================================================================================
+// Parthenon performance portable AMR framework
+// Copyright(C) 2020-2023 The Parthenon collaboration
```
brryan: Just 2023 right?

src/outputs/output_utils.cpp
brryan: What is the potentially needed generalization?

src/outputs/output_utils.hpp
```diff
@@ -0,0 +1,212 @@
+//========================================================================================
+// Parthenon performance portable AMR framework
+// Copyright(C) 2020-2023 The Parthenon collaboration
```
brryan: ditto re: `(C) 2023`

src/outputs/parthenon_xdmf.cpp
```diff
@@ -0,0 +1,413 @@
+//========================================================================================
+// Parthenon performance portable AMR framework
+// Copyright(C) 2020-2023 The Parthenon collaboration
```
brryan: 2023

src/outputs/parthenon_xdmf.hpp
```diff
@@ -0,0 +1,37 @@
+//========================================================================================
+// Parthenon performance portable AMR framework
+// Copyright(C) 2020-2023 The Parthenon collaboration
```
brryan, line 3: 2023

src/utils/utils.hpp
brryan: Would make it possible to write out a formatted error in one line instead of setting up the string stream first.

tst/regression/test_suites/particle_leapfrog_outflow/particle_leapfrog_outflow.py
```diff
@@ -32,12 +32,18 @@ def Prepare(self, parameters, step):
         return parameters
 
     def Analyse(self, parameters):
+        sys.path.insert(
+            1,
+            parameters.parthenon_path
+            + "/scripts/python/packages/parthenon_tools/parthenon_tools",
```
brryan, line 38: Should we insert this path only if an installed parthenon_tools isn't found? Not sure the right way to do that.

tst/regression/test_suites/particle_leapfrog_outflow/particle_leapfrog_outflow.py
```diff
@@ -49,10 +55,7 @@ def Analyse(self, parameters):
...
             ]
         )
-        final_data = structured_to_unstructured(
-            final_data[["x", "y", "z", "vx", "vy", "vz"]]
-        )
         if ref_data.shape != final_data.shape:
             print("TEST FAIL: Mismatch between actual and reference data shape.")
             return False
-        return (final_data == ref_data).all()
+        return (np.abs(final_data - ref_data) <= 1e-10).all()
```
brryan, line 61: Maybe we need some kind of `test_utils.py` that defines things like `soft_equiv()`

src/outputs/output_utils.hpp
brryan: I agree that contiguous particle data is best. I'm not immediately succeeding in parsing the output logic before (I trust that it works because outputs work). But my only question is -- is it possible to easily get out just the particles from a particular block, maybe just at the `phdf` level?

brryan (approved):
This is a great addition -- thank you for the hard work @Yurlungur. Some minor comments but it would be great to merge this in.

Yurlungur:
> I bet this means RNG seeds are being duplicated in a per-block way. I'll look into it.

Thanks good catch. This might just be a small thing in that particular example.

example/particle_leapfrog/parthinput.particle_leapfrog
Yurlungur: not a necessary change here. I changed it because I wanted to run an identical sim to the one in the regression test input deck for debugging.

example/particle_tracers/particle_tracers.cpp
Yurlungur: No---doing that produces overlapping ids. I didn't quite track down what that is, though. I just wanted something sufficiently sparse that no ids overlapped.

Yurlungur: I think `num_tracers_meshblock` is not computed as the total tracer number divided by meshblock number. It's computed by integrated the advected "mass." But that means that since it varies per meshblock it's not useful for a total "ID" space.

example/particle_leapfrog/particle_leapfrog.cpp
Yurlungur: Deprecated now that we have real IO, anyway. :)

scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
Yurlungur: I have something in mind---percolating in my brain. I'm thinking the following might make sense:
1. The ability to "concatenate" snapshots... perhaps by overloading the `+` operator for swarms
2. The ability to re-save a massaged swarm dataset to disk from python for further post-processing
3. The ability to filter a dataset based on a mask.

I think all of that *could* live at the phdf level if written sufficiently generally.

src/outputs/output_utils.cpp
Yurlungur: if `MPI_UNSIGNED_LONG_LONG` isn't the same type as `std::size_t` we would need to infer the proper `MPI_TYPE` passed into `MPI_Allgather` from the system.

src/outputs/output_utils.hpp
Yurlungur: Yes, that's possible. I could add some `phdf` machinery to do this if you want. It's contained in the `offset` and `count` array. The particles in block `b` are given by the `swarm/SwarmVar/var[offsets[b]:offsets[b]+counts[b]+1]`

src/utils/utils.hpp
Yurlungur: I did start writing this macro, but it's a little bit more complicated than I thought it would be. The order of the arguments for the `impl::parthenon_fail` method needs to be swapped so that it can take a variatic pack as the final argument set. Not a huge change, but something to be aware of.

src/interface/swarm.hpp
Yurlungur: Not sure it's needed anymore. I will try removing it.

tst/regression/test_suites/particle_leapfrog_outflow/particle_leapfrog_outflow.py
Yurlungur: For regression tests, I think it's correct to always use the version of `parthenon_tools` packaged with the source code, since the tests should be testing that?

doc/sphinx/src/particles.rst
Yurlungur: Hmm... that's kind of hard because it requires walking a second output block, which the current `Output` class isn't really designed to do. What about this?
```
<parthenon/output0>
swarm = my_particles #swarms you want to output
# per swarm a list of variables
my_particles = id, v, vv
```

Alternatively, what about this?
```
<parthenon/output0>
swarm = my particles #swarms you want to output

<parthenon/output0/my particles>
variables= id, v, vv
```
would one of those work? And if so which do you think you like better?

Yurlungur: The former one requires forbidding spaces in swarm names, but maybe that's a good thing anyway.

brryan: Ah ok -- I have a slight preference for the latter since I think it is a bit more explicit. If the former is substantially easier I think that is fine though.

One could modify the first one probably to be
```
<parthenon/output0>
swarm = my_particles #swarms you want to output
# per swarm a list of variables
my_particles_variables = id, v, vv
```
without much extra effort I would think? That's a bit more explicit.

Regardless, prohibiting spaces in swarm names is probably a good thing.

Yurlungur: I think `<parthenon/output0/my particles>` and
 ```
<parthenon/output0>
my_particles_variables
```
are roughly the same difficulty. I just need to list the swarm names somewhere in `parthenon/output0`, not elsewhere.

File: example/particle_tracers/particle_tracers.cpp (hunk 25 above)
Comment (brryan) [reply example/particle_tracers/particle_tracers.cpp]: Oh ok, I must have misunderstood `num_tracers_meshblock`. Does the `id(n) = ` line as is guarantee no overlaps though? Isn't the stride even per meshblock but what if an earlier meshblock has many more particles than the length of that "average" stride?

Comment (brryan) [reply example/particle_tracers/particle_tracers.cpp]: For simplicity you could probably just do `num_tracers * gid + n` right?

Comment (Yurlungur) [reply example/particle_tracers/particle_tracers.cpp]: I am not sure. I didn't really dig in too much. For sure, we could just do `num_tracers * gid + n`. That's guaranteed to work.

File: doc/sphinx/src/particles.rst (hunk 22 above)
Comment (Yurlungur) [reply doc/sphinx/src/particles.rst]: Both approaches feel a bit weird. But maybe I will go with `<parthenon/output0/my particles>` as that feels more consistent with parthenon patterns.

File: src/outputs/output_utils.hpp (hunk 11 above)
Comment (Yurlungur) [reply src/outputs/output_utils.hpp:39]: Ok you can now do the following
```python
slc = swarm.Block(b)
print(swarm.x[slc]) # prints only the x components of particles on block b
```

doc/sphinx/src/particles.rst
Yurlungur: Ok turns out `<parthenon/output0/my particles>` solution doesn't work without some surgery because of the way the code walks through output blocks.  So I went with the `my_particles_variables` solution.

I did not, however, remove the `swarm_variables` input field. The way it works now is that all variables listed in `swarm_variables` are assumed to be requested by all swarms, whereas variables in the  `my_swarm_name_variables` field are only for `my_swarm_name`. I think this is a good compromise that should offer a nice level fidelity.

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
Yurlungur: done

example/particle_tracers/particle_tracers.cpp
Yurlungur: done

src/interface/swarm_container.hpp
Yurlungur: It's consistent with `meshblock_data` too so it's consistent. I just don't like we do that for either thing. Not gonna modify this now but something to keep in mind.

tst/regression/test_suites/particle_leapfrog_outflow/particle_leapfrog_outflow.py
Yurlungur: :+1: Yeah good idea. Not gonna add it here though.

example/particle_leapfrog/parthinput.particle_leapfrog
Yurlungur: reverted.

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
Yurlungur: Done. There's now a `--particlesize` flag, which takes either a float or a string. If it's a float, that's the size in points. If it's a string, the swarm variable with that name is used for the size. (Future improvement: maybe worth a scale factor too)

src/interface/swarm.hpp
Yurlungur: Does not appear to be required. Will remove.

src/outputs/output_utils.cpp
Yurlungur: yep

src/outputs/output_utils.hpp
Yurlungur: yep

src/outputs/parthenon_xdmf.hpp
Yurlungur: yep

Yurlungur:
@brryan I think I've addressed your comments. Are you happy with my changes?

@BenWibking ping :)

doc/sphinx/src/particles.rst
brryan: OK yeah I agree that's a good compromise, makes sense and exposes maximum granularity in requesting data output -- thanks

Yurlungur: this did require forbidding spaces in swarm names and also forbidding any swarm from just being named "swarm," since that would cause a name collision

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
brryan: Does the swarm class in `phdf` contain a default marker size that can be updated? I can't see the updated code yet here, maybe it still needs to be pushed

src/interface/swarm_container.hpp
brryan: Yep :+1:

doc/sphinx/src/particles.rst
brryan: I think both those restrictions are good to have in place

tst/regression/test_suites/particle_leapfrog_outflow/particle_leapfrog_outflow.py
brryan: Yeah you're right -- I see now too that we do this for all the test suite tests.

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
Yurlungur: Looks like it's present in the branch. 
https://github.com/parthenon-hpc-lab/parthenon/blob/jmm/particle-output/scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py#L75

The `Swarm` class does not have a default marker size. The default marker size is `matplotlib`s default. If you use a variable then the markersize for each particle is just the float for that variable. Probably it should be rescaled from 0 to some reasonable max (10 points maybe?)

Yurlungur: Ok now it rescales. I added a `--maxparticlesize` parameter (with the default given by matplotlibs default markersize). So now if you pass in a swarm variable for the `--particlesize` field, you get particles that vary with that variable with the minimum size = 0 and the maximum `maxparticlesize`. Here's the capability for the `id` field for the `particle_tracers` sim:

Yurlungur:
Ping someone from AthenaPK: @BenWibking @pgrete @forrestglines

src/outputs/output_utils.hpp
BenWibking: The Ascent output needs an update in order to build after the change from `VarInfo` -> `OutputUtils::VarInfo`.

BenWibking: I think the only change needed is on line `ascent.cpp:170`.

BenWibking (changes requested):
`ascent.cpp` needs to be updated to use `VarInfo` in its new namespace.

VisIt 3.3.3 crashes when reading the XDMF outputs for the `particle_tracers` example (with `VisIt: Error - The compute engine running on localhost has exited abnormally.`). I'm not sure if that's supposed to be supported yet, but nonetheless it doesn't appear to work. (VisIt can still read the outputs from the advection example and plot them correctly.)

Paraview 5.11.0 can read the XDMF outputs without crashing (but no particles are viewable).

The regression tests are not able to run on my laptop (they cause it to run out of memory and freeze, requiring a hard reboot), but they also do not run with the current develop, so I don't think it's an issue with this PR.

Yurlungur:
Thanks for the review, @BenWibking !

> `ascent.cpp` needs to be updated to use `VarInfo` in its new namespace.

Fixed.

> 
> VisIt 3.3.3 crashes when reading the XDMF outputs for the `particle_tracers` example (with `VisIt: Error - The compute engine running on localhost has exited abnormally.`). I'm not sure if that's supposed to be supported yet, but nonetheless it doesn't appear to work. (VisIt can still read the outputs from the advection example and plot them correctly.)
> 
> Paraview 5.11.0 can read the XDMF outputs without crashing (but no particles are viewable).

I have given up on getting the xdmf working, at least for now. Help is certainly wanted for getting that working at a later date, but I don't think it should block the PR. That said, I didn't realize visit was actively crashing. To prevent that, I've disabled the particle XDMF for now with a preprocessor macro. Does that work?

> The regression tests are not able to run on my laptop (they cause it to run out of memory and freeze, requiring a hard reboot), but they also do not run with the current develop, so I don't think it's an issue with this PR.

Not sure how to handle that one---I only test on Linux, since my mac isn't set up for development.

BenWibking:
> > VisIt 3.3.3 crashes when reading the XDMF outputs for the `particle_tracers` example (with `VisIt: Error - The compute engine running on localhost has exited abnormally.`). I'm not sure if that's supposed to be supported yet, but nonetheless it doesn't appear to work. (VisIt can still read the outputs from the advection example and plot them correctly.)
> > Paraview 5.11.0 can read the XDMF outputs without crashing (but no particles are viewable).
> 
> I have given up on getting the xdmf working, at least for now. Help is certainly wanted for getting that working at a later date, but I don't think it should block the PR. That said, I didn't realize visit was actively crashing. To prevent that, I've disabled the particle XDMF for now with a preprocessor macro. Does that work?

It works now :)

BenWibking (approved):

Yurlungur:
Huh for some reason, the CI isn't recognizing @BenWibking 's approving review as an approval. Does anyone know what's up with that? To work around that, I will force merge (since we do in fact have two approving reviews) when all tests have passed, unless anyone objects.

BenWibking:
I don't think I have write access to the repo, so it might not count it, although I am a member of the parthenon-hpc-lab org. Not sure how the permissions work.

Yurlungur:
Ah ok. I just gave you write access---you should be able to work on branches now.

Yurlungur:
I would have thought being a member of the org would have transferred permissions... 🤷

PR #831, opened 2023-02-10
Metadata Upgrade

Yurlungur:
## PR Summary

To provide find swarms for output, (i.e., for PR #830) I needed a mechanism to search through swarms by metadata flag. This got me thinking that the way we pull variables from `MeshBlockData` by flag is terribly inefficient, because it's linear in number of variables.

Turns out @jdolence and @lroberts36 saw this effect in riot---where there are many more variables than flags---and fixed it by additionally storing a map from `MetadataFlag` to variable. They already fixed that, so I decided to port that machinery into here for variables and swarms.

While I was working on this, @pdmullen noted two shortcomings of the current infrastructure:
1. There is no way to pack variables with a more complex combination of flags. For example you can't request variables that have *either* of flags `a` and `b`, *both of* flags `d` and `e` and *not* flag `f`. 
2. There is no way to access a user-generated `MetadataFlag` except by storing it in (say) params upon creation.

The latter issue is easy to fix---expose a mechanism to request flags by name. That's much slower than working with a flag, so it should be done sparingly. But no issue there. It was trivial to add so it's in this PR.

For the former issue, I saw an opportunity with all this set machinery flying around, so I decided to try my hand at it. This PR contains the addition of the `Metadata::FlagCollection` class, which understands arithmetic. You can create a flag collection that represents the above scenario with arithmetic notation. For example:
```C++
using FC = Metadata::FlagCollection;
auto flags = (FC({a}, true) + FC({b}, true)) * FC(d) * FC(e) - FC(f);
```
where here `+` represents union, `*` intersection, and `-` set difference. The syntax is a little clumsy and it doesn't exactly follow set syntax. For example, the `FC({a}, true)` signifies that the element `a` should be treated as something one takes unions with. Otherwise, it would be treated internally separately as an intersection object.

This is necessary because all this logic about set operations happens on the *flags* not the variables, but what we actually care about is selecting the right *variables.* 

The machinery is also limited in that it can only work on objects like `MeshBlockData` where you can store a cache of the hash tables. This means it doesn't work on, e.g., `SparsePack`. Maybe @lroberts36 can solve that---I didn't try, as I'm not too familiar with the sparse packing. I recall that we were perhaps planning to remove the old packing machinery with the new sparse packing machinery. If we do that and people like this machinery I've added, hopefully it can make its way into sparse packs. In principal I think that should be straighforward---the relevant hash tables just need to be exposed to `SparsePack` somehow.

I'm still porting over what I need for swarms, but I thought you all should take a look at this now and give some feedback before I fall any deeper down this rabbit hole.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


Yurlungur:
Oh I almost forgot:
- I moved the `Metadata` docs into sphinx since I was touching them anyway.
- I discovered that `Pandoc` can convert our docs basically automatically so I added some instructions on how to do that.

Yurlungur:
The old `VariablePack` machinery always packed variables that contained all requested flags. This informs the default settings/constructor for the `Metadata::FlagCollection` object. By default it assumes you want the intersection of all the flags, not some union or set difference.

Also because I provide a constructor from standard library containers, downstream codes shouldn't notice any change in the interface unless they want to. You can still pack on a `std::vector<MetadataFlag>`. Under the hood, that gets converted to a `FlagCollection` that requests that all flags be present.

tst/unit/test_metadata.cpp
```diff
@@ -64,18 +64,28 @@ TEST_CASE("Built-in flags are registered", "[Metadata]") {
...
-    REQUIRE("TestFlag" == f.Name());
+    REQUIRE(name == f.Name());
+
+    // Metadata should be able to report that this flag exists and
+    // nonexistent flags don't.
+    REQUIRE(Metadata::FlagNameExists(name));
+    REQUIRE(!(Metadata::FlagNameExists("NoCanDoBuddy")));
+
+    // The identical flag should be retrievable
+    auto const f2 = Metadata::FlagFromName(name);
```
pdmullen: Nice.  This is clean.  I suspect this functionality should only ever be used with user defined `Metadata`.  Should the accessor reflect that, e.g., 

`Metadata::UserFlagFromName`

src/interface/mesh_data.hpp
```diff
@@ -350,27 +349,27 @@ class MeshData {
   }
   const auto &PackVariablesAndFluxes() { return PackVariablesAndFluxesImpl(nullptr); }
 
-  // As above, DO NOT use variatic templates here. They shadow each other.
+  // As above, DO NO Tuse variatic templates here. They shadow each other.
```
pdmullen: typo

doc/sphinx/src/metadata.rst
```diff
@@ -0,0 +1,282 @@
...
+
+   using parthenon::Metadata;
+   using parthenon::MetadataFlag;
+
+   MetadataFlag const my_app_flag = Metadata::AllocateNewFlag("MyAppFlag");
+
+These can be used in all the same contexts that the built-in metadata
+flags are used. Parthenon will not interpret them in any way - it’s up
+to the application to interpret them.
+
```
pdmullen: Add documentation here on how to access flag by name.

doc/sphinx/src/metadata.rst
```diff
@@ -0,0 +1,282 @@
...
+By default constructor arguments go into the ``Intersections`` property
+of the ``FlagCollection``. However, if a container is passed into the
+constructor, you can also pass in an optional boolean flag to specify
+whether or not you want to match **any** flags instead of **all**
+flag. This adds the constructor arguments to the ``Unions``
+property of the ``FlagCollection``.
+
+.. code:: cpp
+
+   // Implicit construction form a container, which
```
pdmullen: form -> from

doc/sphinx/src/metadata.rst
```diff
@@ -0,0 +1,282 @@
...
+
+Produces a set s with the "unions" and "intersections" fields of set1
+and an exclusion field containing set1's exlcusion field as well as
+ALL THREE fields (union, intersection, exclusion) contained by set2.
+
+This feels unintuitive, but it makes expressions like
+
+.. code:: cpp
+
+   auto s = {Flag1, FLag2} * {Flag3, Flag4} - {Flag5, Flag6}
```
pdmullen: FLag -> Flag

pdmullen: Sorry... maybe I am still confused on the logic here, this only returns 

```
EITHER Flag1 or Flag2 AND Flag3 AND Flag4 and NOT Flag5 or Flag6
```
if `{Flag1, Flag2}` was constructed via

```FS_t my_set1({Flag1, Flag2}, true);```

and `{Flag3, Flag4}` was constructed via

```FS_t my_set2({Flag1, Flag2}, false);``` or ```FS_t my_set2({Flag1, Flag2});```

Is that right?

**Review by pdmullen — COMMENTED** [review]
What more can I say than wow? We were literally just talking about this at lunch yesterday.  I think this looks great, Jonah.  As you can see from my comments, they are only very minor.  But I think that my comments might also indicate that there may be some learning curve for new users to get the logic right here.  This just stresses the importance of documentation which you delivered on.  

I'd like to actually play around with this in downstream codes, but I think the unit tests demonstrate that everything is working as expected.  

Nice work.

File: tst/unit/test_metadata.cpp (hunk 1 above)
Comment (Yurlungur) [reply tst/unit/test_metadata.cpp]: Oh yes good point. :+1: One could also perhaps add checks to enforce it's only used for user flags. Not sure that's worth it though.

File: src/interface/mesh_data.hpp (hunk 2 above)
Comment (Yurlungur) [reply src/interface/mesh_data.hpp]: Yep good catch
```suggestion
  // As above, DO NOT use variatic templates here. They shadow each other.
```

doc/sphinx/src/metadata.rst
Yurlungur: Good catch
```suggestion
   // Implicit construction from a container, which
```

Yurlungur:
Hmm the advection test is failing so I need to figure out why the new packing logic changed things there. Maybe some implicit logic about variable pack ordering? I will check.

src/interface/mesh_data.hpp
```diff
@@ -352,25 +351,25 @@ class MeshData {
 
   // As above, DO NOT use variatic templates here. They shadow each other.
```
c-prather:
```suggestion
  // As above, DO NOT use variadic templates here. They shadow each other.
```
Not to be pedantic but

src/interface/metadata.cpp
```diff
@@ -79,12 +83,19 @@ class UserMetadataState {
...
+MetadataFlag Metadata::AllocateNewFlag(const std::string &name) {
+  return metadata_state.AllocateNewFlag(name);
 }
 
 std::string const &MetadataFlag::Name() const { return metadata_state.FlagName(*this); }
 
+bool Metadata::FlagNameExists(const std::string &flagname) {
+  return (metadata_state.NamesToFlags().count(flagname) > 0);
+}
+MetadataFlag Metadata::FlagFromName(const std::string &flagname) {
```
c-prather:
```suggestion
MetadataFlag Metadata::GetUserFlag(const std::string &flagname) {
```
This emphasizes both that it's the user flag (and the you should get Parthenon flags with `Metadata::Thing`) but also that it's not creating the flag, just retrieving it.

src/interface/metadata.cpp
```diff
@@ -79,12 +83,19 @@ class UserMetadataState {
 
 parthenon::internal::UserMetadataState metadata_state;
 
-MetadataFlag Metadata::AllocateNewFlag(std::string &&name) {
-  return metadata_state.AllocateNewFlag(std::move(name));
+MetadataFlag Metadata::AllocateNewFlag(const std::string &name) {
```
c-prather:
```suggestion
MetadataFlag Metadata::AddUserFlag(const std::string &name) {
```
In keeping with below.  Never liked `Allocate`, this isn't `malloc`

doc/sphinx/src/metadata.rst
Yurlungur: Ah yeah I was sloppy here. Yes `set1` needs to be constructed with `FS_t my_set1({Flag1, Flag2}, true)`. `set2` needs to be constructed with the `use_union` flag disabled, as you noted. But that's the default, so it could be constructed as `FS_t({Flag1, Flag2}, false)`, `FS_t({Flag1, Flag2})`, or `FS_t(Flag1, Flag2)`.

There's also an object-oriented interface, which I forgot to document, so the same could be achieved with:
```C++
FS_t flags;
flags.TakenUnion(std::vector<MetadataFlag>{Flag1, Flag2});
flags.TakeIntersection(std::vector<MetadataFlag>{Flag3, Flag4});
flags.Exclude(std::vector<MetadataFlag>{Flag5, Flag6});
```
I can also make it so you don't need to pass in a container to make that work. I'll update accordingly...

c-prather (approved):
This is fantastic.  I've been stashing flags in `Params` and I do not recommend it, this is much cleaner.

To clarify: pulling out a flag once per host-side function (e.g. ~10x per step) is fine?  The `FlagByName` lookup isn't too slow, is it?

Yurlungur:
> To clarify: pulling out a flag once per host-side function (e.g. ~10x per step) is fine? The FlagByName lookup isn't too slow, is it?

That's probably fine? The thing is `FlagByName` does string proessing, and hashes a map. So it's obviously slower than bit twiddling. But it's probably fine if you just generate the flag and then use it.

src/interface/metadata.cpp
Yurlungur: Is anyone but you using this? That would be a breaking change. One I'm certainly fine with though.

c-prather: Doesn't changing all the `PackVariables` signatures count as a breaking change already?

Anyway, I don't think anyone else uses this, no -- @pdmullen mentioned looking around for uses and only seeing my `Params` craziness. I certainly hope no one else bothered.

Yurlungur: The `PackVariables` signature is actually unchanged to some degree. The old signature still works because of C++ automatic casting.

c-prather: As in, passing an explicitly constructed `vector<MetadataFlag>{Metadata::Thing, UserMetadataFlag}` to the new functions will transparently construct a `FlagCollection`?  Neat.  Since I use the boolean on the end of `GetNames`, I kind of flipped the "this breaks things" switch and went wild.  As you know, Parthenon changes which are really not part of the interface tend to break things in KHARMA for hacky reasons :)

Yurlungur: > As in, passing an explicitly constructed vector<MetadataFlag>{Metadata::Thing, UserMetadataFlag} to the new functions will transparently construct a FlagCollection?

Yep :)

tst/unit/test_metadata.cpp
Yurlungur: Implementing @bprather 's suggestion of `GetUserFlag`

src/interface/metadata.cpp
Yurlungur: I like this. It will be so.

doc/sphinx/src/metadata.rst
Yurlungur: Done.

doc/sphinx/src/metadata.rst
Yurlungur: Done.

src/interface/metadata.cpp
Yurlungur: Since you're the only one using this @bprather I am fine with changing the interface. Renaming to the suggested name.

Yurlungur:
OK I think advection test is fixed now and I've addressed @bprather and @pdmullen 's comments. This is ready for (re-)review. Might want to try it in a downstream code as a sanity check. Does anyone from AthenaPK want to take a look? @forrestglines @pgrete ?

Yurlungur:
ok well... Now sparse advection is failing. I'll check on that too...

Yurlungur:
The problem with sparse advection was a fascinating edge case:
- It's important that variables in packs have consistent ordering. To enable this, I essentially implemented a heap-sort, by storing the variables to pack in a `std::set<std::shared_ptr<CellVariable<Real>>>`. The set is automatically ordered, as it's a heap under the hood. 
- The same machinery is also used to register variables in the in the mesh refinement infrastructure. (Aside: I am going to create an issue about that infrastructure, as it's a mess.)
- This works quite nicely for dense variables. But for sparse variables, there's a problem. The ordering decided by `std::set<std::shared_ptr<CellVariable<Real>>>` is the *pointer-arithmetic* ordering of the shared pointer, nothing to do with variables. What this means is that if a variable is added or removed, or de-allocated, this ordering changes. You can also imagine this happening if the pointers are ordered differently on different MPI ranks or something. There doesn't seem to be a guarantee of safety here.
- Instead the ordering needs to be enforced some other way.
- One obvious solution is sort by variable name--a lexical sort. This is what we were originally doing and it involves using string comparisons. This would certainly work.
- A more performant solution is to compute a unique identifier for each variable. This is achieved by mapping each full variable name (base name + sparse id) to an integer. This is deterministic and persistent, and enables integer comparisons rather than string or pointer comparisons in the heap sort, which is more performant.

This is now implemented in this PR, with the spin-off feature that variables now have unique IDs that can be used for comparisons, sorting, etc, which should avoid needless string operations.

Hopefully *now* all tests pass and the PR is once again ready for review.

Yurlungur:
The changes to `src/interface/update.hpp` make the `SumData` calls and friends more flexible, but more verbose specializations must be expressed in your driver task list so this MR is a *slightly* breaking change. If people don't like that, I can revert. It required a one line change in `phoebus`.

src/interface/update.hpp
```diff
@@ -66,8 +66,8 @@ TaskStatus UpdateWithFluxDivergence(T *data_u0, T *data_u1, const Real gam0,
                                     const Real gam1, const Real beta_dt);
 
 template <typename F, typename T>
-TaskStatus WeightedSumData(const std::vector<F> &flags, T *in1, T *in2, const Real w1,
-                           const Real w2, T *out) {
+TaskStatus WeightedSumData(const F &flags, T *in1, T *in2, const Real w1, const Real w2,
+                           T *out) {
```
Yurlungur, line 70: This is the breaking change that required 1 line change downstream.

Yurlungur:
Ping @pgrete on review. I have spoken to @lroberts36 about this PR and I think we're on the same page. So just waiting on a AthenaPK person.

doc/sphinx/src/interface/metadata.rst
```diff
@@ -159,7 +159,7 @@ classes may be allocated. The behaviours are the following:
    codes.*
 
 Application Metadata Flags
---------------------------
+~~~~~~~~~~~~~~~~~~~~~~~~~~
```
pgrete:
```suggestion
--------------------------
```

Why this change?
I think this subsection is at the same level as the ones before.

doc/sphinx/src/interface/metadata.rst
```diff
@@ -174,3 +174,130 @@ Applications can allocate their own flags by calling
...
+
+.. code:: cpp
+
+   MetadataFlag const my_app_flag = Metadata::GetUserFlag("MyAppFlag");
+
+Note that this call will return an error if a flag is requested that
+hasn't been registered.
+
+Flag Collections
+~~~~~~~~~~~~~~~~~
```
pgrete:
```suggestion
-----------------
```

As above, to me this is a subsection (`-----`) rather than a subsubsection.

doc/sphinx/src/interface/metadata.rst
```diff
@@ -174,3 +174,130 @@ Applications can allocate their own flags by calling
...
+
+The ``Metadata::FlagCollection`` class provides a way to express a desire for
+a collection of ``Parthenon`` fields that satisfy some combinations of
+``MetadataFlag`` s. In particular, a ``FlagCollection`` specifies for a
```
pgrete:
```suggestion
``MetadataFlag``\ s. In particular, a ``FlagCollection`` specifies for a
```

src/mesh/amr_loadbalance.cpp
```diff
@@ -934,16 +934,17 @@ void Mesh::PrepareSendCoarseToFineAMR(MeshBlock *pb, BufArray1D<Real> &sendbuf,
   auto alloc_subview = Kokkos::subview(sendbuf, std::make_pair(0, p));
   auto alloc_subview_h = Kokkos::create_mirror_view(HostMemSpace(), alloc_subview);
 
-  for (int i = 0; i < pb->pmr->pvars_cc_.size(); ++i) {
-    auto &cc_var = pb->pmr->pvars_cc_[i];
+  int i = 0;
+  for (auto &cc_var : pb->pmr->pvars_cc_) {
     alloc_subview_h(i) = cc_var->IsAllocated() ? 1.0 : 0.0;
-    int nu = cc_var->GetDim(4) - 1;
+    int nu = cc_var->GetDim(4) - 1; // TODO(JMM): looks like this only supports vectors?
```
lroberts36, line 940: Agreed. I think this issue has been called out in a couple other places in `amr_loadbalance.cpp` as well.

src/interface/variable.hpp
```diff
@@ -157,6 +164,13 @@ class CellVariable {
   const int sparse_id_;
   const std::array<int, 6> dims_, coarse_dims_;
 
+  // Machinery for giving each variable a unique ID that is faster to
+  // evaluate than a string. Assumes number of MPI ranks unchanged
+  std::size_t uid_;
+  // Inline static initialization is a C++17 feature. Otherwise static
+  // vars must be initialized in the implementation file.
```
lroberts36: I am not sure this comment is necessary since we assume C++17 in Parthenon now. I would say something about this needing to be a global object so that different instances of `Variable` have the same `uid`s

src/interface/variable.hpp
```diff
@@ -157,6 +164,13 @@ class CellVariable {
   const int sparse_id_;
   const std::array<int, 6> dims_, coarse_dims_;
 
+  // Machinery for giving each variable a unique ID that is faster to
+  // evaluate than a string. Assumes number of MPI ranks unchanged
```
pgrete: Does this mean the sim cannot be restarted with a different number of MPI ranks?

src/mesh/meshblock.cpp
```diff
@@ -179,25 +179,25 @@ void MeshBlock::Initialize(int igid, int ilid, LogicalLocation iloc,
...
+          .vars();
+  for (const auto &v : vars) {
+    RegisterMeshBlockData(v);
   }
 
+  // No RemeshComm
   if (pm->multilevel) {
-    const auto refine_vars =
-        real_container
-            ->GetVariablesByFlag({Metadata::Independent, Metadata::FillGhost}, false)
```
pgrete, line 194: Did the meaning of the `true/false` change between `GetVariablesByFlag({Metadata::Independent, Metadata::FillGhost}, false)` and the `flags({Metadata::Independent, Metadata::FillGhost}, true);` used from above?

src/mesh/meshblock.cpp
```diff
@@ -179,25 +179,25 @@ void MeshBlock::Initialize(int igid, int ilid, LogicalLocation iloc,
...
-            .vars();
+    const auto refine_vars = real_container->GetVariablesByFlag(flags).vars();
     pmr = std::make_unique<MeshRefinement>(shared_from_this(), pin);
     // This is very redundant, I think, but necessary for now
-    for (int n = 0; n < refine_vars.size(); n++) {
+    for (const auto &v : vars) {
```
pgrete: Why is this now going over `vars` and not `refine_vars`?

src/interface/metadata.hpp
```diff
@@ -193,6 +196,108 @@ class Metadata {
...
+  // - ALL of the "intersections" flags,
+  // - and NONE of the "exclusions" flags
+  class FlagCollection {
+   public:
+    FlagCollection() = default;
+    // Container_t should be a stdlib container type
+    // it needs at least the iterator first and last methods
+    template <template <class...> class Container_t, class... extra>
+    FlagCollection(const Container_t<MetadataFlag, extra...> &flags,
+                   bool take_union = false) {
```
lroberts36, line 216:
```c++
template<T, REQUIRES(std::is_same<T::value_type, MetadataFlag>::value)>
FlagCollection(const T &flags, bool take_union = false)
```
I *think* that should work and get rid of the `std::initializer_list` overload. Additionally, I think it is easier to parse than the template template parameters.

src/interface/meshblock_data.cpp
```diff
@@ -70,6 +73,8 @@ void MeshBlockData<T>::AddField(const std::string &base_name, const Metadata &me
   }
 }
 
+// TODO(JMM): To use the set logic in GetVariablesFromFlags
+// that logic would need to be generalized. Not bothering for now.
```
lroberts36: Based on our discussion, I don't think we want to add the option to pass flags here. Rather, the list of desired variables can be determined elsewhere, potentially using sets based on `Metadata` flags

src/interface/meshblock_data.cpp
```diff
@@ -141,6 +146,8 @@ MeshBlockData<T>::MeshBlockData(const MeshBlockData<T> &src,
   CopyFrom(src, true, names, {}, sparse_ids);
 }
 
+// TODO(JMM): To use the set logic in GetVariablesFromFlags
+// that logic would need to be generalized. Not bothering for now.
```
lroberts36: See above.

pgrete (approved):
Looks good to me.
Pending that restart with different MPI ranks still work and that the `vars` <-> `refine_vars` question does what's expected, I'm already approving.

doc/sphinx/src/interface/metadata.rst
Yurlungur: Because it's a mistake. It happened automatically when I merged my old draft with the new docs.

Yurlungur: fixed

doc/sphinx/src/interface/metadata.rst
Yurlungur: fixed

src/interface/variable.hpp
Yurlungur: I am not sure what I meant when I wrote this comment. The basic idea here is a unique hash where variables are named in order of creation. I think that should be deterministically set at startup and not dependent on MPI rank number.

Let me double check it works with a restart with different # of MPI ranks though.

src/mesh/meshblock.cpp
Yurlungur: This is a bug. Not caught because I don't think we have any tests in `parthenon` where `ForceRemeshComm` is present. Fixed now.

src/mesh/meshblock.cpp
Yurlungur: Yes---I wanted the `FlagCollection` object to have the previous behavior as `GetVariablesByFlag` by default, which meant by default it's a set intersection.

But I also wanted the default behaviour of the constructor to be `true` so the optional bool in the `FlagCollection` constructor is `DoUnion` instead of `DoIntersection`. If this is too confusing, I can change the name/defaults so the meaning of `true/false` is retained.

src/interface/variable.hpp
Yurlungur: Ok confirmed. Here's the final value of one of the fields of sparse advection, started on 2 ranks and restarted on 4.

Yurlungur:
Thanks @pgrete . I want to double check that my response to the two unresolved issues makes sense. If so, will pull the trigger.

src/mesh/meshblock.cpp
```diff
@@ -179,25 +179,25 @@ void MeshBlock::Initialize(int igid, int ilid, LogicalLocation iloc,
...
   // variables that we may want to communicate are necessarily relevant
   // to the cost per meshblock.
-  const auto vars = real_container
-                        ->GetVariablesByFlag({Metadata::Independent, Metadata::FillGhost,
-                                              Metadata::ForceRemeshComm},
-                                             false)
-                        .vars();
-  for (int n = 0; n < vars.size(); ++n) {
-    RegisterMeshBlockData(vars[n]);
+  using FS_t = Metadata::FlagCollection;
```
lroberts36: Why `FS_t` and not `FC_t`?

lroberts36 (approved):
This looks good to me, although I am not sure we need to support set operations on flag collections since they don't exactly behave as expected (but we discussed this off line).

Yurlungur:
> although I am not sure we need to support set operations on flag collections since they don't exactly behave as expected (but we discussed this off line)

Yeah I think the path here is use them for now and migrate away when the longer-term solution discussed with sparse packs is implemented.

src/interface/variable.hpp
Yurlungur: Done

src/interface/metadata.hpp
Yurlungur: This does not get rid of the `std::initializer_list` overload but I think it does work.

I think there must be something special going on with initializer list that I don't understand, because I agree it should work based on my mental model of how this should work.

src/interface/meshblock_data.cpp
Yurlungur: Agreed. I think the thing to expose here would be a `std::vector` of unique IDs. I will ad a comment to that effect.

src/interface/meshblock_data.cpp
Yurlungur: :+1:

src/mesh/meshblock.cpp
Yurlungur: Artifact from an earlier draft when this type was called `FlagSet` not `FlagCollection`. Fixed.

src/interface/metadata.hpp
Yurlungur: Updated to the concept version but left the initializer list overload.

Yurlungur: Correction the linter doesn't like it. Reverting to the template template nonsense and making an issue, #839. Sorry.

Yurlungur:
Ok I think this is good to go. Enabling auto-merge.

PR #833, opened 2023-02-13
fix movie2d. Resolves #832.

### LINKED ISSUE #832: Movie python script not working
Trying to use the movie script (which I haven't used in ages) for a 2D problem with mesh refinement, I got 

whereas I expected sth like

image

Yurlungur:
## PR Summary

As @pgrete noted in #832 , the movie script is currently broken. This was due to the change to outputting more than just vector- or scalar-valued variables. Here I fix this by allowing the user to optionally pass in tensor components, which default to 0 for a scalar.

I wanted to also add a test to make sure the movie script was working. And there's a way to do this---we can generate a png file as a gold file and compare to that---but I wasn't sure if we wanted to pull in the additional dependency of Pillow. See [here](https://fabiorosado.dev/blog/comparing-images-with-numpy/). If people are ok with including Pillow as an additional python dependency, then we can add that test later. I don't have the time to do so right now.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
+++ b/scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
@@ -22,7 +22,8 @@ import matplotlib.pyplot as plt
 import matplotlib.patches as mpatches
 
 parser = ArgumentParser(
-    prog="movie2d", description="Plot snapshots of 2d parthenon output",
+    prog="movie2d",
+    description="Plot snapshots of 2d parthenon output",
 )
 parser.add_argument("field", type=str, help="field to plot")
 parser.add_argument(
--- a/scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
+++ b/scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
@@ -1,5 +1,5 @@
 # =========================================================================================
-# (C) (or copyright) 2020-2022. Triad National Security, LLC. All rights reserved.
+# (C) (or copyright) 2020-2023. Triad National Security, LLC. All rights reserved.
 #
 # This program was produced under U.S. Government contract 89233218CNA000001 for Los
 # Alamos National Laboratory (LANL), which is operated by Triad National Security, LLC
--- a/scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
+++ b/scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
@@ -154,7 +154,7 @@ if __name__ == "__main__":
     components = [0, 0]
     if (args.tc is not None) and (args.vc is not None):
         raise ValueError(
-            "Only one of --tensor-components and --vector-component should be set."
+            "Only one of --tensor-component and --vector-component should be set."
         )
     if args.tc is not None:
         components = args.tc
--- a/scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
+++ b/scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
@@ -32,11 +32,11 @@ parser.add_argument(
     default=None,
     help=(
         "Vector component of field to plot. "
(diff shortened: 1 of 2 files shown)


scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
```diff
@@ -13,12 +13,41 @@
...
+    dest="vc",
+    type=float,
+    default=None,
+    help=(
+        "Vector component of field to plot. "
+        + "Mutually exclusive with --tensor-components."
+    ),
+)
+parser.add_argument(
+    "--tensor-components",
```
brryan: This is really `tensor_component` right, because you're pulling out e.g. the `[i,j]` component of a tensor field?

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
```diff
@@ -36,14 +65,30 @@ def read(filename, nGhost=0):
...
+    ye=None,
+    components=[0, 0],
 ):
 
     if xe is None:
         xe = xf
     if ye is None:
         ye = yf
 
+    # get tensor components
```
brryan, line 87: Could we add an `assert len(q.shape) <= 5` here, since technically rank-3 tensor data could arrive from the dump file?

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
```diff
@@ -97,8 +142,18 @@ def plot_dump(
...
+    args = parser.parse_args()
+    field = args.field
+    files = args.files
+    components = [0, 0]
+    if (args.tc is not None) and (args.vc is not None):
+        raise ValueError(
+            "Only one of --tensor-components and --vector-component should be set."
+        )
+    if args.tc is not None:
+        components = tc
```
brryan: Does this need to be `components = args.tc`? Maybe I'm missing something.

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
```diff
@@ -97,8 +142,18 @@ def plot_dump(
...
+    if args.vc is not None:
+        components = [0, vc]
```
brryan: Similarly, `args.vc` here?

brryan (approved):
Thanks for fixing this @Yurlungur, sorry I didn't notice this. A few small comments if you agree with them and want to make those changes, but I'll go ahead and approve now.

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
Yurlungur: no you're right thanks for the catch

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
Yurlungur: Good idea

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
Yurlungur: Good idea :+1:

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
Yurlungur: Yep thanks for the catch.

pgrete (approved):
Perfect, thanks for the quick fix and the `ArgumentParser` is definitely an improvement.

pgrete:
Regarding image diffs, I also thought about those in the context of #810 so I'm in favor of adding those.
I'll open a new issue: https://github.com/parthenon-hpc-lab/parthenon/issues/834

Issue #836, opened 2023-02-17
Open doc issues


pgrete:
Here are a couple of items I noted when refactoring the docs:

- [ ] AMR criteria incomplete/outdated
- [ ] Hierarchy of headings in boundary communication potentially incorrect
- [ ] a LANL person should check if the LANL system specific instructions are still up-to-date
- [ ] Need better structure, e.g.,
  - Getting Started
  - Building/installing
  - Parthenon dev doc
  - Downstream dev doc
  - Usage/input reference
  - Performance
- [ ] need full parameter list
- [ ] need full configure list: already identified missing are
  - TEST_NUMPROC_FLAG
  - SERIAL_WITH_MPIEXEC
- [ ] disentangle/streamline info from `state.md` and more detailed description in separate files (e.g., multiple classes/interface are described twice)
- [ ] "Parthenon documentation" (current `README.rst`) needs to be removed and content separated/organized.
- [ ] Add doc for using docker containers (that already include dependencies)
- [ ] ... to be continued

PR #837, opened 2023-02-17
Migrate docs to sphinx

pgrete:
## PR Summary

~~This is more tedious than anticipated...
Not going to finish today,~~

Ready for review.

I  have not added any new content and preserved the structure almost 1:1.
So I suggest to focus on the deployed version, see https://parthenon-hpc-lab.github.io/parthenon/pgrete/migrate-docs/index.html and just check if there's not major screw-up that I missed.

For the (to follow in a separate PR -- potentially by other people) fine-tuning/updates/reorg, I started to collect item that I noticed: https://github.com/parthenon-hpc-lab/parthenon/issues/836

Finally, I made two technical changes to the docs

1. For the deployments, warnings are not shown as errors and the GitHub Action should fail if the docs don't build without warning (e.g., wrong labels, which caught me off guard a couple of times)

2. I removed the `sectionautolabel` extension as it resulted in quite a mess (with duplicated labels and inconsistent referencing). Based on what I've read in the wake of that mess is that quite often people recommend to be specific about labels, i.e., add one where it's needed (similar to latex) rather than rely on headings (that may change, and thus break links).
So for links within the same file, we can continue to use ``` `could be a section title`_ ``` but across files, we should use the explicit ``` :ref:`my label` ``` notation.

I hope people are fine with those changes.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


Yurlungur:
Thank you putting up with the tedium!

pgrete:
Ready to review, see details above.

.github/workflows/docs.yml
```diff
@@ -30,7 +30,8 @@ jobs:
           run: |
             echo "Repo = ${GITHUB_REPOSITORY}"
             cd doc/sphinx
-            make multiversion
+            # throw warnings as error to make sure docs properly build
+            make multiversion SPHINXOPTS="-W --keep-going -n"
```
Yurlungur, line 34: This is a very good idea

doc/sphinx/conf.py
```diff
@@ -27,7 +27,7 @@
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
-extensions = ["sphinx.ext.autosectionlabel", "sphinx_multiversion"]
```
Yurlungur, line 30: why this change

doc/sphinx/index.rst
```diff
@@ -38,8 +33,11 @@ Community
 .. toctree::
    :maxdepth: 1
    :caption: Contents:
+   :glob:
 
-   src/sphinx-doc
+   src/*
+   src/interface/*
+   src/mesh/*
```
Yurlungur, line 40: 👍

doc/sphinx/src/README.rst
```diff
@@ -0,0 +1,319 @@
...
+Building parthenon
+------------------
+
+See the :ref:`build doc <building>` for details on building Parthenon
+for specific systems.
+
+Description of examples
+-----------------------
+
+-  `Calculate π <https://github.com/parthenon-hpc-lab/parthenon/blob/develop/example/calculate_pi>`_
```
Yurlungur, line 16: does unicode work here?

c-prather (approved):
This looks like a clean port to me.  I also immediately noticed some of the things in #836, so the faster we can get this merged and start working on that...

tagardiner (approved):
I read through the documentation in the Sphinx format.  This is a significant improvement over the original markup pages.  I noticed a few details in reading through, like links that should probably be changed to github.com/parthenon-hpc-lab/parthenon/..., but we can get to that later.  The only other thing I noted is the link to "Module Index" on the home-page links to a non-existent page.  Thanks!

doc/sphinx/src/README.rst
```diff
@@ -0,0 +1,319 @@
+Parthenon documentation
```
Yurlungur: Does it still make sense for this file to be called `README.rst`? I would suggest something like overview. Similarly, I don't think this page should be titled `Parthenon documentation`, as that's a little vague. It's all documentation.

My suggestion is:
```suggestion
Overview
```

doc/sphinx/src/sphinx-doc.rst
```diff
@@ -24,29 +24,27 @@ Building documentation locally
...
+python. These instructions also *do not* require admin access and are usable
 with shared machines or python distributions.
 
 First, ensure that you are running a modern version of python (i.e. python 3 of
 some flavor)
 
-.. code-block::
-   :language:bash
+.. code-block:: bash
+
```
Yurlungur, line 34: This is on me thanks for the fix. I had fixed this in my metadata PR. But too late for that now. :)

Yurlungur (approved):
Thanks for going through all this work! Looks good to me with a minor nitpick or two below.

One big-picture thing: These files are not in any discernable order as far as I can tell. We might want to order them somehow at some point.

doc/sphinx/conf.py
pgrete: I removed the sectionautolabel extension as it resulted in quite a mess (with duplicated labels and inconsistent referencing). Based on what I've read in the wake of that mess is that quite often people recommend to be specific about labels, i.e., add one where it's needed (similar to latex) rather than rely on headings (that may change, and thus break links).
So for links within the same file, we can continue to use `could be a section title`_ but across files, we should use the explicit :ref:`my label` notation.

doc/sphinx/src/README.rst
pgrete: At least the local build and the CI build properly render this for me.

doc/sphinx/src/README.rst
pgrete: Agreed, I already listed this as the second to last item in https://github.com/parthenon-hpc-lab/parthenon/issues/836
I'm happy to make the suggested small change here though.

doc/sphinx/conf.py
Yurlungur: Ok that makes sense to me.

pgrete:
I updated a couple of links, removed the non-existent module index, and renamed the overview.
Pulling the trigger once the docs deployed correctly.

PR #838, opened 2023-02-17
uid registry

### LINKED ISSUE #839: Implicit conversions for `FlagCollection`
cpplint enforces that single-argument constructors be marked explicit. This doesn't make sense in a heavily templated code because, for example, what is the meaning of marking this explicit?
```C++
template<typename T>
MyClass(const T &arg)
```
worse, marking it `explicit` breaks template resolution, as I discovered in #831 when trying to implement a simpler constructor based on @lroberts36 's simple concepts machinery. I suggest we remove this check from our copy of cpplint. (However, I wanted to make an issue, rather than just do it, as I think this one merits collaboration consensus.)

Yurlungur:
## PR Summary

This PR builds on and depends on #831 . Here I take the unique identifier machinery for variables and thread it through packing so that key logic now uses lists of unique IDs, which are integers, rather than strings. This is essentially a performance boost for now, though I'm not sure how much it matters in practice. It probably helps in the limit of many small meshblocks. However, in the long term, it exposes these unique identifiers for use in sparse packs.

@lroberts36 also expressed interest in reorganizing the structure of this PR slightly to further decouple the unique IDs from `MeshBlockData`, so this is a place for us to collaborate on that.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


lroberts36 (approved):
This LGTM @Yurlungur. Please take a look and see if you find any issues with my changes. 

I think I misunderstood from our conversation where the unique ID reqistry was being stored. Having it as a static variable in `CellVariable` seems like a good solution that is nicely decoupled. I was thinking that this was somehow stored in `MeshBlockData` objects, but that is clearly not the case. I am not sure the unique IDs are consistent across ranks, but I guess that doesn't matter since they aren't shared across ranks. 

The only real change I ended up making was exposing a static method in `CellVariable` by which you can add variable names to the name to id map externally and calling that from within `AddFieldImpl` in `StateDescriptor` so variable unique IDs are available as early as possible and can be queried without access to a cell variable itself. This should be all I need for the sparse pack update I think. 

I also fixed #839 here, as well as removed the compiler warnings about __host__ __device__ destructors by explicitly declaring the destructors. I also slightly renamed a couple of things. 

I will thread this stuff into the `SparsePack`s in a separate PR.

src/interface/metadata.hpp
```diff
@@ -207,26 +207,23 @@ class Metadata {
...
       if (take_union) {
         unions_.insert(flags.begin(), flags.end());
       } else { // intersection
         intersections_.insert(flags.begin(), flags.end());
       }
     }
     // Constructor that takes a brace-enclosed initializer list
+    // Required since the type of {...} cannot be deduced since it
+    // could be the type of any object that could be initialized by
+    // an initializer list
```
Yurlungur, line 222: Ah mystery solved!

src/interface/metadata.hpp
```diff
@@ -207,26 +207,23 @@ class Metadata {
...
     // TODO(JMM): The cast to to a vector here implies some extra
     // copies which aren't great. Don't do this too much I guess.
-    // Also I don't totally understand why the templated constructor
-    // above doesn't capture this one.
     FlagCollection(std::initializer_list<MetadataFlag> flags, bool take_union = false)
-        : FlagCollection(FlagVec(flags), take_union) {}
+        : FlagCollection(FlagVec(std::move(flags)), take_union) {}
```
Yurlungur, line 226: Thanks for the catch @lroberts36

src/interface/variable.hpp
```diff
@@ -118,7 +118,9 @@ class CellVariable {
   inline std::string getAssociated() { return m_.getAssociated(); }
 
   KOKKOS_FORCEINLINE_FUNCTION
-  std::size_t GetUniqueID() const { return uid_; }
+  Uid_t GetUniqueID() const { return uid_; }
+
+  static Uid_t GetUniqueID(const std::string &var_label) { return get_uid_(var_label); }
```
Yurlungur, line 123: It's a little unintuitive that this *sets* the unique ID if it's not present (my fault for the naming) but I think that's probably fine. It's a greedy get.

src/interface/variable.hpp
```diff
@@ -165,9 +167,8 @@ class CellVariable {
   const std::array<int, 6> dims_, coarse_dims_;
 
   // Machinery for giving each variable a unique ID that is faster to
-  // evaluate than a string. uid is determined by order of variable
-  // creation, which is deterministic.
-  std::size_t uid_;
+  // evaluate than a string. Assumes number of MPI ranks unchanged
```
Yurlungur:
```suggestion
  // evaluate than a string. Safe so long as the number of MPI ranks
  // does not change while the code is running (restarts are fine).
```

Yurlungur (commented):
Thanks @lroberts36 LGTM. I removed the `[WIP]` and will add changelog and check formatting one last time. Some comments below.

Yurlungur:
@pgrete this is ready for review and it needs one more pair of eyes. Do you think Athena++ eyes are necessary? 

Alternatively does anyone else want to give this a quick look? @bprather @pdmullen @forrestglines ?

Yurlungur:
Ok it looks like sparse advection regression is failing. I'll have a look at this to see what's going on.

src/interface/meshblock_data.hpp
```diff
@@ -148,18 +146,20 @@ class MeshBlockData {
                              "Couldn't find variable '" + label + "'");
     return it->second;
   }
+  std::shared_ptr<CellVariable<T>> GetCellVarPtr(const Uid_t &uid) const {
```
brryan, line 149: Maybe a `PARTHENON_DEBUG_THROWS(varUidMap_.count(uid), "Variable ID not found!");`?

src/interface/meshblock_data.hpp
```diff
@@ -148,18 +146,20 @@ class MeshBlockData {
...
+  CellVariable<T> &Get(const Uid_t &uid) const { return *(varUidMap_.at(uid)); }
 
-  int Index(const std::string &label) noexcept {
-    for (int i = 0; i < (varVector_).size(); i++) {
-      if (!varVector_[i]->label().compare(label)) return i;
-    }
-    return -1;
+  Uid_t UniqueID(const std::string &label) noexcept {
+    auto it = varMap_.find(label);
+    if (it == varMap_.end()) return 0; // Invalid ID
```
brryan: Is it worth promoting this magic number to a named quantity e.g. `vpack_types::invalidID`? I wouldn't necessarily be able to guess whether the right magic number is `0` or `-1` since I have to look elsewhere to know that the unique ID is unsigned.

src/utils/unique_id.hpp
```diff
@@ -19,20 +19,23 @@
 #include <unordered_map>
 
 namespace parthenon {
+using Uid_t = std::size_t; // change this to something else if needed
```
brryan, line 22: I do like that this is promoted to a new name

brryan (approved):
LGTM -- non-blocking comments only

src/interface/meshblock_data.hpp
Yurlungur: Just realized this isn't necessary because `map.at(val)` has bounds checking. I added it anyway though for a nicer error message.

src/interface/meshblock_data.hpp
Yurlungur: Done

src/CMakeLists.txt
```diff
@@ -247,6 +247,9 @@ set_target_properties(parthenon PROPERTIES SOVERSION ${parthenon_VERSION})
 
 target_compile_features(parthenon PUBLIC cxx_std_17)
 
+target_link_options(parthenon PUBLIC "$<$<AND:$<CXX_COMPILER_ID:GNU>,$<VERSION_LESS:$<CXX_COMPILER_VERSION>,9.1>>:-lstdc++fs>" )
+target_link_options(parthenon PUBLIC "$<$<AND:$<CXX_COMPILER_ID:Clang>,$<VERSION_LESS:$<CXX_COMPILER_VERSION>,9.0>>:-lc++fs>" )
```
Yurlungur: This change was introduced to support backwards compatibility for `std::format`, which was introduced in #848 . Neither @pgrete nor I noticed that 

> Using this library may require additional compiler/linker options. GNU implementation prior to 9.1 requires linking with -lstdc++fs and LLVM implementation prior to LLVM 9.0 requires linking with -lc++fs. 

I take responsibility. Wish I'd noticed the fine print.


This set of cmake generator expressions should enable older versions of gcc/clang to link appropriately. Probably should be tested more carefully but it seems to work for us locally.

Yurlungur:
> Ok it looks like sparse advection regression is failing. I'll have a look at this to see what's going on.

Thanks to @lroberts36 I believe this is resolved. At this point it would be good to get review from someone on the AthenaPK or Kharma sides. @pgrete @BenWibking @forrestglines @bprather

src/bvals/cc/build_boundary_buffers.cpp
```diff
@@ -47,7 +47,7 @@ void BuildBoundaryBufferSubset(std::shared_ptr<MeshData<Real>> &md,
                                Mesh::comm_buf_map_t &buf_map) {
   Mesh *pmesh = md->GetMeshPointer();
   ForEachBoundary<BTYPE>(md, [&](sp_mb_t pmb, sp_mbd_t /*rc*/, nb_t &nb, const sp_cv_t v,
-                                 const OffsetIndices &no) {
+                                 const OffsetIndices & /*no*/) {
```
forrestglines, line 50: This parameter doesn't get used? Any reason to keep this around or does this function fulfill an interface requirement?

src/CMakeLists.txt
forrestglines: Is testing older GCC's in the CI worth it to catch these sorts of issues? Probably not?

Yurlungur:
Ping @pgrete and @forrestglines this PR has been waiting for merge for over a month. Since it has two approving PRs from LANL folks, I am going to enable auto-merge if you don't say something by the end of the day next Monday.

forrestglines:
> Ping @pgrete and @forrestglines this PR has been waiting for merge for over a month. Since it has two approving PRs from LANL folks, I am going to enable auto-merge if you don't say something by the end of the day next Monday.
I'm halfway through a review on this one, we discussed this in the meeting today. I should have a review done today

Yurlungur:
Cool. Thanks!

src/CMakeLists.txt
c-prather: This doesn't compile under NVHPC, and probably won't under many compiler IDs we otherwise support (IntelLLVM != Clang, AOCC != Clang, IBM, etc, see [list](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_ID.html)).

Would it be possible to just try to link both libraries, and ignore failures?

c-prather (approved):
Using this in KHARMA now and it passes the tests.  Should be marked as breaks-downstream since it does change some return values I'd consider part of the "interface," which I was using.
Also, it looks like explicitly constructed `std::vector<MetadataFlag>()` are no longer valid in `GetByFlag`, which isn't a problem (use the bare initializer list) but is an incompatibility.

No other comments except the one about the `std::filesystem` bugfix

src/CMakeLists.txt
c-prather: After messing with `find_library` a while, I've decided that libraries are hard even if you don't special-case compilers, and it would be much easier to use something like `find_package(Filesystem)`, i.e., which someone else has already written.

Luckily these exist, I've had good luck with [this one](https://github.com/lethal-guitar/RigelEngine/blob/master/cmake/Modules/FindFilesystem.cmake), released under BSD license.  (There is a rather hiliarously old CMake [bug](https://gitlab.kitware.com/cmake/cmake/-/issues/17834) to make something like this standard). Happy to add the file and the couple of lines required to switch to this, if I get the okay.

Yurlungur:
> Also, it looks like explicitly constructed std::vector<MetadataFlag>() are no longer valid in GetByFlag, which isn't a problem (use the bare initializer list) but is an incompatibility.

Interesting. In my original design this worked. I assume it no longer works because we agreed to forbid implicit type conversions in our constructors. It should be possible to add overloads if this is desired.

src/CMakeLists.txt
Yurlungur: :+1: I'm in favor of this, and thanks for finding it! Please feel free to push directly to this branch.

src/interface/variable.hpp
forrestglines: Is there a conventional name for a greedy get? It seems like a word that should exist but I couldn't think of one.

forrestglines (approved):
Looks good to me. This shouldn't interfere with the current methods we use in AthenaPK using strings to get variables, correct?

UIDs should be the same from run to run so it won't interfere with debugging, right?

Yurlungur:
> This shouldn't interfere with the current methods we use in AthenaPK using strings to get variables, correct?
> 
> UIDs should be the same from run to run so it won't interfere with debugging, right?

Yep and yep. :)

src/bvals/cc/build_boundary_buffers.cpp
Yurlungur: It fulfills an interface requirement.

src/interface/variable.hpp
Yurlungur: Yeah same.

src/CMakeLists.txt
Yurlungur: > Is testing older GCC's in the CI worth it to catch these sorts of issues? Probably not?

It might be worth it... but the problem will just go away with time.

Yurlungur:
Thanks @forrestglines ! And thanks @bprather for the cmake fix. I'll pull the trigger.

PR #840, opened 2023-02-24
Generalize Integrator Infrastructure

Yurlungur:
## PR Summary

I am pushing on high-order methods on parthenon downstream. To prepare for this, I needed some higher-order integrators. For that reason, I've implemented several improvements to our integrator machinery, without breaking any backwards compatibility:

## Class hierarchy

 I have split the `StagedIntegrator` class into a class hierarchy. The base class contains a little shared functionality. There are currently two subclasses: 
-  `LowStorageIntegrator`, which contains the family of low-storage integrators used by Athena++ expressed in Shu Osher form. (This is what @pgrete  updated our integrators to some time ago.) 
- `ButcherIntegrator` which contains a few commonly-used Butcher tableaus plus a 10th order one for fun. The butcher tableau machinery is introduced with IMEX and multi-rate integrators in mind down the line.

I also have templated the `MultiStageDriver` class on integrator type, with the default being `LowStorageIntegrator`, which returns to the old machinery. (To prevent people from having to type `template<>` when inheriting from the `MultiStageDriver`, I name the generic one `MultiStageDriverGeneric` and use a type alias to produce `MultiStageDriver`.)

The intent here is that if one knows they will be only a specific integration scheme, they should inherit from an iterative driver specialized to that specific integrator. However, the class hierarchy opens the possibility of switching between integration schemes at runtime.

In a similar vein, I have added some functions to `update.hpp` that can take the integrator object and do the update by reading it. I have *not* updated our examples or any downstream codes to use these functions, so they may have bugs. Nevertheless I'd like to merge them in if possible and fix them later as I work on my high-order machinery. If people object, I can remove them. I have some discussion points about this below.

## Additional changes:

- There was an ambiguity in our old machinery about stages vs. storage buffers. The integrator now reports not only how many stages it needs, but how many storage buffers.
- I have added a low-storage version of SSPRK4(5) to the `LowStorageIntegrator`. For `parthenon-hydro` and `AthenaPK` this should offer immediate benefits if you want to try it out.
- I have commented where exactly our equations for integration come from in more detail. I found the Athena++ paper wasn't enough for me to fully understand what was going on, so I dig into the low storage paper that is referenced there.
- Our integration coefficients were not tested in isolation. I added a set of unit tests for all our integration schemes.

## Questions

1. I do not believe our methods ever worked in "true" shu osher form, only in minimal storage form. We were lucky that for RK1 and RK2 the minimal storage and true shu osher forms are the same. Should we also implement non-minimal storage SSP integrators with shu osher alpha/beta matrices? (I am probably not going to do this now, but food for thought.)
2. The integration scheme used in our examples and the burgers' benchmark are not totally inline with the minimal storage approach (see point 1). They're a hybrid between shu osher notation and minimal storage 2S* form. I am considering updating them to use the more general integrators, but this will imply significant changes to the task lists for those examples. How do people feel about this?
3. I added some extra `update` functionality to leverage the integrators more easily. However, it's untested. Should we leave it out until I have a test case? Or should we add it now and fix it later after testing?
4. One possibility here is we merge the integrators + the update functionality in this MR and I update one of the examples in a subsequent MR. Thoughts?

@pgrete your thoughts on this PR in particular will be valuable given you're using the low-storage integrators downstream. 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


Yurlungur:
I still need to write updated docs, which I have not done yet.

Yurlungur:
This now pretty much ready for review.

brryan:
Kudos for making this change non-breaking. Only one comment now: isn't the `template<>` thing no longer needed in C++17?

Yurlungur:
> 

Oh yeah that's true---so maybe the `Generic` thing isn't needed in that case...

bensworth:
Simple naming convention, but I would not call classes integrators if they don't do any integration and only contain data

Yurlungur:
> Simple naming convention, but I would not call classes integrators if they don't do any integration and only contain data

This is a fair point. I left the name because that's what it was called before I touched it. If others are open to a different naming convention, I can change the name.

Yurlungur:
ping @pdmullen and @pgrete :)

doc/sphinx/src/integrators.rst
```diff
@@ -0,0 +1,94 @@
...
+:math:`s`.
+
+.. _Ketchson (2010): https://doi.org/10.1016/j.jcp.2009.11.006
+
+.. _Shu Osher: https://doi.org/10.1016/0021-9991(88)90177-5
+
+.. _Athena++ paper: https://doi.org/10.3847/1538-4365/ab929b
+
+The ``LowStorageIntegrator`` contains arrays for ``delta``, ``beta``,
+``gam0``, and ``gam1``. Available integratioon methods are:
```
pdmullen: integratioon -> integration

doc/sphinx/src/integrators.rst
```diff
@@ -0,0 +1,94 @@
...
+
+* ``RK1``, which is simply forward Euler
+
+* ``RK2``, which is Heun's method
+
+* ``VL2``
```
pdmullen: VL2 requires donor cell reconstruction for the predictor stage; maybe this should be documented here?

doc/sphinx/src/integrators.rst
```diff
@@ -0,0 +1,94 @@
...
+
+* ``RK3``, a strong stability preserving variant
+
+* ``RK4``, a strong stability preserving variant
+
+ButcherIntegrator
```
pdmullen, line 82: What is the use case for `ButcherIntegrator`? It is really neat to see a 10th order integrator working in Parthenon, but is it practical (i.e., will people use it?).  Also, why would someone use `ButcherIntegrator` for `RK1`, `RK2`, `RK4` instead of `LowStorageIntegrator `

tst/unit/test_unit_integrators.cpp
```diff
@@ -0,0 +1,251 @@
...
+      }
+    }
+    WHEN("We integrate with LowStorage rk4") {
+      constexpr Real dt = 1e-2;
+      auto integrator = MakeIntegrator<LowStorageIntegrator>("rk4");
+      State_t u;
+      GetInitialData(u);
+      Integrate(integrator, Step2SStar, tf, dt, u);
+      THEN("The final state doesn't differ too much from the true solution") {
+        REQUIRE(std::abs(u[0] - ufinal[0]) <= 1e-2);
```
pdmullen, line 200: thresholds differ for the Butcher integrators, however, low storage RK4 has the same threshold as lower storage RK1

pdmullen (approved):
Looks great, @jonahm-LANL 😸 I don't have too much to contribute... It looks like you mostly covered your bases with unit tests and CI is passing for this PR.  More generally, here are some of the thoughts I had regarding your infrastructure and higher order, in general:

(1) Round-off error conservation gets a bit weird with higher order integrators according to the Athena++ community.  See this issue: https://github.com/PrincetonUniversity/athena/issues/300.  Parthenon may want to address this in the future.  As implemented, I predict that higher order integrators with non-exact coefficients will see some departures from conservation. 
(2) I was initially thinking about how this PR might influence how one would add a sub-cycling integrator for split physics (e.g., super time stepping).  However, I am now not entirely convinced that this infrastructure is what one would use.  Instead, I wonder if the solvers infrastructure might instead be the better place for this kind of thing, e.g., for STS:
```
    auto &solver = tr[i].AddSuperTimeStep("RKL1", ...);
    auto sts = CreateTaskList(..., i, pmesh, tr, solver, mbase, miter, mout, dt);
```

doc/sphinx/src/integrators.rst
BenWibking: For us, the use case is to have the ability to drop (cell by cell) to first-order in space+time when high-order methods produce unphysical states. This is something we are considering in AthenaPK (https://github.com/parthenon-hpc-lab/athenapk/issues/29).

pdmullen: Couldn't you still do this with `LowStorageIntegrator` `RK1` (coupled with donor cell reconstruction, of course)?  Why do you need the `ButcherIntegrator` `RK1`?

BenWibking: I think you need the ButcherIntegrator RK2/RK3/RK4, because it keeps each stage, so you can then reconstruct the lower-order result for a small number of cells from the intermediate stages.

doc/sphinx/src/integrators.rst
Yurlungur: Good point. I'll also add the Stone + Gardiner reference.

doc/sphinx/src/integrators.rst
Yurlungur: I had IMEX and other multi-rate methods in mind, not all of which can be expressed in low-storage form.

tst/unit/test_unit_integrators.cpp
Yurlungur: Yes---that's because I chose a very large timestep. timestep for RK1 is 1e-5. Timestep for rk4 is 1e-2. I'll add a comment.

Yurlungur:
Ping one more person: @BenWibking @bprather @brryan @forrestglines @pgrete this PR needs one more approving review to go in. It has been waiting for review for almost six weeks. If I do not see any signs of life by next Monday I will force merge.

Yurlungur:
Thanks for the review, @pdmullen ! Regarding your comments:

> (1) Round-off error conservation gets a bit weird with higher order integrators according to the Athena++ community. See this issue: https://github.com/PrincetonUniversity/athena/issues/300. Parthenon may want to address this in the future. As implemented, I predict that higher order integrators with non-exact coefficients will see some departures from conservation.

Yes---I'm sure this is correct. Something to keep an eye on for sure.

> (2) I was initially thinking about how this PR might influence how one would add a sub-cycling integrator for split physics (e.g., super time stepping). However, I am now not entirely convinced that this infrastructure is what one would use. Instead, I wonder if the solvers infrastructure might instead be the better place for this kind of thing, e.g., for STS:

Yeah... I haven't thought carefully about this. I tried to make this infrastructure "ready" to extend for such things. But there's definitely questions to answer still as far as that goes.

c-prather (approved):
Using this in KHARMA and broke nothing.  I like making MultiStageDriver a specialization of a new generic version.
I have no way to test the higher-order methods implemented here independently, but the unit tests seem complete.

Yurlungur:
Thanks @pdmullen @bprather ! We now have two approvals. AthenaPK folks: @forrestglines @pgrete speak now-ish or forever hold your peace on the integrator MR. I'll enable auto-merge on Monday morning.

pgrete (approved):
Looks good and bonus points for the additional documentation.
I just found couple of typos that I just fixed directly.

Yurlungur:
Thanks @pgrete ! In that case I'm going to pull the trigger.

PR #843, opened 2023-03-05
Add some guard rails to prevent accidental registration of empty prolongation/restriction ops

Yurlungur:
## PR Summary

Recently @brryan and I tracked down a bug in Phoebus that was caused by the following pattern in a package initialization:

```C++
Metadata m({some set of flags});
if (condition) {
  m.Set(Metadata::FillGhost);
}
```

This was a problem because no prolongation/restriction ops were being set for the variable with this metadata. Default prolongation/restriction ops are provided, but are only automatically registered in the constructor if `IsRefined()` returns true at construct time, not later. So the metadata object was being created with empty prolongation/restriction ops and then set to prolongate and restrict at boundary communication time.

This wasn't a bug in parthenon per se, but I think it was an unexpected and unintuitive requirement on how the downstream code has to interact with the infrastructure.

This PR includes a number of guard rails for this:
1. A check in the `StateDescriptor` when the map from refinement function maps are created.
2. A check in `Restrict` that looks for nullpointers and provides a more useful error message.
3. Automatically set defaults for the prolongation and restriction ops in the state descriptor. Since these are only accessed if the variable `IsRefined`, this should be safer.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/src/interface/metadata.hpp
+++ b/src/interface/metadata.hpp
@@ -1,5 +1,5 @@
 //========================================================================================
-// (C) (or copyright) 2020-2022. Triad National Security, LLC. All rights reserved.
+// (C) (or copyright) 2020-2023. Triad National Security, LLC. All rights reserved.
 //
 // This program was produced under U.S. Government contract 89233218CNA000001 for Los
 // Alamos National Laboratory (LANL), which is operated by Triad National Security, LLC
--- a/src/interface/metadata.hpp
+++ b/src/interface/metadata.hpp
@@ -402,7 +402,7 @@ class Metadata {
 
   // Individual flag setters, using these could result in an invalid set of flags, use
   // IsValid to check if the flags are valid
-  // TODO(JMM):
+  // TODO(JMM): This is dangerous. See Issue #844.
   void Set(MetadataFlag f) { DoBit(f, true); }    ///< Set specific bit
   void Unset(MetadataFlag f) { DoBit(f, false); } ///< Unset specific bit
 
--- a/src/interface/metadata.hpp
+++ b/src/interface/metadata.hpp
@@ -331,11 +331,6 @@ class Metadata {
     if (CountSet({Independent, Derived}) == 0) {
       DoBit(Derived, true);
     }
-    // If variable is refined, set a default prolongation/restriction op
-    if (IsRefined()) {
-      refinement_funcs_ = refinement::RefinementFunctions_t::RegisterOps<
-          refinement_ops::ProlongateCellMinMod, refinement_ops::RestrictCellAverage>();
-    }
 
     // check if all flag constraints are satisfied, throw if not
     IsValid(true);
@@ -406,6 +401,7 @@ class Metadata {
 
   // Individual flag setters, using these could result in an invalid set of flags, use
   // IsValid to check if the flags are valid
+  // TODO(JMM): 
   void Set(MetadataFlag f) { DoBit(f, true); }    ///< Set specific bit
   void Unset(MetadataFlag f) { DoBit(f, false); } ///< Unset specific bit
 
@@ -625,7 +621,9 @@ class Metadata {
 
  private:
   /// the attribute flags that are set for the class
-  refinement::RefinementFunctions_t refinement_funcs_;
+  refinement::RefinementFunctions_t refinement_funcs_ =
+      refinement::RefinementFunctions_t::RegisterOps<
+          refinement_ops::ProlongateCellMinMod, refinement_ops::RestrictCellAverage>();
   std::vector<bool> bits_;
   std::vector<int> shape_ = {1};
   std::vector<std::string> component_labels_ = {};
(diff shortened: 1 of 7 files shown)


src/interface/metadata.hpp
```diff
@@ -625,7 +622,9 @@ class Metadata {
 
  private:
   /// the attribute flags that are set for the class
-  refinement::RefinementFunctions_t refinement_funcs_;
+  refinement::RefinementFunctions_t refinement_funcs_ =
+      refinement::RefinementFunctions_t::RegisterOps<
+          refinement_ops::ProlongateCellMinMod, refinement_ops::RestrictCellAverage>();
```
Yurlungur: I could have left this in the constructor, but this felt more appropriate, it can still be re-set by the user.

src/interface/state_descriptor.cpp
```diff
@@ -271,7 +271,7 @@ bool StateDescriptor::AddFieldImpl(const VarID &vid, const Metadata &m_in,
     return false; // this field has already been added
   } else {
     metadataMap_.insert({vid, m});
-    refinementFuncMaps_.Register(m);
+    refinementFuncMaps_.Register(m, vid.label());
```
Yurlungur, line 274: This is just to enable a more useful error message in the register function.

src/interface/state_descriptor.hpp
```diff
@@ -80,10 +81,21 @@ struct VarIDHasher {
...
       const auto &funcs = m.GetRefinementFunctions();
+      // Guard against uninitialized refinement functions by checking
+      // if the label is the empty string.
+      if (funcs.label().size() == 0) {
+        std::stringstream ss;
+        ss << "Variable " << varname << " registed for refinement, "
+           << "but no prolongation/restriction options found!"
+           << "Please register them with Metadata::RegisterRefinementOps." << std::endl;
+        PARTHENON_THROW(ss);
+      }
```
Yurlungur, line 98: If somehow the registered variable doesn't have a prolongation/restriction op, this will catch it and provide a useful error message. This code should be impossible to reach, but it's in setup so I think it costs us very little to have this check. Especially since we got caught off guard by this once already.

src/prolong_restrict/prolong_restrict.cpp
```diff
@@ -39,6 +40,7 @@ void Restrict(const StateDescriptor *resolved_packages,
   const auto &ref_func_map = resolved_packages->RefinementFncsToIDs();
   for (const auto &[func, idx] : ref_func_map) {
     auto restrictor = func.restrictor;
+    PARTHENON_DEBUG_REQUIRE_THROWS(restrictor != nullptr, "Valid restriction op");
```
Yurlungur, line 43: This should never trigger. Since it's not in setup I put it behind a `DEBUG` flag.

tst/unit/test_metadata.cpp
```diff
@@ -225,5 +225,12 @@ TEST_CASE("Refinement Information in Metadata", "[Metadata]") {
         REQUIRE_THROWS_AS(m.GetRefinementFunctions(), std::runtime_error);
       }
     }
+    WHEN("We set the refinement flag after the fact") {
+      m.Set(Metadata::FillGhost);
+      THEN("The refinement funcs should be present with a valid label") {
+        const auto &my_funcs = m.GetRefinementFunctions();
+        REQUIRE(my_funcs.label().size() > 0);
+      }
+    }
```
Yurlungur: Check that the fail-case downstream works. Note that if we apply the change I propose in #844 , this test becomes unnecessary/impossible again.

pgrete (approved):
LGTM for a quick fix but with a solution from #844 I'd not mind if setting default refinement ops is removed again.

The current design (in theory) allows the answer of `IsRefined()` and `funcs.label().empty()` to differ (which you test/check) and that doesn't feel great from a design point of view.

Yurlungur:
@pgrete with the guard rails I added, we could remove the default prolongation/restriction ops and it would still be an acceptable fix, since the code will no produce a more useful error message. I'd be fine with that if you think that's the better solution.

pgrete:
> @pgrete with the guard rails I added, we could remove the default prolongation/restriction ops and it would still be an acceptable fix, since the code will no produce a more useful error message. I'd be fine with that if you think that's the better solution.

Given that it's low effort, I'd go for that (just for consistency).
In the long run this should be gone with #844 anyway

Yurlungur:
> Given that it's low effort, I'd go for that (just for consistency).

Done.

Yurlungur:
@brryan @lroberts36 @jdolence @bprather can I get at least one more review?

lroberts36 (approved):
This looks good to me. The only thing I might suggest is adding a check to `Metadata::IsValid` that `refinement_funcs_` points somewhere valid if `Metadata::IsRefined()` is true. I guess you could also add a check to `Metadata::Set`, but maybe this has performance implications.

Yurlungur:
> This looks good to me. The only thing I might suggest is adding a check to `Metadata::IsValid` that `refinement_funcs_` points somewhere valid if `Metadata::IsRefined()` is true. I guess you could also add a check to `Metadata::Set`, but maybe this has performance implications.

:+1: Done

Issue #828, opened 2023-02-03
XDMF output for 2-, 3-tensor rank variables


brryan:
#781 changed the HDF5 output such that all tensor rank variables are no longer recast as vectors. This means XDMF output needs to specialize to each tensor rank. With #820 this will be the case for scalar and vector data, but 2- and 3-rank variables will not be supported. This should be straightforward to add, and would also be a good opportunity to include 2- and 3-rank variables in the regression tests somewhere.

PR #851, opened 2023-03-17
Fix xdmf hyperslab definition for vectors

pgrete:
## PR Summary

Our hyperslab definition was off resulting in errors like https://github.com/parthenon-hpc-lab/athenapk/issues/37

The fix set the hyperslab to what was originally planned, i.e., each "slab" is a (vector) component of size `nx3 x nx2 x nx1`.
Note that this just touches the xdmf, i.e. the way tools processing the xdmf file know how to load data from our hdf5 files.
The data itself was correct all along.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/src/outputs/parthenon_hdf5.cpp
+++ b/src/outputs/parthenon_hdf5.cpp
@@ -279,11 +279,11 @@ static void writeXdmfSlabVariableRef(std::ofstream &fid, const std::string &name
       // "3" rows for START, STRIDE, and COUNT for each slab with "5" entries.
       // START: iblock variable(_component)  0   0   0
       // STRIDE: 1               1           1   1   1
-      // COUNT:  1               dims[1]     nx3 nx2 nx1
+      // COUNT:  1               1           nx3 nx2 nx1
       fid << prefix << "    "
           << R"(<DataItem Dimensions="3 5" NumberType="Int" Format="XML">)" << iblock
           << " " << i << " 0 0 0 "
-          << " 1 1 1 1 1 1 " << dims[1] << " " << dims321 << "</DataItem>" << std::endl;
+          << " 1 1 1 1 1 1 1 " << dims321 << "</DataItem>" << std::endl;
       writeXdmfArrayRef(fid, prefix + "    ", hdfFile + ":/", name, dims, ndims, "Float",
                         8);
       fid << prefix << "  "
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
 - [[PR 663]](https://github.com/lanl/parthenon/pull/663) Change bvals_in_one to use sparse boundary buffers and add flux_correction in one.
 
 ### Fixed (not changing behavior/API/variables/...)
+- [[PR 851]](https://github.com/parthenon-hpc-lab/parthenon/pull/851) Fix xdmf hyperslab definition for vectors
 - [[PR 843]](https://github.com/parthenon-hpc-lab/parthenon/pull/843) Add guard rails to prolongation/restriction infrastructure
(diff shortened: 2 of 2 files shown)


Yurlungur (approved):
I always struggle a bit with reading XDMF and today is no exception. If I am reading this correctly, one of these hyperslab descriptions is written per vector/tensor component, and because of that, the extent should be 1--is that right?

If so, I approve.

pgrete:
> I always struggle a bit with reading XDMF and today is no exception. If I am reading this correctly, one of these hyperslab descriptions is written per vector/tensor component, and because of that, the extent should be 1--is that right?

yes, we want a "slab" (chunk of data) for each component as components may potentially be read independently from each other.
The current "count" for the data being read included all components of a vector (tensors are not supported as far as I can tell).
The fixed "count" now makes sure to just take "1" `dim4` index (set by the first "start" parameter).

src/outputs/parthenon_hdf5.cpp
```diff
@@ -279,11 +279,11 @@ static void writeXdmfSlabVariableRef(std::ofstream &fid, const std::string &name
...
       // START: iblock variable(_component)  0   0   0
       // STRIDE: 1               1           1   1   1
-      // COUNT:  1               dims[1]     nx3 nx2 nx1
+      // COUNT:  1               1           nx3 nx2 nx1
       fid << prefix << "    "
           << R"(<DataItem Dimensions="3 5" NumberType="Int" Format="XML">)" << iblock
           << " " << i << " 0 0 0 "
-          << " 1 1 1 1 1 1 " << dims[1] << " " << dims321 << "</DataItem>" << std::endl;
+          << " 1 1 1 1 1 1 1 " << dims321 << "</DataItem>" << std::endl;
       writeXdmfArrayRef(fid, prefix + "    ", hdfFile + ":/", name, dims, ndims, "Float",
```
brryan, line 287: My xdmf expertise is still lacking but just one question -- does this per-component chunking correctly handle named variables? In particular, should the argument to this function call be `names[i]` rather than `name`?

brryan (approved):
One question about component labeling -- I imagine since this works with AthenaPK that everything is fine but just wanted to check.

Yes, our XDMF implementation currently does not support rank-2 or rank-3 data. There is an open issue for this #828 

I'll go ahead and approve, thanks for fixing this @pgrete

src/outputs/parthenon_hdf5.cpp
pgrete: I'm also far from being an xdmf export -- more like a fire fighter putting out loose ends :D

So the output for AthenaPK now looks like
```
      <Attribute Name="Density" Center="Cell">
        <DataItem ItemType="HyperSlab" Dimensions="64 64 64 ">
          <DataItem Dimensions="3 5" NumberType="Int" Format="XML">0 0 0 0 0  1 1 1 1 1 1 1  64 64 64</DataItem>
          <DataItem Format="HDF" Dimensions=" 45 6 64 64 64" Name="cons" NumberType="Float" Precision="8">
            parthenon.restart.00000.rhdf:/cons</DataItem>
        </DataItem>
      </Attribute>
```
The main `Attribute` name is carrying the component label. I imagine that the `DataItem` with `Format=HDF` here refers to the entry  in the hdf file with with `cons` and that the name itself (of that `DataItem` line) is not relevant.

The current setup works in Visit and Paraview as is,so  I'm in favor of keeping it as is for now and revisit if this is causing problems.

brryan: OK that makes sense to me (`Attribute` versus `DataItem Name`), thanks for clarifying. Merging now 🤝

PR #852, opened 2023-03-24
Add UserMeshWorkBeforeOutput

BenWibking:
## PR Summary

This adds a version of `UserWorkBeforeOutput` that operates on the whole `Mesh` at a time, rather than per-`MeshBlock`. To avoid confusion with the existing `UserWorkBeforeOutput` function, this one is called `UserMeshWorkBeforeOutput`. Usage is demonstrated in the `advection` example.

This is necessary to support a use-case in AthenaPK where an MPI reduction is needed to fill the derived output vars.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


pgrete:
I like the idea to have this added to the test suite (though using the existing `UserWorkBeforeOutput` as discussed on Matrix).

BenWibking:
Since I can't run the formatting bot on forks, is it sufficient to run `make lint`/`ninja lint` for formatting and cpplint checks?

BenWibking:
I have added support for the new output function to the advection example, but there is no test that it actually works by reading the outputs. If this is needed, let me know how best to do this.

Yurlungur:
> Since I can't run the formatting bot on forks, is it sufficient to run `make lint`/`ninja lint` for formatting and cpplint checks?

run `make format` too.

Yurlungur (approved):
This all seems reasonable, but we should probably add the ability to register this hook to the docs somewhere.

Yurlungur:
> I have added support for the new output function to the advection example, but there is no test that it actually works by reading the outputs. If this is needed, let me know how best to do this.

Could add the output to the advection output test and check it against the gold file?

pgrete (commented):
Thanks for the PR.

In addition to the "should be mentioned in the docs" (probably both in `README.rst` and `mesh.rst`where the current callback functions are listed) comment, I'm wondering about the naming.
Any strong reason to give this a specific name `FillDerivedVars`?
Downstream codes might use this callback in a different context and there might be confusion with the other non-output related `FillDerived` callbacks.

My preference would be sth like `UserMeshWorkBeforeOutput` which is then called in https://github.com/parthenon-hpc-lab/parthenon/blob/3d566a34aa7ddbc936f624b21ab9550016de1059/src/mesh/mesh.cpp#L1022 along with the per block version.

Finally, @Yurlungur suggestion to just test this against the reference data should also be fairly straightforward, e.g., by reading the original and the derived variable with the `phdf` reader here https://github.com/parthenon-hpc-lab/parthenon/blob/3d566a34aa7ddbc936f624b21ab9550016de1059/tst/regression/test_suites/output_hdf5/output_hdf5.py#L119 and compare both by repeating the calculation in python.

PS: I'm happy to provide more pointers if needed.

Yurlungur:
> My preference would be sth like UserMeshWorkBeforeOutput which is then called in 

This is a good point.

BenWibking:
> My preference would be sth like `UserMeshWorkBeforeOutput` which is then called in
> 
> https://github.com/parthenon-hpc-lab/parthenon/blob/3d566a34aa7ddbc936f624b21ab9550016de1059/src/mesh/mesh.cpp#L1022
> 
> along with the per block version.

Renaming it makes sense, will do.

I can also call this from `ApplyUserWorkBeforeOuput` if it's okay to change the function signature of `ApplyUserWorkBeforeOutput` to add the `Mesh` and `SimTime` arguments.

BenWibking:
> Finally, @Yurlungur suggestion to just test this against the reference data should also be fairly straightforward, e.g., by reading the original and the derived variable with the `phdf` reader here
> 
> https://github.com/parthenon-hpc-lab/parthenon/blob/3d566a34aa7ddbc936f624b21ab9550016de1059/tst/regression/test_suites/output_hdf5/output_hdf5.py#L119
> 
> and compare both by repeating the calculation in python.
> PS: I'm happy to provide more pointers if needed.

Yes, a few more pointers on how to add a new test to the testing infrastructure would be helpful. Is there an example I can follow? I am a bit confused by the overall structure of the test infra Python code.

Yurlungur:
> I can also call this from `ApplyUserWorkBeforeOuput` if it's okay to change the function signature of `ApplyUserWorkBeforeOutput` to add the `Mesh` and `SimTime` arguments.

That's fine with me. I don't know if anyone is using `ApplyUserWorkBeforeOutput` or if it even works.

BenWibking:
Somehow the automatic merge was bad here, and now it won't build :(

BenWibking:
Ah, nevermind. That was downstream breakage due to https://github.com/parthenon-hpc-lab/parthenon/pull/867.

pgrete:
Alright, finally here are the promised pointers for testing:

1. add the new variable to the output block being written (i.e, part of the input file -- in this case I could imagine this to work for either the [`advection_outflow`](https://github.com/parthenon-hpc-lab/parthenon/tree/develop/tst/regression/test_suites/advection_outflow) or the [`output_hdf5`](https://github.com/parthenon-hpc-lab/parthenon/tree/develop/tst/regression/test_suites/output_hdf5) test as both call the advection example) 
2. given that there's a new variable, you  might need to update the gold standard file for the given test. Instructions are here: https://parthenon-hpc-lab.github.io/parthenon/develop/src/tests.html#gold-files
3. Add the data in the output file against each other by editing the python tests case, e.g., [here](https://github.com/parthenon-hpc-lab/parthenon/blob/94caccd69bce1ea4698351b2bf886f2103e16308/tst/regression/test_suites/advection_outflow/advection_outflow.py#L52)
  - I suggest update the return logic of the `Analyse` call similar to [here](https://github.com/parthenon-hpc-lab/parthenon/blob/94caccd69bce1ea4698351b2bf886f2103e16308/tst/regression/test_suites/output_hdf5/output_hdf5.py#L118) so that we can print an error message but keep the test going (which makes it easier to identify what's wrong) see, e.g., [here](https://github.com/parthenon-hpc-lab/parthenon/blob/94caccd69bce1ea4698351b2bf886f2103e16308/tst/regression/test_suites/advection_convergence/advection_convergence.py#L383)
  - Then add a new test for comparing reference data like [here](https://github.com/parthenon-hpc-lab/athenapk/blob/540deb985881c8f2e4a711194c3d022fd08ae671/tst/regression/test_suites/aniso_therm_cond_ring_multid/aniso_therm_cond_ring_multid.py#L111) where you get both orig and derived data and compare the derived data as calculated in python to the one in the `phdf` file directly.

Please let me know if this works as expected or if it's not clear.

BenWibking:
> Alright, finally here are the promised pointers for testing:
> 
> 1. add the new variable to the output block being written (i.e, part of the input file -- in this case I could imagine this to work for either the [`advection_outflow`](https://github.com/parthenon-hpc-lab/parthenon/tree/develop/tst/regression/test_suites/advection_outflow) or the [`output_hdf5`](https://github.com/parthenon-hpc-lab/parthenon/tree/develop/tst/regression/test_suites/output_hdf5) test as both call the advection example)
> 2. given that there's a new variable, you  might need to update the gold standard file for the given test. Instructions are here: https://parthenon-hpc-lab.github.io/parthenon/develop/src/tests.html#gold-files
> 3. Add the data in the output file against each other by editing the python tests case, e.g., [here](https://github.com/parthenon-hpc-lab/parthenon/blob/94caccd69bce1ea4698351b2bf886f2103e16308/tst/regression/test_suites/advection_outflow/advection_outflow.py#L52)
> 
> * I suggest update the return logic of the `Analyse` call similar to [here](https://github.com/parthenon-hpc-lab/parthenon/blob/94caccd69bce1ea4698351b2bf886f2103e16308/tst/regression/test_suites/output_hdf5/output_hdf5.py#L118) so that we can print an error message but keep the test going (which makes it easier to identify what's wrong) see, e.g., [here](https://github.com/parthenon-hpc-lab/parthenon/blob/94caccd69bce1ea4698351b2bf886f2103e16308/tst/regression/test_suites/advection_convergence/advection_convergence.py#L383)
> * Then add a new test for comparing reference data like [here](https://github.com/parthenon-hpc-lab/athenapk/blob/540deb985881c8f2e4a711194c3d022fd08ae671/tst/regression/test_suites/aniso_therm_cond_ring_multid/aniso_therm_cond_ring_multid.py#L111) where you get both orig and derived data and compare the derived data as calculated in python to the one in the `phdf` file directly.
> 
> Please let me know if this works as expected or if it's not clear.

~~@pgrete The new gold files (v21) are here: [parthenon_regression_gold_v21.tgz](https://github.com/parthenon-hpc-lab/parthenon/files/14104993/parthenon_regression_gold_v21.tgz)~~

These are wrong, because the snapshots did not include any of the variables.

Working on item (3) now.

BenWibking:
> 3. Add the data in the output file against each other by editing the python tests case, e.g., [here](https://github.com/parthenon-hpc-lab/parthenon/blob/94caccd69bce1ea4698351b2bf886f2103e16308/tst/regression/test_suites/advection_outflow/advection_outflow.py#L52)
> 
> * I suggest update the return logic of the `Analyse` call similar to [here](https://github.com/parthenon-hpc-lab/parthenon/blob/94caccd69bce1ea4698351b2bf886f2103e16308/tst/regression/test_suites/output_hdf5/output_hdf5.py#L118) so that we can print an error message but keep the test going (which makes it easier to identify what's wrong) see, e.g., [here](https://github.com/parthenon-hpc-lab/parthenon/blob/94caccd69bce1ea4698351b2bf886f2103e16308/tst/regression/test_suites/advection_convergence/advection_convergence.py#L383)

@pgrete I do not understand your suggestion for step 3a (above).

The python script checks whether there is any different at all between the advection_outflow final output and that from the gold standard:

```
        delta = compare(
            [
                "outflow.out0.final.phdf",
                parameters.parthenon_path
                + "/tst/regression/gold_standard/outflow.out0.final.phdf",
            ],
            check_metadata=False,
        )

        return delta == 0
```

Should this instead always `return True`, but set a flag somewhere? Since this is the only test in this python file, what does this accomplish?

Edit: I took a closer look at the `output_hdf5` test and I think I understand what you mean now.

tst/regression/test_suites/advection_outflow/advection_outflow.py
```diff
@@ -50,4 +50,18 @@ def Analyse(self, parameters):
...
+        try:
+            from phdf import phdf
+        except ModuleNotFoundError:
+            print("Couldn't find module to open Parthenon hdf5 files.")
+            return False
+
+        ## compute the derived var 'manually' and compare to the output derived var
+        data_filename = "outflow.out0.final.phdf"
+        data_file = phdf(data_filename)
+        q = data_file.Get("advected")[0]
```
BenWibking, line 68: @pgrete The test fails on this line with this error:
```
            ERROR in phdf.Get: Unable to read advected from file outflow.out0.final.phdf
```
I can't figure out why. Is there something obvious I've missed here?

forrestglines: Maybe `advected` needs to be capitalized to `Advected`? Wasn't that changed recently? It caused a failure in the yt frontend testing

BenWibking: That did not help:
```

            ERROR in phdf.Get: Unable to read Advected from file outflow.out0.final.phdf
            
  File "/Users/benwibking/parthenon/tst/regression/run_test.py", line 204, in <module>
    main(**vars(args))
  File "/Users/benwibking/parthenon/tst/regression/run_test.py", line 88, in main
    test_result = test_manager.Analyse()
                  ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/benwibking/parthenon/tst/regression/utils/test_case.py", line 278, in Analyse
    test_pass = self.test_case.Analyse(self.parameters)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/benwibking/parthenon/tst/regression/test_suites/advection_outflow/../../test_suites/advection_outflow/advection_outflow.py", line 62, in Analyse
    q = data_file.Get("Advected")[0]
        ~~~~~~~~~~~~~~~~~~~~~~~~~^^^
TypeError: 'NoneType' object is not subscriptable

```

BenWibking: The same thing is done here, using `"advected"`: https://github.com/parthenon-hpc-lab/parthenon/blob/a97831ac78ca31eed09b1a2daf6815042efc6f46/tst/regression/test_suites/output_hdf5/output_hdf5.py#L175

BenWibking: I will try moving this to the `output_hdf5` test.

src/mesh/mesh.cpp
```diff
@@ -903,7 +906,12 @@ void Mesh::EnrollBndryFncts_(ApplicationInput *app_in) {
...
-void Mesh::ApplyUserWorkBeforeOutput(ParameterInput *pin) {
+void Mesh::ApplyUserWorkBeforeOutput(Mesh *mesh, ParameterInput *pin,
+                                     SimTime const &time) {
+  // call Mesh version
+  mesh->UserMeshWorkBeforeOutput(mesh, pin, time);
+
+  // call MeshBlock version
   for (auto &pmb : block_list) {
     pmb->UserWorkBeforeOutput(pmb.get(), pin);
   }
```
pgrete, line 917: A topic for discussion might be if we want to allow for both versions to be enrolled simultaneously.
At least for for the initialization function, the general consensus was "no" because that would avoid potential race conditions with respect to the ordering of operations.

tst/regression/test_suites/advection_outflow/advection_outflow.py
pgrete: maybe the variable was not parsed correctly (see comment below) and thus didn't end up in the output file

tst/regression/test_suites/advection_outflow/parthinput.advection_outflow
```diff
@@ -64,4 +64,4 @@ num_vars = 1 # number of variables in variable vector
 <parthenon/output0>
 file_type = hdf5
 dt = 0.5
-variables = advected
+variables = advected my_derived_var
```
pgrete:
```suggestion
variables = advected, my_derived_var
```

tst/regression/test_suites/advection_outflow/advection_outflow.py
BenWibking: That was it.

BenWibking: I'll re-create the gold files with it included.

BenWibking:
@pgrete I've regenerated the gold files so that they actually have the output variables in them. Due to my earlier syntax error, they did not.

[parthenon_regression_gold_v21.tgz](https://github.com/parthenon-hpc-lab/parthenon/files/14143764/parthenon_regression_gold_v21.tgz)

The test passes with the additional derived variable, which is recomputed by the script and checked against its corresponding values in the file:

```console
$ ctest -R "regression_test:advection_outflow" --rerun-failed --output-on-failure
Test project /Users/benwibking/parthenon/build
    Start 66: regression_test:advection_outflow
1/1 Test #66: regression_test:advection_outflow ...   Passed   10.31 sec

100% tests passed, 0 tests failed out of 1

Label Time Summary:
regression    =  10.31 sec*proc (1 test)
serial        =  10.31 sec*proc (1 test)

Total Test time (real) =  10.32 sec
```

src/mesh/mesh.cpp
BenWibking: I'm not sure how to do this, since the function pointers are assigned in different constructors, so by this point, they are both non-`nullptr`, even if neither have been overridden by the application code.

I tried this:
```
// check that UserMeshWorkBeforeOutput and UserWorkBeforeOutput are not both enrolled
PARTHENON_REQUIRE(
      !((block_list[0]->UserWorkBeforeOutput != &UserMeshWorkBeforeOutputDefault) &&
          (mesh->UserMeshWorkBeforeOutput != &UserMeshWorkBeforeOutputDefault)),
      "UserMeshWorkBeforeOutput and UserWorkBeforeOutput cannot both be enrolled "
      "simultaneously!");
```
but the comparison operator for `std::function` is deleted, so it won't compile.

BenWibking:
It looks like the CI is currently failing because there's no release yet for the new gold files.

pgrete:
> It looks like the CI is currently failing because there's no release yet for the new gold files.

I just uploaded/released the new gold file. Let's see how the tests turn out now.

src/mesh/mesh.cpp
pgrete: I think you could mirror what's done, e.g., for `ProblemGenerator` or `PostInit`, in `meshblock.cpp`
```c++
  if (app_in->ProblemGenerator != nullptr) {
    ProblemGenerator = app_in->ProblemGenerator;
    // Only set default block pgen when no mesh pgen is set
  } else if (app_in->MeshProblemGenerator == nullptr) {
    ProblemGenerator = &ProblemGeneratorDefault;
  }
  if (app_in->PostInitialization != nullptr) {
    PostInitialization = app_in->PostInitialization;
    // Only set default post-init when no mesh post-init is set
  } else if (app_in->MeshPostInitialization == nullptr) {
    PostInitialization = &PostInitializationDefault;
  }
```
i.e., only set the default `MeshBlock` one if there's no `Mesh` version.
Then the check for both `!= nullptr` should work.

BenWibking:
@pgrete does the CI only run at certain times of the day? Or is it just backed up with other PRs?

pgrete:
The CI machine got restarted. Services are now back online (though it might be that the extended test won't trigger because the PR comes from a fork).
I'll try to trigger them now.

BenWibking:
@pgrete It looks like the "extended CI" failed. I don't see this on my local machine. Does this error message make sense to you?
```
--------------------------------------------
           Variables=['Blocks', 'Info', 'Input', 'Levels', 'Locations', 'LogicalLocations', 'Params', 'VolumeLocations', 'advected', 'my_derived_var']
--------------------------------------------

Ignoring metadata
____Comparing on a per variable basis with tolerance 1e-12
Tolerance = 1e-12
  advected            : no diffs
  my_derived_var      : no diffs
  File "/__w/parthenon/parthenon/tst/regression/run_test.py", line 204, in <module>
    main(**vars(args))
  File "/__w/parthenon/parthenon/tst/regression/run_test.py", line 93, in main
    raise TestError("Test " + test_manager.test + " failed")
__main__.TestError: Test advection_outflow failed
```

pgrete:
> @pgrete It looks like the "extended CI" failed. I don't see this on my local machine. Does this error message make sense to you?
> 
> ```
> --------------------------------------------
>            Variables=['Blocks', 'Info', 'Input', 'Levels', 'Locations', 'LogicalLocations', 'Params', 'VolumeLocations', 'advected', 'my_derived_var']
> --------------------------------------------
> 
> Ignoring metadata
> ____Comparing on a per variable basis with tolerance 1e-12
> Tolerance = 1e-12
>   advected            : no diffs
>   my_derived_var      : no diffs
> Traceback (most recent call last):
>   File "/__w/parthenon/parthenon/tst/regression/run_test.py", line 204, in <module>
>     main(**vars(args))
>   File "/__w/parthenon/parthenon/tst/regression/run_test.py", line 93, in main
>     raise TestError("Test " + test_manager.test + " failed")
> __main__.TestError: Test advection_outflow failed
> ```

Not sure what's going on.
It looks like things were fine in the test meaning that `return (delta == 0) and np.all(my_derived_var == file_derived_var)` triggered the failure.
I'll try to reproduce locally.

BenWibking:
@pgrete were you able to reproduce the CI failure locally?

pgrete:
> @pgrete were you able to reproduce the CI failure locally?

No... so I'm also puzzled why those fail.

pgrete:
Found the issue. The comparison resulted in difference in one ULP between the output and the derived calculation in Python.
This wasn't easy to see as the test result reported the joint result of both tests (and the provided output from the first test did not show an error).
I now split the error reporting between both tests and allow for one ULP difference between the in-code and post-processed derived variable.

pgrete:
All checks pass (except for the formatting one which still doesn't trigger for forks), so I'll force merge now.

PR #853, opened 2023-03-30
Improve the performance of movie2d.py script

tbhaxor:
## PR Summary

This PR solves a lot of things in the movie2d. Lets go one by one

1. Use a parallel processing using ProcessPoolExecutor. I tested with 10 workers, and it working fine. Although, I am printing the matplotlib threading warning. This improved the performance as shown below
    | Old Code | New Code |
    | :----:   | :----:   |
    |  |  | 

   > **Note** Right click on the images and open them in new-tab to get more clear picture
2. Optionally add time-step to show as the axes title
 
    

3. Check if field exists in the data for each file and break the for-loop on error
4. Remove unused and unaffecting (_idk if that is even a work_) functions
5. Accept the `-d` / `--output-directory` parameter to save the files in the specified directory. It is using pathlib.Path so the functions are pretty simplified as well.
6. Provide switching of `--debug` option from the arguments
7. Generate gif or mp4 movie from the output file images

Motivation: My simulation code used to run faster than movie generation process.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [ ] Docs build

### DIFF
--- a/scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
+++ b/scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
@@ -272,7 +272,7 @@ if __name__ == "__main__":
                 )
 
     if not ERROR_FLAG:
-        logger.info("All files are sent to the processor")
+        logger.info("All frames produced.")
 
         if not args.no_render:
             logger.info(f"Generating {args.movie_format} movie")
--- a/scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
+++ b/scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
@@ -110,8 +110,8 @@ parser.add_argument(
     default="output",
 )
 parser.add_argument(
-    "--no-render",
-    help="Prevent generating the movie from the parsed output files (default: false)",
+    "--render",
+    help="Generate the movie from the parsed output files (default: false)",
     default=False,
     action="store_true",
 )
@@ -274,7 +274,7 @@ if __name__ == "__main__":
     if not ERROR_FLAG:
         logger.info("All frames produced.")
 
-        if not args.no_render:
+        if args.render:
             logger.info(f"Generating {args.movie_format} movie")
             input_pattern = args.output_directory / "*.png"
             ffmpeg_cmd = f"ffmpeg -hide_banner -loglevel error -y -framerate {args.frame_rate} -pattern_type glob -i '{input_pattern}' "
--- a/scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
+++ b/scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
@@ -246,9 +246,7 @@ if __name__ == "__main__":
                 break
 
             q = data.Get(args.field, False, not args.debug)
-            stem = f"{args.prefix}_{str(dump_id).rjust(4, '0')}".strip()
-            stem = re.sub(r"^_", "", stem)
-            name = stem + ".png"
+            name = "{}_{:04d}.png".format(args.prefix, dump_id).strip()
             output_file = args.output_directory / name
 
             # NOTE: After doing 5 test on different precision, keeping 2 looks more promising
--- a/scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
+++ b/scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
@@ -80,8 +80,8 @@ parser.add_argument(
     metavar="PREFIX",
 )
 parser.add_argument(
-    "--debug",
-    help="Enable graph debug mode. (default: false)",
+    "--debug-plot",
+    help="Make plots with an exploded grid and including ghost zones. (default: false)",
     action="store_true",
     default=False,
 )
@@ -244,13 +244,13 @@ if __name__ == "__main__":
                 ERROR_FLAG = True
                 break
 
-            q = data.Get(args.field, False, not args.debug)
+            q = data.Get(args.field, False, not args.debug_plot)
             name = "{}_{:04d}.png".format(args.prefix, frame_id).strip()
             output_file = args.output_directory / name
 
             # NOTE: After doing 5 test on different precision, keeping 2 looks more promising
             current_time = format(round(data.Time, 2), ".2f")
-            if args.debug:
+            if args.debug_plot:
                 pool.submit(
                     plot_dump,
                     data.xg,
--- a/scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
+++ b/scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
@@ -215,8 +215,7 @@ if __name__ == "__main__":
             data = phdf(file_name)
             if args.field not in data.Variables:
                 logging.error(
-                    f'No such field "{args.field}" in {file_name}. \
-This will lead to stop further processing'
+                    f'No such field "{args.field}" in {file_name}. This will lead to stop further processing'
                 )
                 break
 
--- a/scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
+++ b/scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
@@ -15,8 +15,7 @@ from __future__ import print_function
 
 from argparse import ArgumentParser
 
-import os
-import sys
+# import os
 import numpy as np
 import matplotlib.pyplot as plt
 import matplotlib.patches as mpatches
@@ -50,11 +49,11 @@ parser.add_argument(
 parser.add_argument("files", type=str, nargs="+", help="files to plot")
 
 
-def addPath():
-    """add the vis/python directory to the pythonpath variable"""
-    myPath = os.path.realpath(os.path.dirname(__file__))
-    # sys.path.insert(0,myPath+'/../vis/python')
-    # sys.path.insert(0,myPath+'/vis/python')
+# def addPath():
+#     """add the vis/python directory to the pythonpath variable"""
+#     myPath = os.path.realpath(os.path.dirname(__file__))
+#     # sys.path.insert(0,myPath+'/../vis/python')
+#     # sys.path.insert(0,myPath+'/vis/python')
 
 
 def read(filename, nGhost=0):
@@ -147,7 +146,7 @@ def plot_dump(
 
 
 if __name__ == "__main__":
-    addPath()
+    # addPath()
     args = parser.parse_args()
     field = args.field
     files = args.files
--- a/scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
+++ b/scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
@@ -123,7 +123,8 @@ def plot_dump(
 
     fig = plt.figure()
     p = fig.add_subplot(111, aspect=1)
-    p.set_title(f"t = {time_title} seconds")
+    if time_title is not None:
+        p.set_title(f"t = {time_title} seconds")
 
     qm = np.ma.masked_where(np.isnan(q), q)
     qmin = qm.min()
@@ -204,7 +205,7 @@ if __name__ == "__main__":
         components = [0, args.vc]
 
     _x = ProcessPoolExecutor if args.worker_type == "process" else ThreadPoolExecutor
-    current_time = 0.0
+    current_time = 0.0 if args.time_step else None
     with _x(max_workers=args.workers) as pool:
         for dump_id, file_name in enumerate(args.files):
             data = phdf(file_name)
@@ -236,8 +237,9 @@ This will lead to stop further processing'
                 )
             else:
                 pool.submit(plot_dump, data.xng, data.yng, q, current_time, name, True)
-            current_time += args.time_step
-            current_time = round(current_time, ndigits=2)
+            if args.time_step:
+                current_time += args.time_step
+                current_time = round(current_time, ndigits=2)
 
     logger.info("All files are sent to the processor")
 
--- a/scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
+++ b/scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
@@ -25,30 +25,71 @@ from concurrent.futures import ThreadPoolExecutor, ProcessPoolExecutor
 import matplotlib.pyplot as plt
 import matplotlib.patches as mpatches
 
-logging.basicConfig(level=logging.CRITICAL,
-                    format='%(asctime)s [%(levelname)s]\t%(message)s')
+logging.basicConfig(
+    level=logging.CRITICAL, format="%(asctime)s [%(levelname)s]\t%(message)s"
+)
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.DEBUG)
 
-parser = ArgumentParser(prog="movie2d",description="Plot snapshots of 2d parthenon output")
+parser = ArgumentParser(
+    prog="movie2d", description="Plot snapshots of 2d parthenon output"
+)
 
-parser.add_argument("--vector-component", dest="vc", type=float, default=None,
-                    help="Vector component of field to plot. Mutually exclusive with --tensor-component.")
-parser.add_argument("--tensor-component", dest="tc", type=float, nargs=2, default=None, 
-                    help="Tensor components of field to plot. Mutally exclusive with --vector-component.")
-parser.add_argument("--workers", "-w", help="Number of parallel workers to use (default: 10)", 
-                    type=int, metavar="COUNT", default=10)
-parser.add_argument("--worker-type", help="Type of worker to use (default: process)", 
-                    choices=["process", "thread"], default="process", metavar="TYPE")
-parser.add_argument("--time-step", help="Value of dt parameter from parthenon/output* block", 
-                    metavar="DT", default=1.0, type=float)
-parser.add_argument("--output-directory", "-d", help=f"Output directory to save the images (default: {os.getcwd()})", 
-                    type=Path, default=os.getcwd(), metavar="DIR")
-parser.add_argument("--debug", help="Enable graph debug mode. This will run in the single thread. (default: false)", 
-                    action="store_true", default=False)
+parser.add_argument(
+    "--vector-component",
+    dest="vc",
+    type=float,
+    default=None,
+    help="Vector component of field to plot. Mutually exclusive with --tensor-component.",
+)
+parser.add_argument(
+    "--tensor-component",
+    dest="tc",
+    type=float,
+    nargs=2,
+    default=None,
+    help="Tensor components of field to plot. Mutally exclusive with --vector-component.",
+)
+parser.add_argument(
+    "--workers",
+    "-w",
+    help="Number of parallel workers to use (default: 10)",
+    type=int,
+    metavar="COUNT",
+    default=10,
+)
+parser.add_argument(
+    "--worker-type",
+    help="Type of worker to use (default: process)",
+    choices=["process", "thread"],
+    default="process",
+    metavar="TYPE",
+)
+parser.add_argument(
+    "--time-step",
+    help="Value of dt parameter from parthenon/output* block",
+    metavar="DT",
+    default=1.0,
+    type=float,
+)
+parser.add_argument(
+    "--output-directory",
+    "-d",
+    help=f"Output directory to save the images (default: {os.getcwd()})",
+    type=Path,
+    default=os.getcwd(),
+    metavar="DIR",
+)
+parser.add_argument(
+    "--debug",
+    help="Enable graph debug mode. This will run in the single thread. (default: false)",
+    action="store_true",
+    default=False,
+)
 
 parser.add_argument("field", type=str, help="field to plot")
 parser.add_argument("files", type=str, nargs="+", help="files to plot")
+
 
 def plot_dump(
     xf,
@@ -64,7 +105,6 @@ def plot_dump(
     ye=None,
     components=[0, 0],
 ):
-
     if xe is None:
         xe = xf
     if ye is None:
@@ -136,7 +176,6 @@ def plot_dump(
     logger.debug(f"Saved {time_title}s time-step to {output_file}")
 
 
-
 if __name__ == "__main__":
     # addPath()
     args = parser.parse_args()
@@ -148,8 +187,12 @@ if __name__ == "__main__":
         )
 
     if args.workers > 1:
-        logger.warning("Matplotlib is not multi-thread friendly. Read this for more details https://matplotlib.org/stable/users/faq/howto_faq.html#work-with-threads")
-        logger.warning("Try decreasing threads if you encounter any undefined behaviour")
+        logger.warning(
+            "Matplotlib is not multi-thread friendly. Read this for more details https://matplotlib.org/stable/users/faq/howto_faq.html#work-with-threads"
+        )
+        logger.warning(
+            "Try decreasing threads if you encounter any undefined behaviour"
+        )
     # Create output director if does't exists
     args.output_directory.mkdir(0o755, True, True)
     logger.info(f"Total files to process: {len(args.files)}")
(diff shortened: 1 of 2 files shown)


tbhaxor:
CPP linter didnt report any error in the files and also all the formatting related to `scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py` file is fixed in 8e273e24234e3b9ca1aa3426649be577f0783995

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
```diff
@@ -13,63 +13,90 @@
...
 parser.add_argument(
     "--vector-component",
     dest="vc",
     type=float,
     default=None,
-    help=(
-        "Vector component of field to plot. "
-        + "Mutually exclusive with --tensor-component."
-    ),
+    help="Vector component of field to plot. Mutually exclusive with --tensor-component.",
```
Yurlungur, line 44: You may have to change this back---black will want a particular formatting.

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
```diff
@@ -142,45 +172,75 @@ def plot_dump(
...
+                    data.yeg,
+                    components,
+                )
+            else:
+                pool.submit(plot_dump, data.xng, data.yng, q, current_time, name, True)
+            if args.time_step:
+                current_time += args.time_step
+                current_time = round(current_time, ndigits=2)
+
+    logger.info("All files are sent to the processor")
```
Yurlungur: This could be made more performant by having the worker in each pool also load the dump file and the dataset. A wrapper function around `plot_dump` and `phdf ` would need to be written. You could also use `pool.map` instead of `pool.submit`, which would be a little bit more concise.

Yurlungur (changes requested):
Thanks for submitting this! While having the ability to generate frames in parallel with processor pool is a nice addition, I do not think it should be the same script--as  people often use this script as a one-off to generate a few images or to just generate a single frame. My suggestion: rename `movie2d.py` to `snap2d.py`, and then add a movie script that imports `make_frame` and generates frames in parallel as you do here. This has a few advantages:
- The extra complexity of parallel processing is contained, so that folks can still read and use this plot without having to think about processor pools.
- There is a separation of concerns, which makes for more clean and maintainable code.
- In that model you could also call ffmpeg to generate the movie at the end in a seamless way.

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
```diff
@@ -13,63 +13,90 @@
...
+    default="process",
+    metavar="TYPE",
+)
+parser.add_argument(
+    "--time-step",
+    help="Value of dt parameter from parthenon/output* block",
+    metavar="DT",
+    default=1.0,
+    type=float,
+)
```
Yurlungur: Why use this? It would probably be better to simply pull the time of each frame out of the dump file.

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
```diff
@@ -142,45 +172,75 @@ def plot_dump(
...
+            if args.field not in data.Variables:
+                logging.error(
+                    f'No such field "{args.field}" in {file_name}. \
+This will lead to stop further processing'
+                )
+                break
+
+            logger.debug(f"Submitting {file_name}")
+            q = data.Get(args.field, False, not args.debug)
+            name = args.output_directory / Path(str(dump_id).rjust(4, "0") + ".png")
```
Yurlungur: Does the `/` operator append paths? I didn't know that.

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
```diff
@@ -13,63 +13,90 @@
...
+    help=f"Output directory to save the images (default: {os.getcwd()})",
+    type=Path,
+    default=os.getcwd(),
+    metavar="DIR",
+)
+parser.add_argument(
+    "--debug",
+    help="Enable graph debug mode. This will run in the single thread. (default: false)",
+    action="store_true",
+    default=False,
```
Yurlungur: Since the only change here is that it runs in serial, I would call this `--serial` not `--debug` unless it makes additional changes.

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
```diff
@@ -13,63 +13,90 @@
...
+    type=float,
+)
+parser.add_argument(
+    "--output-directory",
+    "-d",
+    help=f"Output directory to save the images (default: {os.getcwd()})",
+    type=Path,
+    default=os.getcwd(),
+    metavar="DIR",
+)
```
Yurlungur, line 75: I like the ability to output to a separate directory! But while we're at it, a filename prefix would probably be just as if not more useful... e.g., `--prefix run_name` would produce files like `run_name_0000.png`.

Yurlungur (commented):
A few more nitpicks below.

It's also worth noting that this script doesn't need to be super feature-complete. It's mostly a utility we use for small things.

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
lroberts36: I think this flag actually sets it to output the exploded plots that show ghost zones (which I called debug plots since they are mainly used to debug issues with downstream codes/Parthenon, potentially confusingly). It doesn't look like it actually gets run on a single thread when this argument is set (see line 222). 

I also think `args.debug` probably shouldn't set the level of logging at line 183.

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
tbhaxor: Yes it does that. Also it is well documented here https://docs.python.org/3/library/pathlib.html#operators

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
tbhaxor: I tried running this but black didn't complain or changed this. Either way it looks more readable to me.

tbhaxor: Do you still want me to revert to the older version for tc and vc?

tbhaxor:
> It's also worth noting that this script doesn't need to be super feature-complete. It's mostly a utility we use for small things.

Even I dont wanted to work on it earlier, but the time it takes to generate output for images is not efficient.

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
tbhaxor: Yeah I didn't know about it. Thanks for the information. It is fixed now

tbhaxor: Yeah I didn't know about it. Thanks for the information. It is fixed now

tbhaxor:
> It's also worth noting that this script doesn't need to be super feature-complete. It's mostly a utility we use for small things.

Even I dont wanted to work on it earlier, but the time it takes to generate output for images is not efficient.

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
tbhaxor: > I also think args.debug probably shouldn't set the level of logging at line 183.

Yes makes sense to me. It is fixed now

tbhaxor: > It doesn't look like it actually gets run on a single thread when this argument is set (see line 222).

That was a typo. I have also fixed that

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
tbhaxor: Added `--prefix`. By default it is `default_run`. You can test it

tbhaxor:
@Yurlungur Under which title do you think I should document this change in the CHANGELOG.md

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
tbhaxor: @Yurlungur I couldn't get what you are saying. Could you give more information here?

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
tbhaxor: While running the black, I couldnt get any warning/error or reformat updates from the black. It is more readable to me now. 

Do you want me to change it back to previous version?

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
tbhaxor: Ohk, but how can we check for the field and break the loop

```py
if args.field not in data.Variables:
                logging.error(
                    f'No such field "{args.field}" in {file_name}. \
This will lead to stop further processing'
                )
                break
```

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
Yurlungur: I prefer the older formatting but if black is happy, it's fine.

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
Yurlungur: Your version produces a cleaner error message, so that's fine. It's just a little bit more concise to construct an array of args to pass in and then `pool.map(make_frame_worker, args)`. You'll still get a useful error message out, but it'll be produced asynchronously by the worker that failed.

Although actually I just thought of something: Currently the `print(data)` line means `movie2d` outputs all the available variables so if I for example call
```python
movie2d.py nonexistent_var my_run.out0.00000.phdf
```
it will output the variables available to plot before the code dies. Your version does not do this. Any error message that runs the movie2d script should provide this so that users can easily get some introspection about the mesh.

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
Yurlungur: :+1:

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
Yurlungur: :+1:

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
Yurlungur: I would suggest a default of the empty string for the `prefix`.

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
Yurlungur: :+1: Probably the help message should not say anything about serial vs parallel then.

Yurlungur:
> Even I dont wanted to work on it earlier, but the time it takes to generate output for images is not efficient.

Has the movie script impeded your workflow? I can run through several hundred outputs in a pretty short amount of time, even in serial.

> @Yurlungur Under which title do you think I should document this change in the CHANGELOG.md

`Changed` I think

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
tbhaxor: ACK

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
tbhaxor: Yes, fixed already

tbhaxor:
Now it also generates movies 

Screenshot_20230405_180336

You can disable it using `--no-render` option

tbhaxor:
> Has the movie script impeded your workflow?

Not that much but on 1000+ files it is very slow

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
tbhaxor: I am now showing all the variables

Screenshot_20230405_180900

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
```diff
@@ -13,63 +13,118 @@
...
+parser.add_argument(
+    "--prefix",
+    help="Prefix for the file name to save",
+    default="",
```
Yurlungur, line 79: :+1:

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
```diff
@@ -13,63 +13,118 @@
...
+    "--movie-filename",
+    help="Basename of the output movie file. (default: output.FORMAT)",
+    default="output",
+)
+parser.add_argument(
+    "--no-render",
+    help="Prevent generating the movie from the parsed output files (default: false)",
+    default=False,
+    action="store_true",
+)
```
Yurlungur: I would switch the logic here. I.e., `--render` should be opt in and `--no-render` should be the default. I often use the script to generate a single frame, and I'm sure I'm not the only one.

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
```diff
@@ -142,45 +200,97 @@ def plot_dump(
...
+                    f'No such field "{args.field}" in {file_name}. This will lead to stop further processing'
+                )
+                logger.info(f"Available fields: {data.Variables}")
+                ERROR_FLAG = True
+                break
+
+            q = data.Get(args.field, False, not args.debug)
+            stem = f"{args.prefix}_{str(dump_id).rjust(4, '0')}".strip()
+            stem = re.sub(r"^_", "", stem)
+            name = stem + ".png"
```
Yurlungur: Regex and auxiliary variable here seems unnecessarily fancy. Why not just do the following?
```suggestion
            name = "{}_{:04d}.png".format(args.prefix, dump_id).strip()
```
sometimes old-style string-formatting is more clear and concise than f-strings. (Also the underscore for the prefix is probably desirable.)

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
```diff
@@ -142,45 +200,97 @@ def plot_dump(
...
-                data.gid,
-                data.xig,
-                data.yig,
-                data.xeg,
-                data.yeg,
-                components,
+
+    _x = ProcessPoolExecutor if args.worker_type == "process" else ThreadPoolExecutor
+    with _x(max_workers=args.workers) as pool:
+        for dump_id, file_name in enumerate(args.files):
```
Yurlungur: `dump_id` is a little confusing as the dumps may not align to the frame numbers. I'd say `frame_index` or `frame_idx` or something like that.

Yurlungur (changes requested):
Thanks for all these changes. Just a few more nitpicks below. It sounds like you prefer to do everything in the same script, rather than split it into a ` snap` script and a `movie` script?

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
```diff
@@ -142,45 +200,97 @@ def plot_dump(
...
+                    plot_dump, data.xng, data.yng, q, current_time, output_file, True
+                )
+
+    if not ERROR_FLAG:
+        logger.info("All files are sent to the processor")
+
+        if not args.no_render:
+            logger.info(f"Generating {args.movie_format} movie")
+            input_pattern = args.output_directory / "*.png"
+            ffmpeg_cmd = f"ffmpeg -hide_banner -loglevel error -y -framerate {args.frame_rate} -pattern_type glob -i '{input_pattern}' "
```
Yurlungur, line 287: Sometimes systems have `libav` and not `ffmpeg`. So it might be worth adding an option to switch ``ffmpeg` for `libav`.

tbhaxor: I have never heard of libav, but will explore an use it. Thanks for suggesting @Yurlungur

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
tbhaxor: Roger that 👍🏻

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
Yurlungur: :+1:

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
```diff
@@ -13,63 +13,118 @@
...
+    help="Prefix for the file name to save",
+    default="",
+    metavar="PREFIX",
+)
+parser.add_argument(
+    "--debug",
+    help="Enable graph debug mode. (default: false)",
+    action="store_true",
+    default=False,
+)
```
lroberts36: It may be better to have a more descriptive help message here, i.e. `help="Make plots with an exploded grid and including ghost zones. (default: false)"`. Maybe the option should be `--debug-plot` or something, just `--debug` is a bit confusing.

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
lroberts36: I agree that `--no-render` should be the default.

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
```diff
@@ -142,45 +200,95 @@ def plot_dump(
...
-    plt.savefig(name, dpi=300)
-    plt.close()
+    fig.savefig(output_file, dpi=300)
+    plt.close(fig=fig)
+    logger.debug(f"Saved {time_title}s time-step to {output_file}")
 
 
 if __name__ == "__main__":
-    addPath()
+    # addPath()
```
lroberts36:
```suggestion
```

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
```diff
@@ -142,45 +200,95 @@ def plot_dump(
...
-                data.yeg,
-                components,
+
+    _x = ProcessPoolExecutor if args.worker_type == "process" else ThreadPoolExecutor
+    with _x(max_workers=args.workers) as pool:
+        for dump_id, file_name in enumerate(args.files):
+            data = phdf(file_name)
+            if args.field not in data.Variables:
+                logger.error(
+                    f'No such field "{args.field}" in {file_name}. This will lead to stop further processing'
```
lroberts36:
```suggestion
                    f'No such field "{args.field}" in {file_name}. Further processing stopped.'
```

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
```diff
@@ -142,45 +200,95 @@ def plot_dump(
...
+                    data.xig,
+                    data.yig,
+                    data.xeg,
+                    data.yeg,
+                    components,
+                )
+            else:
+                pool.submit(
+                    plot_dump, data.xng, data.yng, q, current_time, output_file, True
+                )
```
lroberts36: Maybe I am missing something, but it looks like `components` is not getting passed here. I think that was maybe true of the old code as well, but I think it was a bug.

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
```diff
@@ -142,45 +200,95 @@ def plot_dump(
...
+
+    if not ERROR_FLAG:
+        logger.info("All files are sent to the processor")
```
lroberts36:
```suggestion
        logger.info("All frames produced.")
```

lroberts36 (approved):
This LGTM, but I agree with @Yurlungur that not producing a movie should be the default.

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
tbhaxor: Fixed

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
tbhaxor: Done

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
tbhaxor: It is not even passed in the develop branch. This is the latest commit in the develop branch.

https://github.com/parthenon-hpc-lab/parthenon/blob/a61999403a09a2946bd6ae40b5140169833b844a/scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py#L185

Yurlungur: I think what @lroberts36 is saying is that this is a bug in `develop`

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
tbhaxor: Done, change to `--render`

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
tbhaxor: I see, should I add it then?

Yurlungur: assume it doesn't break functionality, yes please.

lroberts36: yes, I think so

tbhaxor: @Yurlungur

Done, sir

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
```diff
@@ -142,45 +200,94 @@ def plot_dump(
...
+            if args.field not in data.Variables:
+                logger.error(
+                    f'No such field "{args.field}" in {file_name}. Further processing stopped.'
+                )
+                logger.info(f"Available fields: {data.Variables}")
+                ERROR_FLAG = True
+                break
+
+            q = data.Get(args.field, False, not args.debug_plot)
+            name = "{}_{:04d}.png".format(args.prefix, frame_id).strip()
```
Yurlungur: Sorry to change my mind on you. Probably you were right the first time, no need for the underscore.
```suggestion
            name = "{}{:04d}.png".format(args.prefix, frame_id).strip()
```

Yurlungur (approved):
Ok I think this is good to go, pending one last suggestion below. the libav vs ffmpeg thing can wait. Thanks for bearing with all the pickyness on this one.

Yurlungur:
Ok I'd like to make sure the script still works on some edge cases, which I will play with tomorrow (hopefully). After that, since @lroberts36 and I both approve, I can merge it.

tbhaxor:
> Ok I think this is good to go, pending one last suggestion below. the libav vs ffmpeg thing can wait. Thanks for bearing with all the pickyness on this one.

No problem at all sir 😄

tbhaxor:
@Yurlungur any updates?

Yurlungur:
Ok I'm happy. I'm enabling automerge.

tbhaxor:
There are some errors in the cuda11.6 container setup here thats causing auto merge to fail.

Yurlungur:
The CI machine currently is having issues. Should be resolved soon.

Yurlungur:
@tbhaxor I forgot our C++ and python formatting pipeline doesn't work on forks. Can you confirm you ran black on the script?

tbhaxor:
> @tbhaxor I forgot our C++ and python formatting pipeline doesn't work on forks. Can you confirm you ran black on the script?

Reformatted 181d896

Yurlungur:
Great thanks. I will force merge since all the other tests passed.

PR #764, opened 2022-10-10
WIP: Make fluxes separate variables

jdolence:
## PR Checklist

- [ ] Code passes cpplint
- [ ] New features are documented.
- [ ] Adds a test for any bugs fixed. Adds tests for new features.
- [ ] Code is formatted
- [ ] Changes are summarized in CHANGELOG.md
- [ ] CI has been triggered on [Darwin](https://re-git.lanl.gov/eap-oss/parthenon/-/pipelines) for performance regression tests.
- [ ] (@lanl.gov employees) Update copyright on changed files

PR #867, opened 2023-04-28
Rename `CellVariable` to `Variable`, etc.

lroberts36:
## PR Summary

I am going back to start on adding face, etc. variables. PR #764 has gotten substantially out of sync with develop, so I think it is easier at this point to just redo what is there. I plan to do what is in that PR over a few new, smaller PRs. 

This PR takes the first, simplest step and just removes `Cell` and `cc` from the names of things in most places. It also adds checks in the communication routines that variables with `Metadata::WithFluxes` and `Metadata::FillGhosts` set also have `Metadata::Cell` set. 

**I renamed the namespace `cell_centered_bvars` to `var_boundary_comm` and changed the directory `bvals/cc` to `bvals/bnd_flx_communication` so this is a breaking change for downstream codes.** 

There is nothing substantive here I think, but any comments on the naming choices I made are welcome. 

I didn't rename `MeshBlock::pvar_cc`, `MeshBlock::vars_cc_` and `MeshRefinement::pvars_cc_`, or any of the `cc` that show up in `amr_loadbalance.cpp` since these currently should only contain cell centered variables. 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [ ] Docs build


c-prather:
I'm happy with this.  Planning on pulling this into KHARMA before I put in a review but could pass it faster if we want to expedite things.

lroberts36:
@bprather: I think it would be helpful to make sure there isn't some issue with the renaming when it is pulled into KHARMA (and other downstream codes). There is no need for a quick review, I was just going to put up other PRs that depend on this one and I doubt they will go as quickly since this was essentially a search and replace.

example/stochastic_subgrid/stochastic_subgrid_driver.cpp
```diff
@@ -17,7 +17,7 @@
 
 // Local Includes
 #include "amr_criteria/refinement_package.hpp"
-#include "bvals/cc/bvals_cc_in_one.hpp"
+#include "bvals/bnd_flx_communication/bvals_in_one.hpp"
```
brryan: Just thinking out loud here... does `bnd_flx_communication` need to be this verbose/does this provide necessary deconfliction with other (potential) directories in `bvals` or could we just have something like `bvals/comms/bvals_in_one.hpp`? At any rate the `flx` without `var` here makes me think this folder is specifically for flux corrections. 

Or if we need to separate variable and swarm communication into separate folders, maybe `bvals/var_comms` and then later also `bvals/swarm_comms`?

I'm not demanding any names be changed here I'm just bringing this up and it's fine to ignore me on this.

example/stochastic_subgrid/stochastic_subgrid_driver.cpp
```diff
@@ -134,15 +134,15 @@ TaskCollection StochasticSubgridDriver::MakeTaskCollection(BlockList_t &blocks,
 
       const auto any = parthenon::BoundaryType::any;
 
-      tl.AddTask(none, parthenon::cell_centered_bvars::StartReceiveBoundBufs<any>, mc1);
-      tl.AddTask(none, parthenon::cell_centered_bvars::StartReceiveFluxCorrections, mc0);
+      tl.AddTask(none, parthenon::var_boundary_comm::StartReceiveBoundBufs<any>, mc1);
```
brryan: Is the `var_boundary_comm` namespace necessary? Was this just intended to separate cell-centered vars from others, and will this distinction be necessary to the current plan for face/edge fields?

src/bvals/boundary_conditions.cpp
```diff
@@ -221,14 +221,17 @@ void ProlongateGhostCells_(std::shared_ptr<MeshBlockData<Real>> &rc,
...
-    // TODO(LFR): Is this indexing correct for 5 and 6 dimensional fields?
-    int nu = cc_var->GetDim(4) - 1;
-    pmr->ProlongateCellCenteredValues(cc_var.get(), si, ei, sj, ej, sk, ek);
+  for (auto var : rc->GetVariableVector()) {
+    if (!var->IsAllocated()) continue;
+    if (!(var->IsSet(Metadata::Independent) || var->IsSet(Metadata::FillGhost))) continue;
+
+    if (var->IsSet(Metadata::Cell)) {
+      // TODO(LFR): Is this indexing correct for 5 and 6 dimensional fields?
+      int nu = var->GetDim(4) - 1;
```
brryan: `nu` is unused? At any rate I don't think prolongation/restriction has been tested for 5-, 6-D fields.

brryan (approved):
Looks good to me, I'm going to approve this now.

That said I did raise some minor questions about naming. Feel free to ignore but since this is a breaking change now is probably the time to optimize and future-proof (mainly for swarms) these names if such things are necessary.

example/stochastic_subgrid/stochastic_subgrid_driver.cpp
lroberts36: I think I agree with you that the `var_boundary_comm` namespace is unnecessary and was probably put there to distinguish between different types of fields. I guess the only place it could conflict is for swarms (?), since the plan is to unify communication of all types of grid variables. I will leave this up to see if others have thoughts, but if no one says anything I will remove the namespace.

example/stochastic_subgrid/stochastic_subgrid_driver.cpp
lroberts36: I like `comms` better, there was no real reason for the verbosity except that I couldn't think of anything better when I wrote it. I will make the switch.

src/bvals/boundary_conditions.cpp
lroberts36: yes, it is. I removed it.

example/stochastic_subgrid/stochastic_subgrid_driver.cpp
Yurlungur: I think it's fine to remove the namespace. Agree swarms might be a conflict. Defer to @brryan on that.

Yurlungur (approved):
I'm excited to see this finally moving!

example/stochastic_subgrid/stochastic_subgrid_driver.cpp
forrestglines: If we don't remove the name space, would `var_comms` be appropriate? It's slightly less verbose and more in line with the `comms` directory

brryan: Hmm... I think we don't need to preserve this namespace just to deconflict with swarms. Although for swarm packing we will probably want to move to a model like this, e.g. `StartReceiveSomething(std::shared_ptr<MeshData>)` , the communicated swarms aren't really "boundary buffers" (the particles being communicated are literally already on the meshblock they are being moved to) it would be natural for me to just use different names like `StartReceiveSwarmBufs` etc.

forrestglines (approved):
I tested the rename in AthenaPK, works fine with a simple find-and-replace in one file.  The full test suite is still running but it's passing so far.

Edge and face variable will all live together with cell variables under `Variable` with metadata to distinguish them, right?

On the rename of `bvals_cc_in_one.hpp` to `bvals_in_one.hpp`, do you foresee all variable boundary communications living under that header?

example/stochastic_subgrid/stochastic_subgrid_driver.cpp
lroberts36: Ok, I am going to remove the namespace entirely.

lroberts36:
> I tested the rename in AthenaPK, works fine with a simple find-and-replace in one file. The full test suite is still running but it's passing so far.
> 

Thanks for testing it out

> Edge and face variable will all live together with cell variables under `Variable` with metadata to distinguish them, right?
> 

Yeah, that is what is implemented in #868.

> On the rename of `bvals_cc_in_one.hpp` to `bvals_in_one.hpp`, do you foresee all variable boundary communications living under that header?

Yeah, for boundary communication variables of all types will be communicated when `*BoundBufs` are called (see #872).

c-prather (approved):
Using this successfully in KHARMA for a week now.  I don't have boundary communication working downstream, but all the naming was easy to update & I'm happy with the names, and the code works just fine.

PR #868, opened 2023-04-28
Add local face fields

lroberts36:
## PR Summary

This PR starts to implement face, edge, and nodal variables. Specifically, it makes the changes required within the `Variable` and `Metadata` classes to allow `Variable`s to store any type of field and allows these types of variables to be included in packs (both legacy and sparse packs). It does not deal with any non-block local operations on these types of variables (e.g. boundary communication, prolongation, restriction), mainly to keep the PR small. This borrows heavily from #764 and is built on #867. **Please review #867 first if you haven't already.** 

Todo: 
- [x] Include field type dependent shapes in `Metadata` and allocate `Variable` data and coarse data based on these shapes.
- [x] Move all `ParArray6D`s to `ParArrayND`s so we can add an extra index for indexing the face. 
- [x] Allow for non-cell centered variables in packs. **There may need to be a discussion of how we want this interface to look**
- [x] Add test to ensure new variables types are behaving as expected  

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [ ] Docs build


src/bvals/comms/bnd_info.hpp
```diff
@@ -69,9 +69,9 @@ struct BndInfo {
   Coordinates_t coords, coarse_coords; // coords
 
   buf_pool_t<Real>::weak_t buf;         // comm buffer from pool
-  ParArray6D<Real, VariableState> var;  // data variable used for comms
-  ParArray6D<Real, VariableState> fine; // fine data variable for prolongation/restriction
-  ParArray6D<Real, VariableState>
+  ParArrayND<Real, VariableState> var;  // data variable used for comms
+  ParArrayND<Real, VariableState> fine; // fine data variable for prolongation/restriction
+  ParArrayND<Real, VariableState>
```
Yurlungur, line 72: What's the reason for this change?

src/interface/sparse_pack.hpp
```diff
@@ -255,41 +255,43 @@ class SparsePack : public SparsePackBase {
   }
 
   // operator() overloads
-  KOKKOS_INLINE_FUNCTION
-  auto &operator()(const int b, const int idx) const { return pack_(0, b, idx); }
+  template <std::size_t DIR = 0>
+  KOKKOS_INLINE_FUNCTION auto &operator()(const int b, const int idx) const {
+    return pack_(DIR, b, idx);
+  }
```
Yurlungur: How does calling a templated `operator()` look in practice? `pack<DIR>(b, idx)`?

src/interface/variable.hpp
```diff
@@ -137,6 +138,35 @@ class Variable {
...
+    return d.Get(((void)I, DIR)..., std::forward<Args>(args)...);
+  }
+
+  template <std::size_t DIR, class... Args>
+  auto GetFluxTensorComponent(Args &&...args) {
+    return GetTensorComponentImpl<0>(
+        flux[DIR],
+        std::make_index_sequence<MAX_VARIABLE_DIMENSION - sizeof...(Args) - 3>(),
+        std::forward<Args>(args)...);
+  }
```
Yurlungur: I'm not sure I understand what these do.

src/interface/variable_pack.hpp
```diff
@@ -307,16 +308,34 @@ class VariablePack {
...
   ParArray3D<T, VariableState> &operator()(const int n) const {
     assert(IsAllocated(n));
-    return v_(n);
+    assert(v_[0](n).topological_type == TopologicalType::Cell);
+    return v_[0](n);
   }
 
+  KOKKOS_FORCEINLINE_FUNCTION
+  T &operator()(TopologicalElement el, const int n, const int k, const int j,
+                const int i) const {
```
Yurlungur, line 327: Do I always need to pass in a topological element? For face fields, for example, it feels like 1/2/3 should be ok?

src/basic_types.hpp
```diff
@@ -50,6 +52,42 @@ enum class BoundaryType : int {
...
+// NXYZ = edge at (i - 1/2, j - 1/2, k - 1/2)
+enum class TopologicalElement : std::size_t {
+  C = 0,
+  FX = 3,
+  FY = 4,
+  FZ = 5,
+  EYZ = 6,
+  EXZ = 7,
+  EXY = 8,
+  NXYZ = 9
```
Yurlungur: The jumping from 0 to 3 (so you can do modulo arithmetic) probably merits a comment as I think it'll confuse future us.

src/interface/variable_pack.hpp
```diff
@@ -558,9 +591,20 @@ void FillVarView(const VariableVector<T> &vars, bool coarse, ViewOfParArrays<T>
 
           host_al(vindex) = v->IsAllocated();
           if (v->IsAllocated()) {
-            host_cv(vindex) = coarse ? v->coarse_s.Get(k, j, i) : v->data.Get(k, j, i);
+            if (v->IsSet(Metadata::Face) || v->IsSet(Metadata::Edge)) {
+              host_cv0(vindex) = coarse ? v->template GetCoarseTensorComponent<0>(k, j, i)
+                                        : v->template GetTensorComponent<0>(k, j, i);
```
Yurlungur: why not make a wrapper function here that has `coarse` as a bool argument to `GetTensorComponent`? Might save a bit of boiler plate?

src/interface/variable_state.hpp
```diff
@@ -54,6 +61,10 @@ struct VariableState : public empty_state_t {
   int sparse_id;
   int vector_component = NODIR;
   bool initialized = true;
+
+  TopologicalType topological_type;
```
Yurlungur: Nice. :+1: I like storing this in the variable itself.

I think we should be explicit about the default unitialized type though. So I would make this equal to cell by default. (I know that's what it will be anyway, but I think that will improve readability, since this isn't visibly an integral type.)

src/utils/multi_pointer.hpp
```diff
@@ -0,0 +1,33 @@
...
+//========================================================================================
+#ifndef UTILS_MULTI_POINTER_HPP_
+#define UTILS_MULTI_POINTER_HPP_
+
+#include <type_traits>
+
+namespace parthenon {
+
+template <class T, int N>
+struct multi_pointer : multi_pointer<std::add_pointer_t<T>, N - 1> {};
```
Yurlungur, line 22: I can see this is for meta programming... but what is the use of the multi pointer type? A comment might be useful. Also a comment to emphasize this is for metaprogramming, not for actually creating some new class would probably be useful.

tst/unit/test_mesh_data.cpp
```diff
@@ -101,6 +104,15 @@ TEST_CASE("MeshData works as expected for simple packs", "[MeshData]") {
...
               });
         }
+        auto &var = pmbd->Get("v6");
+        par_for(
+            loop_pattern_mdrange_tag, "initialize v6", DevExecSpace(), kb.s, kb.e + 1,
+            jb.s, jb.e + 1, ib.s, ib.e + 1, KOKKOS_LAMBDA(int k, int j, int i) {
+              Real n = i + 1e1 * j + 1e2 * k + 1e3 * 0 + 1e4 * 4 + 1e5 * b;
+              if (k <= kb.e && j <= jb.e) var(0, 0, 0, 0, k, j, i) = 3 * n + 0;
+              if (k <= kb.e && i <= ib.e) var(1, 0, 0, 0, k, j, i) = 3 * n + 1;
+              if (j <= jb.e && i <= ib.e) var(2, 0, 0, 0, k, j, i) = 3 * n + 2;
```
Yurlungur, line 114: This is pretty unintuitive... Is there any way we don't need all these filler zeros for unused indices?

tst/unit/test_mesh_data.cpp
```diff
@@ -126,6 +138,37 @@ TEST_CASE("MeshData works as expected for simple packs", "[MeshData]") {
...
+        using te = parthenon::TopologicalElement;
+        par_reduce(
+            loop_pattern_mdrange_tag, "check imap, scalar", DevExecSpace(), 0,
+            pack.GetDim(5) - 1, kb.s, kb.e + 1, jb.s, jb.e + 1, ib.s, ib.e + 1,
+            KOKKOS_LAMBDA(int b, int k, int j, int i, int &ltot) {
+              Real n = i + 1e1 * j + 1e2 * k + 1e3 * c + 1e4 * v + 1e5 * b;
+              if (n != pack(b, v5lo, k, j, i) && k <= kb.e && j <= jb.e && i <= ib.e)
+                ltot += 1;
+              n = i + 1e1 * j + 1e2 * k + 1e3 * c + 1e4 * 4 + 1e5 * b;
+              if (3 * n + 0 != pack(b, te::FX, v6lo, k, j, i) && k <= kb.e && j <= jb.e)
```
Yurlungur: could I use 1, 2, 3 instead of `te::FX` here?

Yurlungur (changes requested):
This is cool! A few questions/nitpicks below.

src/bvals/comms/bnd_info.hpp
lroberts36: These are now seven dimensional arrays under the hood since we need another index for potentially denoting the face/edge we are asking for. We could go to `ParArray7D` instead everywhere, I am not invested in either choice.

src/interface/variable.hpp
lroberts36: They can be removed as they are mainly an artifact of getting things to work when I changed the dimension of the arrays. Essentially, `GetTensorComponent(t, u, v)` should give you the subview corresponding to either v(0, t, u, v) if the underlying par array is 7D or v(t, u, v) if the underlying par array is 6D. I think it is probably more confusing to have them there since we are fixed to seven dimensional par arrays now, so I will remove them.

src/interface/variable_pack.hpp
lroberts36: Passing in an integer instead of the enum would result in a different overload of `operator()` being called, so we are stuck with using the enum (if we want variable packs to be able to be called with a fake block idx).

tst/unit/test_mesh_data.cpp
lroberts36: I think you would never want to do this in practice, just for testing. This is just to fill a variable with values without using a pack.

tst/unit/test_mesh_data.cpp
lroberts36: see  above

src/bvals/comms/bnd_info.hpp
Yurlungur: I think `ParArray7D` is more descriptive of intent, but if we ever change this again, it'll be annoying. So I don't feel too strongly I suppose.

src/interface/variable.hpp
Yurlungur: :+1:

src/utils/multi_pointer.hpp
lroberts36: `multi_pointer` is just there to so we can create arbitrary rank views via `Kokkos::View<multi_pointer_t<Real, NDIM>, ...>`. I don't think it will have any other use. I will add a comment to this effect.

src/interface/variable_pack.hpp
Yurlungur: Oh I see. Yeah ok. Is that going to cause issues with loops over faces, for example? Maybe we should try to expose some construct for that... like face and edge accessors for the pack that work for the appropriate type so simpler loops can be called?

tst/unit/test_mesh_data.cpp
Yurlungur: Yeah I suppose we always interact with things through packs, so maybe it doesn't matter. I guess I just wonder if instead of setting the slowest-possible index to be the topological location, we could set it to the slowest one in practice? Maybe that's not what we want?

src/interface/variable_pack.hpp
lroberts36: Maybe, but I think I would leave it until there is a clear use case. You can always do things like `for (auto el : {TE::FX, TE::FY, TE::FZ}){...}`, but the spatial index range is different for each of those components.

Yurlungur: Ah that's true. Automatic ranges maybe make the integer looping irrelevant.

tst/unit/test_mesh_data.cpp
lroberts36: I think that necessitates extra logic in a number of places in the code (e.g. boundaries), so I would avoid changing the dimension of the topological index unless absolutely necessary.

Yurlungur: yeah ok

src/interface/sparse_pack.hpp
lroberts36: Oh, good catch. I added this interface before I put in all of the `TopologicalElement` stuff. This should probably be 
```c++
  // operator() overloads
  KOKKOS_INLINE_FUNCTION
  auto &operator()(const int b, const int idx) const { return pack_(0, b, idx); }
  
  // operator() overloads
  KOKKOS_INLINE_FUNCTION
  auto &operator()(const int b, TopologicalElement te, const int idx) const { return pack_((int) te % 3, b, idx); }
  
```  
and similarly below. I think there is no need for the templated call really.

src/interface/variable.hpp
lroberts36: done

src/basic_types.hpp
lroberts36: done

src/interface/variable_pack.hpp
lroberts36: `GetTensorComponent` is now gone

src/interface/variable_state.hpp
lroberts36: done.

src/utils/multi_pointer.hpp
lroberts36: done

Yurlungur (approved):

src/interface/sparse_pack.hpp
lroberts36: done

src/bvals/comms/bnd_info.hpp
lroberts36: I'll leave this open for now in case anyone else has an opinion.

src/basic_types.hpp
```diff
@@ -50,6 +52,45 @@ enum class BoundaryType : int {
...
+};
```
forrestglines, line 76: If the need to iterate over all faces or edges does arise, it might be nice to have shorthand arrays
```
const TopologicalElement ALL_FACES ={ TE:FX, TE:FY, TE:FZ };
const TopologicalElement ALL_EDGES ={ TE:EYZ, TE:EXZ, TE:EXY };
```

src/interface/variable_pack.hpp
```diff
@@ -276,7 +277,7 @@ class VariablePack {
     assert(dims_[0] > 1);
     assert(dims_[1] > 0);
     assert(dims_[2] > 0);
-    assert(dims_[3] == v_.extent(0));
+    assert(dims_[3] == v_[0].extent(0));
```
forrestglines: Are any assertions needed for the additional dimensions for faces and edges?

src/interface/variable_pack.hpp
```diff
@@ -368,7 +394,7 @@ class VariablePack {
   ParArray0D<Coordinates_t> coords;
 
  protected:
-  ViewOfParArrays<T> v_;
+  std::array<ViewOfParArrays<T>, 3> v_;
```
forrestglines: A bit unfortunate that we need 3 views of ParArrays, hopefully it's not an impactful amount of addition allocations/data movements.

Are the three separate views due to each having different dimensionality?

forrestglines:
@lroberts36 Just to let you know, I'm getting through a review of this PR, best bet I'll be able to finish it tomorrow

src/interface/variable_pack.hpp
```diff
@@ -307,15 +307,33 @@ class VariablePack {
...
+    assert(IsAllocated(n));
+    assert(v_((static_cast<std::size_t>(el) % 3) * dims_[3] + n).topological_type ==
+           GetTopologicalType(el));
+    return v_((static_cast<std::size_t>(el) % 3) * dims_[3] + n);
+  }
+
   KOKKOS_FORCEINLINE_FUNCTION
   ParArray3D<T, VariableState> &operator()(const int n) const {
     assert(IsAllocated(n));
+    assert(v_(n).topological_type == TopologicalType::Cell);
```
forrestglines, line 321: Will this function also work for `TopologicalType::Node`?

src/interface/variable_pack.hpp
```diff
@@ -529,13 +554,14 @@ void AppendSparseBaseMap(const VariableVector<T> &vars, PackIndexMap *pvmap) {
 }
 
 template <typename T>
-void FillVarView(const VariableVector<T> &vars, bool coarse, ViewOfParArrays<T> &cv_out,
-                 ParArray1D<int> &sparse_id_out, ParArray1D<int> &vector_component_out,
-                 ParArray1D<bool> &allocated_out, PackIndexMap *pvmap) {
+void FillVarView(const VariableVector<T> &vars, int vsize, bool coarse,
+                 ViewOfParArrays<T> &cv_out, ParArray1D<int> &sparse_id_out,
+                 ParArray1D<int> &vector_component_out, ParArray1D<bool> &allocated_out,
+                 PackIndexMap *pvmap) {
```
forrestglines, line 560: Conceptually, why is `vsize` now added to FillVarView? It's the vector size of the view, right? The sum of the vector size of all vars in the view?

forrestglines (approved):
Looks good so far. I have a few questions relating to feature directions:

1. Right now the different faces, edges, are packaged into one array, which means *some* wasted space. Does the outward facing interface have room to separate out those arrays in case we want to do internal code surgery to save that space?

2. I can see a few useful operations that might go into Parthenon to help with topological elements -- interpolating from faces to cells and vice versa, Curls on faces to cell centers, etc. I think we should get those into Parthenon rather than downstream codes.

src/basic_types.hpp
lroberts36: You are definitely right. I already had to do something like this in the next PR but it would be better to have something global as you suggest. I will make this change.

src/interface/variable_pack.hpp
lroberts36: This got refactored to be a single ParArray since it was triggering HtoD copies. I initially did it this way to mirror what is done for fluxes.

src/interface/variable_pack.hpp
lroberts36: It will, but I am not sure we should remove the assertion. My thinking was that we want people to be explicit about which type of field they are getting from a pack so that there is less opportunity for indexing errors. I am open to changing my mind though.

src/interface/variable_pack.hpp
lroberts36: `vsize` is the number of variables. Since we are using a single `ViewOfParArrays` to store all topological components of a field, `cv_out` could have a size that is three times the number of variables. As a result, the old asserts could trigger even when everything is correct.

src/interface/sparse_pack.hpp
```diff
@@ -255,14 +255,27 @@ class SparsePack : public SparsePackBase {
...
 
   // operator() overloads
-  KOKKOS_INLINE_FUNCTION
-  auto &operator()(const int b, const int idx) const { return pack_(0, b, idx); }
+  using TE = TopologicalElement;
+  KOKKOS_INLINE_FUNCTION auto &operator()(const int b, const TE el, const int idx) const {
+    return pack_(static_cast<int>(el) % 3, b, idx);
+  }
+  KOKKOS_INLINE_FUNCTION auto &operator()(const int b, const int idx) const {
+    return pack_(0, b, idx);
```
brryan, line 266: Can there be a `PARTHENON_DEBUG_REQUIRE(pack_.GetDim(0) == 1, "Suppressed element index for multi-valued topological element");` here? I can imagine the wrong choice of `&operator()` here being an easy mistake that's hard to debug. (I'm not 100% sure that `GetDim(0)` is the right index)

lroberts36: You are right, we need some kind of check here. I think it probably should be something like 
```c++
PARTHENON_DEBUG_REQUIRE(pack_(0, b, idx).topological_type == TopologicalType::Cell, "Suppressed topological element index assumes that this is a cell variable, but it isn't");
```

Yurlungur:
WIth one approval from LANL and one from AthenaPK, are we good to merge this?

lroberts36:
> WIth one approval from LANL and one from AthenaPK, are we good to merge this?

I think so, but let's wait til the next Parthenon meeting.

lroberts36:
@forrestglines:
>     1. Right now the different faces, edges, are packaged into one array, which means _some_ wasted space. Does the outward facing interface have room to separate out those arrays in case we want to do internal code surgery to save that space?

I think it shouldn't be too hard to switch to having no wasted space, but at the expense of three times as many allocations. 

> 
>     2. I can see a few useful operations that might go into Parthenon to help with topological elements -- interpolating from faces to cells and vice versa, Curls on faces to cell centers, etc. I think we should get those into Parthenon rather than downstream codes.

I think things like curls, etc. should maybe go into `update.hpp` but we can save that for a future PR. I have to think a little more about interpolation from one kind of topological element to another, but I would also say let's save that for a future PR.

src/interface/variable_pack.hpp
lroberts36: I think this is fixed now.

src/basic_types.hpp
lroberts36: Actually going to leave this for the next PR.

src/basic_types.hpp
```diff
@@ -50,6 +52,45 @@ enum class BoundaryType : int {
...
+// NXYZ = edge at (i - 1/2, j - 1/2, k - 1/2)
+//
+// The values of the enumeration are chosen so we can do te % 3 to get
+// the correct index for each type of element in Variable::data
+enum class TopologicalElement : std::size_t {
+  C = 0,
+  FX = 3,
+  FY = 4,
+  FZ = 5,
+  EYZ = 6,
```
c-prather: This naming is consistent with the literature and perfectly clear -- however, I found that when using the names often, for my sanity I needed to define:
```
constexpr TE F1 = TE::FX;
constexpr TE F2 = TE::FY;
constexpr TE F3 = TE::FZ;
constexpr TE E1 = TE::EYZ;
constexpr TE E2 = TE::EXZ;
constexpr TE E3 = TE::EXY;
constexpr TE NN = TE::NXYZ;
```
I don't know if these would also be easier for other people, but I found it much nicer to look through e.g. emf calculations with numbers and consistent character widths, vs. remembering that `EXZ` is the one pointing in the Y direction, i.e. direction 2, so needs to be differenced in `j`, etc etc.

I wonder if you're attached to these names, or if we could support both schemes in user code?

c-prather (approved):
As I said I've been using this for a while without problems -- or at least, without problems caused by this code.

Debug builds were my friend when moving to face-centered variables, even for non-segfault issues.  I like them and I think we should push them at users more aggressively as the solution to a broad range of bugs, not just segfaults and bounds errors.

src/basic_types.hpp
lroberts36: I am not attached to those names. I don't think we should support two naming schemes in Parthenon itself. That seems likely to be confusing for new users. Maybe `F1` instead of `FX` is more consistent with other things in the code, but I guess I prefer `X`, `Y`, and `Z` personally. I do like that your suggestions have the same character width.

Yurlungur: I don't have a strong opinion, but I did find the `X`, `Y`, `Z` notation a little hard to context switch to, given the rest of the code uses `1`, `2`, and `3`.

lroberts36: Yeah, it probably makes sense to switch. X, Y, and Z are a little to tied to Cartesian coordinate systems, but I guess I just dislike the current convention since it naturally starts array indexing at 1.

Yurlungur: Just imagine 0 is reserved for time. :D

pgrete: How about aligning those with the conventions used for the coordinates https://parthenon-hpc-lab.github.io/parthenon/develop/src/coordinates.html
We discussed various options back then and it seems there was some agreement on the current ones.

Yurlungur: Fine by me.

lroberts36: @pgrete: I think that makes sense. I guess to be consistent with that notation we would use what @bprather suggested (and also `CC` for cell-centered)?

lroberts36: I am going to switch to 
```c++
enum class TopologicalElement : int { 
  CC = 0; 
  F1 = 3; 
  F2 = 4;
  F3 = 5;
  E1 = 6;
  E2 = 7;
  E3 = 8; 
  NN = 9; 
};
```
unless anyone has any other suggestions.

lroberts36: done

src/bvals/comms/bnd_info.hpp
lroberts36: Ok, no other opinions so I am going to leave as is.

PR #870, opened 2023-05-02
Update component label logic in output to include basename

### LINKED ISSUE #864: Component labels in output
Our current machinery to generate component labels (used e.g., in the hdf5/xdmf output and in Ascent) ignores the (base) label when component labels are provided:
https://github.com/parthenon-hpc-lab/parthenon/blob/01cd49ec65506421a1a0b86dd9ae122de68a663d/src/outputs/outputs.cpp#L112

This results in issues (both for Ascent and ParaView) if two vectors use the same labels, say a "density" for a primitive and conserved variable vector (or magnetic fields) as only one component will show up in the analysis tools (all data is written correctly, it's just the use of the component labels).

I see two obvious options:
1. modify the outputs to that they take the base label into account
2. modify the component label creation function to always take the base label into account

I'm in favor of the second as we'd only need to maintain one piece of code and it'd be in line with the case where no labels are provided, i.e., where the base label is used with `_%d` suffix.

Given that this might affect downstream (analysis) codes input from every group would be welcome, ping @Yurlungur @lroberts36 @bprather 
Finally, it'd be great if we could make a decision on a reasonably short timeframe as this will directly the yt frontend.

pgrete:
## PR Summary

This PR makes the component labels generated by `VarInfo` unique.
In the process I noticed that we cannot let component labels take precedence (and ignore the basename, e.g, for scalar or single component vectors/tensors) as this may result in a name clash between different fields.
Thus, the basename always needs to be included.

Fixes #864

I also fixed made sure that only cell variable are published.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/src/outputs/ascent.cpp
+++ b/src/outputs/ascent.cpp
@@ -29,6 +29,7 @@
 #include "coordinates/coordinates.hpp"
 #include "defs.hpp"
 #include "globals.hpp"
+#include "interface/variable_state.hpp"
 #include "mesh/mesh.hpp"
 #include "outputs/output_utils.hpp"
 #include "outputs/outputs.hpp"
@@ -169,6 +170,11 @@ void AscentOutput::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm,
     auto &mbd = pmb->meshblock_data.Get();
 
     for (const auto &var : mbd->GetCellVariableVector()) {
+      // ensure that only cell vars are added (for now) as the topology above is only
+      // valid for cell centered vars
+      if (!var->IsSet(Metadata::Cell)) {
+        continue;
+      }
       const auto var_info = VarInfo(var);
 
       for (int icomp = 0; icomp < var_info.num_components; ++icomp) {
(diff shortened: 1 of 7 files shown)


src/outputs/ascent.cpp
```diff
@@ -169,6 +170,11 @@ void AscentOutput::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm,
     auto &mbd = pmb->meshblock_data.Get();
 
     for (const auto &var : mbd->GetCellVariableVector()) {
+      // ensure that only cell vars are added (for now) as the topology above is only
+      // valid for cell centered vars
+      if (!var->IsSet(Metadata::Cell)) {
+        continue;
+      }
```
lroberts36, line 177: This is the right fix even for the updates I am making to include face variables, etc. `GetCellVariableVector` is now just `GetVariableVector`.

lroberts36 (approved):
LGTM

Yurlungur (approved):
LGTM

pgrete:
Just updated the gold standard as `Advected_0_0` in the `Info` metadata is now `advected_0_0`.
Enabling auto-merge now.

PR #872, opened 2023-05-03
Boundary communication for non-cell centered fields

lroberts36:
## PR Summary
This PR updates the boundary communication routines to also work with face, edge, and nodal fields. All types of variables are treated on equal footing, different types of variables just have different sets of index ranges that are copied from one block to another during boundary communication. 

The largest change involves refactoring the routines for calculating index ranges for boundary communication and generalizes them to work for faces, edges, and nodes (which have different ranges). Additionally, the necessary changes have been made so that buffer packing also runs over faces/edges in addition to other indices when required. Physical boundary conditions have also been updated to work on any type of variable. 

Additionally, it moves all boundary prolongation and restriction to in-one and unifies these with the boundary communication routines. It also adds the machinery necessary for doing "internal" prolongation (which is not require for cell centered fields). This all jumps the gun a little bit, since I have not tried face fields alongside AMR, but the changes at least still work for cell centered variables. I will work out the kinks of prolongation/restriction for non-face fields in a future PR. These changes allow the removal of some old boundary prolongation machinery (`ProlongateBoundaries`, things related to `BoundaryType::Restricted`). The new machinery is currently less careful about which ghost regions are restricted than the old machinery. This doesn't impact correctness, but could have a (likely small) performance implication. In principle, there is nothing stopping the new code from restricting only where it is absolutely required. I just haven't worked through the logic. 

Given that this has resulted in significant refactoring of the code, I think it is better to put off adding face field tests to another PR so this one doesn't get to large. Therefore, it should work for all of the old cell variable tests but there will be no tests that guarantee that the face, etc. fields are actually working. That being said, I have tested this in a downstream CT MHD code (w/o refinement) and the boundary communication, packing, index ranges, boundary conditions, etc. seem to be working correctly. Prolongation and restriction haven't been tested yet on non-cell centered fields, but that is probably better left to another PR.

TODO:
- [x] Update buffer size calculations to deal with new variable types
- [x] Update index range calculations to deal with new variable types 
- [x] Move to prolongation and restriction in one on boundaries after setting fields.
- [x] Update restriction to be generic
- [x] Update shared prolongation to be generic
- [x] Change registration of prolongation/restriction to work for faces and edges (since they have multiple components)
- [x] Add internal prolongation
- [x] Reason through indexing offsets for internal prolongation
- [x] Update boundary conditions to deal with non-cell centered fields. 
- [ ] Figure out why prolongation in one doesn't work nicely when boundaries are split into local and nonlocal caches. 
- [x] Update sparse test gold file (Order in which prolongation and deallocation are performed has changed, which introduces changes in the results of sparse advection below the sparse threshold)

Magnetic rotor test magnetic energy and divB for some visual evidence things are more or less working:

 
Screen Shot 2023-06-01 at 6 25 14 PM
Screen Shot 2023-06-01 at 6 25 44 PM

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [ ] Docs build


src/bvals/comms/bnd_info.cpp
```diff
@@ -36,202 +37,100 @@
...
       }
+      ++off_idx; // Offsets are listed in X1,X2,X3 order, should never try to access
+                 // with off_idx > 1 since all neighbors must have a non-zero block
+                 // offset in some direction
+    } else if (block_offset[dir] > 0) {
+      s[dir] = bounds[dir].e - Globals::nghost + 1 - top_offset[dir];
+      e[dir] = bounds[dir].e - top_offset[dir];
+    } else {
+      s[dir] = bounds[dir].s + top_offset[dir];
+      e[dir] = bounds[dir].s + Globals::nghost - 1 + top_offset[dir];
```
Yurlungur: I haven't fully reasoned out this new machinery... but it seems like a significant cleanup IMO

src/bvals/comms/boundary_communication.cpp
```diff
@@ -89,47 +89,23 @@ TaskStatus SendBoundBufs(std::shared_ptr<MeshData<Real>> &md) {
...
-
-        Kokkos::single(Kokkos::PerTeam(team_member),
-                       [&]() { sending_nonzero_flags(b) = non_zero; });
+        bool non_zero[3]{false, false, false};
+        for (int iel = 0; iel < bnd_info(b).ntopological_elements; ++iel) {
+          auto &idxer = bnd_info(b).idxer[iel];
+          Kokkos::parallel_reduce(
+              Kokkos::TeamThreadRange<>(team_member, idxer.size()),
+              [&](const int idx, bool &lnon_zero) {
+                const auto [t, u, v, k, j, i] = idxer(idx);
```
Yurlungur, line 101: nice

src/utils/indexer.hpp
```diff
@@ -0,0 +1,102 @@
...
+namespace parthenon {
+
+template <class... Ts>
+struct Indexer {
+  Indexer() : N{}, start{}, _size{} {};
+
+  explicit Indexer(std::pair<Ts, Ts>... Ns)
+      : N{GetFactors(std::make_tuple((Ns.second - Ns.first + 1)...),
+                     std::make_index_sequence<sizeof...(Ts)>())},
+        start{Ns.first...}, end{Ns.second...}, _size(((Ns.second - Ns.first + 1) * ...)) {
```
Yurlungur, line 32: did not know you could variatically unpack with `*`. That's cool.

src/utils/indexer.hpp
```diff
@@ -0,0 +1,102 @@
...
+    std::tuple<Ts...> idxs;
+    (
+        [&] {
+          std::get<Is>(idxs) = idx / std::get<Is>(N);
+          idx -= std::get<Is>(idxs) * std::get<Is>(N);
+          std::get<Is>(idxs) += std::get<Is>(start);
+        }(),
+        ...);
+    return idxs;
+  }
```
Yurlungur, line 68: I like it

src/bvals/comms/bnd_info.cpp
```diff
@@ -420,6 +336,15 @@ BndInfo BndInfo::GetSetBndInfo(std::shared_ptr<MeshBlock> pmb, const NeighborBlo
...
     PARTHENON_FAIL("Buffer should be in a received state.");
   }
+
+  // Still don't understand why, but these have to be set
+  out.si = out.idxer[0].template StartIdx<5>();
+  out.ei = out.idxer[0].template EndIdx<5>();
+  out.sj = out.idxer[0].template StartIdx<4>();
+  out.ej = out.idxer[0].template EndIdx<4>();
+  out.sk = out.idxer[0].template StartIdx<3>();
+  out.ek = out.idxer[0].template EndIdx<3>();
```
Yurlungur: The prolongation/restriction machinery uses them:
```C++
GetLoopBoundsFromBndInfo(const Info_t &info, const int ckbs, const int cjbs, int &sk,
                         int &ek, int &sj, int &ej, int &si, int &ei) {
  sk = info.sk;
  ek = info.ek;
  sj = info.sj;
  ej = info.ej;
  si = info.si;
  ei = info.ei;
  if (DIM < 3) sk = ek = ckbs; // TODO(C++17) make constexpr
  if (DIM < 2) sj = ej = cjbs;
}
```
The simplest way to match the prolong/restrict machinery to your new machinery is to modify that function to use the indexer.

Yurlungur (commented):
I think there's nothing philosophically here that should interfere with the polongation/restriction machinery. 

That said, there probably needs to be a matching step where we figure out the appropriate way to pass bounds to prolongation/restriction on coarse/fine buffers. See below:

src/bvals/comms/bnd_info.cpp
lroberts36: I am pretty sure it is giving the same index ranges as the old code (I ran the test suite calling both and made sure the bounds were always equal), but if you get the chance it might be useful for someone else to go through it and make sure they don't see some edge case I missed.

src/utils/indexer.hpp
lroberts36: Yeah, fold expressions are available in c++17. I think you could do everything they do with template recursion, but fold expressions give much cleaner code.

src/utils/indexer.hpp
lroberts36: yeah, I was pretty proud of this code. It should produce the same assembly if you had written out the indexing by hand. Sadly, it may not be usable because the tuple copy constructor doesn't seem to work on device. Maybe the compiler option `--expt-relaxed-constexpr` would fix this for CUDA but it also seems to not work for HIP.

src/bvals/comms/bnd_info.cpp
lroberts36: Does the prolongation/restriction machinery use `BndInfo` objects that were constructed using `GetSendBndInfo` or `GetSetBndInfo`?

src/utils/indexer.hpp
lroberts36: It looks like the solution to the shortcomings of `std::tuple` on device can be circumvented by just switching to `std::array` so we can use the generic implementation. I didn't check that it produces the same assembly, but I doubt there is any real performance hit.

pgrete: FWIW there's also a `Kokkos::Array`

src/bvals/comms/bnd_info.cpp
lroberts36: I removed `BndInfo::si`, etc. and everything seems to be working. Still a little confused about why it mattered what I did in `GetSetBndInfo`, but probably not worth worrying about.

src/bvals/comms/bnd_info.cpp
Yurlungur: yeah I'll try to do that. I've been out sick the last while.

src/bvals/comms/bnd_info.cpp
Yurlungur: > Does the prolongation/restriction machinery use BndInfo objects that were constructed using GetSendBndInfo or GetSetBndInfo?

Ah good point. No it shouldn't.

> I removed BndInfo::si, etc. and everything seems to be working. Still a little confused about why it mattered what I did in GetSetBndInfo, but probably not worth worrying about.

Now it uses your new machinery? Nice.

benchmarks/burgers/burgers_driver.cpp
```diff
@@ -120,9 +120,6 @@ TaskCollection BurgersDriver::MakeTaskCollection(BlockList_t &blocks, const int
 
     auto fill_deriv = tl.AddTask(update, FillDerived<MeshData<Real>>, mc1.get());
 
-    if (pmesh->multilevel) {
-      tl.AddTask(set | set_local, parthenon::RestrictGhostHalos, mc1, false);
-    }
```
Yurlungur, line 125: I know we talked this out together. But we should double check that this change doesn't break anything. The advection outflow test (I THINK) should catch issues here.

src/basic_types.hpp
```diff
@@ -91,6 +84,39 @@ TopologicalType GetTopologicalType(TopologicalElement el) {
...
+  return (el == TE::FX || el == TE::EXY || el == TE::EYZ || el == TE::NXYZ);
+}
+KOKKOS_INLINE_FUNCTION int TopologicalOffsetJ(TE el) noexcept {
+  return (el == TE::FY || el == TE::EXY || el == TE::EYZ || el == TE::NXYZ);
+}
+KOKKOS_INLINE_FUNCTION int TopologicalOffsetK(TE el) noexcept {
+  return (el == TE::FZ || el == TE::EXZ || el == TE::EYZ || el == TE::NXYZ);
+}
+
+inline constexpr bool IsSubmanifold(TopologicalElement container,
```
Yurlungur: This is pretty hard to parse. I don't know if there's a good solution to that though.

src/bvals/comms/bnd_info.cpp
```diff
@@ -28,308 +28,149 @@
...
+          bounds[dir].e > bounds[dir].s) { // Check that this dimension has ghost zones
+        // We are sending from a coarser level to the coarse buffer of a finer level,
+        // so we need to only send the approximately half of the indices that overlap
+        // with the other block. We also send nghost "extra" zones in the interior
+        // to ensure there is enough information for prolongation. Also note for
+        // non-cell centered values the number of grid points may be odd, so we
+        // pick up an extra zone that is communicated. I think this is ok, but
+        // something to keep in mind if there are issues.
+        const int half_grid = (bounds[dir].e - bounds[dir].s + 1) / 2;
+        s[dir] += face_offset[off_idx] == 1 ? half_grid - Globals::nghost : 0;
```
Yurlungur, line 77: I can't parse all this easily. So I'm mostly trusting that if the tests say it works, it's correct.

src/bvals/comms/boundary_communication.cpp
```diff
@@ -242,119 +219,124 @@ TaskStatus SetBounds(std::shared_ptr<MeshData<Real>> &md) {
...
+
+    // Apply coarse boundary conditions
+    for (int block = 0; block < md->NumBlocks(); ++block) {
+      ApplyBoundaryConditionsOnCoarseOrFine(md->GetBlockData(block), true);
+    }
+
+    // Prolongate from coarse buffer
+    refinement::Prolongate(resolved_packages, cache, pmb->cellbounds, pmb->c_cellbounds);
+    refinement::ProlongateInternal(resolved_packages, cache, pmb->cellbounds,
+                                   pmb->c_cellbounds);
```
Yurlungur, line 264: :+1:

src/bvals/comms/boundary_communication.cpp
```diff
@@ -242,119 +219,124 @@ TaskStatus SetBounds(std::shared_ptr<MeshData<Real>> &md) {
...
+  auto [rebuild, nbound] = CheckReceiveBufferCacheForRebuild<bound_type, false>(md);
+  if (rebuild) RebuildBufferCache<bound_type, false>(md, nbound, BndInfo::GetSetBndInfo);
+  if (nbound > 0 && pmesh->multilevel) {
+    auto pmb = md->GetBlockData(0)->GetBlockPointer();
+    StateDescriptor *resolved_packages = pmb->resolved_packages.get();
+
+    // Prolongate from coarse buffer
+    refinement::Prolongate(resolved_packages, cache, pmb->cellbounds, pmb->c_cellbounds);
+    refinement::ProlongateInternal(resolved_packages, cache, pmb->cellbounds,
+                                   pmb->c_cellbounds);
```
Yurlungur, line 290: Should these be separate calls? Or should `ProlongateInternal` be called at the end of `Prolongate? Maybe it doesn't matter much...

src/bvals/comms/boundary_communication.cpp
```diff
@@ -242,119 +219,124 @@ TaskStatus SetBounds(std::shared_ptr<MeshData<Real>> &md) {
...
+  // auto pro_local = tl.AddTask(cbound | set_local | set, ProlongateBounds<local>, md);
+  // auto pro = tl.AddTask(cbound | set_local | set, ProlongateBounds<nonlocal>, md);
+
+  // auto out = (pro_local | pro);
 
+  // TODO(LFR): Splitting up the boundary tasks while doing prolongation in one
+  //            breaks things, which I haven't completely figured out yet. To
+  //            move to the commented out code above, this needs to be fixed and
+  //            the prolongation and physical boundary conditions need to be removed
+  //            from the end of SetBounds
```
Yurlungur, line 340: weird.

src/mesh/mesh.cpp
```diff
@@ -1110,18 +1110,11 @@ void Mesh::Initialize(bool init_problem, ParameterInput *pin, ApplicationInput *
...
-        RestrictGhostHalos(md, true);
-      }
     }
 
     //  Now do prolongation, compute primitives, apply BCs
     for (int i = 0; i < nmb; ++i) {
       auto &mbd = block_list[i]->meshblock_data.Get();
-      if (multilevel) { // TODO(JMM): Do with meshdata
-        ProlongateBoundaries(mbd);
-      }
```
Yurlungur, line 1126: I'm not sure this one can go way. Unless you've moved it?

src/prolong_restrict/pr_loops.hpp
```diff
@@ -72,6 +58,54 @@ GetLoopBoundsFromBndInfo(const Info_t &info, const int ckbs, const int cjbs, int
...
 // and a version that automatically swaps between them depending on
 // the size of the buffer cache.
+
+template <int DIM, class Stencil, TopologicalElement FEL, TopologicalElement CEL>
+KOKKOS_INLINE_FUNCTION void InnerProlongationRestrictionLoop(
+    team_mbr_t &team_member, std::size_t buf, const BufferCache_t &info,
+    const IndexRange &ckb, const IndexRange &cjb, const IndexRange &cib,
+    const IndexRange &kb, const IndexRange &jb, const IndexRange &ib) {
+  const auto &idxer = info(buf).prores_idxer[static_cast<int>(CEL)];
+  par_for_inner(inner_loop_pattern_ttr_tag, team_member, 0, 0, 0, 0, 0, idxer.size() - 1,
```
Yurlungur: what's going on here with these loop bounds... why not just do a 1d inner loop?

src/prolong_restrict/pr_ops.hpp
```diff
@@ -117,26 +102,36 @@ Real GradMinMod(const Real fc, const Real fm, const Real fp, const Real dxm,
 
 } // namespace util
 
-struct RestrictCellAverage {
-  template <int DIM>
+struct Restrict {
```
Yurlungur: I don't think this should just be named "restrict," as it's definitely a second-order restriction operation, when higher-order restriction operations are possible.

src/prolong_restrict/pr_ops.hpp
```diff
@@ -117,26 +102,36 @@ Real GradMinMod(const Real fc, const Real fm, const Real fp, const Real dxm,
...
+  static constexpr bool OperationRequired(TopologicalElement fel,
+                                          TopologicalElement cel) {
+    return fel == cel;
+  }
```
Yurlungur, line 109: It's a little gross to require specifying this for downstream users.... If we have some sense for what this should be, could we stash it in a base class?

src/prolong_restrict/pr_loops.hpp
```diff
@@ -92,20 +126,76 @@ ProlongationRestrictionLoop(const BufferCache_t &info, const Idx_t &buffer_idxs,
...
+  par_for(
+      DEFAULT_LOOP_PATTERN, "ProlongateOrRestrictCellCenteredValues", DevExecSpace(), 0,
+      0, 0, 0, 0, idxer.size() - 1, KOKKOS_LAMBDA(const int, const int, const int ii) {
+        const auto [t, u, v, k, j, i] = idxer(ii);
+        Stencil::template Do<DIM, FEL, CEL>(t, u, v, k, j, i, ckb, cjb, cib, kb, jb, ib,
+                                            coords, coarse_coords, &coarse, &fine);
+      });
+}
+
+template <int DIM, class Stencil, TopologicalElement... ELs, class... Args>
```
Yurlungur, line 166: Why is `TopologicalElement` variatic here? It doesn't seem like you ever use more than one in the calls below? Or am I missing something in how this parameter pack is working?

Yurlungur (commented):
I think your generalization of the prolongationloops infrastructure makes sense. It's nice you can get that all working for all topological elements with such minimal code duplication and branching. I wonder if we could instantiate the template resolutions for the defaults in separate compilation units to minimize compile times?

benchmarks/burgers/burgers_driver.cpp
lroberts36: Agreed, I think it is probably important to test this PR in some downstream codes as well to make sure it doesn't change answers before merging it in.

src/basic_types.hpp
lroberts36: Is it hard to parse in the sense that it is not clear what it is doing? Would more comments help?

src/bvals/comms/bnd_info.cpp
lroberts36: Yeah, this is a bit tricky to think through. I should probably write up some notes so there is some documentation.

src/bvals/comms/boundary_communication.cpp
lroberts36: Yeah, I didn't think to hard about it but you are probably right that there should just be one `Prolongate` function that does both. I can't think of a scenario where you would do one without the other.

src/mesh/mesh.cpp
lroberts36: What `ProlongateBoundaries` used to do is now done in `SetBoundaries` by calling `refinement::Prolongate`. `SetBoundaries` is called just above this line.

src/prolong_restrict/pr_loops.hpp
lroberts36: I couldn't figure out how to get the parthenon loop abstraction to work for a 1D loop, so I punted to keep moving.

src/prolong_restrict/pr_ops.hpp
lroberts36: Isn't it exact if you interpret the data as being cell/face/edge/node averaged?

src/prolong_restrict/pr_ops.hpp
lroberts36: I think so. It will always be `fel == cel` for the restriction and shared prolongation. And I think it will always be the submanifold thing for the prolongation operator.

src/prolong_restrict/pr_loops.hpp
lroberts36: This was just a way of making a compile time loop over topological elements (e.g. the three faces if you are face field). Probably not necessary and confusing, so better to remove it.

src/basic_types.hpp
Yurlungur: A picture might help. I have trouble translating the acronyms to positions on the mesh in my head.

src/prolong_restrict/pr_loops.hpp
Yurlungur: yeah ok. We can probably resolve with a little elbow grease before merge.

src/prolong_restrict/pr_ops.hpp
Yurlungur: Yes, but we may not want to limit ourselves to element-averaged quantities. Element-averaged and element-centered are only identical at second-order.

For nodes simple copying is exact, I think. But we may want non-averaged face/cell-centered quantities that we want to copy. In which case, simple summing/divide by area will not work. This actually comes up in the 4th-order method for riot. We likely want to communicate cell-centered pointwise quantities after con2prim, which means that at mesh boundaries, restriction will be different than this one.

Maybe call this stencil `RestrictAverage` or something? Just to emphasize it works for averaged quantities?

lroberts36:
> I wonder if we could instantiate the template resolutions for the defaults in separate compilation units to minimize compile times?

I messed around with this on Thursday without success, but I certainly didn't exhaust all avenues. It may be instantiating the templates all over the place, especially since they are inlined. I have to brush up on the details of how template instantiation works across compilation units. I think it is probably worth putting the effort in to figure out how to reduce the compilation time because of the code economy it allows. On the other hand, I don't know how performant this approach is yet.

Yurlungur:
> 

I wonder if the issue is they're being re-instantiated when they get called for registration? I.e., maybe the compilation slowdown is all happening in downstream example code? If so, maybe we can instantiate the *functions* that get registered?

lroberts36:
That is possible. Another thing that might help is to move to using the in-one prolongation/restriction everywhere. We are almost certainly doing two times as many instantiations as necessary with the host vs. device operations.

Yurlungur:
> That is possible. Another thing that might help is to move to using the in-one prolongation/restriction everywhere. We are almost certainly doing two times as many instantiations as necessary with the host vs. device operations.

Yes. Definitely.

src/basic_types.hpp
lroberts36: Ok, added some ascii art and some more comments.

src/prolong_restrict/pr_ops.hpp
lroberts36: Ok, that sounds good

lroberts36: done

c-prather (approved):
https://github.com/parthenon-hpc-lab/parthenon/assets/3892634/80b3046d-0b88-4bb6-ad84-17153715fe4f

lroberts36:
Is that multi-block @bprather?

c-prather:
Not yet, I got excited.  But the periodic boundaries now work.

EDIT: Here's a version with 4 blocks at double res in each direction.  Balsara & Spicer (lack of) upwinding where the previous was from the Athena paper.

https://github.com/parthenon-hpc-lab/parthenon/assets/3892634/db4b311b-cb79-47d5-9e74-7d034d7a9cac

pdmullen:
I will eventually answer the following for myself by reading the source, but I am curious how same-level "shared" faces are currently being handled in this MR:

At same-level MeshBlock boundaries, are shared faces synchronized? Or are they handled as in Athena++ whereby each MeshBlock uniquely evolves a face with no step that directly sets one shared face to the other?

lroberts36:
> At same-level MeshBlock boundaries, are shared faces synchronized? Or are they handled as in Athena++ whereby each MeshBlock uniquely evolves a face with no step that directly sets one shared face to the other?

They are handled as in Athena++ currently. For same-level faces, I think we could synchronize just by changing some index ranges by one and having an averaging check in the buffer unpacking routine (so just changes to a couple of lines of code). Synchronization in other cases might take a bit more work. In the next PR I will add flux correction, which I think is maybe a more natural place to include an option for synchronization at same levels. As you have mentioned, there is a little bit of subtlety to avoid race conditions but I think there is a way to make this work within our in-one framework.

Edit: Actually, I guess if you are willing to forgo averaging and just choose one block that "owns" a given shared element and other blocks get this value, this could be achieved by just changing certain index ranges by one (which is a one or two liner).

pdmullen:
> > At same-level MeshBlock boundaries, are shared faces synchronized? Or are they handled as in Athena++ whereby each MeshBlock uniquely evolves a face with no step that directly sets one shared face to the other?
> 
> They are handled as in Athena++ currently. For same-level faces, I think we could synchronize just by changing some index ranges by one and having an averaging check in the buffer unpacking routine (so just changes to a couple of lines of code). Synchronization in other cases might take a bit more work. In the next PR I will add flux correction, which I think is maybe a more natural place to include an option for synchronization at same levels. As you have mentioned, there is a little bit of subtlety to avoid race conditions but I think there is a way to make this work within our in-one framework.
> 
> Edit: Actually, I guess if you are willing to forgo averaging and just choose one block that "owns" a given shared element and other blocks get this value, this could be achieved by just changing certain index ranges by one (which is a one or two liner).

cool beans~

src/basic_types.hpp
```diff
@@ -62,6 +55,31 @@ enum class BoundaryType : int {
...
+//
+//                      E1(i,j+1,k+1)
+//              NN_+---------x---------+_NN(i+1,j+1,k+1)
+//     (i,j+1,k+1)/|  F3(i,j,k+1)     /|
+//               / |      |          / |
+//           E2_x  |      o         x__|_E2(i+1,j,k+1)
+//    (i,j,k+1)/   x         o     /   x___E3(i+1,j+1,k)
+//            /    |         |___ /____|_F2(i,j+1,k)
+//        NN_+---------x---------+_____|_NN(i+1,j,k+1)
+// (i,j,k+1) |  o  |  E1         |  o__|____F1(i+1,j,k)
```
pdmullen, line 71: nice! this must have taken forever to create...

src/basic_types.hpp
```diff
@@ -91,6 +109,43 @@ TopologicalType GetTopologicalType(TopologicalElement el) {
...
+  return (el == TE::F1 || el == TE::E2 || el == TE::E3 || el == TE::NN);
+}
+KOKKOS_INLINE_FUNCTION int TopologicalOffsetJ(TE el) noexcept {
+  return (el == TE::F2 || el == TE::E3 || el == TE::E1 || el == TE::NN);
+}
+KOKKOS_INLINE_FUNCTION int TopologicalOffsetK(TE el) noexcept {
+  return (el == TE::F3 || el == TE::E2 || el == TE::E1 || el == TE::NN);
+}
+
+// Returns wether or not topological element containee is a boundary of
```
pdmullen, line 125: typos in comment

src/bvals/boundary_conditions.cpp
```diff
@@ -134,66 +98,90 @@ void GenericBC(std::shared_ptr<MeshBlockData<Real>> &rc, bool coarse) {
   const int offset = 2 * ref + (INNER ? -1 : 1);
 
   pmb->par_for_bndry(
-      label, nb, domain, coarse,
+      label, nb, domain, el, coarse,
       KOKKOS_LAMBDA(const int &l, const int &k, const int &j, const int &i) {
         if (!q.IsAllocated(l)) return;
         if (TYPE == BCType::Reflect) {
           const bool reflect = (q.VectorComponent(l) == DIR);
```
pdmullen, line 105: Sorry this is the wrong place to ask this question and I should have figured this out in a previous MR:

Does the current Parthenon implementation imply that face-centered quantities at the, for example, X1-face, always correspond to a vector X1 component? Or could one equally define a non-vector variable there, or even a three-vector (for example in the three face components of a stress tensor).

src/bvals/boundary_conditions.cpp
```diff
@@ -134,66 +98,90 @@ void GenericBC(std::shared_ptr<MeshBlockData<Real>> &rc, bool coarse) {
...
+          q(el, l, k, j, i) = q(el, l, X3 ? ref : k, X2 ? ref : j, X1 ? ref : i);
         }
       });
 }
 
 void OutflowInnerX1(std::shared_ptr<MeshBlockData<Real>> &rc, bool coarse) {
-  GenericBC<X1DIR, BCSide::Inner, BCType::Outflow>(rc, coarse);
+  using TE = TopologicalElement;
+  for (auto el : {TE::CC, TE::F1, TE::F2, TE::F3, TE::E1, TE::E2, TE::E3, TE::NN})
+    GenericBC<X1DIR, BCSide::Inner, BCType::Outflow>(rc, coarse, el);
```
pdmullen, line 118: Different BCs for face vs CC variables should be handled via a user-defined BC, right?

src/bvals/comms/bnd_info.cpp
pdmullen: Yeah, I apologize, I don't know how useful I will be in actually checking the indexing here...

pdmullen (commented):
I am not the right person to review this PR; there is lots of indexing hell in here that someone else might better grasp.

src/bvals/boundary_conditions.cpp
c-prather: I'm successfully using 3-component face-centered variables (velocities) for upwinding.  The machinery is all the same regardless of where a variable is, so I think face-centered fields should support the same things as cell-centered ones.  I haven't verified for myself that it works beyond rank 1 though.

src/basic_types.hpp
lroberts36: It wasn't too bad, although I did start from an ASCII cube I found somewhere on the internet.

src/basic_types.hpp
lroberts36: I think I just made up a word here, "containee" as in the one who is contained. The internet suggests this is non-standard though... or maybe you are just referring to the misspelled whether...

pdmullen: whether ;)

c-prather: FWIW you can suggest changes for these sorts of things to make the fix one-click

src/bvals/boundary_conditions.cpp
lroberts36: Aside from pre-existing issues in the mesh refinement code (which I will fix in the next PR), you should be able to define up to rank 3 face centered quantities. I didn't think through how the reflecting boundary conditions should work in that case though.

I was interested to know if there was some advantage to evolving the stress tensor on  a staggered grid.

src/bvals/boundary_conditions.cpp
lroberts36: Yeah, I think so.

lroberts36:
> I am not the right person to review this PR; there is lots of indexing hell in here that someone else might better grasp.

hm, @Yurlungur said something similar. Probably suggests some better documentation is required.

c-prather:
I don't think code reviews are really where index bugs get caught, documentation or no.  I for one trust a working implementation much better than my eyes.  If/as we find anything downstream we can push back index fixes easily enough.

On the subject of documentation, though, we should update the `par_for_bndry` docs, and we should at some point have user-facing documentation of the `TopologicalElement` struct and the new ways of addressing `CellVariable` and the various packs with it.

lroberts36:
@forrestglines @pgrete @BenWibking: It might be good to have one of you test `AthenaPK` against this branch before merging.

src/basic_types.hpp
```diff
@@ -91,6 +109,43 @@ TopologicalType GetTopologicalType(TopologicalElement el) {
...
+// topological element container
+inline constexpr bool IsSubmanifold(TopologicalElement containee,
+                                    TopologicalElement container) {
```
Yurlungur, line 128: :+1:

src/bvals/boundary_conditions.cpp
```diff
@@ -134,66 +98,90 @@ void GenericBC(std::shared_ptr<MeshBlockData<Real>> &rc, bool coarse) {
...
 }
 
 void OutflowOuterX1(std::shared_ptr<MeshBlockData<Real>> &rc, bool coarse) {
-  GenericBC<X1DIR, BCSide::Outer, BCType::Outflow>(rc, coarse);
+  using TE = TopologicalElement;
+  for (auto el : {TE::CC, TE::F1, TE::F2, TE::F3, TE::E1, TE::E2, TE::E3, TE::NN})
+    GenericBC<X1DIR, BCSide::Outer, BCType::Outflow>(rc, coarse, el);
 }
```
Yurlungur, line 125: Some point soon we really need to add meshdata versions of these.

src/coordinates/uniform_cartesian.hpp
```diff
@@ -166,6 +166,40 @@ class UniformCartesian {
...
+    if constexpr (dir == X1DIR && X1EDGE) {
+      return xmin_[dir - 1] + idx * dx_[dir - 1]; // idx - 1/2
+    } else if constexpr (dir == X2DIR && X2EDGE) {
+      return xmin_[dir - 1] + idx * dx_[dir - 1]; // idx - 1/2
+    } else if constexpr (dir == X3DIR && X3EDGE) {
+      return xmin_[dir - 1] + idx * dx_[dir - 1]; // idx - 1/2
+    } else {
+      return xmin_[dir - 1] + (idx + 0.5) * dx_[dir - 1]; // idx
+    }
+    return 0;
```
Yurlungur: is this default return statement necessary with the `else` clause?

src/coordinates/uniform_cartesian.hpp
```diff
@@ -215,6 +249,32 @@ class UniformCartesian {
...
+    } else if constexpr (el == TE::E1) {
+      return dx_[X1DIR - 1];
+    } else if constexpr (el == TE::E2) {
+      return dx_[X2DIR - 1];
+    } else if constexpr (el == TE::E3) {
+      return dx_[X3DIR - 1];
+    } else if constexpr (el == TE::NN) {
+      return 1.0;
+    }
+    return 0.0;
```
Yurlungur, line 277: Maybe add a parthenon fail for this one?

src/interface/metadata.cpp
```diff
@@ -98,6 +98,71 @@ MetadataFlag Metadata::GetUserFlag(const std::string &flagname) {
 }
 
 namespace parthenon {
+Metadata::Metadata(const std::vector<MetadataFlag> &bits, const std::vector<int> &shape,
+                   const std::vector<std::string> &component_labels,
+                   const std::string &associated)
```
Yurlungur, line 103: Was moving the constructor into the implementation file what fixed compile time issues?

src/interface/sparse_pack.hpp
```diff
@@ -443,6 +443,12 @@ class SparsePack : public SparsePackBase {
   // and it may not even be an lvalue.
   static std::vector<MetadataFlag> AddFlag_(const std::vector<MetadataFlag> &flags,
                                             MetadataFlag mf = Metadata::WithFluxes) {
+    // TODO(LFR): We need to decide if variables w/o fluxes should end up a pack that asks
+    //            for WithFluxes. I actually use this reasonably often in practice, even
+    //            though I was the one who suggested enforcing that WithFluxes
+    //            PackDescriptors include the Metadata::WithFluxes flag.
```
Yurlungur: I don't feel strongly either way. If you're using that, just get rid of this functionality.

src/interface/update.cpp
```diff
@@ -60,7 +60,7 @@ template <>
 TaskStatus FluxDivergence(MeshData<Real> *in_obj, MeshData<Real> *dudt_obj) {
   const IndexDomain interior = IndexDomain::interior;
 
-  std::vector<MetadataFlag> flags({Metadata::WithFluxes});
+  std::vector<MetadataFlag> flags({Metadata::WithFluxes, Metadata::Cell});
   const auto &vin = in_obj->PackVariablesAndFluxes(flags);
```
Yurlungur, line 64: should there be an equivalent functionality for face fields? I suppose it will necessarily look different

src/mesh/domain.hpp
```diff
@@ -140,116 +140,131 @@ class IndexShape {
...
 
-  KOKKOS_INLINE_FUNCTION const IndexRange
-  GetBoundsI(const IndexDomain &domain) const noexcept {
-    return (domain == IndexDomain::interior) ? x_[0] : IndexRange{is(domain), ie(domain)};
+  KOKKOS_INLINE_FUNCTION const IndexRange GetBoundsI(const IndexDomain &domain,
+                                                     TE el = TE::CC) const noexcept {
+    return (domain == IndexDomain::interior && el == TE::CC)
+               ? x_[0]
+               : IndexRange{is(domain, el), ie(domain, el)};
   }
```
Yurlungur, line 148: probably important to have this functionality, but FWIW I think it's probably going to be pretty common for people to use the cell centered domains and add the offsets in by hand

src/prolong_restrict/pr_loops.hpp
Yurlungur: might be worth revisiting this before merge

src/prolong_restrict/pr_loops.hpp
```diff
@@ -72,6 +58,54 @@ GetLoopBoundsFromBndInfo(const Info_t &info, const int ckbs, const int cjbs, int
...
+                });
+}
+
+template <int DIM, class Stencil, TopologicalElement... ELs, class... Args>
+KOKKOS_INLINE_FUNCTION void IterateInnerProlongationRestrictionLoop(Args &&...args) {
+  (
+      [&] {
+        if constexpr (Stencil::OperationRequired(ELs, TE::NN))
+          InnerProlongationRestrictionLoop<DIM, Stencil, ELs, TE::NN>(
+              std::forward<Args>(args)...);
```
Yurlungur, line 83: Feels like there should be a constexpr for all if or something... (This is a random thought. Do not go off and reimplement this)

Yurlungur (approved):
Re-reviewed... I think this mostly looks fine. A few comments below.

One big picture question I have: Suppose someone wants to add a new field which is, e.g., node-centered. And they want to add custom prolongation/restriction operators. Does their operator need to implement all the specializations of topological elements? Or is it enough to implement only the ones they care about?

src/bvals/boundary_conditions.cpp
lroberts36: Agreed.

src/coordinates/uniform_cartesian.hpp
lroberts36: Logically it isn't necessary, but I wasn't sure if it would trigger a warning with some compilers so I put it in there.

src/coordinates/uniform_cartesian.hpp
lroberts36: Possibly, but you should never get there because the if clauses exhaust all possible types of topological elements.

src/interface/metadata.cpp
lroberts36: yes, at least to a large extent. The compile time still isn't great on GPU.

src/coordinates/uniform_cartesian.hpp
Yurlungur: Yeah I know, but I kind of want the fail for readability---a clear message "you should never reach this." Same for the above comment.

src/interface/sparse_pack.hpp
lroberts36: That is what I think as well. After merging your changes, forcing `WithFluxes` variables to have `Metadata::WithFluxes` caused a bug in my code (since the sparse packs I was creating had some variables that had fluxes and others that didn't). It just quietly removes variables that you ask for that don't have fluxes, which inevitably causes segfaults.

src/interface/update.cpp
lroberts36: Yeah, the natural version for face fields is to update them based on a finite-volume curl. The index ranges are different as well, so I think it needs to be a different task (or at least a separate kernel within this task).

src/mesh/domain.hpp
lroberts36: One can do that, but having this option certainly simplifies writing more generic code in my experience. In my opinion, I think we should try to enforce explicitly passing the topological type to `GetBounds*` where practical since it results in code that is easier to understand.

src/prolong_restrict/pr_loops.hpp
lroberts36: I think that is doable... ;)

src/prolong_restrict/pr_ops.hpp
lroberts36: I would vote for leaving this as is in this PR and maybe deal with this in a cleanup PR at the end once we have a little more experience using this stuff in downstream codes.

lroberts36:
> Does their operator need to implement all the specializations of topological elements? Or is it enough to implement only the ones they care about?

They can just implement the ones they care about as long as they have `OperationRequired` return false for topological elements they don't support.

src/coordinates/uniform_cartesian.hpp
lroberts36: yeah, it appears removing this causes warnings in CUDA 11.4 at least. Leaving it in.

src/coordinates/uniform_cartesian.hpp
lroberts36: added a warning and a comment

src/interface/update.cpp
lroberts36: I would put off adding update functions for non-cell centered fields to a future PR, since they are just convenience functions and it might be useful to have a little experience using this machinery in downstream codes before including them.

src/interface/sparse_pack.hpp
lroberts36: Ok, just removed it. Sorry for suggesting that in the first place.

src/prolong_restrict/pr_loops.hpp
lroberts36: Is there an example somewhere of the right call signature for a 1D `par_for_inner`? I messed around with it for a little while and never got things to compile.

forrestglines:
>  It might be good to have one of you test `AthenaPK` against this branch before merging.


Sorry for the delay in review, I'm looking through this today.

c-prather:
As a quick backward compatibility note, the CT branch of KHARMA appears to break none* of our regression tests, which all still use cell-centered fields.  Compiling cell-centered code with the new e.g. `par_for_bndry` took a couple tries, but there don't seem to be bugs introduced by non-obvious changes or anything like that.

*at least, none that weren't already broken in its base branch -- mostly these were related to viscosity, electrons, & other non-B-field features.

Between Flux-CT, Dedner CD, Elliptic cleaning, and now B&S and Gardiner & Stone, I think KHARMA now has 5 magnetic field transports...

src/interface/update.cpp
Yurlungur: sounds good :+1:

src/prolong_restrict/pr_loops.hpp
Yurlungur: Actually I'm looking through `kokkos_abstraction.hpp` and I don't see a 1D implementation of `par_for_inner` for `InnerLoopPatternTTR`. I wonder if that's the issue? I.e., the relevant tag dispatch isn't available? I think the following function added to `kokkos_abstraction.cpp` might fix things:

```C++
template <typename Function>
KOKKOS_INLINE_FUNCTION void par_for_inner(InnerLoopPatternTTR, team_mbr_t team_member,
                                          const int il, const int iu,
                                          const Function &function) {
  Kokkos::parallel_for(Kokkos::TeamThreadRange(team_member, il, iu + 1), function);
}
```

forrestglines:
I've verified that this update works with AthenaPK, tests still pass fine. I also had to fix one instance of `par_for_bndry` in the code but that's it.

I'm also reviewing the code but probably won't have much to add. Feel free to merge if I don't get the review done today.

src/prolong_restrict/pr_loops.hpp
lroberts36: done.

PR #873, opened 2023-05-03
fix no particle output

Yurlungur:
## PR Summary

My PR on particle-IO #830 was untested in the case where a swarm has zero particles. In that case, HDF5 complains, as it appears to dislike outputting arrays with zero length. This trivial PR fixes that edge case by outputting the hdf5 scaffolding but skipping output for swarm vars if the count is zero. Similarly, if the total count is zero, the restart reader does not read a given swarm.

I wasn't sure how to test this, as it needs regression testing but the correctness check is "it runs." I think it's a small enough change a test probably isn't necessary.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/src/outputs/parthenon_hdf5.cpp
+++ b/src/outputs/parthenon_hdf5.cpp
@@ -576,6 +576,9 @@ void PHDF5Output::WriteOutputFileImpl(Mesh *pm, ParameterInput *pin, SimTime *tm
 
   AllSwarmInfo swarm_info(pm->block_list, output_params.swarms, restart_);
   for (auto &[swname, swinfo] : swarm_info.all_info) {
+    if (swinfo.global_count == 0) {
+      continue;
+    }
     const H5G g_swm = MakeGroup(file, swname);
     // offsets/counts are NOT the same here vs the grid data
     hsize_t local_offset[6] = {static_cast<hsize_t>(my_offset), 0, 0, 0, 0, 0};
--- a/src/outputs/parthenon_hdf5.cpp
+++ b/src/outputs/parthenon_hdf5.cpp
@@ -576,9 +576,6 @@ void PHDF5Output::WriteOutputFileImpl(Mesh *pm, ParameterInput *pin, SimTime *tm
 
   AllSwarmInfo swarm_info(pm->block_list, output_params.swarms, restart_);
   for (auto &[swname, swinfo] : swarm_info.all_info) {
-    if (swinfo.global_count == 0) {
-      continue;
-    }
     const H5G g_swm = MakeGroup(file, swname);
     // offsets/counts are NOT the same here vs the grid data
     hsize_t local_offset[6] = {static_cast<hsize_t>(my_offset), 0, 0, 0, 0, 0};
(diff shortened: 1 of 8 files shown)


brryan:
This looks good to me -- nice fix. I can imagine the skipped output causing an issue with some IO software somewhere down the line but I think this has to be the solution due to the zero-length array prohibition. 

 Just to confirm, you used this code to restart the `particle_leapfrog` example with particles disabled and that worked? That is good enough for me w.r.t testing for this.

src/parthenon_manager.cpp
```diff
@@ -379,6 +379,15 @@ void ParthenonManager::RestartPackages(Mesh &rm, RestartReader &resfile) {
     std::vector<std::size_t> counts, offsets;
     std::size_t count_on_rank =
         resfile.GetSwarmCounts(swarmname, myBlocks, counts, offsets);
+    // Compute total count and skip this swarm if total count is zero.
+    std::size_t total_count = count_on_rank;
+#ifdef MPI_PARALLEL
+    MPI_Allreduce(&total_count, &total_count, 1, MPI_UNSIGNED_LONG_LONG, MPI_SUM,
```
pgrete: Can we guarantee that `size_t` is always `MPI_UNSIGNED_LONG_LONG`?

src/outputs/parthenon_hdf5.cpp
```diff
@@ -589,6 +589,9 @@ void PHDF5Output::WriteOutputFileImpl(Mesh *pm, ParameterInput *pin, SimTime *tm
                 global_count, pl_xfer);
 
     const H5G g_var = MakeGroup(g_swm, "SwarmVars");
+    if (swinfo.global_count == 0) {
+      continue;
+    }
```
pgrete, line 594: Just double checking that `global_count` is set correctly here, isn't it?
I'm asking as the `SetCounts` function (which also sets a `global_count [array]) follows afterwards.

src/parthenon_manager.cpp
Yurlungur: No we can't, but I don't know another solution, without introducing a lot of preprocessor nonsense... and `size_t=unsigned long long` seems pretty stable, if not standard guaranteed, on modern systems.

src/outputs/parthenon_hdf5.cpp
Yurlungur: Yeah, the naming could use work, my apologies for that. The value in the `SwarmInfo` struct is total particle number and not hdf5 related:
https://github.com/parthenon-hpc-lab/parthenon/blob/develop/src/outputs/output_utils.cpp#L94

src/parthenon_manager.cpp
pgrete: I just noticed that is the the same as in `MPIPrefixSum` you mention below.
How about adding a static_assert kind of thing to (so not introduce all kinds of versions/types, just check that it matches as a cheap better-safe-than-sorry approach)?
I'm happy if you make the final call on this. I'm already approving as is.

pgrete (approved):

src/parthenon_manager.cpp
Yurlungur: Yeah this is a good idea. I can add that.

Yurlungur: Ok this was slightly more complicated than I thought it would be, but I got there:
1. `unsigned long long int` and `unsigned long int` are typically identical on modern systems, but the type system still treats them as different. So instead of checking type identity, I check `sizeof`. It does however assume `size_t` will always be an unsigned integral type. I think this is safe to assume, but I added a static assert for it anyway.
2. Checking for `size_t` and `unsigned long long` caused `cpplint` to yell at me because these classic types are verboten. So I added an exception in the linter for `static_assert`.

brryan (approved):

PR #874, opened 2023-05-04
WIP: Piecewise cubic monotone interpolation

BenWibking:
## PR Summary

Adds monotone interpolation using cubic Hermite polynomials that can be used on device. However, the `MonotoneInterpolator` itself must be constructed on the host, with the contents of the views passed to it accessible from both host and device (if used).

This depends on Kokkos 4 support in order to use `SharedHostPinnedSpace` (https://github.com/kokkos/kokkos/pull/5405).

Reference: "Monotone piecewise cubic interpolation," SIAM J. Numer. Anal., Vol. 17, No. 2, April 1980. https://doi.org/10.1137/0717021

## PR Checklist

- [ ] Code passes cpplint
- [x] Code is formatted
- [ ] Docs build


src/utils/interpolate.hpp
```diff
@@ -0,0 +1,134 @@
+#ifndef INTERPOLATE_HPP_
```
brryan: I can imagine this name `INTERPOLATE_HPP_` easily occurring in a downstream code. Our usual include guard format would be `UTILS_INTERPOLATE_HPP_`, which is maybe still a bit risky but definitely safer. 

Maybe we should expand all our include guards to have a `PARTHENON_` prefix...

src/utils/interpolate.hpp
```diff
@@ -0,0 +1,134 @@
...
+//
+// Performs monotone interpolation using cubic Hermite polynomials.
+// Reference: "Monotone piecewise cubic interpolation," SIAM J. Numer. Anal.,
+//            Vol. 17, No. 2, April 1980. [https://doi.org/10.1137/0717021]
+//========================================================================================
+
+#include "../kokkos_abstraction.hpp"
+#include <ostream>
+
+template <class VectorContainer>
```
brryan: I would put this class in a `namespace parthenon {}` for consistency with the rest of `utils`

brryan (commented):
This is neat! Two quick thoughts now (mostly because I saw the include guard and got to thinking about that more generally)

src/utils/interpolate.hpp
```diff
@@ -0,0 +1,134 @@
...
+
+  KOKKOS_FUNCTION KOKKOS_FORCEINLINE_FUNCTION auto operator()(Real x) const -> Real;
+  KOKKOS_FUNCTION KOKKOS_FORCEINLINE_FUNCTION auto min() const -> Real { return x_min_; }
+  KOKKOS_FUNCTION KOKKOS_FORCEINLINE_FUNCTION auto max() const -> Real { return x_max_; }
+  auto ConstructVectorContainer(size_t size) -> VectorContainer;
+
+ private:
+  Real x_min_{};
+  Real x_max_{}; // x_min and x_max are the min and max of x_vec_
+  VectorContainer x_vec_{};
```
Yurlungur, line 42: Are these assumed to have reference semantics?

src/utils/interpolate.hpp
```diff
@@ -0,0 +1,134 @@
...
+      d_vec_[i + 1] = beta_i_star * delta_i;
+    }
+  }
+}
+
+template <class T>
+KOKKOS_FUNCTION KOKKOS_FORCEINLINE_FUNCTION auto
+MonotoneInterpolator<T>::operator()(Real x) const -> Real {
+  // to avoid branchy code, do a linear search for the segment I_i
+  // where x_{i} <= x < x_{i+1}.
```
Yurlungur, line 106: Linear search seems like a bad idea IMO. Probably better to just branch.

src/utils/interpolate.hpp
```diff
@@ -0,0 +1,134 @@
...
+auto MonotoneInterpolator<std::vector<Real>>::ConstructVectorContainer(size_t size)
+    -> std::vector<Real> {
+  return std::vector<Real>(size);
+}
+
+template <>
+auto MonotoneInterpolator<PinnedArray1D<Real>>::ConstructVectorContainer(size_t size)
+    -> PinnedArray1D<Real> {
+  return PinnedArray1D<Real>("d", size);
+}
```
Yurlungur: I wonder if this can be generalized using @lroberts36 concepts so all view types could be specialized for at once.

Yurlungur (commented):
I think this is useful overall... I assuem a fleshed out PR would have tests. My one big concern is you show using it for both `std::vector` and a kokkos view... those types have different kinds of semantics: vectors are value semantics and views reference semantics. I worry that could get one in to trouble... but maybe it's fine.

src/utils/interpolate.hpp
BenWibking: I can replace this with a binary search.

src/utils/interpolate.hpp
BenWibking: That would definitely be nice to have.

BenWibking:
@Yurlungur Yes, planning to add tests.

src/utils/interpolate.hpp
BenWibking: Currently `PinnedArray1D` is a typedef for `SharedHostPinnedSpace` in Kokkos 4, so this adds a requirement for Kokkos 4 if we have to explicitly specialize for common vector containers/views.

Yurlungur: I *think* this should just work, but Luke should comment
```C++
template<typename VectorContainer, class = ENABLEIF(implements<kokkos_view(VectorContainer)>::value)>
auto MonotoneInterpolator<VectorContainer>::ConstructVectorContainer(size_t size) {
  return VectorContainer("d", size);
}
```
you'll need to include `utils/concepts_lite.hpp`

src/utils/interpolate.hpp
BenWibking: fixed

src/utils/interpolate.hpp
BenWibking: fixed

src/utils/interpolate.hpp
```diff
@@ -0,0 +1,137 @@
...
+// Reference: "Monotone piecewise cubic interpolation," SIAM J. Numer. Anal.,
+//            Vol. 17, No. 2, April 1980. [https://doi.org/10.1137/0717021]
+//========================================================================================
+
+#include "../kokkos_abstraction.hpp"
+#include <ostream>
+
+namespace parthenon {
+
+template <class VectorContainer>
```
Yurlungur: Can also add some concepts magic here for good measure
```suggestion
template<typename VectorContainer, class = ENABLEIF(implements<contiguous_container(VectorContainer)>::value)>
```
limits us to views with contiguous blocks of memory... but that's probably for the best.

src/utils/interpolate.hpp
lroberts36: I think that should work, but it is probably more consistent to use `REQUIRES` instead  
```c++
template<typename VectorContainer, REQUIRES(implements<kokkos_view(VectorContainer)>::value)>
auto MonotoneInterpolator<VectorContainer>::ConstructVectorContainer(size_t size) {
  return VectorContainer("d", size);
}
```

src/utils/interpolate.hpp
BenWibking: I don't think so... Can you provide an example of how this could cause trouble?

src/utils/interpolate.hpp
BenWibking: I can't get this to compile on my machine. I don't fully understand concepts, so I'm not sure why it's complaining:
```
/Users/benwibking/parthenon/src/utils/interpolate.hpp:54:45: error: nested name specifier 'MonotoneInterpolator<VectorContainer>::' for declaration does not refer into a class, class template or class template partial specialization
auto MonotoneInterpolator<VectorContainer>::ConstructVectorContainer(size_t size) {
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

src/utils/interpolate.hpp
Yurlungur: you're constructing `x_vec_` by copy when you call the interpolator, so you're copying around the data unless `VectorContainer` has refernece semantics. That may be a feature or a bug. If `VectoContainer` is always a `Kokkos::View`, then we're fine though.

Yurlungur (changes requested):
Tests look good, but it would be good to also test on device if you plan to use this tool on GPUs.

Yurlungur:
@BenWibking what's the status of this MR?

PR #876, opened 2023-05-11
Fix restarting from files with ghost zones

c-prather:
## PR Summary

Fix an old issue when restarting from files written including ghost zones (i.e., setting ghost_zones=true in the restart file block).

This is clearly not a common usage, but it's nice for Dirichlet boundary conditions -- as well as some old versions of KHARMA, which included ghosts in restart files for esoteric reasons related to recovering primitive variables.

We could regression-test this by enabling ghost_zones=true in one or another restarting test -- I'm not sure where or I'd go ahead.

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@
 - [[PR 663]](https://github.com/lanl/parthenon/pull/663) Change bvals_in_one to use sparse boundary buffers and add flux_correction in one.
 
 ### Fixed (not changing behavior/API/variables/...)
+- [[PR 876]](https://github.com/parthenon-hpc-lab/parthenon/pull/876) Fix restarting from files which record ghost zones
 - [[PR 873]](https://github.com/parthenon-hpc-lab/parthenon/pull/873) Prevent HDF5 from throwing a fit when a swarm has no particles
 - [[PR 866]](https://github.com/parthenon-hpc-lab/parthenon/pull/866) Add missing guard for HDF5 on restart
 - [[PR 861]](https://github.com/parthenon-hpc-lab/parthenon/pull/861) Fix filesystem include for experimental namespace
--- a/cmake/FindFilesystem.cmake
+++ b/cmake/FindFilesystem.cmake
@@ -1,3 +1,17 @@
+
+#=========================================================================================
+# (C) (or copyright) 2023. Triad National Security, LLC. All rights reserved.
+#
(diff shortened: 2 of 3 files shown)


c-prather:
@par-hermes format

Yurlungur (approved):
Assuming tests pass and you add a changelog I'm fine with this change. Might be worth running by the athenak folks too. @pgrete @forrestglines

pgrete:
Does this mean that the `MeshBlockSize` currently stored in the output file with ghost zones is not the actual "block" size, but the data size?

c-prather:
Yes, the block size recorded includes ghosts if they're enabled -- I don't think this is a huge problem, though it is unintuitive (not to mention redundant with the HDF5 dataset dimensions, I think).  I went with the reader fix to avoid changing the file format, but we should document somewhere that that's how it is.

BTW still happy to record ghosts in one of the restart-based tests, just don't know my way around the infrastructure.  The current tests won't really exercise the new code, since no restarts include ghosts.

c-prather:
Failing check is FindFilesystem.cmake.  I think the correct notice is added in Luke's first PR, so as soon as it goes in I can pull & pass the check.

The `perf-and-regression` tests appear to be stalled on all outstanding PRs though?

pgrete (approved):
While I think that the design/the content in the output should be updated, i.e., block sizes are the block sizes without ghost zones and the dimensions are being derived from `IncludesGhost`, this is not that important to be fixed now.
So I'm happy to approve without test (however, feel free to add a node to the doc that writing/restart from ghosts is untested/unsupported/only for debugging).

pgrete:
> The `perf-and-regression` tests appear to be stalled on all outstanding PRs though?

The more expensive test suite is only triggered once "auto-merge" is enabled.
If you want to force it you can either triggered through the Actions tab manually, or briefly enable/disable auto-merge (as I just did).

lroberts36 (approved):
LGTM

Yurlungur:
@bprather can you clean up formatting then we merge?

c-prather:
Oh this is the copyright thing I inherited from develop.  I'll copy the fix from Luke's PR.

Though, separately, I'm not sure I like pasting a big ol' (C) LANL atop someone else's copyright...

PR #877, opened 2023-05-22
Add flat sparse packs

Yurlungur:
## PR Summary

In riot I recently encountered an interesting  use-case, which I needed to support. It's actually familiar. In the boundary communication kernels, we fuse the loop over meshblocks, over faces, and over variables. So the outermost index has a range that spans ALL variables over ALL boundary buffers and ALL meshblocks in a given `MeshData` object. We do this so that when we run with hierarchical parallelism, there's enough "chunks" of work in the outermost index space to saturate the device... i.e., we need at least as many elements in the outer range as we have warps on the GPU. 

I found myself needing the same capability for loops over variables in the interior of a block. I.e., I needed a hierarchical structure where the outer loop is over meshblocks and variables TOGETHER and the inner loop is over the 3D index space of the meshblock. Because of sparse, this outer loop needs to be a SINGLE loop. The number of variables on each meshblock might vary.

Here I implement this capability in @lroberts36 's `SparsePack`s, which was the easiest place. When built with the `sparse` flag in the generic factory function, or with the `GetFlat` factory function, the sparse pack builds with this shape.

Note that I added a blurb describing this functionality, but the documentation for the `SparsePack` is very sparse, and I didn't fill it in.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


Yurlungur:
P.S., thanks @lroberts36 for your help with this

src/interface/sparse_pack_base.cpp
```diff
@@ -85,21 +85,22 @@ SparsePackBase SparsePackBase::Build(T *pmd, const PackDescriptor &desc) {
...
+      size = 0;
+      nblocks++;
+    }
     for (auto &pv : pmbd->GetCellVariableVector()) {
       for (int i = 0; i < nvar; ++i) {
         if (desc.IncludeVariable(i, pv)) {
           if (pv->IsAllocated()) {
             size += pv->GetDim(6) * pv->GetDim(5) * pv->GetDim(4);
-            ndim = (pv->GetDim(1) > 1 ? 1 : 0) + (pv->GetDim(2) > 1 ? 1 : 0) +
-                   (pv->GetDim(3) > 1 ? 1 : 0);
```
Yurlungur, line 102: Removed since, after discussing with @lroberts36, we agreed this logic isn't actually needed.

tst/unit/test_sparse_pack.cpp
```diff
@@ -167,6 +167,16 @@ TEST_CASE("Test behavior of sparse packs", "[SparsePack]") {
...
       }
 
+      THEN("A flattened sparse pack can correctly load this data in a unified outer "
+           "index space") {
+        using parthenon::variable_names::any;
+        std::vector<parthenon::MetadataFlag> flags{Metadata::WithFluxes};
+        auto sparse_pack = parthenon::SparsePack<any>::GetFlat(&mesh_data, flags);
+        REQUIRE(sparse_pack.GetNBlocks() == 1);
+        // v3 is deallocated on one block.
+        REQUIRE(sparse_pack.GetMaxNumberOfVars() == 5 * NBLOCKS - 3);
```
Yurlungur, line 186: `NBlocks` is 1 and `NumberOfVars` is `blocks*vars` in this mode.

src/interface/sparse_pack.hpp
```diff
@@ -276,6 +338,10 @@ class SparsePack : public SparsePackBase {
                    const int i) const {
     return pack_(0, b, idx)(k, j, i);
   }
+  KOKKOS_INLINE_FUNCTION
+  Real &operator()(int idx, const int k, const int j, const int i) const {
```
brryan, line 359: Would `PARTHENON_DEBUG_REQUIRE(flat_, ...);` be valid here?

src/interface/sparse_pack.hpp
```diff
@@ -276,6 +338,10 @@ class SparsePack : public SparsePackBase {
...
+    return pack_(0, 0, idx)(k, j, i);
+  }
 
   KOKKOS_INLINE_FUNCTION
   Real &operator()(const int b, PackIdx idx, const int k, const int j,
```
brryan: And would `PARTHENON_DEBUG_REQUIRE(!flat_, ...);` be valid here?

src/interface/sparse_pack_base.hpp
```diff
@@ -133,6 +137,10 @@ class SparsePackBase {
     }
     return map;
   }
+  KOKKOS_INLINE_FUNCTION
+  Real &operator()(int idx, const int k, const int j, const int i) const {
```
brryan: Another `PARTHENON_DEBUG_REQUIRE(flat_, ...);` here?

brryan (approved):
LGTM, for my own understanding, how is this intended to be used downstream with sparse variables? Is it just an `if` check for each variable on each meshblock and the work is skipped if that variable isnt allocated on that meshblock?

Yurlungur:
> how is this intended to be used downstream with sparse variables? Is it just an `if` check for each variable on each meshblock and the work is skipped if that variable isnt allocated on that meshblock?

It only packs allocated variables. So if for example, I have a total of 20 variables, but 3 variables are allocated on each of 5 blocks (where the set of variables allocated are different on each block), you get an outer index of length 15.

src/interface/sparse_pack.hpp
Yurlungur: hmm. probably a good idea

src/interface/sparse_pack.hpp
Yurlungur: yes good idea

src/interface/sparse_pack_base.hpp
Yurlungur: yeah good catch

lroberts36:
> the documentation for the SparsePack is very sparse

I see what you did there ;)

src/interface/sparse_pack.hpp
```diff
@@ -200,27 +252,32 @@ class SparsePack : public SparsePackBase {
   KOKKOS_INLINE_FUNCTION int GetLowerBound(const int b) const { return 0; }
 
   KOKKOS_INLINE_FUNCTION int GetUpperBound(const int b) const {
+    PARTHENON_DEBUG_REQUIRE(!flat_, "Bounds not valid for flat packs");
```
lroberts36: Are they not valid if `b==0`? I thought they were.

src/interface/sparse_pack_base.cpp
```diff
@@ -122,12 +123,19 @@ SparsePackBase SparsePackBase::Build(T *pmd, const PackDescriptor &desc) {
...
-  ForEachBlock(pmd, [&](int b, mbd_t *pmbd) {
-    int idx = 0;
-    coords_h(b) = pmbd->GetBlockPointer()->coords_device;
+  int idx = 0;
+  ForEachBlock(pmd, [&](int block, mbd_t *pmbd) {
+    int b = 0;
+    if (!desc.flat) {
+      idx = 0;
+      b = block;
+      coords_h(b) = pmbd->GetBlockPointer()->coords_device;
```
lroberts36, line 143: I guess this makes that no coords object is set for a flat pack since different blocks have different coords. Should we be preventing users from accessing coords from flat packs somehow (at least in debug mode)?

src/interface/sparse_pack_base.cpp
```diff
@@ -122,12 +123,19 @@ SparsePackBase SparsePackBase::Build(T *pmd, const PackDescriptor &desc) {
...
+    }
 
     for (int i = 0; i < nvar; ++i) {
-      pack.bounds_h_(0, b, i) = idx;
+      if (!desc.flat) {
+        pack.bounds_h_(0, b, i) = idx;
+      }
```
lroberts36: Can't we still set this with the actual block index, i.e. `pack.bounds_h_(0, block, i) = idx`? I am not sure what the use case would be, but it would allow all of the `Get*Bound` stuff to work returning the flat index.

src/interface/sparse_pack_base.cpp
```diff
@@ -148,18 +156,8 @@ SparsePackBase SparsePackBase::Build(T *pmd, const PackDescriptor &desc) {
                     PARTHENON_REQUIRE(pack_h(1, b, idx).size() ==
                                           pack_h(0, b, idx).size(),
```
lroberts36: I think this check can probably be removed as well. SparsePacks are probably not where we need to be ensuring that the fluxes have the correct size and we may move to the fluxes having a different size (to conform to the sizes of face variables for instance).

src/interface/sparse_pack_base.cpp
```diff
@@ -169,23 +167,25 @@ SparsePackBase SparsePackBase::Build(T *pmd, const PackDescriptor &desc) {
...
-        // Make the upper bound more negative so a for loop won't iterate once
-        pack.bounds_h_(1, b, i) = -2;
+      if (!desc.flat) {
+        pack.bounds_h_(1, b, i) = idx - 1;
+        if (pack.bounds_h_(1, b, i) < pack.bounds_h_(0, b, i)) {
+          // Did not find any allocated variables meeting our criteria
+          pack.bounds_h_(0, b, i) = -1;
+          // Make the upper bound more negative so a for loop won't iterate once
+          pack.bounds_h_(1, b, i) = -2;
+        }
```
lroberts36: Would have to change this back as well for the bounds to make sense.

Yurlungur:
@lroberts36 you should probably be the other person to review this---seeing as you wrote sparse packs.

src/interface/sparse_pack.hpp
lroberts36: Ok, after reading below I see they are not valid. I think we should make it so that they are valid, and maybe add a default `const int b=0` here so that `GetLowerBound()` and `GetUpperBound()` work for sparse packs (or maybe make them separate functions that have a check if `flat_==true`). I think this is clearer than exposing the pack size through  `GetMaxNumberOfVars()`.

src/interface/sparse_pack_base.cpp
```diff
@@ -169,23 +167,25 @@ SparsePackBase SparsePackBase::Build(T *pmd, const PackDescriptor &desc) {
...
       }
     }
     // Record the maximum for easy access
-    pack.bounds_h_(1, b, nvar) = idx - 1;
+    if (!desc.flat) {
+      pack.bounds_h_(1, b, nvar) = idx - 1;
+    }
```
lroberts36: If this is reverted back to `pack.bounds_h_(1, b, nvar) = idx - 1;`, `GetUpperBound(0)` should correctly return the upper bound of the flat index I think.

src/interface/sparse_pack_base.cpp
```diff
@@ -246,6 +244,10 @@ std::string SparsePackCache::GetIdentifier(const PackDescriptor &desc) const {
   identifier += "____";
   for (int i = 0; i < desc.vars.size(); ++i)
     identifier += desc.vars[i] + std::to_string(desc.use_regex[i]);
+  identifier += "____";
+  identifier += std::to_string(desc.with_fluxes);
+  identifier += std::to_string(desc.coarse);
+  identifier += std::to_string(desc.flat);
```
lroberts36, line 267: This is definitely better than rebuilding every time if there are coarse and fine packs with the same set of variables, for instance.

lroberts36 (changes requested):
This LGTM except that I think we should make sure the pack bounds arrays are meaningful even for flat packs.

src/interface/sparse_pack.hpp
lroberts36: Maybe better to just call them `GetFlatLowerBound()` and `GetFlatUpperBound()`.

src/interface/sparse_pack_base.cpp
lroberts36: Actually, maybe we need this to be `pack.bounds_h_(1, block, nvar) = idx - 1;` for all of the bounds functions to work as expected. The return value of `GetFlatUpperBound()` could just be `pack_.extent_int(2)`;

src/interface/sparse_pack_base.cpp
Yurlungur: Yes that's probably a good idea. That said, it occurs to me we could *can* set the coordinates in a self-consistent way... so maybe we should.

src/interface/sparse_pack_base.cpp
Yurlungur: Is this actually desired behaviour though? I wonder if `GetBounds` should return 0, and nvars always in the flat case? I could add an if statement to that effect maybe?

doc/sphinx/src/interface/sparse.rst
```diff
@@ -137,6 +137,16 @@ simulation. The set of possible operations for a sparse field are:
    ``lroberts36/merge-sparse-with-jdolence-sparse`` that is being used
    in ``Riot``. This should probably be brought into ``develop``, since
    the “sparse sparse pack” access pattern is probably not desirable.
+- *Flatten outer indices:* Most common packs over ``MeshData`` produce
+  a 5-dimensional data structure, where the slowest two moving indices
+  are over ``MeshBlock``s then tensor/vector components of
+  ``Variable``s. However, it is sometimes desirable to construct a
```
pgrete:
```suggestion
  are over ``MeshBlock``\ s then tensor/vector components of
  ``Variable``\ s. However, it is sometimes desirable to construct a
```

src/interface/sparse_pack.hpp
```diff
@@ -183,13 +183,65 @@ class SparsePack : public SparsePackBase {
     Get(pmd, vars, flags, fluxes, coarse);
   }
 
+  template <class T>
+  static SparsePack GetFlat(T *pmd, const std::vector<MetadataFlag> &flags = {}) {
+    const bool coarse = false;
+    const bool fluxes = false;
+    const bool flatten = true;
+    return Get(pmd, flags, fluxes, coarse, flatten);
```
pgrete, line 191: In the long run (e.g., when another parameter is being added, a small refactor to use a struct might be useful) -- not suggesting to do this now.

pgrete (approved):
Looks good to me.
I don't have a strong opinion with respect to the index handling.
In the spirit of our `ParArrayNDs` that transparently handle `0` I could see the appeal to also allow using the standard `()` operator and bounds being used with a `PARTHENON_DEBUG_REQUIRE(!flat || b == 0,...)`.

src/interface/sparse_pack.hpp
Yurlungur: I'm going to make them valid and just re-use the API. I'm not sure if we should come up with new or more generic names, I just chose not to for now. We can revisit that later maybe.

Yurlungur:
After incorporating @lroberts36 's suggestions, I broke a test, so I'm going to need a little time to debug. We should not wait for this PR for the next release. @pgrete .

src/interface/sparse_pack_base.cpp
Yurlungur: done

src/interface/sparse_pack_base.cpp
Yurlungur: Now updated `GetLowerBound` returns 0 and `GetUpperBound` returns total numbers of vars and blocks (i.e., pack size) for flat packs.

src/interface/sparse_pack_base.cpp
Yurlungur: Done

src/interface/sparse_pack_base.cpp
Yurlungur: This works as described now, though under the hood it's a little different.

Yurlungur:
@lroberts36 this is ready for re-review. I believe I've incorporated all your comments.

src/interface/sparse_pack_base.cpp
Yurlungur: yes. Fixed.

src/interface/sparse_pack.hpp
```diff
@@ -318,10 +398,27 @@ class SparsePack : public SparsePackBase {
...
     PARTHENON_DEBUG_REQUIRE(dir > 0 && dir < 4 && with_fluxes_, "Bad input to flux call");
     const int vidx = GetLowerBound(b, t) + t.idx;
     return pack_(dir, b, vidx)(k, j, i);
   }
+
+ private:
+  // Must make a copy of the vector, since input vector is const,
+  // and it may not even be an lvalue.
+  static std::vector<MetadataFlag> AddFlag_(const std::vector<MetadataFlag> &flags,
+                                            MetadataFlag mf = Metadata::WithFluxes) {
```
Yurlungur, line 445: This change was suggested by @lroberts36 offline. It lets us omit the `Metadata::WithFluxes` flag when calling `GetWithFluxes`, as it should be implied.

src/interface/sparse_pack.hpp
```diff
@@ -183,73 +183,133 @@ class SparsePack : public SparsePackBase {
...
   KOKKOS_FORCEINLINE_FUNCTION
   int GetMaxNumberOfVars() const { return pack_.extent_int(2); }
 
   KOKKOS_INLINE_FUNCTION
-  const Coordinates_t &GetCoordinates(const int b) const { return coords_(b)(); }
+  const Coordinates_t &GetCoordinates(const int b = 0) const { return coords_(b)(); }
 
   // Bound overloads
-  KOKKOS_INLINE_FUNCTION int GetLowerBound(const int b) const { return 0; }
+  KOKKOS_INLINE_FUNCTION int GetLowerBound(const int b = 0) const { return 0; }
```
lroberts36: I think this should change for flat packs. `GetLowerBound(b)` with `b > 0` should return `bounds_(0, b, 0)` and `GetUpperBound(b)` should return `bounds_(1, b, nvar_)`.

src/interface/sparse_pack.hpp
```diff
@@ -183,73 +183,133 @@ class SparsePack : public SparsePackBase {
...
 
-  KOKKOS_INLINE_FUNCTION int GetUpperBound(const int b) const {
-    return bounds_(1, b, nvar_);
+  KOKKOS_INLINE_FUNCTION int GetUpperBound(const int b = 0) const {
+    return bounds_(1, !flat_ * b, !flat_ * nvar_);
```
lroberts36: I don't think there should be any check for if the pack is flat here. This is why I think that there actually needs to be a new function that returns the total upper bound 
```c++
KOKKOS_INLINE_FUNCTION int GetUpperBound() const { 
  assert(flat_); 
  return bounds_(1, GetNBlocks() - 1, nvar_); 
}
```

src/interface/sparse_pack.hpp
```diff
@@ -183,73 +183,133 @@ class SparsePack : public SparsePackBase {
...
   }
 
   // Host Bound overloads
-  KOKKOS_INLINE_FUNCTION int GetLowerBoundHost(const int b) const { return 0; }
+  KOKKOS_INLINE_FUNCTION int GetLowerBoundHost(const int b = 0) const { return 0; }
 
-  KOKKOS_INLINE_FUNCTION int GetUpperBoundHost(const int b) const {
-    return bounds_h_(1, b, nvar_);
+  KOKKOS_INLINE_FUNCTION int GetUpperBoundHost(const int b = 0) const {
+    return bounds_h_(1, !flat_ * b, !flat_ * nvar_);
```
lroberts36: Also no flat check here.

src/interface/sparse_pack_base.cpp
```diff
@@ -179,13 +187,12 @@ SparsePackBase SparsePackBase::Build(T *pmd, const PackDescriptor &desc) {
       }
     }
     // Record the maximum for easy access
-    pack.bounds_h_(1, b, nvar) = idx - 1;
+    pack.bounds_h_(1, !desc.flat * b, !desc.flat * nvar) = idx - 1;
```
lroberts36:
```suggestion
    pack.bounds_h_(1, block, nvar) = idx - 1;
```

src/interface/sparse_pack_base.cpp
```diff
@@ -169,8 +176,9 @@ SparsePackBase SparsePackBase::Build(T *pmd, const PackDescriptor &desc) {
         }
       }
 
-      pack.bounds_h_(1, b, i) = idx - 1;
-
+      if (!desc.flat) {
+        pack.bounds_h_(1, b, i) = idx - 1;
+      }
```
lroberts36:
```suggestion
      pack.bounds_h_(1, b, i) = idx - 1;
```

src/interface/sparse_pack_base.cpp
```diff
@@ -114,20 +115,36 @@ SparsePackBase SparsePackBase::Build(T *pmd, const PackDescriptor &desc) {
...
+  ForEachBlock(pmd, [&](int block, mbd_t *pmbd) {
+    int b = 0;
+    if (!desc.flat) {
+      idx = 0;
+      b = block;
+      // JMM: This line could be unified with the coords_h line below,
+      // but it would imply unnecessary copies in the case of non-flat
+      // packs.
+      coords_h(b) = pmbd->GetBlockPointer()->coords_device;
+    }
```
lroberts36, line 144:
```suggestion
    if (!desc.flat) {
      idx = 0;
      b = block;
    }
    coords_h(block) = pmbd->GetBlockPointer()->coords_device;
```

src/interface/sparse_pack_base.cpp
```diff
@@ -114,20 +115,36 @@ SparsePackBase SparsePackBase::Build(T *pmd, const PackDescriptor &desc) {
...
 
     for (int i = 0; i < nvar; ++i) {
-      pack.bounds_h_(0, b, i) = idx;
+      if (!desc.flat) {
+        pack.bounds_h_(0, b, i) = idx;
+      }
```
lroberts36:
```suggestion
      pack.bounds_h_(0, block, i) = idx;
```

src/interface/sparse_pack_base.cpp
```diff
@@ -145,21 +162,11 @@ SparsePackBase SparsePackBase::Build(T *pmd, const PackDescriptor &desc) {
...
-                      PARTHENON_REQUIRE(pack_h(3, b, idx).size() ==
-                                            pack_h(0, b, idx).size(),
-                                        "Different size fluxes.");
-                    }
+                    pack_h(2, b, idx) = pv->flux[2].Get(t, u, v);
+                    pack_h(3, b, idx) = pv->flux[3].Get(t, u, v);
+                  }
+                  if (desc.flat) {
+                    coords_h(idx) = pmbd->GetBlockPointer()->coords_device;
                   }
```
lroberts36:
```suggestion
```

src/interface/sparse_pack_base.cpp
```diff
@@ -114,20 +115,36 @@ SparsePackBase SparsePackBase::Build(T *pmd, const PackDescriptor &desc) {
...
 
+  // For non-flat packs, shape of pack is type x block x var x k x j x i
+  // where type here might be a flux.
+  // For flat packs, shape is type x (some var on some block)  x k x j x 1
+  // in the latter case, coords indexes into the some var on some
+  // block but bounds is reinterpreted as the lower and upper bounds
+  // of the (some var on some block) index range.
   // Size is nvar + 1 to store the maximum idx for easy access
-  pack.bounds_ = bounds_t("bounds", 2, nblocks, nvar + 1);
+  pack.bounds_ = bounds_t("bounds", 2, desc.flat ? 1 : nblocks, !desc.flat * nvar + 1);
```
lroberts36:
```suggestion
  pack.bounds_ = bounds_t("bounds", 2, nblocks, nvar + 1);
```

src/interface/sparse_pack_base.cpp
```diff
@@ -114,20 +115,36 @@ SparsePackBase SparsePackBase::Build(T *pmd, const PackDescriptor &desc) {
...
   pack.bounds_h_ = Kokkos::create_mirror_view(pack.bounds_);
 
-  pack.coords_ = coords_t("coords", nblocks);
+  pack.coords_ = coords_t("coords", desc.flat ? max_size : nblocks);
```
lroberts36, line 130:
```suggestion
  pack.coords_ = coords_t("coords", nblocks);
```

lroberts36 (commented):
I think we were on a different page about how to keep the (block, variable) indexing working for flat packs. I included some possible changes that are along the lines of what I was thinking. I think we probably don't want to do it because of the extra H2D copy, but another array could be added that maps from flat idx to block idx (along with functions for querying the pack for which block an flat idx corresponds to) so that e.g. the correct coordinates could be accessed from a flat pack.

Yurlungur:
@lroberts36 I don't think I understand what you want. Can we talk over MM or something?

lroberts36:
yep, I am on MM

Yurlungur:
@lroberts36 take a look now. I've re-enabled original functionality for `GetLowerBounds` and `GetUpperBounds` as we discussed. Also added `GetSize` as a transparent way to get total size of the array.

src/interface/sparse_pack_base.cpp
lroberts36: I think this is potentially confusing that for regular backs `GetCoords` takes the block number as an argument while for flat packs it takes the variable flat index.

src/interface/sparse_pack_base.hpp
```diff
@@ -133,6 +137,11 @@ class SparsePackBase {
...
+    PARTHENON_DEBUG_REQUIRE(flat_, "Accessor only valid for flat packs");
+    return pack_(0, 0, idx)(k, j, i);
+  }
```
lroberts36: Is there a reason for providing `operator()` in `sparse_pack_base.hpp`? I was trying to confine the interface to just `SparsePack`.

lroberts36 (approved):
There are a couple more comments/suggestions below, but this LGTM and I approve

src/interface/sparse_pack_base.hpp
Yurlungur: I think this is a mistake. I'll remove it.

Yurlungur:
@pgrete the cuda regression test appears to be failing, without a useful error message, but it works on my machine. I'll dig in a little bit, but I wonder if something is wrong with the CI machine?

src/interface/sparse_pack_base.cpp
Yurlungur: I want to be able to make this call, as I think there's a clear use-case: i.e., you need coords for your stencil op. However, I would be willing to give `GetCoords` two different names?

Yurlungur:
Weird problem went away. Pulling the trigger.

PR #879, opened 2023-05-24
Zero buffer allocation and (nearly) in-one AMR 

### LINKED ISSUE #825: AMR prolong/restriction may not work for rank >=2 variables
Looking at the final changes in #818 I noticed https://github.com/parthenon-hpc-lab/parthenon/blob/b866f72e80b7fa6133bdcf4ca3cd65d6df276235/src/mesh/amr_loadbalance.cpp#L1033

On first sight, this looks like it may be an issue for rank >= 2 variables as the data is not packed but directly accessed.
Do you agree @Yurlungur @lroberts36 @pdmullen ?

This probably also applies to
https://github.com/parthenon-hpc-lab/parthenon/blob/b866f72e80b7fa6133bdcf4ca3cd65d6df276235/src/mesh/amr_loadbalance.cpp#L510
https://github.com/parthenon-hpc-lab/parthenon/blob/b866f72e80b7fa6133bdcf4ca3cd65d6df276235/src/mesh/amr_loadbalance.cpp#L872
https://github.com/parthenon-hpc-lab/parthenon/blob/b866f72e80b7fa6133bdcf4ca3cd65d6df276235/src/mesh/amr_loadbalance.cpp#L940
https://github.com/parthenon-hpc-lab/parthenon/blob/b866f72e80b7fa6133bdcf4ca3cd65d6df276235/src/mesh/amr_loadbalance.cpp#L976
https://github.com/parthenon-hpc-lab/parthenon/blob/b866f72e80b7fa6133bdcf4ca3cd65d6df276235/src/mesh/amr_loadbalance.cpp#L1084
https://github.com/parthenon-hpc-lab/parthenon/blob/b866f72e80b7fa6133bdcf4ca3cd65d6df276235/src/mesh/amr_loadbalance.cpp#L1130
https://github.com/parthenon-hpc-lab/parthenon/blob/b866f72e80b7fa6133bdcf4ca3cd65d6df276235/src/mesh/amr_loadbalance.cpp#L1204
https://github.com/parthenon-hpc-lab/parthenon/blob/b866f72e80b7fa6133bdcf4ca3cd65d6df276235/src/mesh/amr_loadbalance.cpp#L1247

lroberts36:
## PR Summary
This PR ports the version of `RedistributeAndRefineMeshBlocks` I wrote a while back on the Riot branch. It does two things:
1. Removes the need for allocating any buffers for communicating block data during mesh refinement. This potentially gives a significant memory savings when using sparse data.
2. Reorganizes things so that all of the restrictions happen in one place at the beginning of the routine and all of the prolongations happen at the end of the routine, which puts us very close to switching to using prolongation and restriction in-one. 

@jdolence and I wrote two versions of the low/zero storage communication routines for Riot. The one he wrote was very slightly faster on the small set of problems we ran with both codes (at most 3% faster), but does not restructure where prolongation and restriction occurs. It is also a somewhat more incremental change to `RedistributeAndRefineMeshBlocks`. If you would like to take a look at that code, it is on lroberts36/merge-sparse-with-jdolence-sparse. I am not wedded to the version of the code in this PR, so please take a look and see if you think this is a good direction or if we should just close this PR without merging.

The main reason I put this PR up is that I would like to have an updated version of `RedistributeAndRefineMeshBlocks` in before I make the changes required to do mesh refinement with non-cell centered variables in #872.  

This also fixes a bug for some MPI implementations in an `MPI_Allreduce` call.

- [x] Test after #890 is in and make sure this PR gives bitwise similar results

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


Yurlungur (approved):
This is a pretty substantive cleanup, and a needed re-org. I am in favor of these changes. I can't work out all the possible edge cases by reading the code, but hopeful that the automated tests would have caught any issues.

lroberts36:
>  but hopeful that the automated tests would have caught any issues.

Oh, don't worry. They are catching lots of things...

lroberts36:
Ok, automated tests defeated.

src/outputs/output_utils.cpp
```diff
@@ -137,7 +137,7 @@ std::size_t MPISum(std::size_t val) {
                 "size_t is unsigned and integral");
   static_assert(sizeof(std::size_t) == sizeof(unsigned long long int),
                 "MPI_UNSIGNED_LONG_LONG same as size_t");
-  MPI_Allreduce(&val, &val, 1, MPI_UNSIGNED_LONG_LONG, MPI_SUM, MPI_COMM_WORLD);
+  MPI_Allreduce(MPI_IN_PLACE, &val, 1, MPI_UNSIGNED_LONG_LONG, MPI_SUM, MPI_COMM_WORLD);
```
pgrete: This should probably also have an `PARTHENON_MPI_CHECK` around.

src/mesh/amr_loadbalance.cpp
```diff
@@ -45,6 +45,241 @@
...
+  const int ox1 = ((fine_loc.lx1 & 1LL) == 1LL);
+  const int ox2 = ((fine_loc.lx2 & 1LL) == 1LL);
+  const int ox3 = ((fine_loc.lx3 & 1LL) == 1LL);
+  int tag = CreateAMRMPITag(lid_recv, ox1, ox2, ox3);
+  if (var->IsAllocated()) {
+    PARTHENON_MPI_CHECK(MPI_Isend(var->data.data(), var->data.size(), MPI_PARTHENON_REAL,
+                                  dest_rank, tag, comm, &req));
+  } else {
+    PARTHENON_MPI_CHECK(
+        MPI_Isend(var->data.data(), 0, MPI_PARTHENON_REAL, dest_rank, tag, comm, &req));
```
pgrete, line 75: Do we conceptually need those (so that we can check if the comms are done)?
Does this translate to a nullop?

src/mesh/amr_loadbalance.cpp
```diff
@@ -45,6 +45,241 @@
...
+    PARTHENON_MPI_CHECK(MPI_Get_count(&status, MPI_PARTHENON_REAL, &size));
+#endif
+    if (size > 0) {
+      if (!pmb->IsAllocated(var->label())) pmb->AllocateSparse(var->label());
+      auto fb = var_in->data;
+#ifdef MPI_PARALLEL
+      PARTHENON_MPI_CHECK(MPI_Recv(var->data.data(), var->data.size(), MPI_PARTHENON_REAL,
+                                   send_rank, tag, comm, MPI_STATUS_IGNORE));
+      fb = var->data;
+#endif
```
pgrete, line 117: can `fb` just be set once after the `#endif`?

src/mesh/amr_loadbalance.cpp
```diff
@@ -45,6 +45,241 @@
...
+    if (size > 0) {
+      if (!pmb->IsAllocated(var->label())) pmb->AllocateSparse(var->label());
+      auto cb = var_in->coarse_s;
+#ifdef MPI_PARALLEL
+      // This has to be an MPI_Recv w/o buffering
+      PARTHENON_MPI_CHECK(MPI_Recv(var->coarse_s.data(), var->coarse_s.size(),
+                                   MPI_PARTHENON_REAL, send_rank, tag, comm,
+                                   MPI_STATUS_IGNORE));
+      cb = var->coarse_s;
+#endif
```
pgrete, line 202: as above: is one `cb =` statement after the `#endif` enough?

src/mesh/amr_loadbalance.cpp
```diff
@@ -45,6 +45,241 @@
...
+      const int ks = (ox3 == 0) ? 0 : (kb_int.e - kb_int.s + 1) / 2;
+      const int js = (ox2 == 0) ? 0 : (jb_int.e - jb_int.s + 1) / 2;
+      const int is = (ox1 == 0) ? 0 : (ib_int.e - ib_int.s + 1) / 2;
+      auto cb = var->coarse_s;
+      const int nt = fb.GetDim(6) - 1;
+      const int nu = fb.GetDim(5) - 1;
+      const int nv = fb.GetDim(4) - 1;
+      parthenon::par_for(
+          DEFAULT_LOOP_PATTERN, "FillSameRankCoarseToFineAMR", DevExecSpace(), 0, nt, 0,
```
pgrete: the name is not correct, isn't it?

src/mesh/amr_loadbalance.cpp
```diff
@@ -45,6 +45,241 @@
...
+                                   MPI_PARTHENON_REAL, send_rank, tag, comm,
+                                   MPI_STATUS_IGNORE));
+      cb = var->coarse_s;
+#endif
+      auto fb = var->data;
+      const int nt = fb.GetDim(6) - 1;
+      const int nu = fb.GetDim(5) - 1;
+      const int nv = fb.GetDim(4) - 1;
+      parthenon::par_for(
+          DEFAULT_LOOP_PATTERN, "FillSameRankCoarseToFineAMR", DevExecSpace(), 0, nt, 0,
```
pgrete: name?

src/mesh/amr_loadbalance.cpp
```diff
@@ -45,6 +45,241 @@
...
+          nu, 0, nv, kb.s, kb.e, jb.s, jb.e, ib.s, ib.e,
+          KOKKOS_LAMBDA(const int t, const int u, const int v, const int k, const int j,
+                        const int i) {
+            fb(t, u, v, k + ks, j + js, i + is) = cb(t, u, v, k, j, i);
+          });
+      // We have to block here w/o buffering so that the write is guaranteed to be
+      // finished before we get here again
```
pgrete: Why/how could we get in here again?

src/mesh/amr_loadbalance.cpp
```diff
@@ -45,6 +45,241 @@
...
+}
+
+#ifdef MPI_PARALLEL
+MPI_Request SendSameToSame(int lid_recv, int dest_rank, Variable<Real> *var,
+                           MeshBlock *pmb, Mesh *pmesh) {
+  MPI_Request req;
+  MPI_Comm comm = pmesh->GetMPIComm(var->label());
+  int tag = CreateAMRMPITag(lid_recv, 0, 0, 0);
+  if (var->IsAllocated()) {
+    // This assumes we have two ghost zones
```
pgrete: Does this imply that running with `nghost != 2` does not work correctly any more?
If so, I'd consider this blocking (though the fix should be straightforward, isn't it)?

pgrete: Also could you please add a comment above those lines for what they're doing.
"Update (de)reference counter" (or similar -- if I follow correctly)

src/mesh/amr_loadbalance.cpp
```diff
@@ -462,368 +697,183 @@ void Mesh::RedistributeAndRefineMeshBlocks(ParameterInput *pin, ApplicationInput
...
+        }
       }
-    } else {
-      if (ranklist[on] != Globals::my_rank) nrecv++;
     }
   }
+  Kokkos::fence();
+
+#ifdef MPI_PARALLEL
+  // Step 7. Send data from old to new blocks
```
pgrete: We got two Steps 7 now

src/mesh/amr_loadbalance.cpp
```diff
@@ -462,368 +697,183 @@ void Mesh::RedistributeAndRefineMeshBlocks(ParameterInput *pin, ApplicationInput
...
-#ifdef MPI_PARALLEL
-  if (nrecv != 0) {
-    int test;
-    std::vector<bool> received(nrecv, false);
-    int rb_idx;
+  bool all_received;
+  int niter = 0;
+  if (block_list.size() > 0) {
+    std::vector<bool> finished(
+        std::max((nbe - nbs + 1), 1) * FindMeshBlock(nbs)->vars_cc_.size() * 8, false);
```
pgrete, line 816: Add comment? (It's not immediately obvious to me where the `>vars_cc_.size() * 8` comes from.

src/mesh/amr_loadbalance.cpp
```diff
@@ -462,368 +697,183 @@ void Mesh::RedistributeAndRefineMeshBlocks(ParameterInput *pin, ApplicationInput
...
           }
-          rb_idx++;
         }
       }
       // rb_idx is a running index, so we repeat the loop until all vals are true
-    } while (!std::all_of(received.begin(), received.begin() + rb_idx,
-                          [](bool v) { return v; }));
-    Kokkos::fence();
+    } while (!all_received && niter < 1e7);
+    if (!all_received) PARTHENON_FAIL("AMR Receive failed");
```
pgrete, line 867: No need to fence here any more?

pgrete (commented):
Nice cleanup @lroberts36 
I'm in favor of this approach and think that it makes the code also more readable.
I just got a couple of questions for clarification.

src/mesh/amr_loadbalance.cpp
lroberts36: Yeah, these are required. The receiver has to look for a message from every field, allocated or not, since it doesn't know the allocation status *a priori*. The zero size message still gets sent (i.e. it is not a null op), the receiver can probe for it and check its size, and if it is zero it knows the field was not allocated on the sending block.  We could send a single message from each block to each other block containing the allocation status of all fields that are being sent (and get rid of the zero size messages), but unless there is a clear performance win I think it is not worth it since it makes the code a bit more complex and coupled.

src/mesh/amr_loadbalance.cpp
lroberts36: No, if there is no MPI it needs to point to the data of a variable on another block (`var_in->data`) while with MPI the data from the other block is received via MPI into `var->data`. The real issue here might be that I am still using MPI communication for same rank to same rank. That could be changed by putting in a second `if` statement in a few places though, so that data is copied directly.

src/mesh/amr_loadbalance.cpp
lroberts36: See above.

src/outputs/output_utils.cpp
lroberts36: done

src/mesh/amr_loadbalance.cpp
lroberts36: fixed

src/mesh/amr_loadbalance.cpp
lroberts36: fixed

src/mesh/amr_loadbalance.cpp
lroberts36: This should read "at *least* two ghost zones", sorry. I have added a comment to the code explaining what is going on and added a `PARTHENON_REQUIRE(nghost>1)`. Let me know if you still think this is blocking.

src/mesh/amr_loadbalance.cpp
lroberts36: We have 7 - eps and 7 now ;). 

But we also no longer have steps 3 through 6, I think I will just remove the Step comments.

src/mesh/amr_loadbalance.cpp
lroberts36: This is just to make sure that this array has enough space in the worst case scenario (where every block on the rank is a new coarser block so that it is receiving a message from eight separate finer blocks for each variable). I have added a comment to this effect since I also forgot where the magic 8 comes from.

src/mesh/amr_loadbalance.cpp
lroberts36: Isn't this fence required in the old code because `recvbuf` is used in the set kernels but is then deleted below?

src/mesh/amr_loadbalance.cpp
lroberts36: When writing from fine to coarse, there are eight old fine blocks that write to a single coarse block. Each of these correspond to a `TryRecvFineToCoarse` call that does an `MPI_Recv` into the same coarse buffer. If the set kernel hasn't finished, the coarse buffer data can be overwritten before it is copied over. I have updated the comment to be more clear about this.

This implementation was meant to minimize memory usage, but this may not be the ideal from a performance perspective. If this becomes an issue, we could make a pool of buffers that we can pull from and only release them at the end of `RedistributeAndRefineMeshBlocks` (after a `Kokkos::fence()`).

src/mesh/amr_loadbalance.cpp
pgrete: Thanks for the explanation. Makes sense. The "packing" message might end up on our top performance concerns sooner or later anyway.

src/mesh/amr_loadbalance.cpp
pgrete: My bad. I missed the distinction between `var` and `var_in`.
Regarding using MPI on the same rank, I opened https://github.com/parthenon-hpc-lab/parthenon/issues/886 to keep a track.

src/mesh/amr_loadbalance.cpp
pgrete: Understood. Thanks for the clarification.

src/mesh/amr_loadbalance.cpp
pgrete: Perfect.

src/mesh/amr_loadbalance.cpp
pgrete: IIRC we added this because (under the assumption we'd allow for multiple device execution spaces. e.g., cuda streams) all data transfers should be done prior to additional operations.
I don't think that a single fence here would hurt performance, but it might be a safer choice (in the long run) given that  `ProlongateCellCenteredValues` is called right afterwards.
What do you think? (and feel free to merge).

pgrete (approved):
Thanks for the clarifications. One last comment left but feel free to merge.

src/mesh/amr_loadbalance.cpp
lroberts36: I agree that it is probably not a performance bottleneck and may save us from issues down the road. I will add a fence before prolongation.

pgrete:
@lroberts36 Should we re-review (as this was in WIP mode again in between)?

lroberts36:
@pgrete It is probably worth a quick second look through. Nothing about the structure of the PR changed, but there are a couple of small edge cases I caught related to sparse variables. Additionally, this now relies on #890 being in before it can be merged, since #890 fixes some sparse issues with boundary communication.

src/mesh/amr_loadbalance.cpp
```diff
@@ -45,6 +45,255 @@
...
+  MPI_Comm comm = pmesh->GetMPIComm(var->label());
+  int tag = CreateAMRMPITag(lid_recv, 0, 0, 0);
+
+  int test;
+  MPI_Status status;
+  PARTHENON_MPI_CHECK(MPI_Iprobe(send_rank, tag, comm, &test, &status));
+  if (test) {
+    int size;
+    PARTHENON_MPI_CHECK(MPI_Get_count(&status, MPI_PARTHENON_REAL, &size));
+    if (size > 2) {
```
pgrete, line 276: There is no corner case where this could be an issue, is there?
Or would this trigger for 1D sims with `nghost=2`?

src/mesh/amr_loadbalance.cpp
```diff
@@ -449,383 +698,205 @@ void Mesh::RedistributeAndRefineMeshBlocks(ParameterInput *pin, ApplicationInput
...
+    int on = newtoold[nn];
+    LogicalLocation &oloc = loclist[on];
+    LogicalLocation &nloc = newloc[nn];
+    auto pmb = FindMeshBlock(nn);
+    if (nloc.level > oloc.level) {
+      const IndexDomain interior = IndexDomain::interior;
+      IndexRange cib = pmb->c_cellbounds.GetBoundsI(interior);
+      IndexRange cjb = pmb->c_cellbounds.GetBoundsJ(interior);
+      IndexRange ckb = pmb->c_cellbounds.GetBoundsK(interior);
+      // Need to restrict this block before doing sends
```
pgrete, line 884: This comment is not up-to-date any more isn't it?

pgrete (approved):
LGTM
Will enable auto-merge now (the two minor comments are for a posteriori clarification)

src/mesh/amr_loadbalance.cpp
lroberts36: I 1D think you would have to have `2*nghost + nx <= 2` which I think is impossible.

src/mesh/amr_loadbalance.cpp
lroberts36: Oh, yeah. That looks like something I copy and pasted from the restriction above.

Issue #376, opened 2020-11-20
Do the remaining work to thread the new boundary machinery through the `MeshData` and `MeshBlockData` machinery


Yurlungur:
Needed as an enhancement of #332 to include MRs #358 and #359 :
- Currently the boundary machinery accepts only `MeshBockData`, not `MeshData`. This should be rectified at some point.
- Prolongation and Restriction need to be `pack_in_one`-ified.

PR #884, opened 2023-06-08
Add constant derivative boundary condition and expose `GenericBC`

lroberts36:
## PR Summary

This PR adds:
 - The option of a constant derivative boundary condition (which I needed when experimenting with evolving the vector potential).
 - The option of a constant value boundary condition (which I also needed at some point)
 - Slightly refactors the internal pieces of `GenericBC` to use sparse packs and allow for the selection of which variables to apply boundary conditions on 
 - Moves `GenericBC` to a header so that downstream codes can compose the generic boundary conditions for subsets of variables. This should make it easier to create user defined boundary conditions that behave differently for different variables, e.g.
```c++
 pman.app_input->boundary_conditions[parthenon::BoundaryFace::inner_x1] = 
    [](std::shared_ptr<MeshBlockData<Real>> &rc, bool coarse) { 
      using namespace parthenon; 
      using namespace parthenon::BoundaryFunction; 
      GenericBC<X1DIR, BCSide::Inner, BCType::Outflow, variable_names::any>(rc, coarse); 
      GenericBC<X1DIR, BCSide::Inner, BCType::ConstantDeriv, myvar_t>(rc, coarse); 
      GenericBC<X1DIR, BCSide::Inner, BCType::Fixed, myvar2_t>(rc, coarse, 1.0); 
    };
```
- Additionally, it fixes a small bug in sparse packs, ensures that the vector components of fields in a sparse  pack are correctly set, exposes `MeshData` level `ApplyBoundaryConditions` tasks, and removes some unused code.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [ ] Docs build


Yurlungur (approved):
This is a nice improvement. Thanks for adding it. Assuming everything works, I'm in favor. :+1:

lroberts36:
This at least partially closes #376

lroberts36:
I used this in a downstream code that does CT using a vector potential (which requires a constant derivative boundary condition to give outflow boundary conditions for B, while the other conserved variables have an outflow boundary condtion) and everything seems to work well.

src/bvals/boundary_conditions_generic.hpp
```diff
@@ -0,0 +1,125 @@
...
+  // used for reflections
+  const int offset = 2 * ref + (INNER ? -1 : 1);
+
+  // used for derivatives
+  const int offsetin = INNER;
+  const int offsetout = !INNER;
+  pmb->par_for_bndry(
+      label, nb, domain, el, coarse,
+      KOKKOS_LAMBDA(const int &l, const int &k, const int &j, const int &i) {
+        if (TYPE == BCType::Reflect) {
```
brryan, line 92: good place for an `if constexpr`

brryan (approved):
LGTM, this is nice to have. I think the `if constexpr` change would be worthwhile performance-wise if you agree and don't mind doing that before merge.

I thought about whether we should have a bunch more functions like `ConstantInnerX1` etc. but I'm not sure that's a good idea. As in your use case different variables often have different BCs. Maybe the right way to do this would be to tag variables with their boundary conditions in their metadata, and use that to construct a BC function at runtime? Not a task for this PR of course.

Yurlungur:
>  As in your use case different variables often have different BCs. Maybe the right way to do this would be to tag variables with their boundary conditions in their metadata, and use that to construct a BC function at runtime? Not a task for this PR of course.

I have a plan for this... metadata is one option, per-package BCs would be another. I've been meaning to refactor the code to add this forever, but I keep putting it off...

src/bvals/boundary_conditions_generic.hpp
lroberts36: I tried this but it created all sorts of crazy GPU compiler errors. I think this is because `if constexpr` implies that the code inside the statement doesn't even get compiled if it evaluates to false (I don't have an explanation of why this causes crazy cuda errors, but it seems plausible). I *think* there shouldn't be a substantial performance difference without the `constexpr` since the branch prediction should be very easy.

brryan: Huh OK well if it doesn't work it doesn't work -- maybe I don't understand all the subtleties of `if constexpr`. Happy to drop this for now at least.

pdmullen (commented):
LGTM 💯

pdmullen (approved):

PR #885, opened 2023-06-11
Separate and expose `PackDescriptor` for `SparsePack`s

lroberts36:
## PR Summary

Implements the updates to the `SparsePack` interface as discussed by myself, @Yurlungur, @pdmullen, and @jdolence. `PackDescriptor`s are now exposed to users and can be created separately from the sparse pack itself. Now, when a pack is actually built based on a `PackDescriptor`, it is done based on the `Uid_t` of variables so there should be no expensive string comparisons. 

```c++
auto desc = parthenon::MakePackDescriptor<v1, v2>(&my_package_descriptor, {Metadata::WithFluxes}, {PDOpt::WithFluxes});
...
auto sparse_pack = desc.MakePack(&mesh_data);
```

Currently, everything is working but I think there are maybe a couple of features that I forgot to add that were discussed on Friday. Currently, it lets you define a variable selector function in a `MakePackDescriptor` free function so that it is straightforward to write new selectors (e.g. if you are not a fan of regexes). It also adds a function to `SparsePack`s that allows you to pull out pointers to variables in a single line via a structured binding, as suggested by @pdmullen. 

One thing we could do is explicitly sort the variables within each variable group in a pack (by allowing for a custom comparator to be passed to `PackDescriptor`). This would allow downstream codes to guarantee ordering of variables within the pack, which is currently required in some downstream codes.

## PR Checklist

- [x] Code passes cpplint
- [x] Updates tests for new features.
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/src/interface/meshblock_data.hpp
+++ b/src/interface/meshblock_data.hpp
@@ -85,6 +85,7 @@ class MeshBlockData {
   }
   auto GetParentPointer() const { return GetBlockPointer(); }
   void SetAllowedDt(const Real dt) const { GetBlockPointer()->SetAllowedDt(dt); }
+  Mesh *GetMeshPointer() const { return GetBlockPointer()->pmy_mesh; }
 
   IndexRange GetBoundsI(const IndexDomain &domain) const {
     return GetBlockPointer()->cellbounds.GetBoundsI(domain);
--- a/src/interface/meshblock_data.hpp
+++ b/src/interface/meshblock_data.hpp
@@ -151,7 +151,7 @@ class MeshBlockData {
     return varUidMap_.at(uid);
   }
 
-  auto &GetUidMap() { return varUidMap_; }
+  const auto &GetUidMap() const { return varUidMap_; }
 
   Variable<T> &Get(const std::string &base_name, int sparse_id = InvalidSparseID) const {
     return *GetVarPtr(MakeVarLabel(base_name, sparse_id));
--- a/src/interface/meshblock_data.hpp
+++ b/src/interface/meshblock_data.hpp
@@ -151,6 +151,8 @@ class MeshBlockData {
     return varUidMap_.at(uid);
   }
 
+  auto &GetUidMap() { return varUidMap_; }
+
   Variable<T> &Get(const std::string &base_name, int sparse_id = InvalidSparseID) const {
     return *GetVarPtr(MakeVarLabel(base_name, sparse_id));
   }
--- a/src/interface/sparse_pack.hpp
+++ b/src/interface/sparse_pack.hpp
@@ -1,5 +1,5 @@
 //========================================================================================
-// (C) (or copyright) 2020-2022. Triad National Security, LLC. All rights reserved.
+// (C) (or copyright) 2020-2023. Triad National Security, LLC. All rights reserved.
 //
 // This program was produced under U.S. Government contract 89233218CNA000001 for Los
 // Alamos National Laboratory (LANL), which is operated by Triad National Security, LLC
--- a/src/interface/sparse_pack.hpp
+++ b/src/interface/sparse_pack.hpp
@@ -118,7 +118,20 @@ class SparsePack : public SparsePackBase {
   SparsePack() = default;
 
   explicit SparsePack(const SparsePackBase &spb) : SparsePackBase(spb) {}
-
+  
+  class Descriptor : public impl::PackDescriptor {
+   public: 
+    Descriptor(const impl::PackDescriptor& desc_in) : impl::PackDescriptor(desc_in) {}
+    template <class T>
+    SparsePack MakePack(T* pmd) const {
+      return SparsePack(SparsePackBase::GetPack(pmd, *this));
+    }
+    SparsePackIdxMap GetMap() const {
+      PARTHENON_REQUIRE(sizeof...(Ts) == 0, "Should not be getting an IdxMap for a type based pack"); 
+      return SparsePackBase::GetIdxMap(*this);
+    }
+  };
+  /*
   // Make a `SparsePack` from variable_name types in the type list Ts..., creating the
   // pack in `pmd->SparsePackCache` if it doesn't already exist. Variables can be
   // accessed on device via instance of types in the type list Ts...
@@ -126,7 +139,8 @@ class SparsePack : public SparsePackBase {
   template <class T>
   static SparsePack Get(T *pmd, const std::vector<MetadataFlag> &flags = {},
                         bool fluxes = false, bool coarse = false, bool flatten = false) {
-    const impl::PackDescriptor desc(std::vector<std::string>{Ts::name()...},
+    const impl::PackDescriptor desc(pmd->GetMeshPointer()->resolved_packages.get(), 
+                                    std::vector<std::string>{Ts::name()...},
                                     std::vector<bool>{Ts::regex()...}, flags, fluxes,
                                     coarse, flatten);
     return SparsePack(SparsePackBase::GetPack(pmd, desc));
@@ -145,7 +159,8 @@ class SparsePack : public SparsePackBase {
   Get(T *pmd, const VAR_VEC &vars, const std::vector<MetadataFlag> &flags = {},
       bool fluxes = false, bool coarse = false, bool flatten = false) {
     static_assert(sizeof...(Ts) == 0, "Cannot create a string/type hybrid pack");
-    impl::PackDescriptor desc(vars, flags, fluxes, coarse, flatten);
+    impl::PackDescriptor desc(pmd->GetMeshPointer()->resolved_packages.get(),
+      vars, flags, fluxes, coarse, flatten);
     return {SparsePack(SparsePackBase::GetPack(pmd, desc)),
             SparsePackBase::GetIdxMap(desc)};
   }
@@ -237,7 +252,7 @@ class SparsePack : public SparsePackBase {
     const bool flatten = true;
     return Get(pmd, vars, flags, fluxes, coarse, flatten);
   }
-
+  */
   // Methods for getting parts of the shape of the pack
   KOKKOS_FORCEINLINE_FUNCTION
   int GetNBlocks() const { return nblocks_; }
@@ -438,6 +453,29 @@ class SparsePack : public SparsePackBase {
     return pack_(dir, b, vidx)(k, j, i);
   }
 };
+
+enum class PDOpt {WithFluxes, Coarse, Flatten};
+template <class... Ts> 
+inline auto MakePackDescriptor(StateDescriptor *psd, 
+                               const std::vector<MetadataFlag> &flags = {},
+                               bool fluxes = false, 
+                               bool coarse = false, 
+                               bool flatten = false) { 
+  static_assert(sizeof...(Ts) > 0, "Must have at least one variable type for type pack");
+  impl::PackDescriptor base_desc(psd, std::vector<std::string>{Ts::name()...},
+                                 std::vector<bool>{Ts::regex()...}, flags, fluxes,
+                                 coarse, flatten); 
+  return typename SparsePack<Ts...>::Descriptor(base_desc);
+}
+
+template <class VAR_VEC>
+inline auto MakePackDescriptor(StateDescriptor *psd, 
+                               const VAR_VEC &vars, 
+                               const std::vector<MetadataFlag> &flags = {},
+                               bool fluxes = false, bool coarse = false, bool flatten = false) {
+  impl::PackDescriptor base_desc(psd, vars, flags, fluxes, coarse, flatten);
+  return typename SparsePack<>::Descriptor(base_desc);
+}
 
 } // namespace parthenon
 
--- a/src/interface/sparse_pack.hpp
+++ b/src/interface/sparse_pack.hpp
@@ -344,27 +344,75 @@ class SparsePack : public SparsePackBase {
 
 enum class PDOpt { WithFluxes, Coarse, Flatten };
 
+inline auto MakePackDescriptor(StateDescriptor *psd, const std::vector<std::string> &vars,
+                               const std::vector<bool> &use_regex,
+                               const std::vector<MetadataFlag> &flags = {},
+                               const std::set<PDOpt> &options = {}) {
+  PARTHENON_REQUIRE(vars.size() == use_regex.size(),
+                    "Vargroup names and use_regex need to be the same size.");
+  auto selector = [&](int vidx, const VarID &id, const Metadata &md) {
+    if (flags.size() > 0) {
+      for (const auto &flag : flags) {
+        if (!md.IsSet(flag)) return false;
+      }
+    }
+
+    if (use_regex[vidx]) {
+      if (std::regex_match(std::string(id.label()), std::regex(vars[vidx]))) return true;
+    } else {
+      if (vars[vidx] == id.label()) return true;
+      if (vars[vidx] == id.base_name && id.sparse_id != InvalidSparseID) return true;
+    }
+    return false;
+  };
+
+  impl::PackDescriptor base_desc(psd, vars, selector, options.count(PDOpt::WithFluxes),
+                                 options.count(PDOpt::Coarse),
+                                 options.count(PDOpt::Flatten));
+  return typename SparsePack<>::Descriptor(base_desc);
+}
+
 template <class... Ts>
 inline auto MakePackDescriptor(StateDescriptor *psd,
                                const std::vector<MetadataFlag> &flags = {},
                                const std::set<PDOpt> &options = {}) {
   static_assert(sizeof...(Ts) > 0, "Must have at least one variable type for type pack");
-  impl::PackDescriptor base_desc(
-      psd, std::vector<std::string>{Ts::name()...}, std::vector<bool>{Ts::regex()...},
-      flags, options.count(PDOpt::WithFluxes), options.count(PDOpt::Coarse),
-      options.count(PDOpt::Flatten));
-  return typename SparsePack<Ts...>::Descriptor(base_desc);
+
+  std::vector<std::string> vars{Ts::name()...};
+  std::vector<bool> use_regex{Ts::regex()...};
+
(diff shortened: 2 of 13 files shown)


src/interface/meshblock_data.hpp
```diff
@@ -85,6 +85,7 @@ class MeshBlockData {
   }
   auto GetParentPointer() const { return GetBlockPointer(); }
   void SetAllowedDt(const Real dt) const { GetBlockPointer()->SetAllowedDt(dt); }
+  Mesh *GetMeshPointer() const { return GetBlockPointer()->pmy_mesh; }
```
Yurlungur, line 88: :+1: pretty annoying not to have this

src/interface/meshblock_data.hpp
```diff
@@ -150,6 +151,8 @@ class MeshBlockData {
     return varUidMap_.at(uid);
   }
 
+  auto &GetUidMap() { return varUidMap_; }
```
Yurlungur: should this be a const reference?

src/interface/sparse_pack.hpp
```diff
@@ -119,124 +120,26 @@ class SparsePack : public SparsePackBase {
...
+    template <class T>
+    SparsePack GetPack(T *pmd) const {
+      return SparsePack(SparsePackBase::GetPack(pmd, *this));
+    }
+
+    SparsePackIdxMap GetMap() const {
+      PARTHENON_REQUIRE(sizeof...(Ts) == 0,
+                        "Should not be getting an IdxMap for a type based pack");
+      return SparsePackBase::GetIdxMap(*this);
+    }
```
Yurlungur, line 141: I think it's fine not to mix type vs string based packs, but I vaguely remember us coming up with a solution while we were talking to mix them unambiguously.

src/interface/sparse_pack.hpp
```diff
@@ -437,8 +340,71 @@ class SparsePack : public SparsePackBase {
     const int vidx = GetLowerBound(b, t) + t.idx;
     return pack_(dir, b, vidx)(k, j, i);
   }
+
+  template <class... VTs>
+  auto GetPtrs(const int b, const TE el, int k, int j, int i, VTs... vts) {
+    return std::make_tuple({&(*this)(b, el, vts, k, j, i)...});
+  }
```
Yurlungur: :+1:

src/interface/sparse_pack.hpp
```diff
@@ -437,8 +340,71 @@ class SparsePack : public SparsePackBase {
...
 };
 
+inline auto MakePackDescriptor(StateDescriptor *psd, const std::vector<std::string> &vars,
```
Yurlungur, line 351: What is the implication of passing in the state descriptor here? Obviously, that's needed to get at the UIDs... but should we warn people somewhere to only use the `resolved_pacakges` state descriptor? Or alternatively provide an overload that takes a pointer to the mesh so we can unambiguously get the right one?

src/interface/sparse_pack.hpp
```diff
@@ -437,8 +340,71 @@ class SparsePack : public SparsePackBase {
...
+}
+
+inline auto MakePackDescriptor(StateDescriptor *psd, const std::vector<std::string> &vars,
+                               const std::vector<MetadataFlag> &flags = {},
+                               const std::set<PDOpt> &options = {}) {
+  return MakePackDescriptor(psd, vars, std::vector<bool>(vars.size(), false), flags,
+                            options);
+}
+
+inline auto MakePackDescriptor(
```
Yurlungur, line 397: Can we also get a UID overload for the pack descriptor?

src/interface/sparse_pack_base.cpp
```diff
@@ -263,12 +323,12 @@ template SparsePackBase &SparsePackCache::BuildAndAdd<MeshBlockData<Real>>(
...
-  identifier += "____";
-  for (int i = 0; i < desc.vars.size(); ++i)
-    identifier += desc.vars[i] + std::to_string(desc.use_regex[i]);
-  identifier += "____";
+  for (const auto &vgroup : desc.var_groups) {
+    for (const auto &[vid, uid] : vgroup) {
+      identifier += std::to_string(uid) + "_";
+    }
+    identifier += "|";
+  }
```
Yurlungur, line 331: It probably matters not at all, since a string is just a vec of chars, but if the identifier is now just a string of UIDs, perhaps there's some compression/performance to be gained here by using a `std::vector<std::size_t>` instead of a s string?

(Actually it could also be the string is more performant, since a char is 1 byte vs 8 bytes for a long int. I guess do we expect UIDs to be 4 digits or longer?)

src/interface/var_id.hpp
```diff
@@ -0,0 +1,56 @@
...
+/// However, sometimes we also need to be able to separate the base name from the sparse
+/// ID. Instead of relying on the fact that they are separated by a "_", we store them
+/// separately in VarID struct. This way we know that a dense variable "foo_3" does not
+/// have a sparse ID and a sparse field "foo_3" has base name "foo" and sparse ID 3,
+/// however, the two VarIDs representing them are still considered equal, so that we find
+/// such duplicates
+/// TODO(JMM): Using VarID machinery for prolongation/restriction
+/// implies that all vars in a sparse pool have the same custom
+/// prolongation/restriction operators.
+struct VarID {
```
Yurlungur, line 36: I like moving this into its own file. :+1:

src/interface/update.cpp
```diff
@@ -149,9 +149,10 @@ TaskStatus SparseDealloc(MeshData<Real> *md) {
   const IndexRange kb = md->GetBoundsK(IndexDomain::entire);
 
   auto control_vars = md->GetMeshPointer()->resolved_packages->GetControlVariables();
-  const auto tup = SparsePack<>::Get(md, control_vars, {Metadata::Sparse});
-  auto pack = std::get<0>(tup);
-  auto packIdx = std::get<1>(tup);
+  auto desc = MakePackDescriptor(md->GetMeshPointer()->resolved_packages.get(),
```
Yurlungur, line 152: Probably for a separate PR, but should we look at using sparse packs in the other parts of the built in machinery, instead of legacy packs? We could in fact store the pack descriptors... either in params for an internal parthenon package, or with tricks like static variables, for a bit of a performance boost.

Yurlungur (approved):
Overall I like this, and I think it's a nice improvement over what we had. A few higher level comments, and then some lower level ones.

- I think it's pretty important we preserve ordering. Are you imagining the downstream code provides the order preserving `MakeStateDescriptor` function? Is there a reason parthenon shouldn't provide that?
- In our big picture discussion, we talked about making it easy to create and register type-based variables. This could be in a future PR, but I wonder if we should expose some utility functions/macros for doing that so a user could declare the types and register the var with less boiler plate?
- I think this machinery gets us very close to the point where we can move completely to sparse packs over legacy packs. If sparse packs can provide (via the descriptor machinery) a more performant interface, even ignoring the per block caching vs per mesh caching, I think we should move to make the switch soon. Pull off the band aid so to speak.

src/interface/meshblock_data.hpp
lroberts36: Yeah, I was surprised it wasn't there.

src/interface/sparse_pack.hpp
lroberts36: Yeah, I think this is actually very easy to relax, it just requires writing a new `MakeDescriptor` that puts together a vector of variable groups from types and a vector of strings.

src/interface/sparse_pack.hpp
lroberts36: My thinking was that a `StateDesciptor` is actually where you want to go for all of the field definitions (since in some cases `MeshBlockData` doesn't have all of the fields present). It may be that you want to create a pack descriptor that only gets fields from a certain package as well. That being said, I agree it would be good to have an overloaded `MakePackDesciptor` that takes `MeshData` or `MeshBlockData` and gets the `resolved_packages` state descriptor.

src/interface/sparse_pack.hpp
lroberts36: You mean like
```c++
inline auto MakePackDescriptor(StateDescriptor *psd, const std::vector<Uid_t> &vars,
                               const std::vector<MetadataFlag> &flags = {},
                               const std::set<PDOpt> &options = {}) {
  
  auto selector = [&](int vidx, const VarID &id, const Metadata &md) {
    if (flags.size() > 0) {
      for (const auto &flag : flags) {
        if (!md.IsSet(flag)) return false;
      }
    }
      if (vars[vidx] == Variable::GetUniqueID(id.label())) return true;
    }
    return false;
  };
  std::string var_str; 
  for (auto v : vars) {var_str.push_back(std::to_string(v));} 
  impl::PackDescriptor base_desc(psd, var_str, selector, options);
  return typename SparsePack<>::Descriptor(base_desc);
}

File: src/interface/sparse_pack_base.cpp (hunk 7 above)
Comment (lroberts36) [reply src/interface/sparse_pack_base.cpp:331]: Yeah, I thought a little bit about how to do that and it seemed like more complication than it was worth (since you can't just flatten the array of variable groups). And, as you point out, it is not clear that it will be faster anyway.

File: src/interface/var_id.hpp (hunk 8 above)
Comment (lroberts36) [reply src/interface/var_id.hpp:36]: Yeah, the hardest part about this PR was fixing circular dependencies.

File: src/interface/sparse_pack.hpp (hunk 6 above)
Comment (Yurlungur) [reply src/interface/sparse_pack.hpp:397]: I think the optional flag vector can be dropped too. It can be assumed they've already been filtered.

File: src/interface/update.cpp (hunk 9 above)
Comment (lroberts36) [reply src/interface/update.cpp:152]: Yeah, I think we should just save that for if/when we remove legacy packs. When I first put in sparse packs, I tried to sneak them into update but it didn't seem like people were on board...

**Comment by lroberts36** [comment]
> * I think it's pretty important we preserve ordering. Are you imagining the downstream code provides the order preserving `MakeStateDescriptor` function? Is there a reason parthenon shouldn't provide that?

I am not quite sure what you mean here. Ordering is preserved in the sense that the first type in the type list corresponds to the first variable group which has indices in the pack [0, n_1] (where n_1 is the total number of 3D fields that correspond to that variable group), the second type corresponds to indices [n_1 + 1, n_1 +  n_2], etc. Within those groups, there is some choice about how to order the fields. Currently I explicitly sort them first lexicographically by their base name and then by their sparse index. I think we just need to make sure this conforms to what happens in legacy packs.

**Comment by lroberts36** [comment]
> * In our big picture discussion, we talked about making it easy to create and register type-based variables. This could be in a future PR, but I wonder if we should expose some utility functions/macros for doing that so a user could declare the types and register the var with less boiler plate?

Yeah, I think the best solution to this is just a macro like the one we use in Riot. Something like 
```c++
#define VARIABLE(ns,nm,sz) \
  struct nm : public parthenon::variable_names::base_t<false, sz> { \
    template <class... Ts> \
    KOKKOS_INLINE_FUNCTION nm(Ts &&...args) \
        : parthenon::variable_names::base_t<false, sz>(std::forward<Ts>(args)...) {} \
    static std::string name() { return "" #ns "." #nm ""; } \
  }; 
```

Yurlungur:
> I am not quite sure what you mean here. 

I essentially mean what is done in legacy packs.

> Ordering is preserved in the sense that the first type in the type list corresponds to the first variable group which has indices in the pack [0, n_1] (where n_1 is the total number of 3D fields that correspond to that variable group), the second type corresponds to indices [n_1 + 1, n_1 + n_2], 

I think this conforms to legacy packs, which are ordered by the order of the string vector passed in, then by sparse index.

> Within those groups, there is some choice about how to order the fields. Currently I explicitly sort them first lexicographically by their base name and then by their sparse index. I think we just need to make sure this conforms to what happens in legacy packs.

I guess the way this might not match legacy packs is if variable types correspond to many variables, as they can with, e.g., regex.

Yurlungur:
> > * In our big picture discussion, we talked about making it easy to create and register type-based variables. This could be in a future PR, but I wonder if we should expose some utility functions/macros for doing that so a user could declare the types and register the var with less boiler plate?
> 
> Yeah, I think the best solution to this is just a macro like the one we use in Riot. Something like
> 
> ```c++
> #define VARIABLE(ns,nm,sz) \
>   struct nm : public parthenon::variable_names::base_t<false, sz> { \
>     template <class... Ts> \
>     KOKKOS_INLINE_FUNCTION nm(Ts &&...args) \
>         : parthenon::variable_names::base_t<false, sz>(std::forward<Ts>(args)...) {} \
>     static std::string name() { return "" #ns "." #nm ""; } \
>   }; 
> ```

Yeah @jdolence and I worked something like this out... @jdolence may have it lying around.

Yurlungur:
> > > * In our big picture discussion, we talked about making it easy to create and register type-based variables. This could be in a future PR, but I wonder if we should expose some utility functions/macros for doing that so a user could declare the types and register the var with less boiler plate?
> > 
> > 
> > Yeah, I think the best solution to this is just a macro like the one we use in Riot. Something like
> > ```c++
> > #define VARIABLE(ns,nm,sz) \
> >   struct nm : public parthenon::variable_names::base_t<false, sz> { \
> >     template <class... Ts> \
> >     KOKKOS_INLINE_FUNCTION nm(Ts &&...args) \
> >         : parthenon::variable_names::base_t<false, sz>(std::forward<Ts>(args)...) {} \
> >     static std::string name() { return "" #ns "." #nm ""; } \
> >   }; 
> > ```
> 
> Yeah @jdolence and I worked something like this out... @jdolence may have it lying around.

Or the thing you just posted... duh. Clearly I've lost my reading comprehension tonight. I also kind of want a type based variable registration infrastructure, so you don't even have to request the string.

lroberts36:
> I guess the way this might not match legacy packs is if variable types correspond to many variables, as they can with, e.g., regex.

I think that it is not possible for legacy packs to have multiple base names in a single variable group, so I guess that just means there is nothing we have to conform to in that case.

Yurlungur:
> > I guess the way this might not match legacy packs is if variable types correspond to many variables, as they can with, e.g., regex.
> 
> I think that it is not possible for legacy packs to have multiple base names in a single variable group, so I guess that just means there is nothing we have to conform to in that case.

I think that's right. But we should test just to be safe.

src/interface/meshblock_data.hpp
lroberts36: yeah, probably so. I was just lazy and wanted to be able to use `[id]`. Fixed now

src/interface/sparse_pack.hpp
lroberts36: I would say we leave this for another PR.

src/interface/sparse_pack.hpp
lroberts36: I think we should maybe think about splitting the `MakePackDescriptor` functions out into their own file so that they are easy to find. Maybe we can do this in another PR and add a few different options for making pack descriptors.

Yurlungur: :+1: that sounds good

src/interface/sparse_pack.hpp
```diff
@@ -437,8 +340,71 @@ class SparsePack : public SparsePackBase {
     const int vidx = GetLowerBound(b, t) + t.idx;
     return pack_(dir, b, vidx)(k, j, i);
   }
+
+  template <class... VTs>
+  auto GetPtrs(const int b, const TE el, int k, int j, int i, VTs... vts) {
+    return std::make_tuple({&(*this)(b, el, vts, k, j, i)...});
```
pdmullen: This functionality needs to be called inside a `par_for_outer` but before a `par_for_inner`.  I worry about calling `std::make_tuple` inside a GPU kernel.  I know that `std::vector`s are not permitted, are we certain that tuples are allowed?

src/interface/sparse_pack.hpp
```diff
@@ -437,8 +340,71 @@ class SparsePack : public SparsePackBase {
...
+  PARTHENON_REQUIRE(vars.size() == use_regex.size(),
+                    "Vargroup names and use_regex need to be the same size.");
+  auto selector = [&](int vidx, const VarID &id, const Metadata &md) {
+    if (flags.size() > 0) {
+      for (const auto &flag : flags) {
+        if (!md.IsSet(flag)) return false;
+      }
+    }
+
+    if (use_regex[vidx]) {
```
pdmullen, line 364: ;-)

src/interface/sparse_pack.hpp
```diff
@@ -437,8 +340,71 @@ class SparsePack : public SparsePackBase {
     const int vidx = GetLowerBound(b, t) + t.idx;
     return pack_(dir, b, vidx)(k, j, i);
   }
+
+  template <class... VTs>
+  auto GetPtrs(const int b, const TE el, int k, int j, int i, VTs... vts) {
```
pdmullen: `KOKKOS_INLINE_FUNCTION`?

src/interface/var_id.hpp
```diff
@@ -0,0 +1,56 @@
...
+#include <string>
+
+namespace parthenon {
+
+inline std::string MakeVarLabel(const std::string &base_name, int sparse_id) {
+  return base_name +
+         (sparse_id == InvalidSparseID ? "" : "_" + std::to_string(sparse_id));
+}
+
+/// We uniquely identify a variable by it's full label, i.e. base name plus sparse ID.
```
pdmullen:
```suggestion
/// We uniquely identify a variable by its full label, i.e. base name plus sparse ID.
```

pdmullen (approved):
LGTM

src/interface/sparse_pack.hpp
lroberts36: After a fix and adding pulling out pointers to the sparse pack test, it seems to at least compile on the darwin nv-devkit nodes. We'll see if it passes all of the tests on GitHub

src/interface/sparse_pack.hpp
lroberts36: done

Issue #630, opened 2022-01-02
Clang/Cuda/Debug build of unit tests not working


pgrete:
Using Clang (clang-12 and clang-13 confirmed) for target device Cuda with build type `Debug` fails with
```
[ 63%] Building CXX object tst/unit/CMakeFiles/unit_tests.dir/kokkos_abstraction.cpp.o
In file included from <built-in>:1:
In file included from /usr/lib/clang/13.0.0/include/__clang_cuda_runtime_wrapper.h:41:
In file included from /usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../include/c++/11.1.0/cmath:41:
In file included from /usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../include/c++/11.1.0/x86_64-pc-linux-gnu/bits/c++config.h:571:
In file included from /usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../include/c++/11.1.0/x86_64-pc-linux-gnu/bits/os_defines.h:39:
/usr/include/features.h:397:4: warning: _FORTIFY_SOURCE requires compiling with optimization (-O) [-W#warnings]
#  warning _FORTIFY_SOURCE requires compiling with optimization (-O)
   ^
1 warning generated when compiling for sm_86.
ptxas fatal   : Cannot take address of function '__cxa_pure_virtual' 
clang-13: error: ptxas command failed with exit code 255 (use -v to see invocation)
clang version 13.0.0
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
clang-13: note: diagnostic msg: 
********************

PLEASE ATTACH THE FOLLOWING FILES TO THE BUG REPORT:
Preprocessed source(s) and associated run script(s) are located at:
clang-13: note: diagnostic msg: /tmp/kokkos_abstraction-5ff6b9.cu
clang-13: note: diagnostic msg: /tmp/kokkos_abstraction-f6f273.cu
clang-13: note: diagnostic msg: /tmp/kokkos_abstraction-5ff6b9.sh
clang-13: note: diagnostic msg: 

********************
make[2]: *** [tst/unit/CMakeFiles/unit_tests.dir/build.make:174: tst/unit/CMakeFiles/unit_tests.dir/kokkos_abstraction.cpp.o] Error 255
[...]

PR #887, opened 2023-06-12
param restarts

### LINKED ISSUE #683: Store arbitrary/mesh data in output
Do we have a way to store (and restore) arbitrary data (e.g., not tied to a block) in our output files apart from abusing the "input file"?
Specifically, I need to save a (small) array of complex numbers that is identical across all ranks and blocks.

Yurlungur:
## PR Summary

We've discussed previously that we now use `Params` for many purposes, including storing mesh-level data. In this PR, I implement scaffolding for this use case. In particular, I add:
- The ability to output `Kokkos::View` and `ParArray*D` params to disk
- The ability to read a mutable param from a restart file to set the value. In particular, a param can now be marked `Immutable`, `Mutable`, or `Restart`. Here `Mutable` is as it was before (though now this is an enum instead of a bool.) `Restart` is a superset of `Mutable`. A `Restart` param is mutable, but also re-read from the restart file upon restart. It is first set to whatever the package initialization provides, then overwritten.
- Unit tests for the new Param machinery.

This new machinery enables some important use-cases. For Phoebus, it lets us stash spacetime information in `Params`, which is useful in some cases. I imagine that for things like turbulent forcing, or derived integral quantities, the Athena-PK team may also be interested.

## Limitations, technical questions

- I needed to enumerate the types that can be written to/read from file. So it's a limited list.
- Because now more params are output, some regression tests are now failing, as the gold file contains less information. I can modify the tests to be less strict (I believe in large part the unit tests now cover the code paths they were checking), or I can create a new gold file. I don't know which is better. I'd like input. **Please let me know which option you would prefer.**
- Outputting `Params` or reading `Params` that differ on different MPI ranks is undefined behavior. This was already true, but I now document it. I think permitting different `Params` on different ranks would be a pretty big lift, and I recommend against. At least for now.
- On a related note, since params are treated as attributes in the HDF5 file, they must be written and read collectively and identically.

@bprather I know I've talked to you about this design. Please take a look at this and let me know if it meets your constraints. If I recall, there was some use-case you had, where you did *not* want the infrastructure to modify your mutable params, which is why I added the three mutability levels.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/doc/sphinx/src/inputs.rst
+++ b/doc/sphinx/src/inputs.rst
@@ -31,16 +31,16 @@ General parthenon options such as problem name and parameter handling.
 
 Options related to time-stepping and printing of diagnostic data.
 
-+------------------------------+---------+--------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
-| Option                       | Default | Type   | Description                                                                                                                                                                                                                                 |
-+==============================+=========+========+=============================================================================================================================================================================================================================================+
-|| tlim                        || none   || float || Stop criterion on simulation time.                                                                                                                                                                                                         |
-|| nlim                        || -1     || int   || Stop criterion on total number of steps taken. Ignored if < 0.                                                                                                                                                                             |
-|| perf_cycle_offset           || 0      || int   || Skip the first N cycles when calculating the final performance (e.g., zone-cycles/wall_second). Allows to hide the initialization overhead in Parthenon, which usually takes place in the first cycles when Containers are allocated, etc. |
-|| ncycle_out                  || 1      || int   || Number of cycles between short diagnostic output to standard out containing, e.g., current time, dt, zone-update/wsec. Default: 1 (i.e, every cycle).                                                                                      |
-|| ncycle_out_mesh             || 0      || int   || Number of cycles between printing the mesh structure (e.g., total number of MeshBlocks and number of MeshBlocks per level) to standard out. Use a negative number to also print every time the mesh was modified. Default: 0 (i.e, off).   |
-|| ncrecv_bdry_buf_timeout_sec || -1.0   || Real  || Timeout in seconds for the `ReceiveBoundaryBuffers` tasks. Disabed (negative) by default. Typically no need in production runs. Useful for debugging MPI calls.                                                       |
-+------------------------------+---------+--------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
++------------------------------+---------+--------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+| Option                       | Default | Type   | Description                                                                                                                                                           |
++==============================+=========+========+=======================================================================================================================================================================+
+|| tlim                        || none   || float || Stop criterion on simulation time.                                                                                                                                   |
+|| nlim                        || -1     || int   || Stop criterion on total number of steps taken. Ignored if < 0.                                                                                                       |
+|| perf_cycle_offset           || 0      || int   || Skip the first N cycles when calculating the final performance (e.g., zone-cycles/wall_second). Allows to hide the initialization overhead in Parthenon.             |
+|| ncycle_out                  || 1      || int   || Number of cycles between short diagnostic output to standard out containing, e.g., current time, dt, zone-update/wsec. Default: 1 (i.e, every cycle).                |
+|| ncycle_out_mesh             || 0      || int   || Number of cycles between printing the mesh structure to standard out. Use a negative number to also print every time the mesh was modified. Default: 0 (i.e, off).   |
+|| ncrecv_bdry_buf_timeout_sec || -1.0   || Real  || Timeout in seconds for the `ReceiveBoundaryBuffers` tasks. Disabed (negative) by default. Typically no need in production runs. Useful for debugging MPI calls.      |
++------------------------------+---------+--------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------+
 
 
 ``<parthenon/mesh>``
--- a/doc/sphinx/src/interface/state.rst
+++ b/doc/sphinx/src/interface/state.rst
@@ -38,10 +38,22 @@ several useful features and functions.
   all blocks just like dense variables, however, in a future upgrade, they
   will only be allocated on those blocks where the user explicitly
   allocates them or non-zero values are advected into.
-- ``void AddParam<T>(const std::string& key, T& value, bool is_mutable)``
-  adds a parameter (e.g., a timestep control coefficient, refinement
-  tolerance, etc.) with name ``key`` and value ``value``. If
-  ``is_mutable`` is true, parameters can be more easily modified.
+- ``void AddParam<T>(const std::string& key, T& value, Mutability
+  mutability)`` adds a parameter (e.g., a timestep control
+  coefficient, refinement tolerance, etc.) with name ``key`` and value
+  ``value``. The enum ``mutability`` can take on three values:
+  ``Mutability::Immutable``, ``Mutability::Mutable``, and
+  ``Mutability::Restart``. Paramters that are ``Immutable`` cannot be
+  modified. Parameters that are ``Mutable`` or ``Restart`` can be
+  modified via the ``MutableParam`` and ``UpdateParam``
+  options. Parameters that are ``Restart`` will be re-read from the
+  restart file and updated upon restart. In contrast, ``Mutable``
+  params not marked ``Restart`` are updated only by user code, not
+  automatically. Note that not all parameter types can be output to
+  HDF5 file. However, most common scalar, vector, and ``Kokkos`` view
+  types are supported. Note also that by default the ``Mesh`` object
+  and each ``MeshBlock`` object all own their own copies of all
+  packages.
 - ``void UpdateParam<T>(const std::string& key, T& value)``\ updates a
   parameter (e.g., a timestep control coefficient, refinement tolerance,
   etc.) with name ``key`` and value ``value``. A parameter of the same
--- a/doc/sphinx/src/interface/state.rst
+++ b/doc/sphinx/src/interface/state.rst
@@ -38,8 +38,8 @@ several useful features and functions.
   all blocks just like dense variables, however, in a future upgrade, they
   will only be allocated on those blocks where the user explicitly
   allocates them or non-zero values are advected into.
-- ``void AddParam<T>(const std::string& key, T& value, Mutability
-  mutability)`` adds a parameter (e.g., a timestep control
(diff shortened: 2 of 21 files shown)


Yurlungur:
Issue #683 discusses this path vs an alternative path adding variables at the `Mesh` level. I think the `Mesh` level infrastructure has advantages for sure. But `Params` currently work, and this has a relatively clean implementation.

src/outputs/parthenon_hdf5.hpp
```diff
@@ -202,49 +210,58 @@ template <>
...
 
-template <typename T>
-std::vector<T> HDF5ReadAttributeVec(hid_t location, const std::string &name) {
-  std::vector<T> res;
-  auto type = getHDF5Type(res.data());
+template <typename D, typename S>
+void HDF5WriteAttribute(const std::string &name, const ParArrayGeneric<D, S> &view,
+                        hid_t location) {
+  return HDF5WriteAttribute(name, view.KokkosView(), location);
+}
```
lroberts36, line 248: With a couple of small changes, we could probably make `ParArrayGeneric` conform to the `kokkos_view` concept and get rid of this overload. Probably not worth it though.

src/outputs/parthenon_hdf5.hpp
```diff
@@ -255,6 +272,72 @@ std::vector<T> HDF5ReadAttributeVec(hid_t location, const std::string &name) {
...
+  }
+}
+
+template <typename D, typename S>
+void HDF5ReadAttribute(hid_t location, const std::string &name,
+                       ParArrayGeneric<D, S> &pararray) {
+  // forces compiler to pass by non-const reference into the next function
+  D &view = pararray.KokkosView();
+  HDF5ReadAttribute(location, name, view);
+}
```
lroberts36: Obviously this means that anything stored in the `ParArray` state is lost when restarting. I think this is fine (since the state is more to easily transfer properties of fields between host and device), but we may want to add a comment somewhere to that effect.

lroberts36 (approved):
If this is passing tests, it LGTM. That being said, I am a little queasy about the general philosophy of putting more and more stuff into params, especially things that may be better thought of as simulation state (e.g. quantities that are evolved but are not directly tied to the grid). I haven't thought carefully about how it would work, but it seems like it would make more logical sense to have a way of registering fields at the mesh level of arbitrary shape and being able to write those as hdf5 datasets (instead of as metadata). It seems worth it to have a discussion at the Parthenon meeting about this.

Edit: Hadn't carefully read #683 before writing this. I think I generally agree with @pgrete suggestion there.

Yurlungur:
> If this is passing tests

It is not currently passing tests. Both for the regression test reason mentioned above, and for some other reason on the cuda machines. I need to debug. The fully general functionality for writing/reading arbitrary shapes was tricky to get right.

> I am a little queasy about the general philosophy of putting more and more stuff into params, especially things that may be better thought of as simulation state (e.g. quantities that are evolved but are not directly tied to the grid). I haven't thought carefully about how it would work, but it seems like it would make more logical sense to have a way of registering fields at the mesh level of arbitrary shape and being able to write those as hdf5 datasets (instead of as metadata). It seems worth it to have a discussion at the Parthenon meeting about this.

This is fair. Yes I'd like to discuss at the parthenon meeting. I think my solution is significantly more flexible than @pgrete 's suggestion, as it's type aware, capable of looking at shape, and knows how to handle both host- and device-side data, but this may or may not be a good thing. 

I do think my solution could be relatively straightforwardly translated into a `Mesh` level data solution... just by sticking a `Params`-like object at the mesh level.

doc/sphinx/src/interface/state.rst
```diff
@@ -38,10 +38,25 @@ several useful features and functions.
...
+  restart file and updated upon restart. In contrast, ``Mutable``
+  params not marked ``Restart`` are updated only by user code, not
+  automatically. Note that not all parameter types can be output to
+  HDF5 file. However, most common scalar, vector, and ``Kokkos`` view
+  types are supported. Note also that if the value of a ``Param`` is
+  different on different MPI ranks, this will result in undefined
+  behaviour.
+- ``void AddParam<T>(const std::string& key, T& value, bool is_mutable=false)``
+  is the same as above, but adds only ``Immutable`` or ``Mutable`` params,
+  not ``Restart`` params.
```
forrestglines, line 59: It this behavior of Mutable parameters at restarts a departure from the old functionality? The cluster simulations I'm running use AddParams to store a few mutable quantities between files that probably should be carried between restarts. I should probably go fix it

src/outputs/parthenon_hdf5.hpp
```diff
@@ -202,49 +210,58 @@ template <>
...
+    pdata = view_h.data();
+  }
+  auto type = getHDF5Type(pdata);
+  H5A const attribute = H5A::FromHIDCheck(
+      H5Acreate(location, name.c_str(), type, data_space, H5P_DEFAULT, H5P_DEFAULT));
+  PARTHENON_HDF5_CHECK(H5Awrite(attribute, type, pdata));
+}
+
+template <typename T, REQUIRES(implements<scalar(T)>::value)>
+void HDF5WriteAttribute(const std::string &name, const T &value, hid_t location) {
```
forrestglines, line 238: So it this what might enable writing arbitrary types to as attributes? I'm not sure how to understand `REQUIRES(implements<scalar(T)>::value)`

doc/sphinx/src/interface/state.rst
forrestglines: I think the new `AddParams` function got left out of this PR, I don't any changes to `state_descriptor.hpp`

forrestglines (changes requested):
I think the `AddParams` from the documentation needs to be added to `StateDescriptor`

forrestglines:
Restarts with AthenaPK seem to work fine, AthenaPK tests are passing so far

doc/sphinx/src/interface/state.rst
Yurlungur: Oh you're right. I only added them to `Params` not to state descriptor. I'll fix that.

src/outputs/parthenon_hdf5.hpp
Yurlungur: That's using @lroberts36 concepts. It means it knows how to write things like a single double or an integer, but it doesn't do arbitrary types. You'll see I also have specializations for `std::vector`, `ParArray` and `Kokkos::View`. To do other types, one would need to add the appropriate template specialization.

Yurlungur:
Big picture are you happy with the design @forrestglines ?

forrestglines:
> Big picture are you happy with the design @forrestglines ?

I think it's good, I see it as an extension of the existing design and not any significant departure.

If it's possible to add our own custom types to Param restarts in the future then it might make some of the usage of Params in my cluster pgen cleaner.

Yurlungur:
> If it's possible to add our own custom types to Param restarts in the future then it might make some of the usage of Params in my cluster pgen cleaner.

I think this is possible, but I'd rather leave it to a future PR. I think it requires an additional interface to register custom I/O ops

doc/sphinx/src/interface/state.rst
Yurlungur: Fixed.

Yurlungur:
@forrestglines how do things look now? I fixed the missing function in `StateDescriptor`. I have now updated the gold files, so regression tests should pass. 

Also, because I always forget how to add regression tests, I added a doc page for it. And I fixed some errors I got when building the docs locally.

doc/sphinx/src/inputs.rst
```diff
@@ -31,16 +31,16 @@ General parthenon options such as problem name and parameter handling.
...
-|| ncycle_out_mesh             || 0      || int   || Number of cycles between printing the mesh structure (e.g., total number of MeshBlocks and number of MeshBlocks per level) to standard out. Use a negative number to also print every time the mesh was modified. Default: 0 (i.e, off).   |
-|| ncrecv_bdry_buf_timeout_sec || -1.0   || Real  || Timeout in seconds for the `ReceiveBoundaryBuffers` tasks. Disabed (negative) by default. Typically no need in production runs. Useful for debugging MPI calls.                                                       |
-+------------------------------+---------+--------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
++------------------------------+---------+--------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```
Yurlungur, line 34: Sphinx didn't seem to like this ultrawide table. I had to shrink it to get it to render.

Yurlungur:
> That being said, I am a little queasy about the general philosophy of putting more and more stuff into params, especially things that may be better thought of as simulation state (e.g. quantities that are evolved but are not directly tied to the grid). I haven't thought carefully about how it would work, but it seems like it would make more logical sense to have a way of registering fields at the mesh level of arbitrary shape and being able to write those as hdf5 datasets (instead of as metadata). It seems worth it to have a discussion at the Parthenon meeting about this.

Just to make sure comment is here for posterity: We discussed in the last parthenon call and decided to go with this design ans there are open questions about a variable based approach, and it was decided an in-hand solution was desired. That said the whole team agreed with the concern about putting more stuff into params.

Yurlungur:
Clang tests currently failing due to #630 .

Yurlungur:
Discussed with @bprather and currently he's happy. Calling it good assuming tests pass after the fix and @forrestglines approves.

c-prather (approved):
LGTM.  Mostly I'm happy that this is opt-in (at least, the reading -- iiuc it will make restart files bigger regardless, which I think is a small enough price to pay).

I don't think I carry around the kind of global state that would benefit from this, but it may prove useful to me at some point.

forrestglines:
The merge seems to have broken compilation for my build of Parthenon within Athenapk:
```
/vast/home/forrestglines/code/athenapk-project/athenapk/external/parthenon/src/interface/variable.cpp(217): error: identifier "mem_size" is undefined

1 error detected in the compilation of "/vast/home/forrestglines/code/athenapk-project/athenapk/external/parthenon/src/interface/variable.cpp".
make[2]: *** [parthenon/src/CMakeFiles/parthenon.dir/build.make:412: parthenon/src/CMakeFiles/parthenon.dir/interface/variable.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
```

EDIT: It's a simple fix, only happens when building without sparse

forrestglines (approved):
@jonahm-LANL I pushed a tiny fix, I think it's ready to merge

Yurlungur:
Oops good catch. Thanks for the fix @forrestglines !

Yurlungur:
Forcing merge since tests pass and the cuda/clang tests are currently broken. Thanks @pgrete for confirming.

PR #890, opened 2023-06-15
Fix Bugs in Sparse Communication

lroberts36:
## PR Summary

This fixes a couple of bugs in sparse communication:

1. Because the field `BndInfo::allocated` was being repurposed to refer to the block being allocated for send `BndInfo` and to refer to the buffer being allocated for receive `BndInfo`, in some cases boundary prolongation would not occur when it should for sparse variables. Through a chain of events, this could show up as a dependence of results on the number of ranks that sparse problems were run on (although the differences were of order the sparse threshold). 
2. In thinking through this, I also realized that for sparse fields splitting the boundary communication into local and nonlocal is not safe, since a field can be allocated by one set of boundary communication after the other boundary communication has finished. This may not have been a real issue to date, as we have always assumed the default value to fill sparse fields with is zero, but if the default value was non-zero it could (maybe?) cause an issue.
3. The `dealloc_count` of fields was not passed during different rank, same-to-same communication during AMR. This meant that the step at which a particular block was deallocated could vary with the number of ranks.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/src/bvals/comms/bnd_info.cpp
+++ b/src/bvals/comms/bnd_info.cpp
@@ -217,13 +217,14 @@ BndInfo BndInfo::GetSetBndInfo(std::shared_ptr<MeshBlock> pmb, const NeighborBlo
   out.buf = buf->buffer();
   auto buf_state = buf->GetState();
   if (buf_state == BufferState::received) {
-    out.allocated = true;
+    out.buf_allocated = true;
     PARTHENON_DEBUG_REQUIRE(v->IsAllocated(), "Variable must be allocated to receive");
   } else if (buf_state == BufferState::received_null) {
-    out.allocated = false;
+    out.buf_allocated = false;
   } else {
     PARTHENON_FAIL("Buffer should be in a received state.");
   }
+  out.allocated = v->IsAllocated();
 
   int Nv = v->GetDim(4);
   int Nu = v->GetDim(5);
@@ -273,6 +274,8 @@ BndInfo BndInfo::GetSetBndInfo(std::shared_ptr<MeshBlock> pmb, const NeighborBlo
           nb.ni, pmb->loc, el, {Nt, Nu, Nv}, pmb->c_cellbounds);
       out.var = v->coarse_s.Get();
       out.refinement_op = RefinementOp_t::Prolongation;
+      out.prores_idxer[idx] = CalcSetIndices<InterfaceType::CoarseToFine, true>(
+          nb.ni, pmb->loc, el, {Nt, Nu, Nv}, pmb->c_cellbounds);
     } else {
       out.var = v->data.Get();
       out.idxer[idx] = CalcSetIndices<InterfaceType::FineToCoarse>(
@@ -303,17 +306,6 @@ BndInfo BndInfo::GetSetBndInfo(std::shared_ptr<MeshBlock> pmb, const NeighborBlo
     }
   }
 
-  if (buf_state == BufferState::received) {
-    // With control variables, we can end up in a state where a
-    // variable that is not receiving null data is unallocated.
-    // for allocated to be set, the buffer must be sending non-null
-    // data and the receiving variable must be allocated
-    out.allocated = v->IsAllocated();
-  } else if (buf_state == BufferState::received_null) {
-    out.allocated = false;
-  } else {
-    PARTHENON_FAIL("Buffer should be in a received state.");
-  }
   return out;
 }
 
--- a/src/bvals/comms/boundary_communication.cpp
+++ b/src/bvals/comms/boundary_communication.cpp
@@ -290,28 +290,13 @@ ProlongateBounds<BoundaryType::local>(std::shared_ptr<MeshData<Real>> &);
 template TaskStatus
 ProlongateBounds<BoundaryType::nonlocal>(std::shared_ptr<MeshData<Real>> &);
 
-TaskStatus ApplyCoarseBoundaryConditions(std::shared_ptr<MeshData<Real>> &md) {
-  if (!md->GetMeshPointer()->multilevel) return TaskStatus::complete;
-  TaskStatus stat = TaskStatus::complete;
-  for (int block = 0; block < md->NumBlocks(); ++block) {
-    auto bstat = ApplyBoundaryConditionsOnCoarseOrFine(md->GetBlockData(block), true);
-    // if (bstat != TaskStatus::complete) stat = bstat;
-  }
-  return stat;
-}
-
-TaskStatus ApplyFineBoundaryConditions(std::shared_ptr<MeshData<Real>> &md) {
-  if (!md->GetMeshPointer()->multilevel) return TaskStatus::complete;
-  for (int block = 0; block < md->NumBlocks(); ++block) {
-    ApplyBoundaryConditionsOnCoarseOrFine(md->GetBlockData(block), false);
-  }
-  // ApplyBoundaryConditions is guaranteed to return complete, so this is safe
-  return TaskStatus::complete;
-}
-
 // Adds all relevant boundary communication to a single task list
 TaskID AddBoundaryExchangeTasks(TaskID dependency, TaskList &tl,
                                 std::shared_ptr<MeshData<Real>> &md, bool multilevel) {
+  // TODO(LFR): Splitting up the boundary tasks while doing prolongation can cause some
+  //            possible issues for sparse fields. In particular, the order in which
+  //            fields are allocated and then set could potentially result in different
+  //            results if the default sparse value is non-zero.
   const auto any = BoundaryType::any;
   // const auto local = BoundaryType::local;
   // const auto nonlocal = BoundaryType::nonlocal;
@@ -335,9 +320,12 @@ TaskID AddBoundaryExchangeTasks(TaskID dependency, TaskList &tl,
   auto send = tl.AddTask(dependency, SendBoundBufs<any>, md);
   auto recv = tl.AddTask(dependency, ReceiveBoundBufs<any>, md);
   auto set = tl.AddTask(recv, SetBounds<any>, md);
-  auto cbound = tl.AddTask(set, ApplyCoarseBoundaryConditions, md);
-  auto pro = tl.AddTask(cbound, ProlongateBounds<any>, md);
-  auto fbound = tl.AddTask(pro, ApplyFineBoundaryConditions, md);
+  auto pro = set;
+  if (md->GetMeshPointer()->multilevel) {
+    auto cbound = tl.AddTask(set, ApplyBoundaryConditionsOnCoarseOrFineMD, md, true);
+    pro = tl.AddTask(cbound, ProlongateBounds<any>, md);
+  }
+  auto fbound = tl.AddTask(pro, ApplyBoundaryConditionsOnCoarseOrFineMD, md, false);
 
   return fbound;
 }
--- a/src/bvals/comms/boundary_communication.cpp
+++ b/src/bvals/comms/boundary_communication.cpp
@@ -224,14 +224,14 @@ TaskStatus SetBounds(std::shared_ptr<MeshData<Real>> &md) {
         int idx_offset = 0;
         for (int iel = 0; iel < bnd_info(b).ntopological_elements; ++iel) {
           auto &idxer = bnd_info(b).idxer[iel];
-          if (bnd_info(b).allocated) {
+          if (bnd_info(b).buf_allocated && bnd_info(b).allocated) {
             Kokkos::parallel_for(Kokkos::TeamThreadRange<>(team_member, idxer.size()),
                                  [&](const int idx) {
                                    const auto [t, u, v, k, j, i] = idxer(idx);
                                    bnd_info(b).var(iel, t, u, v, k, j, i) =
                                        bnd_info(b).buf(idx + idx_offset);
                                  });
-          } else if (bnd_info(b).var.size() > 0) {
+          } else if (bnd_info(b).allocated) {
             const Real default_val = bnd_info(b).var.sparse_default_val;
             Kokkos::parallel_for(Kokkos::TeamThreadRange<>(team_member, idxer.size()),
                                  [&](const int idx) {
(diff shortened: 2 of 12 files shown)


src/bvals/comms/bnd_info.cpp
```diff
@@ -303,17 +306,6 @@ BndInfo BndInfo::GetSetBndInfo(std::shared_ptr<MeshBlock> pmb, const NeighborBlo
     }
   }
 
-  if (buf_state == BufferState::received) {
-    // With control variables, we can end up in a state where a
-    // variable that is not receiving null data is unallocated.
-    // for allocated to be set, the buffer must be sending non-null
-    // data and the receiving variable must be allocated
```
Yurlungur, line 310: Is this comment still relevant?

src/bvals/comms/boundary_communication.cpp
```diff
@@ -252,16 +252,6 @@ TaskStatus SetBounds(std::shared_ptr<MeshData<Real>> &md) {
...
-
-    // Apply coarse boundary conditions
-    for (int block = 0; block < md->NumBlocks(); ++block) {
-      ApplyBoundaryConditionsOnCoarseOrFine(md->GetBlockData(block), true);
-    }
-
-    // Prolongate from coarse buffer
-    refinement::Prolongate(resolved_packages, cache, pmb->cellbounds, pmb->c_cellbounds);
-    refinement::ProlongateInternal(resolved_packages, cache, pmb->cellbounds,
-                                   pmb->c_cellbounds);
```
Yurlungur, line 264: I guess these need to be split out into tasks again because of the issue with local vs nonlocal send/receive?

src/mesh/mesh.cpp
```diff
@@ -1086,24 +1086,53 @@ void Mesh::Initialize(bool init_problem, ParameterInput *pin, ApplicationInput *
     }
 
     // send FillGhost variables
+    bool can_delete;
+    do {
```
Yurlungur, line 1090: Should these do-while loops have some upper iteration count in case something goes really wrong?

src/mesh/mesh.cpp
```diff
@@ -1115,18 +1144,23 @@ void Mesh::Initialize(bool init_problem, ParameterInput *pin, ApplicationInput *
...
+      }
+      ApplyFineBoundaryConditions(md);
       // Call MeshData based FillDerived functions
       Update::FillDerived(md.get());
     }
 
+    for (int i = 0; i < nmb; ++i) {
+      auto &mbd = block_list[i]->meshblock_data.Get();
+      // Call MeshBlockData based FillDerived functions
+      Update::FillDerived(mbd.get());
```
Yurlungur, line 1175: Potential merge conflict with my precomm fill derived PR. Just need to make sure git doesn't put the precomm stuff below the boundaries/comms

Yurlungur (approved):
Sounds like this was a tricky one to track down. Thanks for chasing through it!

src/mesh/mesh.cpp
lroberts36: Yeah, probably

src/bvals/comms/bnd_info.cpp
lroberts36: It is still relevant but I think this is not the right place to have it.

src/bvals/comms/boundary_communication.cpp
lroberts36: Yeah, you can't prolongate until you have received all data. As it was before, I think it would be safe with local and nonlocal, but you would prolongate twice and only the second time through would be correct. I think it is just easier to think through and less error prone to keep these separate.

src/bvals/comms/bnd_info.cpp
Yurlungur: just noticed it's not added anywhere else. If relevant maybe stick it somewhere else in the source?

src/bvals/comms/boundary_communication.cpp
Yurlungur: :+1:

src/bvals/comms/bnd_info.cpp
lroberts36: Thinking about this a little more, this is naturally taken care of now that I disambiguated the `BndInfo::allocated`. The comment probably doesn't need to be anywhere.

src/mesh/mesh.cpp
lroberts36: It looks like it merged to the correct place

src/bvals/comms/bnd_info.cpp
Yurlungur: cool :+1:

src/bvals/comms/boundary_communication.cpp
```diff
@@ -310,38 +300,28 @@ TaskStatus ApplyCoarseBoundaryConditions(std::shared_ptr<MeshData<Real>> &md) {
   return stat;
 }
 
+TaskStatus ApplyFineBoundaryConditions(std::shared_ptr<MeshData<Real>> &md) {
+  if (!md->GetMeshPointer()->multilevel) return TaskStatus::complete;
+  TaskStatus stat = TaskStatus::complete;
+  for (int block = 0; block < md->NumBlocks(); ++block) {
+    auto bstat = ApplyBoundaryConditionsOnCoarseOrFine(md->GetBlockData(block), false);
+    // if (bstat != TaskStatus::complete) stat = bstat;
```
pdmullen: Why is this commented out?

src/mesh/mesh.cpp
```diff
@@ -1096,24 +1096,58 @@ void Mesh::Initialize(bool init_problem, ParameterInput *pin, ApplicationInput *
...
+        auto &md = mesh_data.GetOrAdd("base", i);
+        if (!sent[i]) {
+          if (SendBoundaryBuffers(md) != TaskStatus::complete) {
+            all_sent = false;
+          } else {
+            sent[i] = true;
+          }
+        }
+      }
+    } while (!all_sent);
```
pdmullen: @jonahm-LANL's comments above regarding do-while termination here as well... can we just turn these into `for` loops with `break` statements?

src/mesh/mesh.cpp
```diff
@@ -1096,24 +1096,58 @@ void Mesh::Initialize(bool init_problem, ParameterInput *pin, ApplicationInput *
...
-          all_received = false;
+        if (!received[i]) {
+          if (ReceiveBoundaryBuffers(md) != TaskStatus::complete) {
+            all_received = false;
+          } else {
+            received[i] = true;
+          }
         }
       }
     } while (!all_received);
```
pdmullen: as above...

pdmullen (commented):
LGTM

src/bvals/comms/boundary_communication.cpp
lroberts36: Oh, I should probably just remove that. `ApplyBoundaryConditionsOnCoarseOrFine` always returns `TaskStatus::complete`.

src/mesh/mesh.cpp
lroberts36: A possibly better solution to this is to create and execute a `TaskList` here, which can be generated with a single call to `AddBoundaryExchangeTasks`. I think I might save this for another PR though.

pdmullen: :+1:

src/bvals/comms/boundary_communication.cpp
lroberts36: done

pgrete:
@lroberts36 I was about to review but got confused by the merge conflict with `develop`.
Would you mind updating (as I'm not sure what to make of the removed commented code that we talked about in #896 with respect to the split local/nonlocal comm).

lroberts36:
@pgrete: I think I just removed those comments without thinking about it. I just added the comments back. I think this is good to review now, although I am not sure what is going on with the CI. It looks like there are some things failing, but I am not sure if that is due to the issue you mentioned.

src/bvals/comms/boundary_communication.cpp
```diff
@@ -333,15 +332,13 @@ TaskID AddBoundaryExchangeTasks(TaskID dependency, TaskList &tl,
 
   // auto out = (pro_local | pro);
 
-  // TODO(LFR): Splitting up the boundary tasks while doing prolongation in one
-  //            breaks things, which I haven't completely figured out yet. To
-  //            move to the commented out code above, this needs to be fixed and
-  //            the prolongation and physical boundary conditions need to be removed
-  //            from the end of SetBounds
```
pgrete, line 340: This comment is still valid given that the commented code above is also still there, isn't it?

src/bvals/comms/boundary_communication.cpp
```diff
@@ -310,6 +300,15 @@ TaskStatus ApplyCoarseBoundaryConditions(std::shared_ptr<MeshData<Real>> &md) {
...
 }
 
+TaskStatus ApplyFineBoundaryConditions(std::shared_ptr<MeshData<Real>> &md) {
+  if (!md->GetMeshPointer()->multilevel) return TaskStatus::complete;
+  for (int block = 0; block < md->NumBlocks(); ++block) {
+    ApplyBoundaryConditionsOnCoarseOrFine(md->GetBlockData(block), false);
+  }
+  // ApplyBoundaryConditions is guaranteed to return complete, so this is safe
+  return TaskStatus::complete;
+}
```
pgrete: Why this function (and the one above) given that `ApplyBoundaryConditionsOnCoarseOrFineMD` exists?

src/mesh/mesh.cpp
```diff
@@ -1084,24 +1084,58 @@ void Mesh::Initialize(bool init_problem, ParameterInput *pin, ApplicationInput *
...
+    do {
+      can_delete = true;
+      for (auto &[k, comm] : boundary_comm_map) {
+        can_delete = comm.IsSafeToDelete() && can_delete;
+      }
+      test_iters++;
+    } while (!can_delete && test_iters < 1e10);
+    if (test_iters >= 1e10)
+      PARTHENON_FAIL(
+          "Too many iterations waiting to delete boundary communication buffers.");
```
pgrete: `PARTHENON_REQUIRE` instead of `if ... PARTHENON_FAIL`?

src/mesh/mesh.cpp
pgrete: Until this other PR is there, could we add counters?
I'm mostly concerned about burning computer time by sth going wrong without other processes knowing about it (and triggering a fail).

src/mesh/mesh.cpp
```diff
@@ -1113,18 +1147,23 @@ void Mesh::Initialize(bool init_problem, ParameterInput *pin, ApplicationInput *
...
-      // Call MeshBlockData based FillDerived functions
-      Update::FillDerived(mbd.get());
-    }
     for (int i = 0; i < num_partitions; i++) {
       auto &md = mesh_data.GetOrAdd("base", i);
+      if (multilevel) {
+        ApplyCoarseBoundaryConditions(md);
+        ProlongateBoundaries(md);
+      }
+      ApplyFineBoundaryConditions(md);
```
pgrete: From `ApplyFineBoundaryConditions` above:
```c++
if (!md->GetMeshPointer()->multilevel) return TaskStatus::complete;
```

Does this logic check out? 
IIRC we also want the fine boundary conditions to be always set because those are the ones next to the active zones, isn't it?

src/outputs/parthenon_hdf5.cpp
```diff
@@ -539,6 +539,8 @@ void PHDF5Output::WriteOutputFileImpl(Mesh *pm, ParameterInput *pin, SimTime *tm
                 vinfo.nx6 * vinfo.nx5 * vinfo.nx4 * vinfo.nx3 * vinfo.nx2 * vinfo.nx1;
           }
           memset(tmpData.data() + index, 0, varSize * sizeof(OutT));
+          for (int i = 0; i < varSize; ++i)
+            (tmpData.data() + index)[i] = std::numeric_limits<Real>::quiet_NaN();
```
pgrete: If I'm reading this right, then you're overriding all the `0` from the `memset` right above, correct?
If yes, is the `memset` still required?

pgrete (commented):
Couple of minor comments and a "big" question on the fine boundaries <-> multilevel logic.

src/bvals/comms/boundary_communication.cpp
lroberts36: Right, I will add it back

src/bvals/comms/boundary_communication.cpp
lroberts36: I was not aware of `ApplyBoundaryConditionsOnCoarseOrFineMD` I think. I will switch to this.

src/mesh/mesh.cpp
lroberts36: Good catch, that was a left over from copy/paste of the coarse boundary conditions. I fixed this on another branch (after debugging for an embarrassingly long time), but I guess the fix never made it over here.

src/outputs/parthenon_hdf5.cpp
lroberts36: Oh, this is debugging code. I will switch back to zeros for sparse. I generally think it is more useful to plot sparse with NaNs so you can see where the sparse fields have been deallocated easily, but I think everyone agreed on filling with zeros at some point.

src/mesh/mesh.cpp
lroberts36: This is fixed by switching to `ApplyBoundaryConditionsOnCoarseOrFineMD`.

src/mesh/mesh.cpp
lroberts36: changed.

src/bvals/comms/boundary_communication.cpp
lroberts36: It looks like maybe I added `ApplyBoundaryConditionsOnCoarseOrFineMD`? Apparently my memory is completely gone

src/mesh/mesh.cpp
lroberts36: I added some simple counters, although I am not sure what a good choice for the maximum number of times through the loop is.

lroberts36:
@pgrete: I think this is ready for a re-review and approval. You are correct about the logic being wrong for fine boundaries and multi-level, that was a copy and paste bug. This is now fixed.

pgrete (approved):
Thanks for the updates.

I just did a quick final fix (promoted the iteration counters to `int64_t` as the threshold of 1e10 was beyond the standard `int` limits) and am pulling the trigger now.

pgrete:
Looks like regression tests are failing, e.g.,
```
 ____Comparing on a per variable basis with tolerance 1e-12
Tolerance = 1e-12
Diff in sparse_1            
    bkji: (  71,   0,   0,   0)
    zyx: (0.000000,-0.121094,-0.246094)
    err_mag: 9.908438e-07
    f0: 5.5776e-07
    f1: 1.5486e-06
  sparse_1            : differs
Diff in sparse_0            
    bkji: ( 132,   0,   0,   0)
    zyx: (0.000000,-0.121094,0.191406)
    err_mag: 9.908438e-07
    f0: 5.5776e-07
    f1: 1.5486e-06
  sparse_0            : differs
Diff in sparse_2            
    bkji: ( 183,   0,   0,   7)
    zyx: (0.000000,0.066406,-0.191406)
    err_mag: 9.908438e-07
    f0: 5.5776e-07
    f1: 1.5486e-06
  sparse_2            : differs
Diff in sparse_3            
    bkji: ( 244,   0,   0,   7)
    zyx: (0.000000,0.066406,0.246094)
    err_mag: 9.908438e-07
    f0: 5.5776e-07
    f1: 1.5486e-06
  sparse_3            : differs
  File "/__w/parthenon/parthenon/tst/regression/run_test.py", line 204, in <module>
    main(**vars(args))
  File "/__w/parthenon/parthenon/tst/regression/run_test.py", line 93, in main
    raise TestError("Test " + test_manager.test + " failed")
__main__.TestError: Test sparse_advection failed
```
Might be related to the fine boundary conditions?

lroberts36:
@pgrete: Oh, I think I know what is going on. I updated the sparse gold file with the bugs in. I think this test succeeds with the old gold files with the bug fixes in. I will make sure this is true and update.

Issue #891, opened 2023-06-19
Timers based load balancing


Yurlungur:
Re-enabling the timers based load balancing in Athena++ is made more complex by the fact that in Parthenon kernels loop over collections of meshblocks, so it's not clear what granularity of timing is possible. Can we time within a loop? And if we can only time at the level of the whole loop, how do we distribute time cost across blocks? Perhaps there is some Kokkos functionality we can leverage here?

Partly motivated by discussions with @lroberts36

Yurlungur:
Perhaps memory footprint could be used to help inform how to distribute time cost across collections of blocks?

BenWibking:
WarpX spent a huge amount of effort to try to get this working with the profiling interfaces, but it apparently slowed down performance unacceptably to just turn it on.

An alternative might be to add a timer within the most expensive kernel and just accept the additional register usage (4 registers). For example, see: https://github.com/ECP-WarpX/WarpX/blob/92cc1054bfa9aeffbfa8e914eab350acb9c048f1/Source/Particles/Deposition/ChargeDeposition.H#L101

Implementation of `KernelTimer` here: https://github.com/ECP-WarpX/WarpX/blob/development/Source/ablastr/parallelization/KernelTimer.H

PR #892, opened 2023-06-19
Meshblock Cost Function and Diagnostics

### LINKED ISSUE #576: Default load balancing options/behavior
```c++
  if (pin->GetOrAddString("loadbalancing", "balancer", "default") == "automatic")
    lb_automatic_ = true;
  else if (pin->GetOrAddString("loadbalancing", "balancer", "default") == "manual")
    lb_manual_ = true;
```
I'm kind of surprised by this logic (but it's also in Athena++).
Looks like the "default" means neither "automatic" nor "manual". So if "automatic" is not default, what is?
I'll create an issue as it is unrelated to this PR.

_Originally posted by @pgrete in https://github.com/lanl/parthenon/pull/575#discussion_r677177156_

Yurlungur:
## PR Summary

Again part of the effort to get riot onto parthenon develop. Here I pull several features in from riot:

### Memory Diagnostics Per Meshblock

Whenever variables or swarms are allocated or deallocated, memory used on a given meshblock is updated in a running total. Note this is only "interior" data. I also add the ability to look at the comm buffer pool, but this is a per-mesh quantity since the pool is shared. Not per block. I also didn't add the equivalent logic for boundary swarms, as I didn't fully understand the swarm machinery, and it's probably a negligible contribution. @brryan should probably take a look at the memory diagnostic logic for swarms in case I missed something.

I showcase this capability by modifying the sparse example to show memory footprint in bytes per cycle.

### Manual load balancing

- I "made public" manual cost-based load balancing again. It's enabled in the input deck by setting `parthenon/loadbalancing/balancer=manual`. 
- I also explicitly disable timers based load balancing, as this isn't fully threaded through the infrastructure anymore. I think it's not trivial to figure out what to do here, and I created an issue, #891 to discuss.
- I clarify what it means to use the manual vs default load balancing. Automatic load balancing is timers. Manual is the cost-based where downstream codes are expected to call the meshblock handles. Default is just meshblock count.

I also document both of the above new features.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Current develop
 
 ### Added (new features/APIs/variables/...)
+- [[PR 892]](https://github.com/parthenon-hpc-lab/parthenon/pull/892) Cost-based load balancing and memory diagnostics
 - [[PR 872]](https://github.com/parthenon-hpc-lab/parthenon/pull/872) Boundary communication for non-cell centered fields
 - [[PR 877]](https://github.com/parthenon-hpc-lab/parthenon/pull/877) Add flat sparse packs
 - [[PR 868]](https://github.com/parthenon-hpc-lab/parthenon/pull/868) Add block-local face, edge, and nodal fields and allow for packing
--- a/doc/sphinx/src/interface/sparse.rst
+++ b/doc/sphinx/src/interface/sparse.rst
@@ -412,3 +412,29 @@ block. The remaining changes are as follows:
    it's not yet allocated on the receiving block. We then proceed to
    read the buffer only if the variable is allocated on the receiving
    block.
+
+Memory Footprint Reporting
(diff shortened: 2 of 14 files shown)


Yurlungur:
Resolves #576

pdmullen (approved):
LGTM

lroberts36 (approved):
LGTM

Yurlungur:
Athenapk folks: @pgrete @forrestglines do you want to take a look or should I just pull the trigger? I think the impact on athenapk is minimal.

Yurlungur:
If I don't hear from AthenaPK folks by the end of tomorrow (Tuesday), I'll just pull the trigger. I don't think there's anything in here that should impact you.

PR #894, opened 2023-06-20
map instead of unordered_map for sparse pool

Yurlungur:
## PR Summary

The sparse pool uses an unordered map. However, sparse ids have an implicit ordering, they are integers after all. While not a bug exactly, the unordered map can mangle the ordering and can produce un-intuitive behavior where a user may request sparse IDs and expect the pool to return an ordered list. This PR changes the map to an ordered one (i.e., a heap) so traversal through it always produces sparse IDs sorted from smallest to largest. 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - [[PR 868]](https://github.com/parthenon-hpc-lab/parthenon/pull/868) Add block-local face, edge, and nodal fields and allow for packing
 
 ### Changed (changing behavior/API/variables/...)
+- [[PR 894]](https://github.com/parthenon-hpc-lab/parthenon/pull/894) Demand that sparse pool order sparse ids
 - [[PR 885]](https://github.com/parthenon-hpc-lab/parthenon/pull/885) Expose PackDescriptor and use uids in SparsePacks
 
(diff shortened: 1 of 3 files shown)


pdmullen (approved):
LGTM

lroberts36 (approved):
LGTM

Yurlungur:
Cuda CI test seems to be failing here, but it's working when I run on a local cuda machine. @lroberts36 is reporting similar issues. @pgrete is there perhaps something wrong with the CI machine?

pgrete:
> Cuda CI test seems to be failing here, but it's working when I run on a local cuda machine. @lroberts36 is reporting similar issues. @pgrete is there perhaps something wrong with the CI machine?

Cuda tests did eventually pass, didn't they? And yes, as I mention on Matrix yesterday the machine is more heavily used. now so it's likely just a fight for resources.
Given that it's not "my" node, there's currently very little I can do to fix this.

Yurlungur:
> Cuda tests did eventually pass, didn't they? 

Yes eventually. Perhaps just needs to be timed for nighttime in the EU

PR #900, opened 2023-06-28
Add Morton numbers and expand `LogicalLocation` functionality

lroberts36:
## PR Summary

In preparation for dealing with topological elements that are shared between blocks, this PR adds some utilities for calculating Morton numbers from logical locations and expands the functionality of `LogicalLocation`. In particular:
1. A `LogicalLocation` instance now includes its Morton number. 
2. `LogicalLocation` member variables are private and are accessed through const accessors to prevent the logical location and Morton number from being inconsistent.
3. `>`, `<`, and `==` operators are introduced for `LogicalLocation` since the locations have a Morton order. 
4. A custom hash function for `LogicalLocation` is provided so that we can make `std::unordered_map`s and `std::set`s with `LogicalLocation` as a key. 
5. Member functions for testing if a `LogicalLocation` contains, is contained in, or is a neighbor of another `LogicalLocation` are added. 
6. Member functions for getting parent and daughter `LogicalLocation`s are added.  
7. Member functions for getting a set of possible `LogicalLocation`s abutting a given topological element are added. 
8. Function for determining a blocks ownership of topological elements relative to a list of blocks.

A unit test that exercises this new functionality is included. The `LogicalLocation` of a neighbor is now stored in the `NeighborBlock` class.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


Yurlungur:
planning to review this ASAP. Didn't get a chance today.

lroberts36:
Thanks @Yurlungur. I thought of a couple things I could add to this PR tonight anyway.

src/bvals/bvals_base.cpp
```diff
@@ -51,9 +51,9 @@ NeighborIndexes BoundaryBase::ni[NMAX_NEIGHBORS];
 //                          int ibid, int itargetid, int ifi1=0, int ifi2=0)
 // \brief Set neighbor information
 
-void NeighborBlock::SetNeighbor(int irank, int ilevel, int igid, int ilid, int iox1,
-                                int iox2, int iox3, NeighborConnect itype, int ibid,
-                                int itargetid,
+void NeighborBlock::SetNeighbor(LogicalLocation inloc, int irank, int ilevel, int igid,
```
Yurlungur, line 54: Assuming here that `irank` is mpi rank? I'm not sure I like the idea of that information being exposed unless it has to be. What is that used for here? It's not needed for the space filling curve is it?

src/utils/morton_number.hpp
```diff
@@ -0,0 +1,108 @@
...
+  z = (z >> (chunk * NBITS)) & lowest_nbits_mask;
+
+  // Return the interleaved section of the morton number
+  return InterleaveZeros<3, NBITS>(z) << 2 | InterleaveZeros<3, NBITS>(y) << 1 |
+         InterleaveZeros<3, NBITS>(x);
+}
+} // namespace impl
+
+struct MortonNumber {
+  uint64_t most, mid, least;
```
Yurlungur: Is this just to fit 192 bits? If so, I find the naming confusing. I'm not sure I have a better name though. Perhaps a length 3 array called "digits" or something?

src/defs.hpp
```diff
@@ -56,21 +60,211 @@ struct LogicalLocation { // aggregate and POD type
   // These values can exceed the range of std::int32_t even if the root grid has only a
   // single MeshBlock if >30 levels of AMR are used, since the corresponding max index =
   // 1*2^31 > INT_MAX = 2^31 -1 for most 32-bit signed integer type impelementations
-  std::int64_t lx1, lx2, lx3;
-  int level;
+
+ private:
```
Yurlungur: `LogicalLocation` is now complex enough that maybe the struct deserves its own header file. Doesn't feel appropriate for it to live in `decs.h` anymore.

Also maybe it should be made a `class`. I know those are the same as structs, but now that it has private member data, I feel like the `class` designation better conveys intent.

src/defs.hpp
```diff
@@ -56,21 +60,211 @@ struct LogicalLocation { // aggregate and POD type
...
+  }
+
+  bool IsContainedIn(const LogicalLocation &container) const {
+    if (container.level() > level_) return false;
+    const std::int64_t shifted_lx1 = lx1_ >> (level_ - container.level());
+    const std::int64_t shifted_lx2 = lx2_ >> (level_ - container.level());
+    const std::int64_t shifted_lx3 = lx3_ >> (level_ - container.level());
+    return (shifted_lx1 == container.lx1()) && (shifted_lx2 == container.lx2()) &&
+           (shifted_lx3 == container.lx3());
+  }
```
Yurlungur: Really nice to have this kind of check (and friends below) in the `LogicalLocation` itself. I feel like that's a lot cleaner as far as logic goes.

src/defs.hpp
```diff
@@ -56,21 +60,211 @@ struct LogicalLocation { // aggregate and POD type
...
+    return LogicalLocation(level_, lx1_ + ox1, lx2_ + ox2, lx3_ + ox3);
+  }
+
+  LogicalLocation GetParent() const {
+    if (level_ == 0) return *this;
+    return LogicalLocation(level_ - 1, lx1_ >> 1, lx2_ >> 1, lx3_ >> 1);
+  }
+
+  std::vector<LogicalLocation> GetDaughters() const {
+    std::vector<LogicalLocation> daughters;
```
Yurlungur: I'm sure that returning a `std::vector` here is fine. But the performance curmudgeon in me at least wants to minimize the number of resizes the vector might do internally.
```suggestion
    std::vector<LogicalLocation> daughters;
    daughters.reserve(8);
```

src/defs.hpp
```diff
@@ -56,21 +60,211 @@ struct LogicalLocation { // aggregate and POD type
...
+            if (0 <= lx1 && lx1 < n_cells_level && 0 <= lx2 && lx2 < n_cells_level &&
+                0 <= lx3 && lx3 < n_cells_level) {
+              locs.emplace_back(loc.level(), lx1, lx2, lx3);
+              auto parent = locs.back().GetParent();
+              if (IsNeighbor(parent)) locs.push_back(parent);
+            }
+          }
+        }
+      }
+    };
```
Yurlungur: Should this definitely be a lambda and not its own function?

src/defs.hpp
```diff
@@ -56,21 +60,211 @@ struct LogicalLocation { // aggregate and POD type
...
+  }
+
+ private:
+  bool ownership[3][3][3];
+};
+
+inline block_ownership_t
+DetermineOwnership(const LogicalLocation &main_block,
+                   const std::set<LogicalLocation> &allowed_neighbors) {
+  block_ownership_t main_owns;
```
Yurlungur: the above ownership stuff can probably also go in the above-mentioned header

src/defs.hpp
```diff
@@ -56,21 +60,211 @@ struct LogicalLocation { // aggregate and POD type
...
+  const bool &operator()(int ox1, int ox2, int ox3) const {
+    return ownership[ox1 + 1][ox2 + 1][ox3 + 1];
+  }
+  bool &operator()(int ox1, int ox2, int ox3) {
+    return ownership[ox1 + 1][ox2 + 1][ox3 + 1];
+  }
+
+ private:
+  bool ownership[3][3][3];
+};
```
Yurlungur: Feels like a bit much for a 3^3 array of bools... why not just do
```C++
using block_ownership_t = bool[3][3][3];
```
and live with the bracket indexer? Or are you imagining this struct might become more complex?

Yurlungur (approved):
This is a nice addition and I can see it coming in handy both for code extensions and code cleanup going forward.

src/bvals/bvals_base.cpp
lroberts36: IIRC, I you need it for building the boundary buffers. It doesn't enter into the Morton order.

src/utils/morton_number.hpp
lroberts36: yeah, it is just to fit 192 bits. I thought about the naming a little bit, didn't come up with anything good, and then just moved on. It probably would be more easily extensible to store it as an array as you suggest.

src/bvals/bvals_base.cpp
Yurlungur: why isn't it in the older function if it's needed?

src/defs.hpp
lroberts36: Agreed on both accounts

src/defs.hpp
lroberts36: I agree about possible performance issues, but I am trying not to pre-optimize any of this code. I would rather get the logic right, keep things clean, and then look for hot spots, given that the logic can be a little complex.

src/defs.hpp
lroberts36: I mostly didn't want to have to remember to type `+1` everywhere and remember how to return fixed length arrays.

src/defs.hpp
Yurlungur: I think it's probably going to be fine performance-wise. It's not like this is performance-critical code. I just suggest the `reserve` call because it's one-liner.

src/defs.hpp
Yurlungur: oh my eyes completely glossed over the `+1`s. Yeah fair enough.

src/bvals/bvals_base.cpp
lroberts36: I think it is in the older one...

Yurlungur: oh I'm just blind.

src/defs.hpp
lroberts36: Well, it relies on capturing `irange`, `krange`, and `jrange` which depend on the offsets (as well as `loc`). I guess it would make sense to call it `GetSameLevelOrCoarserTopologicalElementNeighbors` and just pass `ox1`, `ox2` and `ox3` to it and calculate the ranges internally.

Yurlungur: I'm fine with it living where it currently does. Just wondered if it might belong elsewhere. Either way is fine.

src/utils/morton_number.hpp
lroberts36: switched to an array called `bits`.

src/defs.hpp
lroberts36: I think I will leave it here for now.

src/defs.hpp
lroberts36: done

src/defs.hpp
lroberts36: done

src/defs.hpp
lroberts36: added

pdmullen (approved):
LGTM!

tst/unit/test_logical_location.cpp
```diff
@@ -0,0 +1,350 @@
...
+    }
+  }
+}
+
+TEST_CASE("Morton Numbers", "[Morton Numbers]") {
+  constexpr int type_size = 8 * sizeof(uint64_t);
+  GIVEN("some interleave constants") {
+    std::map<int, uint64_t> ics_2d, ics_3d;
+
+    for (int pow = 1; pow <= type_size; pow *= 2) {
```
pgrete: I'm not a big fan of the `pow` variable name given the name clash with `pow` from math (or `std::pow`), but also don't feel strongly about this (so it's just a comment).

tst/unit/test_logical_location.cpp
```diff
@@ -0,0 +1,350 @@
...
+          }
+          lx2 = lx2 % cur_place;
+
+          if (lx1 / cur_place == 1) {
+            hand_morton[62 - (3 * i + 2)] = 1;
+          }
+          lx1 = lx1 % cur_place;
+        }
+        // Check that we have the correct Morton number
+        REQUIRE(hand_morton.to_ullong() == leaf.morton().bits[0]);
```
pgrete, line 172: If I understood correctly, the Morton number is split into 3 chunks with 64bit each (to encode x,y, z locations).
So why is it okay to just check one chunk here, because the others are 0 anyway? Wouldn't it be more consistent to also check for that?

src/mesh/logical_location.hpp
```diff
@@ -0,0 +1,276 @@
+//========================================================================================
+// Athena++ astrophysical MHD code
+// Copyright(C) 2014 James M. Stone <jmstone@princeton.edu> and other code contributors
+// Licensed under the 3-clause BSD License, see LICENSE file for details
```
pgrete, line 4: Is this still valid?

src/mesh/logical_location.hpp
```diff
@@ -0,0 +1,276 @@
...
+  }
+  // LFR: These are old comparison operators. Greater gets used for sorting
+  //      the derefinenment list, but we may want to remove them at some
+  //      point to avoid confusion with the comparators below.
+  static bool Lesser(const LogicalLocation &left, const LogicalLocation &right) {
+    return left.level() < right.level();
+  }
+  static bool Greater(const LogicalLocation &left, const LogicalLocation &right) {
+    return left.level() > right.level();
+  }
```
pgrete: How about renaming them to be more specific in the meantime (assuming that removing them is a larger lift)?

src/mesh/logical_location.hpp
```diff
@@ -0,0 +1,276 @@
...
+
+} // namespace parthenon
+
+template <>
+struct std::hash<parthenon::LogicalLocation> {
+  std::size_t operator()(const parthenon::LogicalLocation &key) const noexcept {
+    // TODO(LFR): Think more carefully about what the best choice for this key is,
+    // probably the least significant sizeof(size_t) * 8 bits of the morton number
+    // with 3 * (level - 21) trailing bits removed.
+    return key.morton().bits[0];
```
pgrete, line 263: Similar to the question below: Why is it okay to just pick the first chunk of bits?

Bonus question: Do you foresee a case where the current choice of key causes issues? (I'm asking in terms of how important/relevant the `TODO` is)

pgrete (approved):
Nice work -- lots of bits flying around.
I only got a couple of clarification questions (though feel free to ignore and merge if you're on a tight schedule).
Similarly (assuming you can spare another 15 min or so), a (very brief) documentation would be great.
I remember that I had a similar question to the Athena++ devs in terms of "What's the best way to get a global index of a cell in an AMR sim?" it would  be nice if we have that kind of info readily available in our docs.

tst/unit/test_logical_location.cpp
lroberts36: You are correct on all accounts. I am only calculating the most significant bits by hand and the rest should be zero. I will add specific requirements for that.

src/mesh/logical_location.hpp
lroberts36: I am not sure, I left it in since there is still some code that presumably came from Athena++, although it is a small fraction of the file.

src/mesh/logical_location.hpp
lroberts36: Well, this is just for hashing so we can choose whatever hash we would like I think (e.g. we could have this function always return one since there is no requirement for the hash to be unique, although this would of course give terrible performance). For less that 21 levels of refinement, the most significant bits should be a perfect hash, but for more refinement it is no longer perfect. Then, it seems like it might be better to use the bits described in the comment to minimize possible collisions. Since we are often just making `std::unordered_map`s of `LogicalLocations` in a small region of the grid, if we just used the most significant bits the hashes of all blocks around a corner that was at a refinement level > 22 would all collide if we just hashed with the most significant bits. Using the least significant bits would result in a locally perfect hash (most of the time).

tst/unit/test_logical_location.cpp
lroberts36: Agreed, will change this.

lroberts36:
@pgrete re-documentation: My plan was to write some comprehensive documentation as a separate PR. I know I say this all the time about updates and never do it, but I think it will be much easier to write a comprehensive document at one time than doing it piecemeal (and I swear I am not lying about eventually doing it).

lroberts36:
@pgrete re-documentation: My plan was to write some comprehensive documentation as a separate PR. I know I say this all the time about updates and never do it, but I think it will be much easier to write a comprehensive document at one time than doing it piecemeal (and I swear I am not lying about writing up everything eventually ;)).

src/mesh/logical_location.hpp
lroberts36: I have to go look at where they get used. It may only be one place in the code, possibly `bvals`

lroberts36: Ok, I should have read my comment. Only `Greater` is used in one place in AMR. I just replace that use with a lambda and removed the old comparators.

src/mesh/logical_location.hpp
lroberts36: I am going to leave it as is for now.

src/mesh/logical_location.hpp
lroberts36: I am just going to leave it.

PR #901, opened 2023-06-29
Functionality for `Indexer` masking

lroberts36:
## PR Summary

This PR add the functionality for determining ownership based masks for boundary communication and implements those masks in the boundary communication routines. The basic problem this PR tries to solve is how to synchronize elements that are shared between blocks (e.g. the edge of a zone that is on the edge of a block may be shared by up to four blocks and those blocks may be on different refinement levels). Rather than let these evolve separately (as is done in Athena++, I think), one particular block is chosen to own each element and it communicates the value it has to the other blocks that share the element (potentially after restriction on a fine block). Ownership is determined first by the most refined block and then if multiple sharing blocks are at the same refinement level, the block with the maximum Morton number is chosen to own the element. 

The index space for owned elements of a particular type on a given block may not be rectangular. As an example, consider a nodal field on a block that has one neighbor block on a corner that is more refined. When nodal zones are passed to a face neighbor (that is also a neighbor of the corner neighbor) and the main block has higher ownership status than the face neighbor, the set of nodes that need to be communicated have one node removed at the corner. To deal with this, all zones possibly required by a neighbor block are packed into a communication buffer but when the boundaries are set in the receiving block it is done with a mask that takes into account ownership. The logic for building these ownership masks first requires building a 3^3 array for each block describing which of its faces, edges, and nodes it owns (blocks always own their interior). This is implemented in `DetermineOwnership`. This can then be transformed into a 3^3 array describing the ownership of a block for face, edge, and nodal fields on each of the possible elements of the block itself. Finally, this field type ownership array can be transformed into the mask by accounting for where in the interior of the block the buffer is being filled from. These two operations are performed in `GetIndexRangeMaskFromOwnership`. These methods are the meat of this PR. 

Additionally, some unit tests are included and everything seems to work with SMR in parthenon-mhd.  I also moved the `MakePackDescriptor` stuff to its own header file.

A list of some stuff I checked/did: 

- [x] Check logic for finding masks based on topological element, boundary of sending block, and sending block ownership is correct as implemented in the test code.
- [x] Move mask finding logic out of tests 
- [x] Include masking information in bound info objects (assuming no ownership for now).
- [x] Add more complete test suite.
- [x] Test against parthenon-mhd for uniform mesh
- [x] Test against parthenon-mhd for SMR mesh

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


src/interface/make_pack_descriptor.hpp
```diff
@@ -0,0 +1,114 @@
...
+#include "interface/mesh_data.hpp"
+#include "interface/meshblock_data.hpp"
+#include "interface/metadata.hpp"
+#include "interface/sparse_pack.hpp"
+#include "interface/state_descriptor.hpp"
+#include "mesh/mesh.hpp"
+
+namespace parthenon {
+
+inline auto MakePackDescriptor(StateDescriptor *psd, const std::vector<std::string> &vars,
```
Yurlungur, line 38: :+1: I approve of moving these. Worth noting though that this is a (very minor) breaking change for downstream codes. 

I was going to suggest putting this in the `parthenon/package.hpp` header file and sticking `MakePackDescriptor` in the parthenon package prelude namespace. But I see you beat me to it. :D

Yurlungur: Also, feel like sneaking
```C++
template <class... Ts>
inline auto MakePackDescriptor(SparsePack<Ts...> pack,
                               StateDescriptor *psd,                                
                               const std::vector<MetadataFlag> &flags = {},
                               const std::set<PDOpt> &options = {}) ) {
  return parthenon::MakePackDescriptor<Ts...>(psd, flags, options);
}
```
in here?

src/mesh/logical_location.hpp
```diff
@@ -244,6 +264,64 @@ DetermineOwnership(const LogicalLocation &main_block,
...
 }
 
+// Given a topological element, ownership array of the sending block, and offset indices
+// defining the location of an index region within the block (i.e. the ghost zones passed
+// across the x-face or the ghost zones passed across the z-edge), return the index range
+// masking array required for masking out unowned regions of the index space. ox? defines
+// buffer location on the owner block
+inline auto GetIndexRangeMaskFromOwnership(TopologicalElement el,
+                                           const block_ownership_t &sender_ownership,
+                                           int ox1, int ox2, int ox3) {
```
Yurlungur, line 285: This seems like it was a bit tricky to work out but the resultant code is pretty clean. I like it.

Yurlungur (approved):
Nice to see this. Excited to try face-centered fields with AMR out :D

src/interface/make_pack_descriptor.hpp
lroberts36: added.

lroberts36: I guess this is only breaking for Riot, right @Yurlungur?

src/mesh/logical_location.hpp
lroberts36: Getting to this point definitely required a bit of thought.

src/interface/make_pack_descriptor.hpp
Yurlungur: I think that's right. And it should be a pretty easy fix in riot, so I think the breaking change is worth it.

lroberts36: Just was deciding if I should mark the PR as breaking.

src/bvals/comms/bnd_info.cpp
```diff
@@ -82,25 +82,30 @@ Indexer6D CalcLoadIndices(const NeighborIndexes &ni, TopologicalElement el,
                  // offset in some direction
     } else if (block_offset[dir] > 0) {
       s[dir] = bounds[dir].e - Globals::nghost + 1 - top_offset[dir];
-      e[dir] = bounds[dir].e - top_offset[dir];
+      e[dir] = bounds[dir].e;
     } else {
-      s[dir] = bounds[dir].s + top_offset[dir];
+      s[dir] = bounds[dir].s;
```
pgrete, line 87: Is the offset not required any more because that the topological element at at index is not owned?

src/bvals/comms/bnd_info.cpp
```diff
@@ -128,15 +133,28 @@ Indexer6D CalcSetIndices(const NeighborIndexes &ni, LogicalLocation loc,
       }
       ++off_idx;
     } else if (block_offset[dir] > 0) {
-      s[dir] = bounds[dir].e + 1;
+      s[dir] = bounds[dir].e + 1 - top_offset[dir];
       e[dir] = bounds[dir].e + ghosts;
     } else {
       s[dir] = bounds[dir].s - ghosts;
-      e[dir] = bounds[dir].s - 1;
+      e[dir] = bounds[dir].s - 1 + top_offset[dir];
```
pgrete, line 140: Same as above: has the offset been added because this element now "owns"?

src/bvals/comms/bnd_info.cpp
```diff
@@ -152,9 +170,9 @@ int GetBufferSize(std::shared_ptr<MeshBlock> pmb, const NeighborBlock &nb,
   const int isize = cb.ie(in) - cb.is(in) + 2;
   const int jsize = cb.je(in) - cb.js(in) + 2;
   const int ksize = cb.ke(in) - cb.ks(in) + 2;
-  return (nb.ni.ox1 == 0 ? isize : Globals::nghost) *
-         (nb.ni.ox2 == 0 ? jsize : Globals::nghost) *
-         (nb.ni.ox3 == 0 ? ksize : Globals::nghost) * v->GetDim(6) * v->GetDim(5) *
+  return (nb.ni.ox1 == 0 ? isize : Globals::nghost + 1) *
+         (nb.ni.ox2 == 0 ? jsize : Globals::nghost + 1) *
+         (nb.ni.ox3 == 0 ? ksize : Globals::nghost + 1) * v->GetDim(6) * v->GetDim(5) *
```
pgrete, line 175: Is the buffer size increased to account for send/recv data that is (not) owned?

pgrete (approved):
Kudos for all that index magic.
I definitely cannot say that I am able to mentally process all the fine grained details.
I trust that if there's sth slightly off we'll track it down in various downstream codes.

src/bvals/comms/bnd_info.cpp
lroberts36: Previously (which only worked for single level meshes), I was not not sending shared elements between blocks. That meant that I had to offset the starting/stopping indices for buffer filling by the topological offset so I didn't grab these shared zones. Now, with ownership those shared zones are always passed in the buffers but they aren't always copied into the receiving block.

src/bvals/comms/bnd_info.cpp
lroberts36: Yep, same explanation as above.

src/bvals/comms/bnd_info.cpp
lroberts36: Yes, I have been not very careful about calculating buffer sizes. I essentially just take the maximum size possible (which corresponds to assuming a nodal field is being communicated). This means that the buffers are sometimes bigger than they need to be, which doesn't pose any issues for correctness. It does mean that the memory footprint is maybe a little larger than it needs to be, but this does mean that all of the buffer arrays can come from the same three memory pools.

lroberts36:
> I trust that if there's sth slightly off we'll track it down in various downstream codes.

I think that is right. I definitely spent a lot of time thinking about it and am pretty convinced things are correct, but at the end of the day I am sure I didn't think through every single edge case and I am relying on things generalizing correctly. The real proof is in it working, although it would be great if someone could think up a unit test which went through every possible case.    

I also condensed the indexing stuff to a single function in the following PR adding AMR. It might be a little easier to track through all of the index logic there.

PR #902, opened 2023-07-06
Add sparse seed nans flag to output

Yurlungur:
## PR Summary

When debugging sparse variables, it's useful to be able to output where sparse variables are specifically de-allocated, as opposed to just exactly zero. This small PR adds a tiny tweak to parthenon_hdf5 output which allows the user to output NaNs for sparse variables which are de-allocated, making them easy to find in output.

This is essentially a debugging tool and it's opt-in. However, it does have the advantage that by replacing `memset` with `std::fill` we could change what that fill value is at a later date to, for example, set it to whatever value the sparse field would be initialized to, if that value is not zero.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/src/outputs/parthenon_hdf5.cpp
+++ b/src/outputs/parthenon_hdf5.cpp
@@ -539,7 +539,7 @@ void PHDF5Output::WriteOutputFileImpl(Mesh *pm, ParameterInput *pin, SimTime *tm
                 vinfo.nx6 * vinfo.nx5 * vinfo.nx4 * vinfo.nx3 * vinfo.nx2 * vinfo.nx1;
           }
           auto fill_val = output_params.sparse_seed_nans
-                              ? std::numeric_limits<OutT>::signaling_NaN()
+                              ? std::numeric_limits<OutT>::quiet_NaN()
                               : 0;
           std::fill(tmpData.data() + index, tmpData.data() + index + varSize, fill_val);
           index += varSize;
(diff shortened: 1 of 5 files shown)


src/outputs/parthenon_hdf5.cpp
```diff
@@ -538,7 +538,10 @@ void PHDF5Output::WriteOutputFileImpl(Mesh *pm, ParameterInput *pin, SimTime *tm
             varSize =
                 vinfo.nx6 * vinfo.nx5 * vinfo.nx4 * vinfo.nx3 * vinfo.nx2 * vinfo.nx1;
           }
-          memset(tmpData.data() + index, 0, varSize * sizeof(OutT));
+          auto fill_val = output_params.sparse_seed_nans
+                              ? std::numeric_limits<OutT>::signaling_NaN()
```
pdmullen: here we use `signaling_NaN()` rather than `quiet_NaN()` because if we are doing any operations on output data then we should signal?

pdmullen (approved):

src/outputs/parthenon_hdf5.cpp
Yurlungur: You know to be honest I didn't think that through carefully. I just reflexively use signalling NaN for this kind of thing basically for that reason. But I'm not married to it, if you'd prefer a quiet NaN.

Yurlungur: Changed to `quiet_NaN()` since that should be a bit more performant and no ops are happening here anyway.

lroberts36 (approved):
LGTM, I heartily approve

PR #904, opened 2023-07-11
AMR for non-cell centered fields

### LINKED ISSUE #736: Plan for face-centered fields

lroberts36:
## PR Summary
This PR does the last few things required to make full AMR work with non-cell centered fields: 

- [x] Separates the prolongation and restriction caches from the boundary caches by creating a new `ProResInfo` type that is similar to `BndInfo`.
- [x] Unifies the boundary index calculation routine further. 
- [x] Switch to restriction in one during `RedistributeAndRefineMeshBlocks`
- [x] Switch to prolongation in one during `RedistributeAndRefineMeshBlocks`
- [x] Updates `TryRecvFineToCoarse` to deal with non-cell centered fields
- [x] Updates `TryRecvCoarseToFine` to deal with non-cell centered fields
- [x] Test that things actually work
- [ ] Come up with some good test problems for Parthenon (any ideas are appreciated)

B^2/2 w/ AMR
Screenshot 2023-07-11 at 2 27 01 PM

divB w/ AMR
Screenshot 2023-07-11 at 2 26 45 PM

## PR Checklist

- [x] Code passes cpplint
- [ ] Adds tests for new features.
- [x] Code is formatted
- [x] Docs build


src/mesh/mesh_refinement.cpp
```diff
@@ -68,70 +68,6 @@ MeshRefinement::MeshRefinement(std::weak_ptr<MeshBlock> pmb, ParameterInput *pin
   }
 }
 
-//----------------------------------------------------------------------------------------
-//  \brief restrict cell centered values
-
-void MeshRefinement::RestrictCellCenteredValues(Variable<Real> *var, int csi, int cei,
-                                                int csj, int cej, int csk, int cek) {
```
Yurlungur, line 75: I am SO happy to see these go away. :+1:

src/mesh/logical_location.hpp
```diff
@@ -100,6 +100,17 @@ class LogicalLocation { // aggregate and POD type
     return (shifted_lx1 == lx1_) && (shifted_lx2 == lx2_) && (shifted_lx3 == lx3_);
   }
 
+  std::array<int, 3> GetOffset(const LogicalLocation &neighbor) const {
```
Yurlungur, line 94: :+1:

pgrete:
I've reviewed the other two open PRs and now will get at reviewing this alongside a test implementation of CT.
In doing so I noticed that we don't have any regression (not unit) tests covering all the new features that came in or am I missing sth?
I'd be great to have some kind of basic regression test also part of Parthenon that illustrate how to use the new machinery in an integrated manner.

lroberts36:
@pgrete: Yeah, there is no regression test in here yet. The issue is I haven't thought of a good regression test (since we can't have physics in Parthenon). One possibility is filling blocks with Morton numbers and communicating for different field types. Then we can check that the boundaries and shared elements are filled with the correct values. The other (kinda stupid) thing that comes to mind is somehow filling the grid with a linear gradient, refining, and making sure we recover the correct values for all topological elements in the interior and exterior of the meshblocks. I guess we could also come up with some sort of physics free constrained transport problem. Any ideas are welcome, a list of things to do for non-cell centered fields after this PR is in #908.

src/bvals/comms/bnd_info.hpp
```diff
@@ -78,11 +71,68 @@ struct BndInfo {
                                  CommBuffer<buf_pool_t<Real>::owner_t> *buf);
 };
 
+struct ProResInfo {
```
Yurlungur, line 74: I *think* it makes sense to split things out this way. At least I don't object.

src/bvals/comms/bnd_info.hpp
```diff
@@ -78,11 +71,68 @@ struct BndInfo {
...
+  int ntopological_elements = 1;
+  // Has to be large enough to allow for maximum integer
+  // conversion of TopologicalElements
+  SpatiallyMaskedIndexer6D idxer[10];
+
+  CoordinateDirection dir;
```
Yurlungur, line 80: Is this just `+1` or `-1`? Or is it `x`, `y`, `z`? Presumably more information is needed to specify buffer?

src/bvals/comms/bnd_info.cpp
```diff
@@ -190,69 +199,129 @@ BndInfo BndInfo::GetSendBndInfo(std::shared_ptr<MeshBlock> pmb, const NeighborBl
...
+  auto elements = v->GetTopologicalElements();
+  out.ntopological_elements = elements.size();
+  for (auto el : {TE::CC, TE::F1, TE::F2, TE::F3, TE::E1, TE::E2, TE::E3, TE::NN})
+    out.idxer[static_cast<int>(el)] =
+        CalcIndices(nb, pmb, el, IndexRangeType::Exterior, true, {Nt, Nu, Nv});
+  out.refinement_op = RefinementOp_t::Prolongation;
+  return out;
+}
+
+ProResInfo ProResInfo::GetSend(std::shared_ptr<MeshBlock> pmb, const NeighborBlock &nb,
```
Yurlungur, line 286: `GetSend`, `GetSet`, etc have a lot of the same boiler plate... I wonder if it could be unified? And if so, I wonder if it's worth it? It might just make things less clear...

lroberts36:
A possible test problem is to evolve 
```math
\partial_t \vec{F} + \nabla \times \vec{G} = 0 
```
```math
\partial_t \vec{G} + \nabla \times \vec{F} = 0
```
which should obey the wave equation but also have $\nabla \cdot \vec{F} = 0$ and $\nabla \cdot \vec{G} = 0$.

Yurlungur (approved):
I'm sorry it took me so long to review this... looks like I had started reviewing it and then I must have gotten interrupted and forgotten because I never finished. Overall this looks good.

One thing: we should probably document the ownership model somewhere in the docs.

Yurlungur:
P.S. I am so excited by this

Yurlungur:
> A possible test problem is to evolve
> 
> which should obey the wave equation but also have ∇⋅F→=0 and ∇⋅G→=0.

:+1: Vector wave equation seems like a good test problem and a nice demonstration of the technology. Maybe as a new example?

src/bvals/comms/bnd_info.hpp
lroberts36: The rational was that it the boundary communication information is unnecessary when doing prolongation and restriction in one during remeshing. Additionally, we may want to not build `ProResInfo` objects for regions that don't require prolongation/restriction at some point to minimize the storage here, since at least as written there are a bunch of `IdxRange`s in `ProResInfo`

lroberts36: Also for GMG, I pair different `BndInfo` creators and `ProResInfo` creators than for normal boundary communication.

Yurlungur: :+1: that makes sense.

src/bvals/comms/bnd_info.hpp
lroberts36: hm, I'm not sure why that is there. It may have just been copy and pasted along from the original `BndInfo` implementation...

src/bvals/comms/bnd_info.cpp
lroberts36: I thought about it a little, but I didn't see an obvious way to unify them. I also think I like that the interface is clear about what type of object you are getting.

Yurlungur: Yeah that's fine with me. I *think* things could be unified but it feels like it comes at the cost of clarity.

lroberts36:
Yeah, we should add it as an example. I think maybe we save that for a documentation/test PR. As you say, it would be good to have documentation about how the ownership model works, etc. It would be especially good to get done before it is forgotten...

Yurlungur:
Yeah I think the ownership model makes sense, and your code is clear. But it's sufficiently different from how Athena++ does it, and there are a sufficiently large number of choices to make, we should clarify which choices we made in a single place.

Yurlungur:
I'm fine with putting docs and an example off into a final PR though.

src/defs.hpp
```diff
@@ -86,6 +69,45 @@ struct RegionSize { // aggregate and POD type; do NOT reorder member declaration
...
+
+  bool &symmetry(CoordinateDirection dir) { return symmetry_[dir - 1]; }
+  const bool &symmetry(CoordinateDirection dir) const { return symmetry_[dir - 1]; }
+};
+
+//----------------------------------------------------------------------------------------
+// enums used everywhere
+// (not specifying underlying integral type (C++11) for portability & performance)
+
+// TODO(felker): C++ Core Guidelines Enum.5: Don’t use ALL_CAPS for enumerators
```
pdmullen: leftovers from Athena++

src/bvals/comms/bnd_info.cpp
lroberts36: Agreed that there is probably some unification that could occur, but I think I am going to leave that be for now (since other PRs based on this one would probably require some complex merging to change things in this PR at this point).

pdmullen (approved):
Not too much to contribute here... LGTM.  The test problems you have presented from various downstream applications give me confidence in the implementation.

src/defs.hpp
lroberts36: I think this all goes in one of the subsequent GMG PRs

c-prather (approved):
Finally using this in earnest and it works great.  Can't yet provide a working downstream due to my own bugs, but the machinery's all there and the interface has been easy to use.
Though it shouldn't delay this PR longer, we should remember to add the `InternalRefinementOp` template arg to `Metadata` when we get a chance, I started implementing the B field internal refinement outside of Parthenon proper and was briefly stung.

Issue #816, opened 2023-01-23
Parthenon Enhancement Proposal 1 -- Support Subclassing StateDescriptor


c-prather:
## Introduction

The Parthenon "Package" is a powerful way to split up code and data/parameters by function, keeping parameters and fields associated with the functions that handle them.  However, current limitations make it difficult for users to extend the definition of a "Package" to better fit their needs.  This "Parthenon Enhancement Proposal" (PEP) explains the desired design pattern, and outlines ways to accommodate it in Parthenon if this is desired.

I partially also intend this as a template for future PEPs, and would welcome format feedback.  Ideally future submissions will not be so verbose.

## Relevant background and current status

A Parthenon package consists of three things:

1. A C++ namespace (by convention, at least)
2. A (function constructing and returning a) `StateDescriptor` object, which is basically a host for a `Parameters` object, which is more or less a `std::map<std::string,std::any>`.  A list of `StateDescriptor` objects is then accessible as a member of any `Mesh` or `MeshBlock` object, i.e., from nearly any function in the code
3. Implementations of zero or more common functions, which are assigned to function pointer members of the `StateDescriptor` object and called at specific times during a Parthenon step (Some of them only if one subclasses a particular `Driver`)

This is a neat paradigm! It makes nearly arbitrary data & parameters automatically accessible nearly anywhere in a code from the `Mesh`/`MeshBlock` objects, while still hierarchically organized.  The function pointer or "callback" structure additionally allows implementing new features in a way that can be enabled & disabled cleanly, without a bunch of `if (feature_enabled)` constructs, which are hard to read and easy to get wrong (e.g. reading options & arrays that don't exist, running an extra op or fix that should have been disabled, etc.)

The full list of current callbacks is (as documented somewhat obliquely in `metadata.md`):
```c++
// Per-package callbacks
void PreFillDerivedBlock(MeshBlockData<Real> *rc);
void PreFillDerivedMesh(MeshData<Real> *rc);
void PostFillDerivedBlock(MeshBlockData<Real> *rc);
void PostFillDerivedMesh(MeshData<Real> *rc);
void FillDerivedBlock(MeshBlockData<Real> *rc);
void FillDerivedMesh(MeshData<Real> *rc);

void PreStepDiagnosticsMesh(SimTime const &simtime, MeshData<Real> *rc);
void PostStepDiagnosticsMesh(SimTime const &simtime, MeshData<Real> *rc);

[...]

PR #907, opened 2023-07-12
PEP 1

c-prather:
This is the minimum viable implementation of PEP 1, #816.  Ideally this would include documentation, but I'm really not sure what to write short of an example.

## PR Summary
See #816 for a description of why this is cool.  tl;dr it takes Parthenon's callback model (e.g., driver calls `Update::FillDerived`, which then calls any package's registered `FillDerived` function) and allows user extensions (e.g., driver calls `MyPackages::ApplyFloors` and that calls a bunch of registered `ApplyFloors` functions depending on which packages are loaded).

See [here](https://github.com/AFD-Illinois/kharma/blob/kharma-next/kharma/kharma_package.hpp) for an example extension package type.

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build


c-prather:
@par-hermes format

src/interface/packages.hpp
```diff
@@ -26,15 +27,61 @@ class Packages_t {
...
+  template <typename T>
+  const std::vector<T*> ListPackagesOfType() const {
+    std::vector<T*> sub_list;
+    for (auto package : packages_) {
+      if (T *cast_package = dynamic_cast<T*>(package.second.get())) {
+        sub_list.append(cast_package);
+      }
+    }
+    return sub_list;
+  }
```
Yurlungur: I don't super object to the new dictionary machinery here, but I wonder if a better approach might be to expose a loop structure. For example:
```C++
template <typename T, typename F>
void ForAllPackagesOfType(F &f) const {
  for (auto &[name, pkg] : packages_) {
    if (T *cast_package = dynamic_cast<T*>(pkg.get())) {
      f(name, package);
    }
  }
}
```

src/interface/packages.hpp
```diff
@@ -26,15 +27,61 @@ class Packages_t {
...
+  template <typename T>
+  const Dictionary<T*> AllPackagesOfType() const {
+    Dictionary<T*> sub_dict;
+    for (auto package : packages_) {
+      if (T *cast_package = dynamic_cast<T*>(package.second.get())) {
+        sub_dict[package.first] = cast_package;
+      }
+    }
+    return sub_dict;
+  }
```
Yurlungur: Looks like you added this twice?

Yurlungur (commented):
This seems pretty minimally invasive, so if it buys you a lot, I'm happy to include it.

src/interface/packages.hpp
c-prather: I considered that, but I think it would be common/useful to add other elements besides just the function call -- e.g. all my loops right now include extra code that can print the function being called, and I can see them as a good point for adding other debugging or common code.

I also sometimes (often) need control of the order that functions are called, e.g. one "floors" function relies on the results of previous.  Currently I just hack it by pulling out the package I know needs to be first, then looping over the others.  A more general approach might be callback "priorities," but that gets hard to follow reading the code and is more complex to implement so it's not clear we'd want it.

Since only maybe half my callers look anything like that prototype, and I gave up trying to cover the different cases.  I'm happy to provide it for other implementations though.

src/interface/packages.hpp
c-prather: Yeah git duplicated everything merging and I didn't catch this part.  Fixed.

src/interface/packages.hpp
Yurlungur: Nah it shouldn't matter much.

c-prather:
@par-hermes format

c-prather:
I've had to additionally mark some `StateDescriptor` members `protected` since I now would like to access them from KHARMA package code.  I think all members should be accessible to subclasses, but if there's anything which should still be `private` I can move it back

Yurlungur:
> I've had to additionally mark some `StateDescriptor` members `protected` since I now would like to access them from KHARMA package code. I think all members should be accessible to subclasses, but if there's anything which should still be `private` I can move it back

I think that's fine

Yurlungur (approved):

pgrete (approved):
LGTM

Yurlungur:
@bprather looks like the CI is blocking due to a dependency? Is that right?

c-prather:
There aren't unmet dependencies -- it looks like the check itself is broken (`cannot find file ...`).  Could force-merge this and wait for a fix to the check, or remove the check entirely with a new PR.  Happy to do either (I think I introduced the check in the first place)

EDIT upstream issue: https://github.com/gregsdennis/dependencies-action/issues/24

jonahm-LANL:
You are likely the one who would fix it. So I defer: do you want to remove it or try to debug it?

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________
From: Ben Prather ***@***.***>
Sent: Wednesday, October 18, 2023 10:04:14 AM
To: parthenon-hpc-lab/parthenon ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [EXTERNAL] Re: [parthenon-hpc-lab/parthenon] PEP 1 (PR #907)


There aren't unmet dependencies -- it looks like the check itself is broken (cannot find file ...). Could force-merge this and wait for a fix to the check, or remove the check entirely with a new PR. Happy to do either (I think I introduced the check in the first place)

c-prather:
Looks like it was fixed upstream already! So, I'm gonna hit go on this.

PR #917, opened 2023-08-09
Update Iterative Tasking Infrastructure

pdmullen:
## PR Summary

I am submitting this PR on behalf of @jdolence who made these changes in `lroberts36/merge-sparse-with-jdolence-sparse` some time ago.  In practice, this PR slightly modifies the task list infrastructure to update iterative tasking, in particular with regional dependencies.  This PR fixes various failure modes exhibited in downstream codes whereby (1) iterative tasks failed to iterate, (2) hangs were experienced, or (3) unexpected results were obtained with MPI both with and without regional dependencies.  

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [ ] Docs build

### DIFF
--- a/src/tasks/task_list.hpp
+++ b/src/tasks/task_list.hpp
@@ -381,7 +381,7 @@ class TaskRegion {
   }
   void AddRegionalDependencies(const std::string &reg_dep_id, const int list_index,
                                const TaskID &id) {
-    AddDepdencies(reg_dep_id, list_index, id);
+    AddDependencies(reg_dep_id, list_index, id);
     global[reg_dep_id] = false;
   }
   void AddGlobalDependencies(const int reg_dep_id, const int list_index,
@@ -390,7 +390,7 @@ class TaskRegion {
   }
   void AddGlobalDependencies(const std::string &reg_dep_id, const int list_index,
                              const TaskID &id) {
-    AddDepdencies(reg_dep_id, list_index, id);
+    AddDependencies(reg_dep_id, list_index, id);
     global[reg_dep_id] = true;
   }
 
@@ -466,7 +466,7 @@ class TaskRegion {
   }
 
  private:
-  void AddDepdencies(const std::string &label, const int list_id, const TaskID &tid) {
+  void AddDependencies(const std::string &label, const int list_id, const TaskID &tid) {
     id_for_reg[label][list_id] = tid;
     lists[list_id].MarkRegional(tid);
     all_done[label].val = 0;
--- a/src/tasks/task_list.hpp
+++ b/src/tasks/task_list.hpp
@@ -24,6 +24,7 @@
 #include <stdexcept>
 #include <string>
 #include <tuple>
+#include <unordered_map>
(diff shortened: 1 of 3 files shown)


Yurlungur (approved):
Thanks for porting this in. Sorry I missed it on the first batch of updates.

src/tasks/task_list.hpp
```diff
@@ -44,17 +45,18 @@ TaskID AddTaskHelper(TaskList *, Task);
...
   // NOTE: we must capture the object pointer
-  template <class T, class U, class... Args>
-  TaskID AddTask(TaskID const &dep, TaskStatus (T::*func)(Args...), U *obj,
-                 Args &&...args) {
+  template <class T, class U, class... Args1, class... Args2>
+  TaskID AddTask(TaskID const &dep, TaskStatus (T::*func)(Args1...), U *obj,
+                 Args2 &&...args) {
     return this->AddTask_(TaskType::iterative, 1, dep, [=]() mutable -> TaskStatus {
-      return (obj->*func)(std::forward<Args>(args)...);
+      return (obj->*func)(std::forward<Args2>(args)...);
```
lroberts36, line 59: I don't immediately see why we need two separate variadic template parameters here. I am mostly just curious what issue the single list was causing. Maybe it was something to do with default arguments?

src/tasks/task_list.hpp
```diff
@@ -393,7 +466,12 @@ class TaskRegion {
   }
 
  private:
-  bool HasRun(const int reg_id) {
+  void AddDepdencies(const std::string &label, const int list_id, const TaskID &tid) {
```
lroberts36:
```suggestion
  void AddDependencies(const std::string &label, const int list_id, const TaskID &tid) {
```

lroberts36 (approved):
LGTM, thanks for porting this. I did not go through the logic in detail, but that is probably not as useful as the test that it actually works in Riot.

src/tasks/task_list.hpp
jdolence: Unfortunately I can't remember _exactly_ what situation led to this change, but I can give you the basic idea.  What happened was that I wrote a task with some set of reasonable arguments (I thought) but the compiler yelled at me about type mismatches when I actually tried to add it to a task list.  Don't hold me to this, but I think it was a `const` vs non-`const` issue.  I convinced myself at the time that there was no reason we should be preventing people from writing tasks the way I had written the problematic one, and this transformation generalized the interface to allow it.

pdmullen:
Extended regression is failing upon enabling auto-merge. Is this definitely pointing to an issue with the changes in this PR? Or could it be unrelated? 

Asking before I jump down the rabbit hole.

Yurlungur:
> Extended regression is failing upon enabling auto-merge. Is this definitely pointing to an issue with the changes in this PR? Or could it be unrelated?

Given that automerge eventually let it go, sounds like the issue was just the test machine overloaded.

PR #920, opened 2023-08-10
Remove global non-optional print statements from Parthenon

c-prather:
Generally, Parthenon has taken an approach I quite like, of printing to console only when there is an error which the user (of Parthenon, not the code end-user) should be aware of.  Generally, it instead provides methods by which a downstream can print, which allows for a lot of flexibility in what information is presented to which logs, at what verbosity levels, etc.

I'm proposing this PR to check whether this is indeed the policy, and clean up some straggling non-error prints if that's the decision going forward.

After this PR, in KHARMA the only lines Parthenon prints are the per-step diagnostics and run-end diagnostics, both from the `EvolutionDriver`.  These seem universally useful for any code using `EvolutionDriver`, and are also just defaults, not forced -- they are easy to override in a subclass if the user wants to quash or reformat them.

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build


Yurlungur:
I think the three outputs you want to surpress here are broadly useful, and I would prefer they be preserved somehow. Perhaps this lends itself to some kind of logging machinery---if the log level is above "quiet" then these setup lines are output.

That said, why are these lines a concern? They're output once at setup ever. So it's not a particularly verbose output.

c-prather:
I agree completely that these outputs are useful; I'll probably keep them nearly as-is in KHARMA, at least at a suitable verbosity level.  However, I don't think they're so useful as to merit inclusion in the output of every run of every Parthenon downstream ever, regardless of configuration, runtime or compile-time options, which driver (if any) is in use, etc etc.

The reason I favored removal is that these outputs are really easy to add back to downstreams.  The `PreExecute` function can always be overridden in a `Driver` subclasses, and the variables can be printed as a one-liner from any function with access to a mesh pointer:
```c++
std::cout << "# Variables in use:\n" << *(pmesh->resolved_packages) << std::endl;
```
What's more, I'm sure downstream codes have more relevant information they could add to those notes, or irrelevant stuff they could suppress (e.g. none of my simulations will have `Swarm`s for a while yet).  Providing the info in every run, in a single generic way, seems less trusting of the user than I expect from Parthenon.

I'm fine with exposing a verbosity level option for Parthenon (so long as there's always an option to print nothing), but we already print so little that it seems like more code lines would be spent adding the option than would be spent duplicating the current functionality in every downstream.  I especially don't want to be in a position of providing general logging functionality to the user code, as it would need to be flexible and therefore complex -- logging packages exist for a reason, and I see no need for Parthenon to become another.

Yurlungur:
I don't feel particularly strongly either way, but IMO why make people overload parthenon to generate this output, instead of just printing it automatically? The level of trust is the same, as the downstream user can always disable it.

c-prather:
For the line in `PreExecute` I agree.  It's basically the same situation as the per-step & ending outputs.  Not too attached to removing that, I put it in another commit specifically to revert easily.

The variables table is impossible to disable, though, without output redirection or other tricks.  It seemed like there were a few other useful outputs during mesh creation, maybe putting it behind a `mesh_verbose` option (with like an overall structure & meshblock number printout, etc) would be the best of both worlds?

Yurlungur:
> For the line in `PreExecute` I agree. It's basically the same situation as the per-step & ending outputs. Not too attached to removing that, I put it in another commit specifically to revert easily.

The line in `PreExecute` is the one I'm least attached to... it's not a particularly useful line.

> The variables table is impossible to disable, though, without output redirection or other tricks. It seemed like there were a few other useful outputs during mesh creation, maybe putting it behind a `mesh_verbose` option (with like an overall structure & meshblock number printout, etc) would be the best of both worlds?

I'd be fine with this being in some kind of mesh structure output. You could also move the variables table into the `PreExecute` driver function. There's no reason it has to be in `Mesh`.

c-prather:
Good idea having it in `PreExecute`, I like that.  I think the current commit should print identical output to `develop` unless you override `PreExecute`, but I'll test this quick.

Maybe the broader policy here is to only print normal operation messages from `Driver` class methods?  That way Parthenon the library is silent for flexibility, but Parthenon the framework/driver has some sensible defaults that don't leave the user clueless as to what it's doing.

Yurlungur:
> Maybe the broader policy here is to only print normal operation messages from Driver class methods? That way Parthenon the library is silent for flexibility, but Parthenon the framework/driver has some sensible defaults that don't leave the user clueless as to what it's doing.

:+1: I like it.

Yurlungur (approved):

pgrete:
I also like the suggested solution.

pgrete (approved):

PR #921, opened 2023-08-11
Provide more flexible ways to add subset MeshBlockData objects to DataCollections

jdolence:
## PR Summary
There are lots of examples where it would be useful to have more flexibility in adding new `MeshBlockData` objects to `DataCollection`s that are slices of other `MeshBlockData` objects (like "base").  We already _sort of_ support this via the overload to `DataCollection::Add` that includes a list of names of variables, but in its current form this is of limited use in practice.  This PR attempts to make this much more flexible and useful.

The main features included in this PR are:
- New overloads to add `MeshBlockData` or `MeshData` objects to `DataCollection`s.  The preexisting versions that take (label, source) or (label, source, var_names) still exist and function as before.  In addition, an `AddShallow` function (with the same overloads) was added that yields a shallow copy, even for non-`Metadata::OneCopy` fields.
- It now works to `Add` directly to the `Mesh`-level `DataCollection`.  When `Add` is called with a `MeshData` source object, the new `MeshBlockData` objects that get created get added to the `MeshBlock` level `DataCollection`s.  The implementation is a bit of a hack, but it works for now and makes things much cleaner downstream.
- `StateDescriptor` now has an overloaded `GetVariablesNames` member function that returns a `std::vector<std::string>` of variable names given a list of names, `Metadata::FlagCollectoin`, and/or a list of sparse ids.  This is meant to produce the names that get passed into the `DataCollection::Add*` calls, so provides the means of creating new `Mesh*Data` objects that are filtered on, for example, `Metadata` and/or sparse ids.
- `GetVariableNames` overloads were also added to `Mesh`, which just forwards the call on to the `resolved_state` `StateDescriptor` functions.  This again is for convenience so that downstream codes don't have to pull out the `resolved_state` when they just want a list of names from the simulation state as a whole.
- Parthenon will now automatically add a `Metadata` flag with the same name as the `StateDescriptor` the field is added to every field, sparse pool, and swarm.  `StateDescriptor` stashes the new flag in a `Param` and provides a function to retrieve the flag for convenient access later.
- Given a list of names, a `Metadata::FlagCollection`, and/or sparse ids, there are new functions that additionally take two `MeshBlockData` pointers or two `MeshData` pointers and return the unique ids of the variables that satisfy the filters and exist in both objects.  This is required when trying to work with `Mesh*Data` objects that may have different fields present.  With just the intersection of unique ids, you can write functions that can, for example, add values from one `MeshData` object into the corresponding fields in another `MeshData` object quite straightforwardly.
- New `PackDescriptor` code that allows you to produce a `SparsePack` given a list of unique ids.

With all this stuff in place, you can do things like make registers for the RHSs of subsets of equations.  We have integrated this branch into our downstream code and successfully make use of all these features.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


jdolence:
@pdmullen @bprather @Yurlungur @lroberts36 @pgrete I think this one is ready to review.  I'm going to leave Draft in place until I do a bit more testing downstream, but so far things look good so I suspect I'll be removing Draft very soon.

src/amr_criteria/amr_criteria.hpp
```diff
@@ -17,12 +17,14 @@
 #include <string>
 
 #include "defs.hpp"
-#include "interface/meshblock_data.hpp"
+//#include "interface/meshblock_data.hpp"
```
lroberts36:
```suggestion
```

src/interface/data_collection.cpp
```diff
@@ -22,43 +22,54 @@
...
-                                          const std::shared_ptr<T> &src,
-                                          const std::vector<std::string> &flags) {
+std::shared_ptr<T> &
+DataCollection<T>::Add(const std::string &name, const std::shared_ptr<T> &src,
+                       const std::vector<std::string> &field_names, const bool shallow) {
   auto it = containers_.find(name);
   if (it != containers_.end()) {
-    if (!(it->second)->Contains(flags)) {
-      PARTHENON_THROW(name + "already exists in collection but does not contain flags");
+    if (!(it->second)->Contains(field_names)) {
```
lroberts36, line 30: I guess `T` has to implement `Contains` and `Copy` (and maybe some other stuff)? It might be useful to document this in the header or create a concept that `T` must adhere to.

src/interface/meshblock_data.cpp
```diff
@@ -114,54 +106,12 @@ void MeshBlockData<T>::CopyFrom(const MeshBlockData<T> &src, bool shallow_copy,
...
-        // otherwise we get an empty pool
-        const auto &sparse_pool = resolved_packages_->GetSparsePool(name);
-
-        // add all sparse ids of the pool
-        for (const auto iter : sparse_pool.pool()) {
-          // this variable must exist, if it doesn't something is very wrong
-          const auto &v = varMap_.at(MakeVarLabel(name, iter.first));
-          add_var(v);
-          found = true;
-        }
```
lroberts36, line 129: I guess this will now only copy full variable names and not sparse variables with the base name? That seems cleaner to me, but I guess this does change behavior?

src/interface/meshblock_data.hpp
```diff
@@ -97,33 +87,23 @@ class MeshBlockData {
...
-  /// from the given list of sparse ids.
-  ///
-  /// @param sparse_ids The list of sparse ids to include
-  /// @return New container with slices from all variables
-  std::shared_ptr<MeshBlockData<T>> SparseSlice(const std::vector<int> &sparse_ids) const;
+  /// Create copy of MeshBlockData, possibly with a subset of named fields,
+  /// and possibly shallow.  Note when shallow=false, new storage is allocated
+  /// for non-OneCopy vars, but the data from src is not actually deep copied
+  void Copy(const MeshBlockData<T> *src, const std::vector<std::string> &names,
+            const bool shallow);
```
lroberts36: I wonder if there is a better name than `Copy` for this method (since `Copy` implies that fields are copied if you don't read the comment). Maybe `CreateTwin` or something.

src/interface/metadata.hpp
```diff
@@ -510,6 +511,16 @@ class Metadata {
     return AllFlagsSet(FlagVec{flag, std::forward<Args>(args)...});
   }
 
+  template <template <class...> class Container_t, class... extra>
+  bool NoFlagsSet(const Container_t<MetadataFlag, extra...> &flags) const {
+    return std::none_of(flags.begin(), flags.end(),
+                        [this](MetadataFlag const &f) { return IsSet(f); });
+  }
```
lroberts36:
```suggestion
  template <class Container_t, 
            REQUIRES(std::is_same<Container_t::value_type, MetadataFlag>::value)>
  bool NoFlagsSet(const Container_t &flags) const {
    return std::none_of(flags.begin(), flags.end(),
                        [this](MetadataFlag const &f) { return IsSet(f); });
  }
```

src/interface/state_descriptor.cpp
```diff
@@ -443,4 +444,84 @@ StateDescriptor::CreateResolvedStateDescriptor(Packages_t &packages) {
...
+          names.insert(name);
+        }
+      }
+    }
+    // now handle sparse fields
+    for (const auto &p : sparsePoolMap_) {
+      auto &name = p.first;
+      auto &pool = p.second;
+      auto &meta = pool.shared_metadata();
+      printf("checking %s\n", name.c_str());
```
lroberts36: Debugging `printf`?

lroberts36:
```suggestion
```

src/interface/state_descriptor.cpp
```diff
@@ -443,4 +444,84 @@ StateDescriptor::CreateResolvedStateDescriptor(Packages_t &packages) {
...
+      if (MetadataUtils::MatchFlags(flags, meta)) {
+        printf("matched flag\n");
```
lroberts36:
```suggestion
```

src/interface/state_descriptor.cpp
```diff
@@ -443,4 +444,84 @@ StateDescriptor::CreateResolvedStateDescriptor(Packages_t &packages) {
   return state;
 }
 
+std::vector<std::string>
+StateDescriptor::GetVariableNames(const std::vector<std::string> &req_names,
+                                  const Metadata::FlagCollection &flags,
+                                  const std::vector<int> &sparse_ids) {
```
lroberts36: Maybe add a comment here or in header about how these requested (?) names, flags, and sparse ids are combined to get a list of variable names.

lroberts36 (approved):
LGTM, comments aren't blocking.

src/interface/data_collection.cpp
```diff
@@ -22,43 +22,54 @@
...
-
-  containers_[name] = c;
   return containers_[name];
 }
+template <typename T>
+std::shared_ptr<T> &DataCollection<T>::Add(const std::string &name,
+                                           const std::shared_ptr<T> &src,
+                                           const std::vector<std::string> &field_names) {
+  return Add(name, src, field_names, false);
+}
```
Yurlungur: Any reason this is implemented with an overload and not a default boolean for the above method?

src/interface/data_collection.cpp
```diff
@@ -22,43 +22,54 @@
...
+std::shared_ptr<T> &
+DataCollection<T>::AddShallow(const std::string &name, const std::shared_ptr<T> &src,
+                              const std::vector<std::string> &field_names) {
+  return Add(name, src, field_names, true);
+}
+template <typename T>
+std::shared_ptr<T> &DataCollection<T>::Add(const std::string &name,
+                                           const std::shared_ptr<T> &src) {
+  return Add(name, src, {}, false);
+}
```
Yurlungur: Same as above. Could be done with default argument.

src/interface/data_collection.cpp
```diff
@@ -22,43 +22,54 @@
...
+                                           const std::shared_ptr<T> &src,
+                                           const std::vector<std::string> &field_names) {
+  return Add(name, src, field_names, false);
+}
+template <typename T>
+std::shared_ptr<T> &
+DataCollection<T>::AddShallow(const std::string &name, const std::shared_ptr<T> &src,
+                              const std::vector<std::string> &field_names) {
+  return Add(name, src, field_names, true);
+}
```
Yurlungur: The two overloads of `AddShallow` could be combined with a default argument for `field_names`

src/interface/meshblock_data.cpp
Yurlungur: Was going to ask the same question. Any need to be able to copy some subset of a sparse variable in this way?

src/interface/meshblock_data.cpp
```diff
@@ -114,54 +106,12 @@ void MeshBlockData<T>::CopyFrom(const MeshBlockData<T> &src, bool shallow_copy,
...
-}
-
-// provides a container that has a single sparse slice
-template <typename T>
-std::shared_ptr<MeshBlockData<T>>
-MeshBlockData<T>::SparseSlice(const std::vector<int> &sparse_ids) const {
-  auto c = std::make_shared<MeshBlockData<T>>();
-  c->CopyFrom(*this, true, {}, {}, sparse_ids);
-  return c;
-}
```
Yurlungur, line 163: No constructor overloads, always use copy now? That's cleaner from a self-describing interface point of view I think

src/interface/meshblock_data.hpp
```diff
@@ -664,6 +659,14 @@ class MeshBlockData {
                                            vpack_types::VPackKey_t *key);
 };
 
+template <typename T, typename... Args>
+std::vector<Uid_t> UidIntersection(MeshBlockData<T> *mbd1, MeshBlockData<T> *mbd2,
+                                   Args &&...args) {
+  auto u1 = mbd1->GetVariableUIDs(std::forward<Args>(args)...);
+  auto u2 = mbd2->GetVariableUIDs(std::forward<Args>(args)...);
```
Yurlungur, line 666: Not that this matters much, but this code is identical between `MeshData` and `MeshBlockData`. I wonder if there's a way to reduce code duplication, either by templating on data type, or by having this version reference the `MeshBlockData` version?

src/utils/unique_id.cpp
```diff
@@ -11,9 +11,22 @@
...
+
+#include "unique_id.hpp"
 
 namespace parthenon {
 template class UniqueIDGenerator<std::string>;
+
+std::vector<Uid_t> UidIntersection(std::vector<Uid_t> v1, std::vector<Uid_t> v2) {
+  std::vector<Uid_t> vout;
+  std::sort(v1.begin(), v1.end());
+  std::sort(v2.begin(), v2.end());
```
Yurlungur, line 26: Is sorting necessary here?

src/interface/state_descriptor.hpp
```diff
@@ -100,11 +100,19 @@ class StateDescriptor {
   StateDescriptor(const StateDescriptor &s) = delete;
 
   // Preferred constructor
-  explicit StateDescriptor(std::string const &label) : label_(label) {}
+  explicit StateDescriptor(std::string const &label) : label_(label) {
+    if (Metadata::FlagNameExists(label)) {
+      AddParam("MetadataFlag", Metadata::GetUserFlag(label));
```
pdmullen: `MetadataFlag` is pretty ambiguous.  How about `MetadataPackageFlag` or similar?

Also, I don't see issue with it, but why are we going through `Param`s here? Is it only such that we can access the flag via "`MetadataFlag`" as follows
```c++
MetadataFlag GetMetadataFlag() { return params_.Get<MetadataFlag>("MetadataFlag"); }
```
hence bypassing the fact that the user flag actually takes on `label`?

Yurlungur (approved):
LGTM. This will be a nice improvement. Do you need changelog + formatting? Also remove the draft prefix?

pdmullen (approved):
LGTM!

src/interface/meshblock_data.cpp
jdolence: Yeah, this isn't built into the Copy function anymore.  Instead, calling the GetVariablesName function with the sparse base name expands into all (or the requested subset of) the associated sparse variables.  I could add this functionality back, but I suspect we're the only consumers of this at the moment and I've already moved us over to this branch.  I think the fewer places we can deal with base name -> expanded list the better, so this moves us in that direction.

src/interface/meshblock_data.hpp
jdolence: Not opposed to changing the name, but not sure to what.  It's kind of a weird operation this function does, or rather a weird set of operations.  Sometimes it allocates new space.  Sometimes it just copies pointers.  Sometimes it does both.  Sometimes on all fields.  Sometimes on a subset of fields.  It never actually copies any data, so it's not a great name.  Certainly open to suggestions for names that better capture what this monstrosity really does.

lroberts36: `Replicate`, `MakeTwin`. I think the thing is not to have the name necessarily explain what the function does, but rather have it not be something common like `Copy` that you read inline and automatically assume some behavior.

src/utils/unique_id.cpp
jdolence: `std::set_intersection` requires sorted arguments or the behavior is undefined.

src/interface/meshblock_data.hpp
jdolence: How about an overload of `Initialize`.  Or `InitFrom`.  or something like that...

lroberts36: An `Initialize` overload sounds good to me.

jdolence:
@pgrete I think this one is ready to go.  Did you want to look through it first or should I just pull the trigger.  I'm pretty sure this won't break anything for you downstream.

src/amr_criteria/amr_criteria.hpp
jdolence: had already gotten this one.

c-prather:
After pulling and fiddling a bunch I like all this as is too.

It's prompted me to dump some more of the other functionality I was duplicating in KHARMA into Parthenon too, I'll open a separate PR with the additions.

src/interface/data_collection.cpp
jdolence: just added a comment in the header and a TODO to create the concept.

src/interface/meshblock_data.hpp
jdolence: Done

src/interface/metadata.hpp
jdolence: Added this, and also updated the other *FlagsSet functions accordingly.

src/interface/state_descriptor.cpp
jdolence: thanks, got it.

src/interface/state_descriptor.cpp
jdolence: Done

src/interface/data_collection.cpp
jdolence: reduced this down to one `Add` that actually does something and requires all arguments, one `Add` that has a default empty `std::vector<std::string>` and just calls the first `Add`, and one `AddShallow`.  Good enough?

src/interface/data_collection.cpp
jdolence: see above

src/interface/data_collection.cpp
jdolence: see above

src/interface/meshblock_data.hpp
jdolence: I did the latter.

src/interface/state_descriptor.hpp
jdolence: Updated the name.  Used `Param`s because it was convenient.  We use it for everything else after all.  Wanted the function that returns the flag in order to reduce the number of times anyone had to put things between double quotes that could introduce bugs via typos.

src/interface/data_collection.cpp
Yurlungur: Yep I'm happy with that.

src/utils/unique_id.cpp
Yurlungur: Ah ok. my bad.

jdolence:
@pgrete didn't give you much time to respond, but I'm going to pull the trigger.  We have a least two separate codes that have confirmed backwards compatibility, so I'm pretty confident this won't break anything for you.  Obviously, if you have comments on anything in this after it's merged, we can work on getting them addressed.

PR #925, opened 2023-08-17
Quick fixes

jdolence:
## PR Summary

Fix a typo in the previous merged PR, fix bug in FlagCollection where `Empty()` did not check if `exclusions_` was empty, and silence a compiler warning related to no return statement in a non-void function.


jdolence:
Just a few quick fixes.  I'm going to merge as soon as tests pass.

lroberts36 (approved):
LGTM

pdmullen (approved):

pgrete (approved):

pgrete:
@jdolence I only now got to have a look at the merged PR.
Definitely a win in usability.
For clarification I have two conceptual questions (that might be added here in the docs):

1. Naming: Does `GetMetadataFlag()` really "return a ``MetadataFlag`` that is  automatically added to..."? Wouldn't in that case `GetOrAdd` or sth similar make more sense in the context of existing names? I'm asking because I'm trying to make sense of the following line `if (!newm.IsSet(GetMetadataFlag())) newm.Set(GetMetadataFlag());` that in similar fashion was added in multiple places. Shouldn't the first `Get...` in the condition already take care of it? Or maybe I didn't fully understood the entire concept.

2. For `AddShallow`: Which container/collection does the shallow copy point to? Is there any way I could accidentally create a data collection that contains a variable (that is allocated) from two containers and then make a shallow copy of that?

jdolence:
> @jdolence I only now got to have a look at the merged PR. Definitely a win in usability. For clarification I have two conceptual questions (that might be added here in the docs):
> 
>     1. Naming: Does `GetMetadataFlag()` really "return a `MetadataFlag` that is  automatically added to..."? Wouldn't in that case `GetOrAdd` or sth similar make more sense in the context of existing names? I'm asking because I'm trying to make sense of the following line `if (!newm.IsSet(GetMetadataFlag())) newm.Set(GetMetadataFlag());` that in similar fashion was added in multiple places. Shouldn't the first `Get...` in the condition already take care of it? Or maybe I didn't fully understood the entire concept.
> 
>     2. For `AddShallow`: Which container/collection does the shallow copy point to? Is there any way I could accidentally create a data collection that contains a variable (that is allocated) from two containers and then make a shallow copy of that?

For the first question: When the `StateDescriptor` object is created, a new "user-defined" metadata flag is created.  It takes on the same name as the label of the `StateDescriptor`.  You can retrieve that flag at any time by calling the `StateDescriptor`'s `GetMetadataFlag()` function.  Then as you add fields/sparse pools/swarms to that `StateDescriptor`, this new flag is appended to the metadata that is provided by the caller.  So as far as downstream codes are concerned, calling `GetMetadataFlag()` on a `StateDescriptor` returns a metadata flag that we guarantee is tied to every variable registered in that package.  Does that address your question/confusion?

For the second question:  I'm not sure I completely understand the scenario you describe.  In general, the shallow copied variables will point at whatever variables are in the source `MeshData`/`MeshBlockData` object passed into the `AddShallow` function.  Maybe you can clarify what you mean?

pgrete:
> > @jdolence I only now got to have a look at the merged PR. Definitely a win in usability. For clarification I have two conceptual questions (that might be added here in the docs):
> > ```
> > 1. Naming: Does `GetMetadataFlag()` really "return a `MetadataFlag` that is  automatically added to..."? Wouldn't in that case `GetOrAdd` or sth similar make more sense in the context of existing names? I'm asking because I'm trying to make sense of the following line `if (!newm.IsSet(GetMetadataFlag())) newm.Set(GetMetadataFlag());` that in similar fashion was added in multiple places. Shouldn't the first `Get...` in the condition already take care of it? Or maybe I didn't fully understood the entire concept.
> > 
> > 2. For `AddShallow`: Which container/collection does the shallow copy point to? Is there any way I could accidentally create a data collection that contains a variable (that is allocated) from two containers and then make a shallow copy of that?
> > ```
> 
> For the first question: When the `StateDescriptor` object is created, a new "user-defined" metadata flag is created. It takes on the same name as the label of the `StateDescriptor`. You can retrieve that flag at any time by calling the `StateDescriptor`'s `GetMetadataFlag()` function. Then as you add fields/sparse pools/swarms to that `StateDescriptor`, this new flag is appended to the metadata that is provided by the caller. So as far as downstream codes are concerned, calling `GetMetadataFlag()` on a `StateDescriptor` returns a metadata flag that we guarantee is tied to every variable registered in that package. Does that address your question/confusion?

Ah, yes, thanks. That makes sense. I interpreted the "automatically generated" as happening inside the `Get` function (rather than the `Get` function just returning that is automatically generated somewhere else and thus available by default.

> For the second question: I'm not sure I completely understand the scenario you describe. In general, the shallow copied variables will point at whatever variables are in the source `MeshData`/`MeshBlockData` object passed into the `AddShallow` function. Maybe you can clarify what you mean?

My bad. I think conceptually, I was wondering about sth like putting the a variable into two containers, make a compound container, and then a shallow copy of that, but I just realized we don't have compound containers (I confused those with added variable through different flags, names, etc to a single container).

PR #928, opened 2023-08-22
Fix refinement next to refined blocks

### LINKED ISSUE #927: Parthenon does not respect ownership of shared prolongated elements
I've implemented a face-field CT scheme in KHARMA based on the new Parthenon non-cell AMR, and ran across a really subtle issue.  After chatting with @lroberts36 we think it would affect any simulation in terms of accuracy, but it especially affects face-field CT as it causes a magnetic field divergence to develop.  I think it's worth documenting here before I put in the fix, as I'm not sure I caught every case, and it's a universal issue, not to mention illustrative of the ownership model.

What's weirdest here is that the symptom appears the *second* block refinement.  That is, one full refinement and several subsequent steps later, we would have no magnetic field divergence (this is KHARMA's divB variable plotted with movie2d.py):

Then, as a part of the next re-mesh, so long as it involved refining in a block adjacent to the first, we'd pick up a nasty divergence:

The bug here is that the shared prolongation operator does not inherently respect the ownership model.  Refining, say, a face to level 2 , we end up writing the outer faces of each of our level-2 blocks, even if another block at refinement 2 might own those faces according to our model.  In order to have up-to-date outer faces, we need to be pulling them from the block which actually owns them, not restricting them and using the restriction results.

Normally this would only be an accuracy issue, as the prolongated values decently resemble the true values, and in any case are almost immediately replaced by those values on the next MPI sync.  However, for constrained transport it's crucial: the internal prolongation operator in a CT scheme must know the exact inner and outer face values *on the fine grid* in order to choose interior field values which will remain divergence-free.  If it chooses values based on the prolongated face, but these values are then immediately replaced by the true owner's values, the field gains a divergence.

Unfortunately, the ownership hierarchy needed for filling new buffers isn't the same as the hierarchy used for normal MPI syncs either, as brand-new blocks must be placed according to their parent/originating data, not their current level.  So, some reordering has to be done, and a list kept of new "untrustworthy" blocks to demote on the first (and only the first) MPI synchronization after re-meshing.

c-prather:
## PR Summary

Fixes #927

Not attached to the carry-around-the-list interface here, or the cheeky double-lambda.

Note that the issue this fixes is really two issues:
1. There must be an MPI sync between ProlongateShared and ProlongateInternal.  This is the primary error and easy to change.
2. During such an MPI sync, one should not prioritize zones from the new block, which were generated by prolongation, over "naturally" finer zones in a previously-refined block.

Most of the code is aimed at dealing with (2), which requires determining and then syncing according to a special priority order, namely:
1. any neighbor blocks at the new refinement level of this block
2. this block
3. any neighbor blocks at the parent refinement level

That is, freshly-refined blocks should own any boundaries with coarser blocks, but none of their boundaries with previously-refined blocks (and should still choose by Morton number among themselves).

Still working on getting this entirely correct, (2) and (3) are muddy in the code as posted.

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build

### DIFF
--- a/src/mesh/amr_loadbalance.cpp
+++ b/src/mesh/amr_loadbalance.cpp
@@ -698,6 +698,7 @@ void Mesh::RedistributeAndRefineMeshBlocks(ParameterInput *pin, ApplicationInput
     oldtonew[mb_idx] = ntot - 1;
 
   current_level = 0;
+  std::unordered_set<LogicalLocation> newly_refined;
   for (int n = 0; n < ntot; n++) {
     // "on" = "old n" = "old gid" = "old global MeshBlock ID"
     int on = newtoold[n];
@@ -705,6 +706,10 @@ void Mesh::RedistributeAndRefineMeshBlocks(ParameterInput *pin, ApplicationInput
       current_level = newloc[n].level();
     if (newloc[n].level() >= loclist[on].level()) { // same or refined
       newcost[n] = costlist[on];
+      // Keep a list of all blocks refined for below
+      if (newloc[n].level() > loclist[on].level()) {
+        newly_refined.insert(newloc[n]);
+      }
     } else {
       double acost = 0.0;
       for (int l = 0; l < nleaf; l++)
@@ -789,7 +794,6 @@ void Mesh::RedistributeAndRefineMeshBlocks(ParameterInput *pin, ApplicationInput
   RegionSize block_size = GetBlockSize();
 
   BlockList_t new_block_list(nbe - nbs + 1);
-  std::set<LogicalLocation> newly_refined;
   for (int n = nbs; n <= nbe; n++) {
     int on = newtoold[n];
     if ((ranklist[on] == Globals::my_rank) &&
@@ -811,7 +815,6 @@ void Mesh::RedistributeAndRefineMeshBlocks(ParameterInput *pin, ApplicationInput
       new_block_list[n - nbs] =
           MeshBlock::Make(n, n - nbs, newloc[n], block_size, block_bcs, this, pin, app_in,
                           packages, resolved_packages, gflag);
-      if (newloc[n].level() > loclist[on].level()) newly_refined.insert(newloc[n]);
     }
   }
 
@@ -936,6 +939,14 @@ void Mesh::RedistributeAndRefineMeshBlocks(ParameterInput *pin, ApplicationInput
     pmb->pbval->SearchAndSetNeighbors(tree, ranklist.data(), nslist.data(),
                                       newly_refined);
   }
+  // Make sure all old sends/receives are done before we reconfigure the mesh
+#ifdef MPI_PARALLEL
+  if (send_reqs.size() != 0)
+    PARTHENON_MPI_CHECK(
+        MPI_Waitall(send_reqs.size(), send_reqs.data(), MPI_STATUSES_IGNORE));
+#endif
+  // Re-initialize the mesh with our temporary ownership/neighbor configurations.
+  // No buffers are different when we switch to the final precedence order.
   Initialize(false, pin, app_in);
 
   // Internal refinement relies on the fine shared values, which are only consistent after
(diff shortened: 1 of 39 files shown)


c-prather:
@par-hermes format

src/mesh/mesh.cpp
```diff
@@ -475,11 +475,6 @@ Mesh::Mesh(ParameterInput *pin, ApplicationInput *app_in, Packages_t &packages,
   }
 
   ResetLoadBalanceVariables();
-
-  // Output variables in use in this run
-  if (Globals::my_rank == 0) {
-    std::cout << "#Variables in use:\n" << *(resolved_packages) << std::endl;
-  }
```
Yurlungur: Isn't this included in another PR? Did that not get merged?

src/mesh/logical_location.hpp
```diff
@@ -251,15 +251,25 @@ struct block_ownership_t {
...
-  auto ownership_less_than = [](const LogicalLocation &a, const LogicalLocation &b) {
+  auto ownership_level = [&](const LogicalLocation &a) {
+    // Newly-refined blocks are treated as higher-level than blocks at their
+    // parent level, but lower-level than previously-refined blocks at their
+    // current level.
+    if (newly_refined.count(a)) return 2 * a.level() - 1;
+    return 2 * a.level();
+  };
+
+  auto ownership_less_than = [ownership_level](const LogicalLocation &a,
```
Yurlungur, line 279: still cheeky IMO

Yurlungur (approved):

Yurlungur:
Check boxes needed :)

Not for this PR, but we really need a regression test

lroberts36:
> Not for this PR, but we really need a regression test

Yeah, we need to write a vector wave equation example I think.

src/mesh/amr_loadbalance.cpp
```diff
@@ -913,29 +915,47 @@ void Mesh::RedistributeAndRefineMeshBlocks(ParameterInput *pin, ApplicationInput
...
+    pmb->pbval->SearchAndSetNeighbors(tree, ranklist.data(), nslist.data());
+  }
+
+#ifdef MPI_PARALLEL
+  if (send_reqs.size() != 0)
+    PARTHENON_MPI_CHECK(
+        MPI_Waitall(send_reqs.size(), send_reqs.data(), MPI_STATUSES_IGNORE));
+#endif
+  Kokkos::Profiling::popRegion(); // AMR: Recv data and unpack
+
```
lroberts36: I *think* this needs to come before `Initialize`, otherwise the buffers being used by the MPI sends may be freed before the communication is complete.

src/mesh/amr_loadbalance.cpp
```diff
@@ -789,6 +789,7 @@ void Mesh::RedistributeAndRefineMeshBlocks(ParameterInput *pin, ApplicationInput
   RegionSize block_size = GetBlockSize();
 
   BlockList_t new_block_list(nbe - nbs + 1);
+  std::set<LogicalLocation> newly_refined;
```
lroberts36: I think we may want this to be a `std::unordered_set` for performance reasons.

lroberts36 (commented):
Overall this looks good to me, but one potential bug and one potential performance enhancement below.

src/mesh/mesh.cpp
c-prather: No, that one's still in queue.  I pinged some people, probably it will still go in before this one and I can avoid having to correct my mistake here.

src/mesh/logical_location.hpp
c-prather: As in, I should collapse it?  This seemed pretty easy to understand, if a little over-engineered

src/mesh/amr_loadbalance.cpp
c-prather: Agreed, I think, the list could potentially be long and we're checking membership.
What's a good hash function for the unordered_set?  Just morton number?

Speaking of that, I think I populate this list incorrectly, since I only pull from `nbs` to `nbe` of the global block list in populating newly_refined, meaning I need to collate the lists from all ranks.  So, also going to fix that.

src/mesh/amr_loadbalance.cpp
c-prather: Yep, will fix.  No harm in moving it up

src/mesh/logical_location.hpp
Yurlungur: no no---it's fine. Cheeky is not a bad thing.

Yurlungur: no no---it's fine. Cheeky is not a bad thing.

src/mesh/amr_loadbalance.cpp
lroberts36: A hash for `LogicalLocation` is injected into the `std::` namespace, so `std::unordered_map<LogicalLocation>` should just work. 

Good catch. Yeah, I think you need the locations on all ranks.

c-prather:
Okay, as updated the PR now gathers new refinements by comparing the global new & old lists of blocks (I think.  The variable naming wasn't completely clear to me in an MPI context and there are no relevant comments).  Testing with multiple MPI ranks now.

Also switched to `unordered_map`, which required a bit of reordering in the `LogicalLocation` header.  The class definition for `LogicalLocation` contains one member function implementation which uses an `unordered_set<LogicalLocation>`, which relies on `std::hash<LogicalLocation>`, which relies on... the class definition of `LogicalLocation`.  Solution:
1. Declare class `LogicalLocation`
2. Declare `std::hash<LogicalLocation>`
3. Implement `LogicalLocation`
4. Implement `std::hash<LogicalLocation>`

Pretty sure this is the simple way to solve it, but maybe I'm missing something.

lroberts36 (approved):
LGTM, thanks for switching over to `std::unordered_set`. I think your solution is the right one.

c-prather:
Would love to just merge this in the next few days if there aren't objections

Yurlungur:
I'm enabling auto merge

PR #930, opened 2023-08-24
remove ParthenonInit and add docs

### LINKED ISSUE #929: Does anyone use `ParthenonInit`
For historical reasons, we have two ways of setting up Parthenon's runtime:
1. First call `ParthenonInitEnv` and then call` ParthenonInitPackagesAndMesh`. The former sets up MPI and Kokkos, and reads in the command line arguments and input deck. The latter calls all the `Initilalize` function for all packages and calls the mesh constructor.
2. `ParthenonInit` simply calls `ParthenonInitEnv` and `ParthenonInitPackagesAndMesh` one after the other.

It's easy to confuse `ParthenonInit` and `ParthenonInitPackagesAndMesh` by writing code like:
```C++
ParthenonManager pman;
pman.ParthenonInit(argc, argv);

// do some code, set app inputs

pman.ParthenonInitPackagesAndMesh();
```
where the intended code was actually

```C++
ParthenonManager pman;
pman.ParthenonInitEnv(argc, argv);

// do some code, set app inputs

pman.ParthenonInitPackagesAndMesh();
```

This seems error prone. Does anyone actually use the older, unified `ParthenonInit` function? If not, perhaps we should just remove it. 

Alternatively, we could add error checking so that if any of the `ParthenonInit*` functions is called twice by the same manager object, an error is raised.

Yurlungur:
## PR Summary

Based on the discussion in #929 , it seems like it might be a good idea to remove the ``ParthenonManager::ParthenonInit`` function. This PR does so. I also add documentation on ``ParthenonManager``, which I couldn't otherwise find.

 **This is a breaking change.** I would like the go-ahead from downstream codes before merging this.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/src/parthenon_manager.cpp
+++ b/src/parthenon_manager.cpp
@@ -40,16 +40,12 @@
 
 namespace parthenon {
 
-ParthenonStatus ParthenonManager::ParthenonInit(int argc, char *argv[]) {
-  auto manager_status = ParthenonInitEnv(argc, argv);
-  if (manager_status != ParthenonStatus::ok) {
-    return manager_status;
-  }
-  ParthenonInitPackagesAndMesh();
-  return ParthenonStatus::ok;
-}
-
 ParthenonStatus ParthenonManager::ParthenonInitEnv(int argc, char *argv[]) {
+  if (called_init_env_) {
+    PARTHENON_THROW("ParthenonInitEnv called twice!");
+  }
+  called_init_env_ = true;
+
   // initialize MPI
 #ifdef MPI_PARALLEL
   if (MPI_SUCCESS != MPI_Init(&argc, &argv)) {
@@ -146,6 +142,11 @@ ParthenonStatus ParthenonManager::ParthenonInitEnv(int argc, char *argv[]) {
 }
 
 void ParthenonManager::ParthenonInitPackagesAndMesh() {
+  if (called_init_packages_and_mesh_) {
+    PARTHENON_THROW("Called ParthenonInitPackagesAndMesh twice!");
+  }
+  called_init_packages_and_mesh_ = true;
+
   // Allow for user overrides to default Parthenon functions
   if (app_input->ProcessPackages != nullptr) {
     ProcessPackages = app_input->ProcessPackages;
--- a/src/parthenon_manager.cpp
+++ b/src/parthenon_manager.cpp
@@ -1,9 +1,9 @@
 //========================================================================================
 // Parthenon performance portable AMR framework
-// Copyright(C) 2020-2022 The Parthenon collaboration
+// Copyright(C) 2020-2023 The Parthenon collaboration
 // Licensed under the 3-clause BSD License, see LICENSE file for details
 //========================================================================================
-// (C) (or copyright) 2020-2022. Triad National Security, LLC. All rights reserved.
+// (C) (or copyright) 2020-2023. Triad National Security, LLC. All rights reserved.
 //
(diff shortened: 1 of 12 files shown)


src/parthenon_manager.cpp
```diff
@@ -40,16 +40,12 @@
 
 namespace parthenon {
 
-ParthenonStatus ParthenonManager::ParthenonInit(int argc, char *argv[]) {
-  auto manager_status = ParthenonInitEnv(argc, argv);
-  if (manager_status != ParthenonStatus::ok) {
-    return manager_status;
+ParthenonStatus ParthenonManager::ParthenonInitEnv(int argc, char *argv[]) {
+  if (called_init_env_) {
+    PARTHENON_THROW("ParthenonInitEnv called twice!");
```
lroberts36, line 45: nice.

src/parthenon_manager.cpp
```diff
@@ -146,6 +142,11 @@ ParthenonStatus ParthenonManager::ParthenonInitEnv(int argc, char *argv[]) {
 }
 
 void ParthenonManager::ParthenonInitPackagesAndMesh() {
+  if (called_init_packages_and_mesh_) {
+    PARTHENON_THROW("Called ParthenonInitPackagesAndMesh twice!");
+  }
+  called_init_packages_and_mesh_ = true;
+
```
lroberts36, line 149: nice.

lroberts36 (approved):
Thanks, LGTM.

c-prather (approved):
LGTM.  I'd have done static vars with function scope but this is more flexible.  Could see e.g. erroring in driver.Execute if init wasn't called, or other hand-holding like that if there are common errors.  It also allows multiple `ParthenonManager`s I guess, though I'm not sure MPI & Kokkos would react well to that anyway.

For my own edification: are there other reasons to prefer members?

Yurlungur:
> LGTM. I'd have done static vars with function scope but this is more flexible. Could see e.g. erroring in driver.Execute if init wasn't called, or other hand-holding like that if there are common errors. It also allows multiple `ParthenonManager`s I guess, though I'm not sure MPI & Kokkos would react well to that anyway.
> 
> For my own edification: are there other reasons to prefer members?

To be honest, I didn't think about it very hard. Members seemed natural and easy to reason about, since the object is already stateful. If people prefer statics, that's fine with me.

Yurlungur:
Ping someone from AthenaPK @BenWibking @forrestglines @pgrete . Can we merge this in?

src/parthenon_manager.cpp
```diff
@@ -40,16 +40,12 @@
...
   }
-  ParthenonInitPackagesAndMesh();
-  return ParthenonStatus::ok;
-}
+  called_init_env_ = true;
 
-ParthenonStatus ParthenonManager::ParthenonInitEnv(int argc, char *argv[]) {
   // initialize MPI
 #ifdef MPI_PARALLEL
   if (MPI_SUCCESS != MPI_Init(&argc, &argv)) {
```
forrestglines, line 51: At some point we'd like to optionally start with`MPI_Init_thread` which is apparently required by HDF5 subfiling. Another time and PR though

forrestglines (approved):
I can confirm that this PR works for AthenaPK (Had some unrelated trouble with ctest on Frontier since apparently the python script couldn't find the generated outputs in the filesystem).

PR #931, opened 2023-08-25
Pack on a subset of blocks in a MeshData object

jdolence:
## PR Summary
This PR includes three things:
- It slightly expands the capability of `SparsePacks` by adding a new optional argument to `GetPack` that is a `std::vector<bool>` of flags that determine which blocks in a `MeshData` object actually get packed.
- A new `MetadataFlag` was added, `Metadata::AllocOnNewBlocks`, intended to force allocation of sparse fields whenever a new `MeshBlock` is created through AMR/load balancing.
- A refactor of the `DataCollection` add functionality so that when `Add` is called on a `MeshData` object, it appropriately calls `Add` on the block-level `DataCollection`s.  The benefit of this is that it can avoid having to always make entirely new `MeshBlockData` objects, instead just using the previously created objects when they're available.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/src/interface/mesh_data.cpp
+++ b/src/interface/mesh_data.cpp
@@ -10,8 +10,9 @@
 // license in this material to reproduce, prepare derivative works, distribute copies to
 // the public, perform publicly and display publicly, and to permit others to do so.
 //========================================================================================
-#include "mesh/mesh.hpp"
 #include "mesh_data.hpp"
+
+#include "mesh/mesh.hpp"
 
 namespace parthenon {
 
--- a/src/interface/mesh_data.cpp
+++ b/src/interface/mesh_data.cpp
@@ -1,0 +1,35 @@
+//========================================================================================
+// (C) (or copyright) 2020-2023. Triad National Security, LLC. All rights reserved.
+//
+// This program was produced under U.S. Government contract 89233218CNA000001 for Los
+// Alamos National Laboratory (LANL), which is operated by Triad National Security, LLC
+// for the U.S. Department of Energy/National Nuclear Security Administration. All rights
+// in the program are reserved by Triad National Security, LLC, and the U.S. Department
+// of Energy/National Nuclear Security Administration. The Government is granted for
+// itself and others acting on its behalf a nonexclusive, paid-up, irrevocable worldwide
+// license in this material to reproduce, prepare derivative works, distribute copies to
+// the public, perform publicly and display publicly, and to permit others to do so.
+//========================================================================================
+#include "mesh_data.hpp"
+#include "mesh/mesh.hpp"
+
+namespace parthenon {
+
+template <typename T>
+void MeshData<T>::Initialize(const MeshData<T> *src,
+                             const std::vector<std::string> &names, const bool shallow) {
+  if (src == nullptr) {
+    PARTHENON_THROW("src points at null");
+  }
+  pmy_mesh_ = src->GetParentPointer();
+  const int nblocks = src->NumBlocks();
+  block_data_.resize(nblocks);
+  for (int i = 0; i < nblocks; i++) {
+    pmy_mesh_->block_list[i]->meshblock_data.Add(stage_name_, src->GetBlockData(i), names,
+                                                 shallow);
+  }
+}
+
+template class MeshData<Real>;
+
+} // namespace parthenon
--- a/src/interface/mesh_data.cpp
+++ b/src/interface/mesh_data.cpp
@@ -25,8 +25,8 @@ void MeshData<T>::Initialize(const MeshData<T> *src,
   const int nblocks = src->NumBlocks();
   block_data_.resize(nblocks);
   for (int i = 0; i < nblocks; i++) {
-    pmy_mesh_->block_list[i]->meshblock_data.Add(stage_name_, src->GetBlockData(i), names,
-                                                 shallow);
+    block_data_[i] = pmy_mesh_->block_list[i]->meshblock_data.Add(
+        stage_name_, src->GetBlockData(i), names, shallow);
   }
 }
 
--- a/src/interface/mesh_data.cpp
+++ b/src/interface/mesh_data.cpp
@@ -10,8 +10,8 @@
 // license in this material to reproduce, prepare derivative works, distribute copies to
 // the public, perform publicly and display publicly, and to permit others to do so.
 //========================================================================================
-#include "mesh_data.hpp"
 #include "mesh/mesh.hpp"
+#include "mesh_data.hpp"
 
 namespace parthenon {
 
--- a/src/interface/meshblock_data.cpp
+++ b/src/interface/meshblock_data.cpp
@@ -54,7 +54,7 @@ void MeshBlockData<T>::Initialize(
     AddField(q.first.base_name, q.second, q.first.sparse_id);
   }
 
-  Metadata::FlagCollection flags({Metadata::Sparse, Metadata::AllocOnNewBlocks});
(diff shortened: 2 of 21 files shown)


src/interface/mesh_data.cpp
```diff
@@ -0,0 +1,35 @@
...
+  pmy_mesh_ = src->GetParentPointer();
+  const int nblocks = src->NumBlocks();
+  block_data_.resize(nblocks);
+  for (int i = 0; i < nblocks; i++) {
+    block_data_[i] = pmy_mesh_->block_list[i]->meshblock_data.Add(
+        stage_name_, src->GetBlockData(i), names, shallow);
+  }
+}
+
+template class MeshData<Real>;
```
Yurlungur, line 34: I *think* this should help with compile times.

src/interface/meshblock_data.cpp
```diff
@@ -53,6 +53,12 @@ void MeshBlockData<T>::Initialize(
   for (auto const &q : resolved_packages->AllFields()) {
     AddField(q.first.base_name, q.second, q.first.sparse_id);
   }
+
+  Metadata::FlagCollection flags({Metadata::Sparse, Metadata::AllocOnNewBlocks});
+  auto vars = GetVariablesByFlag(flags);
+  for (auto &v : vars.vars()) {
+    AllocateSparse(v->label());
+  }
```
Yurlungur: Nice

src/interface/swarm_container.hpp
```diff
@@ -43,7 +43,8 @@ class SwarmContainer {
   // Public Methods
   //-----------------
   // Constructor does nothing
-  SwarmContainer() {}
+  SwarmContainer() = default;
+  explicit SwarmContainer(const std::string &name) : swarm_name_(name) {}
```
Yurlungur, line 47: Not that I object, but why this change?

src/interface/sparse_pack.hpp
```diff
@@ -131,8 +131,8 @@ class SparsePack : public SparsePackBase {
     // accessed on device via instance of types in the type list Ts...
     // The pack will be created and accessible on the device
     template <class T>
-    SparsePack GetPack(T *pmd) const {
-      return SparsePack(SparsePackBase::GetPack(pmd, *this));
+    SparsePack GetPack(T *pmd, const std::vector<bool> &include_block = {}) const {
+      return SparsePack(SparsePackBase::GetPack(pmd, *this, include_block));
```
Yurlungur, line 135: Is the `include_block` functionality expected to be used in userspace? Or do you plan to have some mechanism for automatically choosing which blocks to include?

Yurlungur (commented):
Should `AllocOnNewBlocks` maybe be called `ForceAllocOnNewBlocks`?

Should probably test that the `include_block` stuff works in the unit tests. Also missing changelog. ;)

Yurlungur (approved):
Just going to approve now, so I don't forget, pending tests, changelog, formatting, etc.

src/interface/sparse_pack.hpp
jdolence: user space.  i think that's a lot more flexible for downstream codes.  i don't presume to know how someone may eventually use this, beyond the use case we have.  and for us, there is essentially zero burden to passing this in.

src/interface/swarm_container.hpp
jdolence: this was part of refactoring how the `Add` functionality works with `MeshData`.  all container like objects now have names which can be passed into their constructors.

src/interface/sparse_pack.hpp
Yurlungur: Yeah ok fair enough

jdolence:
> Should `AllocOnNewBlocks` maybe be called `ForceAllocOnNewBlocks`?
> 
> Should probably test that the `include_block` stuff works in the unit tests. Also missing changelog. ;)

I thought about `ForceAlloc...` vs `Alloc...`.  `Force` just feels really long and I decided it didn't really say anything new.  I don't feel strongly about this though, so I can change it if there's consensus around it.

Yurlungur:
> I thought about `ForceAlloc...` vs `Alloc...`. `Force` just feels really long and I decided it didn't really say anything new. I don't feel strongly about this though, so I can change it if there's consensus around it.

IMO `Alloc...` is confusing as it implies that the default behavior would be to never allocate... whereas the real default behavior is to allocate sometimes.

src/interface/meshblock_data.cpp
```diff
@@ -53,6 +53,12 @@ void MeshBlockData<T>::Initialize(
   for (auto const &q : resolved_packages->AllFields()) {
     AddField(q.first.base_name, q.second, q.first.sparse_id);
   }
+
+  Metadata::FlagCollection flags({Metadata::Sparse, Metadata::ForceAllocOnNewBlocks});
+  auto vars = GetVariablesByFlag(flags);
+  for (auto &v : vars.vars()) {
+    AllocateSparse(v->label());
+  }
```
lroberts36, line 61: Is the `Metadata::Sparse` required? Shouldn't that flag have to be set for `Metadata::ForceAllocOnNewBlocks` to be set?

src/mesh/amr_loadbalance.cpp
```diff
@@ -134,7 +134,9 @@ bool TryRecvCoarseToFine(int lid_recv, int send_rank, const LogicalLocation &fin
             });
       }
     } else {
-      if (pmb->IsAllocated(var->label())) pmb->DeallocateSparse(var->label());
+      if (pmb->IsAllocated(var->label()) &&
+          !var->metadata().IsSet(Metadata::ForceAllocOnNewBlocks))
+        pmb->DeallocateSparse(var->label());
```
lroberts36, line 139: Why no change for `TryRecvFineToCoarse`?

src/mesh/amr_loadbalance.cpp
```diff
@@ -296,7 +298,9 @@ bool TryRecvSameToSame(int lid_recv, int send_rank, Variable<Real> *var, MeshBlo
       pmb->pmr->DereferenceCount() = counter_subview_h(0);
       var->dealloc_count = counter_subview_h(1);
     } else {
-      if (pmb->IsAllocated(var->label())) pmb->DeallocateSparse(var->label());
+      if (pmb->IsAllocated(var->label()) &&
+          !var->metadata().IsSet(Metadata::ForceAllocOnNewBlocks))
+        pmb->DeallocateSparse(var->label());
```
lroberts36, line 303: Should there be an explicit allocation here if `Metadata::ForceAllocOnNewBlocks` is set?

lroberts36 (approved):
LGTM

src/mesh/amr_loadbalance.cpp
jdolence: That's handled in `MeshBlockData::Initialize`

src/interface/meshblock_data.cpp
jdolence: As it's written right now, `Metadata::ForceAllocOnNewBlocks` has no effect on any variable that doesn't have `Metadata::Sparse`.  Those fields will always be allocated anyway, of course.  I could add some kind of check that'll throw if you set `Metadata::ForceAllocOnNewBlocks` for a dense variable, but I didn't because you still get the behavior you expect.

src/mesh/amr_loadbalance.cpp
jdolence: `DeallocateSparse` isn't called there.  Not sure if it should be.  It's actually not totally clear to me why it needs to be called in the other cases.  Maybe you know the answer to that?

PR #937, opened 2023-09-06
Fix parsing files with multiple line continuations

### LINKED ISSUE #936: `ParameterInput` does not support multiple line continuations
I had occasion to allow writing parameter files with some complex outputs, e.g.
```
<parthenon/output0>
file_type = hdf5
dt = 10.0
single_precision_output = true
variables = prims.rho, prims.u, prims.uvec, prims.B, &
            prims.Ktot

<parthenon/output1>
file_type = hdf5
dt = 1e20
single_precision_output = false
variables = coords.Xnative, coords.Xsph, &
            coords.gcon, coords.gcov, coords.gdet, &
            coords.lapse, coords.conn
```

Which `ParameterInput` parsed sagely as the following:

```
<parthenon/output0>
file_type               = hdf5                                               
dt                      = 10.0                                               
single_precision_output = true                                               
variables               = prims.rho, prims.u, prims.uvec, prims.B,prims.Ktot 
<parthenon/output1>
file_type               = hdf5                                                                                                                                          
dt                      = 1e20                                                                                                                                          
single_precision_output = false                                                                                                                                         
variablesvariables      = prims.rho, prims.u, prims.uvec, prims.B,prims.Ktotcoords.Xnative, coords.Xsph,coords.gcon, coords.gcov, coords.gdet,coords.lapse, coords.conn
```

Appears we're not clearing some static variables where we should!  I'm hunting it down but figured people should know.  And also want something to point at when people suggest implementing our own parsers...

c-prather:
Fixes #936

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 - [[PR 885]](https://github.com/parthenon-hpc-lab/parthenon/pull/885) Expose PackDescriptor and use uids in SparsePacks
 
 ### Fixed (not changing behavior/API/variables/...)
+- [[PR 937]](https://github.com/parthenon-hpc-lab/parthenon/pull/937) Fix multiple line continuations
 - [[PR 933]](https://github.com/parthenon-hpc-lab/parthenon/pull/933) Remove extraneous debug check
 - [[PR 917]](https://github.com/parthenon-hpc-lab/parthenon/pull/917) Update Iterative Tasking Infrastructure
 - [[PR 890]](https://github.com/parthenon-hpc-lab/parthenon/pull/890) Fix bugs in sparse communication and prolongation
--- a/src/parameter_input.cpp
+++ b/src/parameter_input.cpp
@@ -3,7 +3,7 @@
 // Copyright(C) 2014 James M. Stone <jmstone@princeton.edu> and other code contributors
 // Licensed under the 3-clause BSD License, see LICENSE file for details
 //========================================================================================
-// (C) (or copyright) 2020-2022. Triad National Security, LLC. All rights reserved.
+// (C) (or copyright) 2020-2023. Triad National Security, LLC. All rights reserved.
 //
 // This program was produced under U.S. Government contract 89233218CNA000001 for Los
 // Alamos National Laboratory (LANL), which is operated by Triad National Security, LLC
@@ -188,6 +188,9 @@ void ParameterInput::LoadFromStream(std::istream &is) {
       param_name = multiline_name;
       param_value = multiline_value;
       param_comment = multiline_comment;
+      multiline_name = "";
+      multiline_value = "";
+      multiline_comment = "";
     }
 
     if (!continuing) {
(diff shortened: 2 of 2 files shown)


lroberts36 (approved):
LGTM, I agree that it might be desirable to re-write this at some point (although I am sure no one will ever have time for that).

Yurlungur (commented):
I don't fully understand how this change resolves the above issue, but seems like it soils be pretty easy to add a test to catch this, right?

pgrete:
I now stared at this for some time and also don't fully understand how this fixes the issue.
@bprather , assuming that you've now understood the logic (as you fixed it), would you mind adding additional comments to the existing code so that future us have an easier time to parse the logic (and review)?

c-prather:
I gave it a shot?

I'm honestly not too sure how I could make it clearer or what I could say, sorry.   The `multiline_x` vars are buffers used for line continuations.  These buffers were not cleared when treating multiple line continuations (they reside outside the loop over lines in the file), so they accumulated more data over time, every time there was a line continuation.  Clearing them fixes the problem.

This is a very common first mistake when writing state-machine parsers (I've made it myself).  At risk of sounding sanctimonious, I think this is decent evidence for why we shouldn't be maintaining our own file parser when the problem's been solved well in so many other places.

pgrete:
I was referring to
```
    if (continuing || continuation) {   // <--------------------- this logic
      multiline_name += param_name;
      multiline_value += param_value;
      multiline_comment += param_comment;
      continuing = true;
    }
    if (continuing && !continuation) { // <--------------------- and that logic (in combination with what happens inside those statements)
      continuing = false;
      param_name = multiline_name;
      param_value = multiline_value;
      param_comment = multiline_comment;
      // Clear the multiline_x buffers to hold the next multi-line value
      multiline_name = "";
      multiline_value = "";
      multiline_comment = "";
    }
```
For example, I'm reading `if (continuing && !continuation)` as "if sth is continuing and at the same time it's not a continuation then..." and cannot really tell what's going on.
Maybe it's just wording, but it seems @Yurlungur also had similar issues parsing the logic in that function.

c-prather:
`continuing` means *we* are continuing, in the parser, from a previous line.  `if (continuing && !continuation)` means we've continued from the previous line, but this line does not contain a continuation character.  Therefore, it indicates we should flush the current continuation buffer into the line data, because there aren't more lines to read.

The conditionals handle the 2 extra states possible when parsing a line, for a total of 3:
1. Normal line, no continuation
2. Continuation appears at the end of this line, we should append contents to a buffer
3. No continuation appears at the end of this line, we should append contents and then flush the buffer

This is a bog-standard 3-state parser, like CS101 assignment simple, which it looks like Josh threw together very early in Parthenon development. I heavily expect that if either of you spend a few minutes with it, you could understand exactly what's going on, as I did to fix the bug flushing buffers.  I do agree the variables could be named better, and I'll add basic comments outlining the states so it can be understood a little more easily.

Yurlungur (approved):
Thanks @bprather the naming is much clearer now. I know we don't want to put too much effort into this, as it's just a simpler parser, but I do think some unit tests are in order if anyone ever has time.

(Or a rewrite, as @lroberts36 suggests.)

pgrete (approved):
Thanks @bprather 
The updated comments make this indeed much easier to parse (pun intended ;))

PR #948, opened 2023-10-04
Add MG Solver Interface

lroberts36:
## PR Summary

This PR moves the infrastructure for making multigrid and BiCGStab task lists from the `poisson_gmg` example to `src/solvers` and attempts to create somewhat plug and play solvers for ~general problems. Also works iterative tasking through solvers.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


src/solvers/solver_utils.hpp
```diff
@@ -135,6 +135,165 @@ struct Stencil {
   }
 };
 
+namespace impl {
```
Yurlungur: Should there be a `SolverUtils` namespace for these to live in?

src/solvers/solver_utils.hpp
```diff
@@ -135,6 +135,165 @@ struct Stencil {
   }
 };
 
+namespace impl { 
+template <class in, class out, bool only_md_level = false>
+TaskStatus CopyData(std::shared_ptr<MeshData<Real>> &md) {
```
Yurlungur: :+1:

src/solvers/bicgstab_solver.hpp
```diff
@@ -0,0 +1,276 @@
...
+  BICGVARIABLE(x, v);
+  BICGVARIABLE(x, h);
+  BICGVARIABLE(x, s);
+  BICGVARIABLE(x, t);
+  BICGVARIABLE(x, r);
+  BICGVARIABLE(x, p);
+  BICGVARIABLE(x, u);
+ 
+ public:
+  BiCGSTABSolver(StateDescriptor *pkg, BiCGSTABParams params_in, equations eq_in = equations())
```
Yurlungur: clever way to add properties of a package

src/solvers/bicgstab_solver.hpp
```diff
@@ -0,0 +1,276 @@
...
+ 
+ public:
+  BiCGSTABSolver(StateDescriptor *pkg, BiCGSTABParams params_in, equations eq_in = equations()) 
+      : preconditioner(pkg, MGParams(), eq_in), params_(params_in), 
+        iter_counter(0), eqs_()  { 
+    using namespace refinement_ops;
+    auto mu = Metadata(
+      {Metadata::Cell, Metadata::Independent, Metadata::FillGhost, Metadata::WithFluxes, Metadata::GMGRestrict});
+    auto m_no_ghost = Metadata({Metadata::Cell, Metadata::Derived, Metadata::OneCopy});
+    pkg->AddField(u::name(), mu);
```
Yurlungur: There could be a name collision here if multiple packages try to add bicgstabsolvers. You could add `Metadata::Private` to the metadata flags. That would prefix all the variable names with the package name for the purposes of string-based name resolution.

I guess since they're templated on `x` maybe that's not an issue?

(I don't *think* the types will overlap, because the solver class is templated on x, rhs, and equations.)

src/solvers/mg_solver.hpp
```diff
@@ -0,0 +1,338 @@
...
+  MGVARIABLE(u, res_err); // residual on the way up and error on the way down
+  MGVARIABLE(u, temp); // Temporary storage 
+  MGVARIABLE(u, u0); // Storage for initial solution during FAS
+  MGVARIABLE(u, D); // Storage for (approximate) diagonal
+
+ public:
+  MGSolver(StateDescriptor *pkg, MGParams params_in, equations eq_in = equations()) 
+      : params_(params_in), iter_counter(0), eqs_() { 
+    using namespace parthenon::refinement_ops;
+    auto mres_err = Metadata({Metadata::Cell, Metadata::Independent, Metadata::FillGhost,
```
Yurlungur: See above comment about the solver var registration.

Yurlungur (commented):
I like the class-based design!

src/solvers/solver_utils.hpp
lroberts36: I changed this to `solvers::utils` instead of `solvers::impl`

src/solvers/bicgstab_solver.hpp
lroberts36: There is potential for a name collision, but only if you instantiate two solvers with the same type `x` since the `u::name()` should give something like `x::name() + ".u"`. It probably makes sense to make the fields private, but even then if you have to solvers that work on the same variable in a package there would be a collision. How to deal with these internal fields in cases where there is more than one solver needs some more thought, since we maybe don't want to allocate separate internal fields for each one (although that depends on the exact use case).

example/poisson_gmg/poisson_driver.cpp
```diff
@@ -24,539 +24,109 @@
...
+TaskCollection PoissonDriver::MakeTaskCollection(BlockList_t &blocks) {
   using namespace parthenon;
   using namespace poisson_package;
   TaskCollection tc;
   TaskID none(0);
 
   auto pkg = pmesh->packages.Get("poisson_package");
-  auto max_iterations = pkg->Param<int>("max_iterations");
+  auto solver = pkg->Param<std::string>("solver");
+  auto flux_correct = pkg->Param<bool>("flux_correct");
```
pdmullen, line 71: when would one not use flux correction?

src/interface/sparse_pack.hpp
```diff
@@ -142,7 +142,23 @@ class SparsePack : public SparsePackBase {
     // accessed on device via instance of types in the type list Ts...
     // The pack will be created and accessible on the device
     template <class T>
-    SparsePack GetPack(T *pmd, const std::vector<bool> &include_block = {}) const {
+    SparsePack GetPack(T *pmd, std::vector<bool> include_block = {},
+                       bool only_fine_two_level_composite_blocks = true) const {
```
pdmullen, line 147: Despite the long, descriptive variable name, I am still unsure what this is doing.

pdmullen (approved):
Looks great 😸

example/poisson_gmg/poisson_driver.cpp
lroberts36: Mostly for testing reasons. As things are written now it is not included in raw MG. There is a way to include it in MG, but I found that I still got good convergence with MG-BiCGSTAB when I include flux correction in BiCGSTAB but neglect it in the MG preconditioner (a description of how to include flux correction in MG is in those notes I wrote up).

src/interface/sparse_pack.hpp
lroberts36: If a mesh data object corresponds to a `GridType::two_level_composite_mesh`, then if `only_fine_two_level_composite_blocks` is true the pack will only include blocks that are on the finer level of the two.

src/solvers/solver_utils.hpp
```diff
@@ -135,6 +145,155 @@ struct Stencil {
   }
 };
 
+namespace utils {
+template <class in, class out, bool only_fine_on_composite = true>
+TaskStatus CopyData(const std::shared_ptr<MeshData<Real>> &md) {
```
Yurlungur, line 150: This seems more generic than just solvers. Maybe deserves a space in some kind of just... `var` or `grid` utils...

src/solvers/solver_utils.hpp
```diff
@@ -135,6 +145,155 @@ struct Stencil {
...
+  using TE = parthenon::TopologicalElement;
+  TE te = TE::CC;
+  IndexRange ib = md->GetBoundsI(IndexDomain::entire, te);
+  IndexRange jb = md->GetBoundsJ(IndexDomain::entire, te);
+  IndexRange kb = md->GetBoundsK(IndexDomain::entire, te);
+
+  auto desc = parthenon::MakePackDescriptor<in, out>(md.get());
+  auto pack = desc.GetPack(md.get(), {}, only_fine_on_composite);
+  parthenon::par_for(
+      DEFAULT_LOOP_PATTERN, "SetPotentialToZero", DevExecSpace(), 0,
```
Yurlungur: incorrect kernel name

src/solvers/solver_utils.hpp
```diff
@@ -135,6 +145,155 @@ struct Stencil {
...
+  if (only_interior) {
+    // The neighbors array will only be set for a block if its a leaf block
+    for (int b = 0; b < nblocks; ++b)
+      include_block[b] = md->GetBlockData(b)->GetBlockPointer()->neighbors.size() == 0;
+  }
+
+  auto desc = parthenon::MakePackDescriptor<a_t, b_t, out>(md.get());
+  auto pack = desc.GetPack(md.get(), include_block, only_fine_on_composite);
+  parthenon::par_for(
+      DEFAULT_LOOP_PATTERN, "SetPotentialToZero", DevExecSpace(), 0,
```
Yurlungur: incorrect kernel name

src/solvers/solver_utils.hpp
```diff
@@ -135,6 +145,155 @@ struct Stencil {
...
+  using TE = parthenon::TopologicalElement;
+  TE te = TE::CC;
+  std::vector<bool> include_block(nblocks, true);
+  auto desc = parthenon::MakePackDescriptor<var>(md.get());
+  auto pack = desc.GetPack(md.get(), include_block, only_fine_on_composite);
+  const size_t scratch_size_in_bytes = 0;
+  const int scratch_level = 1;
+  const int ng = parthenon::Globals::nghost;
+  parthenon::par_for_outer(
+      DEFAULT_OUTER_LOOP_PATTERN, "Print", DevExecSpace(), scratch_size_in_bytes,
```
Yurlungur: Ironically also incorrect kernel name... since all the others are call` SetToZero`

src/solvers/solver_utils.hpp
```diff
@@ -135,6 +145,155 @@ struct Stencil {
...
+      md, wa, wb, false);
+}
+
+template <class var, bool only_fine_on_composite = true>
+TaskStatus SetToZero(const std::shared_ptr<MeshData<Real>> &md) {
+  int nblocks = md->NumBlocks();
+  using TE = parthenon::TopologicalElement;
+  TE te = TE::CC;
+  std::vector<bool> include_block(nblocks, true);
+  auto desc = parthenon::MakePackDescriptor<var>(md.get());
```
Yurlungur, line 222: Should this be variatic so the user can set multiple vars to zero at once?

src/interface/sparse_pack.hpp
```diff
@@ -142,7 +143,23 @@ class SparsePack : public SparsePackBase {
...
+      // If this is a composite grid MeshData object, only include blocks on
+      // the finer level
```
Yurlungur: and if `only_fine_two_level_composit_blocks` is set

src/solvers/solver_utils.hpp
```diff
@@ -13,11 +13,21 @@
...
 #define SOLVERS_SOLVER_UTILS_HPP_
 
+#include <memory>
 #include <string>
+#include <utility>
 #include <vector>
 
 #include "kokkos_abstraction.hpp"
 
+#define INTERNALSOLVERVARIABLE(base, varname)                                            \
```
Yurlungur: Since this macro is publicly visible, should it have the `PARTHENON_` prefix? I know that's pretty unwieldy...

example/poisson_gmg/poisson_package.cpp
```diff
@@ -77,60 +80,60 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
...
+  eq.do_flux_cor = flux_correct;
+
+  parthenon::solvers::MGParams mg_params;
+  mg_params.max_iters = max_poisson_iterations;
+  mg_params.residual_tolerance = res_tol;
+  mg_params.do_FAS = do_FAS;
+  mg_params.smoother = smoother_method;
+  parthenon::solvers::MGSolver<u, rhs, PoissonEquation> mg_solver(pkg.get(), mg_params,
+                                                                  eq);
+  pkg->AddParam<>("MGsolver", mg_solver, parthenon::Params::Mutability::Mutable);
```
Yurlungur, line 144: This is super nice now... modulo of vourse that one has to implement the `PoissonEquation`

example/poisson_gmg/poisson_equation.hpp
```diff
@@ -0,0 +1,218 @@
...
+
+namespace poisson_package {
+
+// This class implement methods for calculating A.x = y and returning the diagonal of A,
+// where A is the the matrix representing the discretized Poisson equation on the grid.
+// Here we implement the Laplace operator in terms of a flux divergence to (potentially)
+// consistently deal with coarse fine boundaries on the grid. Only the routines Ax and
+// SetDiagonal need to be defined for interfacing this with solvers. The other methods
+// are internal, but can't be marked private or protected because they launch kernels
+// on device.
```
Yurlungur, line 34: It might be useful to specify somewhere (for example, the docs ;) ) exaclty the API expected of an equations type. Could even impose it with a concept.

example/poisson_gmg/poisson_equation.hpp
```diff
@@ -0,0 +1,218 @@
...
+      auto recv_flxcor = tl.AddTask(send_flxcor, parthenon::ReceiveFluxCorrections, md);
+      flux_res = tl.AddTask(recv_flxcor, parthenon::SetFluxCorrections, md);
+    }
+    return tl.AddTask(flux_res, FluxMultiplyMatrix<x_t, out_t>, md);
+  }
+
+  // Calculate an approximation to the diagonal of the matrix A and store it in diag_t.
+  // For a uniform grid or when flux correction is ignored, this diagonal calculation
+  // is exact. Exactness is (probably) not required since it is just used in Jacobi
+  // iterations.
```
Yurlungur, line 58: Also, this is a me problem, but I always have trouble unwrapping the shape of the mesh into matrix form... Not on you to document that though.

example/poisson_gmg/poisson_driver.cpp
```diff
@@ -24,559 +24,108 @@
...
+          get_err,
+          [](PoissonDriver *driver, int partition) {
+            if (partition != 0) return TaskStatus::complete;
+            driver->final_rms_error =
+                std::sqrt(driver->err.val / driver->pmesh->GetTotalCells());
+            if (Globals::my_rank == 0)
+              printf("Final rms error: %e\n", driver->final_rms_error);
+            return TaskStatus::complete;
+          },
+          this, i);
```
Yurlungur, line 128: nice :+1:

Yurlungur (approved):
A few nitpicks below, but I consider them nonblocking. This is a much nicer interface. Thanks for cleaning it up!

src/solvers/solver_utils.hpp
lroberts36: Maybe so, but I think the `only_fine_on_composite` stuff maybe makes it confusing for general usage? Let's see what others think.

src/solvers/solver_utils.hpp
lroberts36: fixed.

src/solvers/solver_utils.hpp
lroberts36: fixed.

src/solvers/solver_utils.hpp
lroberts36: haha, not sure how that happened. fixed.

src/solvers/solver_utils.hpp
lroberts36: I messed around with that a bit and had some trouble getting it to work on device. There is probably a solution, but maybe save for later.

src/interface/sparse_pack.hpp
lroberts36: fixed.

src/solvers/solver_utils.hpp
pdmullen: Perhaps just documenting `only_fine_on_composite` in a Doxygen comment could be the solution 😉

src/solvers/solver_utils.hpp
lroberts36: no, that is a good point. I changed it.

example/poisson_gmg/poisson_equation.hpp
lroberts36: Yeah, I think this is best left to the method documentation.

src/solvers/solver_utils.hpp
lroberts36: ha, maybe so

example/poisson_gmg/poisson_equation.hpp
lroberts36: I added a couple of comments in the the solver headers. It could be imposed with a concept, but I think doing that is a little tricky since these are template methods. Also, it is not a really general thing so a concept might be overkill.

Yurlungur: Works for me

c-prather (approved):
Started to move things in KHARMA over to this and I like the design.  It ends up looking a lot like the previous turnkey solvers but with less hacking.

If I need changes going forward for some reason I'll make/post them myself.

Yurlungur:
@lroberts36 what's the status of this? Can it go in?

lroberts36:
@Yurlungur: I thought @pgrete wanted to look at it, but otherwise it is ready to go.

example/poisson_gmg/poisson_equation.hpp
```diff
@@ -0,0 +1,218 @@
...
+ public:
+  bool do_flux_cor = false;
+
+  // Add tasks to calculate the result of the matrix A (which is implicitly defined by
+  // this class) being applied to x_t and store it in field out_t
+  template <class x_t, class out_t, class TL_t>
+  parthenon::TaskID Ax(TL_t &tl, parthenon::TaskID depends_on,
+                       std::shared_ptr<parthenon::MeshData<Real>> &md) {
+    auto flux_res = tl.AddTask(depends_on, CalculateFluxes<x_t>, md);
+    if (do_flux_cor && !(md->grid.type == parthenon::GridType::two_level_composite)) {
```
pgrete, line 45: Two level composite means that the block contains data from child blocks on different levels?
I saw the enum defining the grid types, but I not sure I could parse all the different types by their names.

example/poisson_gmg/poisson_equation.hpp
```diff
@@ -0,0 +1,218 @@
...
+      auto start_flxcor =
+          tl.AddTask(flux_res, parthenon::StartReceiveFluxCorrections, md);
+      auto send_flxcor = tl.AddTask(flux_res, parthenon::LoadAndSendFluxCorrections, md);
+      auto recv_flxcor = tl.AddTask(send_flxcor, parthenon::ReceiveFluxCorrections, md);
```
pgrete: Is there a technical reason that the receive depends on the send rather than the start receive?

example/poisson_gmg/poisson_driver.cpp
```diff
@@ -24,559 +24,108 @@
...
-      int mg_reg_dep_id = 0;
-      TaskRegion &region = tc.AddRegion(num_partitions);
-      for (int i = 0; i < num_partitions; ++i) {
-        TaskList &tl = region[i];
-        AddMultiGridTasksLevel(region, tl, none, i, mg_reg_dep_id, max_level, min_level,
-                               max_level);
-      }
+    auto solve = zero_u;
+    auto &itl = tl.AddIteration("Solver");
+    if (solver == "BiCGSTAB") {
```
pgrete, line 103: Given that these map to solvers provided by Parthenon, could we also provide an enum class type interface to be used in downstream codes?

src/interface/sparse_pack.hpp
pgrete: Just saw this comment (after I wrote the one on "composite..." above).
Given that @pdmullen also stumbled across the naming, it might be an indication that some more doc (e.g, around where the enum class is defined) would be useful.

src/solvers/bicgstab_solver.hpp
```diff
@@ -0,0 +1,294 @@
...
+
+  TaskID AddTasks(TaskList &tl, IterativeTasks &itl, TaskID dependence, int i,
+                  Mesh *pmesh, TaskRegion &region, int &reg_dep_id) {
+    using namespace utils;
+    auto &md = pmesh->mesh_data.GetOrAdd("base", i);
+    iter_counter = 0;
+
+    // Initialization: x <- 0, r <- rhs, rhat0 <- rhs,
+    // rhat0r_old <- (rhat0, r), p <- r, u <- 0
+    // TODO(LFR): Fix this to calculate the actual residual
```
pgrete: Is the TODO still valid?

src/solvers/solver_utils.hpp
```diff
@@ -135,6 +145,155 @@ struct Stencil {
...
+        IndexRange ib = cb.GetBoundsI(IndexDomain::interior, te);
+        IndexRange jb = cb.GetBoundsJ(IndexDomain::interior, te);
+        IndexRange kb = cb.GetBoundsK(IndexDomain::interior, te);
+        parthenon::par_for_inner(parthenon::inner_loop_pattern_simdfor_tag, member, kb.s,
+                                 kb.e, jb.s, jb.e, ib.s, ib.e, [&](int k, int j, int i) {
+                                   const int nvars = pack.GetUpperBound(b, var()) -
+                                                     pack.GetLowerBound(b, var()) + 1;
+                                   for (int c = 0; c < nvars; ++c)
+                                     pack(b, te, var(c), k, j, i) = 0.0;
+                                 });
```
pgrete: Why are the inner loop bounds not taking the variable `c` into account?
The way it's currently written we might loose some optimization because the `c` loop results in data not being aligned with the `i` index, isn't it?

src/solvers/solver_utils.hpp
```diff
@@ -135,6 +145,155 @@ struct Stencil {
...
+  auto pack = desc.GetPack(md.get());
+  Real gsum(0);
+  parthenon::par_reduce(
+      parthenon::loop_pattern_mdrange_tag, "DotProduct", DevExecSpace(), 0,
+      pack.GetNBlocks() - 1, kb.s, kb.e, jb.s, jb.e, ib.s, ib.e,
+      KOKKOS_LAMBDA(const int b, const int k, const int j, const int i, Real &lsum) {
+        const int nvars = pack.GetUpperBound(b, a_t()) - pack.GetLowerBound(b, a_t()) + 1;
+        for (int c = 0; c < nvars; ++c)
+          lsum += pack(b, te, a_t(c), k, j, i) * pack(b, te, b_t(c), k, j, i);
+      },
```
pgrete: Similar to the question above: Are we leaving out vectorization here because of the explicit inner loop (rather than using hierarchical parallelism)?

pgrete (approved):
Thanks for the cleanup. The new interface is definitely easier to parse (though admittedly I'm still not able to parse everything, see minor comments).

One question on the doc: Are you planning to move the content from the pdf doc you prepared to the Parthenon doc at some point?

PS: I got sidetracked last week tracking down a (very likely) comm bug that I am reliably able to reproduce/trace back to when the original MG PR got merged it.
I still trying to determine if it's MPI lib/machine specific more Parthenon related.

example/poisson_gmg/poisson_equation.hpp
lroberts36: A two-level composite grid corresponds to all of the nodes at a given level (i.e. both leaf and internal nodes) and all of the leaf nodes at the next-coarser level that have neighbors on the finer level. When there is no AMR, these levels just correspond to levels in the GMG hierarchy, but with AMR the coarse neighbor blocks are required so that during communication prolongation at the internal fine-coarse boundaries uses the most up to date data. The details are described in the pdf I included previously.

example/poisson_gmg/poisson_equation.hpp
lroberts36: No, I think I was just going on autopilot when writing those tasks.

src/solvers/solver_utils.hpp
lroberts36: Yeah, I agree that it would be better to wrap the `par_for_inner` as 
```c++
const int nvars = pack.GetUpperBound(b, var()) -
                             pack.GetLowerBound(b, var()) + 1;
for (int c = 0; c < nvars; ++c) {
  parthenon::par_for_inner(...) {
      pack(b, te, var(c), k, j, i) = 0.0;
    });
}
```
and something similar below. It was written initially without accounting for the fact that variables could be more than size one, and I just quickly hacked this in there. Currently, this doesn't have any noticeable impact on performance.

src/solvers/solver_utils.hpp
lroberts36: yes, see above.

src/solvers/bicgstab_solver.hpp
lroberts36: No, it can be removed.

example/poisson_gmg/poisson_driver.cpp
lroberts36: I am not sure if we want to do that because the solvers may end up being composable (e.g. allow for different preconditioners and base level solvers via template parameters).

lroberts36:
@pgrete:
> One question on the doc: Are you planning to move the content from the pdf doc you prepared to the Parthenon doc at some point?

I am not sure exactly what to do there. I don't know how well our documentation supports latex. I mostly wrote those notes so I wouldn't forget what is implemented. Maybe at some point they could be the appendix of a paper that uses this stuff.
 
> PS: I got sidetracked last week tracking down a (very likely) comm bug that I am reliably able to reproduce/trace back to when the original MG PR got merged it. I still trying to determine if it's MPI lib/machine specific more Parthenon related.

Let me know what you find or if you need help debugging.

src/solvers/bicgstab_solver.hpp
lroberts36: done.

example/poisson_gmg/poisson_equation.hpp
lroberts36: fixed.

src/solvers/solver_utils.hpp
lroberts36: I fixed this in one place where there was already hierarchical parallelism being used. Other places in utils I just added a todo saying the change could be made if these kernels were impacting performance.

PR #961, opened 2023-10-19
don't guard greater than 32767 tags if not an MPI build

gshipman:
## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


Yurlungur (approved):
Good catch! To get it to pass the style check, can you run `make format` and `make lint` in the build directory? That should autoformat the code and hopefully then if the linter is happy the style check will pass

src/bvals/comms/tag_map.cpp
```diff
@@ -45,13 +45,20 @@ void TagMap::AddMeshDataToMap(std::shared_ptr<MeshData<Real>> &md) {
...
     auto &pair_map = it->second;
     int idx = 0;
     std::for_each(pair_map.begin(), pair_map.end(),
                   [&idx](auto &pair) { pair.second = idx++; });
-    if (idx > 32767)
-      PARTHENON_FAIL("Number of tags exceeds the maximum allowed by the MPI standard.");
+    #ifdef MPI_PARALLEL
+    if (idx > (*(int*)max_tag))
+      PARTHENON_FAIL("Number of tags exceeds the maximum allowed by this MPI version.");
+    #endif
```
lroberts36:
```suggestion
    #ifdef MPI_PARALLEL
    if (idx > (*reinterpret_cast<int*>(max_tag)) && it->first != Globals::my_rank)
      PARTHENON_FAIL("Number of tags exceeds the maximum allowed by this MPI version.");
    #endif
```

lroberts36: Exceeding the maximum number of tags is fine if the sender and receiver are on the same rank (since MPI is not used in this case). The linter complains about the c-style cast. I think the `reinterpret_cast` is the right thing to do, but not totally sure.

lroberts36 (approved):
LGTM, with a couple of suggestions.

src/bvals/comms/tag_map.cpp
```diff
@@ -45,13 +45,20 @@ void TagMap::AddMeshDataToMap(std::shared_ptr<MeshData<Real>> &md) {
 }
 
 void TagMap::ResolveMap() {
+  #ifdef MPI_PARALLEL
+  int flag;
+  void *max_tag; // largest supported MPI tag value
+  MPI_Comm_get_attr( MPI_COMM_WORLD, MPI_TAG_UB, &max_tag, &flag);
+  #endif
```
pgrete: Is there a reason we should 
- not use `PARTHENON_MPI_CHECK` for the MPI call
- not check for the result of `flag` before we continue to work with `max_tag` below?

gshipman:
Hi folks, can we get a final review and merge?

src/bvals/comms/tag_map.cpp
Yurlungur: @gshipman I think we were waiting for you to address @pgrete 's question here... after that, from my perspective, it's ready for merge.

gshipman: @Yurlungur see latest commit. Should be good if checks pass. Thx

Yurlungur:
Works for me. Thanks @gshipman

pgrete:
@par-hermes format

pgrete:
Extended test suite ran. I'm force merging now.

PR #962, opened 2023-10-19
Add support for in-situ histograms/profiles

pgrete:
## PR Summary
Add support for in-situ histograms/profiles

Detailed doc here: https://parthenon-hpc-lab.github.io/parthenon/pgrete/histogram-outputs/src/outputs.html#histograms

## PR Checklist

- [ ] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


doc/sphinx/src/outputs.rst
```diff
@@ -158,6 +161,134 @@ This will produce a text file (``.hst``) output file every 1 units of
...
+
+with the following parameters
+
+- ``num_histograms=INT``
+   The number of histograms defined in this block.
+   All histogram definitions need to be prefix with ``hist#_`` where ``#`` is the
+   histogram number starting to count from ``0``.
+   All histograms will be written to the same output file with the "group" in the
+   output corresponding to the histogram number.
+- ``hist#_ndim=INT`` (either ``1`` or ``2``)
```
BenWibking: maybe just `hist_dim`?

doc/sphinx/src/outputs.rst
```diff
@@ -158,6 +161,134 @@ This will produce a text file (``.hst``) output file every 1 units of
...
+- ``hist#_x_edges_min=FLOAT``
+   Minimum value (inclusive) of the bins in the first dim.
+   Used/required only for ``lin`` and ``log`` edge type.
+- ``hist#_x_edges_max=FLOAT``
+   Maximum value (inclusive) of the bins in the first dim.
+   Used/required only for ``lin`` and ``log`` edge type.
+- ``hist#_x_edges_num_bins=INT`` (must be ``>=1``)
+   Number of equally spaced bins between min and max value in the first dim.
+   Used/required only for ``lin`` and ``log`` edge type.
+- ``hist#_x_edges_list=FLOAT,FLOAT,FLOAT,...`` (comma separated list of increasing values)
```
BenWibking: Does this list include the innermost and outmost edges?

doc/sphinx/src/outputs.rst
```diff
@@ -158,6 +161,134 @@ This will produce a text file (``.hst``) output file every 1 units of
...
+Parthenon supports calculating flexible 1D and 2D histograms in-situ that
+are written to disk in HDF5 format.
+Currently supported are
+
+- 1D and 2D histograms
+- binning by variable or coordinate (x1, x2, x3 and radial distance)
+- counting samples and or summing a variable  
+- weighting by volume and/or variable
+
+The output format follows ``numpy`` convention, so that plotting data
```
BenWibking, line 176: Does this mean it follows C-ordering for 2D arrays?

doc/sphinx/src/outputs.rst
```diff
@@ -158,6 +161,134 @@ This will produce a text file (``.hst``) output file every 1 units of
...
+A ``<parthenon/output*>`` block containing one simple and one complex
+example might look like::
+
+   <parthenon/output8>
+   file_type = histogram # required, sets the output type
+   dt = 1.0              # required, sets the output interval
+   num_histograms = 2    # required, specifies how many histograms are defined in this block
+
+   # 1D histogram ("standard", i.e., counting occurance in bin)
+   hist0_ndim = 1
```
BenWibking: The `hist0_`, `hist1_`, etc. prefixes are very ugly. Is it necessary to support more than 1 histogram per output file?

src/outputs/histogram.cpp
```diff
@@ -0,0 +1,554 @@
...
+          } else if (x_var_type == VarType::X3) {
+            x_val = coords.Xc<3>(k, j, i);
+          } else if (x_var_type == VarType::R) {
+            x_val = Kokkos::sqrt(SQR(coords.Xc<1>(k, j, i)) + SQR(coords.Xc<2>(k, j, i)) +
+                                 SQR(coords.Xc<3>(k, j, i)));
+          } else {
+            x_val = x_var(b, x_var_component, k, j, i);
+          }
+          if (x_val < x_edges(0) || x_val > x_edges(x_edges.extent_int(0) - 1)) {
+            return;
```
BenWibking: This makes sense to do by default. It would also be useful to have the option to accumulate values in the first or last bins if they are outside the histogram range.

src/outputs/histogram.cpp
```diff
@@ -0,0 +1,554 @@
...
+            } else if (y_var_type == VarType::R) {
+              y_val =
+                  Kokkos::sqrt(SQR(coords.Xc<1>(k, j, i)) + SQR(coords.Xc<2>(k, j, i)) +
+                               SQR(coords.Xc<3>(k, j, i)));
+            } else {
+              y_val = y_var(b, y_var_component, k, j, i);
+            }
+
+            if (y_val < y_edges(0) || y_val > y_edges(y_edges.extent_int(0) - 1)) {
+              return;
```
BenWibking: Same comment as above.

BenWibking (commented):
This will be very nice to have!

My major comments are that it would be nice to have:
- less ugly runtime parameters (not having to use the `histX_` prefixes)
- having a runtime option to not discard values if they are outside the given range, and
- an option to output in a text-based (csv, json, ...?) format.

In particular, regarding the first point, is it necessary to have more than one histogram per output file?

doc/sphinx/src/outputs.rst
pgrete: I used `ndim` because it's also used in other places to specify/check for dimensionality.

doc/sphinx/src/outputs.rst
pgrete: Yes, I'll update the `doc` to make this clear.

doc/sphinx/src/outputs.rst
pgrete: yes, so the data output data uses `[x,y]` indexing with `y` being the fast index. I'll update the doc do be more specific here, too.

doc/sphinx/src/outputs.rst
pgrete: I'm also not a big fan of this style, but it's the best I came up with so I'm open for suggestions (that do not involve moving to yaml/json/... for parsing inputs, which now seems to come up more often *wink* @bprather 

The main motivation for this approach is to not clobber the (shared) filesystem with many new files.
I can see that one easily calculate 10+ histograms in a simulation at high cadence, which would result in many tiny files (that'd need to be managed again).
To mitigate this, I initially even tried to put all histograms of all times in a single file, but gave up because of HDF5 (deleting/updating data cannot be easily done).

src/outputs/histogram.cpp
pgrete: Agreed. Should be easy enough to be added.

doc/sphinx/src/outputs.rst
BenWibking: You could append to a yaml/json output file to accomplish that... is avoiding an extra library dependency the issue?

BenWibking: This might be what you tried, but it is *supposed* to be possible to do streaming writes using HDF5 with the SWMR interface: https://docs.h5py.org/en/stable/swmr.html

Yurlungur: > To mitigate this, I initially even tried to put all histograms of all times in a single file, but gave up because of HDF5 (deleting/updating data cannot be easily done).

> This might be what you tried, but it is supposed to be possible to do streaming writes using HDF5 with the SWMR interface: https://docs.h5py.org/en/stable/swmr.html

I have in the past used extendable datasets to output all data at all times in a single file in HDF5. I do not recommend it. The main issue is that this strategy is not robust against failure. If the code crashes during an output, the entire file is corrupted, not just the time you're trying to output. It's also less robust against the kind of thing we often do, where we restart a simulation and steer it differently upon restart, for example. For these reasons I am a strong believer in 1 file per output time.

> I'm also not a big fan of this style, but it's the best I came up with so I'm open for suggestions

What about simply assuming one `output` block per histogram, but combining them into a single file if the `dt` is the same?

Yurlungur: Or alternatively have a `file_index` be a parameter per block, and all blocks with the same `file_index` are in the same file and assumed to have the same `dt`. If multiple blocks set `dt` this is either undefined behavior, or a runtime error.

doc/sphinx/src/outputs.rst
```diff
@@ -158,6 +161,134 @@ This will produce a text file (``.hst``) output file every 1 units of
...
+
+- ``num_histograms=INT``
+   The number of histograms defined in this block.
+   All histogram definitions need to be prefix with ``hist#_`` where ``#`` is the
+   histogram number starting to count from ``0``.
+   All histograms will be written to the same output file with the "group" in the
+   output corresponding to the histogram number.
+- ``hist#_ndim=INT`` (either ``1`` or ``2``)
+   Dimensionality of the histogram.
+- ``hist#_x_variable=STRING`` (variable name or special coordinate string ``HIST_COORD_X1``, ``HIST_COORD_X2``, ``HIST_COORD_X3`` or ``HIST_COORD_R``)
```
Yurlungur: I'm a little bit confused by this interface. Is the following accurate?
1. In 2D `hist_x_variable` is the x axis and `hist_binned_variable` is the y axis? With `hist_weight` as the weighting quantity?
2. In 1D is it just `hist_binned_variable` with `hist_x_variable` no longer used?

doc/sphinx/src/outputs.rst
```diff
@@ -158,6 +161,134 @@ This will produce a text file (``.hst``) output file every 1 units of
...
+   see next parameter.
+   For a scalar variable, the component needs to be specified as ``0`` anyway.
+   If sampling (i.e., counting the number of value inside a bin) is to be used the special
+   string ``HIST_ONES`` can be set.
+- ``hist#_binned_variable_component=INT``
+   Component index of the variable to be binned.
+   Used/required only if a variable is binned and not ``HIST_ONES``.
+- ``hist#_weight_by_volume=BOOL`` (``true`` or ``false``)
+   Apply volume weighting to the binned variable. Can be used simultaneously with binning
+   by a different variable.
```
Yurlungur: Is there an option to normalize the histogram? Or maybe that's done in post? I *think* it would be easy to achieve by re-using the weight-by-volume. just divide by total physical volume in the weight.

src/outputs/histogram.cpp
```diff
@@ -0,0 +1,554 @@
...
+  auto edge_dbin = std::numeric_limits<Real>::quiet_NaN();
+
+  const auto edge_type_str = pin->GetString(block_name, prefix + "type");
+  if (edge_type_str == "lin" || edge_type_str == "log") {
+    edge_min = pin->GetReal(block_name, prefix + "min");
+    const auto edge_max = pin->GetReal(block_name, prefix + "max");
+    PARTHENON_REQUIRE_THROWS(edge_max > edge_min,
+                             "Histogram max needs to be larger than min.")
+
+    const auto edge_num_bins = pin->GetReal(block_name, prefix + "num_bins");
```
Yurlungur: probably a premature optimization, but consider reserving space in `edges_in` for `num_edge_bins` once we know what that is.

src/outputs/histogram.cpp
```diff
@@ -0,0 +1,554 @@
...
+    }
+
+  } else if (edge_type_str == "list") {
+    edge_type = EdgeType::List;
+    edges_in = pin->GetVector<Real>(block_name, prefix + "list");
+    //  required by binning index function
+    PARTHENON_REQUIRE_THROWS(std::is_sorted(edges_in.begin(), edges_in.end()),
+                             "Bin edges must be in order.");
+    PARTHENON_REQUIRE_THROWS(edges_in.size() >= 2,
+                             "Need at least one bin, i.e., two edges.");
```
Yurlungur: Does `is_sorted` fail if two edges are the same? Might want a check that edge 1 is strictly greater than edge 0, not just >=. Of course that's an edge (pun intended) case since this is floating point arithmetic. But I think it could happen with user-specified inputs.

src/outputs/histogram.cpp
```diff
@@ -0,0 +1,554 @@
...
+}
+
+Histogram::Histogram(ParameterInput *pin, const std::string &block_name,
+                     const std::string &prefix) {
+  ndim = pin->GetInteger(block_name, prefix + "ndim");
+  PARTHENON_REQUIRE_THROWS(ndim == 1 || ndim == 2, "Histogram dim must be '1' or '2'");
+
+  x_var_name = pin->GetString(block_name, prefix + "x_variable");
+  x_var_component = -1;
+  if (x_var_name == "HIST_COORD_X1") {
```
Yurlungur: The `x_var` and `y_var` parsing is almost identical. Wonder if it's worth abstracting into a little helper function?

src/outputs/histogram.cpp
```diff
@@ -0,0 +1,554 @@
...
+  } else {
+    x_var_type = VarType::Var;
+    x_var_component = pin->GetInteger(block_name, prefix + "x_variable_component");
+    // would add additional logic to pick it from a pack...
+    PARTHENON_REQUIRE_THROWS(x_var_component >= 0,
+                             "Negative component indices are not supported");
+  }
+
+  std::tie(x_edges, x_edges_type, x_edge_min, x_edge_dbin) =
+      GetEdges(pin, block_name, prefix + "x_edges_");
```
Yurlungur: Are you using `std::tie` here instead of structured binding because `x_edges` and friends are member variables of the class?

src/outputs/histogram.cpp
```diff
@@ -0,0 +1,554 @@
...
+    // would add additional logic to pick it from a pack...
+    PARTHENON_REQUIRE_THROWS(weight_var_component >= 0,
+                             "Negative component indices are not supported");
+  }
+}
+
+// Computes a 1D or 2D histogram with inclusive lower edges and inclusive rightmost edges.
+// Function could in principle be templated on dimension, but it's currently not expected
+// to be a performance concern (because it won't be called that often).
+void CalcHist(Mesh *pm, const Histogram &hist) {
```
Yurlungur: Not that it matters much, but I found it a little surprising that `CalcHist` is a free standing function, not a member function of `Histogram`. Either way is fine obviously.

src/utils/signal_handler.cpp
```diff
@@ -61,10 +64,8 @@ void SignalHandlerInit() {
 
 OutputSignal CheckSignalFlags() {
   if (Globals::my_rank == 0) {
-    // TODO(the person bumping std to C++17): use std::filesystem::exists
-    struct stat buffer;
     // if file "output_now" exists
-    if (stat("output_now", &buffer) == 0) {
+    if (fs::exists("output_now")) {
```
Yurlungur, line 68: :+1:

src/utils/sort.hpp
```diff
@@ -31,6 +31,28 @@
 
 namespace parthenon {
 
+// Returns the upper bound (or the array size if value has not been found)
+// Could/Should be replaced with a Kokkos std version once available (currently schedule
+// for 4.2 release).
+template <class T>
```
Yurlungur, line 40: Since you use `extent_int` this is not an arbitrary array type. Maybe a good opportunity for concepts?
```suggestion
template<class T, class = ENABLEIF(implements<kokkos_view(Data)>::value)>
```

src/utils/sort.hpp
```diff
@@ -31,6 +31,28 @@
...
+      l = m + 1;
+    } else {
+      r = m;
+    }
+  }
+  if (l < arr.extent_int(0) && val >= arr(l)) {
+    l++;
+  }
+  return l;
+}
```
Yurlungur, line 57: Hmm... so since this is essentially the serial implementation, this isn't really general right? Bad things happen if you reduce on this with more than one thread? I assume that this is never called in a performance-critical context, so that's fine? But still, I'd like to at least add some guard rails. I see a few options:
1. Add a big comment warning to only reduce on size 1
2. Remove device markings and implement a host-side version that calls `par_reduce` on this kernel with 1 thread
3. Implement a parallel version that can be properly reduced over... perhaps just the dumbest thing of a linear search with a `std::max` for the index

tst/regression/CMakeLists.txt
```diff
@@ -127,6 +127,7 @@ endif()
 # Any external modules that are required by python can be added to REQUIRED_PYTHON_MODULES
 # list variable, before including TestSetup.cmake.
 list(APPEND REQUIRED_PYTHON_MODULES numpy)
+list(APPEND REQUIRED_PYTHON_MODULES h5py)
```
Yurlungur: Was the lack of h5py dependency previously in error? Or is this a newly needed dependency?

src/outputs/histogram.cpp
```diff
@@ -0,0 +1,554 @@
...
+
+  const int num_partitions = pm->DefaultNumPartitions();
+
+  for (int p = 0; p < num_partitions; p++) {
+    auto &md = pm->mesh_data.GetOrAdd("base", p);
+
+    const auto x_var_pack_string = x_var_type == VarType::Var
+                                       ? std::vector<std::string>{hist.x_var_name}
+                                       : std::vector<std::string>{};
+    const auto x_var = md->PackVariables(x_var_pack_string);
```
Yurlungur, line 268: Not hugely important... but I wonder if it would be best to use the `SparsePack` machinery here? It *might* lead to a performance improvement if you built the `PackDescriptor`s as `static` variables, as that way they are only built once?

src/outputs/histogram.cpp
```diff
@@ -0,0 +1,554 @@
...
+            x_val = coords.Xc<1>(k, j, i);
+          } else if (x_var_type == VarType::X2) {
+            x_val = coords.Xc<2>(k, j, i);
+          } else if (x_var_type == VarType::X3) {
+            x_val = coords.Xc<3>(k, j, i);
+          } else if (x_var_type == VarType::R) {
+            x_val = Kokkos::sqrt(SQR(coords.Xc<1>(k, j, i)) + SQR(coords.Xc<2>(k, j, i)) +
+                                 SQR(coords.Xc<3>(k, j, i)));
+          } else {
+            x_val = x_var(b, x_var_component, k, j, i);
```
Yurlungur, line 305: worth noting that (I think?) this even works for tensor components, since you're using a pack, and that will flatten them out...

src/outputs/histogram.cpp
```diff
@@ -0,0 +1,554 @@
...
+                                      : binned_var(b, binned_var_component, k, j, i);
+          auto weight = weight_by_vol ? coords.CellVolume(k, j, i) : 1.0;
+          weight *= weight_var_component == -1
+                        ? 1.0
+                        : weight_var(b, weight_var_component, k, j, i);
+          res(y_bin, x_bin) += val_to_add * weight;
+        });
+    // "reduce" results from scatter view to original view. May be a no-op depending on
+    // backend.
+    Kokkos::Experimental::contribute(result.KokkosView(), scatter);
```
Yurlungur, line 400: Cool! I assume this is more performant than atomic ads?

Yurlungur (changes requested):
This is really cool! I'm very excited to take advantage of it. A few comments/questions below.

Also, should we have something in `phdf.py` for reading histograms? Maybe not as they seem to be pretty self-descriptive in file?

src/outputs/histogram.cpp
pgrete: Good point. I added the check.

src/utils/sort.hpp
pgrete: I'm not sure I follow.
This is an inline function that is called by each thread, does not modify any data, and returns a single value (at a dedicated memory location).
Where would that cause issues?

tst/regression/CMakeLists.txt
pgrete: I missed that we already have this check as part of the `if (ENABLE_HDF5)` conditional above. I removed this line again.

src/outputs/histogram.cpp
pgrete: It's because the variable names already exists.
With a structured binding we'd have to copy/move the result to the existing variables.

src/outputs/histogram.cpp
pgrete: done

src/outputs/histogram.cpp
pgrete: done

src/outputs/histogram.cpp
pgrete: The `ScatterView` transparently switches between using atomics and temporary arrays depending on the backend, so this is actually be a no-op if atomics were used inside the kernel.

src/outputs/histogram.cpp
pgrete: I think so, too, but in the absence of tests I wouldn't immediately trust it.

src/outputs/histogram.cpp
pgrete: Probably yes. I just went with what I'm used to and these parts are likely not performance critical anyway.
Do you feel strongly about moving to `SparsePacks`?

pgrete:
> Also, should we have something in `phdf.py` for reading histograms? Maybe not as they seem to be pretty self-descriptive in file?

There's an example in the doc on how to read and plot the data, e.g.,
```
with h5py.File("parthenon.out8.histograms.00040.hdf", "r") as infile:
  # 1D histogram
   x = infile["0/x_edges"][:]
   y = infile["0/data"][:]
   plt.plot(x, y)
   plt.show()
```
so I'm not sure what else we should provide from the `phdf` point of view.

src/utils/sort.hpp
pgrete: I'm probably using it wrong because when I tried adding this I got
```
[build] /home/pgrete/src/parthenon/src/outputs/histogram.cpp: In lambda function:
[build] /home/pgrete/src/parthenon/src/outputs/histogram.cpp:318:34: error: no matching function for call to ‘upper_bound(const parthenon::ParArrayGeneric<Kokkos::View<double*, Kokkos::LayoutRight, Kokkos::HostSpace>, parthenon::empty_state_t, int>&, double&)’
[build]   318 |               x_bin = upper_bound(x_edges, x_val) - 1;
[build]       |                       ~~~~~~~~~~~^~~~~~~~~~~~~~~~
[build] In file included from /home/pgrete/src/parthenon/src/outputs/histogram.cpp:53:
[build] /home/pgrete/src/parthenon/src/utils/sort.hpp:39:28: note: candidate: ‘template<class T, class> int parthenon::upper_bound(const T&, Real)’
[build]    39 | KOKKOS_INLINE_FUNCTION int upper_bound(const T &arr, Real val) {
[build]       |                            ^~~~~~~~~~~
[build] /home/pgrete/src/parthenon/src/utils/sort.hpp:39:28: note:   template argument deduction/substitution failed:
[build] /home/pgrete/src/parthenon/src/utils/sort.hpp:38:20: error: no type named ‘type’ in ‘struct std::enable_if<false, int>’
[build]    38 | template <class T, class = ENABLEIF(implements<kokkos_view(T)>::value)>
[build]       |                    ^~~~~
```

src/outputs/histogram.cpp
pgrete: Made it a member function.

doc/sphinx/src/outputs.rst
pgrete: That probably depends on the type of normalization.
Are you thinking of a PDF here?
I'm currently leaning towards letting normalizations be done during post processing so that we just provide "raw" data,
In case you agree, I already extended the doc to reflect that no normalization will be done.

Yurlungur: :+1: I think either way is fine.

src/utils/sort.hpp
Yurlungur: Oooh maybe `ParArrayNDGeneric` doesn't satisfy the `kokkos_view` concept? @lroberts36 which concept should @pgrete be using here?

src/utils/sort.hpp
Yurlungur: Doesn't this function only work if you call it on either 1 array per thread? The way you're calling it, you launch a single thread on the array. That's all I mean. If someone is expecting this to be some kind of `kokkos_scan` based function that leverages multiple threads, that isn't the case.

src/outputs/histogram.cpp
Yurlungur: No I don't feel strongly in this instance---it's not important. I'm mainly just thinking about this from the perspective of eventually moving away from two separate pack infrastructures.

src/outputs/histogram.cpp
Yurlungur: Neat! TIL

Yurlungur:
> I'm not sure what else we should provide from the `phdf` point of view.

I suspected as much. It's simple enough I don't think we need to provide anything else.

doc/sphinx/src/outputs.rst
pgrete: Hmm, not quite. I might need to update the documentation (I'm happy for suggestions to make it more clear).

When I wrote this/chose the names, I thought about it in terms of plots.
So for both 1D and 2D histograms the "x" variable is the one on the bottom axis, i.e., always used.
Then the "binned" variable is the "data" variable to be shown (or put into "bins"), i..e, for a 1D histogram this variable will correspond to the line (i.e., the y values) and for the 2D histogram this will be the "z" dim.
The "y" variable itself is only used for 2D histograms and is the variable on the left/y axis.

I realize this might be/is confusing, though I didn't come up with better naming scheme that works for 1D and 2D histograms while not changing the meaning of an input parameter.

Yurlungur: Hmm I'm still not sure I understand. Maybe a few examples would help me reason through it. In 1D, I often want to compute something like the top panel of this figure:
image
Note that each "curve" is basically "mass in a bin" vs bins of Ye. What variables should I pick in your framework to make this happen? My intuition would be `hist_x_variable` set to `electron_fraction`, `hist_binned_variable` set to density, and volume-weighting turned on... is that right?

pgrete:
I now pushed all the change I mentioned in the comments.
@BenWibking @Yurlungur feel free to close the conservation that you consider resolved (and potentially re-review).

Yurlungur:
> I now pushed all the change I mentioned in the comments. @BenWibking @Yurlungur feel free to close the conservation that you consider resolved (and potentially re-review).

Way ahead of you and already done ;)

doc/sphinx/src/outputs.rst
pgrete: I gave this some more thought and am now sure how we should go ahead.
Originally, I liked @Yurlungur idea to have a default file_index and multiple output blocks as it looked clean.
Though when thinking about the implementation, I realized that this might introduce some arbitrary ordering in the resulting output file depending on the order of the `output#` blocks and number of different `file_index` given, which may result in additional fine tuning/logic when going from input file to postprocessing.

As ugly as the currently solution is, it's from that point of view unambiguous. All histograms in one output block are in one file and in the same order they are defined.
So I'm currently leaning towards keeping this verbose option but am also open to further suggestions (or going with the solution that @Yurlungur suggested if people prefer that).

doc/sphinx/src/outputs.rst
pgrete: Exactly right.
And to add a 2D example,
image
Here, the x variable is radius, the y variable plasma beta, and the "data" is again mass (calculated from binning the density field and apply volume weighting).

[edit] the line is the mass weighted mean plasma beta derived from the histogram data a posteriori [/edit]

BenWibking (approved):
My comments have been addressed, so I'm happy with it now.

doc/sphinx/src/outputs.rst
Yurlungur: I'm obviously biased towards my own solution :P. I also think if one relies on ordering in an hdf5 file, one is asking for trouble.

That said, I don't feel that strongly and am fine to defer to implementers choice :)

doc/sphinx/src/outputs.rst
Yurlungur: Ok thanks for the explanation. This makes sense to me---although it might be useful to embed some of these examples in the docs to clarify. I think my head was stuck on what `matplotlib` requests of you when you build a histogram, which is slightly different, as it conflates some of these variables.

Yurlungur (approved):
I have a few minor concerns remaining, but I consider them non-blocking.

src/utils/sort.hpp
pgrete: The assumption here is that `arr` is not modified while the index/lower bound `l` is determined, so yes, when you use this function side a parallel scan where the `arr` is the result of the scan, then this will cause issue.
Though I'm not sure this (finding and index/doing a binary search in a changing array/container) is even legal for the `std::upper_bound`.
Or maybe I'm still not understanding the issue.

doc/sphinx/src/outputs.rst
pgrete: If it's okay to add these illustrative "not-pure-math/numerics" examples, I'm happy to add those to the docs.

src/utils/sort.hpp
Yurlungur: There's nothing technically wrong with the function, so I think my issue is more conceptual than anything. This function is exactly what you want for the histogram code, I'm just worried about it being misused since it lives in `utils`.

doc/sphinx/src/outputs.rst
Yurlungur: I guess I don't know for sure---but they're from published works right? It seems like it should be ok in the context of documentation? (Mine is from DOI 10.3847/2041-8213/acba16)

doc/sphinx/src/outputs.rst
pgrete: > I'm obviously biased towards my own solution :P. I also think if one relies on ordering in an hdf5 file, one is asking for trouble.

"Order" here referred to the explicit number used in the input file that is also being used as the hdf5 group name in the output file, so it's unique and does not depend on the literal order written down in the input file.

Having said that, it just occurred to me that we could enforce that each histogram has a unique string name (rather than a number) which is then used as prefix (both in the input file as well as the output file).

How about going from
```
num_histograms = 3

# 1D histogram of a variable, binned by a variable
hist0_ndim = 1
hist0_x_variable = advected
hist0_x_variable_component = 0
hist0_x_edges_type = list
hist0_x_edges_list = 1e-9, 1e-4,1e-1, 2e-1, 5e-1 ,1.0
hist0_binned_variable = advected
hist0_binned_variable_component = 0

# 2D histogram of a variable, binned by a coordinate and a different variable
hist1_ndim = 2
hist1_x_variable = HIST_COORD_X1
hist1_x_edges_type = lin
hist1_x_edges_num_bins = 4
...
```
to
```
hist_names = myname, rhoT

# 1D histogram of a variable, binned by a variable
myname_ndim = 1
myname_x_variable = advected
myname_x_variable_component = 0
myname_x_edges_type = list
myname_x_edges_list = 1e-9, 1e-4,1e-1, 2e-1, 5e-1 ,1.0
myname_binned_variable = advected
myname_binned_variable_component = 0

# 2D histogram of a variable, binned by a coordinate and a different variable
rhoT_ndim = 2
rhoT_x_variable = HIST_COORD_X1
rhoT_x_edges_type = lin
rhoT_x_edges_num_bins = 4
...
```

This (using names) would, of course, also work in the context of spreading the info over multiple output blocks (each with a unique name for a histogram), but I'm leaning towards keeping the current overall logic (one output block for one output file for a given dt) over adding additional content from different blocks.

What do people think?

src/utils/sort.hpp
pgrete: I put it specifically in `utils` so that it can be reused as drop in replacement for `https://en.cppreference.com/w/cpp/algorithm/upper_bound` on device until a Kokkos version is available.
So there should be no extra gotchas (as far as I'm aware of) in comparison to `std::upper_bound`.

doc/sphinx/src/outputs.rst
Yurlungur: I like the new convention with names instead of indices. :+1:

src/utils/sort.hpp
Yurlungur: If not already present, can you add a comment that it has exactly the same API as `std::upper_bound` but on device? I think that will disambiguate for me

src/utils/sort.hpp
lroberts36: Maybe this should really be operating on types that follow the `contiguous_container` concept? As written it appears that this requires a one-dimensional Kokkos view, but couldn't the `arr.extent_int(0)` instances be replaced with `arr.size()` and `arr(m)` be replaced with `arr[m]`? Then this should work for anything that is a contiguous container I believe (which both kokkos views and `ParArrayNDGeneric`s satisfy).

src/utils/sort.hpp
pgrete: Added a comment (and highlighted the differences).

doc/sphinx/src/outputs.rst
pgrete: I now added two examples.

doc/sphinx/src/outputs.rst
pgrete: New convention is implemented and documented.

pgrete:
I think I've addressed all the leftover discussed changes.
If there are no objections, I'll enable auto-merge tomorrow (and afterwards make a new Parthenon release).
Also feel free to enable auto-merge earlier, if you're happy this the changes.

PR #965, opened 2023-10-20
Allow leading whitespace in input parameters

brryan:
## PR Summary

Currently, input parameters are read incorrectly. For example,

```
<block>
 var = 1.0
 ```

will read `var` as `= 1.0`, which when called with `GetReal` will cast to `0`. 

This PR fixes a bug where the length of the name was incorrectly calculated if leading whitespace was present in the input parameter file. 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [ ] Docs build

### DIFF
--- a/src/parameter_input.cpp
+++ b/src/parameter_input.cpp
@@ -295,7 +295,6 @@ bool ParameterInput::ParseLine(InputBlock *pib, std::string line, std::string &n
                                std::string &value, std::string &comment) {
   std::size_t first_char, last_char, equal_char, hash_char, cont_char, len;
   bool continuation = false;
-  printf("line: %s\n", line.c_str());
 
   hash_char = line.find_first_of("#"); // find "#" (optional)
   comment = "";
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/parameter_input.cpp [hunk 1]
```diff
@@ -295,6 +295,7 @@ bool ParameterInput::ParseLine(InputBlock *pib, std::string line, std::string &n
                                std::string &value, std::string &comment) {
   std::size_t first_char, last_char, equal_char, hash_char, cont_char, len;
   bool continuation = false;
+  printf("line: %s\n", line.c_str());
```
Yurlungur: debug statement?

Yurlungur (approved):

src/parameter_input.cpp
brryan: yep, fixed

brryan:
BTW I tested this by adding a leading space character to `tlim` in the advection example

pdmullen (approved):

pgrete (approved):
LGTM
How about also adding the one-line test you mention (adding a space to one of the examples input files called by the test suite)?

brryan:
> LGTM How about also adding the one-line test you mention (adding a space to one of the examples input files called by the test suite)?

Good idea @pgrete, done (I changed `tst/regression/test_suites/advection_performance/parthinput.advection_performance` to check for both space and tab characters)

PR #966, opened 2023-10-23
Multigrid MPI Bug Fix

lroberts36:
## PR Summary

It turns out I never bothered to actually test MG with MPI. This PR makes changes to the MG PR for MG to work correctly with MPI. First, it makes some small changes so that empty `MeshData` objects don't cause errors and adds MPI communicators for GMG fields that are involved in restriction. Second, and more importantly, it corrects a bug in the task list that allowed certain parts of composite GMG levels to run ahead of coarser levels in the hierarchy. This resulted in boundary data not being up to date at coarse-fine boundaries when any rank only owned coarse blocks on a composite grid. The multi-grid task list now includes task list dependencies and regional dependencies to ensure that the coarser levels of the hierarchy have finished before the post prolongation steps take place on finer two-level composite grids. To ensure the correct ordering within task lists, `AddMultiGridTasksLevel` is now called recursively. I did *not* need to add extra buffers, which I had mentioned as a possible solution in the Parthenon sync last Thursday.

Once this is reviewed, I will merge it into the main multi-grid PR and then merge that PR into develop.

## PR Checklist

- [x] Code passes cpplint

### DIFF
--- a/example/poisson_gmg/poisson_driver.cpp
+++ b/example/poisson_gmg/poisson_driver.cpp
@@ -79,7 +79,8 @@ TaskID AddSRJIteration(TaskList &tl, TaskID depends_on, int stages, bool multile
       {{0.8723, 0.5395, 0.0000}, {1.3895, 0.5617, 0.0000}, {1.7319, 0.5695, 0.0000}}};
   std::array<std::array<Real, 3>, 3> omega_M3{
       {{0.9372, 0.6667, 0.5173}, {1.6653, 0.8000, 0.5264}, {2.2473, 0.8571, 0.5296}}};
-
+  
+  if (stages == 0) return depends_on;
   auto omega = omega_M1;
   if (stages == 2) omega = omega_M2;
   if (stages == 3) omega = omega_M3;
@@ -148,12 +149,13 @@ TaskID DotProduct(TaskID dependency_in, TaskRegion &region, TaskList &tl, int pa
   return finish_global_adotb;
 }
 
-void PoissonDriver::AddMultiGridTasksLevel(TaskRegion &region, int level, int min_level,
+TaskID PoissonDriver::AddMultiGridTasksLevel(TaskList &tl, TaskID dependency, int partition, 
+                                           int level, int min_level,
                                            int max_level, bool final) {
   using namespace parthenon;
   using namespace poisson_package;
   TaskID none(0);
-  const int num_partitions = pmesh->DefaultNumPartitions();
+  //const int num_partitions = pmesh->DefaultNumPartitions();
 
   auto pkg = pmesh->packages.Get("poisson_package");
   auto damping = pkg->Param<Real>("jacobi_damping");
@@ -161,7 +163,10 @@ void PoissonDriver::AddMultiGridTasksLevel(TaskRegion &region, int level, int mi
   int pre_stages = pkg->Param<int>("pre_smooth_iterations");
   int post_stages = pkg->Param<int>("post_smooth_iterations");
   bool do_FAS = pkg->Param<bool>("do_FAS");
-  if (smoother == "SRJ1") {
+  if (smoother == "none") {
+    pre_stages = 0;
+    post_stages = 0;
+  } else if (smoother == "SRJ1") {
     pre_stages = 1;
     post_stages = 1;
   } else if (smoother == "SRJ2") {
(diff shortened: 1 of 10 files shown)


example/poisson_gmg/poisson_package.hpp
```diff
@@ -62,10 +62,9 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin);
 template <class in, class out, bool only_md_level = false>
 TaskStatus CopyData(std::shared_ptr<MeshData<Real>> &md) {
   using TE = parthenon::TopologicalElement;
-  auto pmb = md->GetBlockData(0)->GetBlockPointer();
-  IndexRange ib = pmb->cellbounds.GetBoundsI(IndexDomain::entire, te);
-  IndexRange jb = pmb->cellbounds.GetBoundsJ(IndexDomain::entire, te);
-  IndexRange kb = pmb->cellbounds.GetBoundsK(IndexDomain::entire, te);
+  IndexRange ib = md->GetBoundsI(IndexDomain::entire, te);
+  IndexRange jb = md->GetBoundsJ(IndexDomain::entire, te);
+  IndexRange kb = md->GetBoundsK(IndexDomain::entire, te);
```
Yurlungur, line 67: :+1:

Yurlungur (approved):
LGTM

example/poisson_gmg/poisson_driver.cpp
```diff
@@ -176,94 +180,102 @@ void PoissonDriver::AddMultiGridTasksLevel(TaskRegion &region, int level, int mi
...
-      // 7. Correct solution on this level with res_err field and store in
-      //    communication field
-      auto update_sol =
-          tl.AddTask(prolongate, AddFieldsAndStore<u, res_err, u, true>, md, 1.0, 1.0);
-
-      // 8. Post smooth using communication field and stored RHS
-      post_smooth = AddSRJIteration<BoundaryType::gmg_same>(tl, update_sol, post_stages,
-                                                            multilevel, md);
+  auto &md = pmesh->gmg_mesh_data[level].GetOrAdd(level, "base", partition);
+  // 0. Receive residual from coarser level if there is one
```
pgrete: Is the comment not aligned with the code ("receive residual from coarse" versus "recv_from_finer") or am I missing sth?

example/poisson_gmg/poisson_driver.cpp
```diff
@@ -292,9 +304,12 @@ TaskCollection PoissonDriver::MakeTaskCollectionMG(BlockList_t &blocks) {
...
-    TaskRegion &region = tc.AddRegion(num_partitions * (max_level + 1));
-    for (int level = max_level; level >= min_level; --level) {
-      AddMultiGridTasksLevel(region, level, min_level, max_level, level == min_level);
+    int mg_reg_dep_id = 0;
+    TaskRegion &region = tc.AddRegion(num_partitions);
+    for (int i = 0; i < num_partitions; ++i) {
+      TaskList &tl = region[i];
+      AddMultiGridTasksLevel(region, tl, none, i, mg_reg_dep_id, max_level, min_level,
+                             max_level);
     }
```
pgrete, line 314: Just double checking: `mg_reg_dep_id` is passed in as reference and then incremented in that function so that for multiple partition, the same logical function are being put in different regions, correct?
And is that also the indented behavior (rather than having each all blocks in each partition being logically in the same region so that all blocks on the mesh progress region by region)?

pgrete (approved):
LGTM (just two clarification comments/questions)

Apart from the bugfix, I like that we don't rely on implicit `block[0]` info for `MeshData` any more (this ties `MeshData` closer to the `Mesh` than to the `MeshBlock`.

example/poisson_gmg/poisson_driver.cpp
lroberts36: oh, yeah, that is a typo

lroberts36:
```suggestion
  // 0. Receive residual from finer level if there is one
```

example/poisson_gmg/poisson_driver.cpp
lroberts36: I had just copied what was done in some other code for this where the `reg_dep_id` was initialized outside of the loop over partitions and then incremented for each partition. It is not clear to me how the regional dependence ids work under the hood, so maybe it is not correct as is. The desired behavior is to have "all blocks in each partition being logically in the same region so that all blocks on the mesh progress region by region".

pgrete: In that case, I suggest to defer the question to @jdolence who IIRC wrote the regional dependencies so that we don't run into unexpected issues with `pack_size != 1` or `!= -1`

PR #968, opened 2023-10-26
Allow for registering boundary conditions by package

lroberts36:
## PR Summary

This PR contains a proposal for adding the ability to register boundary conditions on a per package basis. I am not wedded to this setup, but it was easy to implement and I think provides the functionality that I foresee needing myself. As written, boundary conditions will behave exactly as they used to unless boundary condition functions are registered within a package. 

Per package boundary condition registration is performed by adding to the `UserBoundaryFunctions` vectors:
```c++
template <CoordinateDirection DIR, BCSide SIDE>
auto GetMyBC() {
  return [](std::shared_ptr<MeshBlockData<Real>> &rc, bool coarse) -> void {
    // Implementation of BC here
  };
}

std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
  ...
  using BF = parthenon::BoundaryFace;
  pkg->UserBoundaryFunctions[BF::inner_x1].push_back(GetMyBC<X1DIR, BCSide::Inner>());
  pkg->UserBoundaryFunctions[BF::inner_x2].push_back(GetMyBC<X2DIR, BCSide::Inner>());
  ...
}
```
At package resolution, all of the boundary conditions functions from different packages are put into a vector of function pointers and when physical boundary conditions are called these functions are just sequentially called after calling `Mesh::MeshBndryFnctn`. One clear drawback of the current approach is that if multiple user registered boundary conditions update the same variable, its boundary values will be determined by the last boundary function operating on it that is called (i.e. there is no easy way to set the precedence of boundary functions).

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/src/interface/state_descriptor.cpp
+++ b/src/interface/state_descriptor.cpp
@@ -425,6 +425,12 @@ StateDescriptor::CreateResolvedStateDescriptor(Packages_t &packages) {
     // sort
     field_tracker.CategorizeCollection(name, field_dict, &field_provider);
     swarm_tracker.CategorizeCollection(name, package->AllSwarms(), &swarm_provider);
+
+    // Add package registered boundary conditions
+    for (int i = 0; i < 6; ++i)
+      state->UserBoundaryFunctions[i].insert(state->UserBoundaryFunctions[i].end(),
+                                             package->UserBoundaryFunctions[i].begin(),
+                                             package->UserBoundaryFunctions[i].end());
   }
 
   // check that dependent variables are provided somewhere
--- a/src/interface/state_descriptor.hpp
+++ b/src/interface/state_descriptor.hpp
@@ -431,7 +431,7 @@ class StateDescriptor {
 
   friend std::ostream &operator<<(std::ostream &os, const StateDescriptor &sd);
 
-  std::array<std::vector<BValFunc>, 6> UserBoundaryFunctions;
+  std::array<std::vector<BValFunc>, BOUNDARY_NFACES> UserBoundaryFunctions;
 
  private:
   void InvertControllerMap();
--- a/src/interface/state_descriptor.hpp
+++ b/src/interface/state_descriptor.hpp
@@ -44,6 +44,8 @@ class MeshBlockData;
 template <typename T>
 class MeshData;
 
+using BValFunc = std::function<void(std::shared_ptr<MeshBlockData<Real>> &, bool)>;
+
 /// A little container class owning refinement function properties
 /// needed for the state descriptor.
 /// Note using VarID here implies that custom prolongation/restriction
@@ -428,6 +430,8 @@ class StateDescriptor {
   std::function<void(MeshBlockData<Real> *rc)> InitNewlyAllocatedVarsBlock = nullptr;
 
   friend std::ostream &operator<<(std::ostream &os, const StateDescriptor &sd);
+
+  std::array<std::vector<BValFunc>, 6> UserBoundaryFunctions;
 
  private:
   void InvertControllerMap();
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Current develop
 
 ### Added (new features/APIs/variables/...)
+- [[PR 968]](https://github.com/parthenon-hpc-lab/parthenon/pull/968) Add per package registration of boundary conditions
 - [[PR 948]](https://github.com/parthenon-hpc-lab/parthenon/pull/948) Add solver interface and update Poisson geometric multi-grid example 
 - [[PR 911]](https://github.com/parthenon-hpc-lab/parthenon/pull/911) Add infrastructure for geometric multi-grid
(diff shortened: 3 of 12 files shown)


src/interface/state_descriptor.hpp
```diff
@@ -44,6 +44,8 @@ class MeshBlockData;
 template <typename T>
 class MeshData;
 
+using BValFunc = std::function<void(std::shared_ptr<MeshBlockData<Real>> &, bool)>;
```
Yurlungur, line 47: Was this not declared somewhere previously? I'm super surprised by that... (Or is this a forward declaration?

src/interface/state_descriptor.cpp
```diff
@@ -425,6 +425,12 @@ StateDescriptor::CreateResolvedStateDescriptor(Packages_t &packages) {
     // sort
     field_tracker.CategorizeCollection(name, field_dict, &field_provider);
     swarm_tracker.CategorizeCollection(name, package->AllSwarms(), &swarm_provider);
+
+    // Add package registered boundary conditions
+    for (int i = 0; i < 6; ++i)
+      state->UserBoundaryFunctions[i].insert(state->UserBoundaryFunctions[i].end(),
+                                             package->UserBoundaryFunctions[i].begin(),
+                                             package->UserBoundaryFunctions[i].end());
```
Yurlungur, line 433: Is order preserved in an individual package's set of boundary conditions? It looks like it is?

src/interface/state_descriptor.hpp
```diff
@@ -429,6 +431,8 @@ class StateDescriptor {
 
   friend std::ostream &operator<<(std::ostream &os, const StateDescriptor &sd);
 
+  std::array<std::vector<BValFunc>, 6> UserBoundaryFunctions;
```
Yurlungur: I think there's a constexpr name for number of faces somewhere... might be better than 6? Might not be?

Yurlungur (approved):
I'm a fan of this. This was something I've been wanting to add for some time now and just never sat down and did it. Glad you did it instead. :)

I particularly like the innovation of allowing each package to register *multiple* boundary conditions, as I think that's a clean way of ensuring all the flexibility that could possibly be needed.

While we're poking about the boundary condition machinery, can we add a function that takes `MeshData` instead of `MeshBlockData` even if it just loops over blocks? That way we could put it in an "all MeshData" task list.

src/interface/state_descriptor.hpp
lroberts36: It is declared in `boundary_conditions.hpp`, but when I included that I ran into all sorts of other circular dependency issues. I think this isn't technically a forward declaration, so maybe there is the possibility of a collision in the future?

src/interface/state_descriptor.cpp
lroberts36: Yes, I believe this should preserve the ordering within a package.

src/interface/state_descriptor.hpp
lroberts36: yeah, `BOUNDARY_NFACES` is probably a better choice (although I can't imagine that we will change that number).

lroberts36:
@Yurlungur: We already provide MD boundary conditions in `boundary_conditions.cpp` 
```c++
TaskStatus ApplyBoundaryConditionsOnCoarseOrFineMD(std::shared_ptr<MeshData<Real>> &pmd,
                                                   bool coarse) {
  for (int b = 0; b < pmd->NumBlocks(); ++b)
    ApplyBoundaryConditionsOnCoarseOrFine(pmd->GetBlockData(b), coarse);
  return TaskStatus::complete;
}
```
Are you thinking of something different than this?

src/interface/state_descriptor.hpp
lroberts36: changed.

src/interface/state_descriptor.hpp
Yurlungur: :shrug: putting this here seems like a perfectly reasonable solution.

Yurlungur:
> @Yurlungur: We already provide MD boundary conditions in `boundary_conditions.cpp`
> 
> ```c++
> TaskStatus ApplyBoundaryConditionsOnCoarseOrFineMD(std::shared_ptr<MeshData<Real>> &pmd,
>                                                    bool coarse) {
>   for (int b = 0; b < pmd->NumBlocks(); ++b)
>     ApplyBoundaryConditionsOnCoarseOrFine(pmd->GetBlockData(b), coarse);
>   return TaskStatus::complete;
> }
> ```
> 
> Are you thinking of something different than this?

Nope that's what I was thinking of.I Just forgot it already existed.

pdmullen (approved):
As discussed, I think future extensions might permit better interfacing with user-defined boundary conditions via the problem generator.   @bprather (and perhaps @jonahm-LANL) think that this nestles neatly into your design choice here, if the problem generator itself was a unique package.  

In the interim, this looks ready to go in to me.

Yurlungur:
This has been ready to go for a while. I'm clicking the button.

PR #971, opened 2023-11-06
Add hook for UserWorkBeforeLoop

Yurlungur:
## PR Summary

What it says on the can... This kind of workflow has come up a number of times for us in downstream applications, so I decided to finally stick it into parthenon. The basic need here is that you want to modify some variables after the mesh has been generated and problem generator has been called, for example, because you want to compute some global reduction variable, you want to normalize some field, you want to apply atmospheres, etc.

Now this can be done within a problem generator in some cases, for example with mesh-level problem generators. But it may be desired to do this work for *every* problem (for example). I add a hook in `ApplicationInputs` and in `StateDescriptor`. As usual, you assign a function pointer, and then it gets called by `EvolutionDriver`.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/doc/sphinx/src/interface/state.rst
+++ b/doc/sphinx/src/interface/state.rst
@@ -112,6 +112,13 @@ several useful features and functions.
   deletgates to the ``std::function`` member ``PostStepDiagnosticsMesh``
   if set (defaults to ``nullptr`` an therefore a no-op) to print
   diagnostics after the time-integration advance
+- ``void UserWorkBeforeLoopMesh(Mesh *, ParameterInput *pin, SimTime
+  &tm)`` performs a per-package, mesh-wide calculation after the mesh
+  has been generated, and problem generators called, but before any
+  time evolution. This work is done both on first initialization and
+  on restart. If you would like to avoid doing the work upon restart,
+  you can check for the const ``is_restart`` member field of the ``Mesh``
+  object.
 
 The reasoning for providing ``FillDerived*`` and ``EstimateTimestep*``
 function pointers appropriate for usage with both ``MeshData`` and
--- a/example/advection/advection_package.cpp
+++ b/example/advection/advection_package.cpp
@@ -1,5 +1,5 @@
 //========================================================================================
-// (C) (or copyright) 2020-2021. Triad National Security, LLC. All rights reserved.
+// (C) (or copyright) 2020-2023. Triad National Security, LLC. All rights reserved.
 //
 // This program was produced under U.S. Government contract 89233218CNA000001 for Los
 // Alamos National Laboratory (LANL), which is operated by Triad National Security, LLC
@@ -217,6 +217,7 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
   }
   pkg->CheckRefinementBlock = CheckRefinement;
   pkg->EstimateTimestepBlock = EstimateTimestepBlock;
+  pkg->UserWorkBeforeLoopMesh = AdvectionGreetings;
 
   return pkg;
 }
@@ -224,7 +225,8 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
 void AdvectionGreetings(Mesh *pmesh, ParameterInput *pin, parthenon::SimTime &tm) {
   if (parthenon::Globals::my_rank == 0) {
     std::cout << "Hello from the advection package in the advection example!\n"
-              << "This run is a restart: " << pmesh->is_restart << std::endl;
+              << "This run is a restart: " << pmesh->is_restart << "\n"
+              << std::endl;
   }
 }
 
--- a/example/advection/advection_package.cpp
+++ b/example/advection/advection_package.cpp
@@ -13,12 +13,14 @@
 
 #include <algorithm>
 #include <cmath>
+#include <iostream>
 #include <limits>
 #include <memory>
 #include <string>
 #include <vector>
 
 #include <coordinates/coordinates.hpp>
+#include <globals.hpp>
 #include <parthenon/package.hpp>
 
 #include "advection_package.hpp"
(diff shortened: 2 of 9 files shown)


example/advection/advection_package.cpp
```diff
@@ -219,6 +221,14 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
   return pkg;
 }
 
+void AdvectionGreetings(Mesh *pmesh, ParameterInput *pin, SimTime &tm) {
+  if (GLobals::my_rank == 0) {
```
brryan: `GLobals` is a typo?

example/advection/advection_package.hpp
```diff
@@ -21,6 +21,7 @@ namespace advection_package {
 using namespace parthenon::package::prelude;
 
 std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin);
+  void AdvectionGreetings(Mesh *pmes, ParameterInput *pin, SimTime &tm);
```
brryan: Extra leading whitespace -- needs formatting?

brryan (approved):
This is a nice addition, will clean up some downstream code :)

Maybe just check formatting though.

Yurlungur:
> Maybe just check formatting though.

Thanks for the catches. Fixed. :)

doc/sphinx/src/interface/state.rst
```diff
@@ -112,6 +112,13 @@ several useful features and functions.
   deletgates to the ``std::function`` member ``PostStepDiagnosticsMesh``
   if set (defaults to ``nullptr`` an therefore a no-op) to print
   diagnostics after the time-integration advance
+- ``void UserWorkBeforeLoopMesh(Mesh *, ParameterInput *pin, SimTime
```
pdmullen, line 115: Before reading further, this naming (i.e., `UserWorkBeforeLoopMesh`) made me think that this was going to be called before every `Step()` cycle.  Does anybody else see this as a potential confusion? How about `UserInitFinalize` or something more indicative of its utility?

pdmullen (approved):

doc/sphinx/src/interface/state.rst
Yurlungur: I would find `UserInitFinalize` to be more confusing personally... 

I *think* the current name is unambiguous, as there's already a `PreStepUserWorkInLoop` and a `PostStepUserWorkInLoop`, which is work done per step. And the current name fits more in that scheme. But I don't feel strongly.

src/mesh/mesh.hpp
```diff
@@ -94,6 +94,7 @@ class Mesh {
 
   // data
   bool modified;
+  const bool is_restart;
```
Yurlungur, line 97: I wanted people to be able to easily do different things upon restart vs first initialization, but without having to register two separate, maybe almost identical functions. There might be a better way than storing restart state in the mesh pointer, but it also seemed natural as something inherited from Athena++.

src/interface/state_descriptor.hpp
```diff
@@ -406,6 +406,10 @@ class StateDescriptor {
     if (InitNewlyAllocatedVarsBlock != nullptr) return InitNewlyAllocatedVarsBlock(rc);
   }
 
+  void UserWorkBeforeLoop(Mesh *pmesh, ParameterInput *pin, SimTime &tm) const {
+    if (UserWorkBeforeLoopMesh != nullptr) return UserWorkBeforeLoopMesh(pmesh, pin, tm);
+  }
```
Yurlungur, line 411: This copies the design of the other function hooks in state descriptor, but we could cut out the middle and just check for `nullptr` in the function pointer itself.

src/driver/driver.cpp
```diff
@@ -75,6 +75,16 @@ DriverStatus EvolutionDriver::Execute() {
...
 
+  // Before loop do work
+  // App input version
+  if (app_input->UserWorkBeforeLoop != nullptr) {
+    app_input->UserWorkBeforeLoop(pmesh, pinput, tm);
+  }
+  // packages version
+  for (auto &[name, pkg] : pmesh->packages.AllPackages()) {
+    pkg->UserWorkBeforeLoop(pmesh, pinput, tm);
+  }
```
Yurlungur: Many of the older function hooks store the function pointers in the mesh. I think this is a holdover from Athena++ and in parthenon, it's more natural to just query the app input or the state descriptor directly.

Yurlungur:
Does anyone from the AthenaPK team want to take a look at this before I merge it? Changes are pretty minor and just value-add, so maybe not necessary. @BenWibking @pgrete @forrestglines .

src/driver/driver.cpp
```diff
@@ -75,6 +75,16 @@ DriverStatus EvolutionDriver::Execute() {
   // Defaults must be set across all ranks
   DumpInputParameters();
 
+  // Before loop do work
```
pgrete, line 78: How about adding some profiling regions around (just for completeness)?

src/mesh/mesh.hpp
pgrete: What we've typically been doing is to check for `tm.cycle != 0`, which should also be possible in the new callbacks given that the `SimTime` object is passed.

pgrete (approved):
LGTM

We already used something similar in AthenaPK, but realized it as time or cycle based conditional within the `PreStep` callback.

src/mesh/mesh.hpp
Yurlungur: Good point---`is_restart` feels cleaner/less ambiguous to me, but I'm fine to remove this and tell people to use `tm.cycle` if preferred.

src/driver/driver.cpp
Yurlungur: Good point. Done. Sorry @jdolence this means that if my PR goes in before #969 there will be one exposed profiling region in here that we would either need to ignore or change.

PR #973, opened 2023-11-08
MG performance upgrades

lroberts36:
## PR Summary

This PR makes some changes to make things go faster, including:

1. Caches `PackDescriptor`s for boundary conditions generic using a map. At least in MG, re-creating pack descriptors every call was a significant performance hit. 
2. Re-arranges internal fields created by solvers to minimize the number of fields that share ghost data.
3. Adds mesh data for MG communication that only includes internal fields (so that fields from user packages are not communicated unnecessarily during MG solves).
4. Adds hierarchical parallelism in some of the MG utilities.
5. Adds a setup phase for MG to reduce the number of fields that are restricted.
6. Adds some solver parameters.

This breaks downstream codes that use a solver, since now solvers require a call to `AddSetupTasks` before calling `AddTasks`. 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


src/bvals/boundary_conditions_generic.hpp
```diff
@@ -34,6 +37,35 @@ namespace BoundaryFunction {
...
+namespace impl {
+using desc_key_t = std::tuple<bool, TopologicalType>;
+template <class... var_ts>
+using map_bc_pack_descriptor_t =
+    std::unordered_map<desc_key_t, typename SparsePack<var_ts...>::Descriptor,
+                       tuple_hash<desc_key_t>>;
+
+template <class... var_ts>
+map_bc_pack_descriptor_t<var_ts...>
+GetPackDescriptorMap(std::shared_ptr<MeshBlockData<Real>> &rc) {
```
Yurlungur, line 49: Is this to access all the different topological elements?

src/solvers/bicgstab_solver.hpp
```diff
@@ -78,11 +86,20 @@ class BiCGSTABSolver {
     pkg->AddField(p::name(), m_no_ghost);
   }
 
+  template <class TL_t>
+  TaskID AddSetupTasks(TL_t &tl, TaskID dependence, int partition, Mesh *pmesh) {
+    return preconditioner.AddSetupTasks(tl, dependence, partition, pmesh);
+  }
```
Yurlungur, line 92: Why is this indirection necessary?

Yurlungur (approved):
I think it would be useful to better document the solver parameters and how to use the solver more generically. But overall this LGTM.

src/bvals/boundary_conditions_generic.hpp
lroberts36: Yeah, the way the boundary conditions work we call `GenericBC` for each type of topological element. This just builds a map from (topological element, coarse) to the corresponding pack descriptor so that the map can be `static` in `GenericBC` and pack descriptors only get built once. Rebuilding the pack descriptors was a significant performance hit for MG since the boundary stuff gets called a lot.

src/solvers/bicgstab_solver.hpp
lroberts36: I don't know if it is necessary, but the idea was that solvers would have a generic sort of interface that included `AddSetupTasks`. There is no setup required for BiCGStab, but it is possible that you would need to do some setup for the outer solver and some setup for the inner preconditioner.

lroberts36:
> I think it would be useful to better document the solver parameters and how to use the solver more generically. 

Agreed, but I am not sure that the dust has settled on the solver stuff yet. I mostly want to get these changes in now so the branch doesn't go stale.

src/bvals/boundary_conditions_generic.hpp
Yurlungur: :+1:

src/solvers/bicgstab_solver.hpp
Yurlungur: I see. That makes sense. :+1:

src/interface/sparse_pack.hpp
```diff
@@ -202,6 +201,27 @@ class SparsePack : public SparsePackBase {
       return SparsePack(SparsePackBase::GetPack(pmd, *this, include_block));
     }
 
+    template <class T>
+    SparsePack GetPack(T *pmd, bool only_fine_two_level_composite_blocks = true) const {
```
pdmullen, line 205: Why is this additional function better than just passing the default argument to `include_block` above?

src/solvers/bicgstab_solver.hpp
```diff
@@ -251,9 +275,9 @@ class BiCGSTABSolver {
           solver->residual.val = 0.0;
           return TaskStatus::iterate;
         },
-        this, pmesh, params_.residual_tolerance);
+        this, pmesh, params_.max_iters, ptol);
 
-    return solver_id;
+    return tl.AddTask(solver_id, CopyData<x, u>, md);
```
pdmullen, line 280: In the other included examples, I thought that an iterative TaskList should always finally return the `solver_id` (as opposed to the completion task or any other task).  Should `CopyData` be put inside the `CompletionTask` (if iterations complete), and then maintain `return solver_id;`?

src/solvers/solver_utils.hpp
```diff
@@ -146,34 +148,38 @@ struct Stencil {
...
-        //            pull the loop over vars outside of the innermost loop to promote
-        //            vectorization.
-        const int nvars = pack.GetUpperBound(b, in()) - pack.GetLowerBound(b, in()) + 1;
-        for (int c = 0; c < nvars; ++c)
-          pack(b, te, out(c), k, j, i) = pack(b, te, in(c), k, j, i);
+  static auto desc = parthenon::MakePackDescriptor<in_t, out_t>(md.get());
+  auto pack = desc.GetPack(md.get(), only_fine_on_composite);
+  const int scratch_size = 0;
+  const int scratch_level = 0;
+  parthenon::par_for_outer(
```
pdmullen, line 165: Do you suspect that here a `par_for_outer` is more performant than a flat `par_for` with a standard `for` loop over `nvars` in the kernel?

src/solvers/solver_utils.hpp
```diff
@@ -182,25 +188,29 @@ TaskStatus AddFieldsAndStoreInteriorSelect(const std::shared_ptr<MeshData<Real>>
...
-  parthenon::par_for(
-      DEFAULT_LOOP_PATTERN, "AddFieldsAndStore", DevExecSpace(), 0, pack.GetNBlocks() - 1,
-      kb.s, kb.e, jb.s, jb.e, ib.s, ib.e,
-      KOKKOS_LAMBDA(const int b, const int k, const int j, const int i) {
-        // TODO(LFR): If this becomes a bottleneck, exploit hierarchical parallelism and
-        //            pull the loop over vars outside of the innermost loop to promote
-        //            vectorization.
+  const int scratch_size = 0;
+  const int scratch_level = 0;
+  parthenon::par_for_outer(
```
pdmullen, line 205: same Q as above.

pdmullen (approved):

src/interface/sparse_pack.hpp
lroberts36: IIRC, this structure saved creating a bunch of unnecessary intermediate vectors, which impacted MG performance to some extent.

src/solvers/bicgstab_solver.hpp
lroberts36: The way BiCGStab is now structured (which minimizes the number of fields that are communicated) requires adding a non-iterative task after the iterations (here copying the iteration output into the field defined by the downstream code). Notice that this task is in task list `tl` not `itl`.

src/solvers/solver_utils.hpp
lroberts36: At least on CPU, the inner loop should vectorize. On GPU, I think this should minimize the number of kernel launches. Also, it minimizes the number of times we call the type based accessor, which requires more indirection than directly accessing the pointer.

src/solvers/bicgstab_solver.hpp
```diff
@@ -175,7 +194,8 @@ class BiCGSTABSolver {
     }
 
     // 7. t <- A u
-    auto pre_t_comm = AddBoundaryExchangeTasks<BoundaryType::any>(precon2, itl, md, true);
+    auto pre_t_comm =
+        AddBoundaryExchangeTasks<BoundaryType::any>(precon2, itl, md_comm, multilevel);
```
pgrete, line 198: Does this (differentiating between with uniform grid an multilevel runs) fix a previous bug or add new feature/capability?

src/solvers/mg_solver.hpp
```diff
@@ -97,7 +111,8 @@ class MGSolver {
         &iter_counter);
     auto mg_finest = AddLinearOperatorTasks(itl, none, partition, pmesh);
     auto &md = pmesh->mesh_data.GetOrAdd("base", partition);
-    auto calc_pointwise_res = eqs_.template Ax<u, res_err>(itl, mg_finest, md);
+    auto comm = AddBoundaryExchangeTasks<BoundaryType::any>(mg_finest, itl, md, true);
```
pgrete: Why is `multilevel=true` hardcoded again (contrary to the inverse change above)?

src/solvers/mg_solver.hpp
```diff
@@ -126,13 +141,24 @@ class MGSolver {
...
 
     return AddMultiGridTasksPartitionLevel(tl, dependence, partition, max_level,
                                            min_level, max_level, pmesh);
   }
 
+  template <class TL_t>
+  TaskID AddSetupTasks(TL_t &tl, TaskID dependence, int partition, Mesh *pmesh) {
+    using namespace utils;
+
+    int min_level = 0;
```
pgrete: Why does min level here vary from the min level few lines above?

src/solvers/mg_solver.hpp
```diff
@@ -164,49 +187,87 @@ class MGSolver {
...
+            for (int c = 0; c < nvars; ++c) {
+              Real *Ax = &pack(b, te, Axold_t(c), k, jb.s, ib.s);
+              Real *diag = &pack(b, te, D_t(c), k, jb.s, ib.s);
+              Real *prhs = &pack(b, te, rhs_t(c), k, jb.s, ib.s);
+              Real *xo = &pack(b, te, xold_t(c), k, jb.s, ib.s);
+              Real *xn = &pack(b, te, xnew_t(c), k, jb.s, ib.s);
+              const int npoints =
+                  (jb.e - jb.s + 1) * (ib.e - ib.s + 1 + 2 * nghost) - 2 * nghost;
+              parthenon::par_for_inner(
+                  DEFAULT_INNER_LOOP_PATTERN, member, 0, npoints - 1, [&](const int idx) {
```
pgrete: Am I reading this correctly in the sense that includes more points than necessary (effectively going over the `i` ghost zones, too) to create a 1D loop?

src/solvers/solver_utils.hpp
```diff
@@ -146,34 +148,38 @@ struct Stencil {
...
+      DEFAULT_OUTER_LOOP_PATTERN, "CopyData", DevExecSpace(), scratch_size, scratch_level,
+      0, pack.GetNBlocks() - 1, KOKKOS_LAMBDA(parthenon::team_mbr_t member, const int b) {
+        const int nvars =
+            pack.GetUpperBound(b, in_t()) - pack.GetLowerBound(b, in_t()) + 1;
+        const int npoints = (kb.e - kb.s + 1) * (jb.e - jb.s + 1) * (ib.e - ib.s + 1);
+        for (int c = 0; c < nvars; ++c) {
+          Real *in = &pack(b, te, in_t(c), kb.s, jb.s, ib.s);
+          Real *out = &pack(b, te, out_t(c), kb.s, jb.s, ib.s);
+          parthenon::par_for_inner(DEFAULT_INNER_LOOP_PATTERN, member, 0, npoints - 1,
+                                   [&](const int idx) { out[idx] = in[idx]; });
```
pgrete: This caught my eye on first read. Might be worth adding a warning that this only works because `IndedDomain` is `entire`.
Also given that this is fixed for the kernel, might be worth to move `npoint =` outside the loop.

src/solvers/solver_utils.hpp
```diff
@@ -182,25 +188,29 @@ TaskStatus AddFieldsAndStoreInteriorSelect(const std::shared_ptr<MeshData<Real>>
...
-        //            pull the loop over vars outside of the innermost loop to promote
-        //            vectorization.
+  const int scratch_size = 0;
+  const int scratch_level = 0;
+  parthenon::par_for_outer(
+      DEFAULT_OUTER_LOOP_PATTERN, "AddFieldsAndStore", DevExecSpace(), scratch_size,
+      scratch_level, 0, pack.GetNBlocks() - 1,
+      KOKKOS_LAMBDA(parthenon::team_mbr_t member, const int b) {
         const int nvars = pack.GetUpperBound(b, a_t()) - pack.GetLowerBound(b, a_t()) + 1;
+        const int npoints = (kb.e - kb.s + 1) * (jb.e - jb.s + 1) * (ib.e - ib.s + 1);
```
pgrete: same as above

pgrete (approved):
Just a couple of clarifying questions, so feel free to merge when ready.

Otherwise, I second @Yurlungur comment on documentation ;)

src/solvers/bicgstab_solver.hpp
lroberts36: It is just a performance enhancement. Previously, every `FillGhost` variable was being communicated for every MG communication. Now it is only the fields used by MG.

src/solvers/mg_solver.hpp
lroberts36: Yeah, that is right. This is the pattern we use in Riot all over the place.

Yurlungur: Would be a good opportunity to use `IndexSplit` since we upstreamed that. :)

lroberts36: I think we would need to update `IndexSplit` to deal with differing topological types, but this made me realize the code as written has a bug for non-cell centered topological types as well. That is now fixed.

src/solvers/mg_solver.hpp
lroberts36: Ah, good catch. I have changed that to `pmesh->multilevel`. I don't think it hurt to have `multilevel=true` always, but did imply a few extra flops for uniform meshes.

src/solvers/mg_solver.hpp
lroberts36: It should be the same, I have changed it. Once again, this shouldn't produce incorrect results but does require a little more calculation potentially.

src/solvers/solver_utils.hpp
lroberts36: fixed and commented.

src/solvers/solver_utils.hpp
lroberts36: fixed and commented.

PR #974, opened 2023-11-08
make get parent pointer consistent

Yurlungur:
## PR Summary

For the longest time now, `MeshData::GetParentPointer` has returned a `Mesh*` object, while `MeshBlockData::GetParentPointer` has returned a `std::shared_ptr<MeshBlock>`. This inconsistency is annoying to deal with in templated code that uses both levels of the abstraction hierarchy. It's also bad practice to pass around `std::shared_ptr<T>&` everywhere, which this return value encourages.

In this PR I Make `MeshBlockData::GetParentPointer` return a `MeshBlock*`. This required a few things:
1. To walk around the `weak_ptr` logic used for `MeshBlockData`, I needed to add a funciton that still returned a shared pointer. So there are now two underlying fundamental functions: `GetBlockSharedPointer` and `GetBlockPointer`.
2. I needed change all of the code that calls either `GetBlockPointer` or `GetParentPointer` to expect a `MeshBlock*` instead of a `std::shared_ptr<MeshBlock>`. The main places this showed up are in boundary comms.

This is a very small change that touches a lot of code, but it should be a quick review. @lroberts36 I mostly touched your code, so I think it's important you review this one.

Note that this *may* be a breaking change downstream, but the fix would be as simple as find-replace `std::shared_ptr<MeshBlock>` with `MeshBlock*` and maybe not even that if you're using `auto`.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


src/bvals/comms/build_boundary_buffers.cpp
```diff
@@ -44,70 +44,69 @@ template <BoundaryType BTYPE>
...
-                }
-                return buf_t(chunk, std::make_pair(0, buf_size));
-              })));
-        }
-
-        const int receiver_rank = nb.snb.rank;
-        const int sender_rank = Globals::my_rank;
-
-        int tag = 0;
+  ForEachBoundary<BTYPE>(md, [&](auto pmb, sp_mbd_t /*rc*/, nb_t &nb, const sp_cv_t v) {
```
lroberts36, line 47: I didn't carefully check, but it looks like aside from switching to pointers all the changes in this file are just whitespace related?

lroberts36 (approved):
This LGTM, I completely support switching to raw pointers where functions don't need to take ownership. Thanks for going through and making those changes, I am not sure why I ended up going with smart pointers in the boundary stuff in the first place.

src/bvals/comms/build_boundary_buffers.cpp
Yurlungur: Yeah git and clang-format somehow interacted to cause an ugly looking diff. This happened when I merged your multigrid branch in. Not sure why. But the only thing I changed in, e.g., the `ForEachBoundary` call was replace `std::shared_ptr<MeshBlock>` with `auto` in the lambda.

pgrete:
Given that this is potentially API breaking, should we wait for the next release or should we already get this in (and I'll update the changelog when I make the release)?

Yurlungur:
> Given that this is potentially API breaking, should we wait for the next release or should we already get this in (and I'll update the changelog when I make the release)?

Ah good point. I say make a release first then this goes in immediately after release.

brryan:
Since this is a breaking change dealing with how we manage pointers (and removing `shared_ptr`s from interfaces), is it worth considering whether we should remove the usage of `weak_ptr` entirely in this PR? As far as I'm aware we're always in the situation @lroberts36 described "where functions don't need to take ownership." We have about 25 `std::weak_ptr::lock()`'s in the code and I'm not sure they're contributing anything.

Yurlungur:
> Since this is a breaking change dealing with how we manage pointers (and removing `shared_ptr`s from interfaces), is it worth considering whether we should remove the usage of `weak_ptr` entirely in this PR? As far as I'm aware we're always in the situation @lroberts36 described "where functions don't need to take ownership." We have about 25 `std::weak_ptr::lock()`'s in the code and I'm not sure they're contributing anything.

IMO the weak pointers don't buy us a whole lot, but they're more defensible in the internals of the code. The main place they're used is as back-pointers for objects owned by `MeshBlock`, such as `bvals` and `variable`. I think in user code, those back-pointers are rarely used. The weak pointers buy us (technically) some memory safety as they're non-cyclical but reference counted. They cost us a little user-invisible complexity.

I'm fine keeping them or removing them---I don't feel strongly. That said, since this PR already works, I'd rather get it in and then remove the `weak_ptr`s in a separate PR, rather than letting this one languish and potentially diverge form `develop`.

brryan:
> IMO the weak pointers don't buy us a whole lot, but they're more defensible in the internals of the code. The main place they're used is as back-pointers for objects owned by `MeshBlock`, such as `bvals` and `variable`. I think in user code, those back-pointers are rarely used. The weak pointers buy us (technically) some memory safety as they're non-cyclical but reference counted. They cost us a little user-invisible complexity.
> 
> I'm fine keeping them or removing them---I don't feel strongly. That said, since this PR already works, I'd rather get it in and then remove the `weak_ptr`s in a separate PR, rather than letting this one languish and potentially diverge form `develop`.

Thanks for clarifying. Given your points that the weak pointers are internal and that they do protect against cyclic dependencies, I'm fine to defer thinking about our weak pointer usage to later/never.

src/interface/meshblock_data.hpp
```diff
@@ -97,10 +97,10 @@ class MeshBlockData {
     SetBlockPointer(other.get());
   }
   void SetBlockPointer(const MeshBlockData<T> &other) {
-    pmy_block = other.GetBlockPointer();
+    pmy_block = other.GetBlockSharedPointer();
```
brryan, line 100: Since pointers and shared pointers are coexisting more closely in some cases clarity might be improved by naming shared points `spmy_*` and naming pointers `pmy_*`. I don't think this is blocking but I'm just bringing it up in case you agree.

src/mesh/amr_loadbalance.cpp
```diff
@@ -751,7 +751,7 @@ void Mesh::RedistributeAndRefineMeshBlocks(ParameterInput *pin, ApplicationInput
       auto pmb = FindMeshBlock(on);
       for (auto &var : pmb->vars_cc_) {
         restriction_cache.RegisterRegionHost(
-            irestrict++, ProResInfo::GetInteriorRestrict(pmb, NeighborBlock(), var),
+            irestrict++, ProResInfo::GetInteriorRestrict(pmb.get(), NeighborBlock(), var),
```
brryan, line 754: Would `FindMeshBlock` returning a pointer rather than a shared pointer be more in the spirit of this PR?

brryan (approved):
Two non-blocking comments you're welcome to ignore, just mentioning them in case you think they are worth the effort to make the changes.

Yurlungur:
Now we're post release, I'm clicking the button

PR #980, opened 2023-11-29
WIP: Use non-atomic scatterview in hist

pgrete:
## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build

### DIFF
--- a/src/outputs/histogram.cpp
+++ b/src/outputs/histogram.cpp
@@ -207,8 +207,10 @@ Histogram::Histogram(ParameterInput *pin, const std::string &block_name,
   const auto nybins = ndim_ == 2 ? y_edges_.extent_int(0) - 1 : 1;
 
   result_ = ParArray2D<Real>(prefix + "result", nybins, nxbins);
-  scatter_result =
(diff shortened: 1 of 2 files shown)


BenWibking:
Are you compiling with ` -munsafe-fp-atomics`? It should be safe to use this flag, since (I think) we don't depend on coherent memory while GPU kernels are running (https://www.exascaleproject.org/wp-content/uploads/2022/02/crusher_tips_and_tricks-final.pdf)

pgrete:
Oh, good point(er). I'll test that.

BenWibking:
Did that help?

pgrete:
Indeed it did, so I'm now wondering if the performance issue we saw with Ascent on Frontier are also related to this.

BenWibking:
> Indeed it did, so I'm now wondering if the performance issue we saw with Ascent on Frontier are also related to this.

Ascent histograms are computed on the host, using unified memory support. I can't think of an obvious place it would otherwise use device atomics. Or are you thinking of a different performance issue?

PR #981, opened 2023-12-11
`IndexSplit` for hierarchical parallelism

Yurlungur:
## PR Summary

Here I port in a tool written by @jdolence used for balancing the considerations of GPU vs CPU hardware constraints when using hierarchical parallelism. The `IndexSplit` class lets one automatically fuse different indices in two levels of hierarhical parallelism. For example, put k and j in an outer loop and i in an inner loop, or k in the outer loop and j and i in the inner loop. 

I also add some documentation describing best practices for hierarchical parallelism that we got from our experience in performance tuning downstream codes.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/doc/sphinx/src/nested_par_for.rst
+++ b/doc/sphinx/src/nested_par_for.rst
@@ -121,7 +121,7 @@ get optimal performance. Here we list a few strategies/considerations.
   inner loop per SM (which is related to "occupancy") will depend
   positively on length of the inner loop and negatively on higher
   shared memory usage (scratch pad memory in Kokkos parlance and Local
-  Data Store or LDS on AMD GPUs) and higher register usage. Note that
+  Data Share or LDS on AMD GPUs) and higher register usage. Note that
   the number of SMs and the available shared memory and registers per
   SM will vary between GPU architectures and especially between GPU
   vendors.
--- a/doc/sphinx/src/nested_par_for.rst
+++ b/doc/sphinx/src/nested_par_for.rst
@@ -133,7 +133,7 @@ where here ``md`` is a ``MeshData`` object on which you want to
 operate. ``domain`` specifies where in the ``MeshBlock`` you wish to
 operate, for example ``IndexDomain::Interior``. ``nkp`` and ``njp``
 are the number of points in ``X3`` and ``X2`` respectively that are in
-the outer loop. All remaining points are in the inner loop. Typically
+the outer loop. All remaining points are in the inner loop; each team will iterate over multiple `k` and/or `j` indices to cover the specified `k/j` range. Typically
 ``MeshBlock`` index in the pack is also assumed to be in the outer
 loop. ``nkp`` and ``njp`` also accept special flags
 ``IndexSplit::all_outer`` and ``IndexSplit::no_outer``, which specify
--- a/doc/sphinx/src/nested_par_for.rst
+++ b/doc/sphinx/src/nested_par_for.rst
@@ -129,8 +129,8 @@ can be constructed as
 
   IndexSplit(MeshData<Real> md, IndexDomain domain, const int nkp, const int njp);
 
-where here ``md`` is ``MeshData`` object on which you want to
-operatore. ``domain`` specifies where in the ``MeshBlock`` you wish to
+where here ``md`` is a ``MeshData`` object on which you want to
+operate. ``domain`` specifies where in the ``MeshBlock`` you wish to
 operate, for example ``IndexDomain::Interior``. ``nkp`` and ``njp``
 are the number of points in ``X3`` and ``X2`` respectively that are in
 the outer loop. All remaining points are in the inner loop. Typically
--- a/doc/sphinx/src/nested_par_for.rst
+++ b/doc/sphinx/src/nested_par_for.rst
@@ -110,10 +110,10 @@ get optimal performance. Here we list a few strategies/considerations.
   vector cells in the inner loop is to do a 1D ``SIMDFOR`` inner loop
   but combine the ``j`` and ``i`` indices by simply looping over the
   contiguous memory in a rasterized plane on a block.
-* On Cuda GPUs, the outer loop typically maps to blocks, while the
+* On GPUs, the outer loop typically maps to blocks, while the
   inner maps to threads. To see good performance, you must both
-  provide enough work in the inner loop to create enough threads to fill a streaming multiprocessor (SM) with multiple warps to take advantage of pipelining and enough work
-  in the outer loop to create enough blocks to fill all SMs on the GPU divided by the number of simultaneous streams. The number of warps in flight on the inner loop per SM (which is related to "occupancy") will depend positively on length of the inner loop and negatively on higher shared memory usage (or scratch pad memory in Kokkos parlance) and higher register usage.
+  provide enough work in the inner loop to create enough threads to fill in CUDA terms a streaming multiprocessor (SM, equivalent to a Compute Unit or CU on AMD GPUs) with multiple warps (or wavefronts for AMD) to take advantage of pipelining and enough work
+  in the outer loop to create enough blocks to fill all SMs on the GPU divided by the number of simultaneous streams. The number of warps in flight on the inner loop per SM (which is related to "occupancy") will depend positively on length of the inner loop and negatively on higher shared memory usage (scratch pad memory in Kokkos parlance and Local Data Store or LDS on AMD GPUs) and higher register usage. Note that the number of SMs and the available shared memory and registers per SM will vary between GPU architectures and especially between GPU vendors.
 
 IndexSplit
 -------------
--- a/doc/sphinx/src/nested_par_for.rst
+++ b/doc/sphinx/src/nested_par_for.rst
@@ -136,11 +136,12 @@ available in ``<parthenon/package.hpp>`` in the
 ``parthenon::package::prelude`` namespace.
 
 In our experience ``IndexSplit`` is most beneficial when working with
-small meshblocks, especially in two dimensions. For small blocks, we
-want vectorized operations over contiguous memory for our innermost
-loop, but we want that loop to contain enough work for, e.g., vector
-ops to function. We have often found that the optimal split is to fuse
-j, and i into the inner loop and use k and blocks in the outer loop.
+small meshblocks on CPUs, especially in two dimensions. For small
+blocks, we want vectorized operations over contiguous memory for our
+innermost loop, but we want that loop to contain enough work for,
+e.g., vector ops to function. We have often found that the optimal
+split is to fuse j, and i into the inner loop and use k and blocks in
+the outer loop.
 
 The ``IndexSplit`` class can be constructed as
 
--- a/doc/sphinx/src/nested_par_for.rst
+++ b/doc/sphinx/src/nested_par_for.rst
@@ -51,7 +51,7 @@ Data type for memory in scratch pad/cache memory. Use
 documentation <https://kokkos.github.io/kokkos-core-wiki/ProgrammingGuide/HierarchicalParallelism.html?highlight=hierarchical>`__
 for determining scratch pad memory needs before kernel launch.
 
-Important usage hints
+On Barriers
 ---------------------
 
 In order to ensure that individual threads of a team are synchronized
@@ -70,6 +70,7 @@ write to common variables, see this
 `code <https://github.com/parthenon-hpc-lab/parthenon/issues/659#issuecomment-1346871509>`__
 for an example.
 
+
 Cmake Options
 -------------
 
@@ -86,3 +87,117 @@ GPUs.
 ``#pragma omp simd`` to vectorize the loop, which typically gives better
 vectorization loops than ``PAR_LOOP_INNER_LAYOUT=TVR_INNER_LOOP`` on
 CPUs and so is the default on CPUs.
+
+
+Performance Considerations
+---------------------------
+
+Hierarchical parallelism can produce very performant code, but a
+deeper awareness of how hardware is mapped to threads is required to
+get optimal performance. Here we list a few strategies/considerations.
+
+* On CPU, with `SIMDFOR_INNER_LOOP` you may have trouble vectorizing
+  unless you help the compiler along. One way to do so is to work with
+  raw pointers to contiguous memory, rather than working with views
+  and strides. Even for stencil ops, if you can pull out pointers that
+  represent the different points on the stencil, this can help with
+  vectorization.
+* Similarly on CPUs, due to the cost of starting up a vector op,
+  vectorization will only be a performance win if there's enough work
+  in the inner loop. A minimum of 16 points is required for the op to
+  vectorize at all. Experience shows, however, that at least 64 is
+  really required to see big wins. One strategy for providing enough
+  vector cells in the inner loop is to do a 1D ``SIMDFOR`` inner loop
+  but combine the ``j`` and ``i`` indices by simply looping over the
+  contiguous memory in a rasterized plane on a block.
+* On Cuda GPUs, the outer loop typically maps to warps, while the
+  inner maps to threads. To see good performance, you must both
+  provide enough work in the inner loop to fill a warp and enough work
+  in the outer loop to fill all warps.
+
+IndexSplit
+-------------
+
+To balance the CPU vs GPU hardware considerations of hierarchical
+parallelism, ``Parthenon`` provides a utility, the ``IndexSplit``
+class, defined in the ``utils/index_split.hpp`` header file and
+available in ``<parthenon/package.hpp>`` in the
+``parthenon::package::prelude`` namespace. The ``IndexSplit`` class
+can be constructed as
+
+.. code:: cpp
+
+  IndexSplit(MeshData<Real> md, IndexDomain domain, const int nkp, const int njp);
+
+where here ``md`` is ``MeshData`` object on which you want to
+operatore. ``domain`` specifies where in the ``MeshBlock`` you wish to
+operate, for example ``IndexDomain::Interior``. ``nkp`` and ``njp``
+are the number of points in ``X3`` and ``X2`` respectively that are in
+the outer loop. All remaining points are in the inner loop. Typically
+``MeshBlock`` index in the pack is also assumed to be in the outer
+loop. ``nkp`` and ``njp`` also accept special flags
+``IndexSplit::all_outer`` and ``IndexSplit::no_outer``, which specify
+that all and none of the indices in that direction should be in the
+outer loop.
+
+A second constructor alternatively sets the range for ``X3``, ``X2``,
+and ``X1`` explicitly:
+
+.. code:: cpp
+
+  IndexSplit(MeshData<Real> *md, const IndexRange &kb, const IndexRange &jb,
+             const IndexRange &ib, const int nkp, const int njp);
+
+where here ``kb``, ``jb``, and ``ib`` specify the starting and ending
+indices for ``X3``, ``X2``, and ``X1`` respecively.
+
+An ``IndexSplit`` object is typically used as:
+
+.. code:: cpp
+
+  using namespace parthenon::package::prelude;
+  using parthenon::ScratchPad1D;
+  using parthenon::IndexSplit;
+  using parthenon::par_for_outer;
+  using parthenon::par_for_inner;
+  using parthenon::team_mbr_t;
+  // Initialize index split object
+  IndexSplit idx_sp(md, IndexDomain::interior, nkp, njp);
+  
+  // Request maximum size in i and j in the inner loop, for scratch
+  const int Ni = idx_sp.get_max_ni();
+  const int Nj = idx_sp = get_max_nj();
+  const in tNmax = Ni * Nj;
+  
+  // single scratch array for i,j
+  auto scratch_size = ScratchPad1D<Real>::shmem_size(Nmax);
+  constexpr int scratch_level = 0;
+  
+  // Par for
+  par_for_outer(
+	  DEFAULT_OUTER_LOOP_PATTERN, "KernalOuter", DevExecSpace(), scratch_size,
+	  scratch_level, 0, nblocks - 1, 0, idx_sp.outer_size() - 1,
+	  KOKKOS_LAMBDA(team_mbr_t member, const int b, const int outer_idx) {
+	    ScratchPad1D<Real> scratch(member.team_scratch(scratch_level), Nmax);
+	    // Get index ranges. Note they depend on where we are in the outer index!
+	    // These give us a sense for where we are in k,j space
+	    const auto krange = idx_sp.GetBoundsK(outer_idx);
+	    const auto jrange = idx_sp.GetBoundsJ(outer_idx);
+	    // This is the loop of contiguous inner memory. May contain i and j!
+	    const auto irange = idx_sp.GetInnerBounds(jrange);
+
+	    // Whatever part of k is not in the outer loop can be looped over
+	    // with a normal for loop here
+	    for (int k = krange.s; k <= krange.e; ++k) {
+
+	      // pull out a pointer some variable in some pack. Note
+	      // we pick the 0th index of i at k and jrange.s
+	      Real *var = &pack(b, ivar, k, jrange.s, 0);
+
+	      // Do something with the pointer in the inner loop.
+	      par_for_inner(DEFAULT_INNER_LOOP_PATTERN, member, irange.s, irange.e,
+	        [&](const int i) {
+		  foo(var[i]);
+		});
+	    }
+	  });
--- a/doc/sphinx/src/nested_par_for.rst
+++ b/doc/sphinx/src/nested_par_for.rst
@@ -161,6 +161,12 @@ special flags ``IndexSplit::all_outer`` and ``IndexSplit::no_outer``,
 which specify that all and none of the indices in that direction
 should be in the outer loop.
 
+.. warning::
+
+  Note that, in contrast to ``njp``, ``nkp`` points in the
+  ``k``-direction are not included in the innermost loop bounds. You
+  must loop over ``k`` by hand inside the outer loop body.
+
 A second constructor alternatively sets the range for ``X3``, ``X2``,
 and ``X1`` explicitly:
 
@@ -213,7 +219,8 @@ An ``IndexSplit`` object is typically used as:
 	    const auto krange = idx_sp.GetBoundsK(outer_idx);
 	    const auto jrange = idx_sp.GetBoundsJ(outer_idx);
 	    // This is the loop of contiguous inner memory. May contain i and j!
-	    const auto irange = idx_sp.GetInnerBounds(jrange);
+	    const auto flattened_inner_ijrange = idx_sp.GetInnerBounds(jrange);
+            const int inner_size = flattened_inner_ijrange.e - flattened_inner_ijrange.s + 1;
 
 	    // Whatever part of k is not in the outer loop can be looped over
 	    // with a normal for loop here
@@ -221,10 +228,10 @@ An ``IndexSplit`` object is typically used as:
 
 	      // pull out a pointer some variable in some pack. Note
 	      // we pick the 0th index of i at k and jrange.s
-	      Real *var = &pack(b, ivar, k, jrange.s, 0);
+	      Real *var = &pack(b, ivar, k, jrange.s, flattened_inner_ijrange.s);
 
 	      // Do something with the pointer in the inner loop.
-	      par_for_inner(DEFAULT_INNER_LOOP_PATTERN, member, irange.s, irange.e,
+	      par_for_inner(DEFAULT_INNER_LOOP_PATTERN, member, 0, flattened_inner_size,
 	        [&](const int i) {
 		  foo(var[i]);
 		});
--- a/doc/sphinx/src/nested_par_for.rst
+++ b/doc/sphinx/src/nested_par_for.rst
@@ -153,12 +153,13 @@ where here ``md`` is a ``MeshData`` object on which you want to
 operate. ``domain`` specifies where in the ``MeshBlock`` you wish to
 operate, for example ``IndexDomain::Interior``. ``nkp`` and ``njp``
 are the number of points in ``X3`` and ``X2`` respectively that are in
-the outer loop. All remaining points are in the inner loop; each team will iterate over multiple `k` and/or `j` indices to cover the specified `k/j` range. Typically
-``MeshBlock`` index in the pack is also assumed to be in the outer
-loop. ``nkp`` and ``njp`` also accept special flags
-``IndexSplit::all_outer`` and ``IndexSplit::no_outer``, which specify
-that all and none of the indices in that direction should be in the
-outer loop.
+the outer loop. All remaining points are in the inner loop; each team
+will iterate over multiple `k` and/or `j` indices to cover the
+specified `k/j` range. Typically ``MeshBlock`` index in the pack is
+also assumed to be in the outer loop. ``nkp`` and ``njp`` also accept
+special flags ``IndexSplit::all_outer`` and ``IndexSplit::no_outer``,
+which specify that all and none of the indices in that direction
+should be in the outer loop.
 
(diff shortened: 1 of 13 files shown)


doc/sphinx/src/nested_par_for.rst
```diff
@@ -86,3 +87,117 @@ GPUs.
...
+  in the inner loop. A minimum of 16 points is required for the op to
+  vectorize at all. Experience shows, however, that at least 64 is
+  really required to see big wins. One strategy for providing enough
+  vector cells in the inner loop is to do a 1D ``SIMDFOR`` inner loop
+  but combine the ``j`` and ``i`` indices by simply looping over the
+  contiguous memory in a rasterized plane on a block.
+* On Cuda GPUs, the outer loop typically maps to warps, while the
+  inner maps to threads. To see good performance, you must both
+  provide enough work in the inner loop to fill a warp and enough work
+  in the outer loop to fill all warps.
```
forrestglines:
```suggestion
* On Cuda GPUs, the outer loop typically maps to blocks, while the
  inner maps to threads. To see good performance, you must both
  provide enough work in the inner loop to create enough threads to fill a streaming multiprocessor (SM) with multiple warps to take advantage of pipelining and enough work
  in the outer loop to create enough blocks to fill all SMs on the GPU divided by the number of simultaneous streams. The number of warps in flight on the inner loop per SM (which is related to "occupancy") will depend positively on length of the inner loop and negatively on higher shared memory usage (or scratch pad memory in Kokkos parlance) and higher register usage.
```

I think this is more accurate to the hardware but might be too verbose. Has this been inline with your performance testing? I'm drawing from experience with optimizing GPU code but outside of Parthenon.

doc/sphinx/src/nested_par_for.rst
```diff
@@ -86,3 +87,117 @@ GPUs.
...
+Hierarchical parallelism can produce very performant code, but a
+deeper awareness of how hardware is mapped to threads is required to
+get optimal performance. Here we list a few strategies/considerations.
+
+* On CPU, with `SIMDFOR_INNER_LOOP` you may have trouble vectorizing
+  unless you help the compiler along. One way to do so is to work with
+  raw pointers to contiguous memory, rather than working with views
+  and strides. Even for stencil ops, if you can pull out pointers that
+  represent the different points on the stencil, this can help with
+  vectorization.
```
forrestglines, line 104: This has been echoed by multiple talks at KUG

doc/sphinx/src/nested_par_for.rst
```diff
@@ -86,3 +87,117 @@ GPUs.
...
+available in ``<parthenon/package.hpp>`` in the
+``parthenon::package::prelude`` namespace. The ``IndexSplit`` class
+can be constructed as
+
+.. code:: cpp
+
+  IndexSplit(MeshData<Real> md, IndexDomain domain, const int nkp, const int njp);
+
+where here ``md`` is ``MeshData`` object on which you want to
+operatore. ``domain`` specifies where in the ``MeshBlock`` you wish to
```
forrestglines:
```suggestion
where here ``md`` is a ``MeshData`` object on which you want to
operate. ``domain`` specifies where in the ``MeshBlock`` you wish to
```

doc/sphinx/src/nested_par_for.rst
```diff
@@ -86,3 +87,117 @@ GPUs.
...
+operate, for example ``IndexDomain::Interior``. ``nkp`` and ``njp``
+are the number of points in ``X3`` and ``X2`` respectively that are in
+the outer loop. All remaining points are in the inner loop. Typically
```
forrestglines: So  unless both are `all_outer` or `nkp` equals the full index range, then some fraction of the k iteration or j iteration will iterate within the same team, right?

forrestglines: The parameters for best performance would seem to vary per kernel and GPU hardware, I would think. Can we provide functionality for best guesses that depend on architecture? Like `Kokkos::HintLightWeight`?

Might be beyond the scope of this PR.

doc/sphinx/src/nested_par_for.rst
forrestglines: We can add the equivalent AMD lingo too. The performance considerations are largely the same, except that scratch pad is much smaller and wavefronts are twice as large as warps. (Warps=Wavefronts, Blocks=blocks, SM=CU, shared memory = LocalDataStore(LDS) ).

doc/sphinx/src/nested_par_for.rst
```diff
@@ -86,3 +87,117 @@ GPUs.
...
+  in the outer loop to fill all warps.
+
+IndexSplit
+-------------
+
+To balance the CPU vs GPU hardware considerations of hierarchical
+parallelism, ``Parthenon`` provides a utility, the ``IndexSplit``
+class, defined in the ``utils/index_split.hpp`` header file and
+available in ``<parthenon/package.hpp>`` in the
+``parthenon::package::prelude`` namespace. The ``IndexSplit`` class
```
forrestglines: Can you give a high-level description of how `IndexSplit` behaves for CPUs and GPUs? How it might benefit performance? In what scenario should I use `IndexsSplit`?

src/utils/index_split.cpp
```diff
@@ -0,0 +1,96 @@
...
+#else
+    nkp_ = 1;
+#endif
+  } else if (nkp_ > total_k) {
+    nkp_ = total_k;
+  }
+  if (njp_ == 0) {
+#ifdef KOKKOS_ENABLE_CUDA
+    // TODO(@jdolence) why 4?  Do something better here for a default
+    njp_ = std::min(4, total_j);
```
forrestglines: You should choose a default for `nkp_` and `njp_` such that `nkp_*njp_ >= Num_of_SM/Num_of_streams`. The number of SMs is increasing for each architecture, currently it's 132 for an H100.

I wonder if Kokkos would provide a macro for that number, or we can make a map of those architecture if we really care.

src/utils/index_split.cpp
```diff
@@ -0,0 +1,96 @@
...
+#else
+    njp_ = 1;
+#endif
+  } else if (njp_ > total_j) {
+    njp_ = total_j;
+  }
+
+  // add a tiny bit to avoid round-off issues when we ultimately convert to int
+  target_k_ = (1.0 * total_k) / nkp_ + 1.e-6;
+  target_j_ = (1.0 * total_j) / njp_ + 1.e-6;
```
forrestglines, line 116:
```suggestion
  target_k_ = static_cast<int>((1.0 * total_k) / nkp_ + 1.e-6);
  target_j_ = static_cast<int>((1.0 * total_j) / njp_ + 1.e-6);
```

tst/unit/test_index_split.cpp
```diff
@@ -0,0 +1,157 @@
...
+                      vector_shape);
+    auto pkg = std::make_shared<StateDescriptor>("Test package");
+    pkg->AddField(v1::name(), m);
+    pkg->AddField(v3::name(), m_vector);
+    pkg->AddField(v5::name(), m);
+    BlockList_t block_list = MakeBlockList(pkg, NBLOCKS, N, NDIM);
+
+    MeshData<Real> mesh_data("base");
+    mesh_data.Set(block_list, nullptr, NDIM);
+
```
forrestglines, line 103: There's no test for `1<nkp<N` or `1<njp<N`? For when `nkp` or `njp` don't evenly divide into `N`?

src/utils/index_split.hpp
```diff
@@ -0,0 +1,115 @@
...
+    const auto kf = p / njp_;
+    return {kbs_ + static_cast<int>(kf * target_k_),
+            kbs_ + static_cast<int>((kf + 1) * target_k_) - 1};
+  }
+  KOKKOS_INLINE_FUNCTION
+  IndexRange GetBoundsJ(const int p) const {
+    const auto jf = p % njp_;
+    return {jbs_ + static_cast<int>(jf * target_j_),
+            jbs_ + static_cast<int>((jf + 1) * target_j_) - 1};
+  }
```
forrestglines, line 48: Don't we need a check on the end boundary in case the last team goes over the k/j range? If `kbe_entire%nkp!=0` or `jbe_entire%njp!=0`? In some cases the inner `j` and `k` bounds should be uneven between teams, correct?

Am I missing something that performs that check?

src/interface/mesh_data.hpp
```diff
@@ -419,6 +412,7 @@ class MeshData {
     bvars_cache_.clear();
   }
 
+  int GetNDim() const { return ndim_; }
```
forrestglines, line 415: This change is so that we can create a `MeshData` object without a `Mesh` and just an `ndim` and block list for regression tests, right? Nice!

doc/sphinx/src/nested_par_for.rst
forrestglines:
```suggestion
operate, for example ``IndexDomain::Interior``. ``nkp`` and ``njp``
are the number of points in ``X3`` and ``X2`` respectively that are in
the outer loop. All remaining points are in the inner loop; each team will iterate over multiple `k` and/or `j` indices to cover the specified `k/j` range. Typically
```

tst/unit/test_index_split.cpp
```diff
@@ -0,0 +1,157 @@
...
+    WHEN("We initialize an IndexSplit with all outer k and no outer j") {
+      IndexSplit sp(&mesh_data, IndexDomain::interior, IndexSplit::all_outer,
+                    IndexSplit::no_outer);
+      THEN("The outer range should be appropriate") { REQUIRE(sp.outer_size() == N); }
+      THEN("The inner ranges should be appropriate") {
+        using atomic_view = Kokkos::MemoryTraits<Kokkos::Atomic>;
+        Kokkos::View<int *, atomic_view> nwrong("nwrong", 1);
```
forrestglines, line 110: Do a reduction without actually doing a `kokkos_reduction`. I approve

tst/unit/test_index_split.cpp
```diff
@@ -0,0 +1,157 @@
...
+        parthenon::par_for_outer(
+            DEFAULT_OUTER_LOOP_PATTERN, "Test IndexSplit", DevExecSpace(), 0, 0, 0,
+            sp.outer_size() - 1,
```
forrestglines: This is `N*N -1`, correct
```suggestion
            sp.outer_size() - 1, //N * N - 1
```

tst/unit/test_index_split.cpp
```diff
@@ -0,0 +1,157 @@
...
+      THEN("The inner index ranges should be appropriate") {
+        using atomic_view = Kokkos::MemoryTraits<Kokkos::Atomic>;
+        Kokkos::View<int *, atomic_view> nwrong("nwrong", 1);
+        parthenon::par_for_outer(
+            DEFAULT_OUTER_LOOP_PATTERN, "Test IndexSplit", DevExecSpace(), 0, 0, 0,
+            sp.outer_size() - 1,
+            KOKKOS_LAMBDA(parthenon::team_mbr_t member, const int outer_idx) {
+              const auto krange = sp.GetBoundsK(outer_idx);
+              const auto jrange = sp.GetBoundsJ(outer_idx);
+              const auto irange = sp.GetInnerBounds(jrange);
```
forrestglines, line 146: `krange` and `jrange` could be computed here to check, right?

doc/sphinx/src/nested_par_for.rst
```diff
@@ -86,3 +87,117 @@ GPUs.
...
+	    // Get index ranges. Note they depend on where we are in the outer index!
+	    // These give us a sense for where we are in k,j space
+	    const auto krange = idx_sp.GetBoundsK(outer_idx);
+	    const auto jrange = idx_sp.GetBoundsJ(outer_idx);
+	    // This is the loop of contiguous inner memory. May contain i and j!
+	    const auto irange = idx_sp.GetInnerBounds(jrange);
+
+	    // Whatever part of k is not in the outer loop can be looped over
+	    // with a normal for loop here
+	    for (int k = krange.s; k <= krange.e; ++k) {
```
forrestglines: Since we're performing the same iteration for each thread in the team, this isn't improving occupancy by increasing number of threads per block but increasing work per thread, correct?

forrestglines (changes requested):
Very neat! This is intended to increase work per thread for GPUs and decrease number of blocks -- how has this affected performance? Are we planning on using this in production code?

I could see applying the k/j loop to the inner thread loop increase the amount of work on the finest level for GPUs in order to maximize occupancy. This might help performance for blocks smaller than 64~128.

I've added some text to the documentation that might help make better decisions for choosing number of teams/work per team. I'm drawing from my previous experience/knowledge with CUDA, I don't know if it's been in line with your testing.

I also couldn't figure out how the code might handle the case where `nkp` doesn't divide evenly into the k range so that one team necessarily does less work than the others. It seems the code would just run over the loop bounds as is.

doc/sphinx/src/nested_par_for.rst
Yurlungur: Nice thanks for clarifying this. I added an additional sentence summarizing your comment on AMD GPUs as well.

doc/sphinx/src/nested_par_for.rst
Yurlungur: Nice! It's good to hear others have had the same experience as us.

doc/sphinx/src/nested_par_for.rst
Yurlungur: > Has this been inline with your performance testing?

Yeah you're spot on.

tst/unit/test_index_split.cpp
Yurlungur: yep thanks for adding the clarifying comment

src/interface/mesh_data.hpp
Yurlungur: Yep that's right! :)

tst/unit/test_index_split.cpp
Yurlungur: :grin:

doc/sphinx/src/nested_par_for.rst
Yurlungur: > So unless both are all_outer or nkp equals the full index range, then some fraction of the k iteration or j iteration will iterate within the same team, right?

That's right

> The parameters for best performance would seem to vary per kernel and GPU hardware, I would think. Can we provide functionality for best guesses that depend on architecture? Like Kokkos::HintLightWeight?
> 
> Might be beyond the scope of this PR.

I like this idea, but I do think this is outside of the scope of this one.

doc/sphinx/src/nested_par_for.rst
Yurlungur: Yeah I added a paragraph on the context/benefit of `IndexSplit` and also some roughly guidelines for the optimal split on CPUs vs GPUs. Let me know what you think.

src/utils/index_split.cpp
Yurlungur: Cool thanks for this! I've updated the default to be `njp_ = nkp_ / NSMS` since we don't use streams right now. I also added a comment to update this later once we have access to something that lets us compute this number from, e.g., Kokkos.

doc/sphinx/src/nested_par_for.rst
Yurlungur: Yes---that's right.

doc/sphinx/src/nested_par_for.rst
```diff
@@ -86,3 +87,117 @@ GPUs.
...
+  in the inner loop. A minimum of 16 points is required for the op to
+  vectorize at all. Experience shows, however, that at least 64 is
+  really required to see big wins. One strategy for providing enough
+  vector cells in the inner loop is to do a 1D ``SIMDFOR`` inner loop
+  but combine the ``j`` and ``i`` indices by simply looping over the
+  contiguous memory in a rasterized plane on a block.
+* On Cuda GPUs, the outer loop typically maps to blocks, while the
+  inner maps to threads. To see good performance, you must both
+  provide enough work in the inner loop to create enough threads to fill a streaming multiprocessor (SM) with multiple warps to take advantage of pipelining and enough work
+  in the outer loop to create enough blocks to fill all SMs on the GPU divided by the number of simultaneous streams. The number of warps in flight on the inner loop per SM (which is related to "occupancy") will depend positively on length of the inner loop and negatively on higher shared memory usage (or scratch pad memory in Kokkos parlance) and higher register usage.
```
forrestglines:
```suggestion
* On GPUs, the outer loop typically maps to blocks, while the
  inner maps to threads. To see good performance, you must both
  provide enough work in the inner loop to create enough threads to fill in CUDA terms a streaming multiprocessor (SM, equivalent to a Compute Unit or CU on AMD GPUs) with multiple warps (or wavefronts for AMD) to take advantage of pipelining and enough work
  in the outer loop to create enough blocks to fill all SMs on the GPU divided by the number of simultaneous streams. The number of warps in flight on the inner loop per SM (which is related to "occupancy") will depend positively on length of the inner loop and negatively on higher shared memory usage (scratch pad memory in Kokkos parlance and Local Data Store or LDS on AMD GPUs) and higher register usage. Note that the number of SMs and the available shared memory and registers per SM will vary between GPU architectures and especially between GPU vendors.
```

Added a bit of translation for AMD GPUs. I doubt many will care, but it's there now

Yurlungur: Nice. Thanks!

doc/sphinx/src/nested_par_for.rst
forrestglines: I'm not seeing this new paragraph in the docs. Has it been pushed up? Could you comment on it?

src/utils/index_split.cpp
forrestglines: Sounds good. In terms of naming, is `MAX_CONCURRENT_TEAMS` too verbose? It's more vendor agnostic since AMD uses the terminology Compute Unit (CU). However, I think that might also be a different number since NVIDIA GPUs at least support multiple teams in flight per SM

doc/sphinx/src/nested_par_for.rst
Yurlungur: Sorry I forgot to push yesterday. It's there now.

doc/sphinx/src/nested_par_for.rst
```diff
@@ -86,3 +87,136 @@ GPUs.
...
+class, defined in the ``utils/index_split.hpp`` header file and
+available in ``<parthenon/package.hpp>`` in the
+``parthenon::package::prelude`` namespace.
+
+In our experience ``IndexSplit`` is most beneficial when working with
+small meshblocks, especially in two dimensions. For small blocks, we
+want vectorized operations over contiguous memory for our innermost
+loop, but we want that loop to contain enough work for, e.g., vector
+ops to function. We have often found that the optimal split is to fuse
+j, and i into the inner loop and use k and blocks in the outer loop.
```
Yurlungur: @forrestglines here's the new paragraph

forrestglines: I see! So does this give better performance on GPUs and on CPUs? The mention of vectorization suggests that this is a CPU improvement. In my mental model of GPUs launching fewer blocks vs. many blocks here could only have a detrimental performance if `k` is less than the number of SMs. How did this affect GPU performance in your experience?

forrestglines: Could this functionality be inserted into `par_for` so that all local operations use this more optimal arrangement of teams?

Yurlungur: > I see! So does this give better performance on GPUs and on CPUs? The mention of vectorization suggests that this is a CPU improvement. 

Yeah this is a CPU optimization. I can add a comment saying so. I'm not sure what the optimal choice is for GPUs yet. We haven't explored it too much.

> Could this functionality be inserted into par_for so that all local operations use this more optimal arrangement of teams?

Possibly... I'd say we should put that in in a separate PR though. :)

src/utils/index_split.cpp
Yurlungur: I am calling it `concurrency_` I think, and it's now set at run-time based on the suggestions from the Kokkos slack. Take a look and let me know what you think.

Yurlungur:
Thanks for all your suggestions!

> how has this affected performance? Are we planning on using this in production code?

We are using this in production in riot, where it has definitely helped. In riot we basically always fuse at least the i and j indices in the inner loop.

> I also couldn't figure out how the code might handle the case where nkp doesn't divide evenly into the k range so that one team necessarily does less work than the others. It seems the code would just run over the loop bounds as is.

Yeah for k at least, the model is to manually loop over the leftover k indices. Like so:

```C++
  par_for_outer(
	  DEFAULT_OUTER_LOOP_PATTERN, "KernalOuter", DevExecSpace(), scratch_size,
	  scratch_level, 0, nblocks - 1, 0, idx_sp.outer_size() - 1,
	  KOKKOS_LAMBDA(team_mbr_t member, const int b, const int outer_idx) {
	    ScratchPad1D<Real> scratch(member.team_scratch(scratch_level), Nmax);
	    const auto krange = idx_sp.GetBoundsK(outer_idx);
	    const auto jrange = idx_sp.GetBoundsJ(outer_idx);
	    const auto irange = idx_sp.GetInnerBounds(jrange);

	    // Whatever part of k is not in the outer loop can be looped over
	    // with a normal for loop here
	    for (int k = krange.s; k <= krange.e; ++k) {
	      // Do something with the pointer in the inner loop.
	      par_for_inner(DEFAULT_INNER_LOOP_PATTERN, member, irange.s, irange.e,
	        [&](const int i) {
		  foo(var[i]);
		});
	    }
	  });

```
so the "spillover" gets kind of distributed among all the teams. Does this answer your concern satisfactorily? (I assume that's also the remaining unresolved issues in the discussion?)

forrestglines:
> Yeah for k at least, the model is to manually loop over the leftover k indices. Like so:
> 
> ```c++
> 	    const auto krange = idx_sp.GetBoundsK(outer_idx);
> 	    const auto jrange = idx_sp.GetBoundsJ(outer_idx);
> 	    const auto irange = idx_sp.GetInnerBounds(jrange);
> 
> 	    // Whatever part of k is not in the outer loop can be looped over
> 	    // with a normal for loop here
> 	    for (int k = krange.s; k <= krange.e; ++k) {
> ```
> 
> so the "spillover" gets kind of distributed among all the teams. Does this answer your concern satisfactorily? (I assume that's also the remaining unresolved issues in the discussion?)

So `krange.e - krange.s` will not be equal across different `outer_idx` this hypothetical case? For some `outer_idx` that difference will be smaller or greater?

If yes, then I'm satisfied.

doc/sphinx/src/nested_par_for.rst
```diff
@@ -86,3 +87,146 @@ GPUs.
...
+  enough work in the inner loop to create enough threads to fill in
+  CUDA terms a streaming multiprocessor (SM, equivalent to a Compute
+  Unit or CU on AMD GPUs) with multiple warps (or wavefronts for AMD)
+  to take advantage of pipelining and enough work in the outer loop to
+  create enough blocks to fill all SMs on the GPU divided by the
+  number of simultaneous streams. The number of warps in flight on the
+  inner loop per SM (which is related to "occupancy") will depend
+  positively on length of the inner loop and negatively on higher
+  shared memory usage (scratch pad memory in Kokkos parlance and Local
+  Data Store or LDS on AMD GPUs) and higher register usage. Note that
```
forrestglines:
```suggestion
  Data Share or LDS on AMD GPUs) and higher register usage. Note that
```
My typo

Yurlungur:
@forrestglines thought more about your big-picture question about excess work and how that functionally works out. Also discussed briefly with @jdolence .

Big picture is that mismatched sizes do result in one thread block (for example) doing extra work. There's no special fix for that. The right amount of work gets done, but it might not be distributed perfectly evenly. Obviously that's not ideal, but the `IndexSplit` machinery does let the user choose the sizing for a choice optimal for that loop (and it depends on the loop). My suggestion is that we document this shortcoming and keep in our minds some improved automation for choosing the right `nkp` and `njp` down the line. Would this work for you?

forrestglines:
> Big picture is that mismatched sizes do result in one thread block (for example) doing extra work. 

I agree! But I think I might have miscommunicated my concern. I don't see where in your code that the range of `k` changes for this last (or for some set) of teams. I think this extra/less work is unaccounted for. _I'm suggesting that `IndexSplit` has a bug that will cause segfaults by running over intended bounds of `k`, not a performance concern_. 

I don't see this pattern tested in your unit test so I don't know if it works. I think this should be a  test case in your PR before it is merged. It might already be working in your downstream code but I think there should be a test.

I've been busy at the KUG and fixing cylindrical coordinates in AthenaPK (which I think I've finally found the last bug). If you want, I can write the test that I want to see when I'm back at home later today.

Other than that, I'm very happy with the updated documentation and updated choice of parallelization.

src/utils/index_split.hpp
```diff
@@ -0,0 +1,118 @@
...
+  IndexSplit(MeshData<Real> *md, const IndexRange &kb, const IndexRange &jb,
+             const IndexRange &ib, const int nkp, const int njp);
+  IndexSplit(MeshData<Real> *md, IndexDomain domain, const int nkp, const int njp);
+
+  int outer_size() const { return nkp_ * njp_; }
+  KOKKOS_INLINE_FUNCTION
+  IndexRange GetBoundsK(const int p) const {
+    const auto kf = p / njp_;
+    return {kbs_ + static_cast<int>(kf * target_k_),
+            kbs_ + static_cast<int>((kf + 1) * target_k_) - 1};
```
forrestglines, line 41: To be extra clear, I think this line specifically needs to 
```suggestion
            min(kbs_ + static_cast<int>((kf + 1) * target_k_) - 1,kbe_entire_)};
```

src/utils/index_split.hpp
```diff
@@ -0,0 +1,118 @@
...
+  }
+  KOKKOS_INLINE_FUNCTION
+  IndexRange GetBoundsJ(const int p) const {
+    const auto jf = p % njp_;
+    return {jbs_ + static_cast<int>(jf * target_j_),
+            jbs_ + static_cast<int>((jf + 1) * target_j_) - 1};
```
forrestglines, line 47:
```suggestion
            min(jbs_ + static_cast<int>((jf + 1) * target_j_) - 1,jbe_entire_)};
```

tst/unit/test_index_split.cpp
Yurlungur: Added.

tst/unit/test_index_split.cpp
Yurlungur: Done.

src/utils/index_split.cpp
Yurlungur: @forrestglines Actually these cannot be `static_cast` to int as they're doubles. The casting happens later, and the fractional value of these is needed to get the distribution of work to happen properly.

Yurlungur (commented):
> I don't see where in your code that the range of k changes for this last (or for some set) of teams. I think this extra/less work is unaccounted for. I'm suggesting that IndexSplit has a bug that will cause segfaults by running over intended bounds of k, not a performance concern.

Ah I see... Actually we accidentally eliminated this mechanism in review. The static cast (I highlight it below) is incorrect. That value should be a double. The static casting later when the ranges are computed distributes the extra work. (To be honest, I didn't fully understand this. I copied this code from @jdolence 's implementation, and I missed some details.)

> I don't see this pattern tested in your unit test so I don't know if it works. I think this should be a test case in your PR before it is merged. It might already be working in your downstream code but I think there should be a test.

That's fair. I added tests---let me know what you think.

doc/sphinx/src/nested_par_for.rst
```diff
@@ -86,3 +87,146 @@ GPUs.
...
+	    // This is the loop of contiguous inner memory. May contain i and j!
+	    const auto irange = idx_sp.GetInnerBounds(jrange);
+
+	    // Whatever part of k is not in the outer loop can be looped over
+	    // with a normal for loop here
+	    for (int k = krange.s; k <= krange.e; ++k) {
+
+	      // pull out a pointer some variable in some pack. Note
+	      // we pick the 0th index of i at k and jrange.s
+	      Real *var = &pack(b, ivar, k, jrange.s, 0);
```
lroberts36:
```suggestion
	      Real *var = &pack(b, ivar, k, jrange.s, irange.s);
```

doc/sphinx/src/nested_par_for.rst
```diff
@@ -86,3 +87,146 @@ GPUs.
...
+operate, for example ``IndexDomain::Interior``. ``nkp`` and ``njp``
+are the number of points in ``X3`` and ``X2`` respectively that are in
+the outer loop. All remaining points are in the inner loop; each team
+will iterate over multiple `k` and/or `j` indices to cover the
+specified `k/j` range. Typically ``MeshBlock`` index in the pack is
+also assumed to be in the outer loop. ``nkp`` and ``njp`` also accept
+special flags ``IndexSplit::all_outer`` and ``IndexSplit::no_outer``,
+which specify that all and none of the indices in that direction
+should be in the outer loop.
+
```
lroberts36, line 163: Maybe comment somewhere in here that the nkp points in the k-direction are not included in the innermost loop bounds (in contrast to the njp points).

lroberts36: Since this is the case, I assume I can choose something like `nkp = 8` and `njp = 1` and everything will still work correctly?

doc/sphinx/src/nested_par_for.rst
```diff
@@ -86,3 +87,146 @@ GPUs.
...
+	  DEFAULT_OUTER_LOOP_PATTERN, "KernalOuter", DevExecSpace(), scratch_size,
+	  scratch_level, 0, nblocks - 1, 0, idx_sp.outer_size() - 1,
+	  KOKKOS_LAMBDA(team_mbr_t member, const int b, const int outer_idx) {
+	    ScratchPad1D<Real> scratch(member.team_scratch(scratch_level), Nmax);
+	    // Get index ranges. Note they depend on where we are in the outer index!
+	    // These give us a sense for where we are in k,j space
+	    const auto krange = idx_sp.GetBoundsK(outer_idx);
+	    const auto jrange = idx_sp.GetBoundsJ(outer_idx);
+	    // This is the loop of contiguous inner memory. May contain i and j!
+	    const auto irange = idx_sp.GetInnerBounds(jrange);
```
lroberts36: Maybe call this `flattened_inner_ijrange` instead of `irange` for maximum clarity?

lroberts36 (approved):
This LGTM. I didn't carefully check the indexing logic, but I trust that the tests and Riot have caught any issues.

doc/sphinx/src/nested_par_for.rst
Yurlungur: > Maybe comment somewhere in here that the nkp points in the k-direction are not included in the innermost loop bounds (in contrast to the njp points).

Done

> Since this is the case, I assume I can choose something like nkp = 8 and njp = 1 and everything will still work correctly?

Yep.

doc/sphinx/src/nested_par_for.rst
Yurlungur: done

forrestglines (approved):
I reviewed the new changes, looks good to me now! I very much appreciate the new tests.

src/utils/index_split.cpp
forrestglines: Then this was my original mistake, I misread `target_k_` and `target_j_` to be integers in the header file. My first intuition is to implement these things as integers.

Should we add a `static_cast<int>` everywhere else that we're implicitly changing to int? EDIT: We're already doing that! I should have noticed that

We'd expect a minor performance difference using doubles vs. ints, right? Not worth exploring now.

Yurlungur: :+1: yeah there might be a minor performance difference... using ints instead of doubles would require a rewrite though, as those doubles are what let us interweave work.

Yurlungur:
Thanks @forrestglines @lroberts36 for your detailed reviews. Going to set to auto merge now.

PR #982, opened 2023-12-12
Add some diagnostics for Parthenon-VIBE

Yurlungur:
## PR Summary

As a way to quickly gut check some measure of correctness, and to highlight the history features of parthenon, I add some tooling to output the total "mean-square mass" of the Burger's variables in Parthenon-VIBE into the history file. I also output the total number of meshblocks at the time of history dump.

This can be compared, up to some tolerance, for various configurations or pieces of hardware by diffing using the `burgers_diff.py` script that I also add.

In principle, this could be hooked into a gold file for regression testing, but I think for now this is good enough. It provides some useful machinery for us to use for, e.g., machine acceptance criteria.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/benchmarks/burgers/burgers_diff.py
+++ b/benchmarks/burgers/burgers_diff.py
@@ -25,16 +25,6 @@ parser.add_argument(
     "-t", "--tolerance", type=float, default=1e-8, help="Relative tolerance for diff"
 )
 
-
-def matprint(mat, fmt="14g"):
-    "Print a numpy arraay prettily"
-    col_maxes = [max([len(("{:" + fmt + "}").format(x)) for x in col]) for col in mat.T]
-    for x in mat:
-        for i, y in enumerate(x):
-            print(("{:" + str(col_maxes[i]) + fmt + "}").format(y), end="  ")
-        print("")
-
-
 def get_rel_diff(d1, d2):
     "Get relative difference between two numpy arrays"
     return 2 * np.abs(d1 - d2) / (d1 + d2 + 1e-20)
--- a/benchmarks/burgers/burgers_diff.py
+++ b/benchmarks/burgers/burgers_diff.py
@@ -1,3 +1,16 @@
+#========================================================================================
+# (C) (or copyright) 2023. Triad National Security, LLC. All rights reserved.
+#
+# This program was produced under U.S. Government contract 89233218CNA000001 for Los
+# Alamos National Laboratory (LANL), which is operated by Triad National Security, LLC
+# for the U.S. Department of Energy/National Nuclear Security Administration. All rights
+# in the program are reserved by Triad National Security, LLC, and the U.S. Department
+# of Energy/National Nuclear Security Administration. The Government is granted for
+# itself and others acting on its behalf a nonexclusive, paid-up, irrevocable worldwide
+# license in this material to reproduce, prepare derivative works, distribute copies to
+# the public, perform publicly and display publicly, and to permit others to do so.
+#========================================================================================
+
 #!/usr/bin/env python
 import numpy as np
 from argparse import ArgumentParser
--- a/benchmarks/burgers/burgers_diff.py
+++ b/benchmarks/burgers/burgers_diff.py
@@ -25,21 +25,32 @@ parser.add_argument(
     "-t", "--tolerance", type=float, default=1e-8, help="Relative tolerance for diff"
 )
 
+
 def get_rel_diff(d1, d2):
     "Get relative difference between two numpy arrays"
     return 2 * np.abs(d1 - d2) / (d1 + d2 + 1e-20)
 
 
+def compare_files(file1, file2, tolerance, print_results=True):
+    "Compare file1 and file2 to tolerance. Optionally print results."
+    d1 = np.loadtxt(file1)
+    d2 = np.loadtxt(file2)
+    diffs = get_rel_diff(d1, d2)
+    mask = diffs > tolerance
+    errcode = 0
+    if np.any(mask):
+        errcode = 1
+        if print_results:
+            print("Diffs found!")
+            indices = np.transpose(np.nonzero(mask))
+            print("Diff locations (row, column) =", indices)
+            print("Diffs =", diffs[mask])
+    else:
+        if print_results:
+            print("No diffs found!")
+    return errcode
+
+
 if __name__ == "__main__":
     args = parser.parse_args()
-    d1 = np.loadtxt(args.file1)
-    d2 = np.loadtxt(args.file2)
-    diffs = get_rel_diff(d1, d2)
-    mask = diffs > args.tolerance
-    if np.any(mask):
-        print("Diffs found!")
-        indices = np.transpose(np.nonzero(mask))
-        print("Diff locations (row, column) =", indices)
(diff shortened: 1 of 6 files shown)


jdolence (approved):
I think this looks good.  Thanks!

benchmarks/burgers/burgers_diff.py
```diff
@@ -0,0 +1,45 @@
...
+)
+parser.add_argument("file1", type=str, help="First file in diff")
+parser.add_argument("file2", type=str, help="Second fiel in diff")
+parser.add_argument(
+    "-t", "--tolerance", type=float, default=1e-8, help="Relative tolerance for diff"
+)
+
+def get_rel_diff(d1, d2):
+    "Get relative difference between two numpy arrays"
+    return 2 * np.abs(d1 - d2) / (d1 + d2 + 1e-20)
```
pgrete, line 31: How about using some of the [numpy testing](https://numpy.org/doc/stable/reference/routines.testing.html) functions instead (which may be helpful wrt to floating point comparisons at different scales)?

benchmarks/burgers/burgers_diff.py
```diff
@@ -0,0 +1,45 @@
...
+    d2 = np.loadtxt(args.file2)
+    diffs = get_rel_diff(d1, d2)
+    mask = diffs > args.tolerance
+    if np.any(mask):
+        print("Diffs found!")
+        indices = np.transpose(np.nonzero(mask))
+        print("Diff locations (row, column) =", indices)
+        print("Diffs =", diffs[mask])
+    else:
+        print("No diffs found!")
```
pgrete: Could also add a return code here in addition to printing the results (which would allow a straightforward reuse in other automated scripts).

benchmarks/burgers/burgers_package.cpp
```diff
@@ -364,4 +403,34 @@ TaskStatus CalculateFluxes(MeshData<Real> *md) {
...
+      pack.GetDim(5) - 1, 0, pack.GetDim(4) - 1, kb.s, kb.e, jb.s, jb.e, ib.s, ib.e,
+      KOKKOS_LAMBDA(const int b, const int v, const int k, const int j, const int i,
+                    Real &lresult) {
+        const auto &coords = pack.GetCoords(b);
+        const Real vol = coords.CellVolume(k, j, i);
+        const Real x1 = coords.Xc<X1DIR>(k, j, i);
+        const Real x2 = coords.Xc<X2DIR>(k, j, i);
+        const Real x3 = coords.Xc<X3DIR>(k, j, i);
+        const Real mask = (x1min <= x1) && (x1 <= x1max) && (x2min <= x2) &&
+                          (x2 <= x2max) && (x3min <= x3) && (x3 <= x3max);
```
pgrete, line 437: How about adding a comment that's it's fine to use `<=` on both ends (rather than one `<` and one `<=`) because the cell center coordinates are by construction not on the passed in  edges?

benchmarks/burgers/burgers_package.cpp
```diff
@@ -364,4 +403,34 @@ TaskStatus CalculateFluxes(MeshData<Real> *md) {
...
+        lresult += mask * pack(b, v, k, j, i) * pack(b, v, k, j, i) * vol;
+      },
+      Kokkos::Sum<Real>(result));
+  return result;
```
pgrete, line 441: As volume weighting is used, in principle, we should also divide by the total volume (either here or above where we already access x1/2/3min/max)

benchmarks/burgers/burgers_package.hpp
```diff
@@ -40,6 +43,10 @@ void lr_to_flux(const Real uxl, const Real uxr, const Real uyl, const Real uyr,
   fuz = 0.5 * (sr * uzl * upl - sl * uzr * upr + sl * sr * (uzr - uzl)) * islsr;
 }
 
+struct Region {
+  std::array<Real, 3> xmin, xmax;
+};
+
```
pgrete, line 52: In principle, the `RegionSize` struct could be reused.

pgrete (approved):
LGTM (just some minor comments)

benchmarks/burgers/burgers_diff.py
Yurlungur: This looks cool---I wasn't aware of these. But they all seem to be asserts, whereas I would prefer to print an error message and the location in the array that is off. Is there functionality for that?

gshipman (approved):
@Yurlungur , looks good. Thx!

benchmarks/burgers/burgers_diff.py
Yurlungur: Good idea :+1:

benchmarks/burgers/burgers_package.cpp
Yurlungur: Added. :+1:

benchmarks/burgers/burgers_package.cpp
Yurlungur: Good catch. I now store mesh volume as a parameter and use it to compute the weight.

benchmarks/burgers/burgers_package.hpp
Yurlungur: I know... And I initially planned to do that, but `RegionSize` isn't a `struct` anymore, it's a class with accessors and private scope. (I think this changed in the multigrid PR). It was just more cumbersome than doing my own thing. I'll add a comment explaining.

benchmarks/burgers/burgers_diff.py
pgrete: There are also the non-testing/assert versions of most of the functionality, e.g., https://numpy.org/doc/stable/reference/generated/numpy.isclose.html

benchmarks/burgers/burgers_diff.py
```diff
@@ -0,0 +1,56 @@
...
+            print("Diffs =", diffs[mask])
+    else:
+        if print_results:
+            print("No diffs found!")
+    return errcode
+
+
+if __name__ == "__main__":
+    args = parser.parse_args()
+    compare_files(args.file1, args.file1, args.tolerance, True)
```
pgrete, line 56: This is missing a `sys.exit(compare_files(...))` isn't it?

benchmarks/burgers/burgers_diff.py
Yurlungur: Hmm this still doesn't quite meet needs I think, as it doesn't easily output this diff.

benchmarks/burgers/burgers_diff.py
Yurlungur: oops good catch. I'll submit a quick PR.

gshipman:
@Yurlungur  this looks great. Can you add example usage and update the submodule in the LANL benchmarks site? github.com/lanl/benchmarks

Yurlungur:
> @Yurlungur this looks great. Can you add example usage and update the submodule in the LANL benchmarks site? github.com/lanl/benchmarks

Yes---I'll do that. Thanks for the reminder, @gshipman

PR #983, opened 2023-12-12
Add Contains to Sparse Pack

Yurlungur:
## PR Summary

Syntactic sugar for

```C++
bool var_v_is_in_pack = (pack.GetUpperBoundHost(0, v()) >= 0);
```

This was suggested by @brryan in a downstream PR in Phoebus. I think it helps clarify the cases when we pack on many variables, of which only some are present.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/src/interface/sparse_pack.hpp
+++ b/src/interface/sparse_pack.hpp
@@ -249,13 +249,11 @@ class SparsePack : public SparsePackBase {
 
   template <class... Args>
   KOKKOS_INLINE_FUNCTION bool Contains(const int b, Args... args) const {
-    static_assert(sizeof...(Args) > 0, "Must pass in a variable to check existence");
     return GetUpperBound(b, std::forward<Args>(args)...) >= 0;
   }
 
   template <class... Args>
   KOKKOS_INLINE_FUNCTION bool ContainsHost(const int b, Args... args) const {
-    static_assert(sizeof...(Args) > 0, "Must pass in a variable to check existence");
     return GetUpperBoundHost(b, std::forward<Args>(args)...) >= 0;
   }
 
--- a/src/interface/sparse_pack.hpp
+++ b/src/interface/sparse_pack.hpp
@@ -247,6 +247,18 @@ class SparsePack : public SparsePackBase {
     return bounds_h_(1, b, vidx);
   }
 
+  template <class... Args>
+  KOKKOS_INLINE_FUNCTION bool Exists(const int b, Args... args) const {
+    static_assert(sizeof...(Args) > 0, "Must pass in a variable to check existence");
+    return GetUpperBound(b, std::forward<Args>(args)...) >= 0;
+  }
+
+  template <class... Args>
+  KOKKOS_INLINE_FUNCTION bool ExistsHost(const int b, Args... args) const {
+    static_assert(sizeof...(Args) > 0, "Must pass in a variable to check existence");
+    return GetUpperBoundHost(b, std::forward<Args>(args)...) >= 0;
+  }
+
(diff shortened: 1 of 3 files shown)


src/interface/sparse_pack.hpp
```diff
@@ -247,6 +247,18 @@ class SparsePack : public SparsePackBase {
     return bounds_h_(1, b, vidx);
   }
 
+  template <class... Args>
+  KOKKOS_INLINE_FUNCTION bool Exists(const int b, Args... args) const {
```
brryan: Does this need to be a variadic template if we're (presumably) only checking for the existence of one var at a time? Will
```c++
template <typename T>
KOKKOS_INLINE_FUNCTION bool Exists(const int b, T var) const {
  // etc.
}
```
not work?

brryan (approved):
One question but LGTM and I'll approve now.

src/interface/sparse_pack.hpp
Yurlungur: You're totally right, that would work. I somehow thought the topological element was an argument to `GetUpperBound` but it's not. I can make it a variatic pack still so you can check if the pack contains *anything* (i.e., if you pass in zero arguments) or I can just make it a template on one var. What do you think?

src/interface/sparse_pack.hpp
```diff
@@ -247,6 +247,18 @@ class SparsePack : public SparsePackBase {
...
+    static_assert(sizeof...(Args) > 0, "Must pass in a variable to check existence");
+    return GetUpperBound(b, std::forward<Args>(args)...) >= 0;
+  }
+
+  template <class... Args>
+  KOKKOS_INLINE_FUNCTION bool ContainsHost(const int b, Args... args) const {
+    static_assert(sizeof...(Args) > 0, "Must pass in a variable to check existence");
+    return GetUpperBoundHost(b, std::forward<Args>(args)...) >= 0;
+  }
+
```
lroberts36: Maybe add overloads so one can write 
```c++
pack.Contains<v1, v2>(b);
```
That seems cleaner to me.

lroberts36 (commented):
This is a nice feature

src/interface/sparse_pack.hpp
Yurlungur: Are you imagining that returning true only if v1 and v2 are contained?

lroberts36: yes, isn't that how it works currently?

lroberts36: oh, nevermind. I was thinking the return was equivalent to
```c++
return (GetUpperBoundHost(b, std::forward<Args>(args)) >= 0 && ...);
```

Yurlungur: I can make it work that way :grin:

src/interface/sparse_pack.hpp
Yurlungur: Following @lroberts36 suggestion, this now can return the existence status of multiple vars at once via the pack. Either with `Contains<v1, v3, v5>(b)` or with `Contains(b, v1(), v3(), v5())`

lroberts36 (approved):
LGTM

src/interface/sparse_pack.hpp
Yurlungur: It works that way now. This now can return the existence status of multiple vars at once via the pack. Either with `Contains<v1, v3, v5>(b)` or with `Contains(b, v1(), v3(), v5())`

PR #986, opened 2023-12-14
Bugfix for sparse communication

lroberts36:
## PR Summary
If a variable was deallocated at the end of one step and reallocated during the next call to `ReceiveBoundBufs`, it would appear to `SetBounds` that the `BndInfo` cache was not stale (since the field was allocated on subsequent calls to `SetBounds`) even though the variable itself contained newly allocated memory. This is fixed by checking the value of `Variable::GetAllocationStatus()` in `CheckSendBufferCacheForRebuild` and`CheckReceiveBufferCacheForRebuild` against the value that is now store in `BndInfo`. `GetAllocationStatus()` returns zero if the variable is deallocated and the number of times the variable has been allocated over the simulation otherwise. 

This should not have any impact on codes that do not use sparse variables.
## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ### Changed (changing behavior/API/variables/...)
 
 ### Fixed (not changing behavior/API/variables/...)
+- [[PR986]](https://github.com/parthenon-hpc-lab/parthenon/pull/986) Fix bug in sparse boundary communication BndInfo cacheing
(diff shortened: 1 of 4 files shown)


jdolence (approved):
LGTM.  Thanks for the fix!

pdmullen (approved):
LGTM

Issue #959, opened 2023-10-18
Is `GetOrAdd` safe for varying partition sizes within a single driver stage?


pgrete:
Changing the partition sizes within a driver is generally allowed (currently), but it's unclear of that change is properly reflected in the caching label mechanism.
We/I should double check of there are potential issue as there are/might be places in the code where we implicitly assume a single pack (e.g.. for guaranteed non-blocking reductions across all blocks of a rank).

Yurlungur:
This may also be different between sparse and legacy packs (another reason to move to a single codepath at some point)

pgrete:
And for the GMG `GetOrAdd`, see https://github.com/parthenon-hpc-lab/parthenon/pull/911#discussion_r1353088307

PR #987, opened 2023-12-14
[Breaking Change] Tasking rewrite

jdolence:
## PR Summary

[Breaking Change]
The API for representing iterations through our tasks has changed.  `AddIteration` is replaced with `AddSublist` and the arguments and return values are a little different.  Additionally, `AddCompletionTask` is replaced with `AddTask` with `TaskQualifier`s to specify that it is a completion task (and whatever other qualifiers are appropriate).  `AddRegionalDependencies` is similarly replaced by providing the appropriate `TaskQualifier` to `AddTask`.

It's high time we cleaned up the tasking code.  I take a crack at that in this PR.  For the most part, this should be compatible with downstream codes, but it will be breaking for anyone using `IterativeTasks`.  Some basics:
- Replaces all the while loops over lists with a more proper graph traversal
- Gets rid of all the bitmask stuff, totally reimagining what `TaskID` is
- Completely removes `IterativeTasks` in favor of `TaskList`s owning `TaskList`s

Fun new features:
- Graph traversal/task execution is now threaded.  You pass in a thread pool object to the `TaskCollection` `Execute` function and tasks are launched in parallel across the threads in the pool.  If you pass no pool to `Execute`, it just constructs a pool with one thread, so should execute more or less how it did before.
- Executing a `TaskCollection/TaskRegion/TaskList` is no longer self-destructive.  This allows for a model where a task collection is built once and executed many times.
- The functionality in what was `IterativeTasks` is now more flexibile and the interface is cleaner.
  - You specify a min and max number of iterations when adding the "sublist" via the `TaskList::AddSublist` function.  `{1,1}` means it's just a sublist that won't iterate.
  - You can have arbitrarily nested iterations
  - You can specify more than one completion task, to allow for early aborts from a list
- New `TaskQualifier` specifier.  You can combine the qualifiers below with the `|` operator, just like for dependencies.  The interesting ones are
  - `local_sync`: this replaces the old regional dependencies stuff.  It says that dependent tasks can't execute until all the lists in your rank's region have executed this task
  - `global_sync`: this says that all tasks across all ranks and lists must have executed before dependent tasks can execute
  - `completion`: allows for tasks that conditionally continue execution of the list/cycle back to the beginning (`TaskStatus::iterate`) or abort execution of the rest of the list/stop iterating (`TaskStatus::complete`).  The min/max number of iterations for a list are always honored regardless of the status returned by completion tasks.
  - `once_per_region`: for a region with `N` partitions, a tasked marked with this qualifier will only execute a single time instead of `N` times as usual.

I don't add any specific tests because all of the functionality is well tested in our current regression tests.  The only exception is multithreaded task execution.  For now, this feature shouldn't be used anyway because Parthenon has not been made thread-safe everywhere it needs to be.  There is a note to this effect in the docs.  Also, just calling `Execute()` on a `TaskCollection` (as everyone is presumably doing downstream) leads to single-threaded execution.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


src/CMakeLists.txt
```diff
@@ -213,10 +213,11 @@ add_library(parthenon
...
 
-  tasks/task_id.cpp
-  tasks/task_id.hpp
-  tasks/task_list.hpp
-  tasks/task_types.hpp
+  tasks/tasks.hpp
+  #tasks/task_id.cpp
+  #tasks/task_id.hpp
+  #tasks/task_list.hpp
+  #tasks/task_types.hpp
```
Yurlungur: just delete these ;)

src/basic_types.hpp
```diff
@@ -45,7 +45,7 @@ using Real = double;
 // X3DIR z, phi, etc...
 enum CoordinateDirection { NODIR = -1, X0DIR = 0, X1DIR = 1, X2DIR = 2, X3DIR = 3 };
 enum class BlockLocation { Left = 0, Center = 1, Right = 2 };
-enum class TaskStatus { fail, complete, incomplete, iterate, skip, waiting };
+enum class TaskStatus { complete, incomplete, iterate };
```
Yurlungur, line 48: :+1:

src/defs.hpp
```diff
@@ -107,7 +107,7 @@ struct RegionSize {
 // TODO(felker): C++ Core Guidelines Enum.5: Don’t use ALL_CAPS for enumerators
 // (avoid clashes with preprocessor macros). Enumerated type definitions in this file and:
 // io_wrapper.hpp, bvals.hpp, field_diffusion.hpp,
-// task_list.hpp, ???
+// tasks.hpp, ???
```
Yurlungur, line 110: Is this comment still relevant?

src/tasks/thread_pool.hpp
```diff
@@ -0,0 +1,135 @@
...
+#include <condition_variable>
+#include <functional>
+#include <future>
+#include <mutex>
+#include <queue>
+#include <thread>
+#include <vector>
+
+template <typename T>
+class Queue {
```
Yurlungur: This should probably guarded inside a `threads` namespace or something. Or renamed to `ThreadQueue`. I feel like `Queue` is too generic a name for what this thing is.

src/tasks/thread_pool.hpp
```diff
@@ -0,0 +1,135 @@
...
+      q = queue.front();
+      queue.pop();
+    }
+    return false;
+  }
+  void signal_kill() {
+    std::lock_guard<std::mutex> lock(mutex);
+    std::queue<T>().swap(queue);
+    complete = true;
+    exit = true;
```
Yurlungur, line 58: I'm kind of shocked `exit` is not a protected name.

src/tasks/thread_pool.hpp
```diff
@@ -0,0 +1,135 @@
...
+
+class ThreadPool {
+ public:
+  explicit ThreadPool(const int numthreads = std::thread::hardware_concurrency())
+      : nthreads(numthreads), queue(nthreads) {
+    for (int i = 0; i < nthreads; i++) {
+      auto worker = [&]() {
+        while (true) {
+          std::function<void()> f;
+          auto stop = queue.pop(f);
```
Yurlungur, line 100: Why does `queue.pop` assign a `void*` function by reference, rather than return a tuple of the function and a status? Could  instead do
```C++
auto [stop, f] = queue.pop();
```
with structured binding?

src/tasks/tasks.hpp
```diff
@@ -0,0 +1,488 @@
...
+    return result;
+  }
+
+  const std::vector<Task *> &GetIDs() const { return std::cref(dep); }
+
+  bool empty() const { return (!task && dep.size() == 0); }
+  Task *GetTask() { return task; }
+
+ private:
+  Task *task;
```
Yurlungur: I know it's null initialized but, for clarity:
```suggestion
  Task *task = nullptr;
```

src/tasks/tasks.hpp
```diff
@@ -0,0 +1,488 @@
...
+    if (task) {
+      result.dep.push_back(task);
+    } else {
+      result.dep.insert(result.dep.end(), dep.begin(), dep.end());
+    }
+    if (other.task) {
+      result.dep.push_back(other.task);
+    } else {
+      result.dep.insert(result.dep.end(), other.dep.begin(), other.dep.end());
+    }
```
Yurlungur: Why is this `if/else` branching required here? Why not just always add a task if present and allways append dependencies if present? e.g.,
```suggestion
    if (task != nullptr) result.dep.push_back(task);
    result.dep.insert(result.dep.end(), dep.begin(), dep.end());
    if (other.task != nullptr) result.dep.push_back(other.task);
    result.dep.insert(result.dep.end(), other.dep.begin(), other.dep.end());
```
Is there any danger that you could end up in a state where that wouldn't be correct?

src/tasks/tasks.hpp
```diff
@@ -0,0 +1,488 @@
...
+      if (num_calls == exec_limits.second) status = TaskStatus::complete;
+    }
+    // save the status in the Task object
+    SetStatus(status);
+    return status;
+  }
+  TaskID GetID() { return this; }
+  bool ready() {
+    // check that no dependency is incomplete
+    bool go = true;
```
Yurlungur, line 127:
```suggestion
    // set...
    bool go = true;
```
ready... set... go

src/tasks/tasks.hpp
```diff
@@ -0,0 +1,488 @@
...
+      // make status, request, and comm for this global task
+      global_status.push_back(std::make_shared<int>(0));
+      global_request.push_back(std::make_shared<MPI_Request>(MPI_REQUEST_NULL));
+      // be careful about the custom deleter so it doesn't call
+      // an MPI function after Finalize
+      global_comm.emplace_back(new MPI_Comm, [&](MPI_Comm *d) {
+        int finalized;
+        MPI_Finalized(&finalized);
+        if (!finalized) MPI_Comm_free(d);
+      });
```
Yurlungur: Neat. Should this be guarded in a macro checking if we're doing MPI?

Yurlungur (commented):
This is much cleaner than the old tasking machinery. Some things I like:
1. The graph walking is clear and concise, and dependencies are much easier to reasona bout.
2. The fact that sub lists are recursive means iterations and sublists and regional dependencies are all easier to reason about.
3. Threads!!!!

My one concern would be that walking the graph might be expensive in practice, but I suspect that it's not, compared to everything else we're doing. Especially since the task list can now be built once and not rebuilt (which is something that should probably be default behavior.

src/CMakeLists.txt
jdolence: already got em...

src/defs.hpp
jdolence: it's not for some of those files that don't exist anymore, but I think it's still relevent for the enums in tasks.hpp

src/tasks/tasks.hpp
jdolence: No I think that should work, and it's cleaner.

src/tasks/tasks.hpp
jdolence: yes it should.  I'll add that stuff.

src/tasks/thread_pool.hpp
jdolence: there isn't always a function.  returning it this way would require I do something silly like return an empty lambda or something.  really, i don't like either, so I'll leave it how it is unless I hear more objections.

src/tasks/thread_pool.hpp
jdolence: done

src/tasks/tasks.hpp
jdolence: done

src/tasks/tasks.hpp
jdolence: done

doc/sphinx/src/tasks.rst
```diff
@@ -120,21 +116,52 @@ is shown below.
...
+- ``TaskListStatus Execute(ThreadPool &pool)``: Execute all regions in the
+collection.  Regions are executed completely, in the order they were added,
+before moving on to the next region.  Task execution will take advantage of
+the provided ``ThreadPool`` to (possibly) execute tasks across ``TaskList``s
+in each region concurrently.
+- ``TaskListStatus Execute()``: Same as above, but execution will use an
+internally generated ``ThreadPool`` with a single thread.
+
+NOTE: Work remains to make the rest of
+Parthenon thread-safe, so it is currently not recommended to use a ``ThreadPool``
```
pdmullen: I'd say we should actually `PARTHENON_REQUIRE` running with a single thread.

example/poisson/poisson_driver.cpp
```diff
@@ -81,199 +81,167 @@ TaskCollection PoissonDriver::MakeTaskCollection(BlockList_t &blocks) {
...
     auto start_global_reduce =
-        (i == 0 ? tl.AddTask(loc_red, &AllReduce<Real>::StartReduce, &total_mass, MPI_SUM)
-                : none);
+        tl.AddTask(TaskQualifier::once_per_region, loc_red, &AllReduce<Real>::StartReduce,
+                   &total_mass, MPI_SUM);
 
-    auto start_rank_reduce =
-        (i == 0 ? tl.AddTask(rank_red, &Reduce<int>::StartReduce, &max_rank, 0, MPI_MAX)
-                : none);
+    auto start_rank_reduce = tl.AddTask(TaskQualifier::once_per_region, rank_red,
```
pdmullen, line 99: nice

src/solvers/bicgstab_solver.hpp
```diff
@@ -105,37 +104,33 @@ class BiCGSTABSolver {
...
+        printf("# [0] v-cycle\n# [1] rms-residual\n# [2] rms-error\n");
+      return TaskStatus::complete;
+    });
 
     // BEGIN ITERATIVE TASKS
+    auto [itl, solver_id] = tl.AddSublist(initialize, {1, params_.max_iters});
 
     // 1. u <- M p
-    auto precon1 = initialize;
+    auto precon1 = none;
```
pdmullen, line 118: So now in an iterative list, the first task should always have `none` as a dependency, since the `AddSublist` arguments take in a dependency themselves?

Does this seem transparent to the downstream user? I wonder if we should provide a `itl.begin` `TaskID` that is `== none`.

src/solvers/bicgstab_solver.hpp
```diff
@@ -262,11 +251,9 @@ class BiCGSTABSolver {
           solver->residual.val = 0.0;
           return TaskStatus::iterate;
         },
-        this, pmesh, i, params_.max_iters, params_.residual_tolerance);
-    region.AddGlobalDependencies(reg_dep_id, i, check);
-    reg_dep_id++;
+        this, pmesh, params_.residual_tolerance);
 
-    return check;
+    return solver_id;
```
pdmullen, line 256: Why aren't we returning `check`? If this is the new and correct API, I am not sure if I like it, since `check` is actually where we are assigning `TaskStatus::iterate` vs `TaskStatus::complete`.

src/tasks/tasks.hpp
```diff
@@ -0,0 +1,493 @@
...
+// This program was produced under U.S. Government contract 89233218CNA000001 for Los
+// Alamos National Laboratory (LANL), which is operated by Triad National Security, LLC
+// for the U.S. Department of Energy/National Nuclear Security Administration. All rights
+// in the program are reserved by Triad National Security, LLC, and the U.S. Department
+// of Energy/National Nuclear Security Administration. The Government is granted for
+// itself and others acting on its behalf a nonexclusive, paid-up, irrevocable worldwide
+// license in this material to reproduce, prepare derivative works, distribute copies to
+// the public, perform publicly and display publicly, and to permit others to do so.
+//========================================================================================
+
```
pdmullen: remove whitespace

src/tasks/tasks.hpp
pdmullen: lol

pdmullen (approved):

pdmullen:
To be clear, this is only a breaking change for those invoking iterative tasks, correct?

src/tasks/thread_pool.hpp
Yurlungur: No objections.

doc/sphinx/src/tasks.rst
jdolence: Yeah, I'll do that.

src/solvers/bicgstab_solver.hpp
jdolence: An iteration can now have more than one completion task.  The cleanest way I could come up with to support this kind of thing was to have an ID associated with the whole sublist (in this case solver_id).  So tasks that depend on an iteration being complete now depend on the iterative list being complete instead of a particular task within it.  I know what you mean, but I'd actually argue this is cleaner and clearer.  Dependent tasks don't depend on the completion task, they depend on the iteration being complete, whatever that means for the iteration you're trying to write (which may have multiple places where you might exit).

jdolence:
> To be clear, this is only a breaking change for those invoking iterative tasks, correct?

yes

doc/sphinx/src/tasks.rst
```diff
@@ -3,85 +3,81 @@
...
-with the bitwise or operator (``|``) to specify multiple dependencies.
+The ``TaskList`` class stores a vector of all the tasks and sublists (a nested
+``TaskList``) added to it.  Additionally, it stores various bookkeeping
+information that facilitate more advanced features described below.  Adding
+tasks and sublists are the only way to interact with ``TaskList`` objects.
+
+The basic call to ``AddTask`` takes the task's dependencies, the function to be
+executed, and the arguments to the function as its arguments.  ``AddTask`` returns
+a ``TaskID`` object that can be used in subsequent calls to ``AddTask`` as a
+dependency either on its own or combined with other ``TaskID``s via the ``|``
```
lroberts36, line 35: Maybe it is useful to provide a brief word of explanation for why we choose the bitwise `or` operator for what is sort of like a logical `and` operation (i.e.  `Task1 | Task2` means that this task depends on `Task1` *and* `Task2` being done). This seems like it should be a footnote, but not sure how to do that with rst.

doc/sphinx/src/tasks.rst
```diff
@@ -3,85 +3,81 @@
...
+Task
+----
+
+Though downstream codes never have to interact with the ``Task`` object directly,
+it's useful to describe nonetheless.  A ``Task`` object is essentially a functor
+that stores the necessary data to invoke a downstream code's functions with
+the desired arguments.  Importantly, however, it also stores information that
+relates itself to other tasks, namely the tasks that must be complete before
+it should execute and the tasks that may be available to run after it completes.
+In other words, ``Task``s are nodes in a directed (possibly cyclic) graph, and
```
lroberts36, line 21: Can a TL be cyclic and ever run to completion? I thought being cyclic implied that it would never complete.

doc/sphinx/src/tasks.rst
```diff
@@ -120,21 +116,52 @@ is shown below.
...
+lists in a region on a given MPI rank.  Tasks that depend on a ``local_sync``
+marked task gain dependencies from the corresponding task on all lists within
+a region.  A typical use for this qualifier is to do a rank-local reduction, for
+example before initiating a global MPI reduction (which should be done only once
+per rank, not once per ``TaskList``).  Note that Parthenon links tasks across
+lists in the order they are added to each list, i.e. the ``n``th ``local_sync`` task
+in a list is assumed to be associated with the ``n``th ``local_sync`` task in all
+lists in the region.
+- ``TaskQualifier::global_sync``: Tasks marked with ``global_sync`` implicitly have
+the same semantics as ``local_sync``, but additionally do a global reduction on the
```
lroberts36, line 153: Is there ever a use case for `global_sync`? It seems like there will always be some other MPI communication (like a reduction) associated with any point in the task list that requires all ranks to be at the same point.

example/poisson/poisson_driver.cpp
```diff
@@ -81,199 +81,167 @@ TaskCollection PoissonDriver::MakeTaskCollection(BlockList_t &blocks) {
 
     //--- Demo a few reductions
     // pass a pointer to the variable being reduced into
-    auto loc_red = tl.AddTask(none, poisson_package::SumMass<MeshData<Real>>, md.get(),
-                              &total_mass.val);
-    // make it a regional dependency so dependent tasks can't execute until all lists do
-    // this
-    solver_region.AddRegionalDependencies(reg_dep_id, i, loc_red);
-    reg_dep_id++;
```
lroberts36, line 89: Very glad to see this gone!

src/bvals/comms/boundary_communication.cpp
```diff
@@ -424,11 +424,7 @@ TaskID AddBoundaryExchangeTasks(TaskID dependency, TL_t &tl,
 }
 template TaskID AddBoundaryExchangeTasks<BoundaryType::any, TaskList>(
     TaskID, TaskList &, std::shared_ptr<MeshData<Real>> &, bool);
```
lroberts36: I think templating `AddBoundaryExchangeTasks` on `TL_t` is no longer necessary.

src/solvers/mg_solver.hpp
```diff
@@ -327,19 +318,16 @@ class MGSolver {
...
+          tl, communicate_to_coarse, partition, level - 1, min_level, max_level, pmesh);
 
       // 6. Receive error field into communication field and prolongate
       auto recv_from_coarser =
           tl.AddTask(coarser, ReceiveBoundBufs<BoundaryType::gmg_prolongate_recv>, md);
       auto set_from_coarser =
           tl.AddTask(recv_from_coarser, SetBounds<BoundaryType::gmg_prolongate_recv>, md);
-      auto prolongate = tl.AddTask(
+      auto prolongate = tl.AddTask( // TaskQualifier::local_sync, // is this required?
           set_from_coarser, ProlongateBounds<BoundaryType::gmg_prolongate_recv>, md);
```
lroberts36, line 330: I think this is not required.

src/tasks/tasks.hpp
```diff
@@ -0,0 +1,493 @@
...
+  TaskID operator|(const TaskID &other) const {
+    // calling this operator means you're building a TaskID to hold a dependency
+    TaskID result;
+    if (task != nullptr) result.dep.push_back(task);
+    result.dep.insert(result.dep.end(), dep.begin(), dep.end());
+    if (other.task != nullptr) result.dep.push_back(other.task);
+    result.dep.insert(result.dep.end(), other.dep.begin(), other.dep.end());
+    return result;
+  }
```
lroberts36: This looks like it adds more dependencies than are required. Maybe it is safer to do this, but couldn't you just add a check 
```c++
if (other.task == nullptr) result.dep.insert(result.dep.end(), other.dep.begin(), other.dep.end());
```

src/tasks/tasks.hpp
```diff
@@ -0,0 +1,493 @@
...
+
+  const std::vector<Task *> &GetIDs() const { return std::cref(dep); }
+
+  bool empty() const { return (!task && dep.size() == 0); }
+  Task *GetTask() { return task; }
```
lroberts36, line 84:
```suggestion
  Task *GetTask() { 
    if (task == nulltpr) PARTHENON_DEBUG_WARN("Returning a nullptr task.");
    return task; 
   }
```

src/tasks/tasks.hpp
```diff
@@ -0,0 +1,493 @@
...
+      bool do_mpi = false;
+#ifdef MPI_PARALLEL
+      // make status, request, and comm for this global task
+      global_status.push_back(std::make_shared<int>(0));
+      global_request.push_back(std::make_shared<MPI_Request>(MPI_REQUEST_NULL));
+      // be careful about the custom deleter so it doesn't call
+      // an MPI function after Finalize
+      global_comm.emplace_back(new MPI_Comm, [&](MPI_Comm *d) {
+        int finalized;
+        MPI_Finalized(&finalized);
```
lroberts36: I think the MPI calls should be wrapped in `PARTHENON_MPI_CHECK`

src/tasks/tasks.hpp
```diff
@@ -0,0 +1,493 @@
...
+  }
+  TaskStatus GetStatus() {
+    std::lock_guard<std::mutex> lock(mutex);
+    return task_status;
+  }
+  void reset_iteration() { num_calls = 0; }
+
+ private:
+  std::function<TaskStatus()> f;
+  std::array<std::vector<Task *>, 3> dependent;
```
lroberts36, line 157: What is `dependent` supposed to store? Would having `auto *dep_task = dependent[TaskStatus::complete][3]` imply that `dep_task` requires that the `task_status` of `this` is `TaskStatus::complete`?

src/tasks/tasks.hpp
```diff
@@ -0,0 +1,493 @@
...
+              assert(dependent.size() == 1);
+              auto mytask = *dependent.begin();
+              stat = 0;
+              for (auto dep : mytask->GetDependencies()) {
+                stat = std::max(stat, static_cast<int>(dep->GetStatus()));
+              }
+              MPI_Iallreduce(MPI_IN_PLACE, &stat, 1, MPI_INT, MPI_MAX, comm, &req);
+              return TaskStatus::complete;
+            },
+            exec_limits));
```
lroberts36: Is there a reason this can't be written using `parthenon::AllReduce`? It might simplify the code a little bit, but maybe not really worth making a change.

lroberts36 (approved):
This LGTM, it is great to have the tasking cleaned up so that it doesn't rely on bitmasks anymore and so that iterative task lists are more flexible!

doc/sphinx/src/tasks.rst
jdolence: Yeah not sure how to do the footnote, but I can add a comment.  Unfortunately, the reason is historical, going back to Athena++.  `|` was used because we were combining bit masks.  I agree that `&` would make more sense.  Thought about changing this, but not sure it's worth breaking every downstream code.

doc/sphinx/src/tasks.rst
jdolence: Maybe I'm not using quite the right words.  The graph can have cycles, but at least one node in the cycle acts as a conditional, either continuing the cycle or exiting it depending on some condition.  Is there a better way to describe that?

doc/sphinx/src/tasks.rst
jdolence: I think so.   The important distinction of `global_sync` is that it does a reduction on the statuses across all MPI ranks to determine if a task is complete or not.  Technically this may not be required if all MPI ranks are guaranteed to return the same status, as would be the case if they all checked some condition on some globally reduced quantity.  Given that, we can probably avoid using `global_sync` in some places where one might naively think that a reduction on statuses is required.  But I can imagine use cases where different ranks evaluate different things and the global reduction of statuses would be required.

src/bvals/comms/boundary_communication.cpp
jdolence: OK, I'll remove that then.

src/solvers/mg_solver.hpp
jdolence: OK.  Can't remember why it's there.  There must have been a regional dependency marked here before.

src/tasks/tasks.hpp
jdolence: I think you're right.  Those should just be if/else statements.  I'll fix that.

src/tasks/tasks.hpp
jdolence: Yep.  I'll do that.

src/tasks/tasks.hpp
jdolence: It stores a vector of tasks that might possibly be ready to run for a given task status returned by the owning Task.  So, for example, every Task points at itself for `TaskStatus::incomplete` so that if a task returns that status, it gets added to the queue again.  For `TaskStatus::complete`, the vector of tasks is just all the tasks that have this Task as a dependency.  Does that make sense?

src/tasks/tasks.hpp
jdolence: I'm not sure.  I'd have to think about that.  I suspect it's possible, but there might be some gotcha that's not immediately obvious.

jdolence:
@pgrete I think it would be good to have your eyes on this one, but if you're happy to just accept it without review just let me know.

src/solvers/bicgstab_solver.hpp
jdolence: I don't think it has to have `none`, but it often will.  There's perhaps a little misdirection going on, but for various reasons this ended up being the cleanest solution IMHO.  I think it should work identically if instead of `none` you use the same dependencies as the sublist, but giving the sublist a dependency was necessary (the details of why are now a little foggy for me, but I could probably figure it out quickly if we really want to know).

I'm not sure I follow your suggestion about `itl.begin`.

src/tasks/tasks.hpp
jdolence: I'm not sure we want this.  `GetTask` is called from the `Task` constructor and `nullptr` is a valid return value that has meaning, namely that the `TaskID` associated with the task's dependency is not itself tied to a single task, but a set of tasks.

src/solvers/mg_solver.hpp
lroberts36: I think I added it during a fit of paranoia while testing GMG

doc/sphinx/src/tasks.rst
lroberts36: I think it may be that we just have different mental models for the graph. In my mind, the graph is just defined by task dependencies and its structure is static. Each node in the graph is then associated with a status and I look for tasks that are available to do and try to switch their status. In this model, iterative task lists are just groupings of tasks in the graph where a task can return iterate and that sets the status of all tasks in the iterative list to incomplete.

src/solvers/mg_solver.hpp
jdolence: Yeah current develop has a regional dependency here.  Do you think that's an error then?  Should I fix it here?

lroberts36: Not really an error, just unnecessary. You could leave it here and I will fix it in the MG performance PR.

doc/sphinx/src/tasks.rst
lroberts36: My thought is that if we want to sync all ranks, there is always going to be some information that gets communicated between them (more than just that they are done). Maybe I am not being imaginative enough though.

doc/sphinx/src/tasks.rst
lroberts36: Yeah, I think there is no reason to switch it and I understand why it was done. My concern is now that you removed bitmasks it won't make sense if you look at the code internals.

src/tasks/tasks.hpp
lroberts36: Fair enough. This should probably never be called in down stream code so it is probably safe. I guess this is just an issue that comes out of having `TaskID` do double duty as a single `TaskID` and a collection of tasks ids (which I understand the reason for).

src/tasks/tasks.hpp
lroberts36: yep, that makes sense, thanks. A brief comment in the code might be helpful in the future.

doc/sphinx/src/tasks.rst
jdolence: let me try to make something up quickly.  Say we're pushing particles around that can move across many cells in a given time step.  One algorithm you could code up would have every rank push all the particles that live on it's blocks until the end of the time step or until it leaves the rank's domain (at which point it sends them to the right neighbor).  You could write this as an iterative thing with the completion check being a check of whether all particles on the rank have reached the end of the time step.  Then the completion check is totally local to the rank, but you need to continue iterating globally until everybody agrees things are done.

src/solvers/mg_solver.hpp
jdolence: right.  error was the wrong word.  ok, i'll leave it to you to fix in your PR.

src/tasks/tasks.hpp
jdolence: done

doc/sphinx/src/tasks.rst
lroberts36: Sure, but that could also just be written as a reduction to get the total number of active particles in the simulation.

jdolence:
@pgrete, I'd like to get this merged ASAP.  A couple different codes want to use the new mechanism to express iterations.  Think you can review soon?  Should we just merge?  How should we proceed?

CMakeLists.txt
```diff
@@ -116,6 +116,9 @@ endif()
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 find_package(Filesystem REQUIRED COMPONENTS Experimental Final)
 
+# Require threading for tasks
+find_package(Threads)
+
```
pgrete, line 121: Is there any way this could fail?
If threading for the task list is currently not tested/used anyway, then forcing this dependency might be a little strict, isn't it?

doc/sphinx/src/tasks.rst
```diff
@@ -3,85 +3,83 @@
...
+``AddTask`` signifies that the task has no dependencies.
+
+The ``AddSublist`` function adds a nested ``TaskList`` to the ``TaskList`` on
+which its called.  The principle use case for this is to add iterative cycles
+to the graph, allowing one to execute a series of tasks repeatedly until some
+criteria are satisfied.  The call takes as arguments the dependencies (via
+``TaskID``s combined with ``|``) that must be complete before the sublist
+exectues and, optionally, a ``std::pair<int, int>`` specifying the minimum
+and maximum number of times the sublist should execute.  Passing something like
+``{min_iters, max_iters}`` as the second argument should suffice.  ``AddSublist``
```
pgrete: Is it worth adding a (large) default max number of iterations (similar to our MPI recv timeout)?

I'm asking as sth similar for IO would have saved us an ***** amount of computing time last year so I'm now wondering if there are more places where such a sanity check would be useful.

src/solvers/bicgstab_solver.hpp
```diff
@@ -78,10 +78,10 @@ class BiCGSTABSolver {
     pkg->AddField(p::name(), m_no_ghost);
   }
 
-  TaskID AddTasks(TaskList &tl, IterativeTasks &itl, TaskID dependence, int i,
-                  Mesh *pmesh, TaskRegion &region, int &reg_dep_id) {
+  TaskID AddTasks(TaskList &tl, TaskID dependence, Mesh *pmesh, const int partition) {
     using namespace utils;
-    auto &md = pmesh->mesh_data.GetOrAdd("base", i);
+    TaskID none;
+    auto &md = pmesh->mesh_data.GetOrAdd("base", partition);
```
pgrete, line 84: Flagging https://github.com/parthenon-hpc-lab/parthenon/issues/959 again.

I'm actually starting to wonder if we should remove support (i.e., make it technically impossible) to vary the partition size as quite a few place are now making some implicit assumption through using `GetOrAdd`.

PS: This is independent of this PR.

src/tasks/tasks.hpp
```diff
@@ -0,0 +1,504 @@
...
+  static inline constexpr qualifier_t local_sync{1 << 0};
+  static inline constexpr qualifier_t global_sync{1 << 1};
+  static inline constexpr qualifier_t completion{1 << 2};
+  static inline constexpr qualifier_t once_per_region{1 << 3};
+
+  bool LocalSync() const { return flags & local_sync; }
+  bool GlobalSync() const { return flags & global_sync; }
+  bool Completion() const { return flags & completion; }
+  bool Once() const { return flags & once_per_region; }
+  bool Valid() const { return true; }
```
pgrete: What's this required for?
It's seems to be only used once in an assert statement below.
Is this implicitly checking if the passed object below is not a `nullptr`?

src/tasks/tasks.hpp
```diff
@@ -0,0 +1,504 @@
...
+  TaskStatus operator()() {
+    auto status = f();
+    if (task_type == TaskType::completion) {
+      // keep track of how many times it's been called
+      num_calls += (status == TaskStatus::iterate || status == TaskStatus::complete);
+      // enforce minimum number of iterations
+      if (num_calls < exec_limits.first && status == TaskStatus::complete)
+        status = TaskStatus::iterate;
+      // enforce maximum number of iterations
+      if (num_calls == exec_limits.second) status = TaskStatus::complete;
```
pgrete, line 118: Would it make sense to have another status like `::ran_out_of_iterations_before_X`?

src/tasks/tasks.hpp
```diff
@@ -0,0 +1,504 @@
...
+              if (check) {
+                return static_cast<TaskStatus>(stat);
+              }
+              return TaskStatus::incomplete;
+            },
+            exec_limits));
+#endif         // MPI_PARALLEL
+      } else { // unique_id != 0
+        // just add empty tasks
+        tasks.push_back(std::make_shared<Task>(
```
pgrete, line 287: Those are required in order to keep the ids in sync (as mentioned somewhere else), correct?

src/tasks/tasks.hpp
```diff
@@ -0,0 +1,504 @@
...
+  TaskRegion() = delete;
+  explicit TaskRegion(const int num_lists) : task_lists(num_lists) {
+    for (int i = 0; i < num_lists; i++)
+      task_lists[i].SetID(i);
+  }
+
+  TaskListStatus Execute(ThreadPool &pool) {
+    // for now, require a pool with one thread
+    PARTHENON_REQUIRE_THROWS(pool.size() == 1,
+                             "ThreadPool size != 1 is not currently supported.")
```
pgrete, line 399: :+1: for being explicit

pgrete (approved):
Thanks for the interface cleanup (and infrastructure below) and sorry for the delay (I lost track of this PR after I reviewed the other after the sync).

I only left some clarifying questions.
Otherwise, a more general question would whether the new execute machinery benefits from the threaded implementation as is, or, in other words if we don't support that feature anyway, could this functionality be stashed until it'll actually be used so that there's less "to-be-used-in-the-future(-but-maintained-already-now)" code?

c-prather:
@pgrete I think there's a chance the current threading may already benefit my use case, because I spend a lot of CPU time building variable lists & packs, so there's the potential that any reduced synchronicity speeds things up for me in at least some cases.
I can try to scare up some performance numbers if that would help -- otherwise, I was going to put off moving to this branch/latest Parthenon until outputs/restarts of face-centered fields land, to avoid more fun franken-branch merge conflict bugs.

CMakeLists.txt
jdolence: Frankly I'm not sure.  I haven't seen it fail yet and it seems to pass all our CI tests, but certainly I won't claim to have tried this everywhere.  This is the standard way of enabling threads via cmake, so I'd like to think it's fairly robust.

I think I'll leave this in for now.  If anyone encounters problems, I can try to spin up a solution quickly, which may mean making threads a compile time option.  If it turns out not to be needed, that's just unnecessary complexity though.

doc/sphinx/src/tasks.rst
jdolence: I thought about that some and it's possible I made the wrong choice.  I can see arguments for `max_iters = 1` as the default and `max_iters = some_huge_number` as the default.  I can also see an argument for not providing any default and just requiring downstream users to provide a `{min, max}`.  From the parthenon perspective, no defaults is probably the safest -- it's on the downstream developer to provide something appropriate for what they're doing.  I might just do that.

src/tasks/tasks.hpp
jdolence: It's actually doing nothing now.  When I was first implementing all this, not all combinations of qualifiers were supported, so this was checking to make sure someone didn't try something that wouldn't work.  I think I'll just remove it.

src/tasks/tasks.hpp
jdolence: probably.  I suppose that would allow adding some kind of option that would warn or abort if an iteration failed to properly complete.  That seems like a good idea.  I'll either add an issue so I can deal with this later or add the feature if it looks simple enough.

src/tasks/tasks.hpp
jdolence: Something like that.  It mostly helps in keeping all the dependencies straight, I think.

jdolence:
> Thanks for the interface cleanup (and infrastructure below) and sorry for the delay (I lost track of this PR after I reviewed the other after the sync).
> 
> I only left some clarifying questions. Otherwise, a more general question would whether the new execute machinery benefits from the threaded implementation as is, or, in other words if we don't support that feature anyway, could this functionality be stashed until it'll actually be used so that there's less "to-be-used-in-the-future(-but-maintained-already-now)" code?

The threads currently provide no value, but I've already started working on making other parts of parthenon thread-safe to enable nthreads != 1, so I'm hoping this code won't sit around for long in it's current state.  Given that, I'd rather not do the extra work of removing here just to add back very soon.  It's a fair point/question though -- I'll try to be quick about getting threads working more fully.

CMakeLists.txt
pgrete: sounds good.

pgrete:
> @pgrete I think there's a chance the current threading may already benefit my use case, because I spend a lot of CPU time building variable lists & packs, so there's the potential that any reduced synchronicity speeds things up for me in at least some cases.


That's actually an interesting/important use-case/questions: Is our packing machinery thread safe?

> I can try to scare up some performance numbers if that would help -- otherwise, I was going to put off moving to this branch/latest Parthenon until outputs/restarts of face-centered fields land, to avoid more fun franken-branch merge conflict bugs.

No worries. I think as long we have the handbrakes on (enforcing `nthreads=1`) for now by default, don't spent your coding cycles on franken-branches.

pgrete:
Not sure why the status for CI are not reported. I re-triggering.

pgrete:
CI failed and looks like a legit fail.

jdolence:
> CI failed and looks like a legit fail.

Looks that way, I think.  Haven't been able to reproduce though...

PR #988, opened 2023-12-14
Fix neighbor finding bug

c-prather:
## PR Summary

This PR fixes a bug in the new neighbor finding algorithm in `Mesh::SetSameLevelNeighbors`. Previously, for meshes that only contained only a few blocks in each direction, which had periodic boundaries, and contained blocks at multiple levels of refinement, certain fine blocks that did not touch a nearby coarse block could be flagged as a neighbor. This was because the parent block of the fine block was a neighbor of the coarse block and `LogicalLocation::GetSameLevelOffsets` was finding offsets for this block and then it was never checked whether the fine block was actually a neighbor. This PR adds the required check. The second bug that Ben originally found in `bound_info.cpp` was related to part of the `NeighborBlock` object not being set by the old neighbor finding routines, which he had switched to using to get things to work. This is no longer an issue since the new neighbor finding routine has been updated and we are using its output for doing boundary communication.

Original summary from @bprather:

This fixes two bugs I found by dumb bisection, which I do not yet really understand.

That's not quite true.  I do understand the piece in `loop_utils.hpp` thanks to @lroberts36. The neighbor block list returned by `SetSameLevelNeighbors` contains repeat entries, so checking twice (sometimes 4 times?) can lead to race conditions (in some nebulous way I do *not* understand).

However, I don't get what the piece in `bnd_info.cpp` is doing.  Or rather, I can more or less follow the original form I reverted to, but I have no idea why the change was made.

I'm nearly sure this change breaks something (GMG?) given the original change was likely made for a reason.  And, while I can confirm that this fixes the issues with <=2 meshblocks in `parthenon-mhd` and KHARMA, I'm not sure that there aren't more bugs related to `neighbors` having duplicates -- the underlying fix should include addressing that.

So, this is still WIP.  But I wanted to get the patch out in case this helps others.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ### Changed (changing behavior/API/variables/...)
 
 ### Fixed (not changing behavior/API/variables/...)
+- [[PR988]](https://github.com/parthenon-hpc-lab/parthenon/pull/988) Fix bug in neighbor finding routine for small, periodic, refined meshes
 - [[PR978]](https://github.com/parthenon-hpc-lab/parthenon/pull/978) remove erroneous sparse check
 
 ### Infrastructure (changes irrelevant to downstream codes)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 ### Fixed (not changing behavior/API/variables/...)
 - [[PR988]](https://github.com/parthenon-hpc-lab/parthenon/pull/988) Fix bug in neighbor finding routine for small, periodic, refined meshes
+- [[PR986]](https://github.com/parthenon-hpc-lab/parthenon/pull/986) Fix bug in sparse boundary communication BndInfo cacheing
 - [[PR978]](https://github.com/parthenon-hpc-lab/parthenon/pull/978) remove erroneous sparse check
 
 ### Infrastructure (changes irrelevant to downstream codes)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Current develop
 
 ### Added (new features/APIs/variables/...)
+- [[PR 981]](https://github.com/parthenon-hpc-lab/parthenon/pull/981) Add IndexSplit
 - [[PR 983]](https://github.com/parthenon-hpc-lab/parthenon/pull/983) Add Contains to SparsePack
 - [[PR 968]](https://github.com/parthenon-hpc-lab/parthenon/pull/968) Add per package registration of boundary conditions
 - [[PR 948]](https://github.com/parthenon-hpc-lab/parthenon/pull/948) Add solver interface and update Poisson geometric multi-grid example
--- a/benchmarks/burgers/burgers_diff.py
+++ b/benchmarks/burgers/burgers_diff.py
@@ -12,6 +12,7 @@
 # the public, perform publicly and display publicly, and to permit others to do so.
 # ========================================================================================
 
+import sys
 import numpy as np
 from argparse import ArgumentParser
 
@@ -53,4 +54,4 @@ def compare_files(file1, file2, tolerance, print_results=True):
 
 if __name__ == "__main__":
     args = parser.parse_args()
-    compare_files(args.file1, args.file1, args.tolerance, True)
+    sys.exit(compare_files(args.file1, args.file1, args.tolerance, True))
--- a/doc/sphinx/src/nested_par_for.rst
+++ b/doc/sphinx/src/nested_par_for.rst
@@ -51,7 +51,7 @@ Data type for memory in scratch pad/cache memory. Use
 documentation <https://kokkos.github.io/kokkos-core-wiki/ProgrammingGuide/HierarchicalParallelism.html?highlight=hierarchical>`__
 for determining scratch pad memory needs before kernel launch.
 
-Important usage hints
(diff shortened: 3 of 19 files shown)


c-prather:
The replacement fix works for me in KHARMA.  I can also verify that the old and new neighbor lists now always match, at least for my test mesh setups.

Yurlungur (approved):
LGTM. Is there any way we could add a unit test or something to check for correctness here?

pgrete:
I now tested this fix downstream and unfortunately it doesn't fix the issue on our end.
This is the mesh structure (it's a fully refined mesh but using the `adaptive` machinery to create it not that `static` one with periodic boundary conditions)
```
Root grid = 2 x 2 x 2 MeshBlocks
Total number of MeshBlocks = 512
Number of physical refinement levels = 2
Number of logical  refinement levels = 3
  Physical level = 2 (logical level = 3): 512 MeshBlocks, cost = 512
```
running with `srun -N 16 -n 128 ...` immediately results in
```

cycle=10823 time=3.5000692168925700e+00 dt=1.1150882225762992e-04 zone-cycles/wsec_step=0.00e+00 wsec_total=4.11e+01 wsec_step=5.12e+01 zone-cycles/wsec=0.00e+00 wsec_AMR=0.00e+00
Root grid = 2 x 2 x 2 MeshBlocks
Total number of MeshBlocks = 512
Number of physical refinement levels = 2
Number of logical  refinement levels = 3
  Physical level = 2 (logical level = 3): 512 MeshBlocks, cost = 512
### PARTHENON ERROR
  Condition:   *nrecv_tries_ < 1e8
  Message:     MPI probably hanging after 1e8 receive tries.
  File:        /project/project_465000552/env/src/athenapk/external/parthenon/src/utils/communication_buffer.hpp
  Line number: 331
MPI probably hanging after 1e8 receive tries.
```
whereas running with fewer ranks `srun -N 8 -n 64 ...` works like a charm.

Any ideas where to look next (changing from periodic to outflow boundary conditions did seem to *fix* the issue so it might still be related to the proposed change here)?

On a related note: Did you encounter this issue as part of using the GMG machinery @bprather ?
I'm asking as the fix is in the `gmg_...` source file, which is probably not where I'd have looked when not running with GMG.

lroberts36:
@pgrete: 

> I now tested this fix downstream and unfortunately it doesn't fix the issue on our end. This is the mesh structure (it's a fully refined mesh but using the `adaptive` machinery to create it not that `static` one with periodic boundary conditions)

I would have been surprised if this fixed the issue you are seeing. I *think* this issue that is fixed in this PR should only happen when you have a mesh that is covered by 2 x 2 (x 2) blocks at the coarsest level and only one level of refinement.  

Even if this doesn't fix the issue you are seeing, this PR definitely fixes a bug. Can you approve the changes if they are ok with you? 

> 
> Any ideas where to look next (changing from periodic to outflow boundary conditions did seem to _fix_ the issue so it might still be related to the proposed change here)?

I would be surprised if the neighbor finding somehow depends on the number of ranks. Did you try running with @bprather's original fix  [d97dd26](https://github.com/parthenon-hpc-lab/parthenon/pull/988/commits/d97dd26a4ed2b6adb8d762483c929b449ed2557b)? This just reverted to using the results of the old Athena++ neighbor finding routine, which I would expect to give identical results to the new neighbor finding code on the leaf grid. Another way you could check this is to confirm that `MeshBlock::neighbors` contains the same set of neighbors as the old neighbor array `MeshBlock::pbval::neighbor`.

Also, I can't remember if you mentioned this, but did you just try increasing the hang threshold? The amount of time required to set up the mesh may have just changed.

> 
> On a related note: Did you encounter this issue as part of using the GMG machinery @bprather ? I'm asking as the fix is in the `gmg_...` source file, which is probably not where I'd have looked when not running with GMG.

This neighbor finding routine ended up in `mesh-gmg.cpp` since it is used at each GMG level (in addition to being used for the leaf grid now). I am happy to have it moved if this is a confusing place to have it.

c-prather:
I agree neighbor finding should move somewhere more discoverable eventually, probably in the context of eliminating `MeshBlockTree` entirely as @lroberts36 has proposed, which I think would fix the real source of confusion here.

I think your mesh setup should trigger the issue if it were statically refined by one level in one block, but I don't know about AMR.  I tried starting a few AMR simulations with 2x2 base grids just since I haven't tested it before, but as I suppose I should have expected, I encountered neither my error nor yours (I wasn't using very many AMR ranks or levels).

The one way I could see your issue might be related is if MPI behavior was somehow dependent on the ordering of the neighbor list, which has changed in moving from the old to the new neighbor-finding.  You can try reverting to the old list in `loop_utils` by looking through the change history here, that's how I originally worked around this issue.  Even if the revert fixes things, though, it's just exposing an order-dependence bug somewhere else, which we should address separately -- so either way, I agree we should get this PR in now and deal with the hang in a new one.

pdmullen (approved):
Happy to approve these changes since it is fixing a bug, however, I agree with others that this code seems somewhat misplaced since @bprather was experiencing this bug even without invoking GMG.

pgrete (approved):
Along with the discussion I opened #995 and I'll keep track of the bug I'm chasing separately.
So triggering auto-merge now.

PR #989, opened 2023-12-14
[Breaking Change][WIP] Make Mesh-Level Boundary Conditions Usable without "User" flag

Yurlungur:
## API Breaking change

When registering user boundary functions, the mechanism changes from assigning the function pointer to an array. You now must call `ApplicationInput::RegisterBoundaryCondition` and give the custom boundary condition a name. This also eliminates the need for the `user` boundary flag in the input file, although that flag is implicitly still supported.

## PR Summary

A discussion on the new downstream code headed by @brryan @pdmullen indicated a desire to set mesh-level boundary conditions without using the `user` flag. Here I implement this feature. The changes I made are:
1. `ApplicationInput` now has a registration function where you register either a grid or particle boundary condition, with a function pointer (or a class in the case of swarm bounds) and a name.
2. ALL boundary conditions (except periodic) use this mechanism. Reflecting boundaries are, for example, registered in the constructor for `ApplicationInput` with the `"reflecting"` name.
3. `BoundaryFlag`s are dramatically simplified, indicating (essentially) only whether or not boundaries are periodic.
4. The mesh no longer uses flags to set function pointers. Instead, it just pulls the functions from `ApplicationInput`.
5. For (sort of) backwards compatibility, if you register a function without naming it, it's named "user."
6. Swarm boundaries are now decoupled from mesh boundaries. Although by default, they are set to the same value as the mesh.

The advantage of this is that you can now call (in your main function)
```C++
pman->app_input.RegisterBoundaryCondition(parthenon::BoundaryFace::inner_x1, "my_bc", &my_bc_func);
```
and then in the input file just use
```
<parthenon/mesh>
ix1_bc = my_bc
```

I think the refactored code is also significantly cleaner and easier to read. Also I put a link to the parthenon paper in the readme.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/src/application_input.hpp
+++ b/src/application_input.hpp
@@ -16,6 +16,7 @@
 #include <functional>
 #include <map>
 #include <memory>
+#include <sstream>
 #include <string>
 #include <unordered_map>
 #include <vector>
@@ -23,6 +24,8 @@
 #include "bvals/boundary_conditions.hpp"
 #include "defs.hpp"
 #include "interface/state_descriptor.hpp"
+#include "interface/swarm_boundaries.hpp"
+#include "kokkos_abstraction.hpp"
 #include "parameter_input.hpp"
 #include "parthenon_arrays.hpp"
 #include "utils/error_checking.hpp"
@@ -31,11 +34,61 @@ namespace parthenon {
 
 class ApplicationInput {
  public:
-  ApplicationInput();
+  ApplicationInput() {
+    RegisterBoundaryCondition(BoundaryFace::inner_x1, "outflow",
+                              &BoundaryFunction::OutflowInnerX1);
+    RegisterBoundaryCondition(BoundaryFace::outer_x1, "outflow",
+                              &BoundaryFunction::OutflowOuterX1);
+    RegisterBoundaryCondition(BoundaryFace::inner_x2, "outflow",
+                              &BoundaryFunction::OutflowInnerX2);
+    RegisterBoundaryCondition(BoundaryFace::outer_x2, "outflow",
+                              &BoundaryFunction::OutflowOuterX2);
+    RegisterBoundaryCondition(BoundaryFace::inner_x3, "outflow",
+                              &BoundaryFunction::OutflowInnerX3);
+    RegisterBoundaryCondition(BoundaryFace::outer_x3, "outflow",
+                              &BoundaryFunction::OutflowOuterX3);
+    RegisterBoundaryCondition(BoundaryFace::inner_x1, "reflecting",
+                              &BoundaryFunction::OutflowInnerX1);
+    RegisterBoundaryCondition(BoundaryFace::outer_x1, "reflecting",
+                              &BoundaryFunction::ReflectOuterX1);
+    RegisterBoundaryCondition(BoundaryFace::inner_x2, "reflecting",
+                              &BoundaryFunction::ReflectInnerX2);
+    RegisterBoundaryCondition(BoundaryFace::outer_x2, "reflecting",
+                              &BoundaryFunction::ReflectOuterX2);
+    RegisterBoundaryCondition(BoundaryFace::inner_x3, "reflecting",
+                              &BoundaryFunction::ReflectInnerX3);
+    RegisterBoundaryCondition(BoundaryFace::outer_x3, "reflecting",
+                              &BoundaryFunction::ReflectOuterX3);
+
+    // Currently outflow and periodic are the only particle BCs available
+    RegisterSwarmBoundaryCondition(BoundaryFace::inner_x1, "outflow",
+                                   &DeviceAllocate<ParticleBoundIX1Outflow>);
+    RegisterSwarmBoundaryCondition(BoundaryFace::outer_x1, "outflow",
+                                   &DeviceAllocate<ParticleBoundOX1Outflow>);
+    RegisterSwarmBoundaryCondition(BoundaryFace::inner_x2, "outflow",
+                                   &DeviceAllocate<ParticleBoundIX2Outflow>);
+    RegisterSwarmBoundaryCondition(BoundaryFace::outer_x2, "outflow",
+                                   &DeviceAllocate<ParticleBoundOX2Outflow>);
+    RegisterSwarmBoundaryCondition(BoundaryFace::inner_x3, "outflow",
+                                   &DeviceAllocate<ParticleBoundIX3Outflow>);
+    RegisterSwarmBoundaryCondition(BoundaryFace::outer_x3, "outflow",
+                                   &DeviceAllocate<ParticleBoundOX3Outflow>);
+    RegisterSwarmBoundaryCondition(BoundaryFace::inner_x1, "periodic",
+                                   &DeviceAllocate<ParticleBoundIX1Periodic>);
+    RegisterSwarmBoundaryCondition(BoundaryFace::outer_x1, "periodic",
+                                   &DeviceAllocate<ParticleBoundOX1Periodic>);
+    RegisterSwarmBoundaryCondition(BoundaryFace::inner_x2, "periodic",
+                                   &DeviceAllocate<ParticleBoundIX2Periodic>);
+    RegisterSwarmBoundaryCondition(BoundaryFace::outer_x2, "periodic",
+                                   &DeviceAllocate<ParticleBoundOX2Periodic>);
+    RegisterSwarmBoundaryCondition(BoundaryFace::inner_x3, "periodic",
+                                   &DeviceAllocate<ParticleBoundIX3Periodic>);
+    RegisterSwarmBoundaryCondition(BoundaryFace::outer_x3, "periodic",
+                                   &DeviceAllocate<ParticleBoundOX3Periodic>);
+  }
 
   // ParthenonManager functions
-  std::function<Packages_t(std::unique_ptr<ParameterInput> &)>
-  ProcessPackages = nullptr;
+  std::function<Packages_t(std::unique_ptr<ParameterInput> &)> ProcessPackages = nullptr;
 
   // Mesh functions
   std::function<void(Mesh *, ParameterInput *)> InitUserMeshData = nullptr;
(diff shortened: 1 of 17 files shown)


pdmullen (approved):
I really like the new API.  I am a bit confused by two things, 

(1) Do you think this has to be a breaking change? You have already kept the `user` flag functionality in place... it seems the only breaking change is the function signature for `RegisterBoundaryCondition`.  I wonder if there is a way to overload the function (or similar) so this doesn't need to be breaking.  

(2) There are some changes to `bvals*.cpp` and `mesh.cpp`, etc. that don't seem to be related to the new user-defined BCs API, but rather a cleanup.  Someone besides me should review these with a careful eye --- I worry that these changes could easily slip in some undesired behavior.

src/mesh/mesh.cpp
```diff
@@ -1163,7 +1161,6 @@ bool Mesh::SetBlockSizeAndBoundaries(LogicalLocation loc, RegionSize &block_size
 
 RegionSize Mesh::GetBlockSize(const LogicalLocation &loc) const {
   RegionSize block_size = GetBlockSize();
-  bool valid_region = true;
```
Yurlungur: Oops this was not supposed to be deleted.

Yurlungur:
> I really like the new API. I am a bit confused by two things,
> 
> (1) Do you think this has to be a breaking change? You have already kept the `user` flag functionality in place... it seems the only breaking change is the function signature for `RegisterBoundaryCondition`. I wonder if there is a way to overload the function (or similar) so this doesn't need to be breaking.

Hmm... I could do this I think. Is that really desirable? It would be a little bit of extra code for, IMO, not great justification. Sure, keeping full backwards compatibility is nice. But migrating to the new API should be very low effort for the downstream codes. I can add an overload though if you think it's important.

> (2) There are some changes to `bvals*.cpp` and `mesh.cpp`, etc. that don't seem to be related to the new user-defined BCs API, but rather a cleanup. Someone besides me should review these with a careful eye --- I worry that these changes could easily slip in some undesired behavior.

I *think* I only did cleanup related to the new BCs API. Basically, with the `ApplicationInput` machinery now owning this dictionary of function pointers, it doesn't make sense to keep carrying around an array of flags, as the function pointers can just be pulled from `ApplicationInput`. Happy to wait for other eyes though. Maybe from @lroberts36, @forrestglines , or @pgrete ? I'd like an AthenaPK person's eyes on this, since it's a breaking change.

pdmullen:
> > I really like the new API. I am a bit confused by two things,
> > (1) Do you think this has to be a breaking change? You have already kept the `user` flag functionality in place... it seems the only breaking change is the function signature for `RegisterBoundaryCondition`. I wonder if there is a way to overload the function (or similar) so this doesn't need to be breaking.
> 
> Hmm... I could do this I think. Is that really desirable? It would be a little bit of extra code for, IMO, not great justification. Sure, keeping full backwards compatibility is nice. But migrating to the new API should be very low effort for the downstream codes. I can add an overload though if you think it's important.
> 
> > (2) There are some changes to `bvals*.cpp` and `mesh.cpp`, etc. that don't seem to be related to the new user-defined BCs API, but rather a cleanup. Someone besides me should review these with a careful eye --- I worry that these changes could easily slip in some undesired behavior.
> 
> I _think_ I only did cleanup related to the new BCs API. Basically, with the `ApplicationInput` machinery now owning this dictionary of function pointers, it doesn't make sense to keep carrying around an array of flags, as the function pointers can just be pulled from `ApplicationInput`. Happy to wait for other eyes though. Maybe from @lroberts36, @forrestglines , or @pgrete ? I'd like an AthenaPK person's eyes on this, since it's a breaking change.

I definitely don't mind the breaking change.  Although I felt some pressure from Parthenon syncs to try and limit this where possible.

src/application_input.hpp
```diff
@@ -16,19 +16,77 @@
...
+    RegisterBoundaryCondition(BoundaryFace::outer_x1, "reflecting",
+                              &BoundaryFunction::ReflectOuterX1);
+    RegisterBoundaryCondition(BoundaryFace::inner_x2, "reflecting",
+                              &BoundaryFunction::ReflectInnerX2);
+    RegisterBoundaryCondition(BoundaryFace::outer_x2, "reflecting",
+                              &BoundaryFunction::ReflectOuterX2);
+    RegisterBoundaryCondition(BoundaryFace::inner_x3, "reflecting",
+                              &BoundaryFunction::ReflectInnerX3);
+    RegisterBoundaryCondition(BoundaryFace::outer_x3, "reflecting",
+                              &BoundaryFunction::ReflectOuterX3);
```
pgrete, line 61: I'm wondering if we should register these at all (or register those in the respective applications that use them).
The reason is that reflecting boundary conditions depend either on properly using "vectors" (with the correct number of dimensions), i.e., with downstream application using an Athen~a~-esque approach will not work out of the box.
Not registering them here would allow downstream codes to either explicitly implement and register them or fail loudly (when a user tries to use them in the input file without knowing the Parthenon internals).

Bonus point: This would be in line with the registered `Swarm` functions.

src/mesh/mesh.cpp
```diff
@@ -79,13 +79,25 @@ Mesh::Mesh(ParameterInput *pin, ApplicationInput *app_in, Packages_t &packages,
...
-          GetBoundaryFlag(pin->GetOrAddString("parthenon/mesh", "ox2_bc", "reflecting")),
-          GetBoundaryFlag(pin->GetOrAddString("parthenon/mesh", "ix3_bc", "reflecting")),
-          GetBoundaryFlag(pin->GetOrAddString("parthenon/mesh", "ox3_bc", "reflecting"))},
+      mesh_bc_names{pin->GetOrAddString("parthenon/mesh", "ix1_bc", "reflecting"),
+                    pin->GetOrAddString("parthenon/mesh", "ox1_bc", "reflecting"),
+                    pin->GetOrAddString("parthenon/mesh", "ix2_bc", "reflecting"),
+                    pin->GetOrAddString("parthenon/mesh", "ox2_bc", "reflecting"),
+                    pin->GetOrAddString("parthenon/mesh", "ix3_bc", "reflecting"),
+                    pin->GetOrAddString("parthenon/mesh", "ox3_bc", "reflecting")},
+      // TODO(JMM): This probably needs a per-package treatment too
```
pgrete, line 88: What exactly do you mean? Are you referring to the additional per-package boundary conditions?
And should this be addressed in this PR?

src/bvals/bvals_interfaces.hpp
```diff
@@ -85,7 +85,7 @@ enum {
 // int to index raw arrays (not ParArrayNDs)--> enumerator vals are explicitly specified
 
 // identifiers for boundary conditions
-enum class BoundaryFlag { block = -1, undef, reflect, outflow, periodic, user };
+enum class BoundaryFlag { block = -1, undef, periodic, user };
```
pgrete, line 88: General question: What are "block" boundary conditions?
I looked at the doc and didn't find a corresponding mention. Did I miss it?
Are these for embedded boundaries? Do we support those? Are they used anywhere?

pgrete (commented):
Looks very good to me. I think this is a much cleaner approach (which also allows for cleaner downstream use).
Also I don't mind this being breaking change as downstream codes could implement this change in a non-breaking way (as far as I could tell).
The only thing I'd like to discuss before merge is about the default registration of reflecting boundary conditions (see separate comment).

src/application_input.hpp
Yurlungur: Currently reflecting is the default boundary condition, so that would be a more breaking change. I wouldn't necessarily object to this, though, so long as we still provide ways of adding reflecting and document it.

Would curious on @jdolence and @lroberts36 's opinion on this one, as riot uses reflecting BCs heavily.

src/mesh/mesh.cpp
Yurlungur: So Luke's PR #968 makes boundary conditions per-package. But I'd like to hook that into input parsing at some point. That's what this TODO is for.

src/bvals/bvals_interfaces.hpp
Yurlungur: Block BCs are inherited from Athena++. I have no idea how they work and I don't think we support them. I can remove the flag/all references if you like.

src/application_input.hpp
pgrete: Do you/we have any input files that reply on implicitly specified default boundary conditions?
While this being a breaking change it may not be that breaking in practices (apart from adding the register functions downstream).

src/mesh/mesh.cpp
pgrete: Maybe keep track as an issue then, too (on top of the `TODO` in the code)?

src/bvals/bvals_interfaces.hpp
pgrete: In that case I'm in favor of cleaning up that code as part of this refactor.

src/mesh/mesh.cpp
Yurlungur: :+1: #994

src/application_input.hpp
jdolence: Yeah I don't think we ever implicitly assume a boundary condition, so as long as we add the code downstream to register the parthenon provided reflecting boundary conditions, I think this would be fine.

Yurlungur: :+1: Ok sounds good. I will remove the reflecting then and document how to register it.

pgrete:
@Yurlungur I missed bringing this PR up during the sync last week.
Is is reasonable to push this across the finish line before your leave? (I'm asking as we're about to add reflecting boundary conditions in AthenaPK and it'd be nice/cleaner to directly use the new interface).
It not, don't worry.

If I haven't missed anything the open items are "remove the reflecting then and document how to register it." and checking what's going wrong with the Swarm boundaries on device.

Yurlungur:
I'm sorry @pgrete I'm not sure I can get this done by the time I leave. I dug into the swarm failure some but I have no idea what's wrong with it. It might be best to merge this with a (eventual) refactor of swarm boundaries down the line.

pgrete:
@Yurlungur where does this PR stand?
I'm asking as I was about to fix our downstream PR for reflecting boundary conditions and then remembered that there was some Parthenon PR that would cleanup the required logic downstream.

Yurlungur:
> @Yurlungur where does this PR stand? I'm asking as I was about to fix our downstream PR for reflecting boundary conditions and then remembered that there was some Parthenon PR that would cleanup the required logic downstream.

Thanks to the cleanup to particles @brryan did I can probably revive this PR now. I haven't looked at it in some time, though. So I'd need a little bit of time to bring it up to develop.

src/bvals/bvals_interfaces.hpp
Yurlungur: Block BCs are now used by the forest of octrees machinery.

src/application_input.hpp
Yurlungur: Reflecting BCs no longer automatically registered.

Yurlungur:
Superseded by #1177

Issue #946, opened 2023-09-29
ADIOS I/O


BenWibking:
ORNL has optimized ADIOS2 for its systems and vice versa. In particular, Frontier is optimized for file-per-process I/O and ADIOS2 can do this with a single configuration variable change (with its BP5 backend: https://adios2.readthedocs.io/en/v2.9.1/engines/engines.html#bp5).

See also: https://www.exascaleproject.org/wp-content/uploads/2021/02/ADIOS_tutorial_ECP_AHM_Apr2021_full.pdf

With the BP5 backend, it is also significantly faster than either HDF5 and binary MPI I/O (e.g., https://www.scd.stfc.ac.uk/SiteAssets/Pages/CIUK-2021-Poster-Competition/POSTER_Shrey_Bhardwaj.pdf)

ADIOS also supports directly passing GPU buffers to its APIs, while HDF5 does not: https://adios2.readthedocs.io/en/v2.9.1/advanced/gpu_aware.html

Yurlungur:
How portable is ADIOS as far as data analysis tooling? Is it easy to install, e.g., a Python reader?

BenWibking:
I don't think it's as easy to install as `pip install h5py`, unfortunately. It does have a Spack package, and it seems to usually work.

On the other hand, it does come with its own Python API included: https://adios2.readthedocs.io/en/v2.9.1/api_high/api_high.html#python-high-level-api

Yurlungur:
Hmm my experience with spack for python packages is very poor. If you use spack for python, you better throw away your entire other python stack and use spack for everything. If you let things mix, it's painful.

 Though if the `setup.py build/install` process works ok it. might be fine.

BenWibking:
And Paraview and VisIt support with an XML metadata file (not sure if it's identical to the XDMF metadata format, but it looks very similar): https://adios2.readthedocs.io/en/v2.9.1/ecosystem/visualization.html

BenWibking:
> Hmm my experience with spack for python packages is very poor. If you use spack for python, you better throw away your entire other python stack and use spack for everything. If you let things mix, it's painful.
> 
> Though if the `setup.py build/install` process works ok it. might be fine.

Yeah, I've experienced this...

For a different can of worms, there's a Conda package: https://adios2.readthedocs.io/en/v2.9.1/setting_up/setting_up.html#conda

BenWibking:
[...]

Issue #752, opened 2022-10-06
Update output `final` logic


pgrete:
the `final` output names typically only make sense when the simulation stopped before reaching a limit set by the input file (such as `nlim` or `tlim`), e.g., by reaching a wallclock limit or being interrupted by a signal.
Thus, we could only use `final` in that case so that the last (intended) output automatically also gets a proper id.

c-prather:
I like this.  It was a handful of lines to put together a prototype and it feels like as a "breaking" change it should go in with the file format breaks for next release.

Should hitting nsteps be a "final" or numbered dump?  I'm inclined to say "final," since I find it much more likely that it's off-cadence with other dumps spaced evenly in simulation time.

@Yurlungur may have opinions. Not sure if other downstreams rely on the current behavior, I certainly don't.

pgrete:
> Should hitting nsteps be a "final" or numbered dump? I'm inclined to say "final," since I find it much more likely that it's off-cadence with other dumps spaced evenly in simulation time.

I'm also inclined to say final for the same reason.
The argument could also go along "outputs are determined by a `dt` so everything with this `dt` should get numbers, the last output triggered by anything else (nlim, wallclock, signal/file) should be `final`"
This logic, of course, may change if we allow output based on `dn` for which there are use cases (but I don't need them yet).

tagardiner:
Just my 2 cents - I've only ever used numbered outputs.  It may just be me, but when I choose a dt for output, I typically choose it to be an integer fraction of the simulation run time, i.e. (tend - tstart) = n dt for n an integer.  But, as most things go with simulations, "don't give shotguns to ducks" as John Hawley would say.  This also makes sense to me from the point of restarting a simulation to run longer.

I do still use outputs incremented on time-step from time to time when I want too see what's happening with the solution or refinement or ...  This is especially helpful when restarting around a problematic solution time.

Can we make this question of numbered final output an option?

brryan:
Using `final` only when the simulation is interrupted makes sense to me. We recently merged in #729 which makes the `final` labeling entirely optional (disabled by the user by setting `use_final_label = false` in any output block in the input parameter file). Maybe we could extend the `use_final_label` parameter to accept `always`, `never`, and `only_on_interrupt` options? My preference would be to make `only_on_interrupt` the default.

c-prather:
[...]

PR #990, opened 2023-12-16
Refactor I/O

Yurlungur:
## PR Summary

In a step towards getting the node/face/edge centered data to work, I am moving some of the generic I/O code out of HDF5 into utils. This should also help with issue #946 and getting ADIOS/IO to work. A main part of this effort is abstracting out those big gross nested for loops, which need to be done both at write time and read time, so that we can change things in only one place rather than have to synchronize them across files.

@bprather this is where I'm tweaking things.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


pgrete:
Let me know when you want comments here.

c-prather:
Linking #752 here.  It's fine if you don't want to tackle that in this PR, but given leave I'd like to try addressing that here when you're done with the indexing improvements.

Basically I'd like to move all the logic about output types and file names upward in the stack, and make them all use a common `MakeOutput(vector<string> variable_list, string filename)` call.  Then filenames can be more flexible, and new outputs are easier to define in downstream codes, not to mention implementing all the features in that issue becomes trivial.

src/coordinates/uniform_cartesian.hpp
```diff
@@ -164,18 +164,11 @@ class UniformCartesian {
...
   KOKKOS_FORCEINLINE_FUNCTION Real X(const int idx) const {
-    using TE = TopologicalElement;
-    [[maybe_unused]] bool constexpr X1EDGE =
-        el == TE::F1 || el == TE::E2 || el == TE::E3 || el == TE::NN;
-    [[maybe_unused]] bool constexpr X2EDGE =
-        el == TE::F2 || el == TE::E3 || el == TE::E1 || el == TE::NN;
-    [[maybe_unused]] bool constexpr X3EDGE =
-        el == TE::F3 || el == TE::E1 || el == TE::E2 || el == TE::NN;
-    if constexpr (dir == X1DIR && X1EDGE) {
+    if constexpr (dir == X1DIR && TopologicalOffsetI(el)) {
```
Yurlungur, line 167: This is just miscellaneous cleanup. Not needed in this PR but also a small savings that we might as well do.

src/mesh/mesh.cpp
```diff
@@ -1204,6 +1204,22 @@ int Mesh::GetNumberOfMeshBlockCells() const {
...
+  if (level == CellLevel::same) {
+    return pmb->cellbounds;
+    // TODO(JMM):
+    // } else if (level == CellLevel::fine) {
+    //   return pmb->fine_cellbounds;
+    // }
+  } else { // if (level == CellLevel::coarse) {
+    return pmb->c_cellbounds;
+  }
+}
```
Yurlungur, line 1219: cf @lroberts36

pgrete:
Is this ready for review/comments or is something (important) missing?
I'm asking as I'm working on the OpenPMD/ADIOS backend.

Yurlungur:
It's not 100% all the changes I wanted to include, but we could review it and get it in as a first pass so you could leverage the changes I've made here.

src/basic_types.hpp
```diff
@@ -49,7 +49,7 @@ enum class TaskStatus { fail, complete, incomplete, iterate, skip, waiting };
 
 enum class AmrTag : int { derefine = -1, same = 0, refine = 1 };
 enum class RefinementOp_t { Prolongation, Restriction, None };
-
+enum class CellLevel : int { coarse = -1, same = 0, fine = 1 };
```
pgrete, line 52: :+1:

src/coordinates/uniform_cartesian.hpp
pgrete: I think this makes it even clearer.

src/mesh/mesh.cpp
```diff
@@ -1204,6 +1204,22 @@ int Mesh::GetNumberOfMeshBlockCells() const {
 }
 const RegionSize &Mesh::GetBlockSize() const { return base_block_size; }
 
+const IndexShape &Mesh::GetLeafBlockCellBounds(CellLevel level) const {
+  // TODO(JMM): Luke this is for your Metadata::fine stuff.
+  PARTHENON_DEBUG_REQUIRE(level != CellLevel::fine,
+                          "Currently no access to finer cellbounds");
+  MeshBlock *pmb = block_list[0].get();
```
pgrete, line 1209: What are the implicit assumptions/implications here always using the first block?

src/mesh/mesh.hpp
```diff
@@ -194,6 +195,20 @@ class Mesh {
   std::vector<int> GetNbList() const noexcept { return nblist; }
   std::vector<LogicalLocation> GetLocList() const noexcept { return loclist; }
 
+  // TODO(JMM): Put in implementation file?
+  auto GetLevelsAndLogicalLocationsFlat() const noexcept {
+    std::vector<std::int64_t> levels, logicalLocations;
```
pgrete, line 200: How about making the `logicalLocations` a vector of an array of size 3?
In memory (so to pass, e.g., to hdf5) it'd be identical, but in other places in the code it might be more readable.

src/outputs/outputs.hpp
```diff
@@ -64,12 +64,13 @@ struct OutputParameters {
...
   bool sparse_seed_nans;
   int hdf5_compression_level;
+  bool write_xdmf;
   // TODO(felker): some of the parameters in this class are not initialized in constructor
   OutputParameters()
       : block_number(0), next_time(0.0), dt(-1.0), file_number(0),
         include_ghost_zones(false), cartesian_vector(false),
         single_precision_output(false), sparse_seed_nans(false),
-        hdf5_compression_level(5) {}
+        hdf5_compression_level(5), write_xdmf(true) {}
```
pgrete: Any specific reason to use `true`as default?
hst, histogram, Ascent all do not need/support xdmf.

src/parthenon_manager.cpp
```diff
@@ -336,6 +336,9 @@ void ParthenonManager::RestartPackages(Mesh &rm, RestartReader &resfile) {
 
       // Double note that this also needs to be update in case
       // we update the HDF5 infrastructure!
+      // JMM: Version -1 is the old version. We are now on version 3.
+      // TODO(JMM): Do we need to maintain backwards compatibility
+      // here? It would be ncie to remove this.
```
pgrete: I wouldn't mind to remove the old stuff (potentially in separate PR as it'd also pertain to the phdf reader) and mark is as deprecated.

src/outputs/output_utils.hpp
```diff
@@ -200,6 +203,59 @@ struct AllSwarmInfo {
...
+}
+
+template <typename Data_t, typename idx_t, typename Function_t>
+void PackOrUnpackVar(MeshBlock *pmb, Variable<Real> *pvar, bool do_ghosts, idx_t &idx,
+                     std::vector<Data_t> &data, Function_t f) {
+  auto v_h = pvar->data.GetHostMirrorAndCopy();
+  const auto &Nt = pvar->GetDim(6);
+  const auto &Nu = pvar->GetDim(5);
+  const auto &Nv = pvar->GetDim(4);
+  const IndexDomain theDomain = (do_ghosts ? IndexDomain::entire : IndexDomain::interior);
```
pgrete: Nitpicking style: `the_domain` or `domain`?

src/outputs/output_utils.hpp
```diff
@@ -200,6 +203,59 @@ struct AllSwarmInfo {
...
+  for (auto &pmb : pm->block_list) {
+    f(pmb.get(), data, i);
+  }
+  return data;
+}
+
+template <typename Data_t, typename idx_t, typename Function_t>
+void PackOrUnpackVar(MeshBlock *pmb, Variable<Real> *pvar, bool do_ghosts, idx_t &idx,
+                     std::vector<Data_t> &data, Function_t f) {
+  auto v_h = pvar->data.GetHostMirrorAndCopy();
```
pgrete: Just double checking (as host memory might actually be a constrained in some setups): Calling `GetHostMirrorAndCopy` from a `GetHostMirror` variable does *not* duplicate the memory allocation, correct?

src/outputs/parthenon_hdf5.cpp
```diff
@@ -718,66 +595,128 @@ std::string PHDF5Output::GenerateFilename_(ParameterInput *pin, SimTime *tm,
...
+  Kokkos::Profiling::pushRegion("I/O HDF5: write block metadata");
+  const H5G gBlocks = MakeGroup(file, "/Blocks");
+  const hsize_t num_blocks_local = pm->block_list.size();
+  const hsize_t ndim = pm->ndim;
+  const hsize_t loc_offset[2] = {offset, 0};
+
+  // write Xmin[ndim] for blocks
+  {
+    // JMM: These arrays chould be shared, but I think this is clearer
+    // as to what's going on.
```
pgrete, line 612: :+1: I stumbled across the original reuse pattern in the past...

src/outputs/outputs.cpp
```diff
@@ -263,6 +263,7 @@ Outputs::Outputs(Mesh *pm, ParameterInput *pin, SimTime *tm) {
           num_rst_outputs++;
         }
 #ifdef ENABLE_HDF5
+        op.write_xdmf = pin->GetOrAddBoolean(op.block_name, "write_xdmf", true);
```
pgrete, line 266: Add to the doc that his param now exists?

src/outputs/output_utils.cpp
```diff
@@ -100,6 +100,71 @@ AllSwarmInfo::AllSwarmInfo(BlockList_t &block_list,
...
+                                [=](MeshBlock *pmb, std::vector<Real> &data, int &i) {
+                                  auto xmin = pmb->coords.GetXmin();
+                                  data[i++] = xmin[0];
+                                  if (pm->ndim > 1) {
+                                    data[i++] = xmin[1];
+                                  }
+                                  if (pm->ndim > 2) {
+                                    data[i++] = xmin[2];
+                                  }
+                                });
```
pgrete: Is there a potential danger for misuse by accident here for outputs with ndim != 3 because the returned vector's size will always be `3*num_blocks_local` while there might only be valid data in one or two thirds.

pgrete (commented):
Overall I like these changes. Definitely make the easier to read and digest and allows for more reuse!

src/mesh/mesh.cpp
Yurlungur: That all blocks have the same shape, which is actually not necessarily true in the multigrid case, but in that case there are separate block lists, so this one still is safe for that assumption.

src/outputs/outputs.hpp
Yurlungur: backwards compatibility only. If we're fine breaking that, I'm fine making this default to false.

src/outputs/output_utils.cpp
Yurlungur: Hmm... yes there probably is. I believe this refactor maintains the behavior of the original, but let me think about edge cases here... Maybe the original code was a little dangerous.

src/mesh/mesh.hpp
Yurlungur: In that case, is there any danger of, e.g., padding? With `std::vector<std::array<std::int64_t, 3>>`, I can imagine the compiler padding the end of the array struct... So I don't think it's actually guranteed to be identical in memory. I could make a vector of pointers to construct a C-style array? But that would be what's already here but with extra steps.

Alternatively, I could make something that pulls out the logical locations and then something else that flattens?

src/outputs/outputs.hpp
Yurlungur: Ah, I see what you mean. I've set the default initialized value to `false`, but the default value from `GetOrAddBoolean` when called by `ENABLE_HDF5` to `true`.

src/parthenon_manager.cpp
Yurlungur: :+1: I don't see a way to add the `[[deprecated]]` atribute to arbitrary scope. I think it needs to be added to a function, variable or class definition. But I added a `PARTHENON_WARN`.

src/outputs/output_utils.hpp
Yurlungur: :+1: fixed

src/outputs/output_utils.hpp
Yurlungur: I'm not sure I understand the question---you mean in the case where the memory space and host space are the same right? `GetHostMirrorAndCopy` is a thin wrapper (with our data structures) around `Kokkos::create_mirror_view_and_copy`, which only allocates new memory if the source and target memory spaces are different.

Yurlungur: https://kokkos.github.io/kokkos-core-wiki/API/core/view/create_mirror.html?highlight=create_mirror_view_and_copy#_CPPv4I0DpE27create_mirror_view_and_copy14ImplMirrorTypeRKN4Impl12ViewCtorPropIDp12ViewCtorArgsEERK8ViewType

src/outputs/outputs.cpp
Yurlungur: :+1: done

src/outputs/output_utils.cpp
Yurlungur: I think there's no danger in changing `3` to `ndim` here so I'm just going to make that change.

Yurlungur:
@bprather since @pgrete  wants to re-use this for his I/O stuff, let's get this first pass in now. Do you mind going through and reviewing?

@pgrete I *think* I've addressed all your comments. Please take another look.

src/mesh/mesh.cpp
Yurlungur: Also that's why this is called `GetLeafBlockCellBounds`.

src/outputs/output_utils.hpp
```diff
@@ -200,6 +203,59 @@ struct AllSwarmInfo {
...
+  std::vector<T> data(shape * num_blocks_local);
+  int i = 0;
+  for (auto &pmb : pm->block_list) {
+    f(pmb.get(), data, i);
+  }
+  return data;
+}
+
+template <typename Data_t, typename idx_t, typename Function_t>
+void PackOrUnpackVar(MeshBlock *pmb, Variable<Real> *pvar, bool do_ghosts, idx_t &idx,
```
Yurlungur, line 219: Worth noting, @pgrete this is the one function I am likely to change down the road, if you wanted to use it. I plan to have this machinery return some shape information, and also permute it to support cells/edges/etc.

src/mesh/mesh.cpp
pgrete: Gotcha, I blanked on the cellbounds just returning indices and nothing coordinate related.

src/mesh/mesh.hpp
pgrete: Why would there potentially be padding? And if there's padding this would not be impacted when passing the data pointer with the corresponding `.size()` call, does it?

src/outputs/output_utils.hpp
pgrete: No worries.
I'll keep an eye on this.
Also I might later also modify this function in order to allow for coarsening of the data (so overall I'm very much in favor of having this abstraction/functionality).

src/outputs/output_utils.hpp
pgrete: That's what I thought, which is why I was double checking (as currently there's another `GetHostMirror` right before calling this function is being called).

pgrete (approved):
Thanks for the updates. LGTM!

src/mesh/mesh.hpp
Yurlungur: Maybe I'm not understanding what you want to do---My understanding is that you want `std::vector<std::array<std::int64_t, 3>>` right? But is `std::array<T, N>` not a struct? The C++ standard specifies that structs can have arbitrary padding, so there may be extra bits between `&vec[0]` and `&vec[1]`, not just `3*64`. Or am I misunderstanding something?

c-prather (approved):
LGTM this is a great start making this code more manageable.

Be warned I may use `PackOrUnpackVar` downstream, it's too convenient

pgrete:
I'm not sure why the test failed. I retriggered the test to see if it's reproducible

Yurlungur:
> I'm not sure why the test failed. I retriggered the test to see if it's reproducible

Seems to reproduce. I'll see if I can reproduce on my machine. Looks to be the restart test. Could have sworn I had this working on my machine, but... must have broken something in a recent commit.

Yurlungur:
> > I'm not sure why the test failed. I retriggered the test to see if it's reproducible
> 
> Seems to reproduce. I'll see if I can reproduce on my machine. Looks to be the restart test. Could have sworn I had this working on my machine, but... must have broken something in a recent commit.

Hmm ok I'm confused, because tests pass on my machine. I'll try to spin up a GPU machine and see if the failure is Cuda specific.

Yurlungur:
CI failure located and fixed I believe. A silly bug in my pack or unpack variable

PR #991, opened 2023-12-19
[Breaking change] Add fine field option

lroberts36:
## PR Summary

This PR adds the new flag `Metadata::Fine`, which indicates that the field should have twice the resolution of default fields. Fields with this flag can exchange ghosts, and be prolongated and restricted similarly to normal fields. They can also be output, although none of the plotting scripts have been updated to deal with fine fields. This feature is in use in Assail.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


example/advection/advection_package.cpp
```diff
@@ -167,6 +167,10 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
                  std::vector<int>({vec_size}), advected_labels);
     pkg->AddField(field_name, m);
   }
+  // TODO(LFR): Remove this
+  Metadata m_fine(
+      {Metadata::Cell, Metadata::Independent, Metadata::Fine, Metadata::FillGhost});
+  pkg->AddField("advected_fine", m_fine);
```
Yurlungur: Not sure we should remove this actually... Unless we add a separate test somewhere?

example/advection/advection_package.cpp
```diff
@@ -443,6 +447,68 @@ Real EstimateTimestepBlock(MeshBlockData<Real> *rc) {
...
+          v(b, out, fk + 1, fj, fi + 1) = v(b, in, k, j, i);
+          v(b, out, fk + 1, fj + 1, fi) = v(b, in, k, j, i);
+          v(b, out, fk + 1, fj + 1, fi + 1) = v(b, in, k, j, i);
+        }
+      });
+  return TaskStatus::complete;
+}
+
+// TODO(LFR): Remove this
+TaskStatus PrintFine(MeshData<Real> *md) {
```
Yurlungur: This one we can probably due with hdf5 I/O maybe...

Yurlungur: Perhaps with the help of #990

Yurlungur (approved):
Well this was remarkably simple. I was expecting it to be a little more complicated than this but nice to see it worked out so easily.

example/advection/advection_package.cpp
lroberts36: Yeah, I didn't do anything about the I/O so I doubt the fine fields would work.

example/advection/advection_package.cpp
lroberts36: I just thought we should maybe take it out since there is a test called advection performance and this probably has some impact on performance. Not sure if anyone uses that as an actual performance test though... 

Looking at this makes me want to re-write the example though.

lroberts36:
> Well this was remarkably simple. I was expecting it to be a little more complicated than this but nice to see it worked out so easily.

@Yurlungur: Well, I have only tested for unigrid so not totally sure it works for prolongation and restriction. But at least it runs with AMR w/o segfaults.

Yurlungur:
> > Well this was remarkably simple. I was expecting it to be a little more complicated than this but nice to see it worked out so easily.
> 
> @Yurlungur: Well, I have only tested for unigrid so not totally sure it works for prolongation and restriction. But at least it runs with AMR w/o segfaults.

I *suppose* we should check for correctness with AMR lol.

example/advection/advection_package.cpp
Yurlungur: Yeah that's fair. It's frustrating that we don't have a way to test things like this outside of regression testing. And yes agreed. It needs a rewrite.

example/advection/advection_package.cpp
lroberts36: I/O should now work, so this task is removed.

example/advection/advection_package.cpp
lroberts36: leaving this in here now

Yurlungur (approved):

doc/sphinx/src/interface/metadata.rst
```diff
@@ -55,6 +55,9 @@ mutually exclusive:
    cell-corners. The variable might be volume-averaged, or defined
    pointwise.
 
+For any type of topology, the flag ``Metadata::Fine`` can be set which 
+causes the field will have twice the resolution of normal fields.
```
adamdempsey90:
```suggestion
causes the field to have twice the resolution of normal fields.
```

adamdempsey90 (approved):
Downstream app looks good

doc/sphinx/src/interface/metadata.rst
```diff
@@ -55,6 +55,9 @@ mutually exclusive:
    cell-corners. The variable might be volume-averaged, or defined
    pointwise.
 
+For any type of topology, the flag ``Metadata::Fine`` can be set which 
+causes the field to have twice the resolution of normal fields.
+
```
pgrete: It'd be great to document what's currently tested/supported and what's "handle with care and expect sth go (silently) wrong"

example/advection/advection_driver.cpp
```diff
@@ -136,6 +136,9 @@ TaskCollection AdvectionDriver::MakeTaskCollection(BlockList_t &blocks, const in
     auto update = tl.AddTask(avg_data, UpdateIndependentData<MeshData<Real>>, mc0.get(),
                              mdudt.get(), beta * dt, mc1.get());
 
+    // TODO(LFR): Remove this
+    update = tl.AddTask(update, advection_package::FillFine, mc1.get());
```
pgrete: still to be removed?

example/advection/advection_package.cpp
```diff
@@ -493,6 +496,44 @@ Real EstimateTimestepBlock(MeshBlockData<Real> *rc) {
   return cfl * min_dt;
 }
 
+// TODO(LFR): Remove this
```
pgrete: todo?

src/bvals/comms/bnd_info.cpp
```diff
@@ -109,20 +109,23 @@ SpatiallyMaskedIndexer6D CalcIndices(const NeighborBlock &nb, MeshBlock *pmb,
...
-  auto shape = pmb->cellbounds;
+  bool is_fine_field = v->IsSet(Metadata::Fine);
+  auto shape = is_fine_field ? pmb->f_cellbounds : pmb->cellbounds;
   // Both prolongation and restriction always operate in the coarse
   // index space. Also need to use the coarse index space if the
   // neighbor is coarser than you, wether or not you are setting
   // interior or exterior cells
-  if (prores || nb.loc.level() < loc.level()) shape = pmb->c_cellbounds;
+  if (prores || nb.loc.level() < loc.level())
+    shape = is_fine_field ? pmb->cellbounds : pmb->c_cellbounds;
```
pgrete, line 119: smart

src/outputs/output_utils.cpp
```diff
@@ -125,10 +125,11 @@ int VarInfo::GetDim(int i) const {
 }
 
 std::vector<VarInfo> VarInfo::GetAll(const VariableVector<Real> &vars,
-                                     const IndexShape &cellbounds) {
+                                     const IndexShape &cellbounds,
+                                     const IndexShape &f_cellbounds) {
   std::vector<VarInfo> out;
   for (const auto &v : vars) {
-    out.emplace_back(v, cellbounds);
+    out.emplace_back(v, v->IsSet(Metadata::Fine) ? f_cellbounds : cellbounds);
```
pgrete, line 133: here and below: Where is the info on `f_cellbounds` in the context of `VarInfo` actually being used, i.e., why is this change necessary?

src/mesh/meshblock.cpp
```diff
@@ -204,6 +204,7 @@ MeshBlock::~MeshBlock() = default;
 void MeshBlock::InitializeIndexShapesImpl(const int nx1, const int nx2, const int nx3,
                                           bool init_coarse, bool multilevel) {
   cellbounds = IndexShape(nx3, nx2, nx1, Globals::nghost);
+  f_cellbounds = IndexShape(2 * nx3, 2 * nx2, 2 * nx1, Globals::nghost);
```
pgrete, line 207: Does this translate to a fine block having the same number of ghost zones than a standard block?
If so, is that desired behavior as this might result in weird indexing wrt to fine ghost data (and similarly in the context of prolongation/restriction).

pgrete (commented):
Just did a quick review. I understand where this feature might be useful and just have a couple of comments regarding the current implementation.

src/mesh/meshblock.cpp
lroberts36: Yes. I went with this implementation since it was analogous to the difference between regular data and course data for a field. That way for a fine field, the course data should have the same shape as regular data and I think prolongation and restriction should work out of the box (i.e. I need `f_cellbounds` is to `cellbounds` as `cellbounds` is to `c_cellbounds` for things to "just work out").

src/outputs/output_utils.cpp
lroberts36: I think this eventually gets used to set the size of the datasets that are output to the hdf5 file, so if I try to output a `Fine` field without this change it would only write $N^3$ points instead of $(2N)^3$ points.

example/advection/advection_driver.cpp
lroberts36: Actually, I think just the comment should be removed since this at least tests fine fields a little bit. That being said, including the fine field slows down the advection tests some since the amount of output is substantially increased. I just removed the comment for now.

example/advection/advection_package.cpp
lroberts36: Removed comment.

doc/sphinx/src/interface/metadata.rst
lroberts36: docs updated with user beware language.

lroberts36:
This is going back into WIP for now since there is some issue with restarting `Fine` fields going on. Otherwise, I think the last round of changes should make fine fields work with AMR and boundary conditions.

lroberts36:
Ok, I have now added an example that does advection with fine fields and a test that uses this that checks that restarts work. I didn't include any gold files to check the fine advection against, but the results look very reasonable. I think at this point most of the kinks with fine fields should be worked out. 

As you might notice, there is some overkill in the `fine_advection` example (in particular the Stokes update stuff). My general thinking was that this would be a good starting point to include some sort of test that evolves non-cell centered fields and also could be updated to replace the `sparse_advection` example (which is kind of outdated and a mess).

example/fine_advection/advection_driver.cpp
```diff
@@ -0,0 +1,165 @@
...
+}
+
+// See the advection.hpp declaration for a description of how this function gets called.
+TaskCollection AdvectionDriver::MakeTaskCollection(BlockList_t &blocks, const int stage) {
+  using namespace parthenon::Update;
+  TaskCollection tc;
+  TaskID none(0);
+
+  // Build MeshBlockData containers that will be included in MeshData containers. It is
+  // gross that this has to be done by hand.
```
Yurlungur, line 63: could this be an opportunity to fix this? Maybe we could hide this in something that the default evolution driver just "does"

example/fine_advection/advection_driver.cpp
```diff
@@ -0,0 +1,165 @@
...
+        parthenon::MakePackDescriptor<advection_package::Conserved::scalar>(
+            pmesh->resolved_packages.get(), {parthenon::Metadata::WithFluxes},
+            {parthenon::PDOpt::WithFluxes});
+    using pack_desc_t = decltype(desc);
+
+    static auto desc_fine =
+        parthenon::MakePackDescriptor<advection_package::Conserved::scalar_fine>(
+            pmesh->resolved_packages.get(), {parthenon::Metadata::WithFluxes},
+            {parthenon::PDOpt::WithFluxes});
+    using pack_desc_fine_t = decltype(desc_fine);
```
Yurlungur, line 106: Is this just to get the types?

example/fine_advection/advection_driver.cpp
```diff
@@ -0,0 +1,165 @@
...
+    if (pmesh->ndim > 2) faces.push_back(TE::F3);
+    auto flx = none;
+    auto flx_fine = none;
+    for (auto face : faces) {
+      flx = flx | tl.AddTask(none, advection_package::CalculateFluxes<pack_desc_t>, desc,
+                             face, parthenon::CellLevel::same, mc0.get());
+      flx_fine = flx_fine |
+                 tl.AddTask(none, advection_package::CalculateFluxes<pack_desc_fine_t>,
+                            desc_fine, face, parthenon::CellLevel::fine, mc0.get());
+    }
```
Yurlungur, line 120: Ah I see. Cute. Would it not be better maybe though to just use the var names for the template params rather than the pack descriptor type? I guess it doesn't matter...

example/fine_advection/advection_package.hpp
```diff
@@ -0,0 +1,86 @@
...
+    if (v > 0) ioff = -1;
+  } else if (FACE == TE::F2) {
+    v = pkg->Param<Real>("vy");
+    if (v > 0) joff = -1;
+  } else if (FACE == TE::F3) {
+    v = pkg->Param<Real>("vz");
+    if (v > 0) koff = -1;
+  }
+
+  auto pack = desc.GetPack(md);
```
Yurlungur, line 68: Oh I see... you do use the pack descriptors defined in the driver. I don't know how I feel about that... I guess it's fine. It feels a little different than "canonical" parthenon usage though.

example/fine_advection/parthenon_app_inputs.cpp
```diff
@@ -0,0 +1,119 @@
...
+  if (profile == "wave") profile_type = 0;
+  if (profile == "smooth_gaussian") profile_type = 1;
+  if (profile == "hard_sphere") profile_type = 2;
+  if (profile == "block") profile_type = 3;
+  Real amp = 1.0;
+
+  const int b = 0;
+  const int ndim = pmb->pmy_mesh->ndim;
+  const int nghost = parthenon::Globals::nghost;
+  pmb->par_for(
```
Yurlungur, line 67: should we avoid using `pmb->` now that overloads exit that don't need the loop pattern and exec space specified? I feel like `pmb->par_for` should eventually go away.

example/fine_advection/advection_package.cpp
```diff
@@ -0,0 +1,180 @@
...
+      Metadata({Metadata::Cell, Metadata::Fine, Metadata::Independent,
+                Metadata::WithFluxes, Metadata::FillGhost}));
+
+  pkg->AddField<Conserved::scalar>(Metadata({Metadata::Cell, Metadata::Independent,
+                                             Metadata::WithFluxes, Metadata::FillGhost}));
+  pkg->AddField<Conserved::scalar_fine_restricted>(
+      Metadata({Metadata::Cell, Metadata::Derived, Metadata::OneCopy}));
+  pkg->CheckRefinementBlock = CheckRefinement;
+  pkg->EstimateTimestepMesh = EstimateTimestep;
+  pkg->FillDerivedMesh = RestrictScalarFine;
```
Yurlungur, line 78: if we want to eventually replace sparse advection driver should one of these be a sparse pool? Or were you thinking of replacing dense advection? Maybe it should all be unified in a single new advection example...

example/fine_advection/stokes.hpp
```diff
@@ -0,0 +1,150 @@
...
+
+  parthenon::par_for(
+      PARTHENON_AUTO_LABEL, 0, pack1.GetNBlocks() - 1, pack1.GetLowerBoundHost(0),
+      pack1.GetUpperBoundHost(0), // This is safe for dense vars
+      kb.s, kb.e, jb.s, jb.e, ib.s, ib.e,
+      KOKKOS_LAMBDA(const int b, const int l, const int k, const int j, const int i) {
+        pack_out(b, l, k, j, i) = w1 * pack1(b, l, k, j, i) + w2 * pack2(b, l, k, j, i);
+      });
+  return TaskStatus::complete;
+}
```
Yurlungur: Should this be in `interface/update`?

example/fine_advection/stokes.hpp
```diff
@@ -0,0 +1,150 @@
...
+  IndexRange kb = out->GetBoundsK(cl, IndexDomain::interior, TeVar);
+
+  parthenon::par_for(
+      PARTHENON_AUTO_LABEL, 0, pack_out.GetNBlocks() - 1, pack_out.GetLowerBoundHost(0),
+      pack_out.GetUpperBoundHost(0), // This is safe for dense vars only
+      kb.s, kb.e, jb.s, jb.e, ib.s, ib.e,
+      KOKKOS_LAMBDA(const int b, const int l, const int k, const int j, const int i) {
+        pack_out(b, TeVar, l, k, j, i) = 0.0;
+      });
+}
```
Yurlungur, line 72: should this be made general and put in `interface/update`?

src/coordinates/uniform_cartesian.hpp
```diff
@@ -269,6 +269,52 @@ class UniformCartesian {
...
+        return area_[X3DIR - 1];
+      } else if (el == TE::E1) {
+        return dx_[X1DIR - 1];
+      } else if (el == TE::E2) {
+        return dx_[X2DIR - 1];
+      } else if (el == TE::E3) {
+        return dx_[X3DIR - 1];
+      } else if (el == TE::NN) {
+        return 1.0;
+      }
```
Yurlungur, line 293: Is this all new or can we call the older volume method which defaults to `CellLevel::same`?

src/interface/sparse_pack.hpp
```diff
@@ -333,6 +333,19 @@ class SparsePack : public SparsePackBase {
...
+                                    Args &&...args) const {
+    const int dir = (static_cast<int>(te) % 3) + 1;
+    return flux(b, dir, std::forward<Args>(args)...);
+  }
+
+  template <class... Args>
+  KOKKOS_INLINE_FUNCTION auto &flux(const TopologicalElement te, Args &&...args) const {
+    const int dir = (static_cast<int>(te) % 3) + 1;
+    return flux(dir, std::forward<Args>(args)...);
+  }
```
Yurlungur, line 347: Did these not already exist? What's different here?

src/mesh/mesh.cpp
```diff
@@ -1255,16 +1255,11 @@ int Mesh::GetNumberOfMeshBlockCells() const {
 }
 
 const IndexShape &Mesh::GetLeafBlockCellBounds(CellLevel level) const {
-  // TODO(JMM): Luke this is for your Metadata::fine stuff.
-  PARTHENON_DEBUG_REQUIRE(level != CellLevel::fine,
-                          "Currently no access to finer cellbounds");
```
Yurlungur, line 901: :+1:

Yurlungur (approved):
I like the new example!

example/fine_advection/advection_driver.cpp
lroberts36: I didn't think about it too hard, but it seems like it should not be hard to have `MeshData::GetOrAdd` reflexively call `MeshBlockData::Add` (or something) if the `MeshBlockData` containers don't exist. I guess there must be some subtle issue I am missing?

Yurlungur: I think it should be easy... we just never did it

example/fine_advection/advection_driver.cpp
lroberts36: No, `desc` and `desc_fine` get passed to the tasks and then get used inside the tasks to get packs. I think this allows maximum flexibility for update tasks and makes it clear what is intended for those routines to do in the driver. Obviously this is a different pattern than we have been using, but I kind of like it.

example/fine_advection/advection_driver.cpp
lroberts36: I think this allows a lot more flexibility than just passing var types, for instance you can select on Metadata without having to change anything about the internals of the task or you can use any `MakePackDescriptor` that you choose to write.

example/fine_advection/advection_package.hpp
lroberts36: see above

example/fine_advection/parthenon_app_inputs.cpp
lroberts36: oh, yeah. I copied the old advection example and updated it. I thought I removed all `pmb->par_for`s but apparently not.

example/fine_advection/advection_package.cpp
lroberts36: When I replace the sparse advection test I was going to add some sparse variables here. I also would need to update all the loops to use hierarchical parallelism since they are now written to go over the number of variables in the pack on the first block. This would be a small change though. 

I think it is probably better not to remove the original advection example, since it uses some history stuff and has a number of tests that rely on it. Maybe it should be moved from the example directory to the tst directory though.

example/fine_advection/stokes.hpp
lroberts36: Eventually, but I think for now we should leave it in the example. I was trying to write a parallel set of functionality to `update` here that works with more general fields.

src/coordinates/uniform_cartesian.hpp
lroberts36: the old `Volume` method is templated on the topological element. Here I added a new `Volume` overload that is not templated and takes a topological type and a cell level as an argument. It should give the same results as the templated volume function.

src/interface/sparse_pack.hpp
lroberts36: That they take a `TopologicalElement` argument instead of a `CoordinateDirection`. This exists for `operator()` but not for `SparsePack::flux(...)`

example/fine_advection/stokes.hpp
lroberts36: eventually, see above.

example/fine_advection/advection_package.cpp
lroberts36: I also think we may want to add the vector wave equation to this test.

example/fine_advection/advection_package.hpp
Yurlungur: yeah I don't feel strongly about it now that I understand what's going on.

example/fine_advection/advection_package.cpp
Yurlungur: I like the idea of moving the old advection example to the test directory.

src/coordinates/uniform_cartesian.hpp
Yurlungur: What I meant was that for the `CellLevel::same` branch can you call the old method?

lroberts36: You could, but since the old one is templated on TE and this one is not it doesn't save any lines of code since you would have to call each template by hand. I am not actually sure that this really needs to be templates though.

Yurlungur: Oh I see. Could the old one optionally take a runtime? That might be useful...

lroberts36: Yeah, I think the best solution might be to have the old one just pass the template argument as a regular argument to the new one and have the new one have a default argument of cell level same. I doubt that has any performance impact.

example/fine_advection/advection_driver.cpp
lroberts36: This is fixed in #1092

example/fine_advection/parthenon_app_inputs.cpp
lroberts36: this is fixed in a subsequent PR

example/fine_advection/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
+#=========================================================================================
+# (C) (or copyright) 2020. Triad National Security, LLC. All rights reserved.
```
pgrete: This passed the copyright check?

example/fine_advection/advection_driver.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+  for (int i = 0; i < num_partitions; i++) {
+    auto &tl = single_tasklist_per_pack_region[i];
+    auto &mbase = pmesh->mesh_data.GetOrAdd("base", i);
+    auto &mc0 = pmesh->mesh_data.GetOrAdd(stage_name[stage - 1], i);
+    auto &mc1 = pmesh->mesh_data.GetOrAdd(stage_name[stage], i);
+    auto &mdudt = pmesh->mesh_data.GetOrAdd("dUdt", i);
+
+    auto start_send = tl.AddTask(none, parthenon::StartReceiveBoundaryBuffers, mc1);
+    auto start_flxcor = tl.AddTask(none, parthenon::StartReceiveFluxCorrections, mc0);
```
pgrete, line 90: Independent of this PR but relevant in the context of recent discussion around exhausting MPI limits, it might be worth to change the pattern here to first start all receives (in their own `TaskRegion`) rather than starting them "per pack" as the behavior will then depend on the pack size.

example/fine_advection/advection_driver.cpp
pgrete: Given that this is an "example" to be potentially read by new users, I suggest to add one or two lines of comments above to motivate this already here in the code.

example/fine_advection/parthinput.advection
```diff
@@ -0,0 +1,71 @@
+# ========================================================================================
+#  Athena++ astrophysical MHD code
+#  Copyright(C) 2014 James M. Stone <jmstone@princeton.edu> and other code contributors
+#  Licensed under the 3-clause BSD License, see LICENSE file for details
```
pgrete: outdated

example/fine_advection/advection_driver.cpp
pgrete: And maybe even more general, I'd be great to add a (very) short README to the directory explaining what the example does and what it highlights (from a feature point of view).
I know we didn't do this for other examples yet, but I think it's useful in the long run.

example/fine_advection/advection_package.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+// *************************************************//
+// define the "physics" package Advect, which      *//
+// includes defining various functions that control*//
+// how parthenon functions and any tasks needed to *//
+// implement the "physics"                         *//
+// *************************************************//
+
+namespace advection_package {
+using parthenon::UserHistoryOperation;
```
pgrete: unused

example/fine_advection/stokes.hpp
```diff
@@ -0,0 +1,150 @@
...
+  auto pack2 = pd.GetPack(in2);
+  auto pack_out = pd.GetPack(out);
+
+  IndexRange ib = in1->GetBoundsI(cl, IndexDomain::entire, te);
+  IndexRange jb = in1->GetBoundsJ(cl, IndexDomain::entire, te);
+  IndexRange kb = in1->GetBoundsK(cl, IndexDomain::entire, te);
+
+  parthenon::par_for(
+      PARTHENON_AUTO_LABEL, 0, pack1.GetNBlocks() - 1, pack1.GetLowerBoundHost(0),
+      pack1.GetUpperBoundHost(0), // This is safe for dense vars
```
pgrete, line 48: should there be a safety check that all pack indices are aligned (or at least over the same range)?

example/fine_advection/stokes.hpp
```diff
@@ -0,0 +1,150 @@
...
+  PARTHENON_REQUIRE(ioff == 1 || ioff == 0, "Bad combination of TeVar and TeFlux");
+  PARTHENON_REQUIRE(joff == 1 || joff == 0, "Bad combination of TeVar and TeFlux");
+  PARTHENON_REQUIRE(koff == 1 || koff == 0, "Bad combination of TeVar and TeFlux");
+  PARTHENON_REQUIRE((ioff + joff + koff) == 1, "Bad combination of TeVar and TeFlux");
+  koff = ndim > 2 ? koff : 0;
+  joff = ndim > 1 ? joff : 0;
+
+  parthenon::par_for(
+      PARTHENON_AUTO_LABEL, 0, pack_in.GetNBlocks() - 1, pack_in.GetLowerBoundHost(0),
+      pack_in.GetUpperBoundHost(0), // This is safe for dense vars only
```
pgrete, line 101: as above, bounds check in `l` for in and out packs

src/bvals/boundary_conditions_generic.hpp
```diff
@@ -138,7 +138,7 @@ void GenericSwarmBC(std::shared_ptr<Swarm> &swarm) {
 }
 
 namespace impl {
-using desc_key_t = std::tuple<bool, TopologicalType>;
+using desc_key_t = std::tuple<bool, bool, TopologicalType>;
```
pgrete, line 141: Not suggesting for this PR, but if this gets further expanded it might be worth to introduce enum classes or structs to be passed as parameter (to reduce a potential `{true, false, true, ...}` confusion)

src/bvals/boundary_conditions_generic.hpp
```diff
@@ -179,70 +184,74 @@ void GenericBC(std::shared_ptr<MeshBlockData<Real>> &rc, bool coarse,
...
+          } else if (TYPE == BCType::FixedFace) {
+            q(b, el, l, k, j, i) =
+                2.0 * val - q(b, el, l, X3 ? offset - k : k, X2 ? offset - j : j,
+                              X1 ? offset - i : i);
+          } else if (TYPE == BCType::ConstantDeriv) {
+            Real dq = q(b, el, l, X3 ? ref + offsetin : k, X2 ? ref + offsetin : j,
+                        X1 ? ref + offsetin : i) -
+                      q(b, el, l, X3 ? ref - offsetout : k, X2 ? ref - offsetout : j,
+                        X1 ? ref - offsetout : i);
+            Real delta = 0.0;
```
pgrete, line 238: Given the complexity of this code (and above in this file) is it worth to update the new test case for use of non-periodic boundary conditions?

src/coordinates/uniform_cartesian.hpp
```diff
@@ -269,6 +269,52 @@ class UniformCartesian {
...
+    } else if (cl == CellLevel::fine) {
+      if (el == TE::CC) {
+        return cell_volume_ / 8.0;
```
pgrete, line 296: Is this generally true (i.e., design decision) for <3D sims?
I don't recall how we handle this right now.

tst/regression/test_suites/restart_fine/parthinput.restart_fine
```diff
@@ -0,0 +1,66 @@
+# ========================================================================================
+#  Athena++ astrophysical MHD code
+#  Copyright(C) 2014 James M. Stone <jmstone@princeton.edu> and other code contributors
+#  Licensed under the 3-clause BSD License, see LICENSE file for details
```
pgrete: -> Parthenon

pgrete (approved):
LGTM 
Thanks for putting the extra effort in to make it (almost) feature complete.
I just got a few minor comments.

example/fine_advection/parthinput.advection
lroberts36: You think I should just remove the old copyright? probably so since this is different code.

example/fine_advection/stokes.hpp
lroberts36: The changes to make `stokes.hpp` work for sparse variables in #1103 should fix this.

lroberts36: oh, yeah, I can add a test that the two packs have the same bounds as well.

example/fine_advection/stokes.hpp
lroberts36: see above, will fix in #1103

src/bvals/boundary_conditions_generic.hpp
lroberts36: yes, I think we want test coverage for all types of boundary conditions, different forest configurations, sparse and non-sparse, and regular and fine fields. Once the rest of these PRs that rely on this one are in, I was going to make another PR adding regression tests, add gold files, and maybe remove the old sparse advection test.

src/coordinates/uniform_cartesian.hpp
lroberts36: I think that how we deal with, e.g., 2D simulations is that they are actually 3D (so the volume of a cell is a 3-volume) but we just assume they have translational symmetry in the z-direction. In that case, I think dividing by 8 here for all `ndim` is the right thing to do.

example/fine_advection/advection_driver.cpp
lroberts36: Yeah, that seems reasonable.

src/bvals/boundary_conditions_generic.hpp
lroberts36: Agreed, but I also agree we should leave for a future PR.

example/fine_advection/CMakeLists.txt
lroberts36: not sure, but it is now fixed

example/fine_advection/advection_driver.cpp
lroberts36: added the comment and a brief README.md

example/fine_advection/parthinput.advection
lroberts36: removed

example/fine_advection/advection_package.cpp
lroberts36: removed

example/fine_advection/stokes.hpp
lroberts36: added bounds check

example/fine_advection/stokes.hpp
lroberts36: added bounds check

BenWibking:
This broke the existing use of `pmb->par_for_bndry(...)` in AthenaPK. It looks like this was missed from the "breaking changes" section of the changelog.

If we are not using fine fields, should the `fine` argument of `par_for_bndry` always be `false`?

lroberts36:
@BenWibking: Yes, the `fine` argument should just be set to false. Sorry, I included the `breaks-downstream` label but I guess forgot to put this in the correct section of changelog.

PR #992, opened 2023-12-19
Allow custom PR ops with sparse pools

pdmullen:
## PR Summary
`AddSparsePool` copies the relevant `Metadata` for all entries in the pool.  Calling the `Metadata` constructor upon copying (unintentionally) overwrites any custom P/R operations that were previously enrolled via `m.RegisterRefinementOps`.

My changeset adds the default PR ops as an optional argument to the `Metadata` constructor.  When `IsRefined()`, this default argument is used to set the `refinement_funcs_` private member.  

With this change, when copying `Metadata` for all entries in the `SparsePool`, we can just pass `shared_metadata.GetRefinementFunctions()` to the argument list, resulting in the desired behavior.  

My changeset feels somewhat hacky, and invokes unnecessary `RegisterOps` calls due to the default argument.  Please be harsh... if anything, let this PR bring attention to the issue and a possible resolution, even if we eventually take a wildly different approach to solving it.  

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/src/interface/metadata.hpp
+++ b/src/interface/metadata.hpp
@@ -316,9 +316,13 @@ class Metadata {
 
   // 4 constructors, this is the general constructor called by all other constructors, so
   // we do some sanity checks here
-  Metadata(const std::vector<MetadataFlag> &bits, const std::vector<int> &shape = {},
-           const std::vector<std::string> &component_labels = {},
-           const std::string &associated = "");
+  Metadata(
+      const std::vector<MetadataFlag> &bits, const std::vector<int> &shape = {},
+      const std::vector<std::string> &component_labels = {},
+      const std::string &associated = "",
+      const refinement::RefinementFunctions_t ref_funcs_ =
+          refinement::RefinementFunctions_t::RegisterOps<
+              refinement_ops::ProlongateSharedMinMod, refinement_ops::RestrictAverage>());
 
   // 1 constructor
   Metadata(const std::vector<MetadataFlag> &bits, const std::vector<int> &shape,
@@ -548,7 +552,6 @@ class Metadata {
 
   // Refinement stuff
   const refinement::RefinementFunctions_t &GetRefinementFunctions() const {
-    PARTHENON_REQUIRE_THROWS(IsRefined(), "Variable must be registered for refinement");
     return refinement_funcs_;
   }
   template <class ProlongationOp, class RestrictionOp,
--- a/tst/unit/test_state_descriptor.cpp
+++ b/tst/unit/test_state_descriptor.cpp
@@ -383,7 +383,7 @@ TEST_CASE("Test dependency resolution in StateDescriptor", "[StateDescriptor]") 
       }
     }
 
-    WHEN("We register a dense variable custom prolongation/restriction") {
+    WHEN("We register a dense variable with default prolongation/restriction") {
       pkg1->AddField("dense", m_provides);
       WHEN("We register a sparse variable with custom prolongation/restriction") {
         auto m_sparse_provides_ = m_sparse_provides;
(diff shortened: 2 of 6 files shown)


src/interface/metadata.hpp
```diff
@@ -548,7 +552,6 @@ class Metadata {
 
   // Refinement stuff
   const refinement::RefinementFunctions_t &GetRefinementFunctions() const {
-    PARTHENON_REQUIRE_THROWS(IsRefined(), "Variable must be registered for refinement");
```
Yurlungur, line 551: What is the reason for this change?

Yurlungur (changes requested):
Good catch, thanks for finding this! Big picture, I wonder if a more maintainable solution is to add a copy constructor and assignment operator... probably enough to do
```C++
Metadata::Metadata(const Metadata &other) = default;
friend Metadata &operator=(const Metadata &other) = default;
```
and then use assignment when copying in the sparse pool?

We could also add a test in the test suite that the copy constructor works, and in sparse pool that custom prolong/restrict works?

src/interface/metadata.hpp
pdmullen: We are calling `GetRefinementFunctions()` here: 

https://github.com/parthenon-hpc-lab/parthenon/pull/992/files#diff-e142fce743a6bbfa5d718f415ca450fd5abf0694750a16562c64f7b5e392d5e6R62

With the current design, we need to make this call even for fields that are *not* `IsRefined()`.  In this scenario, we are passing an unset `refinement_funcs_`, which is never actually used because there is an internal check on `IsRefined()` inside the `Metadata` constructor, followed by an additional `IsValid()`.

Yurlungur: Oh I see. So long as it's still safe somewhere, I think this is fine. Tests will fail though (because I think this is a test case.)

pdmullen:
Added a few tests that `refinement_funcs_` contains the expected entries after adding a `SparsePool`.  The newly added test fails under current `develop` (as it should).

tst/unit/test_state_descriptor.cpp
```diff
@@ -383,16 +383,16 @@ TEST_CASE("Test dependency resolution in StateDescriptor", "[StateDescriptor]")
       }
     }
 
-    WHEN("We register a dense variable custom prolongation/restriction") {
+    WHEN("We register a dense variable with default prolongation/restriction") {
```
Yurlungur, line 386: These are still custom right? Not default?

Yurlungur (approved):
LGTM. Thanks for adding the test coverage!

tst/unit/test_state_descriptor.cpp
pdmullen: I don't think so... only the `SparsePool` has enrolled custom P/R ops via `RegisterRefinementOps`.  The dense field just uses the default ones.

Yurlungur: Oh yes you're right. :+1:

brryan (approved):
LGTM, I am not offended by the default call to RegisterOps. Only question is whether metadata is the right place for default values of register ops to live (maybe it is). If not, one could do something like
```c++
using DefaultRegisterOps = RegisterOps<refinement_ops::ProlongateSharedMinMod, refinement_ops::RestrictAverage>;
```
in the refinement functions file.

pdmullen:
> LGTM, I am not offended by the default call to RegisterOps. Only question is whether metadata is the right place for default values of register ops to live (maybe it is). If not, one could do something like
> 
> ```c++
> using DefaultRegisterOps = RegisterOps<refinement_ops::ProlongateSharedMinMod, refinement_ops::RestrictAverage>;
> ```
> 
> in the refinement functions file.

I see your point.  I could imagine a scenario where one wants to handle different fields via different P/R ops, therefore, I think that `Metadata` is a natural place for this to live, however, others may disagree.

PR #996, opened 2024-01-19
Remove dynamic allocations from swarm particle creation

brryan:
## PR Summary

`Swarm`s previously created multiple `ParArray`s during the `AddEmptyParticles` call to create new particles. This PR pre-allocates that memory during resize operations and returns an accessor to new particle indices, resolving that issue.

This necessarily breaks previous usage of the `AddEmptyParticles` function and will affect all downstream codes using particles.

API change:

`ParArray1D<bool> Swarm::AddEmptyParticles(const int num_to_add, ParArrayND<int> &new_indices)`

is now

`NewParticlesContext Swarm::AddEmptyParticles(const int num_to_add)`

where `NewParticlesContext` adds the methods `int GetNewParticlesMaxIndex()` that returns the max index of the new particles -> swarm index map, and `int GetNewParticleIndex(const int n)` converts a new particle index to the swarm index.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [ ] Docs build


example/particle_leapfrog/particle_leapfrog.cpp
```diff
@@ -189,7 +187,9 @@ void ProblemGenerator(MeshBlock *pmb, ParameterInput *pin) {
   // This hardcoded implementation should only used in PGEN and not during runtime
   // addition of particles as indices need to be taken into account.
   pmb->par_for(
-      PARTHENON_AUTO_LABEL, 0, num_particles_this_block - 1, KOKKOS_LAMBDA(const int n) {
+      PARTHENON_AUTO_LABEL, 0, newParticlesContext.GetNewParticlesMaxIndex(),
+      KOKKOS_LAMBDA(const int new_n) {
+        const int n = newParticlesContext.GetNewParticleIndex(new_n);
```
Yurlungur: I like this better!

src/interface/swarm.hpp
```diff
@@ -47,6 +47,31 @@ struct BoundaryDeviceContext {
   ParticleBound *bounds[6];
 };
 
+// This class is returned by AddEmptyParticles. It provides accessors to the new particle
+// memory by wrapping the persistent new_indices_ array.
+class NewParticlesContext {
```
Yurlungur, line 52: This is equivalent to a view I guess, but it's cleaner, more clear design. So I like it.

Yurlungur (approved):
Overall this seems like a pretty minor change, but good for performance. And I like the new API, which feels cleaner. :+1:

src/interface/swarm.hpp
brryan: A view wouldn't give you the max index, though, because here in general the max index of new particles will be less than the preallocated new particle indices `ParArray`'s `size()`, right?

Anyway, the "contexts" pattern seems to be taking over for particles. It does seem pretty clear to me too.

Yurlungur: Ah ok yes I see what you mean. :+1:

brryan:
Alright this is ready for review/merge

Yurlungur:
@pgrete or @bprather maybe one of you can review?

c-prather (approved):
LGTM.  I'm not far enough into particles to be annoyed by the slight interface change, and this is clean and easy enough to understand with the example.

example/particle_leapfrog/particle_leapfrog.cpp
```diff
@@ -175,9 +175,7 @@ void ProblemGenerator(MeshBlock *pmb, ParameterInput *pin) {
 
   Kokkos::deep_copy(pmb->exec_space, ids_this_block, ids_this_block_h);
 
-  ParArrayND<int> new_indices;
-  const auto new_particles_mask =
-      swarm->AddEmptyParticles(num_particles_this_block, new_indices);
+  auto newParticlesContext = swarm->AddEmptyParticles(num_particles_this_block);
```
pgrete:
```suggestion
  auto new_particles_context = swarm->AddEmptyParticles(num_particles_this_block);
```
style also below and in other files

example/particles/particles.cpp
```diff
@@ -278,73 +278,75 @@ TaskStatus CreateSomeParticles(MeshBlock *pmb, const double t0) {
...
-            v(2, n) = v(2, n) - vdN / NdN * z(n);
-
-            // Normalize
-            Real v_tmp = sqrt(v(0, n) * v(0, n) + v(1, n) * v(1, n) + v(2, n) * v(2, n));
-            PARTHENON_DEBUG_REQUIRE(v_tmp > 0., "Speed must be > 0!");
-            for (int ii = 0; ii < 3; ii++) {
-              v(ii, n) *= vel / v_tmp;
-            }
+        // PARTHENON_AUTO_LABEL, 0, swarm->GetMaxActiveIndex(), KOKKOS_LAMBDA(const int n)
+        // {
```
pgrete: Leftover debug code?
```suggestion
```

src/interface/swarm.cpp
```diff
@@ -282,58 +288,52 @@ void Swarm::setPoolMax(const std::int64_t nmax_pool) {
...
-  if (num_to_add <= 0) {
-    new_indices = ParArrayND<int>();
-    return ParArray1D<bool>();
-  }
-
-  while (free_indices_.size() < num_to_add) {
-    increasePoolMax();
-  }
+NewParticlesContext Swarm::AddEmptyParticles(const int num_to_add) {
+  PARTHENON_DEBUG_REQUIRE(num_to_add >= 0, "Cannot add negative numbers of particles!");
```
pgrete, line 296: Does it even makes sense to allow for `num_to_add == 0`?

src/interface/swarm.cpp
```diff
@@ -282,58 +288,52 @@ void Swarm::setPoolMax(const std::int64_t nmax_pool) {
...
 
-  num_active_ += num_to_add;
+    num_active_ += num_to_add;
 
-  Kokkos::deep_copy(new_mask, new_mask_h);
-  Kokkos::deep_copy(mask_, mask_h);
-  blockIndex_.DeepCopy(blockIndex_h);
+    Kokkos::deep_copy(mask_, mask_h);
+    blockIndex_.DeepCopy(blockIndex_h);
+    newIndicesMaxIdx_ = num_to_add - 1;
```
pgrete: Just double checking that the new indices do not need to take into account existing indices here, correct?

src/interface/swarm.cpp
```diff
@@ -366,13 +366,13 @@ void Swarm::Defrag() {
   std::int64_t num_free = (max_active_index_ + 1) - num_active_;
   auto pmb = GetBlockPointer();
 
-  ParArrayND<int> from_to_indices("from_to_indices", max_active_index_ + 1);
-  auto from_to_indices_h = from_to_indices.GetHostMirror();
+  // ParArrayND<int> from_to_indices("from_to_indices", max_active_index_ + 1);
```
```suggestion
```
Leftover?

src/interface/swarm.cpp
```diff
@@ -916,6 +918,7 @@ int Swarm::CountParticlesToSend_() {
     ApplyBoundaries_(total_noblock_particles, noblock_indices);
   }
 
+  // TODO(BRR) don't allocate dynamically
```
pgrete, line 924: This is meant as a future to do and not part of this PR, correct? I'm just asking because thisPR is about dynamic allocation.

src/interface/swarm.cpp
```diff
@@ -1087,11 +1090,11 @@ void Swarm::UnloadBuffers_() {
   auto &bdvar = vbswarm->bd_var_;
 
   if (total_received_particles_ > 0) {
-    ParArrayND<int> new_indices;
-    auto new_mask = AddEmptyParticles(total_received_particles_, new_indices);
+    auto newParticlesContext = AddEmptyParticles(total_received_particles_);
 
-    ParArrayND<int> neighbor_index("Neighbor index", total_received_particles_);
-    ParArrayND<int> buffer_index("Buffer index", total_received_particles_);
+    // TODO(BRR) remove these dynamic allocs
```
pgrete: As above, this is meant as a future to do and not part of this PR, correct? I'm just asking because thisPR is about dynamic allocation.

pgrete (approved):
LGTM and the changes make sense from a performance point of view.
The main sna~~rky~~ke case-y comment I have why you chose you use camel case (and even refactor some existing variables) versus snake case.
I'm general in favor of consistent style across the code base.

src/interface/swarm.cpp
brryan: I think this case can easily appear in downstream codes, if say you have some function that samples a certain number of particles on each meshblock according to some function, that function often says to sample zero particles in certain regions. If we prevent that case here, it requires extra logic on the downstream code side.

src/interface/swarm.cpp
brryan: Yes there are still some outstanding dynamic allocation issues especially related to internal particle list sorting. One of the next things I want to work on is moving all this logic to device with `par_scan`, and that change should naturally clean up these dynamic alloc issues.

src/interface/swarm.cpp
brryan: Yes there are still some outstanding dynamic allocation issues especially related to internal particle list sorting. One of the next things I want to work on is moving all this logic to device with `par_scan`, and that change should naturally clean up these dynamic alloc issues.

src/interface/swarm.cpp
brryan: That's right -- `newIndicesMaxIdx_` is for iterating over 1D arrays with valid data only from indices `[0, newIndicesMaxIdx_]` One of those arrays then returns indices into the regular swarm array which holds all the particles and will generally have valid indices beyond `newIndicesMaxIdx_`.

src/interface/swarm.cpp
brryan: I think this PR is the right place to remove this dynamic alloc so I did :+1: thanks for highlighting it

brryan:
> LGTM and the changes make sense from a performance point of view. The main sna~rky~ke case-y comment I have why you chose you use camel case (and even refactor some existing variables) versus snake case. I'm general in favor of consistent style across the code base.

@pgrete You are absolutely right about the case style issue with variable names, I was mistakenly taking this in the wrong direction. Thanks for pointing that out. If I didnt miss anything, I've switched every variable in swarm over to camel case. This should be ready for merge now if you're happy

brryan:
Alright all concerns look to be addressed to me so setting this to automerge, thanks all for reviews.

PR #998, opened 2024-01-23
tensor indices added to sparse pack

mari2895:
## PR Summary

Added the ability for the typed variables to take tensor indices so that sparse pack can support tensor variables. 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


tst/unit/test_sparse_pack.cpp
```diff
@@ -77,13 +77,76 @@ struct v5 : public parthenon::variable_names::base_t<false> {
   static std::string name() { return "v5"; }
 };
 
+struct v7 : public parthenon::variable_names::base_t<false, 3, 3> {
```
Yurlungur: I just thought of this---maybe we should do `ANYDIM, 3` here, instead of `3, 3` to show that the slowest moving index can be arbitrary size

src/interface/sparse_pack.hpp
```diff
@@ -98,6 +99,17 @@ struct base_t {
...
   explicit base_t(int idx1) : idx(idx1) {}
 
+  template <typename... Args, REQUIRES(all_implement<integral(Args...)>::value)>
+  /*
+    for 2D:, (M, N),
+    idx(m, n) = N*m + n
+    for 3D: (L, M, N)
+    idx(l, m, n) = (M*l + m)*N + n
+                 = l*M*N + m*N + n
+   */
```
Yurlungur, line 108: I think we should keep this comment but it should probably be above the `template<...` line

Yurlungur (approved):
LGTM :+1:

Full disclosure. I helped @mari2895 write this (@mari2895 is Mariam---she's a PhD student working with me). So probably we should at least get @lroberts36 's eyes on this before merge.

src/interface/sparse_pack.hpp
```diff
@@ -98,6 +99,17 @@ struct base_t {
   KOKKOS_INLINE_FUNCTION
   explicit base_t(int idx1) : idx(idx1) {}
 
+  template <typename... Args, REQUIRES(all_implement<integral(Args...)>::value)>
```
lroberts36:
```suggestion
  template <typename... Args, REQUIRES(all_implement<integral(Args...)>::value), REQUIRES(sizeof...(Args) == sizeof...(NCOMP)>
```

src/interface/sparse_pack.hpp
```diff
@@ -114,6 +126,19 @@ struct base_t {
   static int size() { return multiply<NCOMP...>::value; }
```
lroberts36, line 136: Maybe we should have a static member that gives the shape of the type so that it is easy to make sure the metadata is consistent, e.g. 
```c++
static std::vector<int> GetShape() {return std::vector<int>{NCOMP...};}
```

src/interface/sparse_pack.hpp
```diff
@@ -114,6 +126,19 @@ struct base_t {
...
+            REQUIRES(all_implement<integral(TailArgs...)>::value)>
+  KOKKOS_INLINE_FUNCTION static auto GetIndex_(int first, TailArgs... rest) {
+    if constexpr (sizeof...(Tail) == 0)
+      return first;
+    else if constexpr (sizeof...(TailArgs) == 0)
+      return first;
+    else
+      return (multiply<Tail...>::value * first +
+              GetIndex_<Tail...>(std::forward<TailArgs>(rest)...));
+  }
```
lroberts36: FWIW, I think this could be done with a fold expression as 
```c++
template <class... Args>
KOKKOS_INLINE_FUNCTION static auto GetIndex_(Args... args) {
  int idx = 0; 
  ([&]{idx *= NCOMP; idx += args;}(), ...);
  return idx;
}
```

src/interface/sparse_pack.hpp
lroberts36: agreed

src/interface/sparse_pack.hpp
lroberts36: and maybe 
```c++
template <int idx> 
  static constexpr auto GetDim() { 
    return std::get<idx>(std::make_tuple(NCOMP...));
  }
```
so it is easy to get the dimensions of each index in a kernel.

lroberts36 (approved):
This LGTM, approving now but we definitely should check that we are getting the correct number of arguments. The rest of the suggestions are optional.

tst/unit/test_sparse_pack.cpp
```diff
@@ -77,13 +77,76 @@ struct v5 : public parthenon::variable_names::base_t<false> {
...
+      }
+      THEN("A sparse pack can correctly index into tensor types") {
+        auto desc = parthenon::MakePackDescriptor<v7>(pkg.get());
+        auto sparse_pack = desc.GetPack(&mesh_data);
+        int nwrong = 0;
+        par_reduce(
+            loop_pattern_mdrange_tag, "check vector", DevExecSpace(), 0,
+            sparse_pack.GetNBlocks() - 1, kb.s, kb.e, jb.s, jb.e, ib.s, ib.e,
+            KOKKOS_LAMBDA(int b, int k, int j, int i, int &ltot) {
+              for (int l = 0; l < 3; ++l) {
```
lroberts36:
```suggestion
              for (int l = 0; l < v7::GetDim<0>(); ++l) {
```

lroberts36: and similarly below

src/interface/sparse_pack.hpp
Yurlungur: @mari2895 added this but quick question: If the size of args is 0 (i.e., we have zero or 1 argument passed in) does the single-argument constructor (line 100) take precedence? We should ensure it does, as it's important for backwards compatibility that scalar and vector calls work without having to specify NCOMP.

src/interface/sparse_pack.hpp
Yurlungur: @mari2895 if you want to use this, you should first double check that it still works (i.e., that tests pass). You also still want to require args are integral type.

src/interface/sparse_pack.hpp
lroberts36: I checked this in godbolt yesterday while doing the review. I initially thought the templated constructor would cover the single argument constructor, but it appears that the single argument constructor takes precedence. I didn't carefully look at the zero argument case though.

Yurlungur: Good enough for me.

src/interface/sparse_pack.hpp
Yurlungur: :+1: Probably should do metadata consistency checking in another PR down the road.

src/interface/sparse_pack.hpp
mari2895: done

src/interface/sparse_pack.hpp
mari2895: @Yurlungur @lroberts36 ok I am going try out this change and check if it works.

tst/unit/test_sparse_pack.cpp
mari2895: done

tst/unit/test_sparse_pack.cpp
```diff
@@ -77,13 +77,78 @@ struct v5 : public parthenon::variable_names::base_t<false> {
...
+        auto desc = parthenon::MakePackDescriptor<v7>(pkg.get());
+        auto sparse_pack = desc.GetPack(&mesh_data);
+        int nwrong = 0;
+        par_reduce(
+            loop_pattern_mdrange_tag, "check vector", DevExecSpace(), 0,
+            sparse_pack.GetNBlocks() - 1, kb.s, kb.e, jb.s, jb.e, ib.s, ib.e,
+            KOKKOS_LAMBDA(int b, int k, int j, int i, int &ltot) {
+              // 0-th is ANYDIM, 1st is 3.
+              for (int l = 0; l < v7::GetDim<1>(); ++l) {
+                for (int m = 0; m < v7::GetDim<1>(); ++m) {
```
pgrete: Are these correct?
If yes, then some comment explaining why using `GetDim<1>` twice works would be good.
If this this a typo, I also suggest to update `     const std::vector<int> tensor_shape{N, N, N, 3, 3};` above to use different dimensions in the tensor indices.

src/interface/sparse_pack.hpp
```diff
@@ -90,6 +90,7 @@ KOKKOS_INLINE_FUNCTION PackIdx operator+(T offset, PackIdx idx) {
 // device
 namespace variable_names {
 // Struct that all variable_name types should inherit from
+constexpr int ANYDIM = -1234;
```
pgrete: I'm not too familiar with this variable name approach, so it's not 100% clear to me what `ANYDIM` is used for and why it's defined in Parthenon, when it's not being used expect the test.
Would be great if you could add a brief explanation there (also on the choice of the `-1234` magic number).

pgrete (commented):
Welcome Mariam and thanks for your contribution.
I just have two clarifying questions and am then happy to approve.

tst/unit/test_sparse_pack.cpp
mari2895: Hi, nice to meet you and  thank you for the comments! I think, it is correct because both "l" and "m" should go from 0 to 3 and 3 is v7::GetDim<1>(), @Yurlungur and @pgrete please correct me if I am wrong. If you agree, I will add a comment to clarify.

src/interface/sparse_pack.hpp
mari2895: To my understanding, the only purpose of ANYDIM is to highlight that first dim does not matter in our calculations and can be anything. We chose it to be negative number for purpose and 1234 is just random (i think). @Yurlungur might have a better explanation for this.

tst/unit/test_sparse_pack.cpp
Yurlungur: This is for the tensor variable type not the array. You could use the dimensions of the `std::vector` as @pgrete suggests, and I think that might be more clear.

The reason it's `1` both times here is that `0` is `ANYDIM`. That's supposed to test the functionality that the slowest moving index of the variable shape allowed to float---i.e., be set-able at runtime rather than compile time. (It turns out that for indexing arithmetic that size doesn't matter... same as how multi-dimensional C-style arrays can be passed into a function with one floating size.)

@mari2895 I suggest you switch to using `tensor_shape[4]` for `l` and `tensor_shape[3]` for `m` for clarity, as @pgrete suggests.

src/interface/sparse_pack.hpp
Yurlungur: I wrote a bit of an explanation above... but just like with how in C-style arrays, the size of the slowest-moving index can be free floating when you pass a multi-D array into a function, the indexing of a tensor variable doesn't depend on the size of the slowest moving index. This is a way of declaring the variable is of the correct tensor rank, but has a free-floating slowest-moving index. e.g.,
```C++
myvar : base_t<ANYDIM, 3, 3>
```
is a rank 3 tensor, which is Nx3x3, where N can be set at runtime. The reason for `-1234` is just that it's a negative number. It could be anything.

We should maybe add a comment explaining this @mari2895 . We could also add a check to the constructor that `ANYDIM` is only ever the slowest-moving index (i.e., the first one). Could probably be done with a fold expression.... Add the following function in private scope of the class (near `GetIndex`):
```C++
template<typename... Tail, REQUIRES(all_implement<integral(Tail...)>::value>
static constexpr bool CheckArgs_(int head, Tail... tail) {
    return (... && (tail > 0));
}
```
then change the constructor to:
```C++
 template <typename... Args, REQUIRES(all_implement<integral(Args...)>::value),
            REQUIRES(sizeof...(Args) == sizeof...(NCOMP))>
  KOKKOS_INLINE_FUNCTION explicit base_t(Args... args)
      : idx(GetIndex_(std::forward<Args>(args)...)) {
            static_assert(CheckArgs_(NCOMP...), 
    "All dimensions must be strictly positive, "
    "except the first (slowest), which may be ANYDIM.");
      }
```
As always, check that tests pass, run formatting, and linting.

tst/unit/test_sparse_pack.cpp
mari2895: Done

src/interface/sparse_pack.hpp
mari2895: Done, all tests passed, formating and linting run.

Yurlungur: @pgrete does the revised version work for you?

tst/unit/test_sparse_pack.cpp
mari2895: @Yurlungur this change causes some checks to fail.

Yurlungur: Let's discuss over slack.

Yurlungur: OH it's inside of a kokkos loop!

To make them pass, assign a variable above the `par_for` like `int nl = tensor_shape[4]` and `int nm = tensor_shape[3]` and then use `nl` and `nm` in the loop itself

src/interface/sparse_pack.hpp
pgrete: Thanks for the clarification. I guess my brain is just/still struggling to process all the template magic happening here (in combination with not being familiar with the sparse pack approach).
Is the only purpose of `ANYDIM` just to make it a three dimensional (so that the pointer approach under the hood works, which allows to specify the first dimension at runtime)?

Similarly, my original question was also aimed at why `ANYDIM` is required at this place (i.e., part of the public interface) when it's only used in a test. So in other words, do I have to use `ANYDIM` on the downstream side (similar to the test) when I want to use sparse packs or could `ANYDIM` also completely live in the test below (and not being part of the general interface).

At some point I should probably code up my own example to figure out how things work.

tst/unit/test_sparse_pack.cpp
pgrete: Just to be 100% sure (tbh I'm still confused), the picking the correct indices in the original approach should have read

1. (using c-style indices based on `v7(l,m)`)
```
              for (int l = 0; l < v7::GetDim<0>(); ++l) {
                for (int m = 0; m < v7::GetDim<1>(); ++m) {
```
2. or (using c-style indices based on `v7(l,m)` with the `ANYDIM` offset)
```
              for (int l = 0; l < v7::GetDim<1>(); ++l) {
                for (int m = 0; m < v7::GetDim<2>(); ++m) {
```
3. or (using kji based indices in the context of our "dimensions", i.e., GetDim(1) is related to the rightmost `i` index)
```
              for (int l = 0; l < v7::GetDim<1>(); ++l) {
                for (int m = 0; m < v7::GetDim<0>(); ++m) {
```
4. or (using kji based indices in the context of our "dimensions", i.e., GetDim(1) is related to the rightmost `i` index with `ANYDIM` offset)
```
              for (int l = 0; l < v7::GetDim<2>(); ++l) {
                for (int m = 0; m < v7::GetDim<1>(); ++m) {
```
?

Based on your comment above, I think it's either option 2 or 4 but between those, I'm not sure.

pgrete (approved):
Thanks for the clarification.
As you can see, I'm still very  much confused (but this is independent of this PR and more general around sparse packs so I'm approving as is).

src/interface/sparse_pack.hpp
Yurlungur: > Is the only purpose of ANYDIM just to make it a three dimensional (so that the pointer approach under the hood works, which allows to specify the first dimension at runtime)?

That's essentially right. 

> Similarly, my original question was also aimed at why ANYDIM is required at this place (i.e., part of the public interface) when it's only used in a test. 

We don't NEED `ANYDIM` at all. You can put any number you want in the slowest moving index (so long as it's positive) and indexing will work properly. However, it's a clarity thing. However, it's in the tests because I wanted to test that the machinery works with this piece in place.

> in other words, do I have to use ANYDIM on the downstream side (similar to the test) when I want to use sparse packs

No, you can use the specific tensor shape you want without ANYDIM. Or for scalars and vectors, you can ignore the indexing machinery entirely. 

> could ANYDIM also completely live in the test below (and not being part of the general interface).

It could, but I would prefer not. It's an entirely optional piece of parthenon machinery. But it's useful to support, e.g., things like multi-group moments, where the `ANYDIM` index is the number of groups.

tst/unit/test_sparse_pack.cpp
Yurlungur: It would have been option 1 actually... Coming from this line:
```C++
  template <int idx>
  static constexpr auto GetDim() {
    return std::get<idx>(std::make_tuple(NCOMP...));
  }
```
That said, that's pretty counter to our usual way of doing things. @mari2895 I would suggest replacing this with
```C++
  template <int idx>
  static constexpr auto GetDim() {
    return std::get<sizeof...(NCOMP) - idx>(std::make_tuple(NCOMP...));
  }
```
which should match the more proper parthenon style with option 4.

PR #999, opened 2024-01-25
Add a post-initialization hook

pdmullen:
## PR Summary

This PR adds a hook for downstream codes to call a post-initialization function.  It is called immediately after `ProblemGenerator` but before `PreCommFillDerived`.  Potential use cases include setting primitive variables in the problem generator, but calling P2C in `PostInitialization`.  Just like `ProblemGenerator`, there are both `Mesh` and `MeshBlock` variations of `PostInitialization`.  Unlike `ProblemGenerator`, `PostInitialization` takes as argument `MeshBlockData<Real>`.  This could understandably be a debatable choice --- in my use case, it allows me to set `PostInitialization = P2C<MeshBlockData<Real>>` whereas in other contexts in the codebase, I could use `P2C<MeshData<Real>>`.  

EDIT: Following PR discussion, I have ensured consistency between function arguments for `ProblemGenerator` and `PostInitialization`, however, I have opened up issue #1002 to facilitate future work/discussion over the points raised here.  

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/doc/sphinx/src/README.rst
+++ b/doc/sphinx/src/README.rst
@@ -178,6 +178,8 @@ Mesh
 ^^^^
 
 -  ``InitUserMeshData``
+-  ``ProblemGenerator``
+-  ``PostInitialization``
 -  ``PreStepUserWorkInLoop``
 -  ``PostStepUserWorkInLoop``
 -  ``UserWorkAfterLoop``
@@ -188,6 +190,7 @@ MeshBlock
 -  ``InitApplicationMeshBlockData``
 -  ``InitMeshBlockUserData``
 -  ``ProblemGenerator``
+-  ``PostInitialization``
 -  ``UserWorkBeforeOutput``
 
 To redefine these functions, the user sets the respective function
@@ -195,12 +198,12 @@ pointers in the ApplicationInput member app_input of the
 ParthenonManager class prior to calling ``ParthenonInit``. This is
 demonstrated in the ``main()`` functions in the examples.
 
-Note that the ``ProblemGenerator``\ s of ``Mesh`` and ``MeshBlock`` are
-mutually exclusive. Moreover, the ``Mesh`` one requires
-``parthenon/mesh/pack_size=-1`` during initialization, i.e., all blocks
-on a rank need to be in a single pack. This allows to use MPI reductions
-inside the function, for example, to globally normalize quantities. The
-``parthenon/mesh/pack_size=-1`` exists only during problem
+Note that the ``ProblemGenerator``\ s (and ``PostInitialization``\ s) of
+``Mesh`` and ``MeshBlock`` are mutually exclusive. Moreover, the ``Mesh``
+ones requires ``parthenon/mesh/pack_size=-1`` during initialization, i.e.,
+all blocks on a rank need to be in a single pack. This allows to use MPI
+reductions inside the function, for example, to globally normalize quantities.
+The ``parthenon/mesh/pack_size=-1`` exists only during problem
 inititalization, i.e., simulations can be restarted with an arbitrary
 ``pack_size``. For an example of the ``Mesh`` version, see the `Poisson
 example <https://github.com/parthenon-hpc-lab/parthenon/blob/develop/example/poisson/parthenon_app_inputs.cpp>`__.
--- a/src/application_input.hpp
+++ b/src/application_input.hpp
@@ -36,7 +36,8 @@ struct ApplicationInput {
   std::function<void(Mesh *, ParameterInput *)> InitUserMeshData = nullptr;
   std::function<void(Mesh *, ParameterInput *, MeshData<Real> *)> MeshProblemGenerator =
       nullptr;
-  std::function<void(Mesh *, MeshData<Real> *)> MeshPostInitialization = nullptr;
+  std::function<void(Mesh *, ParameterInput *, MeshData<Real> *)> MeshPostInitialization =
+      nullptr;
 
   std::function<void(Mesh *, ParameterInput *, SimTime &)> PreStepMeshUserWorkInLoop =
       nullptr;
@@ -58,7 +59,8 @@ struct ApplicationInput {
       InitApplicationMeshBlockData = nullptr;
   std::function<void(MeshBlock *, ParameterInput *)> InitMeshBlockUserData = nullptr;
   std::function<void(MeshBlock *, ParameterInput *)> ProblemGenerator = nullptr;
-  std::function<void(MeshBlockData<Real> *)> PostInitialization = nullptr;
+  std::function<void(MeshBlockData<Real> *, ParameterInput *)> PostInitialization =
+      nullptr;
   std::function<void(MeshBlock *, ParameterInput *)> MeshBlockUserWorkBeforeOutput =
       nullptr;
 };
--- a/src/application_input.hpp
+++ b/src/application_input.hpp
@@ -59,8 +59,7 @@ struct ApplicationInput {
       InitApplicationMeshBlockData = nullptr;
   std::function<void(MeshBlock *, ParameterInput *)> InitMeshBlockUserData = nullptr;
   std::function<void(MeshBlock *, ParameterInput *)> ProblemGenerator = nullptr;
-  std::function<void(MeshBlockData<Real> *, ParameterInput *)> PostInitialization =
-      nullptr;
+  std::function<void(MeshBlock *, ParameterInput *)> PostInitialization = nullptr;
   std::function<void(MeshBlock *, ParameterInput *)> MeshBlockUserWorkBeforeOutput =
       nullptr;
 };
--- a/src/application_input.hpp
+++ b/src/application_input.hpp
@@ -1,5 +1,5 @@
 //========================================================================================
-// (C) (or copyright) 2020-2022. Triad National Security, LLC. All rights reserved.
+// (C) (or copyright) 2020-2024. Triad National Security, LLC. All rights reserved.
 //
 // This program was produced under U.S. Government contract 89233218CNA000001 for Los
 // Alamos National Laboratory (LANL), which is operated by Triad National Security, LLC
@@ -36,6 +36,7 @@ struct ApplicationInput {
   std::function<void(Mesh *, ParameterInput *)> InitUserMeshData = nullptr;
   std::function<void(Mesh *, ParameterInput *, MeshData<Real> *)> MeshProblemGenerator =
       nullptr;
+  std::function<void(Mesh *, MeshData<Real> *)> MeshPostInitialization = nullptr;
 
   std::function<void(Mesh *, ParameterInput *, SimTime &)> PreStepMeshUserWorkInLoop =
       nullptr;
@@ -57,6 +58,7 @@ struct ApplicationInput {
       InitApplicationMeshBlockData = nullptr;
   std::function<void(MeshBlock *, ParameterInput *)> InitMeshBlockUserData = nullptr;
   std::function<void(MeshBlock *, ParameterInput *)> ProblemGenerator = nullptr;
+  std::function<void(MeshBlockData<Real> *)> PostInitialization = nullptr;
   std::function<void(MeshBlock *, ParameterInput *)> MeshBlockUserWorkBeforeOutput =
       nullptr;
 };
--- a/src/mesh/mesh.cpp
+++ b/src/mesh/mesh.cpp
@@ -3,7 +3,7 @@
 // Copyright(C) 2014 James M. Stone <jmstone@princeton.edu> and other code contributors
 // Licensed under the 3-clause BSD License, see LICENSE file for details
 //========================================================================================
-// (C) (or copyright) 2020-2023. Triad National Security, LLC. All rights reserved.
+// (C) (or copyright) 2020-2024. Triad National Security, LLC. All rights reserved.
 //
 // This program was produced under U.S. Government contract 89233218CNA000001 for Los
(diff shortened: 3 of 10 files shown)


src/application_input.hpp
```diff
@@ -36,6 +36,7 @@ struct ApplicationInput {
   std::function<void(Mesh *, ParameterInput *)> InitUserMeshData = nullptr;
   std::function<void(Mesh *, ParameterInput *, MeshData<Real> *)> MeshProblemGenerator =
       nullptr;
+  std::function<void(Mesh *, MeshData<Real> *)> MeshPostInitialization = nullptr;
```
Yurlungur: Does one need to pass both `Mesh` and `MeshData`? I suppose they could have different subsets of blocks

src/application_input.hpp
```diff
@@ -57,6 +58,7 @@ struct ApplicationInput {
       InitApplicationMeshBlockData = nullptr;
   std::function<void(MeshBlock *, ParameterInput *)> InitMeshBlockUserData = nullptr;
   std::function<void(MeshBlock *, ParameterInput *)> ProblemGenerator = nullptr;
+  std::function<void(MeshBlockData<Real> *)> PostInitialization = nullptr;
   std::function<void(MeshBlock *, ParameterInput *)> MeshBlockUserWorkBeforeOutput =
```
Yurlungur: IMO `MeshBlock` vs `MeshBlockData` is just a choice. EIther is fine, so long as it's documented. THough one could argue consistency is better?

src/mesh/mesh.cpp
```diff
@@ -930,6 +933,10 @@ void Mesh::Initialize(bool init_problem, ParameterInput *pin, ApplicationInput *
       PARTHENON_REQUIRE_THROWS(
           !(ProblemGenerator != nullptr && block_list[0]->ProblemGenerator != nullptr),
           "Mesh and MeshBlock ProblemGenerators are defined. Please use only one.");
+      PARTHENON_REQUIRE_THROWS(
+          !(PostInitialization != nullptr &&
+            block_list[0]->PostInitialization != nullptr),
+          "Mesh and MeshBlock PostInitializations are defined. Please use only one.");
```
Yurlungur, line 939: Is this actually required? Why can't they both be set if desired?

Yurlungur (approved):
LGTM. Minor comments below. Also is there documentation for app inputs? If so we should document this.

We may also want to test it somehow...

src/application_input.hpp
pdmullen: I myself would certainly never use `PostInitialization` in a way that would *require* both `Mesh` and `MeshData`.  Here I just tried to keep the function arguments consistent between the `Mesh` variations of `ProblemGenerator` and `PostInitialization`.

src/application_input.hpp
pdmullen: Perhaps.  I can be open here.  

My particular decision was motivated by the fact that I already had a function that did ops on `MeshData`.  With the choice in this MR, I could just template my pre-existing function on `typename` and have it also operate on `MeshBlockData`.  This is not a possibility if `PostInitialization` takes as argument `*pmb`.  Therefore, I think working with `MeshBlockData` offers more opportunity to downstream users.

src/mesh/mesh.cpp
pdmullen: In the current design, if both `Mesh` and `MeshBlock` variations of `PostInitialized` are set, without this `PARTHENON_REQUIRE_THROWS`, only the `Mesh` variation will be called (see the later `if`/`else` logic).  This mimics handling of `ProblemGenerator`.  We could alter the logic to call both, but then there will be an enforced order (which should come first: `Mesh` or `MeshBlock`).  This seems more confusing than beneficial.  IMHO, if you use the `Mesh` version of `ProblemGenerator`, you should use the `Mesh` version of `PostInitialization`. If you use the `MeshBlock` version of `ProblemGenerator`, you should use the `MeshBlock` version of `PostInitialization`.  To me, it seems not favorable to mix and match or to use both.

c-prather:
Looking forward to using this.  My 2c on AppInput registrations is we should be deprecating the block-based functions and look to eventually providing only one callback, which would take a `MeshData` argument and no longer need the `Mesh` prefix.  Letting applications dig themselves deeper into the `MeshBlockData`/`MeshData` duplication in new development seems counterproductive.

src/mesh/mesh.cpp
c-prather: I think allowing two callbacks with the same name is inviting ordering bugs.  It's not obvious from the naming which will get called first, and the order might be especially important during initialization.  We could document it, but why bother when downstreams can control when/how/if they loop through blocks with a few lines in a single MeshData-wide implementation?

src/application_input.hpp
c-prather: I think consistency is a strong argument *for* this `*Data` version.  There's no compelling reason for the callback functions to break from Parthenon's otherwise really solid pattern of passing MeshData & MeshBlockData pointers everywhere (except the places it passes shared_ptr objects, but that's for another time).  I'd be in favor of changing the `ProblemGenerator` and `UserWork` functions to match, frankly, I think they're just Athena++ holdovers right?

Yurlungur: They are just athena++ holdovers. I am fine with moving them over, so long as we warn downstream users of the breaking change.

src/mesh/mesh.cpp
Yurlungur: :+1: that's fine---we can just leave it as @pdmullen designed.

pdmullen:
> LGTM. Minor comments below. Also is there documentation for app inputs? If so we should document this.
> 
> We may also want to test it somehow...

Updated documentation.

src/application_input.hpp
pdmullen: If we go that route, that should be a separate PR.

c-prather (approved):
This looks great.  I would like to shrink the interface in future as I mentioned, to avoid so much MD/MBD duplication.  But that's a long way off I think, we need to get downstreams mostly on board with going MD-only.

doc/sphinx/src/README.rst
```diff
@@ -178,6 +178,8 @@ Mesh
 ^^^^
 
 -  ``InitUserMeshData``
+-  ``ProblemGenerator``
+-  ``PostInitialization``
 -  ``PreStepUserWorkInLoop``
```
pgrete, line 183: While we're at it, I think in the context of this PR it'd be great to also add `UserWorkBeforeLoop` to the doc and emphasize the difference between `PostInit` and `UserWorkBeforeLoop` (as on first sight, they may seem identical in purpose).

src/application_input.hpp
pgrete: Speaking of consistency (which I'm also in favor of), is there a reason to not also pass `ParameterInput` (also above)?
I see that one could argue that everything should be processed in the `ProblemGenerator` and then potentially re-extracted from `Params` but I imagine that a common downstream usecase is to directly work with `pin`.

Regarding templating, this should also work based on the pointer type if you want to make a function that works both ways (though I'm not sure it's sth we want to encourage as the `MeshData` approach seems the way to go based on everyones experience over the past years).

pgrete (commented):
I get the impression that the discussion on the interface is not quite done yet, is it?

src/application_input.hpp
pdmullen: @pgrete, I think the current design is most amenable to re-purposing pre-existing functions (i.e., those typically used by a driver) for `PostInitialization`.  

If, for example, I use the same function call for `PostInitialization` and `FillDerived`, I may have to jump through some hoops if among the two only `PostInitialization` takes as argument `*pin`.  

I do see your point though that function calls inside `Mesh::Initialize` and the `init_problem` boolean (might) should have access to `*pin`, although traditionally I would just exploit params.  I'd note that `PreCommFillDerivedMesh` (which is called immediately after `PostInitialization`) shares the same function arguments as `PostInitialization`. 

Obviously there are arguments for both.  Happy to proceed on whatever front you all like best.

pdmullen:
> I get the impression that the discussion on the interface is not quite done yet, is it?

Modulo your recent comments regarding `*pin` as potential `PostInitialization` argument (see my response above), I was under the impression that everyone was on the same page.  @bprather does raise the point that maybe we should switch `ProblemGenerator` itself to take `MeshBlockData` as argument rather than `*pmb`, but that is outside the scope of *this* PR (i.e., that would be a breaking change).

doc/sphinx/src/README.rst
pdmullen: On it.

Yurlungur:
@pdmullen I just realized that I implemented but didn't document `UserWorkBeforeLoop` and `UserWorkBeforeLoopMesh` which are called after initialization but before the first cycle. Does that also meet the needs filled by this PR? 

Also my bad on not documenting it....

pgrete:
> @pdmullen I just realized that I implemented but didn't document `UserWorkBeforeLoop` and `UserWorkBeforeLoopMesh` which are called after initialization but before the first cycle. Does that also meet the needs filled by this PR?
> 
> Also my bad on not documenting it....

I assumed not because `UserWorkBeforeLoop` is currently called literally right before entering the loop, i.e., after `dt` etc are being calculated based on what was initialized.
Also `UserWorkBeforeLoop` can be used for restarts (which, IIRC, was one of the main motivations for your implementation) whereas `PostInit` is literally tightly related to the problem setup (and thus is only ever called once per simulation and also have an effect on calculating the initial `dt` etc).

src/application_input.hpp
pgrete: I interpreted above comments as "let's remain consistent for now and change things in one go for the `Data *` approach later/separately", which I'm also in favor of.
Given that this change may happen eventually, I don't feel super strong about the choice right now.

Regarding the `pin` I'm slight in favor of adding it as it's part of initialization (and only ever called once  whereas `PreCommFillDerivedMesh` is repeatedly called during the simulation).
Again, no super strong feeling so I suggest that you discuss this over lunch and then just go with what you decide.

pdmullen:
> > @pdmullen I just realized that I implemented but didn't document `UserWorkBeforeLoop` and `UserWorkBeforeLoopMesh` which are called after initialization but before the first cycle. Does that also meet the needs filled by this PR?
> > Also my bad on not documenting it....
> 
> I assumed not because `UserWorkBeforeLoop` is currently called literally right before entering the loop, i.e., after `dt` etc are being calculated based on what was initialized. Also `UserWorkBeforeLoop` can be used for restarts (which, IIRC, was one of the main motivations for your implementation) whereas `PostInit` is literally tightly related to the problem setup (and thus is only ever called once per simulation and also have an effect on calculating the initial `dt` etc).

This is correct.  `PostInitialization` is very much so tied to initialization (c.f., @pgrete's comments regarding restarts) and is carefully positioned after `ProblemGenerator` and before `PreCommFillDerived`.  

`UserWorkBeforeLoop` cannot handle all use cases.  Imagine a `ProblemGenerator` sets fields marked A, however, `PreCommFillDerived` resets a subset of fields A from fields B (but fields B haven't been set yet!!!).  Then you comm fields marked `FillGhost` then you enter `UserWorkBeforeLoop`.  There is no way to recover here...

TLDR: `PostInitialization` is very tightly integrated into the comms chain.

src/application_input.hpp
pdmullen: I think there is a nice opportunity to clean up some things regarding consistency of function arguments across initialization (and user-defined hooks in general); I suggest we convert that component into its own issue (I'll assign myself).  

In the meantime, if we keep everything as it stands now (without introducing `*pin` as an additional function argument), it saves users who want to repurpose existing functions from having to write something like

```c++
template <typename T>
void MyFillDerived(T *data) {
    return;
}

void MyPostInitialization(MeshBlockData<Real> *data, ParameterInput *pin) {
    MyFillDerived<MeshBlockData<Real>>(data);
    return;
}
```

but overall, the above isn't too bad, and perhaps more explicit... thoughts? @jonahm-LANL @bprather @pgrete

pgrete: > I think there is a nice opportunity to clean up some things regarding consistency of function arguments across initialization (and user-defined hooks in general); I suggest we convert that component into its own issue (I'll assign myself).

:+1: 

> In the meantime, if we keep everything as it stands now (without introducing `*pin` as an additional function argument), it saves users who want to repurpose existing functions from having to write something like
> 
> ```c++
> template <typename T>
> void MyFillDerived(T *data) {
>     return;
> }
> 
> void MyPostInitialization(MeshBlockData<Real> *data, ParameterInput *pin) {
>     MyFillDerived<MeshBlockData<Real>>(data);
>     return;
> }
> ```
> 
> but overall, the above isn't too bad, and perhaps more explicit... thoughts? @jonahm-LANL @bprather @pgrete

I'm still slightly in favor of the `pin` version (conceptually seeing this part of initialization), but happy to follow the majority vote.

pdmullen: Done. 

Here: https://github.com/parthenon-hpc-lab/parthenon/issues/1002
and here: https://github.com/parthenon-hpc-lab/parthenon/pull/999/commits/1ef878e0727e98483b71b29e52cd37809aadad44

pdmullen: After all, I just decided to opt for consistency everywhere and push all this concern to another issue/later date.  https://github.com/parthenon-hpc-lab/parthenon/pull/999/commits/927d67265dc66fca6dc652b6a568b1b5d5d803dc

Sorry for teeter-tottering.

PR #1009, opened 2024-02-22
Forest Of Octrees

### LINKED ISSUE #903: Refactor bvals code
The old bvals code relies on a somewhat complex inheritance structure and much of it is unused after changes to the boundary communication routines. We should go through the old code, remove what is unnecessary, and clean things up/shorten code where possible.

### LINKED ISSUE #912: Possibly remove `MeshBlockTree`
As I am working on including geometric multigrid in Parthenon (#911), I am finding more and more that I would like to move away from relying on `MeshBlockTree`. I could possibly remove `MeshBlockTree` entirely in that PR, but I wanted to get some feedback before I went down that (possibly drastic) path. 

The reasons that I can see for doing this are:
1. *The code would be substantially simplified* - If we move to storing the `LogicalLocation`s of the mesh in standard library containers (i.e. either a `std::map<LogicalLocation, ...>` or a `std::unordered_map<LogicalLocation, ...>`), then all of the `MeshBlockTree` code and most of the code in `bvals_base.cpp` is no longer necessary.  Generally, I have found the code in `bvals` to be some of the hardest to follow within Parthenon (see #903). With the recent changes to `LogicalLocation`, there are methods to find lists of possible neighbor blocks for a given block, determine the neighboring block pairs offsets, etc. which are required for building a `NeighborBlock` object, and other mesh location related utilities. As a result, it is very straightforward to find all of the neighbors of a given block by just finding a list of possible neighbors, and then seeing which of those neighbors are in the mesh. 
2. *It allows easy generalization to including multi-grid coarsened meshes in a `Mesh`* - Storing the levels of the mesh in a vector of `std::map`s makes it very easy to find parent blocks on the coarser mesh and daughter blocks on a finer mesh, as well as find neighbor blocks on the current mesh. This could of course be done if `MeshBlockTree` was extended, but this seems somewhat more complex to me. 
3. *It may make it easier to move to having a distributed mesh structure* - Currently, the structure of the mesh (i.e. the `MeshBlockTree`) is duplicated on all ranks. For very large numbers of ranks, this can result in the mesh structure taking up a significant fraction of each nodes memory. As a result, we may eventually want to move to having the mesh structure not be entirely duplicated on each rank. It seems easier to me to do this if the local mesh structure is stored in a `std::map` than if it is stored in a `MeshBlockTree` (although I admittedly haven't spent a huge amount of time thinking about how to do this).      

Possible drawbacks: 
1. The code is working fine using `MeshBlockTree` currently. 
2. I am not certain of the performance implications of moving away from a `MeshBlockTree` (although it seems likely that we can make a `std::unordered_map` perform faster than the current octree structure). 
3. The swarms code relies on inheriting from `BoundaryBase` and `BoundaryCommunication`, so there will be some refactoring that has to be done there.

lroberts36:
## PR Summary

This PR begins transitioning Parthenon domain from being represented by a single octree to a forest of octrees that cover the mesh domain. This PR switches to a new forest based mesh structure under the hood, but does not provide any user facing mechanism for building generic forests. Rather, it just builds a forest equivalent to the old root grid. Previously, if the base grid wasn't logically cubic parts of the octree domain were masked out. Now, multiple trees will cover the domain without masking. These changes should be transparent to downstream codes. For this limited set of forests, there is a clear way to map between forest logical locations and logical locations in the old Athena++ tree which allows writing the Athena++ locations out to file, building grids based on Athena++ style input, etc.   As a result, the behavior of the code and output should be completely unchanged and the changes made in this PR should have no impact on down stream codes (fingers crossed). Obviously, as-is these changes have limited utility but future PRs built on them should allow for more complex grids to be represented. 

This PR also removes a couple of big things from the code:
- `MeshblockTree` is replaced by `Forest` and `Tree` (the latter of which is based on a hash map of logical locations rather than the pointer tree used by `MeshblockTree`) 
- Most of the old boundary communication machinery (`BoundaryBase`, `BoundaryValue`, `BoundaryBuffer` and `BoundaryVariable`) and `pbval` from `MeshBlock`. There was already a lot of dead code in those classes and this PR required removing the old neighbor finding routines (since they relied on `Mesh::tree`). To make removing this stuff possible, some reasonably small changes to the swarms related communication was required. The changes and reorganization for this make up a lot of the lines of the PR.

This PR breaks multigrid, but fixing that would make the PR even longer (and no one is using it except me I think).

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


src/utils/morton_number.hpp
```diff
@@ -69,6 +69,10 @@ inline bool operator==(const MortonNumber &lhs, const MortonNumber &rhs) {
          (lhs.bits[0] == rhs.bits[0]);
 }
 
+inline bool operator!=(const MortonNumber &lhs, const MortonNumber &rhs) {
+  return !(lhs == rhs);
+}
```
Yurlungur, line 74: Surprised this wasn't autogenerated.

src/utils/indexer.hpp
```diff
@@ -22,6 +22,13 @@
 #include "utils/concepts_lite.hpp"
 
 namespace parthenon {
+template <typename tuple_t>
+constexpr auto get_array_from_tuple(tuple_t &&tuple) {
+  constexpr auto get_array = [](auto &&...x) {
+    return std::array{std::forward<decltype(x)>(x)...};
+  };
+  return std::apply(get_array, std::forward<tuple_t>(tuple));
+}
```
Yurlungur: Nice. We may want to move this to some shared variadic utils or something.

src/outputs/parthenon_hdf5.cpp
```diff
@@ -90,8 +90,8 @@ void PHDF5Output::WriteOutputFileImpl(Mesh *pm, ParameterInput *pin, SimTime *tm
   auto const nx2 = out_jb.e - out_jb.s + 1;
   auto const nx3 = out_kb.e - out_kb.s + 1;
 
-  const int rootLevel = pm->GetRootLevel();
-  const int max_level = pm->GetCurrentLevel() - rootLevel;
+  const int rootLevel = pm->GetAthenaCompositeRootLevel();
```
Yurlungur: Why is this called `Athena`?

src/outputs/output_utils.cpp
```diff
@@ -119,21 +119,23 @@ std::vector<Real> ComputeXminBlocks(Mesh *pm) {
 std::vector<int64_t> ComputeLocs(Mesh *pm) {
   return FlattenBlockInfo<int64_t>(
       pm, 3, [=](MeshBlock *pmb, std::vector<int64_t> &locs, int &i) {
-        locs[i++] = pmb->loc.lx1();
-        locs[i++] = pmb->loc.lx2();
-        locs[i++] = pmb->loc.lx3();
+        auto loc = pmb->pmy_mesh->Forest().GetAthenaCompositeLocation(pmb->loc);
+        locs[i++] = loc.lx1();
+        locs[i++] = loc.lx2();
+        locs[i++] = loc.lx3();
```
Yurlungur: Will this stuff need to change more at some point?

src/mesh/meshblock_tree.cpp
Yurlungur, line 1: I am so *incredibly* happy to see this go. :+1: :+1: :+1:

src/mesh/mesh.hpp
```diff
@@ -100,12 +97,14 @@ class Mesh {
...
 
+  const forest::Forest &Forest() const { return forest; }
+
   // data
   bool modified;
   const bool is_restart;
   RegionSize mesh_size;
   RegionSize base_block_size;
-  BoundaryFlag mesh_bcs[BOUNDARY_NFACES];
+  std::array<BoundaryFlag, BOUNDARY_NFACES> mesh_bcs;
```
Yurlungur, line 107: I don't object, but why this change?

src/mesh/mesh.cpp
```diff
@@ -233,13 +231,13 @@ Mesh::Mesh(ParameterInput *pin, ApplicationInput *app_in, Packages_t &packages,
   // initialize user-enrollable functions
   default_pack_size_ = pin->GetOrAddInteger("parthenon/mesh", "pack_size", -1);
 
+  forest = forest::Forest::AthenaXX(mesh_size, block_size, mesh_bcs);
+  root_level = forest.root_level;
   // calculate the logical root level and maximum level
-  for (root_level = 0; (1 << root_level) < nbmax; root_level++) {
-  }
+  // for (root_level = 0; (1 << root_level) < nbmax; root_level++) {
+  //}
```
Yurlungur: should these lines be removed?

src/mesh/mesh.cpp
```diff
@@ -1192,6 +1193,10 @@ bool Mesh::SetBlockSizeAndBoundaries(LogicalLocation loc, RegionSize &block_size
 //        logical location loc
 
 RegionSize Mesh::GetBlockSize(const LogicalLocation &loc) const {
+  // TODO(LFR): Update this
+  if (loc.tree() >=
+      0) // Implies this is a location in a forest, not in the old Athena tree
```
Yurlungur: autoformatting did these lines dirty

src/defs.hpp
```diff
@@ -109,6 +112,9 @@ struct RegionSize {
 // io_wrapper.hpp, bvals.hpp, field_diffusion.hpp,
 // tasks.hpp, ???
 
+// identifiers for boundary conditions
+enum class BoundaryFlag { block = -1, undef, reflect, outflow, periodic, user };
```
Yurlungur, line 118: Don't we have this somewhere else? Or did you move it?

src/defs.hpp
```diff
@@ -87,6 +87,9 @@ struct RegionSize {
   int &nx(CoordinateDirection dir) { return nx_[dir - 1]; }
   const int &nx(CoordinateDirection dir) const { return nx_[dir - 1]; }
 
+  Real LogicalToActualPosition(Real u, CoordinateDirection dir) const {
+    return u * (xmax_[dir - 1] - xmin_[dir - 1]) + xmin_[dir - 1];
+  }
```
Yurlungur, line 94: I don't fully understand this---will this be extended when blocks can be oriented weirdly with each other?

src/bvals/bvals_base.cpp
Yurlungur, line 1: :+1: :+1: :+1:

src/bvals/comms/tag_map.cpp
```diff
@@ -35,13 +35,34 @@ TagMap::rank_pair_t TagMap::MakeChannelPair(const MeshBlock *pmb,
...
+      if constexpr (BOUND == BoundaryType::gmg_prolongate_recv)
+        return &(pmb->gmg_coarser_neighbors);
+      if constexpr (BOUND == BoundaryType::gmg_prolongate_recv)
+        return &(pmb->gmg_coarser_neighbors);
+      if constexpr (BOUND == BoundaryType::gmg_same)
+        return pmb->loc.level() == md->grid.logical_level
+                   ? &(pmb->gmg_same_neighbors)
+                   : &(pmb->gmg_composite_finer_neighbors);
+      return &(pmb->neighbors);
+    }();
```
Yurlungur, line 59: I'm confused. What is going on here? Are you creating an anonymous function and immediately calling it to generate a neighbor array? Why not use a more traditional constexpr control flow?

src/mesh/logical_location.hpp
```diff
@@ -69,26 +69,51 @@ class LogicalLocation { // aggregate and POD type
...
   const auto &lx1() const { return l_[0]; }
   const auto &lx2() const { return l_[1]; }
   const auto &lx3() const { return l_[2]; }
   const auto &level() const { return level_; }
   const auto &morton() const { return morton_; }
+  const auto &tree() const { return tree_idx_; }
+
+  // Check if this logical location is actually in the domain of the tree,
+  // possibly including a ghost halo around the tree
+  bool IsInTree(int nghost = 0) const {
```
Yurlungur, line 101: these are ghost *blocks* right? not cells?

src/mesh/logical_location.hpp
```diff
@@ -69,26 +69,51 @@ class LogicalLocation { // aggregate and POD type
...
+           (l_[2] >= low) && (l_[2] < up);
+  }
+
+  // Check if a LL is in the ghost halo of the tree it is associated with
+  bool IsInHalo(int nghost) const { return IsInTree(nghost) && !IsInTree(0); }
+
+  int NeighborTreeIndex() const {
+    int i1 = (l_[0] >= 0) - (l_[0] < (1LL << level())) + 1;
+    int i2 = (l_[1] >= 0) - (l_[1] < (1LL << level())) + 1;
+    int i3 = (l_[2] >= 0) - (l_[2] < (1LL << level())) + 1;
```
Yurlungur: We should really document somewhere the bit magic being used for the morton ordering. It's clearly the correct way to do things, but I think someone not familiar will get confused. Or at least we should point to a paper/reference

src/mesh/logical_location.cpp
```diff
@@ -293,6 +358,7 @@ std::unordered_set<LogicalLocation> LogicalLocation::GetPossibleNeighborsImpl(
   return unique_locs;
 }
 
+// TODO(LFR): Remove this
```
Yurlungur, line 361: what needs to be changed? Is this a holdover until more generic trees?

src/bvals/neighbor_block.hpp
```diff
@@ -0,0 +1,182 @@
...
+  BLOCK_BNDRY = -1,
+  BNDRY_UNDEF,
+  REFLECTING_BNDRY,
+  OUTFLOW_BNDRY,
+  USER_BNDRY,
+  PERIODIC_BNDRY,
+  POLAR_BNDRY,
+  POLAR_BNDRY_WEDGE
+};
+#endif
```
Yurlungur, line 81: Is now an opportunity to move this to the version we actually want and simplify this pile of backwards compatibility stuff?

src/bvals/neighbor_block.hpp
Yurlungur, line 1: Good to split this off into its own thing. I approve.

src/bvals/neighbor_block.cpp
```diff
@@ -0,0 +1,166 @@
...
+                    itargetid, fi1, fi2) {}
+
+NeighborBlock::NeighborBlock(Mesh *mesh, LogicalLocation loc, int rank, int gid, int lid,
+                             std::array<int, 3> offsets, NeighborConnect type, int bid,
+                             int target_id, int fi1, int fi2)
+    : snb{rank, loc.level(), lid, gid}, ni{offsets[0], offsets[1], offsets[2],
+                                           fi1,        fi2,        type},
+      bufid{bid}, eid{0}, targetid{target_id}, fid{BoundaryFace::undef}, loc{loc},
+      ownership(true), block_size(mesh->GetBlockSize(loc)) {
+  // TODO(LFR): Look and see if this stuff gets used anywhere
```
Yurlungur, line 117: :+1:

src/bvals/bvals.hpp
Yurlungur, line 1: Much simplified :+1:

src/mesh/forest.hpp
```diff
@@ -0,0 +1,346 @@
...
+    if (!IsFace()) return BoundaryFace::undef;
+    for (int dir = 0; dir < 3; ++dir) {
+      if (static_cast<int>(u[dir]))
+        return static_cast<BoundaryFace>((1 + static_cast<int>(u[dir])) / 2 + 2 * dir);
+    }
+    return BoundaryFace::undef; // Shouldn't get here
+  }
+
+  // Get the logical diretions that are tangent to this element
+  // (in cyclic order, XY, YZ, ZX, XYZ)
```
Yurlungur: I'm not sure I understand what this means. Faces and edges clearly have normal/tangent directions. But what about cells and nodes?

src/mesh/forest.hpp
```diff
@@ -0,0 +1,346 @@
...
+    orient.use_offset = true;
+    orient.offset = {0, 0, 0};
+    ptree->neighbors[13].insert({ptree, orient});
+    return ptree;
+  }
+
+  // Methods for modifying the tree
+  int AddMeshBlock(const LogicalLocation &loc, bool enforce_proper_nesting = true);
+  int Refine(const LogicalLocation &ref_loc, bool enforce_proper_nesting = true);
+  int Derefine(const LogicalLocation &ref_loc, bool enforce_proper_nesting = true);
```
Yurlungur: Are we now allowing proper nesting to break?

src/mesh/forest.cpp
```diff
@@ -0,0 +1,484 @@
...
+    if (dir_flip[dir]) {
+      l_out[abs(dir_connection[dir])] = nblock - 1 - l_in;
+    } else {
+      l_out[abs(dir_connection[dir])] = l_in;
+    }
+  }
+  return LogicalLocation(destination, loc_in.level(), l_out[0], l_out[1], l_out[2]);
+}
+
+LogicalLocation RelativeOrientation::TransformBack(const LogicalLocation &loc_in,
```
Yurlungur: did I miss something? I don't remember `RelativeOrientation` being defined in the header...

Yurlungur (changes requested):
:+1: I'm really happy to see this refactor. The forest machinery seems pretty clean as well. The thing that makes me most nervous is that we have no built in testing for the trees other than the regression testing.

IMO a lot of the new refactored machinery is more cleanly decoupled from the full infrastructure. I think this is a good time to introduce unit tests for the tree machinery. @lroberts36 do you think you could add that? That would also make me less nervous, as I struggled to follow a lot of the fiddly index/bit gymnastics here.

src/utils/morton_number.hpp
lroberts36: yeah, I was too.

src/utils/indexer.hpp
lroberts36: not sure what the best place is, but agreed this probably more belongs somewhere in utils.

src/outputs/parthenon_hdf5.cpp
lroberts36: For lack of a better idea, I called all of the old notions about the mesh `Athena*`.  I guess we can probably always define a root level though, so maybe it is better to just call this `GetRootLevel()` as before.

src/outputs/output_utils.cpp
lroberts36: Yes,  I think in the future it will just need to be 
```c++
auto &loc = pmb->loc;
locs[i++] = loc.tree();
locs[i++] = loc.lx1();
locs[i++] = loc.lx1();
locs[i++] = loc.lx1();
```
(along with changing the size of the `locs` array). This will of course break backward compatibility and we will have to output forest connectivity information somewhere.

src/outputs/parthenon_hdf5.cpp
Yurlungur: Or perhaps just `CompositeRootLevel`? Referencing the "composite" tree?

src/outputs/output_utils.cpp
Yurlungur: Cool. Maybe we add a TODO comment and call it good for now? (Just so I keep it in mind while plumbing?

src/utils/indexer.hpp
Yurlungur: could just stick it in `utils.hpp` if we can't think of somewhere better.

src/bvals/comms/tag_map.cpp
lroberts36: Yes, it creates a lambda and then calls it at that location. I tend to like that pattern more for more complex initializations. As you say, in this case I could do 
```c++
std::vector<NeighborBlock> *neighbors;
if constexpr(....) {
  neighbors = &(pmb->gmg_coarser_neighbors);
} else if (...) {
...
```
but I guess I prefer the pattern that is in there now since it minimizes the number of times an object is initialized.

src/bvals/neighbor_block.hpp
lroberts36: Yeah, I cannot count the number minutes I have wasted searching for where this code is.

src/bvals/neighbor_block.hpp
lroberts36: I thought about this but decided against doing anything because I really wasn't sure how important this backwards compatibility was to downstream codes.

src/bvals/comms/tag_map.cpp
Yurlungur: I find the lambda version confusing, but I won't fight you on it. I was able to disentangle what's going on.

src/defs.hpp
lroberts36: I moved it, I think it was in one of the deleted files.

src/mesh/forest.hpp
lroberts36: Cells have no normal directions, faces have one, edges have two, and nodes have three. Cells have three tangent directions, faces have two, edges have one, and nodes have zero.

src/mesh/forest.hpp
lroberts36: No (at least not for now). These mirror the functionality that was present in `MeshBlockTree` where there were some routines to add blocks that didn't enforce proper nesting. These are used on restart to repopulate a tree that was presumably already properly nested.

src/mesh/logical_location.cpp
lroberts36: Any routines in `LogicalLocation` that are duplicated (but with a name containing `Forest` at the end) won't be used and can be removed once I update multigrid.

src/mesh/logical_location.hpp
lroberts36: yep, ghost blocks.

src/mesh/logical_location.hpp
lroberts36: fair enough. We need to write up documentation about `LogicalLocation`s, trees, and the forest. We can probably enumerate all the bit tricks pretty quickly, e.g. this is just a quick way to get 2^level = # of blocks per side of the tree domain at this level.

src/mesh/mesh.cpp
lroberts36: yeah, will move the comment to a separate line.

src/outputs/parthenon_hdf5.cpp
lroberts36: Ah, I actually wasn't thinking about this correctly before. I think this root level is unique to the Athena style grid.

src/bvals/comms/tag_map.cpp
lroberts36: yeah, obviously it doesn't really buy you anything for pointer initialization but I think it is a useful pattern since default construction of other types of objects can be expensive. It is also a useful pattern for use in class initializer lists.

src/mesh/forest.cpp
lroberts36: It's hidden in the header just above the declaration of `Tree`. Maybe there are getting to be enough classes that this header should be split into multiple files.

src/bvals/comms/tag_map.cpp
brryan: Just to chime in, this syntax would also be confusing to me (since I can see this discussion I now know what's going on) but it might be worthwhile to just add a comment saying `define and use lambda to reduce initializations while setting neighbors` so people like me don't have to wonder whether there is something more or different going on.

brryan:
I don't have any objections to the swarm boundary comms refactor (thank you for working this out @lroberts36!) -- the only thing I would like to do is run the parthenon `particle_leapfrog` example for a large number of meshblocks and see that it still works. Right now this branch isn't building for me on `volta-x86`:
```
"/home/brryan/github/parthenon/example/sparse_advection/sparse_advection_driver.cpp", line 125: error: expected an identifier
      auto restrict =
                    ^

"/home/brryan/github/parthenon/example/sparse_advection/sparse_advection_driver.cpp", line 130: error: expected an expression
        tl.AddTask(restrict, SparseDealloc, mc1.get());
```
but once it is working again I'm happy to check that example.

src/bvals/neighbor_block.hpp
Yurlungur: I *think* it's completely unused. Though I could be wrong about that. @pgrete maybe can chime in.

src/mesh/logical_location.hpp
Yurlungur: Might want to add a comment to that effect.

src/outputs/parthenon_hdf5.cpp
Yurlungur: What about the `Athena` style grid is unique here?

src/bvals/comms/tag_map.cpp
Yurlungur: I would be happy with a comment to that effect.

src/mesh/forest.cpp
Yurlungur: I would be in favor of splitting them out.

src/mesh/mesh.hpp
lroberts36: Because I needed to pass it to a function and I think passing c-style arrays is a bit ugly.

src/mesh/mesh.cpp
lroberts36: removed.

src/mesh/mesh.cpp
lroberts36: fixed

src/defs.hpp
lroberts36: Yes, I we will have to extend this, although I am not entirely sure what the best approach is. For right now, this works for the types of grids supported by this PR and allows for easy construction of `RegionSize`s for blocks a tree region size and trees region sizes from a forest region size.

src/mesh/logical_location.hpp
lroberts36: `ghost` -> `ghost block`

src/outputs/parthenon_hdf5.cpp
lroberts36: Per our conversation, I am going to leave this as `Athena` since it is consistent with the levels of the `Athena` logical locations that are used for output.

src/utils/indexer.hpp
lroberts36: moved to utils.hpp

lroberts36:
> I don't have any objections to the swarm boundary comms refactor (thank you for working this out @lroberts36!) -- the only thing I would like to do is run the parthenon `particle_leapfrog` example for a large number of meshblocks and see that it still works. Right now this branch isn't building for me on `volta-x86`:
> 
> ```
> "/home/brryan/github/parthenon/example/sparse_advection/sparse_advection_driver.cpp", line 125: error: expected an identifier
>       auto restrict =
>                     ^
> 
> "/home/brryan/github/parthenon/example/sparse_advection/sparse_advection_driver.cpp", line 130: error: expected an expression
>         tl.AddTask(restrict, SparseDealloc, mc1.get());
> ```
> 
> but once it is working again I'm happy to check that example.

@brryan: I think I just fixed the issue you were seeing (which I don't think came from this PR?). One of the tasks in the sparse advection example was called `restrict`, which is not a c++ keyword as of c++14 but was until then (https://stackoverflow.com/questions/776283/what-does-the-restrict-keyword-mean-in-c). Not sure why this cropped up now...

src/bvals/comms/tag_map.cpp
lroberts36: comment added.

src/mesh/forest.cpp
lroberts36: reorganized.

lroberts36:
> 👍 I'm really happy to see this refactor. The forest machinery seems pretty clean as well. The thing that makes me most nervous is that we have no built in testing for the trees other than the regression testing.
> 
> IMO a lot of the new refactored machinery is more cleanly decoupled from the full infrastructure. I think this is a good time to introduce unit tests for the tree machinery. @lroberts36 do you think you could add that? That would also make me less nervous, as I struggled to follow a lot of the fiddly index/bit gymnastics here.

@Yurlungur: I think this is ready for re-review. In addition to addressing your other comments, I added a unit test for refining and de-refining a two block forest. That being said, I think the regression tests that use AMR (in particular sparse advection) are more robust tests of the new forest implementation as they require reproducing the exact same mesh structure as the the old tree code at all steps (as the gold files are built on that).

src/mesh/forest/forest.hpp
```diff
@@ -0,0 +1,151 @@
...
+  std::size_t CountMeshBlock() const {
+    std::size_t count{0};
+    for (auto &[id, tree] : trees)
+      count += tree->CountMeshBlock();
+    return count;
+  }
+
+  // TODO(LFR): Probably eventually remove this. This is only meaningful for simply
+  // oriented grids
+  LogicalLocation GetAthenaCompositeLocation(const LogicalLocation &loc) const {
```
Yurlungur: I thought we came up with some name we liked better than Athena? UniTree?

src/mesh/forest/tree.cpp
```diff
@@ -0,0 +1,289 @@
...
+#include "mesh/forest/relative_orientation.hpp"
+#include "mesh/forest/tree.hpp"
+#include "mesh/logical_location.hpp"
+#include "utils/bit_hacks.hpp"
+#include "utils/indexer.hpp"
+
+namespace parthenon {
+namespace forest {
+
+Tree::Tree(Tree::private_t, std::int64_t id, int ndim, int root_level, RegionSize domain,
```
Yurlungur, line 38: What's going on with the `Tree::private_t` first argument? I don't think I've seen that before.

src/utils/indexer.hpp
```diff
@@ -51,6 +52,12 @@ struct Indexer {
     return GetIndicesImpl(idx, std::make_index_sequence<sizeof...(Ts)>());
   }
 
+  KOKKOS_FORCEINLINE_FUNCTION
+  auto GetIdxArray(int idx) const {
+    return get_array_from_tuple(
+        GetIndicesImpl(idx, std::make_index_sequence<sizeof...(Ts)>()));
+  }
```
Yurlungur, line 59: :+1:

Yurlungur (approved):
Thanks for adding the test! I think even from a documentation perspective seeing some usage in the unit test is helpful. LGTM.

src/mesh/forest/tree.cpp
lroberts36: This enforces that `Tree`s can only be created as `shared_ptr`s via `Tree::create`. If the constructor is just made `private` or `protected`, then the call to `std::make_shared<Tree>` in `Tree::create` fails because the constructor is not accessible. This pattern makes the constructor effectively private while still allowing the use of `make_shared` within `static` member functions. (full disclosure: I found this approach on stack overflow and I think it is fairly standard.)

Yurlungur: Ah ok. Thanks for the explanation!

src/mesh/forest/forest.hpp
lroberts36: Yggdrasil?

Yurlungur: lol. I'm in.

pgrete:
I'll get to review a bit later this week (as I also want to run some [performance] tests downstream)

lroberts36:
This PR passes all Riot tests.

src/bvals/comms/bnd_info.cpp
```diff
@@ -408,12 +408,8 @@ ProResInfo ProResInfo::GetSet(MeshBlock *pmb, const NeighborBlock &nb,
...
   bool restricted = false;
   if (mylevel > 0) {
-    for (int k = 0; k < 3; ++k) {
-      for (int j = 0; j < 3; ++j) {
-        for (int i = 0; i < 3; ++i) {
-          restricted = restricted || (pmb->pbval->nblevel[k][j][i] == (mylevel - 1));
-        }
-      }
+    for (const auto &nb : pmb->neighbors) {
+      restricted = restricted || (nb.loc.level() == (mylevel - 1));
```
pgrete, line 412: Looks much nicer!

src/bvals/neighbor_block.hpp
pgrete: Just checked and we don't use it downstream, so this can go.

src/defs.hpp
```diff
@@ -87,6 +87,9 @@ struct RegionSize {
   int &nx(CoordinateDirection dir) { return nx_[dir - 1]; }
   const int &nx(CoordinateDirection dir) const { return nx_[dir - 1]; }
 
+  Real LogicalToActualPosition(Real u, CoordinateDirection dir) const {
```
pgrete, line 92: Can we get a one line comment what this function does, i.e., what the parameter `u` is (especially as it's a `Real` not  an `int`, which is what I'd have assumed from the function name)?

src/mesh/forest/cell_center_offsets.hpp
```diff
@@ -0,0 +1,139 @@
...
+namespace forest {
+enum class Direction : uint { I = 0, J = 1, K = 2 };
+
+// CellCentOffsets defines the position of a topological element
+// within a cell or a neighboring cell via offsets from the cell center. The center of
+// cell is defined by zero offsets in each direction. The faces have
+// a Offset::Low or Offset::Up in one direction and Offset::Middle
+// in the others, etc. The topological position of an element in the
+// is turned into an unsigned index 0...27 via
+// (x_offset + 1) + 3 * (y_offset + 1) + 9 * (z_offset + 1)
```
pgrete: I'm wondering to which degree this is tied to the "forest" or whether this is more general (from where the code should live point of view).

src/mesh/forest/cell_center_offsets.hpp
```diff
@@ -0,0 +1,139 @@
...
+
+// TODO(LFR): Consider switching this to C-style enum within a namespace to avoid
+// static_cast
+enum class Offset : int { Low = -1, Middle = 0, Up = 1 };
```
pgrete: Further up in the bvals, the ordering of offset was 0, -1, 1 IIRC.
I realize these are enums here, so it shouldn't matter, but I'm wondering if there any danger for confusion mid/long term wrt to ordering convention.

src/mesh/forest/cell_center_offsets.hpp
```diff
@@ -0,0 +1,139 @@
...
+    return {static_cast<int>(u[0]), static_cast<int>(u[1]), static_cast<int>(u[2])};
+  }
+
+  BoundaryFace Face() const {
+    if (!IsFace()) return BoundaryFace::undef;
+    for (int dir = 0; dir < 3; ++dir) {
+      if (static_cast<int>(u[dir]))
+        return static_cast<BoundaryFace>((1 + static_cast<int>(u[dir])) / 2 + 2 * dir);
+    }
+    return BoundaryFace::undef; // Shouldn't get here
```
pgrete: Why not a `PARTHENON_FAIL` then?

src/mesh/forest/forest.cpp
```diff
@@ -0,0 +1,182 @@
...
+    std::size_t end = mb_list.size();
+    for (int i = start; i < end; ++i)
+      tree->InsertGid(mb_list[i], gid++);
+  }
+  // The index of blocks in this list corresponds to their gid
+  gids_resolved = true;
+  return mb_list;
+}
+
+Forest Forest::AthenaXX(RegionSize mesh_size, RegionSize block_size,
```
pgrete: That's the legacy Tree, isn't it?

src/mesh/forest/tree.cpp
pgrete: Might be worth adding this explanation to the code as comment (also to educate other devs like me about this pattern when reading the code again at a later point ;) )

src/mesh/forest/tree.cpp
```diff
@@ -0,0 +1,289 @@
...
+                          {ndim > 1 ? -1 : 0, ndim > 1 ? 1 : 0},
+                          {ndim > 2 ? -1 : 0, ndim > 2 ? 1 : 0});
+  std::vector<NeighborLocation> neighbor_locs;
+  for (int o = 0; o < offsets.size(); ++o) {
+    auto [ox1, ox2, ox3] = offsets(o);
+    if (std::abs(ox1) + std::abs(ox2) + std::abs(ox3) == 0) continue;
+    FindNeighborsImpl(loc, ox1, ox2, ox3, &neighbor_locs);
+  }
+
+  const int clev = loc.level() - 1;
```
pgrete: Where is this used (given the following return statement)?

src/mesh/forest/tree.hpp
```diff
@@ -0,0 +1,131 @@
...
+#include "utils/bit_hacks.hpp"
+#include "utils/indexer.hpp"
+
+namespace parthenon {
+namespace forest {
+
+// We don't explicitly allow for periodic boundaries, since we can encode periodicity
+// through connectivity in the forest
+class Tree : public std::enable_shared_from_this<Tree> {
+  // This allows us to ensure that Trees are only created as shared_ptrs
```
pgrete, line 40: Never mind the comment above, the explanation is already here.
When reviewing I should probably read the header files first and then the impl.

src/mesh/logical_location.hpp
```diff
@@ -122,19 +154,20 @@ class LogicalLocation { // aggregate and POD type
...
   GetSameLevelNeighbor(int ox1, int ox2, int ox3,
                        const RootGridInfo &rg_info = RootGridInfo()) const {
-    return LogicalLocation(level(), lx1() + ox1, lx2() + ox2, lx3() + ox3);
+    return LogicalLocation(tree(), level(), lx1() + ox1, lx2() + ox2, lx3() + ox3);
   }
 
-  LogicalLocation GetParent() const {
-    if (level() <= 0) return LogicalLocation(level() - 1, 0, 0, 0);
-    return LogicalLocation(level() - 1, lx1() >> 1, lx2() >> 1, lx3() >> 1);
+  LogicalLocation GetParent(int nlevel = 1) const {
```
pgrete, line 160: nlevel is the offset wrt to number of levels for the parent, i.e., nlevel=2 would give the grandparent?

tst/unit/test_forest.cpp
```diff
@@ -0,0 +1,106 @@
+//========================================================================================
+// Parthenon performance portable AMR framework
+// Copyright(C) 2022 The Parthenon collaboration
+// Licensed under the 3-clause BSD License, see LICENSE file for details
+//========================================================================================
+// (C) (or copyright) 2021-2022. Triad National Security, LLC. All rights reserved.
```
pgrete: Years?

pgrete (approved):
LGTM -- and also nice cleanup of the code and thanks for the unit tests (which are way more readable than the integrated version of construction in the Mesh class.

I also ran some downstream tests and didn't notice anything odd.
In fact, the AMR performance increased by up to 30%-50% (based on the `wsec_AMR` output and for a test with several hundreds of blocks per device).

src/mesh/forest/forest.cpp
lroberts36: Yeah, this should build a forest that is equivalent to the single masked tree that Athena used for non-cubic base grids. (Equivalent in the sense that the block list retrieved for a grid built with this routine should be identical to that retrieved from the old tree code).

src/mesh/logical_location.hpp
lroberts36: yep, that is right.

src/mesh/forest/cell_center_offsets.hpp
lroberts36: I am not sure I see where the ordering is different?

src/mesh/forest/cell_center_offsets.hpp
lroberts36: added.

src/mesh/forest/tree.cpp
lroberts36: that is dead code, removed.

tst/unit/test_forest.cpp
lroberts36: fixed.

src/defs.hpp
lroberts36: added. This just moves from a block local coordinate system running from 0 to 1 to the global coordinate system.

src/mesh/forest/cell_center_offsets.hpp
lroberts36: Yeah, this is probably a little more general than forests and could be used in a few other places. I have moved it to utils. At some point we can maybe thread it through other places in the code (e.g. where `ox1` etc show up).

src/mesh/forest/forest.cpp
lroberts36: I renamed this `Forest::HyperRectangular` for clarity.

src/mesh/forest/forest.hpp
lroberts36: renamed to `GetLegacyTreeLocation` (and similarly for other routines).

brryan:
> I don't have any objections to the swarm boundary comms refactor (thank you for working this out @lroberts36!) -- the only thing I would like to do is run the parthenon `particle_leapfrog` example for a large number of meshblocks and see that it still works. Right now this branch isn't building for me on `volta-x86`:
> 
> ```
> "/home/brryan/github/parthenon/example/sparse_advection/sparse_advection_driver.cpp", line 125: error: expected an identifier
>       auto restrict =
>                     ^
> 
> "/home/brryan/github/parthenon/example/sparse_advection/sparse_advection_driver.cpp", line 130: error: expected an expression
>         tl.AddTask(restrict, SparseDealloc, mc1.get());
> ```
> 
> but once it is working again I'm happy to check that example.

Just tested on `particle_leapfrog` on 32 nodes and it runs and gives the right particle positions at final time, so everything looks good to me.

brryan (approved):
This approval is mainly for the modifications to particles. Thanks for dealing with that @lroberts36

PR #1010, opened 2024-02-29
WIP: Store pmb->gid in VariableState

pdmullen:
## PR Summary

In addition to `sparse_id`, it would be also nice to be able to access the current block `gid` in a `SparsePack` via

```c++
pack(b, var()).gid
```
This utility is added in this MR, by adding `gid` to `VariableState`.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/src/interface/variable.cpp
+++ b/src/interface/variable.cpp
@@ -39,6 +39,8 @@ Variable<T>::Variable(const std::string &base_name, const Metadata &metadata,
                            "Mismatch between sparse flag and sparse ID");
   uid_ = get_uid_(label());
 
+  pmb_gid_ = wpmb.lock().get()->gid;
+
   if (m_.getAssociated() == "") {
     m_.Associate(label());
   }
--- a/src/interface/variable.cpp
+++ b/src/interface/variable.cpp
@@ -1,5 +1,5 @@
 //========================================================================================
-// (C) (or copyright) 2020-2023. Triad National Security, LLC. All rights reserved.
+// (C) (or copyright) 2020-2024. Triad National Security, LLC. All rights reserved.
 //
 // This program was produced under U.S. Government contract 89233218CNA000001 for Los
 // Alamos National Laboratory (LANL), which is operated by Triad National Security, LLC
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Current develop
 
 ### Added (new features/APIs/variables/...)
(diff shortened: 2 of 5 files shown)


Yurlungur (approved):
This is a nice solution... Better than putting it directly into the pack. :+1:

pgrete (approved):

pdmullen:
> This is a nice solution... Better than putting it directly into the pack. 👍

Credit to @lroberts36

src/interface/variable.cpp
```diff
@@ -39,6 +39,8 @@ Variable<T>::Variable(const std::string &base_name, const Metadata &metadata,
                            "Mismatch between sparse flag and sparse ID");
   uid_ = get_uid_(label());
 
+  pmb_gid_ = wpmb.lock().get()->gid;
+
```
lroberts36, line 43: It is possible that after a re-mesh when amr is turned on a block can have its `gid` changed. I don't think there is any mechanism to propagate this through to the variables, so I think that the `VariableState` could end up with a stale `gid` in this case. 

The two possible solutions I think of. First, maybe adding a method to `MeshBlock` called `UpdateGid(int gid)` which updates the block `gid` and the `VariableState` of all variables (but I guess this depends on iterating through all of the containers). This method would need to be called in amr_loadbalance.cpp instead of just setting the block `gid`s by hand. The other possibility is to just make sure that the values are updated during sparse packing, which is a little less clean I think but probably easier to implement.

lroberts36 (changes requested):
See my comment below.

src/interface/variable.cpp
pgrete: Just double checking: are our gids always guaranteed to be contiguous  (starting from 0)?
We have some code paths that implicitly assume that there's always a block with gid 0.

lroberts36: Yes, `gid`s are defined to always start at zero and be contiguous (i.e. they are just the indices of all the blocks in a std::vector that has been sorted by the Morton number of the blocks' logical locations). The issue is that after remeshing, the Morton number of a block that hasn't been refined or derefined doesn't change but it's `gid` can change since blocks with lower Morton numbers may have been added/removed from the grid.

brryan:
@pdmullen can this PR be closed now that #1167 is merged?

pdmullen:
> @pdmullen can this PR be closed now that #1167 is merged?

Yes.  thanks for the reminder!

Issue #515, opened 2021-05-11
Potential inconsistency when reading writing hdf strings


pgrete:
We recently encountered an interesting bug.

A regression test (run in the context of #511 ) consistently failed with 
```
4:     dataset[:,:,:,:,idx_component]
4: TypeError: 'NoneType' object is not subscriptable
```

We were able to isolate the issue that showed up on Python 3.8.x but not on Python 3.7.x after realizing that the default `h5py` version shipped in those environments made a major jump, e.g.,
```bash
$ python
Python 3.7.4 (default, Dec 22 2020, 12:15:41) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import h5py
>>> print(h5py.__file__)
/mnt/home/gretephi/.local/lib/python3.7/site-packages/h5py/__init__.py
>>> print(h5py.__version__)
2.10.0
```
versus
```bash
$ python
Python 3.8.2 (default, Feb 16 2021, 21:08:08) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import h5py
>>> print(h5py.__file__)
/mnt/home/gretephi/.local/lib/python3.8/site-packages/h5py/__init__.py
>>> print(h5py.__version__)
3.1.0
```

The bug showed up because the name comparison to identify a variable in the dataset failed as the names were read as byte array
```python
>>> datafile.Info["VariableNames"]
array([b'Density', b'MomentumDensity1', b'MomentumDensity2',
       b'MomentumDensity3', b'TotalEnergyDensity', b'Density',
       b'Velocity1', b'Velocity2', b'Velocity3', b'Pressure'],
      dtype=object)
```
which was a major change in `h5py` 3.x.x:
"_Variable-length UTF-8 strings in datasets are now read as bytes objects instead of str_", see https://docs.h5py.org/en/latest/whatsnew/3.0.html#breaking-changes-deprecations

We were able to fix the issue by  explicitly converting to string, e.g.,
```python
component_names = datafile.Info["VariableNames"].astype(str)
```

Given that the fix works I haven't looked into how we actually write the attributes and whether changing the encoding (when writing from Parthenon or reading with `phdf`) may potentially solve this automatically and/or make this more(?) consistent, see some additional details on string encoding here: https://docs.h5py.org/en/stable/strings.html#encodings

[...]

PR #1019, opened 2024-03-11
[breaking] output more topological elements

Yurlungur:
## PR Summary

What it says on the can

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


c-prather:
So many feature PRs with net negative lines! Approach looks great, can test/debug later this week.  This is a super important piece of putting KHARMA back on Parthenon develop. I'll have another interesting little piece to push upstream in the next week too, and then I can stop frankenbranching!  Exciting.

Yurlungur:
I'm not done yet. This is currently just cleanup. But the feature will be in soon. Hopefully before you look. :)

pgrete (commented):
Just had quick look at the PR following our discussion on Thursday and it looks like a nice cleanup (though I didn't review the code in great detail yet).
Note, that I'm about to make the `RestartReader` a base class so that I can derive for hdf5 and OpenPMD.
Do you expect to touch the restart parts (beyond the removed legacy code)?

Yurlungur:
> Just had quick look at the PR following our discussion on Thursday and it looks like a nice cleanup (though I didn't review the code in great detail yet). Note, that I'm about to make the `RestartReader` a base class so that I can derive for hdf5 and OpenPMD. Do you expect to touch the restart parts (beyond the removed legacy code)?

I will need to touch the utility functions called by the restart reader, such as the pack/unpack variables. I will also need to change `ReadBlocks` in `restart.hpp`.

pgrete:
> > Just had quick look at the PR following our discussion on Thursday and it looks like a nice cleanup (though I didn't review the code in great detail yet). Note, that I'm about to make the `RestartReader` a base class so that I can derive for hdf5 and OpenPMD. Do you expect to touch the restart parts (beyond the removed legacy code)?
> 
> I will need to touch the utility functions called by the restart reader, such as the pack/unpack variables. I will also need to change `ReadBlocks` in `restart.hpp`.

Right, but this is all going to be "internal" to restarting from hdf5, correct?
The OpenPMD logic will a little bit different, so I'm mostly interested if you expect changes to the interface that's being called (not what happens within the functions themselves).

Yurlungur:
> > > Just had quick look at the PR following our discussion on Thursday and it looks like a nice cleanup (though I didn't review the code in great detail yet). Note, that I'm about to make the `RestartReader` a base class so that I can derive for hdf5 and OpenPMD. Do you expect to touch the restart parts (beyond the removed legacy code)?
> > 
> > 
> > I will need to touch the utility functions called by the restart reader, such as the pack/unpack variables. I will also need to change `ReadBlocks` in `restart.hpp`.
> 
> Right, but this is all going to be "internal" to restarting from hdf5, correct? The OpenPMD logic will a little bit different, so I'm mostly interested if you expect changes to the interface that's being called (not what happens within the functions themselves).

Ah I see... I'm not sure. I was thinking of trying to loop the `VarInfo` through *reading* restarts too, to re-use that code... but it might be more trouble than it's worth.

pgrete:
> > > > Just had quick look at the PR following our discussion on Thursday and it looks like a nice cleanup (though I didn't review the code in great detail yet). Note, that I'm about to make the `RestartReader` a base class so that I can derive for hdf5 and OpenPMD. Do you expect to touch the restart parts (beyond the removed legacy code)?
> > > 
> > > 
> > > I will need to touch the utility functions called by the restart reader, such as the pack/unpack variables. I will also need to change `ReadBlocks` in `restart.hpp`.
> > 
> > 
> > Right, but this is all going to be "internal" to restarting from hdf5, correct? The OpenPMD logic will a little bit different, so I'm mostly interested if you expect changes to the interface that's being called (not what happens within the functions themselves).
> 
> Ah I see... I'm not sure. I was thinking of trying to loop the `VarInfo` through _reading_ restarts too, to re-use that code... but it might be more trouble than it's worth.

Alright, as far as I can tell right now my refactor won't be that intrusive, so I'll just go ahead and then we can mix/match later (especially as I'll probably use more from the change you implement here than the other way around).

Yurlungur:
This branch now depends on #1027

Yurlungur:
@pgrete @pdmullen @bprather @lroberts36 @jdolence this is ready for friendly testing and review. The following are still TODOs before merge:
- [x] Test on a realistic problem. I will be testing on partheon-MHD

The following needs to happen at some point, but I might do in a separate PR:
- [x] Update XDMF to support topological elements
- [x] Update `phdf.py` and maybe `movie2d.py` to support topological elements
- [ ] Add a regression test for topological element I/O. Talked to @lroberts36 about this, and I might try to make a vector wave equation example.

Curious on your thoughts. This turned out to be a *much* bigger lift than I expected, due to the large number of fiddly index gymnastics which were all hardcoded throughout.

Yurlungur:
Note that this depends on #1027 so the diff will be cleaner after that change goes in.

src/interface/variable.hpp
```diff
@@ -81,14 +81,19 @@ class Variable {
   KOKKOS_FORCEINLINE_FUNCTION
   auto GetDim(const int i) const {
     // we can't query data.GetDim() here because data may be unallocated
-    assert(0 < i && i <= MAX_VARIABLE_DIMENSION && "ParArrayNDs are max 6D");
+    PARTHENON_DEBUG_REQUIRE(0 < i && i <= MAX_VARIABLE_DIMENSION, "Index out of bounds");
```
Yurlungur, line 84: The error message here was previously incorrect.

src/outputs/histogram.cpp
```diff
@@ -472,12 +472,13 @@ void HistogramOutput::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm
...
-    const std::array<hsize_t, H5_NDIM> local_offset({0, 0, 0, 0, 0, 0, 0});
-    std::array<hsize_t, H5_NDIM> local_count({0, 0, 0, 0, 0, 0, 0});
-    std::array<hsize_t, H5_NDIM> global_count({0, 0, 0, 0, 0, 0, 0});
+    // As we're reusing the interface from the existing hdf5 output,
+    // we have to define everything as H5_NDIM arrays. Counts will be
+    // set for each histogram individually below. All
+    // zero-initialized
+    const std::array<hsize_t, H5_NDIM> local_offset = {0};
+    std::array<hsize_t, H5_NDIM> local_count = {0};
+    std::array<hsize_t, H5_NDIM> global_count = {0};
```
Yurlungur, line 481: Technically these arrays are zero-initialized no matter what I set in the initializer list. However, initializer lists initialize all the elements up to the length of the list, so if you have an array of dimension that changes, an initializer list of fixed size can be misleading. I set the first element of the array to zero just for clarity.

src/outputs/output_utils.hpp
```diff
@@ -41,31 +46,89 @@ namespace parthenon {
...
+  // Includes topological element shape
+  int TensorSize() const;
+  // Size of region that needs to be filled with 0s if not allocated
+  int FillSize(const IndexDomain domain) const;
+  // number of elements of data that describe variable shape
+  int GetNDim() const;
+
+  template <typename T>
+  int FillShape(const IndexDomain domain, T *data) const {
+    if (where == MetadataFlag({Metadata::None})) {
```
Yurlungur, line 81: Broadly, `Metadata::None` is a special snowflake everywhere. I couldn't figure out how to not special-case it. But most of the special-casing is hidden.

src/outputs/output_utils.hpp
```diff
@@ -96,11 +159,22 @@ struct VarInfo {
...
+  static std::vector<VarInfo> GetAll(const VariableVector<Real> &vars,
+                                     const IndexShape &cellbounds);
+
+  bool operator==(const std::string &other) const { return other == label; }
+
+ private:
+  // TODO(JMM): Probably nx_ and rnx_ both not necessary... but it was
+  // easiest for me to reason about it this way.
+  std::array<int, VNDIM> nx_;
+  std::vector<int> rnx_;
```
Yurlungur, line 181: This is just the limitations of my ability to reason about index gymnastics. I think this could be simplified but I'm not sure it's worthwhile to do so.

tst/unit/test_output_utils.cpp
Yurlungur, line 1: Delta lines would be much closer to zero or negative, except I added tests.

src/interface/variable.hpp
```diff
@@ -212,7 +217,7 @@ class ParticleVariable {
 
   KOKKOS_FORCEINLINE_FUNCTION
   auto GetDim(const int i) const {
-    PARTHENON_DEBUG_REQUIRE(0 < i && i <= 6, "ParArrayNDGenerics are max 6D");
+    PARTHENON_DEBUG_REQUIRE(0 < i && i <= 6, "Index out of bounds");
```
lroberts36, line 217: Should this be `MAX_VARIABLE_DIMENSION` instead of 6?

src/interface/variable.hpp
```diff
@@ -81,14 +81,19 @@ class Variable {
   KOKKOS_FORCEINLINE_FUNCTION
   auto GetDim(const int i) const {
     // we can't query data.GetDim() here because data may be unallocated
-    assert(0 < i && i <= MAX_VARIABLE_DIMENSION && "ParArrayNDs are max 6D");
+    PARTHENON_DEBUG_REQUIRE(0 < i && i <= MAX_VARIABLE_DIMENSION, "Index out of bounds");
     return dims_[i - 1];
   }
 
+  KOKKOS_FORCEINLINE_FUNCTION
+  auto GetDim() const { // TODO(JMM): should this be host-only?
```
lroberts36: It seems like this should be host only. I didn't think `Variable` was meant to be used on device.

src/outputs/output_utils.cpp
```diff
@@ -33,6 +33,105 @@
 namespace parthenon {
 namespace OutputUtils {
 
+Triple_t<int> VarInfo::GetNKJI(const IndexDomain domain) const {
```
lroberts36: FWIW, I initially read this to mean that it should be returning four indices n, k, j, i.

src/outputs/output_utils.cpp
```diff
@@ -33,6 +33,105 @@
...
+  int nx3 = 1, nx2 = 1, nx1 = 1;
+  // TODO(JMM): I know that this could be done by hand, but I'd rather
+  // rely on the loop bounds machinery and this should be cheap.
```
lroberts36, line 39: I support this.

src/outputs/output_utils.cpp
```diff
@@ -33,6 +33,105 @@
...
+}
+
+// Includes topological element shape
+int VarInfo::TensorSize() const {
+  if (where == MetadataFlag({Metadata::None})) {
+    return Size();
+  } else {
+    return std::accumulate(rnx_.begin(), rnx_.end() - 3, 1, std::multiplies<int>());
+  }
+}
```
lroberts36: Maybe it is overly pedantic, but it would be clearer to have this return the actual tensor shape (i.e. t, u, v) and have a separate function for `TopologicalElementSize`. Haven't seen how many places `TensorSize` shows up though.

src/outputs/output_utils.hpp
```diff
@@ -41,31 +46,89 @@ namespace parthenon {
...
                    // variables
   MetadataFlag where;
   bool is_sparse;
   bool is_vector;
+  IndexShape cellbounds;
   std::vector<std::string> component_labels;
-  int Size() const { return nx6 * nx5 * nx4 * nx3 * nx2 * nx1; }
+  std::vector<TopologicalElement> topological_elements;
+  int ntop_elems;
+  bool element_matters;
```
lroberts36, line 66: It would be useful to comment on what this denotes. I think it is used to store info about the topological element size?

src/outputs/output_utils.hpp
```diff
@@ -216,31 +290,35 @@ std::vector<T> FlattenBlockInfo(Mesh *pm, int shape, Function_t f) {
 
 // mirror must be provided because copying done externally
 template <typename Data_t, typename idx_t, typename Function_t>
-void PackOrUnpackVar(MeshBlock *pmb, Variable<Real> *pvar, bool do_ghosts, idx_t &idx,
-                     std::vector<Data_t> &data, Function_t f) {
-  const auto &Nt = pvar->GetDim(6);
-  const auto &Nu = pvar->GetDim(5);
-  const auto &Nv = pvar->GetDim(4);
+void PackOrUnpackVar(const VarInfo &info, Variable<Real> *pvar, bool do_ghosts,
+                     idx_t &idx, std::vector<Data_t> &data, Function_t f) {
```
lroberts36: Are `data` and `pvar` necessary here? Looks like they might actually be accessed by lambda capture.

lroberts36 (approved):
Nice work, LGTM! 

I tested this downstream in `parthenon-mhd` on both a uniform grid and with SMR (not easy to test AMR right now because of #1031). Restarts (which stress face, edge and nodal fields there, but probably edge fields the most) give bitwise agreement with a run without an intermediate restart. HDF5 output also seems to work for non-cell centered values (although I have only checked that the shapes are correct since I haven't written scripts to plot up non-cell centered data).

src/interface/variable.hpp
Yurlungur: Nope. That's inside `ParticleVariable`.

src/interface/variable.hpp
Yurlungur: Done.

src/outputs/output_utils.cpp
Yurlungur: I will rename it `GetNumKJI`

src/outputs/output_utils.cpp
Yurlungur: It was originally used in more places. Now it is only used internally. This is an easy change.

src/outputs/output_utils.hpp
Yurlungur: :+1:

src/outputs/output_utils.hpp
Yurlungur: You're right! Will remove.

lroberts36:
@Yurlungur: I think when this goes in we can probably update our Parthenon capabilities list to include "Support for cell, face, edge, and node centered variables".

Yurlungur:
OK this is now ready for merge after sufficient approval/testing. Plotted below is the z-component of the vector potential for a rotor problem.

Yurlungur:
@pgrete @bprather this is fully ready for merge with 1 more review.

src/outputs/output_utils.cpp
```diff
@@ -33,6 +33,109 @@
...
+    auto jb = cellbounds.GetBoundsJ(domain, el);
+    auto ib = cellbounds.GetBoundsI(domain, el);
+    ks = kb.s; // pads are only upper indices
+    js = jb.s;
+    is = ib.s;
+    ke = std::max(ke, kb.e);
+    je = std::max(je, jb.e);
+    ie = std::max(ie, ib.e);
+  }
+  IndexRange kb{ks, ke}, jb{is, ie}, ib{is, ie};
```
adamdempsey90:
```suggestion
  IndexRange kb{ks, ke}, jb{js, je}, ib{is, ie};
```

Yurlungur: Oops. Thank you for the fix!

src/interface/variable.hpp
pgrete: But it's also `  std::array<int, MAX_VARIABLE_DIMENSION> dims_;` below, isn't it?

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
```diff
@@ -208,17 +208,20 @@ def plot_dump(
...
-    if len(q.shape) == 5:
-        q = q[:, components[-1], 0, :, :]
-    if len(q.shape) == 4:
-        q = q[:, 0, :, :]
+    ntensors = len(q.shape[1:-3])
+    if components:
+        if len(components) != ntensors:
+            print("value error!", len(components), ntensors, q.shape)
+            raise ValueError(
+                "Tensor rank not the same as number of specified components"
```
pgrete: Any reason for the separate preceding print statement rather than putting everything in the raised error message?

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
```diff
@@ -208,17 +208,20 @@ def plot_dump(
...
+            )
+        for c in components:
+            if c > (q.shape[1] - 1):
+                print("Value error!", c, q.shape)
+                raise ValueError("Component out of bounds")
```
pgrete: Same here.

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
```diff
@@ -208,17 +208,20 @@ def plot_dump(
...
-        q = q[:, 0, :, :]
+    ntensors = len(q.shape[1:-3])
+    if components:
+        if len(components) != ntensors:
+            print("value error!", len(components), ntensors, q.shape)
+            raise ValueError(
+                "Tensor rank not the same as number of specified components"
+            )
+        for c in components:
+            if c > (q.shape[1] - 1):
```
pgrete, line 229: What's exactly happening here (in combination with the `q = q[:, c]` below.
The `for c in ...` make me believe there's more than one component, but then it's checked against a fixed index in q.

scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
```diff
@@ -99,20 +99,29 @@ class phdf:
...
+           if found.
 
-            Default is to return a flat array of length TotalCells.
-            However if flatten is set to False, a 4D (or 5D if vector)
-            array is returned that has dimensions [NumBlocks, Nz, Ny,
-            Nx] where NumBlocks is the total number of blocks, and Nz,
-            Ny, and Nx are the number of cells in the z, y, and x
-            directions respectively.
+           If variable is a vector, each element of the returned numpy
+           array is a vector of that length.
```
pgrete: I'm having trouble parsing that statement. Does "length" here refer to the flattened length, i.e. you get a 2D numpy array back with `[c,TotalCells]` with `c` being the vector.
In that case, how about moving this statement after the first sentence of the following paragraph as (so that there's an explanation for "length".

scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
```diff
@@ -99,20 +99,29 @@ class phdf:
...
+           Nx are the number of cells in the z, y, and x directions
+           respectively.
+
+           If flatten is False and interior is True, only non-ghost data
+           will be returned. This array will correspond to the coordinates
+           xg and xng, etc.
+
+           By default, node, face, and edge centered variables are
+           averaged to cell centers for visualization. This can be
+           disabled by setting average = False.
```
pgrete: Nitpicking, but how about naming this more verbose like `average_to_cell_centers` (or sth similar) as `average` could be understood in various ways, e.g., get the mean variable quantity.

scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
```diff
@@ -492,18 +502,17 @@ def findBlockIdxInOther(self, other, ib, tol=1e-10, verbose=False):
...
-        """
-        Reads data for the named variable from file.
+    def Get(self, variable, flatten=True, interior=False, average=True):
+        """Reads data for the named variable from file.
 
         Returns None if variable is not found in the file or the data
         if found.
 
         If variable is a vector, each element of the returned numpy
         array is a vector of that length.
```
pgrete: Ah, I now see where the sentence above came from -- though my confusion remains ;)

scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
```diff
@@ -542,114 +542,134 @@ def Get(self, variable, flatten=True, interior=False):
             )
             return None
 
+        try:
+            self.varTopology[variable] = self.fid[variable].attrs["TopologicalLocation"]
+        except:
+            self.varTopology[variable] = "Cell"
+        if type(self.varTopology[variable]) != str:
+            # This is stupid and seems to be python version dependent.
```
pgrete: Is this #515 again, i..e, using `.astype(str)` should do the trick?

scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
```diff
@@ -542,114 +542,134 @@ def Get(self, variable, flatten=True, interior=False):
...
-                        self.TotalCells, vShape[-1]
+        simdim = (vShape[-1] > 1) + (vShape[-2] > 1) + (vShape[-3] > 1)
+        if average and average_able:
+            v = self.varData[variable]
+            vnew = v.copy()
+            vnew = vnew[..., :-1]
+            if simdim >= 2:
+                vnew = vnew[..., :-1, :]
+            if simdim >= 3:
+                vnew = vnew[..., :-1, :, :]
```
pgrete: What's happening here?
Is this just creating `vnew` with the appropriate shape?

scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
```diff
@@ -542,114 +542,134 @@ def Get(self, variable, flatten=True, interior=False):
...
+                        + v[..., :-1, 1:, 1:]
+                        + v[..., :-1, :-1, 1:]
+                        + v[..., :-1, 1:, :-1]
+                        + v[..., 1:, :-1, :-1]
+                    )
+            else:
+                print(
+                    "WARNING: Topology {} for var {} cannot be averaged. Resetting.".format(
+                        self.varTopology[variable], variable
+                    )
```
pgrete: How about raising an error/fail to make sure that it's not unintentionally used? From my point of view Python/numpy is sometimes too robust to go along and doing magic on arrays that was not supposed to happen.

scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
```diff
@@ -542,114 +542,134 @@ def Get(self, variable, flatten=True, interior=False):
...
-                    return self.varData[variable][:, :, :, :]
-                elif vShape[2] == 1:
-                    return self.varData[variable][:, :, :, nghost:-nghost]
-                elif vShape[1] == 1:
-                    return self.varData[variable][:, :, nghost:-nghost, nghost:-nghost]
-                else:
-                    return self.varData[variable][
-                        :, nghost:-nghost, nghost:-nghost, nghost:-nghost
-                    ]
+            print(vShape)
```
pgrete: leftover debug statement?

scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
```diff
@@ -542,114 +542,134 @@ def Get(self, variable, flatten=True, interior=False):
...
+            if vShape[-1] == 1:
+                return self.varData[variable][:]
+            elif vShape[-2] == 1:
+                return self.varData[variable][..., nghost:-nghost]
+            elif vShape[-3] == 1:
+                return self.varData[variable][..., nghost:-nghost, nghost:-nghost]
```
pgrete, line 665: Note to myself, check below what outputting ghost actually means wrt to indices of non-cell centered fields.

src/outputs/output_utils.hpp
```diff
@@ -215,32 +293,35 @@ std::vector<T> FlattenBlockInfo(Mesh *pm, int shape, Function_t f) {
...
-    // TODO(JMM): Add topological elements here
-  } else { // metadata none
-    kb = {0, pvar->GetDim(3) - 1};
-    jb = {0, pvar->GetDim(2) - 1};
-    ib = {0, pvar->GetDim(1) - 1};
+  // shape as written to or read from. contains additional padding
+  // in orthogonal directions.
+  // e.g., Face1-centered var is shape (N1+1)x(N2+1)x(N3+1)
+  // format is
+  // topological_elems x tensor_elems x block_elems
```
pgrete, line 303: Followup up on my comment above: I see, so it mirrors our pattern of allocation too much in order to reduce complexity with different shapes.

src/outputs/parthenon_hdf5.cpp
```diff
@@ -249,18 +250,9 @@ void PHDF5Output::WriteOutputFileImpl(Mesh *pm, ParameterInput *pin, SimTime *tm
...
-  for (auto &v : vars) {
-    all_vars_info.emplace_back(v);
-  }
-
-  // sort alphabetically
-  std::sort(all_vars_info.begin(), all_vars_info.end(),
-            [](const VarInfo &a, const VarInfo &b) { return a.label < b.label; });
+  // get list of all vars, just use the first block since the list is
+  // the same for all blocks
+  auto all_vars_info = VarInfo::GetAll(get_vars(pm->block_list.front()), cellbounds);
```
pgrete, line 255: I wrote the same piece of code for the openpmd output :D

src/outputs/parthenon_hdf5.cpp
```diff
@@ -308,74 +300,34 @@ void PHDF5Output::WriteOutputFileImpl(Mesh *pm, ParameterInput *pin, SimTime *tm
...
     if (output_params.hdf5_compression_level > 0) {
+      std::array<hsize_t, H5_NDIM> chunk_size;
+      std::fill(chunk_size.begin(), chunk_size.end(), 1);
+      for (int i = 1; i < ndim; ++i) {
+        chunk_size[i] = local_count[i];
+      }
+      if (vinfo.where != MetadataFlag(Metadata::None)) {
+        std::fill(&(chunk_size[0]), &(chunk_size[0]) + ndim - 3, 1);
+      }
+      PARTHENON_HDF5_CHECK(H5Pset_chunk(pl_dcreate, ndim, chunk_size.data()));
```
pgrete, line 330: :+1:  on moving this inside the soft disabled logic

src/outputs/parthenon_hdf5.cpp
```diff
@@ -308,74 +300,34 @@ void PHDF5Output::WriteOutputFileImpl(Mesh *pm, ParameterInput *pin, SimTime *tm
...
+    // be passed, which may result in slower output.
     if (output_params.hdf5_compression_level > 0) {
+      std::array<hsize_t, H5_NDIM> chunk_size;
+      std::fill(chunk_size.begin(), chunk_size.end(), 1);
+      for (int i = 1; i < ndim; ++i) {
+        chunk_size[i] = local_count[i];
+      }
+      if (vinfo.where != MetadataFlag(Metadata::None)) {
+        std::fill(&(chunk_size[0]), &(chunk_size[0]) + ndim - 3, 1);
+      }
```
pgrete, line 329: What's the actual resulting chunk size here?
We want `(1,1,1,1,nx3,nx2,nx1)` (+/- extension to nx123 based on topological elements), don't we?

src/outputs/parthenon_hdf5_types.hpp
```diff
@@ -34,15 +34,16 @@
 #include <tuple>
 #include <vector>
 
+#include "kokkos_abstraction.hpp"
 #include "utils/error_checking.hpp"
 
 namespace parthenon {
 namespace HDF5 {
 
 // Number of dimension of HDF5 field data sets (block x nv x nu x nt x nz x ny x nx)
```
pgrete: topo elem index is now also part, isn't it?

src/outputs/parthenon_xdmf.cpp
```diff
@@ -150,26 +153,23 @@ void genXDMF(std::string hdfFile, Mesh *pm, SimTime *tm, int nx1, int nx2, int n
...
         continue;
       }
-
+      ndim = vinfo.FillShape<hsize_t>(domain, &(dims[1])) + 1;
       const int num_components = vinfo.num_components;
+      nx3 = dims[ndim - 3];
+      nx2 = dims[ndim - 2];
+      nx1 = dims[ndim - 1];
+      dims321 =
+          std::to_string(nx3) + " " + std::to_string(nx2) + " " + std::to_string(nx1);
```
pgrete: How does this interact with the `dims321` set above? Or in other words, is this required here or a leftover from the non-cell centered field attempts?

src/outputs/restart_hdf5.cpp
```diff
@@ -180,64 +181,28 @@ void RestartReaderHDF5::ReadParams(const std::string &name, Params &p) {
...
-    }
+  hsize_t offset[VNDIM], count[VNDIM];
+  std::fill(offset + 1, offset + VNDIM, 0);
+  std::fill(count + 1, count + VNDIM, 1);
+
+  offset[0] = static_cast<hsize_t>(range.s);
+  count[0] = static_cast<hsize_t>(range.e - range.s + 1);
+  const IndexDomain domain = hasGhost ? IndexDomain::entire : IndexDomain::interior;
+
+  if (file_output_format_version >= HDF5::OUTPUT_VERSION_FORMAT - 1) {
```
pgrete, line 205: Maybe add a comment that this is hardcoded logic supporting restarts from 3 with current 4? (or be more specific about the numbers so that it's clear it needs an update/care when the format is changed again -- which will hopefully never happen)

src/parthenon_manager.cpp
```diff
@@ -240,17 +240,10 @@ void ParthenonManager::RestartPackages(Mesh &rm, RestartReader &resfile) {
             << std::endl;
 
   const auto file_output_format_ver = resfile.GetOutputFormatVersion();
-  if (file_output_format_ver == -1) {
+  if (file_output_format_ver < HDF5::OUTPUT_VERSION_FORMAT - 1) {
```
pgrete, line 244: As above wrt to implied version numbers here.

src/parthenon_manager.cpp
```diff
@@ -343,25 +334,14 @@ void ParthenonManager::RestartPackages(Mesh &rm, RestartReader &resfile) {
...
-              }
-            }
-          }
-        }
-      } else if (file_output_format_ver == 2 ||
-                 file_output_format_ver == HDF5::OUTPUT_VERSION_FORMAT) {
-        OutputUtils::PackOrUnpackVar(pmb.get(), v.get(), resfile.hasGhost, index, tmp,
-                                     [&](auto index, int t, int u, int v, int k, int j,
-                                         int i) { v_h(t, u, v, k, j, i) = tmp[index]; });
+      if (file_output_format_ver >= HDF5::OUTPUT_VERSION_FORMAT - 1) {
```
pgrete, line 339: same here

tst/unit/test_output_utils.cpp
```diff
@@ -0,0 +1,281 @@
+//========================================================================================
+// Athena++ astrophysical MHD code
+// Copyright(C) 2014 James M. Stone <jmstone@princeton.edu> and other code contributors
+// Licensed under the 3-clause BSD License, see LICENSE file for details
```
pgrete: Is this still appropriate for the new file or should this be PArthenon?

tst/unit/test_output_utils.cpp
```diff
@@ -0,0 +1,281 @@
...
+
+    // vector-valued var with a single value at each edge
+    const std::string scalar_edge = "scalar_edge";
+    m = Metadata({Metadata::Edge, Metadata::Independent});
+    pkg->AddField(scalar_edge, m);
+
+    const std::vector<std::string> var_names = {scalar_cell, tensor_cell, tensor_none,
+                                                vector_face, scalar_edge};
+
+    auto pmb = std::make_shared<MeshBlock>(NSIDE, NDIM);
```
pgrete, line 94: Given the bug @adamdempsey90 found wrt to indices, would it make sense to test non-cubic blocks? (though I'm not suggesting to change it here as I'm not sure how much effort that would be and if it's worth it in the context of this PR).

pgrete (commented):
Quite a rabbit hole, which came out nicely clean wrt to the changes to the hdf5 piece alone.
I'm done with the code review (with a couple of minor comments), and will approve tomorrow once I confirmed things are working (particularly restarting version 3 outputs) in AthenaPK.

src/interface/variable.hpp
Yurlungur: So it is. That said, `NumComponents` and `GetHostMirrorAndCopy` etc all assume 6D. I'd like to table this discussion, but I've made an issue #1040.

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
Yurlungur: The exceptions raised within a thread (since the code is thread-parallel) were not being propagated, so the script was just failing silently. This was my hacky way of dealing with it.

I just went back in and fixed it properly, by forcing python to raise exceptions.

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
Yurlungur: see above.

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
Yurlungur: the intent here is that `q` always has the shape `[blocks, some number of tensor components, k, j, i]` so `q[1]` is *always* the index of the slowest-moving tensor component. Then `q = q[:,c]` sets the slowest moving tensor-component to `c` but leaves the block index free floating. If you iterate with this, q keeps changing shape, so you can walk throw all tensor components, slowest first. I added a comment in the code to explain.

scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
Yurlungur: I don't think I wrote this line. I just moved it/reformated it. I also don't understand what it means. I've deleted it and modified the passage below it to explain.

scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
Yurlungur: Done.

scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
Yurlungur: Confusing sentence removed and documentation clarified... I hope. :)

scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
Yurlungur: Yes! I forgot about this. Thank you for the cleaner fix.

scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
Yurlungur: Yeah that's right. In the code below I want to be able to index into the properly shaped array. This lets me do that.

scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
Yurlungur: Done.

scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
Yurlungur: Removed.

src/outputs/output_utils.hpp
Yurlungur: that's right---so we output more data than we need to, but it simplifies things.

src/outputs/parthenon_hdf5.cpp
Yurlungur: Great minds think alike! ;)

src/outputs/parthenon_hdf5.cpp
Yurlungur: ~~I mimicked what was previously being done, which is that the chunk size is `(tensor_dims, nx3, nx2, nx1)`, individual components of a tensor are not chunked individually. We can change this if you like---I just stuck with preserving the original functionality.~~

src/outputs/parthenon_hdf5_types.hpp
Yurlungur: Right you are. :+1:

src/outputs/parthenon_xdmf.cpp
Yurlungur: `dims321` above can be deleted. It's only relevant here.

src/outputs/restart_hdf5.cpp
Yurlungur: Comment added.

src/parthenon_manager.cpp
Yurlungur: Also added more verbose error message.

src/parthenon_manager.cpp
Yurlungur: Done.

tst/unit/test_output_utils.cpp
Yurlungur: Agree. Fixed.

tst/unit/test_output_utils.cpp
Yurlungur: This seems like it would be a little bit of a pain in the unit tests here... but I agree. See issue #1041 .

src/outputs/parthenon_hdf5.cpp
Yurlungur: Oops what I said above is *not* correct. The chunk size for `Metadata::None` is just the shape of the variable on the block, with the chunk in blocks = 1. The chunk size for `Metadata::MeshTied` is `(1, 1, 1, 1, nx3+o, nx2+o, nx1+o)`, where `o` is the topological offset. It's 1 for face/edge/node and 0 for cells.

pgrete (approved):
Thanks for the updates.
I also just verified that things work in AthenaPK.
Ready to merge!

PR #1023, opened 2024-03-13
fix param bool bug

### LINKED ISSUE #1021: Simple `bool` in `Params` prevents restarts
@BenWibking discovered an interesting bug (already mentioned on Matrix).

We don't quite understand why/how this could happens.

Upon restart the code fails with
```
### PARTHENON ERROR
  Condition:   rank == view_rank
  Message:     input and output view are same rank
  File:        /home/pgrete/src/athenapk/external/parthenon/src/outputs/parthenon_hdf5.hpp
  Line number: 203
input and output view are same rank
```
Adding some debug info
```
  std::cout << "Reading var " << name << " with rank " << view_rank << " and expected rank " << rank << std::endl;
```
right in front of the failing lines results in
```
Reading var Hydro/vertical_driving_only with rank 1 and expected rank 0
```

That param is a simple bool
```
    auto vertical_driving_only =
        pin->GetOrAddBoolean("precipitator/driving", "vertical_driving_only", false);
    hydro_pkg->AddParam("vertical_driving_only", vertical_driving_only, parthenon::Params::Mutability::Restart);
```
with no additional modifications of that param in the codebase.

Also it seems to be properly stored:
```
   ATTRIBUTE "Hydro/vertical_driving_only" {
      DATATYPE  H5T_STD_U8LE
      DATASPACE  SCALAR
      DATA {
      (0): 0
      }
   }
```

Commenting the `AddParam` results in restart that don't crash anymore.

I don't understand how that's happening. We have other `bool`s stored in `Params` that do not cause issues...

Any ideas? (particularly @Yurlungur as you're also looking at outputs)

Yurlungur:
## PR Summary

Issue #1021 , raised by by @BenWibking @pgrete  is that  `parthenon` will fail to read a single boolean from the params file upon restart. The issue, it turns out, was caused by me. To save code, scalar values of hdf5 params are read as vectors. For all types except booleans, this works very well. For booleans, unfortunately `std::vector<bool>` and `std::vector<string>` are special, and not implemented in the same way as `std::vector<T>`. 

I thought I could short-circuit this, and introduced this bug. I tried to use a `Kokkos::View` to read a `std::vector<bool>`. This works for true vectors of bools, but not for the scalar case. 

This PR removes the optimistic hack I had introduced and reads boolean params properly.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [ ] Docs build

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 ### Changed (changing behavior/API/variables/...)
 
 ### Fixed (not changing behavior/API/variables/...)
+- [[PR1023]](https://github.com/parthenon-hpc-lab/parthenon/pull/1023) Fix broken param of a scalar bool
 - [[PR1012]](https://github.com/parthenon-hpc-lab/parthenon/pull/1012) Remove accidentally duplicated code
 - [[PR992]](https://github.com/parthenon-hpc-lab/parthenon/pull/992) Allow custom PR ops with sparse pools
 - [[PR988]](https://github.com/parthenon-hpc-lab/parthenon/pull/988) Fix bug in neighbor finding routine for small, periodic, refined meshes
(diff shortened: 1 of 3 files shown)


c-prather (approved):
LGTM.  I don't read params in KHARMA to test this, but the new unit test seems adequate.

BenWibking (approved):

lroberts36 (approved):
LGTM. It is very obnoxious that `std::vector<bool>` is such a unique snowflake

PR #1024, opened 2024-03-14
Add features to history output

brryan:
## PR Summary

We want additional control in how history outputs are produced. This PR:

- [x] Adds the ability to enroll history outputs on a per-package basis
- [x] Adds the ability to have multiple history output files with different output quantities and different cadences [**breaking change** by changing e.g. `my_output.hst` to `my_output.out2.hst`]
- [x] Merges MPI reductions by operation to reduce communication
- [x] Adds the ability to enroll vector quantities as history outputs. 
- [x] Include example vector output in advection example.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [ ] Docs build


example/advection/parthinput.advection
```diff
@@ -83,6 +83,11 @@ file_type = hst
 dt = 0.05
 
 <parthenon/output4>
+file_type = hst
+dt = 0.1
+packages = advection_app
```
brryan, line 88: Example for including only history output from specific package (`advection_app` has no enrolled history outputs so `advection.out4.hst` only has `t` and `dt` fields).

src/outputs/history.cpp
```diff
@@ -45,72 +45,116 @@ namespace parthenon {
...
+      usr_op = MPI_MIN;
+    }
+
+    if (Globals::my_rank == 0) {
+      PARTHENON_MPI_CHECK(MPI_Reduce(MPI_IN_PLACE, &(results[op])[0], results[op].size(),
+                                     MPI_PARTHENON_REAL, usr_op, 0, MPI_COMM_WORLD));
+    } else {
+      PARTHENON_MPI_CHECK(MPI_Reduce(&(results[op])[0], &(results[op])[0],
+                                     results[op].size(), MPI_PARTHENON_REAL, usr_op, 0,
+                                     MPI_COMM_WORLD));
```
brryan, line 151: I evaluate history results per-processor separately by MPI operation and then each MPI operation is called only once over that vector.

src/outputs/outputs.hpp
```diff
@@ -147,9 +149,20 @@ struct HistoryOutputVar {
       : hst_op(hst_op_), hst_fun(hst_fun_), label(label_) {}
 };
 
+struct HistoryOutputVec {
```
brryan, line 152: I treated vector output separately from scalar history output because I didn't see another way to do this apart from a breaking change where enrolled history functions *only* return `std::vector<Real>` i.e. all existing history implementations would need to switch.

src/outputs/history.cpp
```diff
@@ -45,72 +45,116 @@ namespace parthenon {
...
 
-  // only the master rank writes the file
-  // create filename: "file_basename" + ".hst".  There is no file number.
+  // Only the master rank writes the file.
+  // Create filename: "file_basename.out" + [output number] + ".hst". There is no file
+  // number.
   if (Globals::my_rank == 0) {
     std::string fname;
     fname.assign(output_params.file_basename);
+    fname.append(".out" + std::to_string(output_params.block_number));
```
brryan: E.g. `advection.hst` is now something like `advection.out3.hst` which is changed behavior.

example/advection/advection_package.cpp
```diff
@@ -375,6 +382,45 @@ void PostFill(MeshBlockData<Real> *rc) {
...
+  std::vector<Real> result(nvec, 0);
+
+  // We choose to apply volume weighting when using the sum reduction.
+  // Downstream this choice will be done on a variable by variable basis and volume
+  // weighting needs to be applied in the reduction region.
+  const bool volume_weighting = std::is_same<T, Kokkos::Sum<Real, HostExecSpace>>::value;
+
+  for (int n = 0; n < nvec; n++) {
+    T reducer(result[n]);
+    pmb->par_reduce(
```
Yurlungur: might as well use `par_reduce` rather than `pmb->par_reduce`?

example/advection/advection_package.cpp
```diff
@@ -375,6 +382,45 @@ void PostFill(MeshBlockData<Real> *rc) {
...
+  const int nvec = 3;
+
+  std::vector<Real> result(nvec, 0);
+
+  // We choose to apply volume weighting when using the sum reduction.
+  // Downstream this choice will be done on a variable by variable basis and volume
+  // weighting needs to be applied in the reduction region.
+  const bool volume_weighting = std::is_same<T, Kokkos::Sum<Real, HostExecSpace>>::value;
+
+  for (int n = 0; n < nvec; n++) {
```
Yurlungur, line 406: would be nice to do this all at once. Dunno if it's possible though.

src/outputs/history.cpp
```diff
@@ -49,68 +49,111 @@ void HistoryOutput::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm,
...
+  std::vector<UserHistoryOperation> ops = {
+      UserHistoryOperation::sum, UserHistoryOperation::max, UserHistoryOperation::min};
+  for (auto &op : ops) {
+    results[op] = {};
+    labels[op] = {};
+  }
+
+  // If "packages" not provided, output history for all packages
+  auto &requested_packages = output_params.packages;
+  Dictionary<std::shared_ptr<StateDescriptor>> packages;
```
Yurlungur, line 64: You could use a `Packages_t` here but that might be overkill

src/outputs/history.cpp
```diff
@@ -49,68 +49,111 @@ void HistoryOutput::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm,
...
+  if (requested_packages.empty()) {
+    packages = pm->packages.AllPackages();
+  } else {
+    for (const auto &pkg : pm->packages.AllPackages()) {
+      if (std::find(requested_packages.begin(), requested_packages.end(), pkg.first) !=
+          requested_packages.end()) {
+        packages[pkg.first] = pkg.second;
+      }
+    }
```
Yurlungur: This seems a little backwards. In the limit of `n_total_packages > n_requested_packages`, this is probably better:
```suggestion
    const auto &all_packages = pm->packages.AllPackages();
    for (const auto & pkg_name : requested_packages) {
      const auto &it = all_packages.find(pkg_name);
      if (it != all_packages.end()) {
        packages[it.first] = it.second;
      }
   }
```

src/outputs/history.cpp
Yurlungur: this is really nice

src/outputs/history.cpp
Yurlungur: It might be better to make this something machine readable like `%02d` and no `out` something like that?

Yurlungur (approved):
Thanks for adding this. This is a really nice quality of life improvement. A few nitpicks below, but nothing blocking.

src/outputs/history.cpp
brryan: I switched this to use the existing `output_params.file_id` that we use for `.outN.` for other output types like restart and HDF5. I think it is good to keep the `out` because it unifies the output block numbering which is now further being used for histories from the input files.

example/advection/advection_package.cpp
brryan: Good point I just got that from copy/pasting the scalar version (I updated that to `par_reduce` as well).

example/advection/advection_package.cpp
brryan: 🤷

src/outputs/history.cpp
Yurlungur: ok that makes sense. Would still prefer 0-padding for the output number.

src/outputs/history.cpp
brryan: Looking at `Packages_t` I do think it would be overkill here.

src/outputs/history.cpp
brryan: Yeah this is an improvement, fixed, thanks

src/outputs/history.cpp
brryan: I switched to using the existing `outN` string calculated here: https://github.com/parthenon-hpc-lab/parthenon/blob/cd96dd78141d0d09a957c671f16e78941069e6fe/src/outputs/outputs.cpp#L138

I think if we want to add zero-padding to that, that should be a separate PR/discussion since it could break everybody's file readers.

Yurlungur: kk works for me.

example/advection/parthinput.advection
pdmullen: `packages = ` seems to signal to me that I could supply a comma separated list of packages to be enrolled in output4.  Is that the case? Will keep reading...

src/outputs/history.cpp
```diff
@@ -49,68 +49,113 @@ void HistoryOutput::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm,
...
+        labels[hist_vec.hst_op].push_back(label);
       }
-#endif
+    }
+  }
 
-      all_results.emplace_back(result);
-      all_labels.emplace_back(hist_var.label);
+  // Need fence so result is ready prior to MPI call
+  Kokkos::fence();
```
pdmullen, line 129: Should the fence be inside the `#ifdef MPI_PARALLEL`?

src/outputs/outputs.cpp
```diff
@@ -196,6 +196,11 @@ Outputs::Outputs(Mesh *pm, ParameterInput *pin, SimTime *tm) {
         }
       }
 
+      if (op.file_type == "hst") {
+        op.packages = pin->GetOrAddVector<std::string>(pib->block_name, "packages",
```
pdmullen, line 200: Q above now answered.

pdmullen (approved):
LGTM!

src/outputs/history.cpp
brryan: You're right, fixed, thanks

example/advection/advection_package.cpp
pgrete: In principle, this would be possible if we'd include a reduction to array capability in Parthenon.
AFAIK this is not part of the standard Kokkos API (but the example in the Kokkos wiki illustrate how to write such as custom reducer).
We also recently had a use case for this downstream but I went for atomic instead of a reduction array/view because the section was not performance critical.
Having that said, this feature might be of broader interest and Parthenon could offer some standard reduction array with a user/compile time defined number of entries.

example/advection/advection_package.cpp
```diff
@@ -400,9 +446,9 @@ Real AdvectionHst(MeshData<Real> *md) {
   // weighting needs to be applied in the reduction region.
   const bool volume_weighting = std::is_same<T, Kokkos::Sum<Real, HostExecSpace>>::value;
 
-  pmb->par_reduce(
-      PARTHENON_AUTO_LABEL, 0, advected_pack.GetDim(5) - 1, kb.s, kb.e, jb.s, jb.e, ib.s,
-      ib.e,
+  parthenon::par_reduce(
+      parthenon::loop_pattern_mdrange_tag, PARTHENON_AUTO_LABEL, DevExecSpace(), 0,
+      advected_pack.GetDim(5) - 1, kb.s, kb.e, jb.s, jb.e, ib.s, ib.e,
```
pgrete, line 451: Why this change? (also using above using the more verbose par_reduce over the terse pmb one)?

src/outputs/history.cpp
```diff
@@ -49,68 +49,113 @@ void HistoryOutput::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm,
...
+  auto &requested_packages = output_params.packages;
+  Dictionary<std::shared_ptr<StateDescriptor>> packages;
+  if (requested_packages.empty()) {
+    packages = pm->packages.AllPackages();
+  } else {
+    const auto &all_packages = pm->packages.AllPackages();
+    for (const auto &pkg_name : requested_packages) {
+      const auto &it = all_packages.find(pkg_name);
+      if (it != all_packages.end()) {
+        packages[(*it).first] = (*it).second;
```
pgrete, line 72: Should we throw an error if the packages has not been found? To prevent accidental typos in the input file.

tst/regression/test_suites/output_hdf5/output_hdf5.py
```diff
@@ -119,12 +119,15 @@ def Analyse(self, parameters):
...
-        hst_3d = np.genfromtxt("advection_3d.hst")
+        hst_2d = np.genfromtxt("advection_2d.out1.hst")
+        hst_3d = np.genfromtxt("advection_3d.out1.hst")
         ref_results = [
             ["time", 1.0, 1.0],
             ["dt", 1.75781e-03, 3.12500e-03],
             ["total", 7.06177e-02, 1.39160e-02],
+            ["power0", 7.06177e-02, 1.39160e-02],
+            ["power1", 3.88112e-02, 2.59597e-03],
+            ["power2", 2.65948e-02, 7.19427e-04],
```
pgrete: This should be `advected_powers#`, isn't it?
Also, if there are no objection, I'd prefer the indices separated by `_` as we're using that kind of convention in other places already.

pgrete (approved):
Really nice change. Thanks!

example/advection/advection_package.cpp
Yurlungur: Actually a broader conversation... which maybe I should start and/or just implement, is the `par_for`s that aren't tied to meshblock shouldn't need to take in an exec space. We should provide overloads that accept default arguments. (Not your problem for this MR though.)

example/advection/advection_package.cpp
brryan: This was requested by @Yurlungur earlier because the loop isn't associated with `pmb`. He then also suggested adding default `parthenon::par_*` in the future to avoid having to be more verbose like you say.

I think the switch to terse `parthenon::par_reduce` would ultimately be reasonable, so I'm inclined to leave this change as is in anticipation of that unless you feel strongly.

src/outputs/history.cpp
brryan: Yeah I agree. One could imagine someone wanting to write out history from specific packages and automatically not getting that output if they enable/disable individual packages at runtime, but I think catching typos is more important. Also, a way to disable packages at runtime is to enroll an empty package with the same name, which would work for the above. 

Changed to error if this `if` statement is `false`.

example/sparse_advection/sparse_advection_driver.cpp
```diff
@@ -122,12 +122,12 @@ TaskCollection SparseAdvectionDriver::MakeTaskCollection(BlockList_t &blocks,
                              mdudt.get(), beta * dt, mc1.get());
 
     // do boundary exchange
-    auto restrict =
+    auto prolong_restrict =
```
brryan: `restrict` is a C keyword, and was not compiling for me. I'm actually not sure how we've been getting away with this for months...

example/advection/advection_package.cpp
Yurlungur: We discussed this on the last call---and I believe everyone is on board with deprecating the `pmb` interface.

example/sparse_advection/sparse_advection_driver.cpp
Yurlungur: huh. Good catch.

tst/regression/test_suites/output_hdf5/output_hdf5.py
brryan: Yes, the other labels were actually wrong as well (e.g. `total` should have been `total_advected`). I changed all the variable names to be corrected and also added a test of the column headers.

And I also added the `_` by default for vector-length history output so it is now e.g. `advected_powers_2`. 

IN ADDITION I also enforced a space between each column so we can parse the labels more easily. In `develop` currently we can get headers like `# [0]=time    [1]=dt   [2]=my_long_var_name[3]=my_other_longvar_name` which can be annoying to parse via something simple like `.split()` in python. I don't think this is a breaking change because whitespace between variable names was already a possibility (for short variable names) but I should mention it.

example/sparse_advection/sparse_advection_driver.cpp
lroberts36: I think it worked because `restrict` is only sort of a keyword, see https://github.com/parthenon-hpc-lab/parthenon/pull/1009#issuecomment-2009801991.

brryan: ahh gotcha, thanks @lroberts36 for making that connection for me, and the context (and sorry for blaming your PR for this before!). Maybe my build recipe is defaulting to C++11 somehow? Regardless, I'd rather just change this to a never-was-a-keyword if there are no objections

brryan:
@pgrete @Yurlungur @pdmullen My plan is to set this to automerge sometime tomorrow -- last call for objections!

Thanks for your reviews all

example/sparse_advection/sparse_advection_driver.cpp
lroberts36: I agree it should be changed, but the change may conflict with #1009 where it is already changed. Simple enough to deal with though.

brryan: Merge conflict resolved :+1:

example/advection/parthinput.advection
```diff
@@ -83,6 +83,11 @@ file_type = hst
...
+
+<parthenon/output5>
 file_type = ascent
 dt = -0.05   # soft disabled by default, as Ascent is an optional dependency
```
pgrete, line 92: I think this is what's causing the failed regression test as ascent outputs are manually enabled in `.github/workflows/ci-extended.yml` via `parthenon/output4/dt=0.05`

brryan: 🚨 thank you for noticing this @pgrete! 

I also modified `.github/workflows/ci-extended.yml` to give an error message if `test` fails.

PR #1027, opened 2024-03-19
Refactor `RestartReader` as abstract class

pgrete:
## PR Summary

The default `RestartReader` is now an abstract base class, which now also provides a clean interface to support OpenPMD restarts.
This got a little more ugly/more changed code in the interface than I'd hoped for given that our templated Get/Read functions are not supported in combination with virtual function.
Thus, I decided to hide these internals by separating out `TimeInfo` and `MeshInfo` similar to the existing `SparseInfo`.
Only `ReadBlocks` and `ReadSwarmVar` now have duplicated interfaces to allow for different types.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [ ] Docs build


src/outputs/restart.hpp
```diff
@@ -83,263 +63,63 @@ class RestartReader {
...
-    int rank;
-    hsize_t count;
-    std::vector<hsize_t> dims;
+  struct MeshInfo {
+    int nbnew, nbdel, nbtotal, root_level, includes_ghost, n_ghost;
+    std::vector<std::string> bound_cond;
+    std::vector<int> block_size;
+    std::vector<Real> grid_dim;
+    std::vector<int64_t> lx123;
+    std::vector<int> level_gid_lid_cnghost_gflag; // what's this?!
```
lroberts36, line 75: Do we want all of this stuff mashed together going forward? For the forest stuff, we would also have to add a tree index here.

src/outputs/restart_hdf5.hpp
```diff
@@ -0,0 +1,339 @@
...
+#endif // ENABLE_HDF5
+  }
+
+ public:
+  // Gets data for all blocks on current rank.
+  // Assumes blocks are contiguous
+  // fills internal data for given pointer
+  void ReadBlocks(const std::string &name, IndexRange range, std::vector<Real> &dataVec,
+                  const std::vector<size_t> &bsize, int file_output_format_version,
+                  MetadataFlag where, const std::vector<int> &shape = {}) const override {
```
lroberts36: Should this be declared here and defined in the cpp file?

src/parthenon_manager.cpp
```diff
@@ -100,11 +105,15 @@ ParthenonStatus ParthenonManager::ParthenonInitEnv(int argc, char *argv[]) {
     pinput = std::make_unique<ParameterInput>(arg.input_filename);
   } else if (arg.res_flag != 0) {
     // Read input from restart file
-    restartReader = std::make_unique<RestartReader>(arg.restart_filename);
+    if (fs::path(arg.restart_filename).extension() == ".rhdf") {
+      restartReader = std::make_unique<RestartReaderHDF5>(arg.restart_filename);
+    } else {
+      PARTHENON_FAIL("HELP!");
```
lroberts36:
```suggestion
      PARTHENON_FAIL("Unsupported restart file format!");
```

lroberts36 (approved):
LGTM

src/outputs/restart.hpp
pgrete: If you mean "all mesh related stuff" with "all", then I'd imagine so.
I separate "time" and "mesh" similar to the separate `SparseInfo` though I'm also open to suggestion for other ways.
The key issue was trying to resolve is that `GetAttr<T>` in a templated version could not be used in the abstract class anymore, so I introduced this interface so that all the `GetAttr<T>` calls are private to the derived class.

src/outputs/restart_hdf5.hpp
pgrete: yes, this was a leftover from the original implementation and I'll clean it up.

src/parthenon_manager.cpp
pgrete: done

src/mesh/mesh.cpp
```diff
@@ -505,13 +506,14 @@ Mesh::Mesh(ParameterInput *pin, ApplicationInput *app_in, RestartReader &rr,
...
+  auto mesh_info = rr.GetMeshInfo();
   // All ranks read HDF file
-  nbnew = rr.GetAttr<int>("Info", "NBNew");
-  nbdel = rr.GetAttr<int>("Info", "NBDel");
-  nbtotal = rr.GetAttr<int>("Info", "NumMeshBlocks");
-  root_level = rr.GetAttr<int>("Info", "RootLevel");
+  nbnew = mesh_info.nbnew;
+  nbdel = mesh_info.nbdel;
+  nbtotal = mesh_info.nbtotal;
+  root_level = mesh_info.root_level;
```
Yurlungur, line 514: :+1:

src/mesh/mesh.cpp
```diff
@@ -553,14 +555,11 @@ Mesh::Mesh(ParameterInput *pin, ApplicationInput *app_in, RestartReader &rr,
...
-  const auto includesGhost = rr.GetAttr<int>("Info", "IncludesGhost");
-  const auto nGhost = rr.GetAttr<int>("Info", "NGhost");
-
   for (auto &dir : {X1DIR, X2DIR, X3DIR}) {
     block_size.xrat(dir) = mesh_size.xrat(dir);
-    block_size.nx(dir) =
-        blockSize[dir - 1] - (blockSize[dir - 1] > 1) * includesGhost * 2 * nGhost;
+    block_size.nx(dir) = mesh_info.block_size[dir - 1] -
+                         (mesh_info.block_size[dir - 1] > 1) * mesh_info.includes_ghost *
+                             2 * mesh_info.n_ghost;
```
Yurlungur, line 562: With different topological elements, this information will be less useful than it used to be. But I'm not sure where all the dependencies are on this.

src/outputs/restart.hpp
Yurlungur: Does this belong in restart, actually? Or should it be elsehwere?

src/outputs/restart.hpp
```diff
@@ -83,262 +64,62 @@ class RestartReader {
...
-    // type
-    T *typepointer = nullptr;
-    handle.type = getHDF5Type(typepointer);
-    const H5T dset_type = H5T::FromHIDCheck(H5Dget_type(handle.dataset));
-    status = PARTHENON_HDF5_CHECK(H5Tequal(handle.type, dset_type));
-    PARTHENON_REQUIRE_THROWS(status > 0, "Type mismatch for dataset " + name);
+  struct TimeInfo {
+    Real time, dt;
+    int ncycle;
+  };
```
Yurlungur: Same for this. Don't we already have an appropriate struct for this actually?

src/outputs/restart.hpp
```diff
@@ -83,262 +64,62 @@ class RestartReader {
...
-
-    // Read data from file
-    PARTHENON_HDF5_CHECK(H5Dread(hdl.dataset, hdl.type, memspace, hdl.dataspace,
-                                 H5P_DEFAULT, dataVec.data()));
-#endif // ENABLE_HDF5
-  }
+  virtual void ReadBlocks(const std::string &name, IndexRange range,
+                          std::vector<Real> &dataVec, const std::vector<size_t> &bsize,
+                          int file_output_format_version, MetadataFlag where,
+                          const std::vector<int> &shape = {}) const = 0;
```
Yurlungur, line 92: I may change this interface. But it's fine for now.

src/outputs/restart.hpp
```diff
@@ -83,262 +64,62 @@ class RestartReader {
...
-    }
-
-    return res[0];
-  }
+  virtual void ReadSwarmVar(const std::string &swarmname, const std::string &varname,
+                            const std::size_t count, const std::size_t offset,
+                            const Metadata &m, std::vector<Real> &dataVec) = 0;
+  virtual void ReadSwarmVar(const std::string &swarmname, const std::string &varname,
+                            const std::size_t count, const std::size_t offset,
+                            const Metadata &m, std::vector<int> &dataVec) = 0;
```
Yurlungur, line 102: Unfortunate that these can't remain templated... On the other hand, maybe this will speed up compile times...

src/parthenon_manager.cpp
```diff
@@ -28,16 +28,18 @@
 
 #include "amr_criteria/refinement_package.hpp"
 #include "config.hpp"
-#include "driver/driver.hpp"
+#include FS_HEADER
```
Yurlungur, line 31: I don't recall---Why is this needed? Is this related to `std::filesystem`? If so, I suggest we might want to avoid using it... as it's been problematic on some machines.

src/outputs/restart_hdf5.hpp
```diff
@@ -0,0 +1,256 @@
...
+#include <hdf5.h>
+
+#include "interface/metadata.hpp"
+#include "outputs/parthenon_hdf5_types.hpp"
+
+using namespace parthenon::HDF5;
+// TODO(someone) the following "else" is very ugly but fixes missing types when not
+// using hdf5. The issue is that some of the restart machinery lives outside of
+// restart.cpp and restart.hpp and uses the interfaces below. We should overall update
+// the logic to reduce the ifdefs for HDF5 in this (and other restart related) files.
```
Yurlungur: gross. Maybe we should move these definitions to the `parthenon_hdf5_types.hpp` file? Then that file can always be included where relevant?

Yurlungur (approved):
This all seems reasonable. Some nitpicky comments below. 

I think our changesets are *effectively* orthogonal, but will be annoying to merge because you move code between files. Assuming all of this works, I'm going to merge your branch into mine... I don't want to deal with having to migrate my changes to restarts into separate files.

src/outputs/restart.hpp
pgrete: All/most these variables live as member of `Mesh`.
Here,they are being used only for restarts, i.e., these are temp structures that contain the necessary info that is being *read* from the restart file and then passed to the appropriate places in the code allowing to hide the backend specific `GetAttr<T>` calls.
This is mirroring the existing `SparseInfo` struct used for restarts and I split this into `Time` and `Mesh` from a logical point of view.

src/mesh/mesh.cpp
pgrete: Yeah, I was also wondering about this, but decided to keep this PR on a refactor and not cleanup.

src/parthenon_manager.cpp
pgrete: It's needed for checking the end of the filename and addresses the issue with `std::experimental::filesystem` versus `std::filesystem` for some old compilers/stdlibs.

Are you suggesting to remove it?

src/outputs/restart_hdf5.hpp
pgrete: Let me double check.
This PR might actually have resolved the issue as what's need in other places is just the knowledge of the abstract class, which does not need hdf5.
The only item left might be writing Params, but that could potentially be alleviated by adding a `DumpParams` interface that accepts the package as paramter rather than having a `DumpToHdf5` call living inside params.

src/outputs/restart.hpp
Yurlungur: For `MeshInfo` I can see that.. for `TimeInfo` though, why not just re-use `SimTime`? It contains almost the same information and is a tiny container struct that can be passed around.

src/parthenon_manager.cpp
Yurlungur: We've had trouble with `std::filesystem` on some lanl systems, so I lean towards removing that library from our dependencies... but that's maybe a longer discussion and doesn't belong in this PR.

Yurlungur: (i.e., the discussion doesn't belong in this PR. your changeset can stay.)

src/outputs/restart.hpp
pgrete: Do you mean to introduce a struct in `Mesh` that holds this info so it can be reused (similar to `SimTime`)?

src/outputs/restart.hpp
pgrete: Done.

src/outputs/restart_hdf5.hpp
pgrete: Good news. Given that the `DatasetHandle` (and all associate calls) are now private to the hdf5 reader, the `else` statement could be removed.

pgrete:
I think I've addressed all comments. The only thing potentially left is what to do about params, i.e., whether to keep the hdf5 parts in Params or move it to the appropriate reader/writer (and passing Params as an argument).
I'm leaning towards the latter as it'd result in confining external library interfaces to fewer places.

src/outputs/restart.hpp
Yurlungur: Yeah, that's what I meant... but we don't have to do that now... or maybe ever.

PR #1031, opened 2024-03-25
Bug fix in non-cell centered AMR

lroberts36:
## PR Summary
On develop, there is a bug that shows up when performing AMR on non-cell centered fields. Coming out of `RedistributeAndRefineMeshBlocks`, `FillDerived` may have been called using uninitialized non-cell centered data at certain locations on newly refined blocks. Additionally, the boundaries of those non-cell centered fields may have been filled with uninitialized neighbor data at certain locations. 

This PR fixes that by breaking out the different tasks in `Mesh::Initialize` and calling them in the correct order to ensure that non-cell centered fields are filled correctly in the interior before calling `FillDerived` (and `PreCommFillDerived`). Additionally, a second communication is added to make sure the ghost values are correctly filled. The extra communication uses mesh data that only contains the non-cell centered fields with FillGhost and does not occur if none of these fields exist. 

This bug strongly suggests that we need to add a regression test that stresses non-cell centered fields, but I think that will be a bit of work and too big for this PR. As @Yurlungur suggests in #1019, a transverse vector wave equation solver may be a good choice for this test. 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/src/mesh/mesh.cpp
+++ b/src/mesh/mesh.cpp
@@ -917,7 +917,7 @@ void Mesh::ApplyUserWorkBeforeOutput(Mesh *mesh, ParameterInput *pin,
   }
 }
 
-void Mesh::BuildCommunicationBuffers() {
+void Mesh::BuildBoundaryBuffers() {
   const int num_partitions = DefaultNumPartitions();
   const int nmb = GetNumMeshBlocksThisRank(Globals::my_rank);
 
@@ -1138,7 +1138,7 @@ void Mesh::Initialize(bool init_problem, ParameterInput *pin, ApplicationInput *
 
     PreCommFillDerived();
 
-    BuildCommunicationBuffers();
+    BuildBoundaryBuffers();
 
     CommunicateBoundaries();
 
--- a/src/mesh/mesh.cpp
+++ b/src/mesh/mesh.cpp
@@ -917,6 +917,161 @@ void Mesh::ApplyUserWorkBeforeOutput(Mesh *mesh, ParameterInput *pin,
   }
 }
 
+
+void Mesh::BuildCommunicationBuffers() {
+  const int num_partitions = DefaultNumPartitions();
+  const int nmb = GetNumMeshBlocksThisRank(Globals::my_rank); 
+
+  // Build densely populated communication tags
+  tag_map.clear();
+  for (int i = 0; i < num_partitions; i++) {
+    auto &md = mesh_data.GetOrAdd("base", i);
+    tag_map.AddMeshDataToMap<BoundaryType::any>(md);
+    for (int gmg_level = 0; gmg_level < gmg_mesh_data.size(); ++gmg_level) {
+      auto &mdg = gmg_mesh_data[gmg_level].GetOrAdd(gmg_level, "base", i);
+      // tag_map.AddMeshDataToMap<BoundaryType::any>(mdg);
+      tag_map.AddMeshDataToMap<BoundaryType::gmg_same>(mdg);
+      tag_map.AddMeshDataToMap<BoundaryType::gmg_prolongate_send>(mdg);
+      tag_map.AddMeshDataToMap<BoundaryType::gmg_restrict_send>(mdg);
+      tag_map.AddMeshDataToMap<BoundaryType::gmg_prolongate_recv>(mdg);
+      tag_map.AddMeshDataToMap<BoundaryType::gmg_restrict_recv>(mdg);
+    }
+  }
+  tag_map.ResolveMap();
+
+  // Create send/recv MPI_Requests for swarms 
+  for (int i = 0; i < nmb; ++i) {
+    auto &pmb = block_list[i];
+    pmb->swarm_data.Get()->SetupPersistentMPI();
+  }
+
+  // Wait for boundary buffers to be no longer in use 
+  bool can_delete;
+  std::int64_t test_iters = 0;
+  constexpr std::int64_t max_it = 1e10;
+  do {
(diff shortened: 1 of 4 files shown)


Yurlungur (approved):
I like to see this split up. Would also like this fix in ASAP.

c-prather (approved):
KHARMA doesn't hit this (which is all I tested when I introduced it lol).  So, I'm happy to approve the good reorg now, and if it introduces a regression for me I'll figure it out later.

However, this feels like the sort of organization thing @pgrete should weigh in on before it gets upstreamed.

lroberts36:
> KHARMA doesn't hit this (which is all I tested when I introduced it lol). So, I'm happy to approve the good reorg now, and if it introduces a regression for me I'll figure it out later.

@bprather: It would probably be good to at least make sure this doesn't break things for you before it goes in since things are a bit reorganized relative to what was there before.

c-prather:
> @bprather: It would probably be good to at least make sure this doesn't break things for you before it goes in since things are a bit reorganized relative to what was there before.

Yeah, I'll be pulling this as I can, but since KHARMA's currently based on a pretty forked branch it may be a bit before I could spot any regression.  So, marking that I'm happy if this goes in before I get a chance.

I think if the regression tests & parthenon-mhd work, KHARMA's unlikely to hit anything crippling from this patch (and thus isn't crucial as a data point before putting this in).  And, if AMR breaks in KHARMA I'm pretty sure no one cares -- no one's using it or planning to soon, and it works in the release anyway.

lroberts36:
This branch also passes the Riot test suite

src/mesh/mesh.cpp
```diff
@@ -917,6 +917,161 @@ void Mesh::ApplyUserWorkBeforeOutput(Mesh *mesh, ParameterInput *pin,
...
+  const int nmb = GetNumMeshBlocksThisRank(Globals::my_rank);
+
+  // Build densely populated communication tags
+  tag_map.clear();
+  for (int i = 0; i < num_partitions; i++) {
+    auto &md = mesh_data.GetOrAdd("base", i);
+    tag_map.AddMeshDataToMap<BoundaryType::any>(md);
+    for (int gmg_level = 0; gmg_level < gmg_mesh_data.size(); ++gmg_level) {
+      auto &mdg = gmg_mesh_data[gmg_level].GetOrAdd(gmg_level, "base", i);
+      // tag_map.AddMeshDataToMap<BoundaryType::any>(mdg);
```
pgrete:
```suggestion
```
Or comment why it's commented

src/mesh/mesh.cpp
```diff
@@ -917,6 +917,161 @@ void Mesh::ApplyUserWorkBeforeOutput(Mesh *mesh, ParameterInput *pin,
   }
 }
 
+void Mesh::BuildCommunicationBuffers() {
```
pgrete: Maybe add a comment what differentiate `BuildCommunicationBuffers` from` BuildBoundaryBuffers`?

pgrete (approved):
I think the refactor makes sense.
One thing I don't fully understand is why this didn't come up for cell centered fields or, in other words, were is the logic split between handling cell centered and non cell-centered fields. Is this related to the ownership model for which there's no block/buffer taking precedence?
So  maybe some additional comments why this separate treatment is required would be useful for future reference.

Finally, given that we don't have non-cc fields in AthenaPK, I don't have other means to test so I'm trusting the Parthenon internal and Riot results/testing.

src/mesh/mesh.cpp
lroberts36: I guess we could just call this `BuildBoundaryBuffers` since it is just building the tag map then calling `BuildBoundaryBuffers`.

src/mesh/mesh.cpp
lroberts36: It is redundant (it was also commented out on develop).

lroberts36:
> I think the refactor makes sense. One thing I don't fully understand is why this didn't come up for cell centered fields or, in other words, were is the logic split between handling cell centered and non cell-centered fields. Is this related to the ownership model for which there's no block/buffer taking precedence? So maybe some additional comments why this separate treatment is required would be useful for future reference.

@pgrete: The key difference with cell centered fields is that internal prolongation is a no-op for them so all cell-centered values on newly refined blocks are correctly filled after a call to `ProlongateShared`. Previously, `Initialize` (i.e. communicating boundary data and calling fill derived) was called after `ProlongateShared` but before `ProlongateInternal`. This meant that if you only had cell centered data everything worked fine, but if you had non-cell centered elements on the new fine grid that were not shared with the coarse grid they were uninitialized during boundary communication and fill derived (which obviously gives junk on the grid and in the boundaries). Now, if there are non-cell centered values, we communicate them after `ProlongateShared` so that shared elements respect the correct ownership. Then we call `ProlongateInternal`, at which point the internal values of all fields are correctly filled. Then we can safely FillDerived and communicate everything so that the interior and boundaries are correct after refinement.  

> Finally, given that we don't have non-cc fields in AthenaPK, I don't have other means to test so I'm trusting the Parthenon internal and Riot results/testing.

I think the real test was in `parthenon-mhd` (where AMR was broken before adding this fix), Riot doesn't have any non-cell centered values. Really, I was more concerned to check that I didn't do something in the re-organization that impacted cell centered fields in some unexpected way.

src/mesh/mesh.cpp
lroberts36: done.

PR #1028, opened 2024-03-20
[Trivial] Reorganize `LogicalLocation` code

lroberts36:
## PR Summary

Just move some files around and split up some files to make the code easier to navigate. Also removes a little bit of dead code.
## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Changes are summarized in CHANGELOG.md
- [x] (@lanl.gov employees) Update copyright on changed files
- [x] #1009 is merged

PR #1035, opened 2024-03-26
Fix Multigrid Solver in Forest

lroberts36:
## PR Summary
This PR updates the geometric multigrid infrastructure to work with the new forest of trees based mesh. Generally, I think this has resulted in a much cleaner implementation.

The forest neighbor finding algorithm is extended to find neighbors in different types of meshes (i.e. the standard leaf mesh and two-level composite meshes at different levels). Additionally, `gid`s are assigned to internal nodes in the forest when `Forest::GetMeshBlocksAndResolveGids` is called. The way the level of gmg grids is defined is also simplified so that it just corresponds to the logical level of finer blocks that are members of the grid (which means we store the `gmg_block_lists` in a map instead of a vector since the index can be negative). All of these changes should be internal and not be visible to downstream codes. Because we no longer have to worry about periodic root grids (with periodicity being taken care of by the forest connectivity) and because we use the forest information explicitly for gmg neighbor finding, a *lot* of the complex neighbor finding code in `LogicalLocation` is no longer necessary. This PR also removes a lot of other code that died in switching to a forest.    

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build
- [x] #1028 is merged into develop


src/interface/data_collection.cpp
```diff
@@ -70,15 +70,13 @@ GetOrAdd_impl(Mesh *pmy_mesh_,
...
       containers_[md_label]->Set(partitions[i], pmy_mesh_);
-      if (gmg_level >= 0) {
-        int min_gmg_logical_level = pmy_mesh_->GetGMGMinLogicalLevel();
-        containers_[md_label]->grid = GridIdentifier{GridType::two_level_composite,
-                                                     gmg_level + min_gmg_logical_level};
+      if (gmg_level) {
+        containers_[md_label]->grid = GridIdentifier::two_level_composite(*gmg_level);
       } else {
-        containers_[md_label]->grid = GridIdentifier{GridType::leaf, 0};
+        containers_[md_label]->grid = GridIdentifier::leaf();
```
lroberts36, line 79: gmg levels can now be negative, so I switch to a `std::optional` for `gmg_level`. If it is not set, it implies that we are working with meshdata for the regular leaf grid.

lroberts36:
Passes `Riot` tests and works with `parthenon-mhd`.

Yurlungur (approved):
LGTM. Love to see the negative line count.

lroberts36:
hm, I managed to break something in the last couple commits. Will get back to it on Monday. Likely an issue related to `LogicalLocation`s with negative levels.

lroberts36:
> hm, I managed to break something in the last couple commits. Will get back to it on Monday. Likely an issue related to `LogicalLocation`s with negative levels.

now fixed, was due to a bug introduced by a merge.

pgrete:
Is it safe to change the target branch to develop now given that #1028 has been merged?

lroberts36:
@pgrete: The target branch is now `develop`.

src/mesh/forest/logical_location.cpp
```diff
@@ -145,71 +143,27 @@ bool LogicalLocation::IsNeighborOfTEForest(const LogicalLocation &in,
...
-    }
+LogicalLocation LogicalLocation::GetParent(int nlevel) const {
+  // Shift all locations so that ghost tree locations have positive values,
+  // then shift back. Looks overly complicated to deal with ghosts and negative
+  // levels, but basically boils down to
+  // return LogicalLocation(tree(), level() - nlevel, lx1() >> nlevel, lx2() >> nlevel,
+  // lx3() >> nlevel); for most cases
+  const int norig = 1LL << std::max(level(), 0);
+  const int nparent = 1LL << std::max(level() - nlevel, 0);
+  constexpr int nmax_tree_offset = 5;
```
pgrete, line 154: Is this a magic number/should be be obvious why it's 5?

src/mesh/forest/tree.cpp
```diff
@@ -43,16 +43,21 @@ Tree::Tree(Tree::private_t, std::int64_t id, int ndim, int root_level, RegionSiz
...
-            internal_nodes.emplace(my_id, l, i, j, k);
+            internal_nodes.emplace(LocMapEntry(loc, -1, -1));
           }
         }
       }
     }
   }
+
+  // Build in negative levels
+  for (int l = -20; l < 0; ++l) {
```
pgrete, line 58: Is -20 just a large enough number that we should never reach it?

src/mesh/mesh.cpp
```diff
@@ -1182,86 +1182,23 @@ std::shared_ptr<MeshBlock> Mesh::FindMeshBlock(int tgid) const {
...
-          loc.l(dir - 1) == 0 ? mesh_bcs[GetInnerBoundaryFace(dir)] : BoundaryFlag::block;
-      block_bcs[GetOuterBoundaryFace(dir)] = loc.l(dir - 1) == nrbx_ll - 1
-                                                 ? mesh_bcs[GetOuterBoundaryFace(dir)]
-                                                 : BoundaryFlag::block;
-    } else {
-      block_bcs[GetInnerBoundaryFace(dir)] = mesh_bcs[GetInnerBoundaryFace(dir)];
-      block_bcs[GetOuterBoundaryFace(dir)] = mesh_bcs[GetOuterBoundaryFace(dir)];
-    }
-  }
+  block_size = forest.GetBlockDomain(loc);
```
pgrete, line 1191: Unused var?

pgrete (approved):
I looked at the code and followed as much as I could (which, admittedly is not too much given that I haven't really internalized all the logic and associated magic).
Anyway, negative lines are always welcome and I just stumbled across two numbers that I weren't able to immediate connect.

src/mesh/forest/logical_location.cpp
lroberts36: I think it could be anything that doesn't overflow when multiplied by `norig`.

src/mesh/forest/tree.cpp
lroberts36: yeah, for that to not be enough would imply we had a meshblock size of (2^20)^D.

src/mesh/mesh.cpp
lroberts36: No, `block_size` is passed by reference. The input to this routine is `loc` and the output is `block_size` and `block_bcs`. I agree that this is a weird design, but it is what was originally in the code.

PR #1037, opened 2024-03-27
Add SwarmPacks

pdmullen:
This MR introduces the notion of a `SwarmPack`.  It is a stripped down version of `SparsePack`, tuned towards particle swarms, to enable packing over `MeshBlock`s.  I have preliminarily settled on a design wherein one can pack a `SwarmPack` via a `std::vector<std::string>` or the type-based variable prescription previously used in `SparsePack`s.  `SwarmPack`s are unique from `SparsePack`s in that we must be able to handle both `Real` and `integer` types.  In this MR, I do not allow `SwarmPack`s to contain mixed types, however, one can create a pack of `Real` swarm variables and a separate pack for `int` swarm variables.  The design follows:

by `std::vector<std::string>` (where the template argument specifies that the data type):

```c++
std::vector<std::string> vars{swarm_position::x::name(),
                              swarm_position::y::name(),
                              swarm_position::z::name()};
static auto desc = MakeSwarmPackDescriptor<Real>(swarm_name, vars);
auto pack = desc.GetPack(md);
```
by type-based vars (where the data type is inferred from var-types):

```c++
static auto desc = MakeSwarmPackDescriptor<swarm_position::x,
                                           swarm_position::y,
                                           swarm_position::z>(swarm_name);
auto pack = desc.GetPack(md);
```

There are so many opportunities to further clean up things here... One could imagine (1) allowing mixed types in packs, (2) allowing for packing via Metadata or regex, or (4) combining `SparsePack`s and `SwarmPack`s, to name a few.  

This MR adds the `SwarmPack` machinery to the `particle_leapfrog` test.

Finally, I'd note that this MR also transitions `swarm_data` to belong to `meshblock_data` rather than `pmb`.  This likely has consequence when considering copies of `mesh_data` registers (e.g., in RK staging).  Any undesired manifestations from this change should be uncovered as we continue to test downstream.  For one, the current design requires that downstream users interface with particles via the `base` `mesh_data` register.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build
- [x] Include caching for `SwarmPack`s (cache cleared whenever `Swarm::setPoolMax` is called)


scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
```diff
@@ -59,15 +59,15 @@ def Get(self, variable):
 
     @property
     def x(self):
-        return self.Get("x")
+        return self.Get("swarm.x")
```
Yurlungur, line 62: Why this name change here?

src/interface/make_swarm_pack_descriptor.hpp
```diff
@@ -0,0 +1,59 @@
...
+#include "interface/metadata.hpp"
+#include "interface/state_descriptor.hpp"
+#include "interface/swarm_pack.hpp"
+#include "interface/variable.hpp"
+#include "mesh/mesh.hpp"
+
+namespace parthenon {
+
+template <typename TYPE>
+inline auto MakeSwarmPackDescriptor(const std::string &swarm_name,
```
Yurlungur, line 36: not super important but I'm not sure if I love this name. It's self-consistent with swarms though, so maybe it's right. I can't think of anything better.

src/interface/mesh_data.hpp
```diff
@@ -445,6 +457,8 @@ class MeshData {
   MapToMeshBlockVarPack<T> varPackMap_;
   MapToMeshBlockVarFluxPack<T> varFluxPackMap_;
   SparsePackCache sparse_pack_cache_;
+  SwarmPackCache<int> swarm_pack_int_cache_;
+  SwarmPackCache<Real> swarm_pack_real_cache_;
```
Yurlungur, line 500: It feels like we're duplicating a lot of code between swarm packs and variable packs... but maybe there's no way around that.

src/interface/meshblock_data.hpp
```diff
@@ -209,6 +211,17 @@ class MeshBlockData {
     return GetAllVariables(sparse_ids).unique_ids();
   }
 
+  DataCollection<SwarmContainer> swarm_data;
+  std::vector<std::shared_ptr<Swarm>> GetAllSwarms() {
+    return swarm_data.Get()->GetSwarmVector();
+  }
```
Yurlungur: It looks like one change here is that `swarm`s are now more consistently included in `MeshBlockData`? That's nice. I approve.

src/interface/swarm_pack.hpp
```diff
@@ -0,0 +1,207 @@
...
+
+  KOKKOS_INLINE_FUNCTION
+  std::size_t VariableIdx() { return vidx; }
+  KOKKOS_INLINE_FUNCTION
+  int Offset() { return offset; }
+
+ private:
+  std::size_t vidx;
+  int offset;
+};
```
Yurlungur: This looks identical to the variable index machinery from sparse packs... can the sparse pack one be re-used rather than copy-pasting code?

src/interface/swarm_pack.hpp
```diff
@@ -0,0 +1,207 @@
...
+  }
+  KOKKOS_INLINE_FUNCTION
+  static bool regex() { return REGEX; }
+  KOKKOS_INLINE_FUNCTION
+  static int ndim() { return sizeof...(NCOMP); }
+  KOKKOS_INLINE_FUNCTION
+  static int size() { return multiply<NCOMP...>::value; }
+
+  const int idx;
+};
```
Yurlungur: Same as above... is there a reason to not just re-use the variable name machinery from sparse packs?

src/interface/swarm_pack.hpp
```diff
@@ -0,0 +1,207 @@
...
+  using SwarmPackBase<TYPE>::nblocks_;
+
+  explicit SwarmPack(const SwarmPackBase<TYPE> &spb) : SwarmPackBase<TYPE>(spb) {
+    if constexpr (sizeof...(Ts) != 0) {
+      static_assert(std::is_same<TYPE, typename GetDataType<Ts...>::value>::value,
+                    "Type mismatch in SwarmPack! When passing type-based variables as "
+                    "template argument to SwarmPack, ensure that the first template "
+                    "parameter is a data type (e.g., Real or int) that matches the "
+                    "data type of subsequent variable types!");
+    }
```
Yurlungur, line 58: Nice. :+1: You can probably move the `sizeof` into the assert by requiring `sizeof...(Ts) == 0 || std::is_same`

src/interface/make_swarm_pack_descriptor.hpp
```diff
@@ -0,0 +1,59 @@
...
+template <typename TYPE, class... Ts>
+inline auto MakeSwarmPackDescriptor(const std::string &swarm_name) {
+  static_assert(sizeof...(Ts) > 0, "Must have at least one variable type for type pack");
+
+  std::vector<std::string> vars{Ts::name()...};
+
+  return typename SwarmPack<TYPE, Ts...>::Descriptor(
+      static_cast<impl::SwarmPackDescriptor<TYPE>>(
+          MakeSwarmPackDescriptor<TYPE>(swarm_name, vars)));
+}
```
Yurlungur: I think because you have the `PackDescriptor` you can infer `TYPE` from `Ts` in only this function so you can do the automatic type inferrence we discussed. And I *think* only this function needs to change. Maybe this is sufficient?
```suggestion
template <class... Ts>
inline auto MakeSwarmPackDescriptor(const std::string &swarm_name) {
  static_assert(sizeof...(Ts) > 0, "Must have at least one variable type for type pack");
  using TYPE = typename GetDataType<Ts...>::value;

  std::vector<std::string> vars{Ts::name()...};

  return typename SwarmPack<TYPE, Ts...>::Descriptor(
      static_cast<impl::SwarmPackDescriptor<TYPE>>(
          MakeSwarmPackDescriptor<TYPE>(swarm_name, vars)));
}
```
because the two `MakePackDescriptor` functions have different numbers of input variables, I *think* this should just work and should be unambiguous.

src/interface/swarm_pack.hpp
```diff
@@ -0,0 +1,207 @@
...
+  KOKKOS_INLINE_FUNCTION int GetLowerBound(const int b, const TIn &) const {
+    const int vidx = GetTypeIdx<TIn, Ts...>::value;
+    return bounds_(0, b, vidx);
+  }
+
+  template <class TIn, REQUIRES(IncludesType<TIn, Ts...>::value)>
+  KOKKOS_INLINE_FUNCTION int GetUpperBound(const int b, const TIn &) const {
+    const int vidx = GetTypeIdx<TIn, Ts...>::value;
+    return bounds_(1, b, vidx);
+  }
```
Yurlungur: For swarms the bounds stuff probably doesn't need to depend on block, as swarms are not sparse.

Yurlungur: That said, it would be nice to add the `Contains` and the `Host` variants, so that one can check bounds on host and so that one can check whether or not a pack contains a particular swarm var.

src/interface/swarm_pack_base.hpp
```diff
@@ -0,0 +1,265 @@
...
+  for (int b = 0; b < pmd->NumBlocks(); ++b) {
+    auto &pmbd = pmd->GetBlockData(b);
+    func(b, pmbd.get());
+  }
+}
+
+template <class T, class F>
+inline auto ForEachBlock(T *pmbd, F func) -> decltype(T().GetBlockPointer(), void()) {
+  func(0, pmbd);
+}
```
Yurlungur: Is there a reason you can't re use these loops from sparse packs?

src/interface/swarm_pack_base.hpp
```diff
@@ -0,0 +1,265 @@
...
+    std::vector<int> vardims;
+    ForEachBlock(pmd, [&](int b, mbd_t *pmbd) {
+      auto swarm = pmbd->GetSwarm(desc.swarm_name);
+      int size = 0;
+      nblocks++;
+      for (auto &pv : swarm->template GetVariableVector<TYPE>()) {
+        for (int i = 0; i < nvar; ++i) {
+          if (desc.IncludeVariable(i, pv)) {
+            size += pv->GetDim(6) * pv->GetDim(5) * pv->GetDim(4) * pv->GetDim(3) *
+                    pv->GetDim(2);
```
Yurlungur, line 100: I wonder if swarms should just be capable of reporting their tensorsize so you don't have to compute this product here?

src/interface/swarm_pack_base.hpp
```diff
@@ -0,0 +1,265 @@
...
+            ndim = 1;
+            int vardim = (pv->GetDim(6) > 1 ? 1 : 0) + (pv->GetDim(5) > 1 ? 1 : 0) +
+                         (pv->GetDim(4) > 1 ? 1 : 0) + (pv->GetDim(3) > 1 ? 1 : 0) +
+                         (pv->GetDim(2) > 1 ? 1 : 0);
```
Yurlungur: same for dimension...

src/interface/swarm_pack_base.hpp
```diff
@@ -0,0 +1,265 @@
...
+    pack.nblocks_ = nblocks;
+
+    // Allocate the views
+    int leading_dim = 1;
+    pack.pack_ = pack_t("data_ptr", leading_dim, nblocks, max_size);
+    auto pack_h = Kokkos::create_mirror_view(pack.pack_);
+
+    int bounds_leading_size = 2 + 1 + *max_element(vardims.begin(), vardims.end());
+    pack.bounds_ = bounds_t("bounds", bounds_leading_size, nblocks, nvar);
+    auto bounds_h = Kokkos::create_mirror_view(pack.bounds_);
```
Yurlungur: because swarms aren't sparse, I don't think the bounds view needs to depend on blocks.

src/interface/swarm_variable_types.hpp
```diff
@@ -0,0 +1,38 @@
...
+    static std::string name() { return #ns "." #varname; }                               \
+  }
+
+namespace swarm_position {
+
+SWARM_VARIABLE(parthenon::Real, swarm, x);
+SWARM_VARIABLE(parthenon::Real, swarm, y);
+SWARM_VARIABLE(parthenon::Real, swarm, z);
+
+} // namespace swarm_position
```
Yurlungur: I like having this just be provided. :+1:

src/outputs/parthenon_xdmf.cpp
Yurlungur, line 1: This never worked anyway... but thanks for keeping it consistent.

Yurlungur (commented):
I am super happy to see this be added! Worth noting it does break downstream because you moved where swarms live to be in meshblock data (which is where they belong). Comments below.

scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
pdmullen: Swarm positions are now a default type defined here: 

https://github.com/parthenon-hpc-lab/parthenon/pull/1037/files#diff-e1cb8671fdfab655d0e0da72dfa34efaa82ad67a4f7f235274509c290b4052ffR32

The corresponding string name from this approach is now "swarm.x".

src/interface/make_swarm_pack_descriptor.hpp
pdmullen: I couldn't come up with anything better.  However, now I am wondering if we do `MakeSwarmPackDescriptor`, should the call to actually get the pack be `desc.GetSwarmPack(md)`.  Right now, it is `desc.GetPack(md)` which overlaps with `SparsePacks` 😬

src/interface/mesh_data.hpp
pdmullen: Yeah I went the extreme of duplicating everything at first.  Probably lots of opportunity to clean up (see your below comments too...)

src/interface/meshblock_data.hpp
pdmullen: As the MR text shows, we may need to think about how this change works in practice, particularly with multiple mesh data registers, e.g., with RK sub-staging.

src/interface/swarm_pack.hpp
pdmullen: I think this one was actually directly copied and pasted.  Will address...

src/interface/swarm_pack.hpp
pdmullen: This one is a bit different from the `SparsePack` iteration.  Now the data type is a template parameter... but there is no reason why we couldn't also do this for `SparsePack`s.  When enrolling new variables, it might be advantageous to have distinct `variable_names` and `swarm_variable_names` `namespaces`, let me ponder this one a bit.... but I do recognize that copy-paste=bad.

src/interface/swarm_pack.hpp
pdmullen: That should work since the right operand won't be evaluated when the left is `false`.

src/interface/make_swarm_pack_descriptor.hpp
pdmullen: ah yes! 😸 I think this is a viable path.  

I couldn't get the automatic deduction to work prior to introducing `MakeSwarmPackDescriptor` when trying to stay compatible with both string based and var-type based packs.  But now that this functionality is introduced, I think your path is a clear one to avoiding extraneous template arguments.

src/interface/swarm_pack.hpp
pdmullen: I did a lot of pruning for `SwarmPack`s.  It looks like some of the functionality I ripped out might be useful to others...

src/interface/swarm_pack_base.hpp
pdmullen: this was copy-pasted.  will address...

src/interface/swarm_pack_base.hpp
pdmullen: let me ponder this one

src/interface/swarm_pack_base.hpp
pdmullen: ...and this one

src/interface/swarm_variable_types.hpp
pdmullen: Should it be `swarm_default_types.hpp`?

src/outputs/parthenon_xdmf.cpp
pdmullen: 😉

scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
Yurlungur: but is there a reason to make it `swarm.x` and not `x`? I guess you put it in a namespace, which is why?

src/interface/make_swarm_pack_descriptor.hpp
Yurlungur: I think either is fine. Since the descriptor is a different type than the sparse pack descriptor.

src/interface/swarm_pack.hpp
```diff
@@ -0,0 +1,207 @@
...
+  KOKKOS_INLINE_FUNCTION
+  static int size() { return multiply<NCOMP...>::value; }
+
+  const int idx;
+};
+
+} // namespace swarm_variable_names
+
+template <typename TYPE, class... Ts>
+class SwarmPack : public SwarmPackBase<TYPE> {
```
lroberts36, line 40: I wonder if we could just extend/template `SparsePack` to provide the functionality given by `SwarmPack` and `SparsePack` (although the would have to inherit different from different base classes). Reduces code duplication but may result in some gross SFINAE.

For instance, we could change calls like 
```c++
  KOKKOS_INLINE_FUNCTION
  Real &operator()(int b, int idx, const int k, const int j, const int i) const {
    return pack_(0, b, idx)(k, j, i);
  }
 ```
 to 
 ```c++
  template <class... Args> 
  KOKKOS_INLINE_FUNCTION
  Real &operator()(int b, int idx, Args&&... args) const {
    return pack_(0, b, idx)(std::foward<Args>(args)...);
  }
 ```
 which would work for both swarm and field shapes.

File: src/interface/swarm_variable_types.hpp (hunk 14 above)
Comment (Yurlungur) [reply src/interface/swarm_variable_types.hpp]: maybe `swarm_default_names.hpp`?

File: src/interface/swarm_pack.hpp (hunk 9 above)
Comment (Yurlungur) [reply src/interface/swarm_pack.hpp]: I would remove the dependence on `b` but add the host variants and re-introduce `Contains`?

File: scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py (hunk 1 above)
Comment (pdmullen) [reply scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py:62]: yepp it's the namespace.

File: src/interface/make_swarm_pack_descriptor.hpp (hunk 8 above)
Comment (pdmullen) [reply src/interface/make_swarm_pack_descriptor.hpp]: This worked 😄

File: src/interface/swarm_pack.hpp (hunk 7 above)
Comment (pdmullen) [reply src/interface/swarm_pack.hpp:58]: hmm this didn't work... reverting for now... will look into more depth soon

File: example/particle_leapfrog/particle_leapfrog.cpp [hunk 16]
```diff
@@ -207,46 +207,74 @@ void ProblemGenerator(MeshBlock *pmb, ParameterInput *pin) {
...
+  // Make a SwarmPack via strings to get ids
+  // NOTE(@pdmullen): since we are constructing the pack via strings, we must specify
+  // the datatype associated with ids (i.e., int).  We also extract an indexing map.
+  std::vector<std::string> vars_id{"id"};
+  static auto desc_id = MakeSwarmPackDescriptor<int>(swarm_name, vars_id);
+  auto pack_id = desc_id.GetPack(md);
+  auto pack_id_map = desc_id.GetMap();
+  parthenon::SwarmPackIdx spi_id(pack_id_map["id"]);
+
+  // Make a SwarmPack via strings to get v (note that v is a vector!)
```
brryan, line 236: vector and scalar variables work together in a single pack right?

src/interface/meshblock_data.cpp
```diff
@@ -54,6 +54,18 @@ void MeshBlockData<T>::Initialize(
     AddField(q.first.base_name, q.second, q.first.sparse_id);
   }
 
+  auto &swarm_container = swarm_data.Get();
+  swarm_container->SetBlockPointer(pmb);
+  for (auto const &q : resolved_packages->AllSwarms()) {
+    swarm_container->Add(q.first, q.second);
```
brryan, line 60: As we discussed, this may need to be modified to support `Metadata::OneCopy` for `Swarm`s.

src/interface/meshblock_data.hpp
```diff
@@ -209,6 +211,17 @@ class MeshBlockData {
     return GetAllVariables(sparse_ids).unique_ids();
   }
 
+  DataCollection<SwarmContainer> swarm_data;
```
brryan: I don't know that this needs to be a `DataCollection<SC>` rather than just a `SwarmContainer` because I don't think we need "stages" inside an already-staged `MeshBlockData`.

src/interface/swarm_pack.hpp
brryan: Is this assertion checked in a unit test? the argument of the `static_assert` is a bit opaque to me

src/outputs/output_utils.cpp
```diff
@@ -1,6 +1,6 @@
 //========================================================================================
 // Parthenon performance portable AMR framework
-// Copyright(C) 2023 The Parthenon collaboration
+// Copyright(C) 2024 The Parthenon collaboration
```
brryan: 2023-2024?

brryan (commented):
After a brief first read, this looks great to me so far.

example/particle_leapfrog/particle_leapfrog.cpp
pdmullen: They work together  😸

src/interface/meshblock_data.cpp
pdmullen: Agreed.  I still need to address the potential consequences of moving `swarm_data` to `MeshBlockData`.  Not only with RK substaging but also partitioning.

src/interface/meshblock_data.hpp
pdmullen: Yeah this first pass just ensured that what used to belong to `pmb` is now owned by `MeshblockData`, but as you allude to, there may be some inconsistencies that need to be resolved (see above comment).

src/outputs/output_utils.cpp
pdmullen: thanks!

src/interface/swarm_pack.hpp
pdmullen: this assertion is not currently in a unit test (but I did test it downstream).  I am having a heck of a time getting unit tests to work with swarms because the unit tests are isolated (i.e., they don't follow the same initialization pathways a normal downstream code would see).

src/interface/swarm_pack.hpp
brryan: These may unify naturally in the future if we add support for `int` type fields to parthenon...

src/interface/swarm_pack.hpp
brryan: Yeah if this can be removed in favor of reusing `PackIdx` I would encourage that. Looks like it should work.

src/interface/swarm_pack.hpp
brryan: Because `T` has a default `Real` parameter can `SparsePack` just use this version out of the box? Not sure if it'll interfere with using `NCOMP` if `T=Real` isn't at the end of the tempalte parameter list.

brryan (commented):
This is great work and I don't have any concerns here apart from thinking it would be best to share `struct`s that are already identical up to default template parameters.

src/interface/mesh_data.hpp
pdmullen: Addressed duplication to some extent in most recent commits.

src/interface/swarm_pack.hpp
pdmullen: Addressed duplication to some extent in most recent commits.

src/interface/meshblock_data.hpp
```diff
@@ -258,6 +261,22 @@ class MeshBlockData {
     return GetAllVariables(sparse_ids, flux).unique_ids();
   }
 
+  // Queries related to swarm data
+  DataCollection<SwarmContainer> &GetSwarmData(int n = 0) {
```
brryan: Fail on `n > 0`? It looks like we rely downstream on only using the `Get()` method to get the default stage from `DataCollection<SwarmContainer>`. Does anything go wrong if we just get rid of the `DataCollection` and store a `SwarmContainer` directly (and still fail here for `n != 0`)?

src/interface/swarm_default_names.hpp
```diff
@@ -0,0 +1,37 @@
...
+//========================================================================================
+#ifndef INTERFACE_SWARM_DEFAULT_NAMES_HPP_
+#define INTERFACE_SWARM_DEFAULT_NAMES_HPP_
+
+#include <string>
+#include <utility>
+
+#include "swarm_pack.hpp"
+
+#define SWARM_VARIABLE(type, ns, varname)                                                \
```
brryan, line 21: Should this macro be advertised in the docs and used downstream?

example/particle_leapfrog/particle_leapfrog.cpp
```diff
@@ -273,31 +294,36 @@ TaskCollection ParticleDriver::MakeParticlesUpdateTaskCollection() const {
...
     }
   }
 
+  TaskRegion &tr = tc.AddRegion(num_partitions);
+  for (int i = 0; i < num_partitions; i++) {
+    auto &tl = tr[i];
+    auto &base = pmesh->mesh_data.GetOrAdd("base", i);
+    auto transport = tl.AddTask(none, TransportParticles, base.get(), &integrator);
+  }
+
```
lroberts36, line 316: nice

src/interface/meshblock_data.cpp
lroberts36: Is this issue now resolved?

src/interface/pack_utils.hpp
```diff
@@ -0,0 +1,197 @@
+//========================================================================================
```
lroberts36, line 1: It feels like this file should maybe not be in the interface directory, but not sure where it should go. Utils is not quite right either.

lroberts36: I didn't check the functions here carefully since I think they are all moved from elsewhere. Is that right?

src/interface/swarm.cpp
```diff
@@ -237,6 +239,30 @@ void Swarm::setPoolMax(const std::int64_t nmax_pool) {
...
+  auto &int_md_cache = pmd->GetSwarmPackCache<int>().pack_map;
+  if (real_md_cache.size() > 0) real_md_cache.clear();
+  if (int_md_cache.size() > 0) int_md_cache.clear();
+
+  // (3): MeshBlockData
+  auto pmbd = pmb->meshblock_data.Get().get();
+  auto &real_mbd_cache = pmbd->GetSwarmPackCache<Real>().pack_map;
+  auto &int_mbd_cache = pmbd->GetSwarmPackCache<int>().pack_map;
+  if (real_mbd_cache.size() > 0) real_mbd_cache.clear();
+  if (int_mbd_cache.size() > 0) int_mbd_cache.clear();
```
lroberts36: This feels like it should live in `MeshData` and `MeshBlockData`, i.e. there should be a function `ClearSwarmCaches` in those that just gets called here.

src/interface/swarm_pack.hpp
lroberts36: resolving this for now, although I think this would be nice to do in a future PR.

src/interface/swarm_pack_base.hpp
```diff
@@ -0,0 +1,267 @@
...
+                      PARTHENON_REQUIRE(
+                          pack_h(0, b, idx).size() > 0,
+                          "Seems like this variable might not actually be allocated.");
+                      idx++;
+                    }
+                  }
+                }
+              }
+            }
+            bounds_h(2, b, i) = (pv->GetDim(6) > 1 ? 1 : 0) +
```
lroberts36: Maybe a brief comment that you are storing the tensor size of the swarm variable in `bounds`? At least I think that is what is being done here.

src/interface/swarm_pack_base.hpp
```diff
@@ -0,0 +1,267 @@
...
+  static SwarmPackBase<TYPE> BuildAndAdd(T *pmd, const SwarmPackDescriptor<TYPE> &desc) {
+    auto &pack_map = pmd->template GetSwarmPackCache<TYPE>().pack_map;
+    pack_map[desc.identifier] = Build<T>(pmd, desc);
+    return pack_map[desc.identifier];
+  }
+
+  template <class T>
+  static SwarmPackBase<TYPE> Get(T *pmd, const impl::SwarmPackDescriptor<TYPE> &desc) {
+    auto &pack_map = pmd->template GetSwarmPackCache<TYPE>().pack_map;
+    if (pack_map.count(desc.identifier) > 0) {
```
lroberts36, line 174: I guess swarm packs are static aside from the supplemental part, so if they exist in the cache they must be fresh?

lroberts36 (approved):
Nice, this LGTM. Some small non-blocking questions/comments below.

example/particle_leapfrog/particle_leapfrog.cpp
```diff
@@ -207,45 +207,66 @@ void ProblemGenerator(MeshBlock *pmb, ParameterInput *pin) {
...
           // drift
-          x(n) += v(0, n) * 0.5 * dt;
-          y(n) += v(1, n) * 0.5 * dt;
-          z(n) += v(2, n) * 0.5 * dt;
-
-          bool on_current_mesh_block = true;
-          swarm_d.GetNeighborBlockIndex(n, x(n), y(n), z(n), on_current_mesh_block);
+          pack_pos(b, swarm_position::x(), n) += pack_v(b, iv + 0, n) * 0.5 * dt;
+          pack_pos(b, swarm_position::y(), n) += pack_v(b, iv + 1, n) * 0.5 * dt;
+          pack_pos(b, swarm_position::z(), n) += pack_v(b, iv + 2, n) * 0.5 * dt;
```
Yurlungur, line 264: Was the missing v a bug before?

src/interface/meshblock_data.hpp
Yurlungur: Actually I'm a bit confused. What does `n` signify here? Is there a reason to allow us to pass in `n` at all? I'd rather fail at compile time than run time.

src/interface/swarm_pack_base.hpp
pdmullen: correct

src/interface/pack_utils.hpp
pdmullen: happy to move this... although I agree that it doesn't belong in `utils`.  Really, this stuff is just what is shared between `SparsePack`s and `SwarmPack`s.  Maybe a later merging of the two will eliminate this file.

example/particle_leapfrog/particle_leapfrog.cpp
pdmullen: 😸

src/interface/meshblock_data.cpp
Yurlungur: What happens currently if I copy a `MeshBlockData` object? I assume swarms just simply aren't copied? Agree with @brryan that in the fullness of time it would be nice to just treat them as `OneCopy` (or even make copies of them) but this is fine for now IMO.

src/interface/pack_utils.hpp
Yurlungur, line 1: Nice. I like this stuff all shared in a utils file.

src/interface/meshblock_data.cpp
pdmullen: I now currently require that `swarm_data` is accessed through the `base` `MeshData` register.  Is there any possible scenario where this is not adequate?  How about with partitions?

src/interface/pack_utils.hpp
```diff
@@ -0,0 +1,197 @@
...
+          idx += args;
+        }(),
+        ...);
+    return idx;
+  }
+};
+template <bool REGEX, int... NCOMP>
+struct base_t : public var_base_t<REGEX, Real, NCOMP...> {
+  using var_base_t<REGEX, Real, NCOMP...>::var_base_t;
+};
```
Yurlungur, line 129: Nice. I like this solution for the type templates.

src/interface/swarm_default_names.hpp
Yurlungur: :+1: In general we need some documentation for sparse packing and features as a whole.

src/interface/pack_utils.hpp
lroberts36: yeah, I just feel like the `interface` directory already has too much stuff in it so was thinking if there was a way to not add more.

src/interface/meshblock_data.hpp
```diff
@@ -258,6 +261,23 @@ class MeshBlockData {
     return GetAllVariables(sparse_ids, flux).unique_ids();
   }
 
+  // Queries related to swarm data
+  std::shared_ptr<SwarmContainer> &GetSwarmData(int n = 0) {
```
Yurlungur, line 270: Why not just remove `n` from the argument list here?

src/interface/swarm_pack_base.hpp
```diff
@@ -0,0 +1,267 @@
...
+  using contexts_h_t = typename ParArray1D<SwarmDeviceContext>::HostMirror;
+  using max_active_indices_t = ParArray1D<int>;
+  using desc_t = impl::SwarmPackDescriptor<TYPE>;
+  using idx_map_t = std::unordered_map<std::string, std::size_t>;
+
+  // Build supplemental entries to SwarmPack that change on a cadence faster than the
+  // other pack cache
+  template <class T>
+  static void BuildSupplemental(T *pmd, const SwarmPackDescriptor<TYPE> &desc,
+                                SwarmPackBase<TYPE> &pack) {
```
Yurlungur, line 57: Nice solution here for the rate at which things can be cached.

src/interface/swarm_pack_base.hpp
Yurlungur: Resolving this for now. But I still think it's worth doing down the road.

src/interface/swarm_pack_base.hpp
Yurlungur: Resolving this for now. But I still think it's worth doing down the road.

src/interface/swarm_pack_base.hpp
Yurlungur: It's actually not clear to me what the bounds calls/array does for swarms. Do we need one?

src/kokkos_abstraction.hpp
```diff
@@ -215,6 +215,17 @@ inline void kokkos_dispatch(ParallelScanDispatch, Args &&...args) {
...
+// this pattern does not support reductions yet
+template <typename Tag, typename Function>
+inline void par_dispatch(LoopPatternSimdFor, const std::string &name,
+                         DevExecSpace exec_space, const int &il, const int &iu,
+                         const Function &function) {
+  PARTHENON_INSTRUMENT_REGION(name)
+#pragma omp simd
+  for (auto i = il; i <= iu; i++)
+    function(i);
+}
```
Yurlungur, line 227: did we really not have a 1D simdfor pattern before? That's kind of shocking.

Yurlungur (approved):
I still have some minor nitpicks but nothing I would call blocking. Thanks for this cleanup @pdmullen it's really great to have swarm packs.

src/interface/pack_utils.hpp
Yurlungur: Can this just go in `utils`?

src/interface/meshblock_data.hpp
pdmullen: I require that both `MeshBlockData` and `MeshData` versions of this function have the same function signature and arguments.  In this way, we can template downstream functions on data type and interface with swarms correctly.  I did add a check on `n` in a recent commit.

example/particle_leapfrog/particle_leapfrog.cpp
pdmullen: sorry which missing v?

src/interface/meshblock_data.hpp
Yurlungur: Ah ok for mesh data `n` is the block index? Makes sense. :+1:

example/particle_leapfrog/particle_leapfrog.cpp
Yurlungur: I just can't read. Nothing is missing.

src/interface/meshblock_data.hpp
pdmullen: Resolved via comment elsewhere.  Addressed your concern @brryan.

src/interface/pack_utils.hpp
pdmullen: Agreed that `interface` directory has too much.  

Also agreed that it doesn't belong in `utils`.  

Resolving this for now until we merge `SparsePack`s and `SwarmPack`s or refactor what is contained in `interface`.

src/interface/swarm_pack_base.hpp
pdmullen: I removed this extraneous debugging info from bounds.  Thanks for pointing me to additional cleanup.

src/interface/swarm.cpp
pdmullen: Addressed in [59c0d1c](https://github.com/parthenon-hpc-lab/parthenon/pull/1037/commits/59c0d1c3118de6fa5d40af115c818d1c5c67b419)

src/interface/swarm_pack.hpp
pdmullen: Let's keep it for now, anticipating a merge b/w `SparsePack`s and `SwarmPack`s.  There was other significant cleanup to `bounds` in a recent commit following your advice.

pgrete:
FYI I'm happy with the discussion and don't plan a detailed review. So just go ahead an merge when everyone is happy.

PR #1042, opened 2024-04-02
[Small] Cleanup of `NeighborBlock`

lroberts36:
## PR Summary
Cleanup of `NeighborBlock` code to remove extraneous classes and fields that were essentially unused. Also updates code to use `CellCenterOffsets` in a few places. **This PR possibly breaks downstream codes that directly access `NeighborBlock`, but the fix should be simple since no information was removed from `NeighborBlock` just slightly reorganized`**.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted


Yurlungur (approved):
Nice cleanup :+1:

pdmullen (approved):
LGTM; but this is a breaking change for downstream codes.

lroberts36:
> LGTM; but this is a breaking change for downstream codes.

@pdmullen: where does it break things?

Issue #1032, opened 2024-03-25
When outputting a variable, Parthenon also outputs all versions with suffixes


c-prather:
TIL this wasn't an intentional feature.  I agree with the voiced consensus upon discussing it, which is that we should support a more explicit form of this trick.

When outputting variables, one specifies a list:
```
variables = my.variable.1, my.variable.2, another.variable
```
However, if your code contains the variables above, it will actually still output them all if given only the prefix:
```
variables = my.variable, another.variable
```
That is, given `name`, parthenon outputs all variables which match `name*`.

While disgusting, this is actually also really convenient.  By naming variables `prims.rho, prims.u, etc` I can ensure that regardless of which are enabled or the theory I'm evolving, all primitive variables will be output so long as I specify a list `prims, etc, etc`.  However, it's definitely more bug than feature, and can be a problem if you have variables with suffixes (in my case, I discovered it by naming fields `prims.B` and `prims.Bf` for faces).

Thus it seems like we should replace the limited-match behavior with globbing.  i.e., rather than treat `prims` == `prims*`, require the user to write `prims*` if they actually want to match all suffixes.  Note some jerk may also expect support for `prims.*.face` or `*.B*` or other monstrosities (or god forbid regex) so we should decide where the madness stops, and implement/document a suitable compromise.

pgrete:
I also wasn't aware of this "feature" and am in favor of keeping (but documenting/properly implementing it).
The constraint might be to disallow `*` in the variable names (which is probably a good idea anyway).

PR #1043, opened 2024-04-04
Make fluxes separate variables

lroberts36:
## PR Summary
Following # #764, this PR promotes fluxes to be their own variables and then implements flux correction through the regular boundary communication machinery. Now when a variable has the `Metadata::WithFluxes` flag set, a new variable is created with the flags `Metadata::Flux` and `Metadata::OneCopy` and the name `bnd_flux::` + original var name (the prepending is required because of issue #1032). This variable is packed transparently so that packs that are created with fluxes pack associated flux variables automatically and calls to `pack.flux(...)` are unchanged. As a result, this PR should have no impact on downstream codes. 
 
Notes: 

- Allows "fluxes" (in the generalized Stokes theorem sense) for non-cell centered fields. This is fully implemented but *untested*.
- Removes all of the old face flux correction code paths that had to be maintained separately from the boundary values code. 
- Should communicate the same information as was communicated by the previous flux correction routines, i.e. only shared elements are communicated on fine to coarse boundaries, not the entire ghost halo on all boundaries. 
- This may not produce bitwise identical results to the old code, since the order in which field elements are summed during restriction may not be the same as the order that the old flux correction routines did the restriction sums in. Nevertheless, any deviation beyond machine precision is not expected.  
- The flux fields have the same sizes as their parent fields (i.e. the shape of a face flux field in the tuvkji directions is the same as that of the a cell centered field). This is to be consistent with the previous shape of flux fields and to allow for flattened loops that go over more than just the `i` indices of a field (as is commonly done in Riot for instance).   

## PR Checklist

- [x] #1042 is merged into `develop`
- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


lroberts36:
Passes Riot tests.

lroberts36:
Passes parthenon-mhd tests.

example/advection/advection_driver.cpp
```diff
@@ -124,9 +124,7 @@ TaskCollection AdvectionDriver::MakeTaskCollection(BlockList_t &blocks, const in
     auto &mc1 = pmesh->mesh_data.GetOrAdd(stage_name[stage], i);
     auto &mdudt = pmesh->mesh_data.GetOrAdd("dUdt", i);
 
-    auto send_flx = tl.AddTask(none, parthenon::LoadAndSendFluxCorrections, mc0);
-    auto recv_flx = tl.AddTask(none, parthenon::ReceiveFluxCorrections, mc0);
-    auto set_flx = tl.AddTask(recv_flx, parthenon::SetFluxCorrections, mc0);
+    auto set_flx = parthenon::AddFluxCorrectionTasks(none, tl, mc0, pmesh->multilevel);
```
Yurlungur, line 127: :+1:

example/stochastic_subgrid/stochastic_subgrid_package.cpp
```diff
@@ -377,7 +377,7 @@ TaskStatus CalculateFluxes(std::shared_ptr<MeshBlockData<Real>> &rc) {
   const int nx1 = pmb->cellbounds.ncellsi(IndexDomain::entire);
   const int nvar = advected.GetDim(4);
   size_t scratch_size_in_bytes = parthenon::ScratchPad2D<Real>::shmem_size(nvar, nx1);
-  parthenon::ParArray4D<Real> x1flux = rc->Get("advected").flux[X1DIR].Get<4>();
+  parthenon::ParArray4D<Real> x1flux = rc->Get("bnd_flux::advected").data.Get(0, 0, 0);
```
Yurlungur, line 380: Ok what's with the `Get(1, 0, 0)`? Direction 0, but what are the other two 0s for?

tst/unit/test_data_collection.cpp
```diff
@@ -97,7 +97,9 @@ TEST_CASE("Adding MeshBlockData objects to a DataCollection", "[DataCollection]"
           KOKKOS_LAMBDA(const int i) { v2(0) = 222; });
       auto x = d.Add("part", mbd, {"var2", "var3"});
       THEN("Requesting the missing variables should throw") {
-        REQUIRE_THROWS(x->Get("var1"));
+        // This no longer call PARTHENON_REQUIRE_THROWS, it just call PARTHENON_REQUIRE
+        // since throwing does not work nicely with threads
+        // REQUIRE_THROWS(x->Get("var1"));
```
Yurlungur: Not for this PR, but this should be fixed. See, e.g., issue #1046 .

tst/unit/test_meshblock_data_iterator.cpp
```diff
@@ -135,7 +135,10 @@ TEST_CASE("Can pull variables from containers based on Metadata",
       }
     }
 
-    auto v = mbd.PackVariables();
+    using FS_t = Metadata::FlagCollection;
+    auto flags =
+        (FS_t({Metadata::Independent, Metadata::Derived}, true) - FS_t(Metadata::Flux));
+    auto v = mbd.PackVariables(flags);
```
Yurlungur, line 141: HAHA! Kind of happy to see flag collection used here...

src/mesh/meshblock.cpp
```diff
@@ -190,9 +190,8 @@ void MeshBlock::Initialize(int igid, int ilid, LogicalLocation iloc,
   const auto vars =
       real_container->GetVariablesByFlag(flags + FC_t({Metadata::ForceRemeshComm}, true))
           .vars();
-  for (const auto &v : vars) {
-    RegisterMeshBlockData(v);
-  }
+  for (const auto &v : vars)
+    vars_cc_.push_back(v);
```
Yurlungur, line 194: Where is `vars_cc_` still used actually?

src/prolong_restrict/pr_loops.hpp
```diff
@@ -130,16 +130,28 @@ ProlongationRestrictionLoop(const ProResInfoArr_t &info, const Idx_t &buffer_idx
...
+          if (info(buf).include_el[static_cast<int>(TE::F1)])
+            IterateInnerProlongationRestrictionLoop<DIM, Stencil, TE::F1>(
                 team_member, buf, info, ckb, cjb, cib, kb, jb, ib);
-          if (info(buf).fine.topological_type == TopologicalType::Edge)
-            IterateInnerProlongationRestrictionLoop<DIM, Stencil, TE::E3, TE::E2, TE::E1>(
+          if (info(buf).include_el[static_cast<int>(TE::F2)])
+            IterateInnerProlongationRestrictionLoop<DIM, Stencil, TE::F2>(
                 team_member, buf, info, ckb, cjb, cib, kb, jb, ib);
-          if (info(buf).fine.topological_type == TopologicalType::Node)
+          if (info(buf).include_el[static_cast<int>(TE::F3)])
```
Yurlungur: Maybe we should make a method in `BufferInfo` that takes a topological type enum so we don't have to do this static cast "downstream"? might be easier to read.

src/interface/state_descriptor.cpp
```diff
@@ -271,6 +271,20 @@ bool StateDescriptor::AddFieldImpl(const VarID &vid, const Metadata &m_in,
...
+        mFlags.push_back(Metadata::Face);
+      else if (m.IsSet(Metadata::Face))
+        mFlags.push_back(Metadata::Edge);
+      else if (m.IsSet(Metadata::Edge))
+        mFlags.push_back(Metadata::Node);
+      Metadata mf(mFlags, m.Shape());
+      auto fId = VarID{"bnd_flux::" + vid.base_name, vid.sparse_id};
+      AddFieldImpl(fId, mf, control_vid);
+      m.SetFluxName(fId.label());
+    }
```
Yurlungur: I'm not sure why, but I find the location of this logic in `StateDescriptor` a little distasteful. But I don't know of a better place for it.

src/interface/sparse_pack.hpp
```diff
@@ -161,6 +161,13 @@ struct any : public base_t<true> {
   KOKKOS_INLINE_FUNCTION any(Ts &&...args) : base_t<true>(std::forward<Ts>(args)...) {}
   static std::string name() { return ".*"; }
 };
+
+struct any_nonflux : public base_t<true> {
+  template <class... Ts>
+  KOKKOS_INLINE_FUNCTION any_nonflux(Ts &&...args)
+      : base_t<true>(std::forward<Ts>(args)...) {}
+  static std::string name() { return "^(?!bnd_flux::).+"; }
```
Yurlungur: Should it be documented somewhere that fluxes have this name? Actually for output purposes we probably want that too. i.e., we want to be able to output fluxes for debugging. So users should know what the name format is. And also should be forbidden from trying to manually register variables with this pattern.

src/bvals/comms/bvals_utils.hpp
```diff
@@ -103,6 +103,9 @@ void InitializeBufferCache(std::shared_ptr<MeshData<Real>> &md, COMM_MAP *comm_m
   pcache->buf_vec.clear();
   pcache->idx_vec = std::vector<std::size_t>(key_order.size());
   std::for_each(std::begin(key_order), std::end(key_order), [&](auto &t) {
+    if (comm_map->count(std::get<2>(t)) == 0) {
+      PARTHENON_FAIL("Asking for buffer that doesn't exist");
+    }
```
Yurlungur, line 108: What's going on here?

src/interface/mesh_data.cpp
```diff
@@ -25,19 +25,14 @@ void MeshData<T>::Initialize(const MeshData<T> *src,
...
-    }
-  } else {
-    for (int i = 0; i < nblocks; i++) {
-      block_data_[i] = pmy_mesh_->block_list[i]->meshblock_data.Add(
-          stage_name_, src->GetBlockData(i), names, shallow);
-    }
+  PARTHENON_REQUIRE(nblocks == bl.size(), "Somehow these sizes don't agree.");
+  for (int i = 0; i < nblocks; i++) {
+    block_data_[i] =
+        bl[i]->meshblock_data.Add(stage_name_, src->GetBlockData(i), names, shallow);
```
Yurlungur: :+1:

Yurlungur (approved):
This is a really nice cleanup. Love to see the whole flux correction code get removed. That said, this probably merits some documentation.

example/stochastic_subgrid/stochastic_subgrid_package.cpp
lroberts36: I believe `ParArrayND::Get<D>` returns a `D` dimensional sub array where the first `N-D` indices of the original array are set to zero. `ParArrayND::Get(int... args)` returns and `N-sizeof...(args)` dimension array with the leading indices set to `args...`. Because the leading index is the topological element index (i.e. flux direction index), we need to change a little bit about how we pull out the subarray.

tst/unit/test_data_collection.cpp
lroberts36: I know all too well... don't ask me why this PR took a week to get done instead of a day and a half.

src/mesh/meshblock.cpp
lroberts36: For communicating data during remeshing.

src/bvals/comms/bvals_utils.hpp
lroberts36: This was part of a debugging "exercise" and I saw no reason to remove it once things were fixed. It is just checking that a comm buffer corresponding to the requested key exists. I think previously the map would throw an error if you asked for a non-existent key, but now we can't trust exceptions.

Yurlungur: :+1:

example/stochastic_subgrid/stochastic_subgrid_package.cpp
Yurlungur: Yeesh I should have known that... since I wrote `ParArrayND::Get`...

src/interface/sparse_pack.hpp
lroberts36: yeah, maybe we should disallow `::` in variable names? That would give us some flexibility for creating more auto-generated variables. Also `bnd_flux` was just what Josh had chosen in the his old PR making fluxes variables. We could change that.

Also, I guess there is no rule that fluxes have to be named `bnd_flux::*`. A downstream code could declare a field with `Metadata::Flux` with any name, not explicitly associate it with any other field, and it would still get communicated during the flux correction routines. Maybe this should read `any_nonautomaticflux`.

src/interface/state_descriptor.cpp
lroberts36: The only other place I could think is to add a static method `Metadata::MakeFluxFieldMetadata(const Metadata& in)`. But the calls to `AddFieldImpl` and `SetFluxName` would still probably need to be here.

Yurlungur: Yeah... That feels a little better to me, but I honestly can't express why. I think this is fine as is for now.

src/interface/sparse_pack.hpp
Yurlungur: I am in favor of disallowing `::` in variable names. And agree about `any_nonautoflux`

c-prather:
I had thought one of the advantages of fluxes as fields would be consistency: sizeof(flux(cell)) == sizeof(face) and so on.  Always making them the same size as parent, rather than the usual size for their new location, feels a bit inconsistent -- not to mention that some stuff like CT might need that last face/edge, or else need to use an extra ghost cell.  Maybe we should have an option whether to declare the flux smaller for backward compatibility/easier multi-D indexing, or larger for consistency with the other variables which live in the same place? `Metadata::FluxCompat`?

lroberts36:
@bprather: This should have no impact for consistency of face fluxes/edges or edge fluxes/nodes, which are all the same size in memory (3 or 1, nt, nu, nv, nz + 2 * nghost + 1, ny + 2 * nghost + 1, nx + 2 * nghost + 1). It only matters for cells fluxes vs face fields. I am not aware of a case where that one missing layer of ghosts on the positive sides of the block are used, since the flux arrays should never get their ghosts filled (which maybe should be enforced in the metadata to prevent fields with both `Metadata::FillGhost` and `Metadata::Flux`?).

c-prather:
Ah, fair enough, I didn't consider that the memory would be the same for face field/edge flux variables anyway.

Re: ghosts, sometimes KHARMA fills EMFs to the last ghost edges of domain boundaries, and expects a consistent face value, for e.g. preparing Dirichlet/frozen boundary conditions.  But, this should be fine since the edge and face variables will both be (N+1)^3.

I'm not aware of anyone requiring the last face value of any face-centered variable which could conceivably be marked a flux.  So I guess the concern is purely aesthetic, and in any case non-blocking.

lroberts36:
@Yurlungur: Some documentation added/fixed to be consistent.

src/interface/sparse_pack.hpp
lroberts36: renamed `any_nonflux` but keeping this conversation open to hear others thoughts on reserving `::`

src/prolong_restrict/pr_loops.hpp
lroberts36: done

lroberts36:
The issue with the `riot` tests mentioned during the Parthenon call today has been figured out. It had to do with a kernel grabbing all variables in a `MeshData` object and zeroing them out. 

Because fluxes are now variables, a change to how `MeshData` and `MeshBlockData` are built when `DataCollection::Add(const std::string &name, const std::shared_ptr<T> &src, const std::vector<ID_t> &fields, const bool shallow)` is called was required for backward compatibility. Rather than just add the variables contained in `fields`, we now add those variables and the flux variables associated with them. This ensures that when packs are created from these `MeshData` and `PDOpt::WithFluxes` is requested, the fluxes are in fact available to pack. 

The issue in Riot was that a `MeshData` object was created through `DataCollection::Add`, then a pack was created from this `MeshData` using `variable_names::any`, and then all of the fields were zeroed before they were accumulated into. This meant that the one-copy fluxes were zeroed during this call since they were automatically included in the `MeshData` and then sucked up by `any`. The solution chosen to fix this was to make `variable_names::any` only select variables that didn't include `bnd_flux::` in their name. I am not sure this is the best way to deal with it, but that should fix things for now.

doc/sphinx/src/interface/metadata.rst
```diff
@@ -130,6 +130,8 @@ variables are copied or not in multiple stages.
    must be filled via communication or boundary conditions. This is not
    always required. ``OneCopy`` variables, for example, may not need
    this.
+-  ``Metadata::Flux`` specifies that elements shared with neighbor blocks 
+   at coarse-fine boundaries are communicated during flux correction.
```
pgrete: "just" communicated or also "corrected"?

doc/sphinx/src/interface/metadata.rst
```diff
@@ -143,10 +145,23 @@ classes may be allocated. The behaviours are the following:
...
--  If ``Metadata::WithFluxes`` is set, the flux vector for the variable
-   is allocated. Note that it is necessary to set both
-   ``Metadata::WithFluxes`` and ``Metadata::FillGhosts`` to send flux
-   corrections across meshblocks.
+-  If ``Metadata::WithFluxes`` is set, a new one-copy variable with the 
+  correct topological type for a flux in the generalized Stokes theorem 
+  sense (e.g. if the ``WithFluxes`` variable has ``Metadata::Cell`` 
+  set the new variable will have ``Metadata::Face``) will be created in
+  the package with the name ``bnd_flx::name_of_original_variable`` and 
+  `Metadata::Flux` and `Metadata::OneCopy``. When creating packs that
```
pgrete:
```suggestion
  ``Metadata::Flux`` and ``Metadata::OneCopy``. When creating packs that 
```

example/particle_tracers/particle_tracers.cpp
```diff
@@ -265,8 +265,7 @@ TaskStatus CalculateFluxes(MeshBlockData<Real> *mbd) {
 
   auto advected = mbd->Get("advected").data;
 
-  auto x1flux = mbd->Get("advected").flux[X1DIR].Get<4>();
-
+  parthenon::ParArray4D<Real> x1flux = mbd->Get("bnd_flux::advected").data.Get(1, 0, 0);
```
pgrete: This should be `Get(0,0,0)` doesn't it?
And if so, what's this test doing if it passes when using the wrong fluxes?

Separately, (along with @Yurlungur similar comment on `Get` below), what about a more intuitive interface like
GetFlux(varname, direction) that hides the flux name and extra `0` indices?

src/bvals/comms/bnd_info.cpp
```diff
@@ -219,43 +248,49 @@ int GetBufferSize(MeshBlock *pmb, const NeighborBlock &nb,
...
+    idx++;
   }
-  return out;
+}
+
+BndInfo BndInfo::GetSendBndInfo(MeshBlock *pmb, const NeighborBlock &nb,
+                                std::shared_ptr<Variable<Real>> v,
+                                CommBuffer<buf_pool_t<Real>::owner_t> *buf) {
+  auto idx_range_type = IndexRangeType::BoundaryInteriorSend;
+  if (nb.offsets.IsCell()) idx_range_type = IndexRangeType::InteriorSend;
```
pgrete: I'd be great to get a comment in the code here, what's going on.
I tried to guess, but didn't come up with sth (same for the following `GetSetBndInfo` function).

src/bvals/comms/boundary_communication.cpp
```diff
@@ -422,4 +433,12 @@ AddBoundaryExchangeTasks<BoundaryType::any>(TaskID, TaskList &,
 template TaskID
 AddBoundaryExchangeTasks<BoundaryType::gmg_same>(TaskID, TaskList &,
                                                  std::shared_ptr<MeshData<Real>> &, bool);
+
+TaskID AddFluxCorrectionTasks(TaskID dependency, TaskList &tl,
+                              std::shared_ptr<MeshData<Real>> &md, bool multilevel) {
+  if (!multilevel) return dependency;
```
pgrete, line 439: I see how this makes the code path in the driver cleaner (wrt to dependencies) but somehow passing a variable in and just use it to immediately return feels odd (just commenting, not saying this should change).

src/interface/data_collection.hpp
```diff
@@ -49,7 +49,7 @@ class DataCollection {
                           const std::vector<ID_t> &fields, const bool shallow) {
     auto it = containers_.find(name);
     if (it != containers_.end()) {
-      if (fields.size() && !(it->second)->ContainsExactly(fields)) {
+      if (fields.size() && !(it->second)->Contains(fields)) {
```
pgrete, line 52: Why this change? IIRC this line was recently touched by the uuid PR.

src/interface/meshblock_data.hpp
```diff
@@ -134,7 +143,18 @@ class MeshBlockData {
...
+        add_var(var);
+        // Add the associated flux as well if not explicitly
+        // asked for
+        if (var->IsSet(Metadata::WithFluxes)) {
+          auto flx_name = var->metadata().GetFluxName();
+          bool found = false;
+          for (const auto &v2 : vars) {
+            if (src->GetVarPtr(v2)->label() == flx_name) found = true;
+          }
+          if (!found) add_var(src->GetVarPtr(flx_name));
```
pgrete, line 156: Do we need to check if the found variable has certain properties (required and incompatible -- like `OneCopy` or similar)?

src/interface/metadata.cpp
```diff
@@ -331,11 +334,12 @@ Metadata::GetArrayDims(std::weak_ptr<MeshBlock> wpmb, bool coarse) const {
     // This variable is not necessarily tied to any specific
     // mesh element, so dims will be used as the actual array
     // size in each dimension
-    assert(N >= 1 && N <= MAX_VARIABLE_DIMENSION);
+    assert(N >= 1 && N < MAX_VARIABLE_DIMENSION);
```
pgrete, line 337: Was that a bug?

src/interface/metadata.cpp
```diff
@@ -331,11 +334,12 @@ Metadata::GetArrayDims(std::weak_ptr<MeshBlock> wpmb, bool coarse) const {
...
     for (int i = 0; i < N; i++)
       arrDims[i] = shape[i];
     for (int i = N; i < MAX_VARIABLE_DIMENSION; i++)
       arrDims[i] = 1;
+    if (IsSet(Flux)) arrDims[MAX_VARIABLE_DIMENSION - 1] = 3;
```
pgrete, line 342: Does this imply that we can have a flux that is not mesh tied?

src/interface/sparse_pack.hpp
pgrete: The `::` may or may not cause a problem as it's adisallowed character by the OpenPMD standard.
I'm actually not sure if this is a hard problem or not (as I assume that some default tools may not work with that approach, but we probably would dump these flux variable only for debugging anyway, don't we)?
Also, we could in principle overwrite the reserved pattern for output and do a name matching when reading them back in again.

src/interface/sparse_pack.hpp
```diff
@@ -156,11 +156,21 @@ struct base_t {
...
 };
+
+struct any_nonautoflux : public base_t<true> {
+  template <class... Ts>
+  KOKKOS_INLINE_FUNCTION any_nonautoflux(Ts &&...args)
+      : base_t<true>(std::forward<Ts>(args)...) {}
+  static std::string name() { return "^(?!bnd_flux::).+"; }
+};
+
+using any = any_nonautoflux;
```
pgrete, line 175: May that have some unintended side-effects?

src/interface/state_descriptor.cpp
```diff
@@ -271,6 +271,20 @@ bool StateDescriptor::AddFieldImpl(const VarID &vid, const Metadata &m_in,
...
+      std::vector<MetadataFlag> mFlags = {Metadata::OneCopy, Metadata::Flux};
+      if (m.IsSet(Metadata::Sparse)) mFlags.push_back(Metadata::Sparse);
+      if (m.IsSet(Metadata::Cell))
+        mFlags.push_back(Metadata::Face);
+      else if (m.IsSet(Metadata::Face))
+        mFlags.push_back(Metadata::Edge);
+      else if (m.IsSet(Metadata::Edge))
+        mFlags.push_back(Metadata::Node);
+      Metadata mf(mFlags, m.Shape());
+      auto fId = VarID{"bnd_flux::" + vid.base_name, vid.sparse_id};
```
pgrete: Should the prefix potentially be a const string at a more central location?

src/mesh/mesh.cpp
```diff
@@ -1253,23 +1250,14 @@ void Mesh::SetupMPIComms() {
...
       PARTHENON_REQUIRE_THROWS(ret.second, "Communicator with same name already in map");
-
-      if (multilevel) {
-        MPI_Comm mpi_comm_flcor;
-        PARTHENON_MPI_CHECK(MPI_Comm_dup(MPI_COMM_WORLD, &mpi_comm_flcor));
-        const auto ret =
-            mpi_comm_map_.insert({pair.first.label() + "_flcor", mpi_comm_flcor});
-        PARTHENON_REQUIRE_THROWS(ret.second,
-                                 "Flux corr. communicator with same name already in map");
-      }
```
pgrete, line 1272: Not required any more because as separate vars they already have their separate comms, correct?

tst/unit/test_data_collection.cpp
pgrete: Should this be reintroduced now?

pgrete (approved):
Nice cleanup!
I just got a couple of interface/doc comments -- nothing big.

tst/unit/test_data_collection.cpp
lroberts36: yep. It is now reintroduced.

src/interface/sparse_pack.hpp
lroberts36: Ok, I added a global string `internal_varname_separator = "::"` and another global string `internal_fluxname = "bnd_flux"` and no longer explicitly write `bnd_flux::` anywhere (except in tests). This should make it easy to change the internal naming scheme if it becomes necessary for OpenPMD.

doc/sphinx/src/interface/metadata.rst
lroberts36: added corrected for clarity.

doc/sphinx/src/interface/metadata.rst
lroberts36: fixed in another commit

example/particle_tracers/particle_tracers.cpp
lroberts36: Yes, it is disturbing that this test passes with that bug. I am not really sure what it is supposed to be doing though... 

The `Get` function called here is really `ParArrayGeneric::Get(Args... args)`, so I don't think we can reasonably make a flux based overload. I don't think we really need to support it from `MeshBlockData` either, since I don't think we want to encourage the pattern that is used here (variables should be accessed through packs rather than pulling out raw `ParArray`s directly from `MeshBlockData`). That being said, I agree that the `ParArrayGeneric::Get` interface is a bit confusing and should possibly be changed. That is for another time though I think...

src/bvals/comms/bnd_info.cpp
lroberts36: comment added.

src/interface/data_collection.hpp
lroberts36: This is sort of related to the bugfix for Riot described in the comment above. Since we now automatically add flux variables associated with the variables in `fields`, `ContainsExactly` now fails when called with the original `fields` since there are extra flux fields in `it->second`.

src/interface/meshblock_data.hpp
lroberts36: I don't think that we want to do that, at least for backward compatibility. If a variable is in a `MeshBlockData` object, it's flux should be available there as well. At least that was my thinking.

src/interface/metadata.cpp
lroberts36: I think so.

src/interface/metadata.cpp
lroberts36: yeah, although this is completely untested and it is not clear to me that it makes sense for non-mesh tied variables to have fluxes.

src/interface/sparse_pack.hpp
lroberts36: I don't think so, since there are few places where `any` is currently used and those places assumed that fluxes would not get sucked up when using `any`. I think in the future we, should avoid using `any` and use either `any_nonautoflux` or `any_withautoflux`.

src/interface/state_descriptor.cpp
lroberts36: good point, changed.

src/mesh/mesh.cpp
lroberts36: yep

src/bvals/comms/boundary_communication.cpp
lroberts36: I tend to like this better than not taking `multilevel` as an argument and having people include 
```c++
auto flx_cor = prev_task;
if (multilevel) {
  flx_cor = AddFluxCorrectionTasks(flx_cor, ...); 
}
```

src/interface/sparse_pack.hpp
lroberts36: Checking that people don't add variables with `::` is considerably harder because of how all the package resolution works. I went down that path for a little while and eventually gave up.

PR #1049, opened 2024-04-11
Catch task failures from threads

### LINKED ISSUE #1046: Fatal errors in `Task`s are not propagated upward
Was debugging an issue where a task wasn't executing, even when explicitly added to the task list, verified to be in the resulting list, etc.  I could execute anything if it was marked depending on `TaskID(0)`, but not if it depended on other tasks.

The problem was that a previous task was throwing a `runtime_error`.  Error thrown, thread exits, thread pool starts up another without any pending tasks.  `pool.wait` finishes, execution continues silently.

I don't imagine this is really what we intend, or tasks become very dangerous to use.  We should check on the thread pool (however that looks with `std::threads`, I'm no expert) rather than just moving on when it dies with `wait()`.

c-prather:
Previously, due to using a worker thread, Parthenon would drop any exceptions from Tasks executed in a TaskList.  Exceptions would crash the thread, but execution would continue as if it had finished normally.

This PR checks the futures associated with each task, to make sure there are not exceptions.  As a bonus, we can add a potentially non-fatal failure return TaskStatus::fail, which will be propagated up to the Driver.

This has in the past failed the `sparse_advection` tests, but that may have been due to incidental changes over the course of debugging and testing.  It seems to pass all tests reliably now, 3-4 times in a row (except a couple of gmg tests which it also fails on my machine even with threading disabled).  Should there turn out to still be a rare race condition, I've extended `ThreadVector` such that vectors of `Task`s or `TaskList`s could be swapped over to be thread-safe transparently.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/src/tasks/tasks.hpp
+++ b/src/tasks/tasks.hpp
@@ -1,5 +1,5 @@
 //========================================================================================
-// (C) (or copyright) 2023. Triad National Security, LLC. All rights reserved.
+// (C) (or copyright) 2023-2024. Triad National Security, LLC. All rights reserved.
 //
 // This program was produced under U.S. Government contract 89233218CNA000001 for Los
 // Alamos National Laboratory (LANL), which is operated by Triad National Security, LLC
--- a/src/tasks/tasks.hpp
+++ b/src/tasks/tasks.hpp
@@ -32,7 +32,7 @@
 
 namespace parthenon {
 
-enum class TaskListStatus { complete }; // doesn't feel like we need this...
+enum class TaskListStatus { complete, fail };
 enum class TaskType { normal, completion };
 
 class TaskQualifier {
@@ -423,7 +423,10 @@ class TaskRegion {
     // then wait until everything is done
     pool.wait();
 
-    return TaskListStatus::complete;
+    // Check the results, so as to fire any exceptions from threads
+    // Return failure if a task failed
+    return (pool.check_task_returns() == TaskStatus::complete) ?
+            TaskListStatus::complete : TaskListStatus::fail;
   }
 
   TaskList &operator[](const int i) { return task_lists[i]; }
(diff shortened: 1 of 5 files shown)


src/tasks/tasks.hpp
```diff
@@ -423,7 +423,10 @@ class TaskRegion {
     // then wait until everything is done
     pool.wait();
 
-    return TaskListStatus::complete;
+    // Check the results, so as to fire any exceptions from threads
+    // Return failure if a task failed
+    return (pool.check_task_returns() == TaskStatus::complete) ? TaskListStatus::complete
+                                                               : TaskListStatus::fail;
```
Yurlungur, line 429: This is clever.

Yurlungur (approved):

src/tasks/thread_pool.hpp
```diff
@@ -89,6 +91,37 @@ class ThreadQueue {
   bool waiting = false;
 };
 
+template <typename T>
+class ThreadVector {
```
lroberts36, line 95: I guess this is just a thread safe vector? If so, maybe a short comment clarifying it would be helpful.

src/tasks/thread_pool.hpp
```diff
@@ -117,21 +150,37 @@ class ThreadPool {
   void kill() { queue.signal_kill(); }
 
   template <typename F, class... Args>
-  std::future<typename std::result_of<F(Args...)>::type> enqueue(F &&f, Args &&...args) {
+  void enqueue(F &&f, Args &&...args) {
     using return_t = typename std::result_of<F(Args...)>::type;
     auto task = std::make_shared<std::packaged_task<return_t()>>(
         [=, func = std::forward<F>(f)] { return func(std::forward<Args>(args)...); });
-    std::future<return_t> result = task->get_future();
+    // If we're listing Prathenon "Tasks" (all current uses) keep the returns
```
lroberts36, line 157: I see you are trying to remake Parthenon in your image...

lroberts36 (approved):
LGTM, thanks for the fix!

PR #1052, opened 2024-04-12
find_package(Threads) on install

pbrady:
Runs `find_package(Threads)` when consuming installed cmake target.

Fixes issue with missing `Threads::Threads` target.


Yurlungur (approved):
Trivial so I don't think it needs a changelog. Why is formatting CI test failing though?

Yurlungur:
@pgrete CI seems to be failing at the initialize containers stage.

pgrete (approved):

pgrete:
> @pgrete CI seems to be failing at the initialize containers stage.

This was a driver/library mismatch on the CI machine. Should be fine now.

PR #1054, opened 2024-04-15
pre-post receives

MoraruMaxim:
Fixing the issue of too many unexpected messages on Slingshot (only works with non-sparse fields).


Yurlungur (approved):
Just to double check, this won't interfere with sparse will it @lroberts36 ?

lroberts36:
@Yurlungur: This will not interfere with sparse fields, but calls to `TryStartReceive` don't pre-post ireceives for sparse fields (since the message size is apriori unknown).

PR #1055, opened 2024-04-16
Refactor Mesh Constructors

lroberts36:
## PR Summary

This PR splits up the `Mesh` constructors to remove code duplication between the regular and restart mesh constructors. The main rationale for doing this is to split up the constructors so that new constructors for non-hyper-rectangular meshes can easily be added.

In the process of going through this PR, I noticed that some of the user function overrides were not set in the restart constructor. **Someone please confirm that it is ok to set all of these on restart.**

Almost all of the changes should be just a reorganization of pre-existing code. One exception is that on restart the base mesh is built from the parameter input file, and then the base mesh properties are checked against the restart file (as opposed to directly building from the restart file information). I think there is no reason why the parameter input should be able to change the mesh properties on restart, but please correct me if I am wrong. 

## PR Checklist

- [x] #1043 is merged
- [x] #1071 is merged
- [x] Code passes cpplint
- [x] Code is formatted


lroberts36:
Passes `Riot` and `parthenon-mhd` tests.

src/mesh/mesh.hpp
```diff
@@ -74,9 +74,15 @@ class Mesh {
...
   friend class MeshBlock;
-  friend class MeshBlockTree;
   friend class MeshRefinement;
 
+  struct base_constructor_selector_t {};
+  Mesh(ParameterInput *pin, ApplicationInput *app_in, Packages_t &packages,
+       base_constructor_selector_t);
+  struct hyper_rectangular_constructor_selector_t {};
+  Mesh(ParameterInput *pin, ApplicationInput *app_in, Packages_t &packages,
+       hyper_rectangular_constructor_selector_t);
```
Yurlungur, line 84: tag dispatch. I like it.

Yurlungur (approved):
This is a nice cleanup. :+1: I don't know the answer about the function hooks so I defer to @pgrete on this one but what you did seems reasonable to me.

pdmullen (approved):
LGTM

lroberts36:
`riot` and `parthenon-mhd` tests still passing.

pgrete:
It looks like the extended tests don't pass (any more?)

lroberts36:
Ah, there was a bug. I was mapping the derefinement counts based on the legacy logical location and then trying to find them based on the forest logical location (so effectively every derefinement count was set to zero). I think this wasn't showing up previously because the test was still setting a very high number for the derefinement_count until the latest PR was merged in. Should all be fixed now.

pgrete (approved):
Further testing and debugging point towards the issue I'm seeing at scale being independent of this PR. So let's get this in as all other tests at smaller scale pass.

PR #1057, opened 2024-04-16
Fix history output after restarts

### LINKED ISSUE #1056: Restart with new history output
When running the advection example with the older input file (that's missing the block for the new history output), I get a crash on restart
```
cycle=224 time=3.9375000000000154e-01 dt=1.7578125000000000e-03 zone-cycles/wsec_step=4.47e+06 wsec_total=2.44e+00 wsec_step=5.98e-02 zone-cycles/wsec=4.47e+06 wsec_AMR=1.53e-05
cycle=225 time=3.9550781250000155e-01 dt=1.7578125000000000e-03 zone-cycles/wsec_step=4.39e+06 wsec_total=2.50e+00 wsec_step=6.09e-02 zone-cycles/wsec=4.39e+06 wsec_AMR=1.49e-05
cycle=226 time=3.9726562500000157e-01 dt=1.7578125000000000e-03 zone-cycles/wsec_step=4.42e+06 wsec_total=2.56e+00 wsec_step=6.05e-02 zone-cycles/wsec=4.42e+06 wsec_AMR=1.55e-05

cycle=227 time=3.9902343750000158e-01 dt=1.7578125000000000e-03 zone-cycles/wsec_step=4.46e+06 wsec_total=2.62e+00 wsec_step=5.99e-02 zone-cycles/wsec=4.46e+06 wsec_AMR=1.49e-05
### PARTHENON ERROR
  Message:     History output for package "" requested but package is not available!

  File:        /home/pgrete/src/parthenon/src/outputs/history.cpp
  Line number: 77
History output for package "" requested but package is not available!

Backtrace:
                  [0x5f929523da85]
                  [0x5f9295235caa]
                  [0x5f9295235cdb]
                  [0x5f92950c27a0]
                  [0x5f92951f4351]
                  [0x5f92950e4e28]
                  [0x5f9295049011]
                  [0x5f9294f585e6]
                  [0x7e2c2da43cd0]
__libc_start_main [0x7e2c2da43d8a]
                  [0x5f9294f5b5f5]
[haerke:212372] *** Process received signal ***
[haerke:212372] Signal: Aborted (6)
[haerke:212372] Signal code:  (-6)
[haerke:212372] [ 0] /usr/lib/libc.so.6(+0x3c770)[0x7e2c2da5a770]
[haerke:212372] [ 1] /usr/lib/libc.so.6(+0x8d32c)[0x7e2c2daab32c]
[haerke:212372] [ 2] /usr/lib/libc.so.6(gsignal+0x18)[0x7e2c2da5a6c8]
[haerke:212372] [ 3] /usr/lib/libc.so.6(abort+0xd7)[0x7e2c2da424b8]
[haerke:212372] [ 4] /home/pgrete/src/parthenon/build-develop/example/advection/advection-example(+0x306ce0)[0x5f9295235ce0]
[haerke:212372] [ 5] /home/pgrete/src/parthenon/build-develop/example/advection/advection-example(+0x1937a0)[0x5f92950c27a0]
[haerke:212372] [ 6] /home/pgrete/src/parthenon/build-develop/example/advection/advection-example(+0x2c5351)[0x5f92951f4351]
[haerke:212372] [ 7] /home/pgrete/src/parthenon/build-develop/example/advection/advection-example(+0x1b5e28)[0x5f92950e4e28]
[haerke:212372] [ 8] /home/pgrete/src/parthenon/build-develop/example/advection/advection-example(+0x11a011)[0x5f9295049011]
[haerke:212372] [ 9] /home/pgrete/src/parthenon/build-develop/example/advection/advection-example(+0x295e6)[0x5f9294f585e6]
[haerke:212372] [10] /usr/lib/libc.so.6(+0x25cd0)[0x7e2c2da43cd0]
[haerke:212372] [11] /usr/lib/libc.so.6(__libc_start_main+0x8a)[0x7e2c2da43d8a]
[haerke:212372] [12] /home/pgrete/src/parthenon/build-develop/example/advection/advection-example(+0x2c5f5)[0x5f9294f5b5f5]
[haerke:212372] *** End of error message ***

```

I haven't investigated yet, just reporting that it exists.

brryan:
## PR Summary

@pgrete identified a bug in my recent history PR that caused history outputs, that use the default case of outputting all packages' histories, to fail after restarts. 

This was because I used `pkgs = pin->GetOrAddVector(block, packages, std::vector())` to get the names of packages, and I checked if `pkgs` was empty to determine whether to do the default case of outputting all packages. Well, if you do `GetOrAdd` like that, the parameters class will tell the restart file that you actually had an input file that looked like:
```
<parthenon/output3>
packages=
```
which, after restart, `GetOrAddVector` will interpret as a length-1 vector that contains `""`. 

I addressed this by first checking for the presence of `packages` in the output block, and only if it exists calling `GetVector`. My initial opinion is that this is *not* a problem with the `GetOrAddVector` implementation but I'm curious if others think we should modify that behavior. 

Fixes #1056 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [ ] Docs build

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - [[PR 1004]](https://github.com/parthenon-hpc-lab/parthenon/pull/1004) Allow parameter modification from an input file for restarts
 
 ### Fixed (not changing behavior/API/variables/...)
+- [[PR 1057]](https://github.com/parthenon-hpc-lab/parthenon/pull/1057) Fix history output after restarts
 - [[PR 1024]](https://github.com/parthenon-hpc-lab/parthenon/pull/1024) Add features to history output
 - [[PR 1031]](https://github.com/parthenon-hpc-lab/parthenon/pull/1031) Fix bug in non-cell centered AMR
 
--- a/src/outputs/history.cpp
+++ b/src/outputs/history.cpp
@@ -62,7 +62,15 @@ void HistoryOutput::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm,
   // If "packages" not provided, output history for all packages
   auto &requested_packages = output_params.packages;
   Dictionary<std::shared_ptr<StateDescriptor>> packages;
+
+    printf("block: %s\n", output_params.block_name.c_str());
+    printf("All my packages???\n");
+    for (const auto &pkg_name : requested_packages) {
+      printf("  %s\n", pkg_name.c_str());
+    }
+
   if (requested_packages.empty()) {
+    printf("Here??\n");
     packages = pm->packages.AllPackages();
   } else {
     const auto &all_packages = pm->packages.AllPackages();
--- a/src/outputs/history.cpp
+++ b/src/outputs/history.cpp
@@ -62,7 +62,6 @@ void HistoryOutput::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm,
   // If "packages" not provided, output history for all packages
   auto &requested_packages = output_params.packages;
   Dictionary<std::shared_ptr<StateDescriptor>> packages;
-
   if (requested_packages.empty()) {
     packages = pm->packages.AllPackages();
   } else {
--- a/src/outputs/history.cpp
+++ b/src/outputs/history.cpp
@@ -63,14 +63,7 @@ void HistoryOutput::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm,
   auto &requested_packages = output_params.packages;
   Dictionary<std::shared_ptr<StateDescriptor>> packages;
 
-    printf("block: %s\n", output_params.block_name.c_str());
-    printf("All my packages???\n");
-    for (const auto &pkg_name : requested_packages) {
-      printf("  %s\n", pkg_name.c_str());
-    }
-
   if (requested_packages.empty()) {
-    printf("Here??\n");
     packages = pm->packages.AllPackages();
   } else {
     const auto &all_packages = pm->packages.AllPackages();
--- a/src/outputs/outputs.cpp
+++ b/src/outputs/outputs.cpp
@@ -200,7 +200,8 @@ Outputs::Outputs(Mesh *pm, ParameterInput *pin, SimTime *tm) {
       }
 
       if (op.file_type == "hst") {
-        // Do not use GetOrAddVector because it will pollute the input parameters for restarts
+        // Do not use GetOrAddVector because it will pollute the input parameters for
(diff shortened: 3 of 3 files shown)


Yurlungur (approved):
LGTM. This is a bit unintuitive---dunno how I feel about how this interacts with `GetOrAdd`. For now I agree we leave it though.

pgrete:
> LGTM. This is a bit unintuitive---dunno how I feel about how this interacts with `GetOrAdd`. For now I agree we leave it though.

I think I know what you mean.

How about fixing it at the places that processes the "packages" var, i.e., if it's empty default back to "all packages" (as if the params was not defined at all). Soft disabling would still be possible by setting a negative `dt`.

pgrete:
Forgot to say: thanks for the quick investigation and fix!

brryan:
> > LGTM. This is a bit unintuitive---dunno how I feel about how this interacts with `GetOrAdd`. For now I agree we leave it though.
> 
> I think I know what you mean.
> 
> How about fixing it at the places that processes the "packages" var, i.e., if it's empty default back to "all packages" (as if the params was not defined at all). Soft disabling would still be possible by setting a negative `dt`.

This is what I did initially in the history output, basically
```c++
auto pkgs = output+params.packages;
if (pkgs.size() == 1) {
  if (pkgs[0] == "") {
    pkgs = std::vector<std::string>();
  }
}
```
but that just felt a little bit fragile to me. I guess I have a preference to not set a vector parameter to an empty default because on restart that empty default is read back in as something that is not an empty default. Maybe the way we regenerate parameter files for restarts should be retooled a bit so an empty default in `GetOrAddVector` will still return that empty default after restart? I don't know if we can/should always read
`packages=`
as an empty vector rather than the one-element `{""}`. 

I'm also not quite sure how a negative `dt` solves this, since we do want the history output during initial run and during restarts where it was failing. Maybe I misunderstand?

> Forgot to say: thanks for the quick investigation and fix!

Thanks for finding this so quickly! Sorry I didn't even think to check this case.

pgrete:
Oh, I didn't realize that `GetOrAddVector` does not support an empty default vector.
My comment was made under that assumption.
I agree that this would be nice to be fixed/reworked, but this is beyond this PR.
In that case, the current solution looks good!

Regarding restarts, what I was trying to say is that using an empty list of packages as default to "all" packages rather than "none" would be reasonable as the equivalent to "none" would be to disable that particular history output (which could equally be achieved via setting a negative `dt`).

pgrete (approved):
Solution looks good give other constraints (see comment in the main thread).

PR #1058, opened 2024-04-17
Vector history not being output if no scalar history present

brryan:
## PR Summary

Running vector history downstream, I just ran into another bug: If a package *only* has vector history outputs enrolled, these will be ignored by an earlier `continue` statement in the loop over history variables triggered by a lack of scalar history outputs. 

Trivial fix, works downstream for me.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [ ] Docs build

### DIFF
--- a/src/outputs/history.cpp
+++ b/src/outputs/history.cpp
@@ -85,22 +85,21 @@ void HistoryOutput::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm,
 
     // Get "base" MeshData, which always exists but may not be populated yet
     auto &md_base = pm->mesh_data.Get();
+    // Populated with all blocks
+    if (md_base->NumBlocks() == 0) {
+      md_base->Set(pm->block_list, pm);
+    } else if (md_base->NumBlocks() != pm->block_list.size()) {
+      PARTHENON_WARN(
+          "Resetting \"base\" MeshData to contain all blocks. This indicates that "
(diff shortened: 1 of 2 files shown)


Yurlungur (approved):

src/outputs/history.cpp
```diff
@@ -102,24 +96,28 @@ void HistoryOutput::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm,
       md_base->Set(pm->block_list, pm);
     }
 
-    for (const auto &hist_var : hist_vars) {
-      auto result = hist_var.hst_fun(md_base.get());
-      results[hist_var.hst_op].push_back(result);
-      labels[hist_var.hst_op].push_back(hist_var.label);
+    // Check if the package has enrolled scalar history functions which are stored in the
+    // Params under the `hist_param_key` name.
+    if (params.hasKey(hist_param_key)) {
```
brryan, line 101: Only potential change in behavior is that, even if no history vars are enrolled, we still update `md_base` if it was empty or if it had a different size than `pm->block_list`. I think this is fine because codes shouldn't be designed around not updating this variable during history output *only* if no history variables were requested.

I also now update `md_base` only once for all packages rather than inside the `for (const auto &pkg : packages)` loop.

pgrete (approved):
LGTM straightforward fix

PR #1060, opened 2024-04-18
Thread UIDs through `data_collection`

Yurlungur:
## PR Summary

This PR adds the ability to request a new `MeshData` or `MeshBlockData` object by requesting a subset of unique IDs of variables from the parent object, rather than by using only variable names. This should make UID-based operations more usable and thus unlock a bit of performance by minimizing string comparisons. Needed in riot and I suspect also KHARMA.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/src/interface/data_collection.hpp
+++ b/src/interface/data_collection.hpp
@@ -49,7 +49,7 @@ class DataCollection {
                           const std::vector<ID_t> &fields, const bool shallow) {
     auto it = containers_.find(name);
     if (it != containers_.end()) {
-      if ((fields.size() > 0) && !(it->second)->ContainsExactly(fields)) {
+      if (fields.size() && !(it->second)->ContainsExactly(fields)) {
         PARTHENON_THROW(name + " already exists in collection but fields do not match.");
       }
       return it->second;
--- a/src/interface/data_collection.hpp
+++ b/src/interface/data_collection.hpp
@@ -19,6 +19,8 @@
 #include <string>
 #include <vector>
 
+#include "utils/error_checking.hpp"
+
 namespace parthenon {
 class Mesh;
 /// The DataCollection class is an abstract container that contains at least a
@@ -42,21 +44,36 @@ class DataCollection {
 
   void SetMeshPointer(Mesh *pmesh) { pmy_mesh_ = pmesh; }
 
-  std::shared_ptr<T> &Add(const std::string &label, const std::shared_ptr<T> &src,
-                          const std::vector<std::string> &flags, const bool shallow);
-  std::shared_ptr<T> &Add(const std::string &label, const std::shared_ptr<T> &src,
-                          const std::vector<std::string> &flags = {});
-  std::shared_ptr<T> &AddShallow(const std::string &label, const std::shared_ptr<T> &src,
-                                 const std::vector<std::string> &flags = {});
-  std::shared_ptr<T> &Add(const std::string &label) {
-    // error check for duplicate names
-    auto it = containers_.find(label);
+  template <typename ID_t>
+  std::shared_ptr<T> &Add(const std::string &name, const std::shared_ptr<T> &src,
+                          const std::vector<ID_t> &fields, const bool shallow) {
+    auto it = containers_.find(name);
     if (it != containers_.end()) {
+      if (!(it->second)->Contains(fields)) {
+        PARTHENON_THROW(name +
+                        " already exists in collection but does not contain fields");
+      }
       return it->second;
     }
-    containers_[label] = std::make_shared<T>();
-    return containers_[label];
+
+    auto c = std::make_shared<T>(name);
+    c->Initialize(src.get(), fields, shallow);
+
+    Set(name, c);
+
+    return containers_[name];
   }
+  template <typename ID_t = std::string>
+  std::shared_ptr<T> &Add(const std::string &label, const std::shared_ptr<T> &src,
+                          const std::vector<ID_t> &fields = {}) {
+    return Add(label, src, fields, false);
+  }
+  template <typename ID_t = std::string>
+  std::shared_ptr<T> &AddShallow(const std::string &label, const std::shared_ptr<T> &src,
+                                 const std::vector<ID_t> &fields = {}) {
+    return Add(label, src, fields, true);
+  }
+  std::shared_ptr<T> &Add(const std::string &label);
 
   auto &Stages() { return containers_; }
   const auto &Stages() const { return containers_; }
--- a/src/interface/data_collection.hpp
+++ b/src/interface/data_collection.hpp
@@ -49,7 +49,7 @@ class DataCollection {
                           const std::vector<ID_t> &fields, const bool shallow) {
     auto it = containers_.find(name);
     if (it != containers_.end()) {
-      if (!(it->second)->ContainsExactly(fields)) {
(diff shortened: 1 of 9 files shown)


pdmullen (approved):

src/interface/mesh_data.cpp
```diff
@@ -61,6 +36,17 @@ void MeshData<T>::Set(BlockList_t blocks, Mesh *pmesh) {
   Set(blocks, pmesh, ndim);
 }
 
+template <typename T>
+std::shared_ptr<MeshBlock> MeshData<T>::GetBlock(const int i, const MeshData<T> *src,
+                                                 const Mesh *pmesh) const {
```
pgrete: Is the `GetBlock` functionality potentially universal, i.e., applicable beyond the `Inititalize` call?
I'm asking as `GetBlock` is a pretty universal name, so if it's special to the `Initialize` call, I suggest to name it more specific (so that the general `GetBlock` call remains available for future use).

tst/unit/test_data_collection.cpp
```diff
@@ -116,5 +116,34 @@ TEST_CASE("Adding MeshBlockData objects to a DataCollection", "[DataCollection]"
         REQUIRE(hxv2(0) == hv2(0));
       }
     }
+    AND_WHEN("We want only a subset of variables in a new MeshBlockData by UID") {
+      // reset vars
+      par_for(
+          loop_pattern_flatrange_tag, "init vars", DevExecSpace(), 0, 0,
+          KOKKOS_LAMBDA(const int i) { v2(0) = 222; });
```
pgrete: What is this doing?
Shouldn't this match the `xv2(0)=22` call below?

pgrete (commented):
In general, looks good to me and nice refactor along the way.
I just don't understand the unit test. Shouldn't `222 != 22` as in `hv2(0) != hxv2(0)`?
Maybe I'm misreading sth.

src/interface/mesh_data.cpp
Yurlungur: @lroberts36 should jump in here---it's not quite universal, I think... as the `MeshData<T> *src` piece specifies this as being for `Initialize`. I will change it to `GetSourceBlock_`.

tst/unit/test_data_collection.cpp
Yurlungur: No it shouldn't---this inits v2 to the *wrong* value as a check that it is overwritten when it is called again on line 108 in the new mesh data object, which is not shallow.

Yurlungur: I didn't actually write this logic, I copied it from the lines above for the add call by string. I'm fine to remove it if you like.

Yurlungur: I added a comment to this effect.

Yurlungur:
Thanks for the review @pgrete I think I've addressed your comments now.

src/interface/mesh_data.cpp
```diff
@@ -16,31 +16,6 @@
...
-              stage_name_, src->GetBlockData(i), names, shallow);
-    }
-  } else {
-    for (int i = 0; i < nblocks; i++) {
-      block_data_[i] = pmy_mesh_->block_list[i]->meshblock_data.Add(
-          stage_name_, src->GetBlockData(i), names, shallow);
-    }
-  }
-}
-
```
lroberts36, line 43: This is going to conflict with changes in #1043

src/interface/data_collection.hpp
```diff
@@ -42,21 +44,36 @@ class DataCollection {
...
-    auto it = containers_.find(label);
+  template <typename ID_t>
+  std::shared_ptr<T> &Add(const std::string &name, const std::shared_ptr<T> &src,
+                          const std::vector<ID_t> &fields, const bool shallow) {
+    auto it = containers_.find(name);
     if (it != containers_.end()) {
+      if (!(it->second)->Contains(fields)) {
+        PARTHENON_THROW(name +
+                        " already exists in collection but does not contain fields");
+      }
```
lroberts36: Does this check that `fields` is exactly the same set of fields as are contained in `it->second` or just that the requested fields are a subset of the fields in `it->second`? I am not sure which is the desired behavior, but I would lean towards the former.

src/interface/mesh_data.hpp
```diff
@@ -242,8 +242,24 @@ class MeshData {
...
+    const int nblocks = src->NumBlocks();
+    block_data_.resize(nblocks);
+
+    grid = src->grid;
+    for (int i = 0; i < nblocks; ++i) {
+      block_data_[i] =
+          GetSourceBlockPointer_(i, src, pmy_mesh_)
+              ->meshblock_data.Add(stage_name_, src->GetBlockData(i), vars, shallow);
+    }
+  }
```
lroberts36: Maybe I am missing something, but this looks like it will just grab the first `nblocks` blocks in the meshblock list for a `Mesh` on a given rank. Doesn't this mean the wrong blocks will be grabbed if there are multiple partitions?

src/interface/meshblock_data.hpp
```diff
@@ -109,8 +110,33 @@ class MeshBlockData {
...
+    if (vars.empty()) {
+      for (auto v : src->GetVariableVector()) {
+        add_var(v);
+      }
+    } else {
+      for (const auto &v : vars) {
+        add_var(src->GetVarPtr(v));
+      }
+    }
+  }
```
lroberts36, line 140: These are also likely to conflict with #1043

lroberts36 (commented):
I am not sure this works with general partitioning, but maybe I am misunderstanding things. See my comments below.

src/interface/data_collection.hpp
Yurlungur: It's the latter. This checks that the requested fields are all available in `it->second`. And I would suggest that's desired behavior.

lroberts36: I guess the concern I have is for communication. If you mistakenly use the same name for two mesh_data, one of which contains a small subset of the fields of the other you will possibly end up communicating extra fields you didn't intend.

Yurlungur: Hmm... That's true. I can change this, but that would be a change in functionality, as I just preserved the existing functionality (probably created by @jdolence ?)

lroberts36: I would just leave it as is.

Yurlungur: Too late fixed it. ;)

src/interface/mesh_data.hpp
Yurlungur: Discussed offline. You're right---this is a bug. Fixed, and added a comment discussing an edge case we need to be concerned about. But not fixing it right now.

Yurlungur:
@lroberts36 thanks for the review. I think I've addressed all your comments.

lroberts36 (approved):
LGTM

pgrete:
With the latest changes the distinction between `if (grid.type == GridType::two_level_composite) {` and using `pmy_mesh_->gmg_block_lists[src->grid.logical_level][i]` is gone.
Is that not required any more?

lroberts36:
@pgrete: Yeah, that is no longer required since the `src` `MeshData` already should point to internal/multigrid blocks.

PR #1062, opened 2024-04-22
UserWorkBeforeRestartOutput

brryan:
## PR Summary

We have a downstream need to call a third party library's restart capability when writing our own restarts. This PR adds hooks for `Mesh` and `MeshBlock` analogs of `UserWorkBeforeOutput` that are only triggered during restart outputs. 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


src/application_input.hpp
```diff
@@ -46,6 +46,8 @@ struct ApplicationInput {
 
   std::function<void(Mesh *, ParameterInput *, SimTime const &)>
       UserMeshWorkBeforeOutput = nullptr;
+  std::function<void(Mesh *, ParameterInput *, SimTime const &)>
```
brryan: It might be nice to make `OutputParameters*` an optional input argument here, so we could get things like the output number, filename, etc. for synchronizing additional write operations. But I don't think, with our function pointers, we could also add that optional argument to the existing functions like `UserMeshWorkBeforeOutput` without breaking downstream.

src/interface/params.cpp
```diff
@@ -109,6 +110,7 @@ void Params::ReadFromRestart(const std::string &prefix, const HDF5::H5G &group)
   // strings
   ReadFromHDF5AllParamsOfType<std::string>(prefix, group);
   ReadFromHDF5AllParamsOfType<std::vector<std::string>>(prefix, group);
+  ReadFromHDF5AllParamsOfType<std::vector<char>>(prefix, group);
```
brryan, line 113: This lets us read binary files into `Params` and keep them in `.rhdf` output files.

Yurlungur:
@brryan is this ready for review? It's still marked WIP

brryan:
> @brryan is this ready for review? It's still marked WIP

@Yurlungur Yes it works but I just wanted to see what people think about including `OutputParameters*` in the function signature for `UserWorkBeforeRestartOutput` (and somehow extending `UserWorkBeforeOutput` in the same way?) so we could get things like output filename downstream.

src/application_input.hpp
Yurlungur: This would be a breaking change, but I don't think it would be a big deal to make. I don't feel strongly.

src/mesh/mesh.cpp
```diff
@@ -911,6 +914,22 @@ void Mesh::ApplyUserWorkBeforeOutput(Mesh *mesh, ParameterInput *pin,
...
+
+  // call Mesh version
+  mesh->UserMeshWorkBeforeRestartOutput(mesh, pin, time);
+
+  // call MeshBlock version
+  for (auto &pmb : block_list) {
+    pmb->UserWorkBeforeRestartOutput(pmb.get(), pin);
+  }
+}
+
```
Yurlungur: Should this be in `Mesh`? Or would it be better to put it in `Driver`? I lean towards `Driver`.

src/pgen/default_pgen.cpp
```diff
@@ -131,6 +131,18 @@ void Mesh::UserMeshWorkBeforeOutputDefault(Mesh *, ParameterInput *, SimTime con
...
+//! \fn void Mesh::UserMeshWorkBeforeRestartOutputDefault(Mesh *pmb, ParameterInput *pin,
+//! SimTime &t)
+//  \brief Function called before generating output files
+//========================================================================================
+
+void Mesh::UserMeshWorkBeforeRestartOutputDefault(Mesh *, ParameterInput *,
+                                                  SimTime const &) {
+  // do nothing
+  return;
+}
```
Yurlungur: I don't think these weakly linked defaults are needed or used anymore are they? We can just put the function pointer in `application_input`?

Would it be desirable also to have a `StateDescriptor` hook for per-package work before restart output?

Yurlungur (approved):
I am fine with outputting `OutputParameters`. Also I wonder if this should be in `StateDescriptor` in addition to or instead of `ApplicationInput`?

Also should the flag be documented in the docs?

A few comments below. However, consider this non-blocking. I'll approve now.

src/mesh/mesh.cpp
brryan: I put this right below `Mesh::ApplyUserWorkBeforeOutput`, if we want to move this we should probably move both.

src/mesh/mesh.cpp
```diff
@@ -903,11 +906,35 @@ void Mesh::EnrollBndryFncts_(ApplicationInput *app_in) {
...
+                                            SimTime const &time) {
+
+  // call Mesh version
+  if (mesh->UserMeshWorkBeforeRestartOutput) {
+    mesh->UserMeshWorkBeforeRestartOutput(mesh, pin, time);
+  }
+
+  // call MeshBlock version
+  for (auto &pmb : block_list) {
+    if (pmb->UserWorkBeforeRestartOutput) {
```
brryan: @Yurlungur I removed the default implementations for `User*WorkBefore*Output` and now just need to check that the function pointer is valid before calling it -- any objections to this change? If not I can propagate to all our user hooks with default implementations.

Yurlungur: I am in favor of this change.

src/mesh/mesh.cpp
Yurlungur: Ah ok. Leave it for now then... My preference is to move all of this stuff into Driver and `StateDescriptor` rather than looping it through Mesh. But that's a bigger refactor maybe.

src/outputs/output_parameters.hpp
```diff
@@ -0,0 +1,58 @@
+//========================================================================================
```
brryan, line 1: Including `outputs/outputs.hpp` everywhere I needed it to make `OutputType` an argument to the `UserWork` callbacks introduced a nasty circular dependency. So now only this file, which doesn't contain any mesh or pack types, is included elsewhere.

src/interface/state_descriptor.hpp
```diff
@@ -412,31 +413,41 @@ class StateDescriptor {
...
+  std::function<void(MeshBlockData<Real> *rc)> PreCommFillDerivedBlock;
+  std::function<void(MeshData<Real> *rc)> PreCommFillDerivedMesh;
+  std::function<void(MeshBlockData<Real> *rc)> PreFillDerivedBlock;
+  std::function<void(MeshData<Real> *rc)> PreFillDerivedMesh;
+  std::function<void(MeshBlockData<Real> *rc)> PostFillDerivedBlock;
+  std::function<void(MeshData<Real> *rc)> PostFillDerivedMesh;
+  std::function<void(MeshBlockData<Real> *rc)> FillDerivedBlock;
+  std::function<void(MeshData<Real> *rc)> FillDerivedMesh;
+  std::function<void(Mesh *, ParameterInput *, SimTime &)> UserWorkBeforeLoopMesh;
+  std::function<void(Mesh *, ParameterInput *, SimTime &)> UserWorkBeforeOutputMesh;
```
brryan: The `= nullptr` and `= {nullptr}` are redundant so I removed them.

src/mesh/mesh.cpp
brryan: Yeah I'd rather defer that and build consensus first

src/pgen/default_pgen.cpp
brryan: You're right, these aren't needed, and are removed.

I agree that a `StateDescriptor` hook in addition is desirable. Just added.

src/interface/state_descriptor.hpp
Yurlungur: C++ trivia for the day. Apparently `std::function` very explicitly has an overload of `operator==` for comparing against `nullptr_t`. The zero-initialized (and initialized to nullptr) `std::function` is actually an *empty* `std::function`.

Yurlungur: https://en.cppreference.com/w/cpp/utility/functional/function/operator_cmp

src/outputs/outputs.cpp
```diff
@@ -439,8 +439,17 @@ void Outputs::MakeOutputs(Mesh *pm, ParameterInput *pin, SimTime *tm,
            ptype->output_params.analysis_flag)))) {
       if (first && ptype->output_params.file_type != "hst") {
         pm->ApplyUserWorkBeforeOutput(pm, pin, *tm);
+        for (const auto &pkg : pm->packages.AllPackages()) {
+          pkg.second->UserWorkBeforeOutput(pm, pin, *tm);
```
pdmullen, line 443: It would be a breaking change, but should this function also take as argument `&(ptype->output_params)`?

pdmullen (approved):

src/interface/state_descriptor.hpp
pgrete: So we traded the `= nullptr` against `if (FunctionCall) {FunctionCall()}`?
Are there any C++ best-practices for one or the other?
To me the original approach looked cleaner (though just commenting -- not saying we should revert).

src/mesh/mesh.cpp
```diff
@@ -901,11 +908,34 @@ void Mesh::EnrollBndryFncts_(ApplicationInput *app_in) {
...
+                                            SimTime const &time,
+                                            OutputParameters *pparams) {
+  // call Mesh version
+  if (mesh->UserMeshWorkBeforeRestartOutput) {
+    mesh->UserMeshWorkBeforeRestartOutput(mesh, pin, time, pparams);
+  }
+
+  // call MeshBlock version
+  for (auto &pmb : block_list) {
+    if (pmb->UserWorkBeforeRestartOutput) {
```
pgrete: Do we want to allow/support both callbacks to be defined?
In many other places, we restricted the use to one version so that's clear what's going to happen.

src/outputs/outputs.cpp
```diff
@@ -439,8 +439,17 @@ void Outputs::MakeOutputs(Mesh *pm, ParameterInput *pin, SimTime *tm,
...
+        }
         first = false;
       }
+      if (ptype->output_params.file_type == "rst") {
+        pm->ApplyUserWorkBeforeRestartOutput(pm, pin, *tm, &(ptype->output_params));
+        for (const auto &pkg : pm->packages.AllPackages()) {
+          pkg.second->UserWorkBeforeRestartOutput(pm, pin, *tm, &(ptype->output_params));
+        }
```
pgrete, line 451: General question: Do we want to/can we make any guarantees on the order of execution here (on other words, should we expect competing work being done here in downstream codes)?
Anyway, might be worth documenting that what we can guarantee and what not in the dev docs above.

pgrete (commented):
As mentioned on Thursday, I'm happy with forwarding output params (even if its' a breaking change).
With that in mind, I could also imagine a solution to the original problem without introducing new callbacks by just checking on output type internally downstream, but having additional callbacks also probably doesn't hurt.

src/interface/state_descriptor.hpp
Yurlungur: No---I think we can still do `if (FunctionCall != nullptr)`. `std::function` always has an overload for comparison against `nullptr_t`. This is, I think, equivalent to the fact that poitners are null initialized. 

Personally, I find explicitly assigning 0 to be a useful "documentation line" of code. But I don't feel strongly either way.

brryan: Yes I don't think we traded any behavior for how we check whether these function pointers are initialized or not -- comparisons with `std::function` behave the same whether we explicitly initialize to `nullptr` or not. 

I went back and forth on this and decided (/agreed with you both) that the self-documenting quality of `= nullptr` and `!= nullptr` is valuable so I reverted to what we've done previously. 

I couldn't find any style guidelines about this in e.g. the google style guide.

src/mesh/mesh.cpp
brryan: Yeah let me remove the `MeshBlock` version of `UserWorkBeforeRestartOutput` I think that is purely redundant. But I'm hesitant to remove the existing `MeshBlock` version of `UserWorkBeforeOutput`-- if we are going to refactor existing hooks I think that should be a separate PR.

src/outputs/outputs.cpp
brryan: I state in the documentation now that the per-mesh callbacks precede the per-package callbacks, but both are always called if they are provided.

brryan:
OK I updated this to reintroduce explicitness around `nullptr` with `std::function` and I removed the per-MeshBlock version of `UserWorkBeforeRestartOutput`. What I did *not* do is:
 - Remove the per-MeshBlock version of `UserWorkBeforeOutput`
 - Change the signature of `UserWorkBeforeOutput` to include `OutputParameters`.
 
 That would pick up two breaking changes so maybe we want to get some consensus on a separate PR that refactors some of these hooks sometime in the future? If so I think this is ready to go, @pgrete if you are OK to approve (and others don't want to pause before merge) I will then go ahead and merge this in.

src/interface/state_descriptor.hpp
pgrete: I was actually referring to other changes (see following highlight)

src/driver/driver.cpp
```diff
@@ -96,17 +96,25 @@ DriverStatus EvolutionDriver::Execute() {
...
       if (Globals::my_rank == 0) OutputCycleDiagnostics();
 
-      pmesh->PreStepUserWorkInLoop(pmesh, pinput, tm);
-      pmesh->PreStepUserDiagnosticsInLoop(pmesh, pinput, tm);
+      if (pmesh->PreStepUserWorkInLoop != nullptr) {
+        pmesh->PreStepUserWorkInLoop(pmesh, pinput, tm);
+      }
+      if (pmesh->PreStepUserDiagnosticsInLoop != nullptr) {
+        pmesh->PreStepUserDiagnosticsInLoop(pmesh, pinput, tm);
+      }
```
pgrete, line 104: Followup up from above: this is the change I was referring to, i.e., now needing to check for the function rather than just calling it.

pgrete (approved):
LGTM

src/driver/driver.cpp
brryan: Oh right I see, yes I added checks that the function pointers are not `null` everywhere necessary in the code I think (if not I would expect the regression tests to break).

PR #1067, opened 2024-04-30
Enables compilation with nvhpc-sdk.

dmageeLANL:
Very minor change to avoid unrecognized flag when building with nvidia-hpc-sdk. This facilitates Venado Acceptance testing, but also should help vendors build parthenon for ats5 benchmarking.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -308,7 +308,8 @@ endif()
 # Therefore, we don't use the Kokkos_ENABLE_CUDA_CONSTEXPR option add the flag manually.
 # Also, not checking for NVIDIA as nvcc_wrapper is identified as GNU so we just make sure
 # the flag is not added when compiling with Clang for Cuda.
-if (Kokkos_ENABLE_CUDA AND NOT CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+if (Kokkos_ENABLE_CUDA AND NOT CMAKE_CXX_COMPILER_ID STREQUAL "Clang"
+   AND NOT CMAKE_CXX_COMPILER_ID STREQUAL "NVHPC" )
    target_compile_options(parthenon PUBLIC --expt-relaxed-constexpr)
 endif()
 


Yurlungur (approved):
Sufficiently minor I don't think we need to include in the changelog.

Does nvhpc automatically include the equivalent of `relaxed-constexpr`?

Yurlungur:
Also should this MR also be applied into parthenon develop?

dmageeLANL:
> Also should this MR also be applied into parthenon develop?

Sure, I mean I don't see why not. I'm planning on using this branch for venado testing because I'm familiar with it from the DAT we had on Crossroads April 29 and it seems to ameliorate the startup MPI issues. That's my main motivation for the MR, but you may find it useful for running on chicoma/venado in general if you're planning on using nvc++.

dmageeLANL:
> Sufficiently minor I don't think we need to include in the changelog.
> 
> Does nvhpc automatically include the equivalent of `relaxed-constexpr`?

I don't know what nvc++ is doing under the hood, but I know that adding that flag makes the build fail on Chicoma. I get these build errors:

``` bash
$~ CC --expt-relaxed-constexpr
nvc++-Error-Unknown switch: --expt-relaxed-constexpr
```

pgrete:
I think this is an `nvcc` flag, so the logic should probably be ~~`CUDA and (not clang or not NVHPC)`~~ [edit] nevermind, the original logic should be fine [/edit]
Alternatively, it might be worth to peek into what Kokkos does this with flag/feature and nvc++.

Yurlungur:
> > Sufficiently minor I don't think we need to include in the changelog.
> > Does nvhpc automatically include the equivalent of `relaxed-constexpr`?
> 
> I don't know what nvc++ is doing under the hood, but I know that adding that flag makes the build fail on Chicoma. I get these build errors:
> 
> ```shell
> $~ CC --expt-relaxed-constexpr
> nvc++-Error-Unknown switch: --expt-relaxed-constexpr
> ```

I have built succesfully on Chicoma with nvhpc, but it required a custom version of the kokkos wrapper. At the time, Kokkos did not support nvhpc as a compiler. Here's my build process on Chicoma:
https://github.com/lanl/phoebus/issues/70

Yurlungur:
Looks like cpplint is failing. Odd. I'm just going to click merge. @pgrete I suggest we keep this change in mind. But hold off on making it in develop until we have our hands on Venado (which should be soon for me).

PR #1068, opened 2024-05-02
Add host labels and an overload for operator<< in sparsepack

Yurlungur:
## PR Summary

I'm chasing down a bug in riot right now and it's been useful to be able to print the contents of a `SparsePack` object. This adds the capability to do so with a `LabelHost` function that can be called on host and a stream operator overload if you just want to dump the whole pack.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/src/interface/sparse_pack_base.cpp
+++ b/src/interface/sparse_pack_base.cpp
@@ -163,7 +163,7 @@ SparsePackBase SparsePackBase::Build(T *pmd, const PackDescriptor &desc,
     leading_dim += 2;
   }
   pack.pack_ = pack_t("data_ptr", leading_dim, pack.nblocks_, max_size);
-  auto pack_h = Kokkos::create_mirror_view(pack.pack_);
+  pack.pack_h_ = Kokkos::create_mirror_view(pack.pack_);
 
   // For non-flat packs, shape of pack is type x block x var x k x j x i
   // where type here might be a flux.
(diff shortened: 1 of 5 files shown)


lroberts36 (approved):
LGTM, seems very helpful

src/interface/sparse_pack_base.cpp
```diff
@@ -163,7 +163,7 @@ SparsePackBase SparsePackBase::Build(T *pmd, const PackDescriptor &desc,
     leading_dim += 2;
   }
   pack.pack_ = pack_t("data_ptr", leading_dim, pack.nblocks_, max_size);
-  auto pack_h = Kokkos::create_mirror_view(pack.pack_);
+  pack.pack_h_ = Kokkos::create_mirror_view(pack.pack_);
```
pgrete, line 166: The implication from this change is that the host mirror is now mirroring the content of the device pack after build, isn't it?
And is this intentional or fixing a bug?

pgrete (approved):
LGTM

src/interface/sparse_pack_base.cpp
Yurlungur: The implication is that now the host mirror contains device views. This is intentional, but also not a bug fix. Previously the host mirror was not persistent in the pack. It's now persistent, which enables us to access, e.g., the `label` method for the packed views, which isn't available otherwise as `pack_` points to device memory.

PR #1053, opened 2024-04-15
Set the correct root level on restart

lroberts36:
## PR Summary

Currently, on restart `Mesh::root_level` and `Mesh::current_level` get set to levels relative to the legacy tree, which is incorrect (see issue #1051). This PR fixes the issue.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Changes are summarized in CHANGELOG.md
- [x] CI has been triggered on [Darwin](https://re-git.lanl.gov/eap-oss/parthenon/-/pipelines) for performance regression tests.
- [x] (@lanl.gov employees) Update copyright on changed files

PR #1047, opened 2024-04-11
Implement general 2D forest meshes

lroberts36:
## PR Summary

This PR implements the minimum necessary technology for defining meshes based on generally connected forests of octrees. I have left details of coordinates to a future PR, so for now each tree must cover a hyper-rectangular domain defined by $\vec{x}_\text{min}$ and $\Delta \vec{x}$. Some rough notes about my implementation are in [ForestOfOctreesNotes.pdf](https://github.com/parthenon-hpc-lab/parthenon/files/15240509/ForestOfOctreesNotes.pdf), which is now also included in `doc/latex/main.pdf`.

The four major things this PR accomplishes:
- Add `forest::Node`, `forest::Edge`, and `forest::Face`  for defining the forest mesh topology. `forest::Face` contains routines for finding neighbors and then calculating the correct logical coordinate transformations to those neighbors. For now, I have stuck to defining 2D forests but the mechanisms used here should generalize to 3D relatively easily in a future PR. 
- Add  `forest::Forest::Make2D` that builds a `Forest` from a `forest::ForestDefinition`. `ForestDefinition` contains a set of `Faces`, boundary conditions defined on `forest::Edge`s, `RegionSize`s associated with each block, and a set of initial `LogicalLocation`s to refine.
- Add `Mesh` constructor as well as `ParthenonManager::ParthenonInitPackagesAndMesh` that takes a `forest::ForestDefinition` and constructs a mesh based on a general 2D forest.
- Thread logical coordinate transformations through boundary communication routines so that neighbor blocks that are non-simply oriented can communicate boundaries. 

Smaller things also included in the PR:
- Rename `RelativeOrientation` -> `LogicalLocationCoordinateTransformation` for clarity.
- Move `MeshBndryFnctn` to `Mesh::forest::trees` instead of `Mesh` and add appropriate registration routines. 
- Adds an example and associated regression test that communicates `gid`s and other information through boundaries to ensure that neighbor communication is occurring correctly. The gold file has just been verified by hand.
- Adds a unit test that performs refinement on non-simply connected forests. 

Things this PR does not do (but maybe a future PR should):
- Deal with physical coordinates on trees beyond the simplest choice. 
- Implement periodic boundary conditions for general forests.
- Allow for 3D meshes.
- Allow for 6+ valent points in the macro-mesh.
- Deal with corner communication for 5-valent corners or generally store information about valence of corners and edges in `MeshBlock`.
- Make a better interface for users to define general forests.
 
[...]

PR #1071, opened 2024-05-09
Fix bug in static refinement

lroberts36:
## PR Summary

The re-definition of `Mesh::root_level` introduced in #1053 did not get correctly propagated through the static refinement related code (which operates on logical locations relative to the legacy tree). This PR fixes that issue, which had cropped up in Artemis.

## PR Checklist

- [x] #1073 is merged
- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


Yurlungur (commented):
Thanks for the catch! To prevent this kind of thing from happening again... maybe we should have a regression test that uses SMR?

lroberts36:
Agreed, we need to add a couple tests that use multi-tree forests. We have at least one that uses SMR, but the bug does not manifest if the base grid only requires one tree.

tst/regression/test_suites/sparse_advection/sparse_advection.py
```diff
@@ -33,6 +33,27 @@ def Prepare(self, parameters, step):
...
+                "parthenon/mesh/nx2=32",
+                "parthenon/time/nlim=50",
+                "parthenon/job/problem_id=sparse_twotree_static",
+                "parthenon/mesh/refinement=static",
+                "parthenon/static_refinement0/x1min=-0.75",
+                "parthenon/static_refinement0/x1max=-0.5",
+                "parthenon/static_refinement0/x2min=-0.75",
+                "parthenon/static_refinement0/x2max=-0.5",
+                "parthenon/static_refinement0/level=3",
+            ]
```
Yurlungur, line 55: :+1:

Yurlungur (approved):
:+1: for fixes and :+1: for tests

pdmullen (approved):
LGTM

tst/regression/test_suites/sparse_advection/sparse_advection.py
```diff
@@ -77,5 +98,31 @@ def Analyse(self, parameters):
...
+                return False
+
+            delta = compare(
+                [
+                    "sparse_twotree.out0.final.phdf",
+                    parameters.parthenon_path
+                    + "/tst/regression/gold_standard/sparse_twotree.out0.final.phdf",
+                ],
+                one=True,
+                tol=1e-12,
```
pgrete, line 112: Any reason for the "loose" tolerance?

tst/regression/test_suites/sparse_advection/sparse_advection.py
```diff
@@ -77,5 +98,31 @@ def Analyse(self, parameters):
                 tol=1e-12,
                 check_metadata=False,
             )
+            if delta != 0:
+                return False
```
pgrete: In other tests with num_steps > 0, this approach sometimes resulted in some difficulties in figuring out where the test actually failed.
So I suggest to either add a `print("check X failed")` before each `return` or add some `passed = False` with a `print` and only a single `return passed` at the end of the `Analyse` function.

pgrete (approved):
LGTM (plus optional comments on the python stuff)

The updated gold files are in #1073, right?

tst/regression/test_suites/sparse_advection/sparse_advection.py
lroberts36: I think I just copied this tolerance from the other checks. Isn't 1.e-12 what we set for most tests? I don't think this test should be bitwise exact between machines.

tst/regression/test_suites/sparse_advection/sparse_advection.py
lroberts36: added.

lroberts36:
> The updated gold files are in #1073, right?

Yeah, I added the new gold files required here (and in #1047) to v23 so that I didn't have to upgold a bunch of times.

PR #1073, opened 2024-05-10
Make AMR and sparse restarts bitwise exact

### LINKED ISSUE #1072: Some results of restarts are not bitwise exact
How I ended up here is a longer story, but today I was looking at restarting the advection example. On current develop in a build directory, if I run 
```bash
rm *hdf*
rm parthinput*
cp ../tst/regression/test_suites/advection_outflow/parthinput.advection_outflow .
echo " " >> parthinput.advection_outflow
echo "<parthenon/output1>" >> parthinput.advection_outflow
echo "file_type = rst" >> parthinput.advection_outflow
echo "dt = 0.05" >> parthinput.advection_outflow
./example/advection/advection-example -i parthinput.advection_outflow parthenon/job/problem_id=gold
./example/advection/advection-example -r gold.out1.00009.rhdf parthenon/job/problem_id=silver
h5diff -c gold.out1.final.rhdf silver.out1.final.rhdf
```
the output I get from the h5diff is
```
attribute: <WallTime of </Info>> and <WallTime of </Info>>
1 differences found
attribute: <File of </Input>> and <File of </Input>>
4397 differences found
dataset: </advected> and </advected>
1550 differences found
dataset: </one_minus_sqrt_one_minus_advected_sq_12> and </one_minus_sqrt_one_minus_advected_sq_12>
1385 differences found
dataset: </one_minus_sqrt_one_minus_advected_sq_37> and </one_minus_sqrt_one_minus_advected_sq_37>
1385 differences found
```

If I switch to restarting from `gold.out1.00005.rhdf` I get
```
attribute: <WallTime of </Info>> and <WallTime of </Info>>
1 differences found
attribute: <File of </Input>> and <File of </Input>>
4396 differences found
dataset: </advected> and </advected>
3987 differences found
dataset: </one_minus_sqrt_one_minus_advected_sq_12> and </one_minus_sqrt_one_minus_advected_sq_12>
3663 differences found
dataset: </one_minus_sqrt_one_minus_advected_sq_37> and </one_minus_sqrt_one_minus_advected_sq_37>
3663 differences found
```
These diffs are much larger than machine precision (I have to take the tolerance of the h5diff to 1.e-5 before the dataset differences go away), but they are small compared to the magnitude of the solution. What is strange is that if I restart from `gold.out1.0000[0-4].rhdf` or `gold.out1.0000[6-8].rhdf` I get bitwise agreement between the gold and silver final outputs. This behavior seems to persist back to at least the most recent release. At least to me, this behavior seems to be a bug.

lroberts36:
## PR Summary
This PR saves the derefinement count for each meshblock and the deallocation count for each sparse variable in the restart output (as well as in regular HDF5 output). This should ensure that restarted simulations give bitwise exact results to simulations that were run from start to finish. 

Previously, AMR simulations that were restarted had the deallocation count of all meshblocks reset to zero. This meant that after restart, a block could be deallocated a few steps after it would have been deallocated if there was no restart. This resulted in differences in simulation results with and without restarts many orders of magnitude larger than machine precision, although still relatively small compared to characteristic scales of the problem. An example of this difference is in issue #1072 (the problem detailed there is fixed by this PR). 

Similarly, the deallocation count of sparse variables was not stored and was reset to zero on restart. This was a known issue, and the restart regression test set the global deallocation count large enough that variables were never deallocated. This PR also stores the deallocation count of variables in the restart file and sets them to the correct values on restart. The restart test now works with a deallocation count of 5.

Also, `MeshRefinement::DereferenceCount` is renamed to `MeshRefinement::DeallocationCount` since dereference count doesn't really make sense.

**WARNING: This PR does not solve the edge case described in issue #1077**

## PR Checklist

- [x] Code passes cpplint
- [x] Adds a test for any bugs fixed.
- [x] Code is formatted
- [x] Docs build


src/mesh/mesh.cpp
```diff
@@ -598,7 +598,7 @@ Mesh::Mesh(ParameterInput *pin, ApplicationInput *app_in, RestartReader &rr,
   auto locLevelGidLidCnghostGflag = mesh_info.level_gid_lid_cnghost_gflag;
   current_level = -1;
   for (int i = 0; i < nbtotal; i++) {
-    loclist[i] = LogicalLocation(locLevelGidLidCnghostGflag[5 * i], lx123[3 * i],
+    loclist[i] = LogicalLocation(locLevelGidLidCnghostGflag[6 * i], lx123[3 * i],
```
Yurlungur: This is exactly where I would have suggested sticking this :+1:

src/outputs/parthenon_hdf5.cpp
```diff
@@ -441,8 +443,8 @@ void PHDF5Output::WriteOutputFileImpl(Mesh *pm, ParameterInput *pin, SimTime *tm
   // write SparseInfo and SparseFields (we can't write a zero-size dataset, so only write
   // this if we have sparse fields)
   if (num_sparse > 0) {
-    WriteSparseInfo_(pm, sparse_allocated.get(), sparse_names, num_sparse, file, pl_xfer,
-                     my_offset, max_blocks_global);
+    WriteSparseInfo_(pm, sparse_allocated.get(), sparse_dealloc_count, sparse_names,
```
Yurlungur, line 447: Nice :+1:

src/outputs/parthenon_hdf5.cpp
```diff
@@ -594,8 +596,8 @@ void PHDF5Output::WriteBlocksMetadata_(Mesh *pm, hid_t file, const HDF5::H5P &pl
 
   {
     // (LOC.)level, GID, LID, cnghost, gflag
-    hsize_t loc_cnt[2] = {num_blocks_local, 5};
-    hsize_t glob_cnt[2] = {max_blocks_global, 5};
+    hsize_t loc_cnt[2] = {num_blocks_local, 6};
```
Yurlungur: Maybe this should be a global const now since we had to change it.

Yurlungur (approved):
LGTM. Thanks for the catch! Some nitpicks below

src/outputs/parthenon_hdf5.cpp
lroberts36: it is not obvious to me where the global should live though

src/outputs/restart.hpp
```diff
@@ -63,6 +65,17 @@ class RestartReader {
 
       return allocated[block * num_sparse + sparse_field_idx];
     }
+
+    int DeallocCount(int block, int sparse_field_idx) const {
+      PARTHENON_REQUIRE_THROWS(allocated != nullptr,
+                               "Tried to get allocation status but no data present");
+      PARTHENON_REQUIRE_THROWS((block >= 0) && (block < num_blocks),
+                               "Invalid block index in SparseInfo::IsAllocated");
```
pdmullen:
```suggestion
                               "Invalid block index in SparseInfo:: DeallocCount");
```

src/outputs/restart.hpp
```diff
@@ -63,6 +65,17 @@ class RestartReader {
...
+      PARTHENON_REQUIRE_THROWS((sparse_field_idx >= 0) && (sparse_field_idx < num_sparse),
+                               "Invalid sparse field index in SparseInfo::IsAllocated");
```
pdmullen:
```suggestion
                               "Invalid sparse field index in SparseInfo:: DeallocCount");
```

pdmullen (approved):
LGTM

src/outputs/parthenon_hdf5.cpp
```diff
@@ -660,6 +662,7 @@ void PHDF5Output::WriteLevelsAndLocs_(Mesh *pm, hid_t file, const HDF5::H5P &pl,
 }
 
 void PHDF5Output::WriteSparseInfo_(Mesh *pm, hbool_t *sparse_allocated,
+                                   const std::vector<int> dealloc_count,
```
pgrete: const ref rather than copy? (not that it'd matter performance wise given the use of this function and I can fix that in the OpenPMD PR which will also touch this function)

src/outputs/parthenon_hdf5.cpp
pgrete: As far as I can tell this "structure" is intrinsically tied to restarts, so one "global" place I could imagine is the restart reader header.

src/outputs/restart_hdf5.cpp
```diff
@@ -102,6 +103,12 @@ RestartReaderHDF5::SparseInfo RestartReaderHDF5::GetSparseInfo() const {
     // Read data from file
     PARTHENON_HDF5_CHECK(H5Dread(hdl.dataset, hdl.type, memspace, hdl.dataspace,
                                  H5P_DEFAULT, static_cast<void *>(info.allocated.get())));
+    info.dealloc_count.resize(hdl_dealloc.count);
+    PARTHENON_HDF5_CHECK(H5Dread(hdl_dealloc.dataset, hdl_dealloc.type,
+                                 H5S::FromHIDCheck(H5Screate_simple(
+                                     hdl_dealloc.rank, hdl_dealloc.dims.data(), NULL)),
+                                 hdl_dealloc.dataspace, H5P_DEFAULT,
+                                 static_cast<void *>(info.dealloc_count.data())));
```
pgrete, line 111: Do you see a case where this might cause issues for your downstream codes (e.g., loading older restart files)?
If so, does it make sense/is it worth to add a default override to support older files?

We don't have sparse variables yet, so it's up to you.

src/parthenon_manager.cpp
```diff
@@ -347,6 +347,10 @@ void ParthenonManager::RestartPackages(Mesh &rm, RestartReader &resfile) {
         // check if the sparse variable is allocated on this block
         if (sparse_info.IsAllocated(pmb->gid, sparse_idxs.at(label))) {
           pmb->AllocateSparse(label);
+          auto dealloc_count = sparse_info.DeallocCount(pmb->gid, sparse_idxs.at(label));
+          // Warning: For this to work, it is required that the controlling variable is
+          // stored in the restart files.
+          pmb->meshblock_data.Get()->GetVarPtr(label)->dealloc_count = dealloc_count;
```
pgrete, line 353: Can this reasonable be triggered in practice? And if so, does it make sense to throw here?

tst/regression/test_suites/restart/restart.py
```diff
@@ -67,11 +75,11 @@ def Analyse(self, parameters):
 
         success = True
 
-        def compare_files(name):
+        def compare_files(name, base="silver"):
             delta = compare(
                 [
                     "gold.out0.%s.rhdf" % name,
-                    "silver.out0.%s.rhdf" % name,
+                    "{}.out0.{}.rhdf".format(base, name),
```
pgrete, line 82: Why not go fully modern? ;)
`f"{base}.out0.{name}.rhdf",`

pgrete (approved):
LGTM thanks for the fix!
Feel free to merge as is if the "restart from old files" is not an issue.

src/parthenon_manager.cpp
lroberts36: ~~Actually, I think that I added this warning is erroneously. I think the sparse information for all fields (not just those that are written to file) is output here. I will remove it.~~

Edit: This can be triggered in practice, but it is non-trivial to check that the controlling variable was stored. I think I will put an issue up rather than try to fix this here.

Edit: Issue #1077

tst/regression/test_suites/restart/restart.py
lroberts36: oh, man, I must be getting old. I thought `"".format()` was the modern way of doing it...

src/outputs/restart_hdf5.cpp
lroberts36: I think Riot is the only code using sparse variables and I don't think we have any legacy data that we need to restart from currently, so I will just leave it as is. This way people will know if they are trying to restart from data that will not give exact results, and it is easy enough to add a workaround in the future if we really need to.

src/outputs/parthenon_hdf5.cpp
lroberts36: fixed to be const ref.

src/outputs/parthenon_hdf5.cpp
lroberts36: Added it there.

tst/regression/test_suites/restart/restart.py
Yurlungur: I'm also still not in the habit of doing fstrings....

PR #1079, opened 2024-05-16
Address XDMF/Visit Issues

pdmullen:
Visualizing HDF5 + XDMF files generated through `develop` with VisIt produces strange artifacts wherein 2D datasets often show a protruding X3 feature (which looks like a skyscraper).  This MR attempts to fix such artifacting by adjusting the data dumped in `HyperSlab` `DataItem`s.  

Parthenon particles cannot be visualized in VisIt currently, and their corresponding mesh is not even dumped in our current XDMF files.  For the life of me, I cannot get VisIt to interpret `VXVYVZ` type geometries for `Polyvertex` meshes.  I find that in order to get Visit to read our `Polyvertex` mesh, I must recast the geometry as an array of type `{{x1, y1, z2}, {x2, y2, z2}...}`.  Even after that change, I cannot get Visit to see the `Polyvertex` mesh unless it is its own XDMF file.  Visit can still plot particles on top of field variables if both XDMF files are read and a correlation is instantiated (VisIt prompts users automatically for this).  This MR enables users to optionally dump out swarm XDMF files, which, if enabled, also dumps restructured particle geometry in the corresponding HDF5 file.  Restart files should remain unchanged.  

This is gross.  Feel free to shoot down this MR. 

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build

### DIFF
--- a/src/outputs/parthenon_hdf5.cpp
+++ b/src/outputs/parthenon_hdf5.cpp
@@ -1,6 +1,6 @@
 //========================================================================================
 // Parthenon performance portable AMR framework
-// Copyright(C) 2020-2023 The Parthenon collaboration
+// Copyright(C) 2020-2024 The Parthenon collaboration
 // Licensed under the 3-clause BSD License, see LICENSE file for details
 //========================================================================================
 // (C) (or copyright) 2020-2024. Triad National Security, LLC. All rights reserved.
--- a/src/outputs/parthenon_hdf5.cpp
+++ b/src/outputs/parthenon_hdf5.cpp
@@ -520,6 +520,12 @@ void PHDF5Output::WriteOutputFileImpl(Mesh *pm, ParameterInput *pin, SimTime *tm
       }
     }
     if (output_params.write_swarm_xdmf) {
+      // TODO(@pdmullen): Here and above, we have worked with temp vectors pos_tmp and
+      // swarm_positions so that we can take the existing swarm position data structures
+      // and recast them into a format that XDMF/VisIt prefers.  Future efforts may (1)
+      // eliminate the extra geometry dump and/or (2) directly write the auxillary
+      // positions via low-level HDF writes, such that the vector manipulation below is
+      // unnecessary.
       const int npart = pos_tmp.size() / 3;
       std::vector<Real> swarm_positions(pos_tmp.size());
       int spcnt = 0;
--- a/src/outputs/parthenon_hdf5.cpp
+++ b/src/outputs/parthenon_hdf5.cpp
@@ -34,6 +34,7 @@
 
 #include "driver/driver.hpp"
 #include "interface/metadata.hpp"
+#include "interface/swarm_default_names.hpp"
 #include "mesh/mesh.hpp"
 #include "mesh/meshblock.hpp"
 #include "outputs/output_utils.hpp"
@@ -485,6 +486,17 @@ void PHDF5Output::WriteOutputFileImpl(Mesh *pm, ParameterInput *pin, SimTime *tm
       local_count[rank] = swinfo.count_on_rank;
       global_count[rank] = swinfo.global_count;
     };
+    auto SetCountsParticlePositions = [&](const SwarmInfo &swinfo) {
+      for (int i = 0; i < 6; ++i) {
+        local_offset[i] = 0; // reset everything
+        local_count[i] = 0;
+        global_count[i] = 0;
+      }
+      local_offset[0] = swinfo.global_offset;
+      local_count[0] = swinfo.count_on_rank;
+      global_count[0] = swinfo.global_count;
+      local_count[1] = global_count[1] = 3;
+    };
     auto &int_vars = std::get<SwarmInfo::MapToVarVec<int>>(swinfo.vars);
     for (auto &[vname, swmvarvec] : int_vars) {
       const auto &vinfo = swinfo.var_info.at(vname);
@@ -493,6 +505,7 @@ void PHDF5Output::WriteOutputFileImpl(Mesh *pm, ParameterInput *pin, SimTime *tm
       HDF5WriteND(g_var, vname, host_data.data(), vinfo.tensor_rank + 1, local_offset,
                   local_count, global_count, pl_xfer, H5P_DEFAULT);
     }
+    std::vector<Real> pos_tmp; // tmp vector to (potentially) hold particle positions
     auto &rvars = std::get<SwarmInfo::MapToVarVec<Real>>(swinfo.vars);
     for (auto &[vname, swmvarvec] : rvars) {
       const auto &vinfo = swinfo.var_info.at(vname);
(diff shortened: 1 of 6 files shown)


src/outputs/parthenon_hdf5.cpp
```diff
@@ -493,14 +505,34 @@ void PHDF5Output::WriteOutputFileImpl(Mesh *pm, ParameterInput *pin, SimTime *tm
...
+      std::vector<Real> swarm_positions(pos_tmp.size());
+      int spcnt = 0;
+      for (int i = 0; i < npart; ++i) {
+        swarm_positions[spcnt++] = pos_tmp[i];
+        swarm_positions[spcnt++] = pos_tmp[npart + i];
+        swarm_positions[spcnt++] = pos_tmp[2 * npart + i];
+      }
+      SetCountsParticlePositions(swinfo);
+      HDF5WriteND(g_var, "swarm_positions", swarm_positions.data(), 2, local_offset,
+                  local_count, global_count, pl_xfer, H5P_DEFAULT);
```
Yurlungur, line 539: This implies (I think) that we're writing swarm positions twice, right? Since you're outputting the fused version for xdmf but still outputting the original in whatever requested shape.

Is that desirable? Or should we change swarms to always be read and written in this way? As a conceit for visualization I think I'm fine doing it this way as the original shape is better IMO and more faithful to parthenon. Could go either way.

src/outputs/parthenon_hdf5.cpp
```diff
@@ -493,14 +505,34 @@ void PHDF5Output::WriteOutputFileImpl(Mesh *pm, ParameterInput *pin, SimTime *tm
...
     for (auto &[vname, swmvarvec] : rvars) {
       const auto &vinfo = swinfo.var_info.at(vname);
       auto host_data = swinfo.FillHostBuffer(vname, swmvarvec);
       SetCounts(swinfo, vinfo);
       HDF5WriteND(g_var, vname, host_data.data(), vinfo.tensor_rank + 1, local_offset,
                   local_count, global_count, pl_xfer, H5P_DEFAULT);
+      if (output_params.write_swarm_xdmf &&
+          (vname == swarm_position::x::name() || vname == swarm_position::y::name() ||
+           vname == swarm_position::z::name())) {
+        pos_tmp.insert(pos_tmp.end(), host_data.begin(), host_data.end());
```
Yurlungur, line 519: Is it necessary to fill this contiguous block like this? I feel like we could probably get around this by using HDF5's lower level memory space machinery to tell it to write swarm::x specifically into the first "column," swarm::y directly into the second column, etc. Maybe not though. Haven't thought about it carefully.

Yurlungur (commented):
Hey thanks for figuring this out! And I apologize for your suffering. I don't think this MR is the optimal implementation of this machinery. But I'm not sure I care about an optimal implementation... XDMF is forcing a bizarre structure on us and I think I'm perfectly happy to just hack in a solution rather than overengineer jumping through XDMF's hoops.

> This MR attempts to fix such artifacting by adjusting the data dumped in HyperSlab DataItems.

Am I missing something? I can't find this change in the diff.

Yurlungur (approved):
I'm approving now to indicate my willingness to just hack this in.

src/outputs/parthenon_hdf5.cpp
pdmullen: That is correct.  This was an intentional choice; but I equally hate it.  Here was my thought process:

(1) As you mention, this is totally wasteful; I am dumping the exact same info twice; just one in a particular format to make XDMF happy.  
(2) If I just switch to the format that XDMF requires, then (a) this would be a breaking change; (b) I would subsequently have to go in and make edits to fix swarm restarts; (c) I would have to make changes to `phdf.py`.
(3) In the spirit of trying be non-breaking, I keep restarts and dumps as is, just adding this new geometry piece iff swarm XDMF files are requested (which weren't available before).

pdmullen:
> Am I missing something? I can't find this change in the diff.

@Yurlungur  https://github.com/parthenon-hpc-lab/parthenon/pull/1079/files#diff-31b92df33d462322a957c4750de1874912d6947009746063e9684dae98314675R515

src/outputs/parthenon_hdf5.cpp
pdmullen: I was already shooting in the dark with the existing Parthenon HDF functions.  I think you are among the few that actually understands the nuts and bolts of those.  If you see how to implement what you propose straightforwardly, I would welcome that strategy with open arms.  This hack I put in is quite ugly.

src/outputs/parthenon_xdmf.cpp
```diff
@@ -462,7 +512,7 @@ static void BlockCoordRegularRef(std::ofstream &xdmf, int nbtot, int ib, int nx,
       "            %d 0 1\n"
       "            1 1 %d\n"
       "          </DataItem>\n",
-      nx + (nx > 1), ib, nx + (nx > 1));
+      nx + 1, ib, nx + 1);
```
Yurlungur, line 515: Oh shoot. I had to change this to support the moving coordinates machinery in 2d I think... Maybe we add a flag for that so this is `nx + (custom_coords && (nx>1))` or something?

Yurlungur (changes requested):
Thanks for pointing the change out. We should test this in Assail before merging...

src/outputs/parthenon_hdf5.cpp
Yurlungur: I don't think the parthenon HDF5 functions can do it. We'd have to go to the lower-level HDF5. Let's talk either tomorrow or next week (depending on when you're around). I can try to suggest a hack.

src/outputs/parthenon_xdmf.cpp
pdmullen: I didn't think that the moving coordinates machinery even used this function.

src/outputs/parthenon_hdf5.cpp
Yurlungur: We should probably get other people's opinion on this... but I think I'm fine with your approach. Doing it the other way would require changes to both `phdf.py` and the restart reader. And I prefer the original layout (obviously... as I designed it.)

Perhaps a question is how does OpenPMD want it?

src/outputs/parthenon_xdmf.cpp
Yurlungur: Oops you're right. I must have changed this when floundering... Sorry about this and good catch.

Yurlungur (approved):

src/outputs/parthenon_hdf5.cpp
pdmullen: In my most recent commit, I have opted to make `write_swarm_xdmf` `false` by default.  I feel this better reflects the discussion here; i.e., if you _really_ want to read particles into VisIt, then you incur the penalty of dumping geometry twice and creating an additional XDMF file.  Furthermore, the default behavior just recovers what is the current behavior in `develop` prior to this MR (but with the field variable XDMF fix).

lroberts36 (approved):
I am happy to approve this if it is working for you guys, but I don't think I can provide much useful feedback since I rarely use Visit and don't know what to expect for the XDMF.

src/outputs/parthenon_hdf5.cpp
Yurlungur: I'm fine with this.

src/outputs/parthenon_hdf5.cpp
Yurlungur: We discussed offline---I think it's possible to do this with low level HDF5 but unclear if this is needed.

PR #1081, opened 2024-05-20
[Trivial] add GetSize and GetIndex to sparse pack

Yurlungur:
## PR Summary

Upstreaming a few features we use in riot. Mainly a cheap way of getting the number of sparse IDs of a variable on a block and the "index" in the pack on a given block.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/src/interface/sparse_pack.hpp
+++ b/src/interface/sparse_pack.hpp
@@ -160,30 +160,30 @@ class SparsePack : public SparsePackBase {
   }
 
   // Host Bound overloads
-  inline int GetLowerBoundHost(const int b) const {
+  int GetLowerBoundHost(const int b) const {
     return (flat_ && (b > 0)) ? (bounds_h_(1, b - 1, nvar_) + 1) : 0;
   }
 
-  inline int GetUpperBoundHost(const int b) const { return bounds_h_(1, b, nvar_); }
+  int GetUpperBoundHost(const int b) const { return bounds_h_(1, b, nvar_); }
 
-  inline int GetLowerBoundHost(const int b, PackIdx idx) const {
+  int GetLowerBoundHost(const int b, PackIdx idx) const {
     static_assert(sizeof...(Ts) == 0);
     return bounds_h_(0, b, idx.VariableIdx());
   }
 
-  inline int GetUpperBoundHost(const int b, PackIdx idx) const {
+  int GetUpperBoundHost(const int b, PackIdx idx) const {
     static_assert(sizeof...(Ts) == 0);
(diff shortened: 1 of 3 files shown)


src/interface/sparse_pack.hpp
```diff
@@ -160,36 +160,54 @@ class SparsePack : public SparsePackBase {
   }
 
   // Host Bound overloads
-  KOKKOS_INLINE_FUNCTION int GetLowerBoundHost(const int b) const {
+  inline int GetLowerBoundHost(const int b) const {
```
lroberts36: I don't remember why these were `KOKKOS_INLINE_FUNCTION`s in the first place (maybe just copy and paste?). That being said, I don't think the `inline` is necessary since class methods defined in the class definition are implicitly `inline` already.

pdmullen (commented):
LGTM! if you are feeling brave, these can also be added to `SwarmPack`s

lroberts36 (approved):
LGTM

src/interface/sparse_pack.hpp
Yurlungur: yeah they should be automatically inline I think. I figured I'd just swap out the markings though... in case there was some documentation value to them being marked inline (probably there isn't).

Yurlungur:
> LGTM! if you are feeling brave, these can also be added to `SwarmPack`s

Can you click approve? Also are these relevant to `SwarmPack`s? I guess `GetIndex` is but since swarms don't have sparse IDs I don't think `GetSize` provides much value.

pdmullen (approved):

pdmullen:
> > LGTM! if you are feeling brave, these can also be added to `SwarmPack`s
> 
> Can you click approve? Also are these relevant to `SwarmPack`s? I guess `GetIndex` is but since swarms don't have sparse IDs I don't think `GetSize` provides much value.

The member functions are more or less consistent between `SparsePack`s and `SwarmPack`s, so it might seem natural to add them, however, as you add, I also don't see an immediate need for them...

src/interface/sparse_pack.hpp
lroberts36: Having the `inline` is a little confusing to me (since it makes me think they must be doing something (i.e. there is something different about these functions relative to other class definition defined methods) when they are really not).

Yurlungur: Ok I'll remove it

Yurlungur: done

PR #1089, opened 2024-05-23
Fix restaring from files not containing the derefinement counter (from #1073)

pgrete:
## PR Summary

In https://github.com/parthenon-hpc-lab/parthenon/pull/1073 the per-block derefinement counter was stored in restart files (and restored on restarts).
This changes the number of entries in the flattened `loc.level-gid-lid-cnghost-gflag` vector (from 5 per block to 6 per block).
If one currently restarts with from an older file this is not taken into account and logical locations (and potentially more are off).
I came across this because of a 
```
### PARTHENON ERROR
  Message:     Somehow didn't find a tree.
  File:        /project/project_465000552/env/src/athenapk/external/parthenon/src/mesh/forest/forest.hpp
  Line number: 128
Somehow didn't find a tree.
```
on restart.

Effectively, this renders restart any older file impossible/erroneous right now.
The proposed fix is not necessarily super-clean but it circumvents bumping the output format version (and introducing similar logic anyway).
But if there's a different/cleaner solution, I'm also open to suggestions.

Note, that this still doesn't fully fix a restart problem I'm tracking down (also in context of reviewing #1055) but it fixes the error above.
It might also be that the restart I'm trying to fix are broken because I used some Parthenon version in between that didn't contain all the mesh/tree related bugfixes that went in over the past weeks.
I'm still investigating, but I'm fairly sure that this PR and the proposed fix are necessary no matter what.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/src/outputs/parthenon_hdf5.cpp
+++ b/src/outputs/parthenon_hdf5.cpp
@@ -603,6 +603,16 @@ void PHDF5Output::WriteBlocksMetadata_(Mesh *pm, hid_t file, const HDF5::H5P &pl
     HDF5Write2D(gBlocks, "loc.level-gid-lid-cnghost-gflag", tmpID.data(), &loc_offset[0],
                 &loc_cnt[0], &glob_cnt[0], pl);
   }
+
+  {
+    // derefinement count
+    hsize_t loc_cnt[2] = {num_blocks_local, 1};
+    hsize_t glob_cnt[2] = {max_blocks_global, 1};
+    std::vector<int> tmpID = OutputUtils::ComputeDerefinementCount(pm);
+    HDF5Write2D(gBlocks, "derefinement_count", tmpID.data(), &loc_offset[0], &loc_cnt[0],
+                &glob_cnt[0], pl);
+  }
+
   Kokkos::Profiling::popRegion(); // write block metadata
 }
 
--- a/src/outputs/restart_hdf5.cpp
+++ b/src/outputs/restart_hdf5.cpp
@@ -138,10 +138,12 @@ RestartReaderHDF5::MeshInfo RestartReaderHDF5::GetMeshInfo() const {
     mesh_info.derefinement_count = ReadDataset<int>("/Blocks/derefinement_count");
   } catch (const std::runtime_error &e) {
     // File does not contain this dataset, so must be older. Set to default value of zero
-    PARTHENON_WARN(
-        "Restarting from an HDF5 file that doesn't contain /Blocks/derefinement_count. \n"
-        "If you are running with AMR, this may cause restarts to not be bitwise exact \n"
-        "with simulations that are run without restarting.");
+    if (Globals::my_rank == 0)
+      PARTHENON_WARN("Restarting from an HDF5 file that doesn't contain "
+                     "/Blocks/derefinement_count. \n"
+                     "If you are running with AMR, this may cause restarts to not be "
+                     "bitwise exact \n"
+                     "with simulations that are run without restarting.");
     mesh_info.derefinement_count = std::vector<int>(mesh_info.nbtotal, 0);
   }
   return mesh_info;
--- a/src/outputs/restart_hdf5.cpp
+++ b/src/outputs/restart_hdf5.cpp
@@ -134,6 +134,12 @@ RestartReaderHDF5::MeshInfo RestartReaderHDF5::GetMeshInfo() const {
   mesh_info.level_gid_lid_cnghost_gflag =
       ReadDataset<int>("/Blocks/loc.level-gid-lid-cnghost-gflag");
 
+  try {
+    mesh_info.derefinement_count = ReadDataset<int>("/Blocks/derefinement_count");
+  } catch (const std::runtime_error &e) {
+    // File does not contain this dataset, so must be older. Set to default value of zero
+    mesh_info.derefinement_count = std::vector<int>(mesh_info.nbtotal, 0);
+  }
   return mesh_info;
 }
 
--- a/src/outputs/restart_hdf5.cpp
+++ b/src/outputs/restart_hdf5.cpp
@@ -138,6 +138,10 @@ RestartReaderHDF5::MeshInfo RestartReaderHDF5::GetMeshInfo() const {
     mesh_info.derefinement_count = ReadDataset<int>("/Blocks/derefinement_count");
   } catch (const std::runtime_error &e) {
     // File does not contain this dataset, so must be older. Set to default value of zero
+    PARTHENON_WARN(
+        "Restarting from an HDF5 file that doesn't contain /Blocks/derefinement_count. \n"
+        "If you are running with AMR, this may cause restarts to not be bitwise exact \n"
+        "with simulations that are run without restarting.");
     mesh_info.derefinement_count = std::vector<int>(mesh_info.nbtotal, 0);
   }
   return mesh_info;
--- a/src/outputs/restart_hdf5.cpp
+++ b/src/outputs/restart_hdf5.cpp
@@ -134,16 +134,18 @@ RestartReaderHDF5::MeshInfo RestartReaderHDF5::GetMeshInfo() const {
   mesh_info.level_gid_lid_cnghost_gflag =
       ReadDataset<int>("/Blocks/loc.level-gid-lid-cnghost-gflag");
 
-  try {
+  auto status =
+      PARTHENON_HDF5_CHECK(H5Lexists(fh_, "Blocks/derefinement_count", H5P_DEFAULT));
+  if (status > 0) {
     mesh_info.derefinement_count = ReadDataset<int>("/Blocks/derefinement_count");
-  } catch (const std::runtime_error &e) {
+  } else {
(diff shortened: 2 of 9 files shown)


Yurlungur (approved):
This is gross but we gotta do what we gotta do

lroberts36 (approved):
This looks like a reasonable fix to me, sorry for not thinking of that before. A possibly cleaner choice, but that would require more work, would be to include the derefinement count in a separate dataset in the hdf5 file.

pgrete:
> This looks like a reasonable fix to me, sorry for not thinking of that before. A possibly cleaner choice, but that would require more work, would be to include the derefinement count in a separate dataset in the hdf5 file.

Given that the other PR is just 8 days in, I'm kind of in favor of the clean fix rather than carrying around a work-around (that might inspire similar extension in the future).
I should be able to get to that early next week.

pgrete:
(We could still merge this in now for an immediate resolution)

lroberts36:
> (We could still merge this in now for an immediate resolution)

Let me see if I can get this to work this morning.

pgrete:
My plan would be to extend the `mesh_info` object with a new vector containing the counters (which would by default set to 0 if not present).

lroberts36:
@pgrete and @Yurlungur: I just switched to making a new dataset and leaving `level-gid-lid...` as it was previous to #1073. I think this should be a clean fix. If the new dataset is not present in the file, HDF5 writes some complaints to screen but the exception is caught and it runs through.

tst/regression/test_suites/restart/parthinput.restart
```diff
@@ -39,9 +39,6 @@ x3max = 1
 ix3_bc = periodic
 ox3_bc = periodic
 
-derefine_count = 999999 # disable this since derefinement counter resets on
-                        # restart and thus leads to different results
-
```
lroberts36, line 44: I am not sure how this got into the restart test.

src/outputs/restart_hdf5.cpp
```diff
@@ -134,6 +134,16 @@ RestartReaderHDF5::MeshInfo RestartReaderHDF5::GetMeshInfo() const {
   mesh_info.level_gid_lid_cnghost_gflag =
       ReadDataset<int>("/Blocks/loc.level-gid-lid-cnghost-gflag");
 
+  try {
+    mesh_info.derefinement_count = ReadDataset<int>("/Blocks/derefinement_count");
+  } catch (const std::runtime_error &e) {
```
Yurlungur: knowing HDF5 I bet there's voluminous output here... I'm fine with this solution, but if we wanted to reduce the amount that HDF5 complains, you could instead check for the existence of `"/Blocks/derefinement_count"` with, e.g., `H5Lexists` (https://docs.hdfgroup.org/hdf5/v1_12/group___h5_l.html#title11). Despite the function being  documented as looking at links, it works for any "name" in an HDF5 file: https://forum.hdfgroup.org/t/check-if-dataset-exists/1454/5

src/outputs/restart_hdf5.cpp
```diff
@@ -134,6 +134,16 @@ RestartReaderHDF5::MeshInfo RestartReaderHDF5::GetMeshInfo() const {
...
+    // File does not contain this dataset, so must be older. Set to default value of zero
+    PARTHENON_WARN(
+        "Restarting from an HDF5 file that doesn't contain /Blocks/derefinement_count. \n"
+        "If you are running with AMR, this may cause restarts to not be bitwise exact \n"
+        "with simulations that are run without restarting.");
```
Yurlungur: Hide this warning behind an `my_rank == 0` check?

tst/regression/test_suites/restart/parthinput.restart
Yurlungur: I think it was there from the beginning but I thought you removed before. :shrug:

Yurlungur (approved):
Thanks for the quick fix, @lroberts36 . This cleaner solution seems good to me. Some nitpicks/hdf5 foo below.

src/outputs/restart_hdf5.cpp
lroberts36: Honestly, I am a little confused by the output. The beginning of `ReadDataset` does 
```c++
    // make sure dataset exists
    auto status = PARTHENON_HDF5_CHECK(H5Oexists_by_name(fh_, name.c_str(), H5P_DEFAULT));
    PARTHENON_REQUIRE_THROWS(
        status > 0, "Dataset '" + name + "' does not exist in HDF5 file " + filename_);
```
and that throw is what I am catching here I think.

src/outputs/restart_hdf5.cpp
lroberts36: fixed.

src/outputs/restart_hdf5.cpp
Yurlungur: Ah. That makes sense. In which case yeah I wouldn't expect HDF5 to complain at all. Bizarre.

Yurlungur: yeah in that case disregard my suggestion.

lroberts36: I am going to leave it as is for now.

lroberts36:
I will leave it to @pgrete to click automerge when he is happy with this.

src/outputs/restart_hdf5.cpp
```diff
@@ -134,6 +134,18 @@ RestartReaderHDF5::MeshInfo RestartReaderHDF5::GetMeshInfo() const {
   mesh_info.level_gid_lid_cnghost_gflag =
       ReadDataset<int>("/Blocks/loc.level-gid-lid-cnghost-gflag");
 
+  try {
+    mesh_info.derefinement_count = ReadDataset<int>("/Blocks/derefinement_count");
+  } catch (const std::runtime_error &e) {
+    // File does not contain this dataset, so must be older. Set to default value of zero
+    if (Globals::my_rank == 0)
```
pgrete: Extend to the check to only print with mesh refinement enabled?

src/outputs/parthenon_hdf5.cpp
```diff
@@ -603,6 +603,16 @@ void PHDF5Output::WriteBlocksMetadata_(Mesh *pm, hid_t file, const HDF5::H5P &pl
...
                 &loc_cnt[0], &glob_cnt[0], pl);
   }
+
+  {
+    // derefinement count
+    hsize_t loc_cnt[2] = {num_blocks_local, 1};
+    hsize_t glob_cnt[2] = {max_blocks_global, 1};
+    std::vector<int> tmpID = OutputUtils::ComputeDerefinementCount(pm);
+    HDF5Write2D(gBlocks, "derefinement_count", tmpID.data(), &loc_offset[0], &loc_cnt[0],
+                &glob_cnt[0], pl);
```
pgrete, line 652: Shall we always write (i..e, a bunch of zeros for static runs) or move the `if (pm->pmr)` from `ComputeDerefinementCount` here to only write the deref counter for multilevel sims?

tst/regression/test_suites/restart/parthinput.restart
pgrete: So what does this mean if the tests still pass independent of this variable?
Is our tolerance too loose?
I somehow assumed that this, in principle, you result in a failed test.
I'm going to trigger the full suite to see what happens.

lroberts36: By setting the `derefine_count` to a huge number, it was ensured that derefinement was never triggered. This meant that restarts didn't depend on on the derefinement counter of each block and were exact in the test. Now that we store the derefinement counter in the restart files, `derefine_count` can be set to any number and the restarts will still be exact (while previously it would not be unless it was set to zero or a huge number).

src/outputs/parthenon_hdf5.cpp
lroberts36: I kind of lean towards always writing to file, with or without refinement, just to keep the structure of the files the same. I think this should happen with things written as they are now.

src/outputs/restart_hdf5.cpp
lroberts36: I think because of the HDF5 spew that comes out when the dataset doesn't exist, we should probably always leave the warning in to provide some explanation to the user of what is going on.

tst/regression/test_suites/restart/parthinput.restart
pgrete: Ah, I missed that the restart test does not compare to our (precalculated) gold files but to a freshly generated state, so, yes, this makes a lot of sense now.

src/outputs/restart_hdf5.cpp
pgrete: Given that the (quite extensive) error message is printed by every rank, I now went for the `H5Lexists` option as @Yurlungur suggested (and we already use for the `SparseInfo` field).

src/outputs/restart_hdf5.cpp
pgrete: Given that I removed the hdf5 error message, I now added the additional check.

PR #1090, opened 2024-05-23
SMR with swarms

brryan:
## PR Summary

Static mesh refinement currently fails with particles; particles are communicated to incorrect fine blocks across refinement boundaries. This PR fixes that by changing the way we set the internal neighbor connectivity matrix in the swarm machinery. Instead of a bunch of hardcoded indexing specific to dimension and no refinement, I create test points for each halo neighbor block at the maximum possible refinement level, and test whether those points are contained by any neighbor blocks. Works for neighbor connectivity including across trees and across periodic boundaries. 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/src/interface/swarm_comms.cpp
+++ b/src/interface/swarm_comms.cpp
@@ -68,8 +68,8 @@ void Swarm::SetNeighborIndices_() {
   for (int k = 0; k < 4; k++) {
     for (int j = 0; j < 4; j++) {
       for (int i = 0; i < 4; i++) {
-        // Midpoint of meshblock at highest possible refinement level in this i,j,k ghost
-        // halo.
+        // Midpoint of meshblock at highest possible refinement level in this i,j,k
+        // ghost halo.
         std::array<Real, 3> x_test = {bsize.xmin_[0] + (i - 0.5) * dx_test[0],
                                       ndim < 2 ? (bsize.xmin_[1] + bsize.xmax_[1]) / 2.
                                                : bsize.xmin_[1] + (j - 0.5) * dx_test[1],
@@ -129,6 +129,36 @@ void Swarm::SetNeighborIndices_() {
       }
     }
   }
+
+  // Draft alternative approach due to LFR utilizing mesh offset comparisons at the
+  // highest refinement level
+  // auto ll_block = pmb->loc.GetDaughter(0, 0, 0);
+  // int finest_level = pmb->loc.level() + 1;
+  // for (auto &n : pmb->neighbors) {
+  //  std::vector<LogicalLocation> dlocs;
+  //  auto &nloc =
+  //      n.loc; // Would need to use the location in the coordinates of the origin tree
+  //  if (nloc.level() == finest_level) {
+  //    dlocs.emplace_back(nloc);
+  //  } else if (nloc.level() == finest_level) {
+  //    dlocs = nloc.GetDaughters(ndim);
+  //  } else if (nloc.level() == finest_level - 2) {
+  //    auto tlocs = nloc.GetDaughters(ndim);
+  //    for (auto &t : tlocs) {
+  //      auto gdlocs = t.GetDaughters(ndim);
+  //      dlocs.insert(dlocs.end(), gdlocs.begin(), gdlocs.end());
+  //    }
+  //  } else {
+  //    PARTHENON_FAIL("Proper nesting is not being respected.");
+  //  }
+  //  for (auto &d : dlocs) {
+  //    const int k = d.lx3() - ll_block.lx3() + 1;
+  //    const int j = d.lx2() - ll_block.lx2() + 1;
+  //    const int i = d.lx1() - ll_block.lx1() + 1;
+  //    if (i >= 0 && i <= 3 && j >= 0 && j <= 3 && k >= 0 && k <= 3)
+  //      neighbor_indices_h(k, j, i) = n.gid;
+  //  }
+  //}
 
   neighbor_indices_.DeepCopy(neighbor_indices_h);
 }
--- a/src/interface/swarm_comms.cpp
+++ b/src/interface/swarm_comms.cpp
@@ -30,55 +30,7 @@ namespace parthenon {
 /// position in terms of neighbor blocks based on spatial position. See
 /// GetNeighborBlockIndex()
 ///
-void Swarm::SetNeighborIndices1D_() {
-  auto pmb = GetBlockPointer();
-  const int ndim = pmb->pmy_mesh->ndim;
-  auto neighbor_indices_h = neighbor_indices_.GetHostMirror();
-
-  // Initialize array in event of zero neighbors
-  for (int k = 0; k < 4; k++) {
-    for (int j = 0; j < 4; j++) {
-      for (int i = 0; i < 4; i++) {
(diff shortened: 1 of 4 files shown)


src/interface/swarm_comms.cpp
```diff
@@ -26,265 +26,105 @@
...
+          if (pmb->boundary_flag[4] == BoundaryFlag::periodic) {
+            if (x_test[2] < msize.xmin(X3DIR)) {
+              x_test[2] = msize.xmax(X3DIR) - (msize.xmin(X3DIR) - x_test[2]);
+            }
+          }
+          if (pmb->boundary_flag[5] == BoundaryFlag::periodic) {
+            if (x_test[2] > msize.xmax(X3DIR)) {
+              x_test[2] = msize.xmin(X3DIR) + (x_test[2] - msize.xmax(X3DIR));
+            }
+          }
```
brryan, line 115: @lroberts36 We talked about a method (EDIT: you came up with this idea and shared it with me) of maximally refining the current block and all neighbor blocks and comparing all refined blocks to accommodate periodic boundary conditions, but because `mesh_size` is the entire extent of the hyperrectangular domain, just applying boundary conditions to the test point works fine (I tested downstream with a periodic BC across two trees). I think this method also results in fewer comparison operations since we are comparing each ghost halo test block with each neighbor, rather than each ghost halo test block with each daughter of each neighbor. 

So I'm inclined to suggest this method get merged in because it's simple, supports the general hyperrectangular case, and is already done. We have a near-term need for SMR downstream. But if you think this is really dangerous in terms of future proofing lets discuss more.

Yurlungur (approved):
Exciting to see this in! And it's a nice cleanup!

src/interface/swarm_comms.cpp
```diff
@@ -26,265 +26,105 @@
...
+                                               : bsize.xmin_[2] + (k - 0.5) * dx_test[2]};
+
+        // Account for periodic boundary conditions by applying BCs to x_test
+        // Assumes mesh is hyper-rectangular and Mesh::mesh_size represents the entire
+        // domain.
+        auto &msize = pmb->pmy_mesh->mesh_size;
+        if (pmb->boundary_flag[0] == BoundaryFlag::periodic) {
+          if (x_test[0] < msize.xmin(X1DIR)) {
+            x_test[0] = msize.xmax(X1DIR) - (msize.xmin(X1DIR) - x_test[0]);
+          }
```
lroberts36, line 86: Do you not also require a check if `x_test[0] > msize.xmax(X1DIR)` (and likewise for other directions)?

lroberts36: oh, nm. The check was just not visible near by in the diff.

src/interface/swarm_comms.cpp
```diff
@@ -26,265 +26,105 @@
...
+        // Loop over neighbor blocks and see if any contains this test point.
+        for (int n = 0; n < pmb->neighbors.size(); n++) {
+          NeighborBlock &nb = pmb->neighbors[n];
+          const auto &nbsize = nb.block_size;
+          if ((x_test[0] > nbsize.xmin_[0] && x_test[0] < nbsize.xmax_[0]) &&
+              (x_test[1] > nbsize.xmin_[1] && x_test[1] < nbsize.xmax_[1]) &&
+              (x_test[2] > nbsize.xmin_[2] && x_test[2] < nbsize.xmax_[2])) {
+            neighbor_indices_h(k, j, i) = n;
+            break;
+          }
```
lroberts36, line 127: I think this is a reasonable way to do it for now, but I think there will be some general cases in which it won't work. Just so the more general way doesn't get forgotten, it might be worth it to add the aspirational code below as a comment
```c++
auto ll_block = pmb->loc.GetDaughter(0, 0, 0); 
int finest_level = pmb->loc.level() + 1;
for (auto &n : pmb->neighbors) {
  std::vector<LogicalLocation> dlocs;
  auto &nloc = n->loc; // Would need to use the location in the coordinates of the origin tree 
  if (nloc.level() == finest_level) {
    dlocs.emplace_back(nloc); 
  } else if (nloc.level() == finest_level) {
    dlocs = nloc.GetDaughters(ndim); 
  } else if (nloc.level() == finest_level - 2) {
    auto tlocs = nloc.GetDaughters(ndim);
    for (auto &t : tlocs) {
      auto gdlocs = t.GetDaughters(ndim); 
      dlocs.insert(dlocs.end(), gdlocs.begin(), gdlocs.end());
    }
   } else {
    PARTHENON_FAIL("Proper nesting is not being respected.");  
  }
  for (auto &d : dlocs) {
    const int k = d.lx3() - ll_block.lx3() + 1;
    const int j = d.lx2() - ll_block.lx2() + 1;
    const int i = d.lx1() - ll_block.lx1() + 1;
    if (i >=0 && i <= 3 && j >=0 && j <=3 && k >=0 && k <= 3)
        neighbor_indices_h(k, j, i) = n->gid;
  }
}
```
using this requires the neighbor location in the logical coordinate system of the origin block, which is only available in #1047.

lroberts36 (approved):
This LGTM. I definitely like to see the negative line count and removing special casing! I think the solution of using the physical coordinates is good for now, but I included some aspirational code for essentially how the neighbor finding should look for general forests.

src/interface/swarm_comms.cpp
brryan: Thanks @lroberts36! I agree this should go in as a comment. One thought now: is this a solution only for a subset of general cases, for example some kind of non-hyperrectangular tiling domain with periodic boundary conditions in some dimension? The more general case, of unstructured meshblock connectivity, will probably need a substantial refactor of swarms; for example `neighbor_indices()` will probably no longer make any sense in that case. I do think something like your solution above would be part of that refactor though.

lroberts36: I guess a basic question is can particles move into corners over a timestep or can they only cross faces? If they can only move across faces, I think this would actually work in the most general case (although we would require performing a coordinate transformation on the particle positions when they move across blocks). If you want them to be able to move into corners and edges, then we need to work a little more since there can be more than one block that connects at a single edge or corner in an unstructured mesh. I think you could always demand that particles only go across faces, but maybe this is too much of a performance hit (since particles would have to be moved across blocks twice to effectively cross a corner).

brryan: Right that was my concern, is that the general allowed approach now is just that particles get pushed ~`dx`  rather than being obliged to move across faces. To go to a fully general mesh I think the right approach would be to implement an alternative approach to pushing particles, maybe through parthenon, that explicitly moves particles between cells via face crossings. A number of details would need to be worked out to do that right I think.

brryan: I did try your code out and it compiled but I ran into `PARTHENON_FAIL("Proper nesting is not being respected.");` at runtime. Still very happy to include as a comment for future use. Thanks for looking at this

lroberts36: Yeah, as written it won't work for multi-tree forests or periodic boundaries. It is a one word change to make it correct when the neighbor logical location in the coordinates of the origin block is available though, I think.

PR #1091, opened 2024-05-23
Add a vector wave equation example

lroberts36:
## PR Summary

This PR adds a vector wave equation to the `fine_field` example. That is, we solve the equations 
```math
\partial_t \vec{C} + \nabla \times \vec{D} = 0
```
```math
\partial_t \vec{D} - \nabla \times \vec{C} = 0
```
```math
\nabla \cdot \vec{C} = 0
```
```math
\nabla \cdot \vec{D} = 0
```
using an upwind constrained transport scheme (currently just using piecewise constant reconstruction). The face centered fields are evolved and prolongation is done using the scheme of Toth & Roe.

TODO: 
- [x] Actually initialize $\vec{C}$ and $\vec{D}$ to something
- [x] Include Toth & Roe prolongation so that divergence constraint is preserved in AMR

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [ ] Docs build


Yurlungur:
> Include Toth & Roe prolongation so that divergence constraint is preserved in AMR

Is this needed if using edge fields?

lroberts36:
> > Include Toth & Roe prolongation so that divergence constraint is preserved in AMR
> 
> Is this needed if using edge fields?

@Yurlungur: No, it shouldn't be. I had started writing it using face fields though.

src/mesh/mesh-amr_loadbalance.cpp
```diff
@@ -213,11 +214,13 @@ bool TryRecvFineToCoarse(int lid_recv, int send_rank, const LogicalLocation &fin
...
-        if (ox2 == 0) jb.e -= TopologicalOffsetJ(te);
+        if (ox3 == 0 && ndim > 2) kb.e -= TopologicalOffsetK(te);
+        if (ox2 == 0 && ndim > 1) jb.e -= TopologicalOffsetJ(te);
         if (ox1 == 0) ib.e -= TopologicalOffsetI(te);
-        const int ks = (ox3 == 0) ? 0 : (kb.e - kb.s + 1 - TopologicalOffsetK(te));
-        const int js = (ox2 == 0) ? 0 : (jb.e - jb.s + 1 - TopologicalOffsetJ(te));
+        const int ks =
+            (ox3 == 0 || ndim < 3) ? 0 : (kb.e - kb.s + 1 - TopologicalOffsetK(te));
+        const int js =
+            (ox2 == 0 || ndim < 2) ? 0 : (jb.e - jb.s + 1 - TopologicalOffsetJ(te));
```
lroberts36, line 223: This fixes a bug which caused the z-component of coarsened face fields to be filled with zeros in 2D. I don't think anyone is likely to have run into it yet though (aside from in this example).

example/fine_advection/advection_driver.cpp
```diff
@@ -89,18 +102,22 @@ TaskCollection AdvectionDriver::MakeTaskCollection(BlockList_t &blocks, const in
...
+    static auto desc = parthenon::MakePackDescriptor<phi>(
+        pmesh->resolved_packages.get(), {parthenon::Metadata::WithFluxes},
+        {parthenon::PDOpt::WithFluxes});
     using pack_desc_t = decltype(desc);
 
-    static auto desc_fine =
-        parthenon::MakePackDescriptor<advection_package::Conserved::scalar_fine>(
-            pmesh->resolved_packages.get(), {parthenon::Metadata::WithFluxes},
-            {parthenon::PDOpt::WithFluxes});
+    static auto desc_fine = parthenon::MakePackDescriptor<phi_fine>(
```
pdmullen, line 104: `phi_fine` could be packed in the same pack as `phi`, right?  This is just a convenience here so that you can assign `pack_desc_fine_t` which is later used as a template argument (e.g., `Stokes`)?

example/fine_advection/advection_driver.cpp
```diff
@@ -115,37 +132,31 @@ TaskCollection AdvectionDriver::MakeTaskCollection(BlockList_t &blocks, const in
...
-    auto flux_div = tl.AddTask(set_flx, Stokes<pack_desc_t>, parthenon::CellLevel::same,
-                               parthenon::TopologicalType::Cell, desc, pmesh->ndim,
-                               mc0.get(), mdudt.get());
-
-    auto flux_div_fine = tl.AddTask(
-        set_flx, Stokes<pack_desc_fine_t>, parthenon::CellLevel::fine,
-        parthenon::TopologicalType::Cell, desc_fine, pmesh->ndim, mc0.get(), mdudt.get());
+    auto vf_dep = none;
+    for (auto edge : std::vector<TE>{TE::E1, TE::E2, TE::E3}) {
+      vf_dep = tl.AddTask(vf_dep, advection_package::CalculateVectorFluxes<C, D>, edge,
```
pdmullen, line 130: I dislike re-using TaskIDs like this (does this actually all work okay under the hood in the tasking infrastructure?).  I understand that this situation is a bit unique in that the for loop is called over a subset of topological elements, but is there a path here that doesn't involve re-using `vf_dep`?

example/fine_advection/advection_package.cpp
```diff
@@ -66,23 +66,44 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
...
+  pkg->AddField<Conserved::phi>(Metadata({Metadata::Cell, Metadata::Independent,
+                                          Metadata::WithFluxes, Metadata::FillGhost}));
+  pkg->AddField<Conserved::phi_fine_restricted>(
       Metadata({Metadata::Cell, Metadata::Derived, Metadata::OneCopy}));
+
+  Metadata m(
+      {Metadata::Face, Metadata::Independent, Metadata::WithFluxes, Metadata::FillGhost});
+  m.RegisterRefinementOps<parthenon::refinement_ops::ProlongateSharedMinMod,
+                          parthenon::refinement_ops::RestrictAverage,
+                          parthenon::refinement_ops::ProlongateInternalTothAndRoe>();
```
pdmullen, line 81: nice

example/fine_advection/advection_package.hpp
```diff
@@ -33,14 +33,23 @@ namespace advection_package {
 using namespace parthenon::package::prelude;
 
 namespace Conserved {
-VARIABLE(advection, scalar);
-VARIABLE(advection, scalar_fine);
-VARIABLE(advection, scalar_fine_restricted);
+VARIABLE(advection, phi);
```
pdmullen, line 36: We are using the `VARIABLE` macro in various places throughout Parthenon now.  Not for this MR, but should we centrally locate this macro so that even downstream applications can use it?

example/fine_advection/advection_package.hpp
```diff
@@ -81,6 +90,120 @@ TaskStatus CalculateFluxes(pack_desc_t &desc, parthenon::TopologicalElement FACE
   return TaskStatus::complete;
 }
 
+template <class var, class flux_var>
```
pdmullen, line 93: clever to template function on `var` (and `flux_var`).

example/fine_advection/parthenon_app_inputs.cpp
```diff
@@ -49,9 +57,8 @@ void ProblemGenerator(MeshBlock *pmb, ParameterInput *pin) {
 
   auto coords = pmb->coords;
 
-  using scalar = advection_package::Conserved::scalar;
-  using scalar_fine = advection_package::Conserved::scalar_fine;
-  static auto desc = parthenon::MakePackDescriptor<scalar, scalar_fine>(data.get());
+  using namespace advection_package::Conserved;
+  static auto desc = parthenon::MakePackDescriptor<phi, phi_fine, C, D>(data.get());
```
pdmullen, line 61: this answers my above question.

example/fine_advection/stokes.hpp
```diff
@@ -24,9 +24,10 @@ namespace advection_example {
 using namespace parthenon::driver::prelude;
 
 template <class pack_desc_t>
-TaskStatus WeightedSumData(parthenon::CellLevel cl, parthenon::TopologicalElement te,
-                           pack_desc_t &pd, MeshData<Real> *in1, MeshData<Real> *in2,
-                           Real w1, Real w2, MeshData<Real> *out) {
+TaskStatus WeightedSumDataElement(parthenon::CellLevel cl,
```
pdmullen, line 27: not for this MR, but this feels like it belongs in `interface/update.cpp` just as much as `UpdateWithFluxDivergence` does.

src/prolong_restrict/pr_ops.hpp
```diff
@@ -371,6 +371,91 @@ struct ProlongateInternalAverage {
...
+  KOKKOS_FORCEINLINE_FUNCTION static void
+  Do(const int l, const int m, const int n, const int k, const int j, const int i,
+     const IndexRange &ckb, const IndexRange &cjb, const IndexRange &cib,
+     const IndexRange &kb, const IndexRange &jb, const IndexRange &ib,
+     const Coordinates_t &coords, const Coordinates_t &coarse_coords,
+     const ParArrayND<Real, VariableState> *,
+     const ParArrayND<Real, VariableState> *pfine) {
+    using namespace util;
+
+    if constexpr (!IsSubmanifold(fel, cel)) {
```
pdmullen, line 401: Can you dumb this branch down for me? i.e., what is it checking again (without using the word submanifold)?

pdmullen (approved):
Very happy to see a topological element regression test going in!

example/fine_advection/advection_driver.cpp
lroberts36: yes, there is nothing preventing you from packing fine and regular variables in the same pack. It is just not desirable here.

example/fine_advection/advection_driver.cpp
lroberts36: I think there is no issue under the hood with re-initializing `TaskID` objects with new task identifiers. 

Although there is probably a way to not reuse the task ID here, it seems like it might be over kilI to change things. Maybe take that with a grain of salt, since I personally like the pattern of reusing TaskID objects as much as possible since it makes it easier to add and remove tasks with sequential dependency when debugging or looping as above.

example/fine_advection/stokes.hpp
lroberts36: Agreed. The next thing I plan to do with this is going to be making the update functions in this example compatible with sparse fields, remove the old sparse example, and add sparse fields to this example. Once that is all working, I think it makes sense to move this stuff to `interface/` although it semi-duplicates some things that are in there currently so we will have to decide how to deal with that.

src/prolong_restrict/pr_ops.hpp
lroberts36: It is just checking if the topological element `fel` corresponds to some part of the boundary of `cel`, e.g. if `cel = TE::F1` then this would be true for `fel = (TE::NN, TE::E2, TE::E3)` but false for all other elements. Internal prolongation only has to occur on these elements.

c-prather (approved):
LGTM, this will be a great extra test!

src/prolong_restrict/pr_ops.hpp
```diff
@@ -371,6 +371,91 @@ struct ProlongateInternalAverage {
   }
 };
 
+struct ProlongateInternalTothAndRoe {
+  static constexpr bool OperationRequired(TopologicalElement fel,
+                                          TopologicalElement cel) {
```
Yurlungur, line 381: I'm sure it was tricky to get this implementation right (thanks @bprather ) but I'm pleasantly surprised by how simple the implementation is here. It might be worth making a comment that this struct is specifically for divergence free calculations though.

src/basic_types.hpp
```diff
@@ -152,7 +152,7 @@ enum class TopologicalElement : std::size_t {
 enum class TopologicalType { Cell, Face, Edge, Node };
 
 KOKKOS_FORCEINLINE_FUNCTION
-TopologicalType GetTopologicalType(TopologicalElement el) {
+constexpr TopologicalType GetTopologicalType(TopologicalElement el) {
```
Yurlungur, line 155: :+1:

example/fine_advection/advection_package.hpp
Yurlungur: I second this 👍

example/fine_advection/parthenon_app_inputs.cpp
```diff
@@ -31,7 +31,15 @@ using namespace parthenon;
...
-
+namespace {
+KOKKOS_INLINE_FUNCTION
+Real sign(Real a, Real b) {
+  if (a == 0.0) {
+    return b > 0.0 ? 1.0 : -1.0;
+  }
+  return a > 0.0 ? 1.0 : -1.0;
+}
+} // namespace
```
Yurlungur, line 42: should this be in parthenon utils? (In another PR...)

example/fine_advection/parthenon_app_inputs.cpp
```diff
@@ -64,7 +71,7 @@ void ProblemGenerator(MeshBlock *pmb, ParameterInput *pin) {
   const int b = 0;
   const int ndim = pmb->pmy_mesh->ndim;
   const int nghost = parthenon::Globals::nghost;
-  pmb->par_for(
+  parthenon::par_for(
```
Yurlungur, line 74: :+1:

example/fine_advection/advection_package.hpp
```diff
@@ -33,14 +33,23 @@ namespace advection_package {
...
 
 namespace Conserved {
-VARIABLE(advection, scalar);
-VARIABLE(advection, scalar_fine);
-VARIABLE(advection, scalar_fine_restricted);
+VARIABLE(advection, phi);
+VARIABLE(advection, phi_fine);
+VARIABLE(advection, phi_fine_restricted);
+VARIABLE(advection, C);
+VARIABLE(advection, D);
```
Yurlungur, line 40: given these could be anything, I guess C and D are fine names but they seem rather non standard. I would prefer A (or E) and B or psi and pi (to go with phi)

Yurlungur (approved):
Very excited to see this modernization and demonstration for how to do constrained transport.

example/fine_advection/parthenon_app_inputs.cpp
lroberts36: yeah, probably... but for later

example/fine_advection/advection_package.hpp
lroberts36: I explicitly tried to avoid $\vec{A}$, $\vec{B}$, and $\vec{E}$.

Yurlungur: I figured... Well my preference is for psi and pi but I suppose it doesn't matter much

lroberts36: I think I am going to leave $\vec{C}$ and $\vec{D}$. $\vec{\psi}$ and $\vec{\pi}$ seem scalary to me, and more importantly keeping it the same requires no search and replace...

example/fine_advection/advection_package.hpp
lroberts36: yeah, I think this makes sense, but maybe for another PR.

src/prolong_restrict/pr_ops.hpp
lroberts36: comment added

lroberts36:
@par-hermes format

PR #1092, opened 2024-05-29
Updates to `DataCollection`

lroberts36:
## PR Summary

This PR makes some changes to `DataCollection` so that calling `pmesh->mesh_data.GetOrAdd(some_label, partition)` works whether or not `some_label` has been previously added to the `MeshBlockData` container. Additionally, it makes it so that `Add` will now work correctly for `MeshData` (previously it didn't use the correct labeling and containers could be overwritten).

Possible other things to do:
- Formalize partitioning and move it to `Mesh`
- Get all of the `DataCollection` interface to work with both `Mesh` and `MeshData` 
- Remove `Initialize` methods and replace them with constructors since they break RAII 

## PR Checklist

- [x] #991 is merged
- [x] Code passes cpplint
- [x] Code is formatted
- [ ] Docs build


src/interface/meshblock_data.hpp
```diff
@@ -122,18 +115,27 @@ class MeshBlockData {
...
+  }
+
+  template <class SRC_t, typename ID_t = std::string>
+  void Initialize(const std::shared_ptr<StateDescriptor> resolved_packages_in,
+                  const std::shared_ptr<SRC_t> src, const std::vector<ID_t> &vars = {},
+                  const bool shallow_copy = false) {
+    if constexpr (!(std::is_same_v<SRC_t, MeshBlockData<Real>> ||
+                    std::is_same_v<SRC_t, MeshBlock>)) {
+      // We don't allow other types
+      static_assert(always_false<SRC_t>);
```
Yurlungur: Is there a reason not to just `static_assert` the boolean expression that you have in `if constexpr` here? Also you can add a string to the `static_assert` which I think is a nice readability enhancement.

src/interface/meshblock_data.hpp
```diff
@@ -155,23 +157,51 @@ class MeshBlockData {
...
+            for (const auto &v2 : vars) {
+              if (src->GetVarPtr(v2)->label() == flx_name) found = true;
+            }
+            if (!found) add_var(src->GetVarPtr(flx_name));
           }
-          if (!found) add_var(src->GetVarPtr(flx_name));
         }
+      } else {
+        PARTHENON_FAIL(
+            "Variable subset selection not yet implemented for MeshBlock input.");
```
Yurlungur, line 187: :+1:

src/interface/meshblock_data.hpp
```diff
@@ -155,23 +157,51 @@ class MeshBlockData {
...
+      }
+    }
+
+    // TODO(LFR): Not sure why we only do this in the MeshBlock case, but this carries
+    // over from the previous iteration.
+    if constexpr (std::is_same_v<SRC_t, MeshBlock>) {
+      if (stage_name_ == "base") {
+        const auto &swarm_container = GetSwarmData();
+        swarm_container->Initialize(resolved_packages, GetBlockSharedPointer());
+      }
```
Yurlungur, line 197: Ping @pdmullen @brryan

src/interface/data_collection.hpp
```diff
@@ -44,9 +51,16 @@ class DataCollection {
...
-  template <typename ID_t>
-  std::shared_ptr<T> &Add(const std::string &name, const std::shared_ptr<T> &src,
+  template <class SRC_t, typename ID_t>
+  std::shared_ptr<T> &Add(const std::string &name, const std::shared_ptr<SRC_t> &src,
                           const std::vector<ID_t> &fields, const bool shallow) {
+    if constexpr (!((std::is_same_v<SRC_t, MeshBlock> &&
+                     std::is_same_v<T, MeshBlockData<Real>>) ||
+                    std::is_same_v<SRC_t, T>)) {
+      // SRC_t and T are incompatible
+      static_assert(always_false<SRC_t>);
```
Yurlungur: Why not just `static_assert` the boolean in the if constexpr and add the comment as a string?

src/interface/data_collection.hpp
```diff
@@ -105,6 +119,28 @@ class DataCollection {
...
+      return stage_label;
+    }
+  }
+  std::string GetKey(const std::string &stage_label, std::optional<int> partition_id,
+                     std::optional<int> gmg_level) {
+    std::string key = stage_label;
+    if (partition_id) key = key + "_part-" + std::to_string(*partition_id);
+    if (gmg_level) key = key + "_gmg-" + std::to_string(*gmg_level);
+    return key;
+  }
```
Yurlungur, line 140: :+1: I approve of separating this out

src/utils/concepts_lite.hpp
```diff
@@ -16,6 +16,13 @@
 #include <type_traits>
 #include <utility>
 
+// This is a class template that is required for doing something like static_assert(false)
+// in constexpr if blocks. Actually writing static_assert(false) will always cause a
+// compilation error, even if it is an unchosen constexpr if block. This is fixed in C++23
+// I think.
+template <class...>
+constexpr std::false_type always_false{};
```
Yurlungur, line 24: I don't think this is necessary if you write the `static_assert`s slightly differently. See above.

Yurlungur (approved):
Thanks for cleaning this up. I think these are good changes.

src/interface/meshblock_data.hpp
lroberts36: Yeah, there is some reason that `static_assert(false)` will always trigger a compiler failure even if that constexpr if branch isn't triggered. Having something that always evaluates to false that has a template argument does not get triggered in contrast. Apparently they fix this stupidity in c++23.

src/interface/data_collection.hpp
lroberts36: see above. String is probably useful though.

src/utils/concepts_lite.hpp
lroberts36: So I just went and tested this on compiler explorer, and `static_assert(false);` seems to work fine on gcc and clang, but StackOverflow leads me to believe that this is not guaranteed to work with all compilers. Adding a text message doesn't change anything I think.

Yurlungur: I think I misunderstood this comment you made then... I thought you were saying `static_assert(false)` doesn't work, while `static_assert(some_constexpr_that_evaluates_to_false, "string")` does. I use the latter everywhere in singularity...

lroberts36: I think the distinction is that `some_constexpr_that_evaluates_to_false` has to somehow depend on template parameters (at least based on my cursory reading of StackOverflow).

Yurlungur: But it does, doesn't it?

lroberts36: [StackOverflow discussion](https://stackoverflow.com/questions/38304847/how-does-a-failed-static-assert-work-in-an-if-constexpr-false-block), based on that I believe 
```c++
if constexpr (std::is_same<T, int>::value) {
  // do something
} else {
  static_assert(false, "blah"); // Technically ill formed
  static_assert(true == false, "blah"); // Technically ill formed
  static_assert(std::is_same<T, int>::value, "blah"); // valid code, since the argument depends on template parameter T
}
```

Yurlungur: Ah I see. Ok. :+1:

src/interface/meshblock_data.hpp
lroberts36: added string to assert.

src/interface/data_collection.hpp
lroberts36: string added

src/interface/meshblock_data.hpp
brryan: I don't see any issue with this. Swarms are `OneCopy`-like and always belong to the base stage so it doesn't really make sense to `Add` a new `MeshBlockData` to the `DataCollection` based on an existing `MeshBlockData`. We have thought a bit about the best way to express this through the parthenon interface but don't have any good thoughts at the moment.

pdmullen: What was the concern here? Sorry I am late to the party... 

We currently adopt a model wherein the "base" MeshBlockData register contains our swarm data.  We have checks that swarms are only accessed through said register.

lroberts36: I just wanted to know if this was the desired behavior and if there were any side effects of what I did.

pdmullen: I'll test downstream first thing tomorrow.

pdmullen:
w.r.t. swarms, I tested this downstream and all looks good.  I noticed that this PR is targeting `lroberts36/add-fine-variables` instead of `develop`... intentional?

lroberts36:
> I noticed that this PR is targeting `lroberts36/add-fine-variables` instead of `develop`... intentional?

@pdmullen: yeah, but only because I was using the new test I added there to check that removing the `MeshBlockData` boiler plate worked.

Yurlungur:
@lroberts36 are we waiting for someone to review this?

lroberts36:
> @lroberts36 are we waiting for someone to review this?

@Yurlungur: @pgrete said he would on Thursday, but I think he has been under the weather the last few days.

src/interface/data_collection.hpp
```diff
@@ -105,6 +120,28 @@ class DataCollection {
...
+      return key;
+    } else {
+      return stage_label;
+    }
+  }
+  std::string GetKey(const std::string &stage_label, std::optional<int> partition_id,
+                     std::optional<int> gmg_level) {
+    std::string key = stage_label;
+    if (partition_id) key = key + "_part-" + std::to_string(*partition_id);
+    if (gmg_level) key = key + "_gmg-" + std::to_string(*gmg_level);
```
pgrete, line 138: Should there be a debug_require that only partition_id or gmg_level or neither is passed?

src/interface/meshblock_data.hpp
pgrete: Did the test happen? @pdmullen

example/fine_advection/advection_driver.cpp
```diff
@@ -62,17 +62,6 @@ TaskCollection AdvectionDriver::MakeTaskCollection(BlockList_t &blocks, const in
...
-  if (stage == 1) {
-    for (int i = 0; i < blocks.size(); i++) {
-      auto &pmb = blocks[i];
-      // first make other useful containers
-      auto &base = pmb->meshblock_data.Get();
-      pmb->meshblock_data.Add("dUdt", base);
-      for (int s = 1; s < integrator->nstages; s++)
-        pmb->meshblock_data.Add(stage_name[s], base);
-    }
-  }
```
pgrete, line 74: Should this be removed in the other examples, too, to decrease likelihood of reusing this pattern?

pgrete (approved):
LGTM

"Remove Initialize methods and replace them with constructors since they break RAII" is possibly meant for a future PR, isn't it? In that case -> new issue?

src/interface/meshblock_data.hpp
lroberts36: They did, there is an old comment from Patrick below.

src/interface/data_collection.hpp
lroberts36: no, the GMG levels are also partitioned. Although to be fair, partitioning has always been broken for multigrid, i.e. it only works when all blocks are included in the partition. This is on my list of things to fix.

example/fine_advection/advection_driver.cpp
lroberts36: Probably, but let's leave it for a future PR. I think it would be nice to work to clean up the examples in the near future.

lroberts36:
> "Remove Initialize methods and replace them with constructors since they break RAII" is possibly meant for a future PR, isn't it? In that case -> new issue?

@pgrete: I think I was just annoyed when I wrote this comment. It is probably not worth the effort to remove the initialization methods and replace them with constructors.

PR #911, opened 2023-07-19
Geometric Multigrid

lroberts36:
Screen Shot 2023-09-26 at 1 13 01 PM
Solution and residual of Poisson equation after fifteen iterations of MG-BiCGStab with a step function, circular right hand side centered at zero, $D=1000$ in the L-shaped region and $D=1$ elsewhere. 

## PR Summary

This PR adds the infrastructure required for implementing geometric multi-grid (GMG) solvers in Parthenon. Additionally, it adds an example that solves the Poisson equation $\nabla \cdot D \nabla u = f$ (possibly with rapidly spatially varying $D$) using both raw FAS multigrid v-cycles with FAC like grids and using this same multigrid solver as a pre-conditioner for the Krylov subspace method BiCGStab. Good convergence is seen on both uniform and non-uniform grids (including flux correction) and the MG-BiCGStab method provides rapid convergence on problems with rapidly varying $D$ (which cannot be efficiently solved by either BiCGStab or MG V-cylces alone). The implementation of a GMG solver task list and its use as a preconditioner is shown in the `poisson_gmg` example, but I have not included generic solver routines within Parthenon itself. Generic solvers probably could be provided, since a user probably only needs to provide a method for applying there matrix to a vector and a method for smoothing (e.g. for doing Jacobi iterations).

In terms of infrastructure, multigrid generally requires a hierarchy of increasingly coarsened meshes, a way to restrict a field from a finer mesh to the next coarser mesh, a way to prolongate data from a coarser mesh to a finer mesh, and a mechanism for intra-level communication between blocks at each GMG level.

To do this in Parthenon, a hierarchy of de-refined block meshes are created from the base mesh during initialization and after each AMR step[^1]. The blocks in these coarsened meshes are assigned a global id (`gid`) by iterating through them in Morton order, but starting from the largest `gid` of the next finest mesh. This gives each block in the GMG hierarchy a unique identifier. Once these de-refined block lists are built, a same level neighbor block finding algorithm is performed on each level[^2], the parent block on the next grid up in the GMG hierarchy is found, and all daughter blocks on the next grid down in the GMG hierarchy are found. All of these relationships are stored as arrays of `NeighborBlock`s in `MeshBlock`. The coarsened levels can go above the root grid of the simulation and the blocks change in size at these coarser levels where required to cover only the root grid. At least for cubic grids and other grids that have a number of root blocks and zones that are a large enough power of two, the coarsening will continue to negative grid levels and possibly to a block with a single zone. 

[...]

PR #1098, opened 2024-06-06
Bugfix: Resymmetrize logical coordinates

lroberts36:
## PR Summary

Originally, Athena++ transformed integer logical coordinates into floating point logical coordinates in the range [-0.5, 0.5]. In #911, I changed this to map to the range [0, 1]. As @pgrete has pointed out, the new range breaks floating point symmetry (see the comment at the end of this athena++ [document](https://github.com/PrincetonUniversity/athena/wiki/Coordinate-Systems-and-Meshes#tips-for-athena-developers) at the very bottom).  This in turn causes static mesh refinement to operate differently due to roundoff error and block coordinates to not be bitwise exact with the original coordinate positions for single tree forests. This PR switches back to symmetrized floating-point logical coordinates and back to the exact algorithm used for finding static refinement regions from before #911.  

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [ ] Docs build


pdmullen (approved):
Round-off error strikes again! Happy to approve assuming that it fixes @pgrete's issue.  Also, I'd like to understand why this only affects SMR? Shouldn't this have affected symmetry more generally?

lroberts36:
It also effects the coordinates of blocks at roundoff, but I don't think this is as big of an issue as the impact on SMR (where the refinement structure was different in some cases with the exact same input). That being said, it probably broke exact symmetry (if it ever existed in any downstream code).

Yurlungur (approved):
LGTM. I also like the new names better.

src/mesh/mesh.cpp
```diff
@@ -1065,22 +1065,26 @@ void Mesh::DoStaticRefinement(ParameterInput *pin) {
...
+    int lxmin, lxmax;
+    for (lxmin = 0; lxmin < lxtot; lxmin++) {
+      Real r = LogicalLocation::IndexToSymmetrizedCoordinate(lxmin + 1,
+                                                             BlockLocation::Left, lxtot);
+      if (mesh_size.SymmetrizedLogicalToActualPosition(r, dir) > ref_size.xmin(dir))
+        break;
+    }
+    for (lxmax = lxmin; lxmax < lxtot; lxmax++) {
+      Real r = LogicalLocation::IndexToSymmetrizedCoordinate(lxmax + 1,
+                                                             BlockLocation::Left, lxtot);
```
pgrete, line 1081: Just double checking, this is not meant to be `BlockLocation::Right` is it? (the new logical location machinery is still a little opaque to me).

pgrete (approved):
LGTM and I can confirm that this results in the expected mesh again (though other issue still persist with restarts, but I'll track those down separately).

I suggest to wait clicking merge on this PR until #1055 is in `develop`, then change the target branch and pull the trigger here.
This way we get a clean history (which may at some point be useful for bisects)

src/mesh/mesh.cpp
lroberts36: It is supposed to be `BlockLocation::Left` here. I agree this is a little confusing, but this gives code that is as close as possible to the old implementation.

PR #1099, opened 2024-06-07
Add capability to output task graph

### LINKED ISSUE #880: Flowchart for tasks
At the Athena meeting (and also other discussions) it became clear that having a flow chart of the tasks is very useful.
In principle this should be very straightforward by just writing out our tasks in a given format that is understood by a plotting library, e.g., https://mermaid.js.org/syntax/flowchart.html

While giving it a first shot, I noticed two difficulties
1. I don't fully understand how dependencies work internally (in terms of blocks of bitsets with fixed width and whether a new id [without a task] is created for combined dependencies)
2. there's no straightforward way to get the name of the function we pass to `AddTask`. Even with C++20 `std::source_location` we would only be able to determine where it was called, but not the name itself

Any thoughts/ideas/input, particularly @jdolence @Yurlungur @lroberts36 
Am I missing sth?

One potentially clean option is to add another `AddTask` overload that takes an optional string as (arbitrary) function name, so that downstream codes could decide to make use of it (or not).

lroberts36:
## PR Summary
This PR adds the functionality to output the graph of a `TaskList`, `TaskRegion`, or `TaskCollection` in the Graphviz format. I do not explicitly add the connectivity between `TaskRegion`s, so they show up as disconnected graphs. It is pretty trivial to pull out the graph because of #917.  Initially, I used the `nameof` library for writing out function names, but I am not actually convinced this is necessary. I think we get everything we want with
```c++
#define TF(...) std::string(#__VA_ARGS__), __VA_ARGS__
```
Let me know if you have thoughts on this. After reading the internet for a while, I am pretty sure that there is no way to get the names of tasks within task lists themselves since they are effectively function pointers and no name information is available (hence the need to allow passing a string naming the task).  

Results look like (from the sparse advection test without refinement and a single block): 
Screen Shot 2024-06-07 at 1 51 18 PM

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [ ] Docs build


example/sparse_advection/sparse_advection_driver.cpp
```diff
@@ -138,20 +138,20 @@ TaskCollection SparseAdvectionDriver::MakeTaskCollection(BlockList_t &blocks,
...
       // Update refinement
       if (pmesh->adaptive) {
         auto tag_refine = tl.AddTask(
-            set_bc, parthenon::Refinement::Tag<MeshBlockData<Real>>, sc1.get());
+            set_bc, TF(parthenon::Refinement::Tag<MeshBlockData<Real>>), sc1.get());
       }
     }
   }
-
+  tc.PrintGraph();
```
Yurlungur: you probably want to only do this once?

pdmullen (approved):
This is really cool and a powerful debugging tool.  My approval comes from testing this utility in various downstream applications.  All looks good to me.  It even appears to be working with the iterative tasking infrastructure.  And it has already uncovered various issues that needed addressing! 😬 

My only comment is a non-blocker but for a future PR:  It would be cool to thread these graphs through outputs so that they could be dumped to file, e.g., at the beginning of a simulation.  In the meantime, however, as a debugging tool, calling `std::cout << my_task_collection` should be sufficient.

src/tasks/tasks.hpp
```diff
@@ -207,17 +262,30 @@ class TaskList {
...
+  TaskID AddTask(const TaskQualifier tq, TaskID dep, Arg1 &&arg1, Args &&...args) {
+    if constexpr (std::is_invocable<Arg1, Args...>::value) {
+      return AddTaskImpl(tq, dep, {}, std::forward<Arg1>(arg1),
+                         std::forward<Args>(args)...);
+    } else if constexpr (std::is_convertible<Arg1, std::string>::value) {
+      return AddTaskImpl(tq, dep, std::forward<Arg1>(arg1), std::forward<Args>(args)...);
+    } else {
+      PARTHENON_FAIL("Bad signature for AddTask.");
+    }
+  }
```
Yurlungur: I'm not going to make you rewrite it. But his seems much less readable, IMO, than just two overloads.

src/tasks/tasks.hpp
```diff
@@ -207,17 +262,30 @@ class TaskList {
...
 
+  template <class Arg1, class... Args>
+  TaskID AddTask(const TaskQualifier tq, TaskID dep, Arg1 &&arg1, Args &&...args) {
+    if constexpr (std::is_invocable<Arg1, Args...>::value) {
+      return AddTaskImpl(tq, dep, {}, std::forward<Arg1>(arg1),
+                         std::forward<Args>(args)...);
+    } else if constexpr (std::is_convertible<Arg1, std::string>::value) {
+      return AddTaskImpl(tq, dep, std::forward<Arg1>(arg1), std::forward<Args>(args)...);
+    } else {
+      PARTHENON_FAIL("Bad signature for AddTask.");
```
Yurlungur: Can this be
 ```suggestion
      static_assert(false, "Bad signature for AddTask.");
```
or something similar? Since this is `constexpr` it should be possible to make this thing fail at compile time.

You could maybe `static_assert` for the two conditions in the if branches before branching?

File: src/tasks/tasks.hpp [hunk 4]
```diff
@@ -363,29 +446,48 @@ class TaskList {
...
 
+  template <class F>
+  std::string MakeUserTaskLabel(std::optional<std::string> label) {
+    if (!label.has_value()) label = "anon";
+#ifdef HAS_CXX_ABI
+    int status;
+    auto signature =
+        std::string(abi::__cxa_demangle(typeid(F).name(), NULL, NULL, &status));
+#else
+    std::string signature = " (...)";
```
Yurlungur, line 473: should we warn if CXX_ABI is not available so the user knows the graph won't look right?

Yurlungur (approved):
LGTM. Looks like now the user has to call WriteTaskGraph by hand? That's probably for the best.

If I interpret this correctly, the task graph appending is optional--and this doesn't break downstream?

Also could we support
```
PARTHENON_ADDTASK(tasklist, dependency, _VA_ARGS_)
```
as sugar for the call around `TF`?

example/sparse_advection/sparse_advection_driver.cpp
lroberts36: Ah, I see. This comment is old, I refactored things so this method (`PrintGraph()`) doesn't exist anymore.

src/tasks/tasks.hpp
lroberts36: It actually requires a bunch of overloads and gross template arguments to work the other way. This turns out to be much cleaner.

lroberts36: ask me how I know...

Yurlungur: Hah. Ok fair enough.

example/sparse_advection/sparse_advection_driver.cpp
Yurlungur: Oops yes sorry. I started reading this before you were ready for merge.

src/tasks/tasks.hpp
lroberts36: Yeah, it is possible but you run into the `static_assert(false)` makes compilation fail no matter which branch is taken. Need to do something like `static_assert(alway_false_t<Arg1>::value)`. I will switch it, I was just being lazy when I wrote this.

src/tasks/tasks.hpp
lroberts36: Yeah, we can add a warning.

lroberts36:
> LGTM. Looks like now the user has to call WriteTaskGraph by hand? That's probably for the best.

Yeah, currently the user has to do something like `std::cout << tc;` to write the graph to screen. Patrick suggested having an option in the driver that would write the task graph to file on the first step. 
 
> If I interpret this correctly, the task graph appending is optional--and this doesn't break downstream?

Yes, this should have no impact on downstream codes. Although there is a label attached to all tasks now, it has no impact unless you call the output stream operator on a `TaskCollection`, `TaskRegion`, or `TaskList`. If you do call the output stream operator before completely building the task list, it is possible that it could break things since the output operator calls the graph completion task (which is required so that all the edges are correctly set before printing). 

The vector of tasks that gets appended to for output of the graph is a temporary and has no impact on the internal task lists of `Task*`. Aside from the issue of calling the output operator to early and finalizing the graph to early, you should be able to write the task graph to file and then execute the task graph with no side effects.
 
> 
> Also could we support
> 
> ```
> PARTHENON_ADDTASK(tasklist, dependency, _VA_ARGS_)
> ```
> 
> as sugar for the call around `TF`?

I don't quite see what this would buy us?

Yurlungur:
> Yeah, currently the user has to do something like `std::cout << tc;` to write the graph to screen. Patrick suggested having an option in the driver that would write the task graph to file on the first step.

:+1: would be kind of nice to have the driver do it automatically.

> Yes, this should have no impact on downstream codes. Although there is a label attached to all tasks now, it has no impact unless you call the output stream operator on a `TaskCollection`, `TaskRegion`, or `TaskList`. If you do call the output stream operator before completely building the task list, it is possible that it could break things since the output operator calls the graph completion task (which is required so that all the edges are correctly set before printing).

Not super important... but can we add guard rails for that?

> The vector of tasks that gets appended to for output of the graph is a temporary and has no impact on the internal task lists of `Task*`. Aside from the issue of calling the output operator to early and finalizing the graph to early, you should be able to write the task graph to file and then execute the task graph with no side effects.

:+1:

> I don't quite see what this would buy us?

I just find `TF` a little ugly and I like that macro better. But it's a taste thing. No need to add it.

lroberts36:
> > Yeah, currently the user has to do something like `std::cout << tc;` to write the graph to screen. Patrick suggested having an option in the driver that would write the task graph to file on the first step.
> 
> 👍 would be kind of nice to have the driver do it automatically.

Yeah, agreed, but let's leave for a future PR.
 
> > Yes, this should have no impact on downstream codes. Although there is a label attached to all tasks now, it has no impact unless you call the output stream operator on a `TaskCollection`, `TaskRegion`, or `TaskList`. If you do call the output stream operator before completely building the task list, it is possible that it could break things since the output operator calls the graph completion task (which is required so that all the edges are correctly set before printing).
> 
> Not super important... but can we add guard rails for that?

I just included a check in `AddTask` that the task region containing it hasn't had `BuildGraph` called.

> 
> I just find `TF` a little ugly and I like that macro better. But it's a taste thing. No need to add it.

Josh and I thought this sort of pattern was a quick to implement, minimally invasive way to update old task list building code, but I agree that it can be a little obscure (which partially inspired the task function acronym choice...).

Yurlungur:
> Yeah, agreed, but let's leave for a future PR.

:+1:

> I just included a check in `AddTask` that the task region containing it hasn't had `BuildGraph` called.

:+1: 

> Josh and I thought this sort of pattern was a quick to implement, minimally invasive way to update old task list building code, but I agree that it can be a little obscure (which partially inspired the task function acronym choice...).

Yeah. That's fine. Like I said it's a taste thing.

PR #1100, opened 2024-06-07
Custom refinement ops propagated to fluxes

brryan:
## PR Summary

Custom refinement operations enrolled in variable metadata were not being propagated to flux fields if `WithFluxes` was active. This PR addresses that.

Note that in this model the refinement operations will need to be able to distinguish between the field and the flux field, presumably by branching based on the `TopologicalElement`. 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/src/interface/state_descriptor.cpp
+++ b/src/interface/state_descriptor.cpp
@@ -281,11 +281,20 @@ bool StateDescriptor::AddFieldImpl(const VarID &vid, const Metadata &m_in,
         mFlags.push_back(Metadata::Edge);
       else if (m.IsSet(Metadata::Edge))
         mFlags.push_back(Metadata::Node);
-      Metadata mf(mFlags, m.Shape());
+
+      Metadata mf;
+      if (m.GetRefinementFunctions().label().size() > 0) {
+        // Propagate custom refinement ops to flux field
+        mf = Metadata(mFlags, m.Shape(), std::vector<std::string>(), std::string(),
+                      m.GetRefinementFunctions());
+      } else {
+        mf = Metadata(mFlags, m.Shape());
+      }
       auto fId = VarID{internal_fluxname + internal_varname_seperator + vid.base_name,
                        vid.sparse_id};
       AddFieldImpl(fId, mf, control_vid);
       m.SetFluxName(fId.label());
+      refinementFuncMaps_.Register(mf, fId.label());
     }
     metadataMap_.insert({vid, m});
     refinementFuncMaps_.Register(m, vid.label());
--- a/src/interface/state_descriptor.cpp
+++ b/src/interface/state_descriptor.cpp
@@ -294,7 +294,6 @@ bool StateDescriptor::AddFieldImpl(const VarID &vid, const Metadata &m_in,
                        vid.sparse_id};
       AddFieldImpl(fId, mf, control_vid);
       m.SetFluxName(fId.label());
-      refinementFuncMaps_.Register(mf, fId.label());
     }
     metadataMap_.insert({vid, m});
     refinementFuncMaps_.Register(m, vid.label());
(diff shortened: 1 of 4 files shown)


src/interface/state_descriptor.cpp
```diff
@@ -281,11 +281,20 @@ bool StateDescriptor::AddFieldImpl(const VarID &vid, const Metadata &m_in,
...
+        mf = Metadata(mFlags, m.Shape(), std::vector<std::string>(), std::string(),
+                      m.GetRefinementFunctions());
+      } else {
+        mf = Metadata(mFlags, m.Shape());
+      }
       auto fId = VarID{internal_fluxname + internal_varname_seperator + vid.base_name,
                        vid.sparse_id};
       AddFieldImpl(fId, mf, control_vid);
       m.SetFluxName(fId.label());
+      refinementFuncMaps_.Register(mf, fId.label());
```
pdmullen: I don't think you need this, since `AddFieldImpl(fId, mf, control_vid);` will call `RefinementFuncMaps_.Register(m, vid.label());` below, where `vid` is actually `fid` given the function arguments above.

Yurlungur (approved):
Nice. Glad this worked. Do we want to propogate the exact same refinement functions? Or do we want to create a second set of refinement functions for fluxes? I guess the topological element is always different, so they're always distinguishable?

src/interface/state_descriptor.cpp
brryan: I think you are right this isn't needed -- removed.

brryan:
> Nice. Glad this worked. Do we want to propogate the exact same refinement functions? Or do we want to create a second set of refinement functions for fluxes? I guess the topological element is always different, so they're always distinguishable?

Yeah the approach we're taking is to have fluxes get the same refinement functions. As you say they are always distinguishable by topological element, and that is a template parameter, so I don't think there is any cost to doing things anyway, and it is compact. If this seems like a bad design choice though lets discuss that here before merge.

pdmullen:
This PR fixes a breaking change for downstream codes needing custom flux correction routines, therefore, we should likely get it in sooner rather than later...

One comment for future consideration... If we have a field that is NOT `Metadata::Independent`, NOT `Metadata::ForceRemeshComm` but it does have `Metadata::WithFluxes`, it seems strange that we have to enroll all custom PR ops for that field even though the only thing it needs is flux correction. 

Actually, is `Metadata::WithFluxes` enough to trigger `IsRefined` so that the flux correction is even enrolled?

pdmullen:
> Actually, is Metadata::WithFluxes enough to trigger IsRefined so that the flux correction is even enrolled?

the answer is yes: https://github.com/parthenon-hpc-lab/parthenon/blob/ff6a9431a5fd8259aabf0c03a62796774c02eb15/src/interface/metadata.hpp#L451

lroberts36 (approved):
LGTM. I agree that this is a general solution since the restriction operation can choose the method based on the topological element, although in some cases it might be cleaner to allow specifying a separate restriction operation for the flux field.

Yurlungur:
> > Nice. Glad this worked. Do we want to propogate the exact same refinement functions? Or do we want to create a second set of refinement functions for fluxes? I guess the topological element is always different, so they're always distinguishable?
> 
> Yeah the approach we're taking is to have fluxes get the same refinement functions. As you say they are always distinguishable by topological element, and that is a template parameter, so I don't think there is any cost to doing things anyway, and it is compact. If this seems like a bad design choice though lets discuss that here before merge.

Nah it's fine. I think as far as user intent goes, maybe a separate function would be clearer. But as far as implementation "cleanliness" it's unclear what the optimal choice is. There's trade-offs. Go ahead and merge this.

PR #1104, opened 2024-06-12
Fix reading restarts due to hidden ghost var

pgrete:
## PR Summary

`hasGhost` was defined both in `restart.hpp` and `restart_hdf5.hpp`
We only set the value inside the constructor of the derived class but in the parthenon manager we work with the base class.
The latter actually set the data based on bounds defined by `hasGhosts` so it made a decision based on uninitialized data.
On Frontier I could reliably reproduce the issue (both with cray and amdclang/hip compilers), whereas on other machines/compilers `hasGhosts` was set by default/accident to `0` in the base class (though also rendering reading restarts with ghosts unusable in current develop).
Instead of just removing the duplicated declaration I decided to remove public variable entirely and go with a getter function.
I also get the data type to `int` for backwards compatibility and in case we eventually decide to do sth more with this rather than a bool check.

PS: With this (hopefully) last PR, we can finally restart/do some analysis with AthenaPK in sync with Parthenon `develop` again.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/src/outputs/restart_hdf5.cpp
+++ b/src/outputs/restart_hdf5.cpp
@@ -53,7 +53,7 @@ RestartReaderHDF5::RestartReaderHDF5(const char *filename) : filename_(filename)
   fh_ = H5F::FromHIDCheck(H5Fopen(filename, H5F_ACC_RDONLY, H5P_DEFAULT));
   params_group_ = H5G::FromHIDCheck(H5Oopen(fh_, "Params", H5P_DEFAULT));
 
-  hasGhost = GetAttr<int>("Info", "IncludesGhost");
+  has_ghost = GetAttr<int>("Info", "IncludesGhost");
 #endif // ENABLE_HDF5
 }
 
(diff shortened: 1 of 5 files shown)


src/outputs/restart_hdf5.cpp
```diff
@@ -245,8 +245,6 @@ void RestartReaderHDF5::ReadBlocks(const std::string &name, IndexRange range,
                            "Buffer (size " + std::to_string(dataVec.size()) +
                                ") is too small for dataset " + name + " (size " +
                                std::to_string(total_count) + ")");
-  PARTHENON_HDF5_CHECK(
-      H5Sselect_hyperslab(hdl.dataspace, H5S_SELECT_SET, offset, NULL, count, NULL));
```
pgrete, line 249: @lroberts36 found that this line is duplicated two lines below, so I cleaned it up (independent of the bugfix in this PR)

BenWibking:
I don't understand how the CI passed with this bug...?

Also: would it make sense to add an action to run the CI with ASAN enabled to catch unitialized value bugs in general?

pdmullen (approved):
You are a godsend @pgrete.  I lost my day to this bug yesterday... I couldn't track it down for the life of me.  Confirmed that this fixes a restart issue encountered with another parth downstream code.

lroberts36 (approved):
LGTM, nice catch!

lroberts36:
> I don't understand how the CI passed with this bug...?

I think that `hasGhost = 0` is the correct value for all of the restart tests, so if it was default initialized to zero then things would work even with the bug.

> Also: would it make sense to add an action to run the CI with ASAN enabled to catch unitialized value bugs in general?

I agree that this would be a good idea.

PR #1106, opened 2024-06-13
Add CMake options for turning on ASAN and HWASAN

BenWibking:
## PR Summary

Add CMake options `ENABLE_ASAN` and `ENABLE_HWASAN` to build with AddressSanitizer and HWAddressSanitizer, respectively, turned on.

ASAN is supported in both [clang](https://clang.llvm.org/docs/AddressSanitizer.html) and [gcc](https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#index-fsanitize_003daddress). HWASAN is support by [clang](https://clang.llvm.org/docs/HardwareAssistedAddressSanitizerDesign.html) and (in theory) [gcc](https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#index-fsanitize_003dhwaddress).

Then the app is run as normal. It will stop and print an error if a memory bug is detected. This should detect almost all out of bounds access, uninitalized values, and other memory errors for CPU builds.

Usage notes:

* Miscellaneous useful env vars to configure its behavior when running:
  ```
  ASAN_OPTIONS=abort_on_error=1:fast_unwind_on_malloc=1:detect_leaks=0 UBSAN_OPTIONS=print_stacktrace=0 LSAN_OPTIONS=suppressions=leak_suppress.txt
  ```

* HWAddressSanitizer only works on Linux ARM platforms but runs significantly faster than AddressSanitizer. In my testing, additional flags are required for HWAsan:
  ```
  -DCMAKE_CXX_FLAGS="-fPIC" -DCMAKE_EXE_LINKER_FLAGS="-fuse-ld=lld" -DCMAKE_SHARED_LINKER_FLAGS="-fuse-ld=lld" -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++
  ```
  (I was not able to get it working with gcc, since building HWASAN support in gcc is optional and distros don't do it.)

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,6 +57,8 @@ option(CHECK_REGISTRY_PRESSURE "Check the registry pressure for Kokkos CUDA kern
 option(TEST_INTEL_OPTIMIZATION "Test intel optimization and vectorization" OFF)
 option(TEST_ERROR_CHECKING "Enables the error checking unit test. This test will FAIL" OFF)
 option(CODE_COVERAGE "Enable code coverage reporting" OFF)
+option(ENABLE_ASAN "Turn on ASAN" OFF)
+option(ENABLE_HWASAN "Turn on HWASAN (currently ARM-only)" OFF)
 
 include(cmake/Format.cmake)
 include(cmake/Lint.cmake)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -290,7 +292,19 @@ if (Kokkos_ENABLE_CUDA AND "${PARTHENON_ENABLE_GPU_MPI_CHECKS}" )
   configure_file(${CMAKE_CURRENT_SOURCE_DIR}/cmake/CTestCustom.cmake.in ${CMAKE_BINARY_DIR}/CTestCustom.cmake @ONLY)
 endif()
 
-
+# option to turn on AddressSanitizer for debugging
+if(ENABLE_ASAN)
+  message(STATUS "Compiling with AddressSanitizer and UndefinedBehaviorSanitizer *enabled*")
+  add_compile_options(-fsanitize=address -fsanitize=undefined -fno-sanitize-recover=all -fsanitize=float-divide-by-zero -fsanitize=float-cast-overflow -fno-sanitize=null -fno-sanitize=alignment)
+  add_link_options(-fsanitize=address -fsanitize=undefined)
```
pgrete, line 299: This will then also be forwarded to all submodules, we include, isn't it?

pgrete (approved):
I think it's a good idea to add those.
How much slower does the code run with those options?
I'm asking wrt to whether it's worth to activate this by default to one of the short or extended tests (once the issue reported in #1107 are fixed).

PS: Changelog entry is missing;)

CMakeLists.txt
BenWibking: Yes, it should be.

lroberts36 (approved):
LGTM, this will definitely be helpful to have. I am going to take a look at the logical location bit shift issue, I think it will be straightforward to fix.

lroberts36:
> I'm asking wrt to whether it's worth to activate this by default to one of the short or extended tests (once the issue reported in #1107 are fixed).

I definitely support having it run on one of the extended tests (unless it takes hours to run with these options on).

BenWibking:
Running the test suite with this turned on is only about a factor of 2-3 slower on my laptop, so I think it should be fine to enable for the extended CI.

PR #1113, opened 2024-06-13
Small Bugfix: Prevent division by zero

lroberts36:
## PR Summary

Remove divisions by zero that are described in #1107. 

## PR Checklist

- [ ] Code passes cpplint
- [ ] New features are documented.
- [ ] Adds a test for any bugs fixed. Adds tests for new features.
- [ ] Code is formatted
- [ ] Changes are summarized in CHANGELOG.md
- [ ] Change is breaking (API, behavior, ...)
  - [ ] Change is *additionally* added to CHANGELOG.md in the breaking section
  - [ ] PR is marked as breaking
  - [ ] Short summary API changes at the top of the PR (plus optionally with an automated update/fix script)
- [ ] CI has been triggered on [Darwin](https://re-git.lanl.gov/eap-oss/parthenon/-/pipelines) for performance regression tests.
- [ ] Docs build
- [ ] (@lanl.gov employees) Update copyright on changed files

PR #1111, opened 2024-06-13
Small Bugfix: Remove possibility of bitshifting negative integers

lroberts36:
## PR Summary

This PR fixes the bitshift related undefined behavior described in #1107 by multiplying by a power of two explicitly rather than performing the same operation (on positive numbers) via a bitshift.

## PR Checklist

- [ ] Code passes cpplint
- [ ] New features are documented.
- [ ] Adds a test for any bugs fixed. Adds tests for new features.
- [ ] Code is formatted
- [ ] Changes are summarized in CHANGELOG.md
- [ ] Change is breaking (API, behavior, ...)
  - [ ] Change is *additionally* added to CHANGELOG.md in the breaking section
  - [ ] PR is marked as breaking
  - [ ] Short summary API changes at the top of the PR (plus optionally with an automated update/fix script)
- [ ] CI has been triggered on [Darwin](https://re-git.lanl.gov/eap-oss/parthenon/-/pipelines) for performance regression tests.
- [ ] Docs build
- [ ] (@lanl.gov employees) Update copyright on changed files

PR #1108, opened 2024-06-13
[Breaking change] Remove NaN payload tags infrastructure

BenWibking:
## PR Summary

This removes the `NaN` payload tags infrastructure.

It is unused, and the unit test causes a runtime error with UndefinedBehaviorSanitizer. See https://github.com/parthenon-hpc-lab/parthenon/issues/1107.

## PR Checklist

- [ ] Code passes cpplint
- [ ] New features are documented.
- [ ] Adds a test for any bugs fixed. Adds tests for new features.
- [ ] Code is formatted
- [x] Changes are summarized in CHANGELOG.md
- [x] Change is breaking (API, behavior, ...)
  - [x] Change is *additionally* added to CHANGELOG.md in the breaking section
  - [x] PR is marked as breaking
  - [ ] Short summary API changes at the top of the PR (plus optionally with an automated update/fix script)
- [ ] CI has been triggered on [Darwin](https://re-git.lanl.gov/eap-oss/parthenon/-/pipelines) for performance regression tests.
- [ ] Docs build
- [ ] (@lanl.gov employees) Update copyright on changed files

PR #1114, opened 2024-06-14
Enable sanitizers for extended CI host build

BenWibking:
## PR Summary

Enable sanitizers (e.g., AddressSanitizer, UndefinedBehaviorSanitizer, etc.) for the extended CI host build. It is not enabled for the CUDA build, since CUDA does not support sanitizers.

Memory leak detection is currently disabled (does not work with OpenMPI, several unit tests have memory leaks).

**Depends on:**
* https://github.com/parthenon-hpc-lab/parthenon/pull/1113
* https://github.com/parthenon-hpc-lab/parthenon/pull/1111
* https://github.com/parthenon-hpc-lab/parthenon/pull/1108
* https://github.com/parthenon-hpc-lab/parthenon/pull/1112

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build


BenWibking:
Tests are expected to fail until the issues are fixed in other PRs: https://github.com/parthenon-hpc-lab/parthenon/actions/runs/9522754529/job/26252926684

Yurlungur (approved):
Will be nice to have this automatically. Hopefully will catch bugs.

lroberts36 (approved):
LGTM, I agree that this will help us catch some subtler issues before they can enter `develop`.

.github/workflows/ci-extended.yml
```diff
@@ -44,10 +44,20 @@ jobs:
...
+            echo "enable_asan=ON" >> $GITHUB_OUTPUT
+          else
+            echo "enable_asan=OFF" >> $GITHUB_OUTPUT
+          fi
+
       - name: Configure
         run: |
           cmake -B build \
             -DCMAKE_BUILD_TYPE=Release \
+            -DENABLE_ASAN=${{ steps.cmake-vars.outputs.enable_asan }} \
```
pgrete, line 60: Is that a common pattern?

Might sth like
```bash
-DENABLE_ASAN=${${{ matrix.device == 'host' }}^^} \
```
work, too?

Or if that one line logic doesn't work using an env var:
```
env:
  ENABLE_ASAN: ${{ matrix.device == 'host' }}
```
with (the `^^` to uppercase the lowercase result in the env var):
```
-DENABLE_ASAN=${ENABLE_ASAN^^} \
```

BenWibking: Bash outputs `1` for true and `0` for false, so I don't think that works, unless CMake translates `1` to `ON`?

pgrete: I think it should 
"True if the constant is 1, ON, YES, TRUE, Y, or a non-zero number (including floating point numbers)." from https://cmake.org/cmake/help/latest/command/if.html

But I think GitHub actions only returns (lowercase) `true` or `false` for expressions so that's why we need to convert it to uppercase.

BenWibking:
Probably need to disable LeakSanitizer for MPI runs due to the memory leaks in OpenMPI we can't do anything about...

BenWibking:
AddressSanitizer and UndefinedBehaviorSanitizer run without workings, but some tests have memory leaks that cause LeakDetector to fail.

BenWibking:
Rather than try to chase down every memory leak in the unit tests, I've just disabled memory leak detection entirely.

Yurlungur:
> Rather than try to chase down every memory leak in the unit tests, I've just disabled memory leak detection entirely.

I think that's the right thing but we should leave the memory leak test in place in the cmake so we can trigger it locally. Because someone should start trying to track these down over time. :)

PR #1115, opened 2024-06-15
[WIP] Take ownership of `Param`s

BenWibking:
## PR Summary

When `Param`s are created or updated, it takes ownership of the object passed with `std::move`.

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build

### DIFF
--- a/src/interface/params.hpp
+++ b/src/interface/params.hpp
@@ -49,28 +49,28 @@ class Params {
   ///
   /// Throws an error if the key is already in use
   template <typename T>
-  void Add(const std::string &key, T &value, Mutability mutability) {
+  void Add(const std::string &key, T &&object, Mutability mutability) {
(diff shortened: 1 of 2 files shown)


BenWibking:
@Yurlungur is there a reason for a custom Params container? Could this be replaced with a wrapper around a `std::unordered_map<std::string, std::any>`?

Although I see that `std::any` only works for copy-constructible objects...

Yurlungur:
Thanks for taking a stab at this!

> @Yurlungur is there a reason for a custom Params container? Could this be replaced with a wrapper around a `std::unordered_map<std::string, std::any>`?

We wrote `Params` when Parthenon was C++14 only. So at the time it wasn't an option. In principle, I think it could be replaced with `std::any` modulo the issue with copy constructible types.

I am somewhat concerned also about how this will interact with I/O. In principle it should all work though.

PR #1119, opened 2024-06-18
Refactor `MeshBlock` Partitioning

### LINKED ISSUE #1097: Formalizing `MeshBlock` partitions for creating `MeshData`
I am putting up this issue as the result of a discussion we had on the 6/6/2024 Parthenon sync call.

In #1092, I cleaned up `DataCollection` a little bit so that `MeshBlockData` did not need to be pre-built by hand before creating `MeshData` objects with `GetOrAdd`. One thing that I changed was that `DataCollection<MeshBlockData>::Add` can now take create new `MeshBlockData` in the collection from either another `MeshBlockData` (as before) or from a `MeshBlock`. It was not possible to do something similar for call to `DataCollection<MeshData>::Add`, since there is no analogy to `MeshBlock` for `MeshData`, i.e. `MeshBlockData` is to `MeshBlock` as `MeshData` is to `??????`. 

Additionally, all block list partitioning is dealt with in `DataCollection<MeshData>::GetOrAdd_impl`. At least to me, this is a little bit of a strange place to have it conceptually. Also, as it is currently written the only partitioning allowed is to the `Mesh::DefaultPackSize()`.

As a result, it seems like it would be a good thing to create an object `BlockPartition` (or some better name). A set of these could live in `Mesh` and `DataCollection<MeshData>::Add` would work on them as well. It seems like it would be straightforward to address most of the issues brought up in #1003, #959, and #419.

lroberts36:
## PR Summary
This PR adds an object `BlockListPartition` that stores a `BlockList_t` and a little more information about a partition of blocks. We then have the analogy `BlockListPartition` is to `MeshData` as `MeshBlock` is to `MeshBlockData`. Then instead of doing block list partitioning inside of `DataCollection`, partitioning is moved to `Mesh` (which at least to me is a more reasonable location for this to live) and methods for getting the partitions are added to `Mesh`. `MeshData` gains an `Initialize` method that takes a `BlockListPartition`. As a result, all incarnations of `DataCollection<MeshData>::Add` are made to "just work" for `MeshData` when it is passed either another `MeshData` object or a `BlockListPartition` object. `DataCollection<MeshData>::GetOrAdd` is deprecated (although still available, it just calls `Add` internally). I have also removed `GetOrAdd` from Parthenon internals and from the examples.

This nearly gets us to working multigrid for multiple partitions, aside from the task bug described in #1125 (at least I hope).

This should close issue #1097.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [ ] Docs build

### DIFF
--- a/example/advection/advection_driver.cpp
+++ b/example/advection/advection_driver.cpp
@@ -62,7 +62,7 @@ TaskCollection AdvectionDriver::MakeTaskCollection(BlockList_t &blocks, const in
   const Real dt = integrator->dt;
   const auto &stage_name = integrator->stage_name;
 
-  auto partitions = pmesh->GetBlockPartitions();
+  auto partitions = pmesh->GetDefaultBlockPartitions();
   int num_partitions = partitions.size();
   // note that task within this region that contains one tasklist per pack
   // could still be executed in parallel
--- a/example/advection/advection_driver.cpp
+++ b/example/advection/advection_driver.cpp
@@ -62,27 +62,21 @@ TaskCollection AdvectionDriver::MakeTaskCollection(BlockList_t &blocks, const in
   const Real dt = integrator->dt;
   const auto &stage_name = integrator->stage_name;
 
-  // first make other useful containers
-  if (stage == 1) {
-    for (int i = 0; i < blocks.size(); i++) {
-      auto &pmb = blocks[i];
-      // first make other useful containers
-      auto &base = pmb->meshblock_data.Get();
-      pmb->meshblock_data.Add("dUdt", base);
-      for (int s = 1; s < integrator->nstages; s++)
-        pmb->meshblock_data.Add(stage_name[s], base);
-    }
-  }
-
-  const int num_partitions = pmesh->DefaultNumPartitions();
-
+  auto partitions = pmesh->GetBlockPartitions();
+  int num_partitions = partitions.size();
   // note that task within this region that contains one tasklist per pack
   // could still be executed in parallel
   TaskRegion &single_tasklist_per_pack_region2 = tc.AddRegion(num_partitions);
   for (int i = 0; i < num_partitions; i++) {
     auto &tl = single_tasklist_per_pack_region2[i];
-    auto &mc0 = pmesh->mesh_data.GetOrAdd(stage_name[stage - 1], i);
-    auto &mc1 = pmesh->mesh_data.GetOrAdd(stage_name[stage], i);
+    // Initialize the base MeshData for this partition
+    // (this automatically initializes the MeshBlockData objects
+    // required by this MeshData object)
+    auto &mbase = pmesh->mesh_data.Add("base", partitions[i]);
+    // Initialize other MeshData objects based on the base container
+    auto &mc0 = pmesh->mesh_data.Add(stage_name[stage - 1], mbase);
+    auto &mc1 = pmesh->mesh_data.Add(stage_name[stage], mbase);
+    auto &mdudt = pmesh->mesh_data.Add("dUdt", mbase);
 
     const auto any = parthenon::BoundaryType::any;
 
@@ -119,10 +113,10 @@ TaskCollection AdvectionDriver::MakeTaskCollection(BlockList_t &blocks, const in
   TaskRegion &single_tasklist_per_pack_region = tc.AddRegion(num_partitions);
   for (int i = 0; i < num_partitions; i++) {
     auto &tl = single_tasklist_per_pack_region[i];
-    auto &mbase = pmesh->mesh_data.GetOrAdd("base", i);
-    auto &mc0 = pmesh->mesh_data.GetOrAdd(stage_name[stage - 1], i);
-    auto &mc1 = pmesh->mesh_data.GetOrAdd(stage_name[stage], i);
-    auto &mdudt = pmesh->mesh_data.GetOrAdd("dUdt", i);
+    auto &mbase = pmesh->mesh_data.Add("base", partitions[i]);
+    auto &mc0 = pmesh->mesh_data.Add(stage_name[stage - 1], mbase);
+    auto &mc1 = pmesh->mesh_data.Add(stage_name[stage], mbase);
+    auto &mdudt = pmesh->mesh_data.Add("dUdt", mbase);
 
     auto set_flx = parthenon::AddFluxCorrectionTasks(none, tl, mc0, pmesh->multilevel);
 
--- a/src/bvals/comms/bvals_utils.hpp
+++ b/src/bvals/comms/bvals_utils.hpp
@@ -105,11 +105,10 @@ void InitializeBufferCache(std::shared_ptr<MeshData<Real>> &md, COMM_MAP *comm_m
   std::for_each(std::begin(key_order), std::end(key_order), [&](auto &t) {
     if (comm_map->count(std::get<2>(t)) == 0) {
       auto key = std::get<2>(t);
-      PARTHENON_FAIL(std::string("Asking for buffer that doesn't exist")
-                     + " (sender: " + std::to_string(std::get<0>(key)) 
-                     + ", receiver: " + std::to_string(std::get<1>(key))
-                     + ", var: " + std::get<2>(key)
-                     + ", location: " + std::to_string(std::get<3>(key)) + ")");
+      PARTHENON_FAIL(std::string("Asking for buffer that doesn't exist") +
+                     " (sender: " + std::to_string(std::get<0>(key)) + ", receiver: " +
+                     std::to_string(std::get<1>(key)) + ", var: " + std::get<2>(key) +
+                     ", location: " + std::to_string(std::get<3>(key)) + ")");
     }
     pcache->buf_vec.push_back(&((*comm_map)[std::get<2>(t)]));
     (pcache->idx_vec)[std::get<1>(t)] = buff_idx++;
--- a/src/bvals/comms/bvals_utils.hpp
+++ b/src/bvals/comms/bvals_utils.hpp
@@ -104,7 +104,12 @@ void InitializeBufferCache(std::shared_ptr<MeshData<Real>> &md, COMM_MAP *comm_m
   pcache->idx_vec = std::vector<std::size_t>(key_order.size());
   std::for_each(std::begin(key_order), std::end(key_order), [&](auto &t) {
     if (comm_map->count(std::get<2>(t)) == 0) {
-      PARTHENON_FAIL("Asking for buffer that doesn't exist");
+      auto key = std::get<2>(t);
+      PARTHENON_FAIL(std::string("Asking for buffer that doesn't exist")
+                     + " (sender: " + std::to_string(std::get<0>(key)) 
+                     + ", receiver: " + std::to_string(std::get<1>(key))
+                     + ", var: " + std::get<2>(key)
+                     + ", location: " + std::to_string(std::get<3>(key)) + ")");
     }
     pcache->buf_vec.push_back(&((*comm_map)[std::get<2>(t)]));
     (pcache->idx_vec)[std::get<1>(t)] = buff_idx++;
--- a/src/driver/driver.cpp
+++ b/src/driver/driver.cpp
@@ -63,7 +63,7 @@ void Driver::PostExecute(DriverStatus status) {
 
 DriverStatus EvolutionDriver::Execute() {
   PreExecute();
-  InitializeBlockTimeStepsAndBoundaries();
+  InitializeBlockTimeSteps();
   SetGlobalTimeStep();
 
   // Before loop do work
@@ -123,7 +123,7 @@ DriverStatus EvolutionDriver::Execute() {
 
       timer_LBandAMR.reset();
       pmesh->LoadBalancingAndAdaptiveMeshRefinement(pinput, app_input);
-      if (pmesh->modified) InitializeBlockTimeStepsAndBoundaries();
+      if (pmesh->modified) InitializeBlockTimeSteps();
       time_LBandAMR += timer_LBandAMR.seconds();
       SetGlobalTimeStep();
 
@@ -188,23 +188,15 @@ void EvolutionDriver::PostExecute(DriverStatus status) {
   Driver::PostExecute(status);
 }
 
-void EvolutionDriver::InitializeBlockTimeStepsAndBoundaries() {
+void EvolutionDriver::InitializeBlockTimeSteps() {
   // calculate the first time step using Block function
   for (auto &pmb : pmesh->block_list) {
     Update::EstimateTimestep(pmb->meshblock_data.Get().get());
   }
   // calculate the first time step using Mesh function
-  pmesh->boundary_comm_map.clear();
-  const int num_partitions = pmesh->DefaultNumPartitions();
-  for (int i = 0; i < num_partitions; i++) {
-    auto &mbase = pmesh->mesh_data.GetOrAdd("base", i);
-    Update::EstimateTimestep(mbase.get());
-    BuildBoundaryBuffers(mbase);
-    for (auto &[gmg_level, mdc] : pmesh->gmg_mesh_data) {
-      auto &mdg = mdc.GetOrAdd(gmg_level, "base", i);
-      BuildBoundaryBuffers(mdg);
-      BuildGMGBoundaryBuffers(mdg);
-    }
+  for (auto &partition : pmesh->GetBlockPartitions()) {
+    auto &mbase = pmesh->mesh_data.Add("base", partition);
+    Update::EstimateTimestep(mbase.get())
   }
 }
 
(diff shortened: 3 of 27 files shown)


src/driver/driver.cpp
```diff
@@ -188,23 +188,15 @@ void EvolutionDriver::PostExecute(DriverStatus status) {
...
-    auto &mbase = pmesh->mesh_data.GetOrAdd("base", i);
+  for (auto &partition : pmesh->GetBlockPartitions()) {
+    auto &mbase = pmesh->mesh_data.Add("base", partition);
     Update::EstimateTimestep(mbase.get());
-    BuildBoundaryBuffers(mbase);
-    for (auto &[gmg_level, mdc] : pmesh->gmg_mesh_data) {
-      auto &mdg = mdc.GetOrAdd(gmg_level, "base", i);
-      BuildBoundaryBuffers(mdg);
-      BuildGMGBoundaryBuffers(mdg);
-    }
```
lroberts36, line 207: I think this is unnecessary, so I removed it. Let me know if I am missing something.

src/bvals/comms/bvals_utils.hpp
```diff
@@ -104,7 +104,11 @@ void InitializeBufferCache(std::shared_ptr<MeshData<Real>> &md, COMM_MAP *comm_m
   pcache->idx_vec = std::vector<std::size_t>(key_order.size());
   std::for_each(std::begin(key_order), std::end(key_order), [&](auto &t) {
     if (comm_map->count(std::get<2>(t)) == 0) {
-      PARTHENON_FAIL("Asking for buffer that doesn't exist");
+      auto key = std::get<2>(t);
+      PARTHENON_FAIL(std::string("Asking for buffer that doesn't exist") +
+                     " (sender: " + std::to_string(std::get<0>(key)) + ", receiver: " +
+                     std::to_string(std::get<1>(key)) + ", var: " + std::get<2>(key) +
+                     ", location: " + std::to_string(std::get<3>(key)) + ")");
```
Yurlungur, line 111: +1 for verbose error message.

src/driver/driver.cpp
Yurlungur: I think you wrote that code in the first place ;)

example/advection/advection_driver.cpp
```diff
@@ -62,27 +62,21 @@ TaskCollection AdvectionDriver::MakeTaskCollection(BlockList_t &blocks, const in
...
   // could still be executed in parallel
   TaskRegion &single_tasklist_per_pack_region2 = tc.AddRegion(num_partitions);
   for (int i = 0; i < num_partitions; i++) {
     auto &tl = single_tasklist_per_pack_region2[i];
-    auto &mc0 = pmesh->mesh_data.GetOrAdd(stage_name[stage - 1], i);
-    auto &mc1 = pmesh->mesh_data.GetOrAdd(stage_name[stage], i);
+    // Initialize the base MeshData for this partition
+    // (this automatically initializes the MeshBlockData objects
+    // required by this MeshData object)
+    auto &mbase = pmesh->mesh_data.Add("base", partitions[i]);
```
pdmullen, line 75: As mentioned during the parth call, the previous name `GetOrAdd` implied that if already available, we need not generate it again --- `Add` also has this feature as used here, however, I wonder if another name would better convey this?

Edit: Looks like @jonahm-LANL also independently raised this concern --- perhaps a rename is warranted, albeit making this a breaking PR would be not great.  I trust your judgement on how to proceed here.

src/mesh/meshblock.hpp
```diff
@@ -460,6 +460,15 @@ class MeshBlock : public std::enable_shared_from_this<MeshBlock> {
...
 using BlockList_t = std::vector<std::shared_ptr<MeshBlock>>;
 
+struct BlockListPartition {
+  BlockListPartition(int p, GridIdentifier g, const BlockList_t &bl, Mesh *pm)
+      : partition{p}, grid{g}, block_list{bl}, pmesh{pm} {}
+  int partition;
+  GridIdentifier grid;
+  BlockList_t block_list;
+  Mesh *pmesh;
+};
```
Yurlungur: Are the partitions modified after being created? Or can the member fields be `const`?

src/mesh/mesh.hpp
```diff
@@ -329,6 +334,10 @@ class Mesh {
   void CommunicateBoundaries(std::string md_name = "base");
   void PreCommFillDerived();
   void FillDerived();
+
+  void BuildBlockPartitions(GridIdentifier grid);
+  std::map<GridIdentifier, std::vector<std::shared_ptr<BlockListPartition>>>
+      block_partitions_;
```
Yurlungur, line 340: I can't tell from the diff is this private or public scope?

src/mesh/mesh.hpp
```diff
@@ -146,6 +145,12 @@ class Mesh {
   int DefaultNumPartitions() {
     return partition::partition_impl::IntCeil(block_list.size(), DefaultPackSize());
   }
+
+  std::vector<std::shared_ptr<BlockListPartition>> &
+  GetBlockPartitions(GridIdentifier grid = GridIdentifier::leaf()) {
```
Yurlungur: should this be `const`?

src/interface/mesh_data.hpp
```diff
@@ -243,39 +243,41 @@ class MeshData {
     }
   }
 
-  void Initialize(BlockList_t blocks, Mesh *pmesh, int ndim,
-                  std::optional<int> gmg_level = {});
-  void Initialize(BlockList_t blocks, Mesh *pmesh, std::optional<int> gmg_level = {});
+  template <typename ID_t>
+  void Initialize(const std::shared_ptr<BlockListPartition> &part,
+                  const std::vector<ID_t> &vars, const bool shallow) {
```
Yurlungur, line 248: Is it necessary to include `shallow` as an option at all here?

Yurlungur (approved):
Overall I like this. Two big picture questions:
1. Does this now preclude different partitioning for different parts of a task list? Does it also preclude packs over different subsets of blocks (for example different refinement levels)?
2. It's a bit weird that `Get` is sometimes `Get` and sometimes `GetOrAdd`. Also weird that a partition index must always requested.

src/driver/driver.cpp
lroberts36: I changed it and added the GMG stuff, but I did that because the first version of the code I looked at built buffers here (but I could never figure out why it did that).

Yurlungur: :+1:

src/interface/mesh_data.hpp
lroberts36: Yeah, it is since `Data Collection` passes this parameter to `Initialize`.

src/mesh/mesh.hpp
lroberts36: it should be private, but let me double check.

src/interface/data_collection.hpp
```diff
@@ -76,17 +70,21 @@ class DataCollection {
...
+
   template <class SRC_t, typename ID_t = std::string>
   std::shared_ptr<T> &AddShallow(const std::string &label,
                                  const std::shared_ptr<SRC_t> &src,
                                  const std::vector<ID_t> &fields = {}) {
     return Add(label, src, fields, true);
   }
+  // TODO(LFR): Is this method used anywhere? Should insertion of empty containers be
+  // allowed?
   std::shared_ptr<T> &Add(const std::string &label);
```
Yurlungur: Is `base` created this way?

src/interface/mesh_data.hpp
Yurlungur: Ah I see. Ok.

lroberts36:
> Overall I like this. Two big picture questions:
> 
> Does this now preclude different partitioning for different parts of a task list? Does it also preclude packs over different subsets of blocks (for example different refinement levels)?

No, you can build `BlockListPartition`s however you might wish and multiple partitions could contain partially overlapping block lists. This PR provides no functionality for anything beyond what the preceding code did in `DataCollection` for partitioning, it just moves it to `Mesh`. That being said, different functionality could be added to mesh or downstream codes could build `BlockListPartition`s themselves based on whatever criteria they desire. The main change here is that partitioning is decoupled from `DataCollection`.

>  It's a bit weird that `Get` is sometimes `Get` and sometimes `GetOrAdd`. Also weird that a partition index must always requested.

`DataCollection<T>::GetOrAdd` was only supported previously for `T = MeshData` and it took a partition index (since partitioning was done internally in `DataCollection`). This PR deprecates that function, but I did not remove it since it would break just about every code downstream (and the old version is what requires a partition index, which I agree is not very robust). `DataCollection<T>::Get` should just get a container only if it exists. 

What is confusing is that `DataCollection<T>::Add` really gets or adds the requested container (depending on if it is already in the map). I would support changing the name, but that would be a breaking change since `Add` is used for `MeshBlockData` in downstream codes.

Yurlungur:
> No, you can build `BlockListPartition`s however you might wish and multiple partitions could contain partially overlapping block lists. This PR provides no functionality for anything beyond what the preceding code did in `DataCollection` for partitioning, it just moves it to `Mesh`. That being said, different functionality could be added to mesh or downstream codes could build `BlockListPartition`s themselves based on whatever criteria they desire. The main change here is that partitioning is decoupled from `DataCollection`.

:+1:

> What is confusing is that `DataCollection<T>::Add` really gets or adds the requested container (depending on if it is already in the map). I would support changing the name, but that would be a breaking change since `Add` is used for `MeshBlockData` in downstream codes.

Yeah I feel ambivalent about this. I think changing the name would be better but not sure it's worth breaking backwards compatibility. Maybe we table that change for a later PR and poll if people think it's worth doing.

src/interface/data_collection.hpp
lroberts36: If I remove the method, Parthenon itself compiles with no issues. I just wonder if any downstream codes use this. It doesn't really make sense when `T = MeshData` though, since the label is not the key in that case.

I am removing it, since I don't see any use case and it seems like it could result in unexpected behavior if it is used.

src/mesh/mesh.hpp
lroberts36: yeah, it is private

src/mesh/meshblock.hpp
lroberts36: made most of the fields constant, but it didn't work for `pmesh`.

src/mesh/mesh.hpp
lroberts36: yep, it can be. changed.

lroberts36:
> 1. Does this now preclude different partitioning for different parts of a task list? Does it also preclude packs over different subsets of blocks (for example different refinement levels)?

I wasn't thinking about the map key when I responded to this earlier, which did rely on the partition id. I now made the key just contain a list of the gids of all blocks in the partition. This should allow arbitrary partitioning without having to make any changes to `Mesh`.

src/interface/data_collection.hpp
```diff
@@ -76,42 +70,45 @@ class DataCollection {
...
   std::shared_ptr<T> &GetOrAdd(const std::string &mbd_label, const int &partition_id);
   std::shared_ptr<T> &GetOrAdd(int gmg_level, const std::string &mbd_label,
                                const int &partition_id);
 
   void PurgeNonBase() {
     auto c = containers_.begin();
     while (c != containers_.end()) {
-      if (c->first != "base") {
+      // Remove anything that doesn't contain the label base
+      if (c->first.find("base") == std::string::npos) {
```
lroberts36: This change made the tests break...

example/advection/advection_driver.cpp
```diff
@@ -62,27 +62,21 @@ TaskCollection AdvectionDriver::MakeTaskCollection(BlockList_t &blocks, const in
...
-      auto &base = pmb->meshblock_data.Get();
-      pmb->meshblock_data.Add("dUdt", base);
-      for (int s = 1; s < integrator->nstages; s++)
-        pmb->meshblock_data.Add(stage_name[s], base);
-    }
-  }
-
-  const int num_partitions = pmesh->DefaultNumPartitions();
-
+  auto partitions = pmesh->GetDefaultBlockPartitions();
```
pdmullen, line 65: You didn't introduce it, but I am curious what `Default` means in this context? 

When I read it, it makes me assume that it would return `partitions` with `partitions.size() == 1` since `pack_size` defaults to `-1` when not specified by user, but I know that is incorrect.

example/advection/advection_driver.cpp
```diff
@@ -62,27 +62,21 @@ TaskCollection AdvectionDriver::MakeTaskCollection(BlockList_t &blocks, const in
...
+  int num_partitions = partitions.size();
   // note that task within this region that contains one tasklist per pack
   // could still be executed in parallel
   TaskRegion &single_tasklist_per_pack_region2 = tc.AddRegion(num_partitions);
   for (int i = 0; i < num_partitions; i++) {
     auto &tl = single_tasklist_per_pack_region2[i];
-    auto &mc0 = pmesh->mesh_data.GetOrAdd(stage_name[stage - 1], i);
```
pdmullen, line 84: When would a user ever want to use `pmesh->mesh_data.GetOrAdd()` instead of `pmesh->mesh_data.Add("base", partitions[i]);` following this PR?

src/driver/driver.cpp
```diff
@@ -188,23 +188,15 @@ void EvolutionDriver::PostExecute(DriverStatus status) {
...
   }
   // calculate the first time step using Mesh function
-  pmesh->boundary_comm_map.clear();
-  const int num_partitions = pmesh->DefaultNumPartitions();
-  for (int i = 0; i < num_partitions; i++) {
-    auto &mbase = pmesh->mesh_data.GetOrAdd("base", i);
+  for (auto &partition : pmesh->GetDefaultBlockPartitions()) {
+    auto &mbase = pmesh->mesh_data.Add("base", partition);
     Update::EstimateTimestep(mbase.get());
-    BuildBoundaryBuffers(mbase);
```
pdmullen, line 202: this (https://github.com/parthenon-hpc-lab/parthenon/blob/a44da26921b7db024a104ed3f18dd59c35ec57a5/src/mesh/mesh.cpp#L642-L652) alone is satisfactory?

src/mesh/mesh-amr_loadbalance.cpp
```diff
@@ -996,6 +997,7 @@ void Mesh::RedistributeAndRefineMeshBlocks(ParameterInput *pin, ApplicationInput
     FillDerived();
 
     // Initialize the "base" MeshData object
+    // TODO(LFR): Is this necessary? Do we every pull out the entire mesh MeshData?
```
pdmullen: typo

src/mesh/mesh.cpp
```diff
@@ -803,7 +830,7 @@ void Mesh::Initialize(bool init_problem, ParameterInput *pin, ApplicationInput *
                           "Mesh PostInitialization requires parthenon/mesh/pack_size=-1 "
```
pdmullen: Can you make the same change that you did for `ProblemGenerator` here?

pdmullen (approved):
LGTM! Tested downstream (without iterative tasklists, without changing `GetOrAdd` -> `pmesh->Add`) and all looked good.

example/advection/advection_driver.cpp
Yurlungur: default partition size is the size set by the input deck. But the user can build partitions of alternate sizes as they see fit.

pdmullen: Thanks 😸

example/advection/advection_driver.cpp
lroberts36: Yeah, I think I want to leave this as is for now. It would also be easy to add a function `auto GetOrAdd(Args&&...args){return Add(std::forward<Args>(args)...);}` eventually.

example/advection/advection_driver.cpp
lroberts36: possibly useful for sweeps.

example/advection/advection_driver.cpp
lroberts36: never, I think.

src/driver/driver.cpp
lroberts36: Sorry, don't completely understand what you are asking here.

I am 95% sure that building the boundary buffers here is unnecessary (and have been for a while), so I removed it in this PR.

src/mesh/mesh.cpp
lroberts36: yeah, good catch

src/mesh/mesh-amr_loadbalance.cpp
lroberts36: fixed

src/driver/driver.cpp
pdmullen: My linked code to `mesh.cpp` is almost verbatim what you deleted from `driver.cpp`.  My question was just: are we deleting this block of code in `driver.cpp` here because this is actually taken care of here in `mesh.cpp`?

example/advection/advection_driver.cpp
pdmullen: Sounds good to me

src/driver/driver.cpp
lroberts36: ah, I see (didn't read carefully enough). Yeah, I think that code in `mesh.cpp` should do what is needed and I think it was strange to have this buffer building tucked away in the driver as well.

PR #1130, opened 2024-06-24
Enable `parthenon::par_reduce` for MD loops with Kokkos 1D Range

acreyes:
## PR Summary

Replaces `KOKKOS_LAMBDA`s with templated functors in `par_dispatch(LoopPatternFlatRange, ...)` that deduce the signature of the provided `function` to allow for extra arguments used in reductions.

Makes it possible to use `par_reduce(DEFAULT_LOOP_PATTERN, ..)` & `par_reduce(name, ...)` 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [ ] Docs build

### DIFF
--- a/src/kokkos_abstraction.hpp
+++ b/src/kokkos_abstraction.hpp
@@ -268,13 +268,13 @@ auto MakeFlatFunctor(F &function, Args... args) {
 
 template <typename Function, typename R, typename T, typename Index, typename... FArgs>
 class FlatFunctor<Function, R (T::*)(Index, Index, Index, FArgs...) const> {
-  int NjNi, Nj, Ni, kl, jl, il;
+  int NjNi, Ni, kl, jl, il;
   Function function;
 
  public:
-  FlatFunctor(const Function _function, const int _NjNi, const int _Nj, const int _Ni,
-              const int _kl, const int _jl, const int _il)
-      : function(_function), NjNi(_NjNi), Nj(_Nj), Ni(_Ni), kl(_kl), jl(_jl), il(_il) {}
+  FlatFunctor(const Function _function, const int _NjNi, const int _Ni, const int _kl,
+              const int _jl, const int _il)
+      : function(_function), NjNi(_NjNi), Ni(_Ni), kl(_kl), jl(_jl), il(_il) {}
   KOKKOS_INLINE_FUNCTION
   void operator()(const int &idx, FArgs &&...fargs) const {
     int k = idx / NjNi;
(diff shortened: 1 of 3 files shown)


Yurlungur (approved):
This is a nice fix. Thanks for implementing it!

src/kokkos_abstraction.hpp
```diff
@@ -258,6 +258,35 @@ par_dispatch(LoopPatternMDRange, const std::string &name, DevExecSpace exec_spac
...
+  void operator()(const int &idx, FArgs &&...fargs) const {
+    int k = idx / NjNi;
+    int j = (idx - k * NjNi) / Ni;
+    int i = idx - k * NjNi - j * Ni;
+    k += kl;
+    j += jl;
+    i += il;
+    function(k, j, i, std::forward<FArgs>(fargs)...);
+  }
+};
```
lroberts36: Probably not for this PR, but I wonder if we could make this work for an arbitrary dimensional index space using `IndexRange` in `utils/` with a little bit of template magic.

lroberts36 (approved):
LGTM, thanks for the fix!

src/kokkos_abstraction.hpp
acreyes: I think it could work. This seems to work and at least gets part way there  https://godbolt.org/z/fbYP5vW6r

lroberts36: This (https://godbolt.org/z/n48951jGK) seems to work, but I didn't explicitly pull out the function signature like you did (which I think matters for references).

acreyes: yeah I remember cuda builds had given me trouble with that. This [one](https://godbolt.org/z/4W37ejM7E) uses some more template helper structs to also handle `IndexRange`s in the constructor. I imagine you could use similar things to handle all the `par_dispatch` overloads as well

src/kokkos_abstraction.hpp
```diff
@@ -258,6 +258,35 @@ par_dispatch(LoopPatternMDRange, const std::string &name, DevExecSpace exec_spac
...
+class FlatFunctor;
+
+template <typename F, typename... Args>
+auto MakeFlatFunctor(F &function, Args... args) {
+  return FlatFunctor<F, decltype(&F::operator())>(function, std::forward<Args>(args)...);
+}
+
+template <typename Function, typename R, typename T, typename Index, typename... FArgs>
+class FlatFunctor<Function, R (T::*)(Index, Index, Index, FArgs...) const> {
+  int NjNi, Nj, Ni, kl, jl, il;
```
pgrete: `Nj` is not used, isn't it? So I suggest to remove it.
Same for the 4D version below.

pgrete (commented):
Thanks for the PR!

Out of curiosity, was there a specific use and/or performance consideration that inspired these changes?

src/kokkos_abstraction.hpp
acreyes: You're right, nice catch

acreyes:
> Thanks for the PR!
> 
> Out of curiosity, was there a specific use and/or performance consideration that inspired these changes?

Not any performance consideration. I was wanting to use `par_reduce` with similar default arguments as I was with `par_for`

src/kokkos_abstraction.hpp
pgrete: Saved for future reference here https://github.com/parthenon-hpc-lab/parthenon/issues/1134

pgrete (approved):

pgrete:
All tests pass. I'm force merging.

PR #1132, opened 2024-06-26
Fix regional dependencies for sub-`TaskList`s and make solvers work for abitrary partitioning

### LINKED ISSUE #1125: `TaskQualifier::*_sync` does not work for iterative task lists
When `TaskRegion::BuildGraph` is called, it connects regional tasks across all of the task lists in `TaskRegion::task_lists` but does not appear to do anything with the task lists contained in `TaskList::sublists`. As a result, the BiCGSTAB solver hangs for more than one partition since the residual is zeroed out by one task for a reduction before other tasks have the chance to check the residual in their completion task in the previous iteration.

I will try to take a look at how to fix this next week.

lroberts36:
## PR Summary
Currently, sublists of `TaskList`s do not have their regional dependencies set correctly. This results in the bug described in #1125. This PR changes `TaskRegion::BuildGraph()` to correctly register regional dependencies among all `TaskList`s and closes #1125. 

Additionally, it takes a second step to make the multigrid solver work when `parthenon/mesh/pack_size != -1` by including "self-communication" between coarse blocks on a finer two-level composite grid and fine leaf blocks on the next coarser two-level composite grid. Although no data is communicated in these communication channels, this introduces a dependency (outside of the tasking framework) that ensures operations on these blocks that are part of multiple gmg levels occur in the correct order.

Getting the multigrid stuff to work was a painful debugging process. As a result, I introduced some capability for more customized naming of tasks and the ability to write task labels to screen as the tasks are called. This turned out to be fairly useful for debugging what was going on.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [ ] Docs build


Yurlungur (commented):
I really feel like @jdolence should look at this one.

src/bvals/comms/bnd_info.cpp
```diff
@@ -177,7 +177,7 @@ CalcIndices(const NeighborBlock &nb, MeshBlock *pmb,
                                 (neighbor_bounds[dir].e - neighbor_bounds[dir].s + 1);
         s[dir] += nb.origin_loc.l(dir) % 2 == 1 ? extra_zones - interior_offset : 0;
         e[dir] -= nb.origin_loc.l(dir) % 2 == 0 ? extra_zones - interior_offset : 0;
-        if (ir_type == IndexRangeType::InteriorSend) {
+        if (ir_type == IndexRangeType::InteriorSend && !prores) {
```
lroberts36, line 180: Changes to `CalcIndices` are required since we are actually passing the correct neighbor information in `GetInterior*`.

src/bvals/comms/bnd_info.hpp
```diff
@@ -59,6 +59,7 @@ struct BndInfo {
   bool allocated = true;
   bool buf_allocated = true;
   int alloc_status;
+  bool same_to_same = false;
```
lroberts36, line 62: This flag denotes that this buffer represents a communication from this block to itself and that the buffer does not need to be filled (a message just needs to be sent).

src/bvals/comms/bvals_utils.hpp
```diff
@@ -35,22 +35,35 @@
...
+                                   BoundaryType btype) {
   const int sender_id = pmb->gid;
   const int receiver_id = nb.gid;
   const int location_idx = nb.offsets.GetIdx();
-  return {sender_id, receiver_id, pcv->label(), location_idx};
+  int other = (nb.gid == pmb->gid && (btype == BoundaryType::gmg_restrict_recv ||
+                                      btype == BoundaryType::gmg_restrict_send))
+                  ? 1
+                  : 0;
+  return {sender_id, receiver_id, pcv->label(), location_idx, other};
```
lroberts36, line 53: Add more information to the communication key so that we can represent same-to-same restriction and same-to-same prolongation as different channels.

src/tasks/tasks.hpp
```diff
@@ -106,7 +113,11 @@ class Task {
   template <typename TID>
   Task(TID &&dep, const std::string &label, const std::function<TaskStatus()> &func,
        std::pair<int, int> limits = {1, 1})
-      : label_(label), f(func), exec_limits(limits) {
+      : Task(std::forward<TID>(dep), label, 0, func, limits) {}
+  template <typename TID>
+  Task(TID &&dep, const std::string &label, int verbose_level,
+       const std::function<TaskStatus()> &func, std::pair<int, int> limits = {1, 1})
+      : label_(label), verbose_level_(verbose_level), f(func), exec_limits(limits) {
```
lroberts36, line 121: Add verbosity flag to tasks, if `verbose_level_ > 0` the task will print its name out to stdout before starting the task. Useful for debugging.

src/tasks/tasks.hpp
```diff
@@ -270,18 +283,23 @@ class TaskList {
...
-      return AddTaskImpl(tq, dep, {}, std::forward<Arg1>(arg1),
+      return AddTaskImpl(tq, dep, {}, 0, std::forward<Arg1>(arg1),
                          std::forward<Args>(args)...);
     } else if constexpr (std::is_convertible<Arg1, std::string>::value) {
-      return AddTaskImpl(tq, dep, std::forward<Arg1>(arg1), std::forward<Args>(args)...);
+      return AddTaskImpl(tq, dep, std::forward<Arg1>(arg1), 0,
+                         std::forward<Args>(args)...);
+    } else if constexpr (is_tuple_t<Arg1>::value) {
+      return AddTaskImpl(tq, dep, std::get<0>(arg1), std::get<1>(arg1), std::get<2>(arg1),
+                         std::forward<Args>(args)...);
```
lroberts36, line 296: Allow passing a `tuple` containing the name of the task, the verbosity level of the task, and the task function itself. Allows easy print out of the order in which tasks are run.

src/tasks/tasks.hpp
```diff
@@ -413,6 +431,12 @@ class TaskList {
     return WriteTaskGraph(stream, tasks);
   }
 
+  std::vector<TaskList *> GetAllTaskLists() {
+    std::vector<TaskList *> list;
+    GetAllTaskListsInternal(list);
+    return list;
+  }
```
lroberts36, line 441: Necessary for linking regional tasks in iterative task lists.

src/bvals/comms/bnd_info.cpp
```diff
@@ -330,6 +331,8 @@ BndInfo BndInfo::GetSetBndInfo(MeshBlock *pmb, const NeighborBlock &nb,
   } else if (buf_state == BufferState::received_null) {
     out.buf_allocated = false;
   } else {
+    printf("%i -> %i (Set %s) is in state %i.\n", nb.gid, pmb->gid, v->label().c_str(),
+           buf_state);
```
Yurlungur: Should you also print MPI rank?

src/interface/mesh_data.hpp
```diff
@@ -290,6 +290,14 @@ class MeshData {
 
   const auto &GetAllBlockData() const { return block_data_; }
 
+  bool ContainsGid(int gid) const {
```
pdmullen, line 293: Any concern about stale `gid`s  here (c.f., https://github.com/parthenon-hpc-lab/parthenon/pull/1010)?

src/tasks/tasks.hpp
```diff
@@ -120,6 +131,9 @@ class Task {
 
   TaskStatus operator()() {
     auto status = f();
+    if (verbose_level_ > 0)
+      std::cout << label_ << " [status = " << static_cast<int>(status) << "]"
+                << std::endl;
```
Yurlungur: Should these include MPI rank? Also `printf` more than stream is probably better because with MPI streams can get muddled. Better to send a whole string to `cout` at once.

Yurlungur (approved):
Very minor nitpicks on the debugging code. Otherwise LGTM.

src/tasks/tasks.hpp
pdmullen: This is a useful feature.  I wonder if the verbosity should be blanket applied to all tasks based off a `verbose` flag set in `pin`, rather than enrolled via each `AddTask`?

EDIT: this proposal could actually be handled via the downstream app

tst/regression/test_suites/poisson_gmg/parthinput.poisson
```diff
@@ -17,6 +17,7 @@ problem_id = poisson
 <parthenon/mesh>
 refinement = static
 multigrid = true 
+pack_size = 1
```
pdmullen, line 20: excellent

pdmullen (approved):
LGTM! I have tested this in a downstream app invoking iterative tasking.  This MR *fixes* (1) hangs and (2) undefined behavior when invoking `pack_size = 1` that currently exist with `parthenon/develop`.

src/bvals/comms/bnd_info.cpp
lroberts36: yeah, we could include that.

src/tasks/tasks.hpp
lroberts36: sure, that is a valid concern. I will include rank as well.

src/interface/mesh_data.hpp
lroberts36: I don't think so, since the `MeshData` should be cleared and recreated on remeshing. This doesn't rely on the gids of the variables themselves.

src/tasks/tasks.hpp
lroberts36: I agree that would be nice, but I think that it requires a bit of threading to make it work (since the tasking stuff doesn't directly access `pin` currently). I would leave this for a future PR. 

In practice, I just manually comment out the `verbose_level_ > 0` test and recompile if I want to see everything...

src/bvals/comms/bnd_info.cpp
lroberts36: done

src/tasks/tasks.hpp
lroberts36: done

PR #1137, opened 2024-07-10
[Breaking] Another stab at more general coordinate support

jdolence:
## PR Summary
Add support for spherical and cylindrical coordinates.

Might break downstream because of removal of `FA` versions of functions.  Because coordinates no longer use variadics, there's no longer a need to disambiguate via function names.  Different args are sufficient.  The fix should be to just remove `FA` from all calls.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [ ] Docs build


src/coordinates/uniform_coordinates.hpp
```diff
@@ -0,0 +1,326 @@
...
+  }
+
+  //----------------------------------------
+  // Xc: Positions at cell centroids
+  //----------------------------------------
+  template <int dir>
+  KOKKOS_FORCEINLINE_FUNCTION Real Xc(const int idx) const {
+    static_assert(dir > 0 && dir < 4);
+    return xmin_[dir - 1] + (idx + 0.5) * dx_[dir - 1];
+  }
```
Yurlungur, line 121: looks like you're shadowing these in derived classes. Should they be marked `overload` or something? Or is that not necessary due to CRTP?

src/coordinates/uniform_spherical.hpp
```diff
@@ -0,0 +1,155 @@
...
+      const Real sth1 = std::sin(th1);
+      const Real cth1 = std::cos(th1);
+      return (th0 * cth0 - th1 * cth1 - sth0 + sth1) / (cth0 - cth1);
+    }
+    return base_t::Xc<X3DIR>(idx);
+  }
+
+  template <int dir>
+  KOKKOS_FORCEINLINE_FUNCTION
+  Real hx(const Real r, const Real th, const Real phi) const {
```
Yurlungur: so this is the diagonal metric `h` factor but what is the `x` for in the name?

Yurlungur: And do we need any additional curvature terms we should track?

src/coordinates/uniform_coordinates.hpp
```diff
@@ -0,0 +1,326 @@
...
+    else if constexpr (dir == X2DIR) return static_cast<const System *>(this)->template Dxc<dir>(j);
+    else if constexpr (dir == X3DIR) return static_cast<const System *>(this)->template Dxc<dir>(k);
+    PARTHENON_FAIL("Unknown dir.");
+    return 0.0;
+  }
+  KOKKOS_FORCEINLINE_FUNCTION Real Dxc(const int dir, const int k, const int j, const int i) const {
+    assert(dir > 0 && dir < 4);
+    if (dir == X1DIR) return static_cast<const System *>(this)->template Dxc<X1DIR>(k, j, i);
+    else if (dir == X2DIR) return static_cast<const System *>(this)->template Dxc<X2DIR>(k, j, i);
+    else if (dir == X3DIR) return static_cast<const System *>(this)->template Dxc<X3DIR>(k, j, i);
```
lroberts36: Are the `if`s necessary here?

src/coordinates/uniform_coordinates.hpp
```diff
@@ -0,0 +1,326 @@
...
+
+  template <int dir>
+  KOKKOS_FORCEINLINE_FUNCTION Real Dxc(const int idx) const {
+    static_assert(dir > 0 && dir < 4);
+    return dx_[dir - 1];
+  }
+  template <int dir>
+  KOKKOS_FORCEINLINE_FUNCTION Real Dxc(const int k, const int j, const int i) const {
+    static_assert(dir > 0 && dir < 4);
+    if constexpr (dir == X1DIR) return static_cast<const System *>(this)->template Dxc<dir>(i);
```
lroberts36: FWIW, could just have one line `return static_cast<const System *>(this)->template Dxc<dir>(std::get<dir - 1>(std::make_tuple(i, j, k)));`

src/coordinates/uniform_coordinates.hpp
```diff
@@ -0,0 +1,326 @@
...
+    else if constexpr (dir == X2DIR) return static_cast<const System *>(this)->template Xc<dir>(j);
+    else if constexpr (dir == X3DIR) return static_cast<const System *>(this)->template Xc<dir>(k);
+    return 0; // To appease compiler
+  }
+
+  //----------------------------------------
+  // Xf: Positions on Faces
+  //----------------------------------------
+  template <int dir>
+  KOKKOS_FORCEINLINE_FUNCTION Real Xf(const int idx) const {
```
lroberts36, line 138: Should this still be supported?

src/coordinates/uniform_coordinates.hpp
```diff
@@ -0,0 +1,326 @@
...
+    return 0; // To appease compiler
+  }
+
+  template <int dir>
+  KOKKOS_FORCEINLINE_FUNCTION Real Xf(const int k, const int j, const int i) const {
+    return Xf<dir, dir>(k, j, i);
+  }
+
+  template <int dir, TopologicalElement el>
+  KOKKOS_FORCEINLINE_FUNCTION Real X(const int idx) const {
```
lroberts36, line 169: Should this still be supported?

src/coordinates/uniform_coordinates.hpp
```diff
@@ -0,0 +1,326 @@
...
+        return area_[X3DIR - 1] / 4.0;
+      } else if (el == TE::E1) {
+        return dx_[X1DIR - 1] / 2.0;
+      } else if (el == TE::E2) {
+        return dx_[X2DIR - 1] / 2.0;
+      } else if (el == TE::E3) {
+        return dx_[X3DIR - 1] / 2.0;
+      } else if (el == TE::NN) {
+        return 1.0;
+      }
```
lroberts36: I guess this still needs to be updated.

src/coordinates/uniform_spherical.hpp
```diff
@@ -0,0 +1,158 @@
...
+    static_assert(dir > 0 && dir < 4);
+    // only the index along dir is relevant in Xc, so offsetting all three is OK
+    return Xc<dir>(idx) - Xc<dir>(idx-1);
+  }
+
+  //----------------------------------------
+  // Xc: Positions at cell centroids
+  //----------------------------------------
+  template <int dir>
+  KOKKOS_FORCEINLINE_FUNCTION Real Xc(const int idx) const {
```
lroberts36, line 54: should this take `k, j,  i`?

src/coordinates/uniform_spherical.hpp
```diff
@@ -0,0 +1,158 @@
...
+      if (el == TE::CC) return CellVolume(k, j, i);
+      else if (el == TE::F1) return FaceArea<X1DIR>(k, j, i);
+      else if (el == TE::F2) return FaceArea<X2DIR>(k, j, i);
+      else if (el == TE::F3) return FaceArea<X3DIR>(k, j, i);
+      else if (el == TE::E1) return EdgeLength<X1DIR>(k, j, i);
+      else if (el == TE::E2) return EdgeLength<X2DIR>(k, j, i);
+      else if (el == TE::E3) return EdgeLength<X3DIR>(k, j, i);
+      else if (el == TE::NN) return 1.0;
+    } else {
+      PARTHENON_FAIL("Have not yet implemented fine fields for UniformSpherical coordinates.");
```
lroberts36: We could just have a coordinates object own another coordinates object that has twice the number of zones and then just call that for fine fields.

jdolence:
I need to add to the docs and such, but this seems to be working downstream.  Some (new) reviews would be nice while I add some documentation.

src/coordinates/uniform_cartesian.hpp
```diff
@@ -13,320 +13,21 @@
...
-  static const char *StaticName() { return name_; }
-
- private:
-  std::array<int, 3> istart_;
-  std::array<Real, 3> xmin_, dx_, area_;
-  Real cell_volume_;
   constexpr static const char *name_ = "UniformCartesian";
 
-  const std::array<Real, 3> &Dx_() const { return dx_; }
+ private:
```
lroberts36:
```suggestion
```

src/coordinates/uniform_coordinates.hpp
```diff
@@ -0,0 +1,368 @@
...
+  }
+  template <int dir, TopologicalElement el>
+  KOKKOS_FORCEINLINE_FUNCTION Real X(const int k, const int j, const int i) const {
+    static_assert(dir > 0 && dir < 4);
+    if constexpr (dir == X1DIR)
+      return X<dir, el>(i);
+    else if constexpr (dir == X2DIR)
+      return X<dir, el>(j);
+    else if constexpr (dir == X3DIR)
+      return X<dir, el>(k);
```
lroberts36, line 190: Should these be `return static_cast<const System *>(this)->template X<dir, el>(i)` etc., or is the idea that `X` will never be defined by a class that inherits from `UniformCoordinates` (which seems to be a reasonable assumption)?

src/coordinates/uniform_coordinates.hpp
lroberts36: I think this function is still inconsistent with `Real Volume(const int k, const int j, const int i)`

src/coordinates/uniform_cylindrical.hpp
```diff
@@ -0,0 +1,164 @@
...
+    if constexpr (dir == X1DIR) {
+      return Xf<X1DIR>(k, j, i) * Dx<X2DIR>() * Dx<X3DIR>();
+    } else if constexpr (dir == X2DIR) {
+      Real r0 = Xf<X1DIR>(k, j, i);
+      Real r1 = Xf<X1DIR>(k, j, i + 1);
+      return 0.5 * (r1 * r1 - r0 * r0) * Dx<X3DIR>();
+    }
+    Real r0 = Xf<X1DIR>(k, j, i);
+    Real r1 = Xf<X1DIR>(k, j, i + 1);
+    return 0.5 * (r1 * r1 - r0 * r0) * Dx<X2DIR>();
```
lroberts36: I think this expression for the area of the $\phi$-face is incorrect. It should be `Dx<X1DIR>() * Dx<X2DIR>`. (Although I am guessing this is never used)

lroberts36 (approved):
LGTM aside from the small comments below. It might also be nice to have a brief comment about each coordinate system at the top of each file (e.g. X1 -> r, X2 -> z, X3 -> phi, etc.).

src/coordinates/uniform_cylindrical.hpp
jdolence: Do you mean `Volume(CellLevel cl, TopologicalElement el, ...)` ought to be doing things like `static_cast<const System *>(this)->CellVolume(k, j, i)` etc instead of using `cell_volume_`, `area[]`, etc.?

src/coordinates/uniform_coordinates.hpp
jdolence: Yeah I don't think `X` should ever be redefined.  I'm not sure if there's a way to enforce that.  Are you aware of any way to do that?

src/coordinates/uniform_cylindrical.hpp
jdolence: > I think this expression for the area of the ϕ -face is incorrect. It should be `Dx<X1DIR>() * Dx<X2DIR>`. (Although I am guessing this is never used)

oof.  this is not even dimensionally correct.  wonder what I was doing...

src/coordinates/uniform_coordinates.hpp
lroberts36: The only way I know of would be to declare the function as`virtual double X(int idx) final {...}`, which would prevent it from being overridden. I think this would imply that calling `X` would require a vtable lookup, so it is probably a non-starter.

src/coordinates/uniform_cylindrical.hpp
lroberts36: > Do you mean `Volume(CellLevel cl, TopologicalElement el, ...)` ought to be doing things like `static_cast<const System *>(this)->CellVolume(k, j, i)` etc instead of using `cell_volume_`, `area[]`, etc.?

Yeah, that is what I was thinking.

jdolence: Think I got all these.

src/coordinates/uniform_cartesian.hpp
jdolence: Got this one and a couple others like it.

src/coordinates/uniform_spherical.hpp
jdolence: I thought about doing this for a sec.  Can we just leave this for future work, if/when someone tries to use fine fields in non-Cartesian coordinates?  It will at least produce an error now that is very obvious.

lroberts36: Yeah, leaving it for the future if it is needed sounds good to me. I don't think of anywhere this is likely to crop up for the time being.

src/coordinates/uniform_coordinates.hpp
lroberts36: It could be that the compiler is smart enough to realize that such calls can be "de-virtualized" though, not sure

src/coordinates/uniform_spherical.hpp
```diff
@@ -0,0 +1,195 @@
...
+      const Real r1sq = r1 * r1;
+      return 0.75 * (r1sq * r1sq - r0sq * r0sq) / (r1sq * r1 - r0sq * r0);
+    } else if constexpr (dir == X2DIR) {
+      const Real th0 = Xf<X2DIR>(idx);
+      const Real sth0 = std::sin(th0);
+      const Real cth0 = std::cos(th0);
+      const Real th1 = Xf<X2DIR>(idx + 1);
+      const Real sth1 = std::sin(th1);
+      const Real cth1 = std::cos(th1);
+      return (th0 * cth0 - th1 * cth1 - sth0 + sth1) / (cth0 - cth1);
```
pdmullen, line 69: not sure where to put this point... maybe here... but there is a choice on how to handle what theta and phi calls return in ghost cells when they are outside 0,pi and/or 0,2pi respectively... i.e., should they always be mapped back to be between 0,pi and 0,2pi respectively?

src/coordinates/uniform_spherical.hpp
```diff
@@ -0,0 +1,197 @@
...
+
+  //----------------------------------------
+  // FaceArea: Area of cell areas
+  //----------------------------------------
+  template <int dir>
+  KOKKOS_FORCEINLINE_FUNCTION Real FaceArea(const int k, const int j, const int i) const {
+    static_assert(dir > 0 && dir < 4);
+    if constexpr (dir == X1DIR) {
+      Real dOmega =
+          Dx<X3DIR>() * (std::cos(Xf<X2DIR>(k, j, i)) - std::cos(Xf<X2DIR>(k, j + 1, i)));
```
pdmullen: Athena++ and Artemis throw in an `abs` on the term in parenthesis.  I am not sure if this is related to my comment above about ghost cells outside the 0,pi range.  Just flagging this, since I am spot checking consistency among different codebases.

pdmullen (commented):
Spot checked for consistency with other codebases and looks good.

src/coordinates/uniform_spherical.hpp
jdolence: is there a question also about negative `r`?

what is the argument that they should be mapped back to [0,pi] or [0,2pi]?  Internally, you can get things like negative `CellVolume` or `FaceArea` in the ghost zones.  I think this is related to your point below about inserting `abs` in some places, though I'm not sure of a context where you need these things.  In the example above (`Xc`) I think it's potentially a very bad idea to enforce results always lying within [0,pi] or [0,2pi].  It would imply a huge jump in, for example, theta_centroid between cells.

src/coordinates/uniform_spherical.hpp
jdolence: I think `abs` in places like this probably make sense.  I'll add those.

src/coordinates/uniform_spherical.hpp
pdmullen: > is there a question also about negative r?

Good question.  I guess I was mostly thinking about phi because of periodicity.  

> what is the argument that they should be mapped back to [0,pi] or [0,2pi]? Internally, you can get things like negative CellVolume or FaceArea in the ghost zones. I think this is related to your point below about inserting abs in some places, though I'm not sure of a context where you need these things. In the example above (Xc) I think it's potentially a very bad idea to enforce results always lying within [0,pi] or [0,2pi]. It would imply a huge jump in, for example, theta_centroid between cells.

Artemis and Athena++ have made the same choices you have made here, and things are working just fine there.  I was just curious if we should anticipate any ambiguity for the downstream user on what the cell positions should returned if outside [0,2pi].  e.g., if x3min=0 and x3max=2pi, would a user be confused to find that `Xf<X3DIR>(kb.e + 2)` doesn't return 0 + dx3f?

src/CMakeLists.txt
```diff
@@ -90,7 +90,18 @@ set(COMPILED_WITH ${CMAKE_CXX_COMPILER})
...
-set(COORDINATE_TYPE UniformCartesian) # TODO: Make this an option when more are available
+set(PARTHENON_COORDINATES "Cartesian" CACHE STRING "Set the coordinate system from {Cartesian, Cylindrical, Spherical} with Cartesian the default")
+
+if (${PARTHENON_COORDINATES} STREQUAL "Cartesian")
+  set(COORDINATE_TYPE UniformCartesian)
+elseif(${PARTHENON_COORDINATES} STREQUAL "Cylindrical")
+  set(COORDINATE_TYPE UniformCylindrical)
+elseif(${PARTHENON_COORDINATES} STREQUAL "Spherical")
+  set(COORDINATE_TYPE UniformSpherical)
+else()
```
Yurlungur: I don't love this switch statement honestly... I'd rather it be more flexible, but I won't push too hard on that.

src/coordinates/coordinates.hpp
```diff
@@ -16,11 +16,18 @@
...
+#include "uniform_spherical.hpp"
 
 namespace parthenon {
 
 using Coordinates_t = COORDINATE_TYPE;
 
+template <typename T>
+constexpr bool IsCoord() {
+  return std::is_same_v<T, Coordinates_t>;
+}
```
Yurlungur, line 29: :+1:

src/coordinates/uniform_cartesian.hpp
```diff
@@ -13,320 +13,19 @@
...
-#include "defs.hpp"
-#include "globals.hpp"
-
-#include <Kokkos_Macros.hpp>
+#include "uniform_coordinates.hpp"
 
 namespace parthenon {
 
-class UniformCartesian {
+class UniformCartesian : public UniformCoordinates<UniformCartesian> {
```
Yurlungur, line 20: Good 'ol CRTP

src/coordinates/uniform_coordinates.hpp
Yurlungur: I think it probably should... if AthenaPK uses it. @pgrete ?

src/coordinates/uniform_coordinates.hpp
Yurlungur: It's probably enough to just document it. Which does mean you need to write docs :P

src/coordinates/uniform_spherical.hpp
Yurlungur: I think the choice here is correct. Agree that physically there's no negative r or negative theta. But I think it's useful to be able to pull that out if needed. And downstream it's easy enough to add an absolute value if desired.

src/coordinates/uniform_spherical.hpp
Yurlungur: My rule of thumb is coordinate positions should not have abs, but "generalized volumes" (i.e., measures) should

Yurlungur (approved):
This came out remarkably clean. Nice work @jdolence ! Ideally, I'd like to see this machinery documented... and also an integrated example that uses it. Not sure what a good example would be...

src/CMakeLists.txt
jdolence: Dealt with this as you've already seen.

Yurlungur (approved):
LGTM!

jdolence:
@pgrete @bprather @AstroBarker @pdmullen @adamdempsey90 Depending on which coordinate functions you're using downstream, this PR may be breaking.  I _believe_ the only breaking change is the removal of the coordinate functions with `FA` in them.  That should be quite easy to fix, I'd think.

Yurlungur:
> @pgrete @bprather @AstroBarker @pdmullen @adamdempsey90 Depending on which coordinate functions you're using downstream, this PR may be breaking. I _believe_ the only breaking change is the removal of the coordinate functions with `FA` in them. That should be quite easy to fix, I'd think.

I checked in Phoebus and indeed we do make FA calls. But we just have to remove the letters FA and the calls will work as desired/expected.

adamdempsey90:
> @pgrete @bprather @AstroBarker @pdmullen @adamdempsey90 Depending on which coordinate functions you're using downstream, this PR may be breaking. I _believe_ the only breaking change is the removal of the coordinate functions with `FA` in them. That should be quite easy to fix, I'd think.

artemis uses it's own implementation of all of this, so we don't call these FA routines

jdolence:
On the advice of others, I'm just going to merge this.  If it breaks things downstream beyond needing to just remove FA, we can always walk it back.

@pgrete, please forgive me...

pgrete:
Looks like I'm late to the party @jdolence 
When bumping AthenaPK I didn't come across issues with the `FA` calls, but I saw that
`Dxf<int dir, int face>(const int idx); //Distance between face centers along dir`
has been removed.
I didn't see any replacement. Did I miss sth or was this removed on purpose?

Issue #1134, opened 2024-07-04
Using `IndexRange` in `FlatFunctor`


pgrete:
yeah I remember cuda builds had given me trouble with that. This [one](https://godbolt.org/z/4W37ejM7E) uses some more template helper structs to also handle `IndexRange`s in the constructor. I imagine you could use similar things to handle all the `par_dispatch` overloads as well

_Originally posted by @acreyes in https://github.com/parthenon-hpc-lab/parthenon/pull/1130#discussion_r1653624413_
            

As discussed above, there might be some general enhancement around this.
Opening this issue with the reference to the comment (and the two above that one) as it included some test code.

PR #1142, opened 2024-07-29
Unify `par_dispatch`, `par_for_outer` & `par_for_inner` overloads

### LINKED ISSUE #185: Use of `IndexRange`s in `par_for` wrappers
As pointed out by @JoshuaSBrown the parallel for wrappers currently take a long list index arguments. Using the `IndexRange` instead would mitigate problems arising from , e.g., incorrect order.

acreyes:
## PR Summary
Provides a single overload for `par_dispatch`, `par_for_outer` & `par_for_inner` that can handle both integer and `IndexRange` launch bounds.

* `par_dispatch` & `par_for_outer` loops are handled by the same `par_dispatch_impl` struct  that constructs the appropriate kokkos policy and functor  for `kokkos_dispatch`
* relies on the `TypeList` struct to hold parameter packs and accompanying type traits to figure out function signatures
* Similar pattern for `par_for_inner`
* Introduces a new loop pattern `LoopPatternCollapse<team, thread, vector>` that can be used to collapse a general ND loop over any combination of kokkos teams, threads and vectors inspired by `#pragma acc collapse` directives
    * specializes to `LoopPatternTPTTR`, `LoopPatternTPTVR`, `LoopPatternTPTTRTVR`, `InnerLoopPatternTTR` and `InnerLoopPatternTVR` patterns
* Fallbacks for incompatible `Tag`s & `Pattern`s that can show up from `DEFAULT_LOOP_PATTERN`
* generalizes tests for `par_for` & `par_reduce` and improves coverage for all patterns up to rank 7 loops

Addresses #1134 
## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build


src/kokkos_abstraction.hpp
```diff
@@ -171,6 +177,43 @@ static struct LoopPatternTPTTRTVR {
...
+struct LoopPatternTeam : std::false_type {};
+
+template <size_t team, size_t thread, size_t vector>
+struct LoopPatternTeam<LoopPatternCollapse<team, thread, vector>, team + thread + vector,
+                       void> : std::true_type {
+  static constexpr size_t Nvector = vector;
+  static constexpr size_t Nthread = thread;
+  static constexpr size_t Nteam = team;
+  using LoopPattern = LoopPatternCollapse<team, thread, vector>;
+};
```
acreyes: This doesn't compile in gcc <12 or nvcc < 24.5

[https://godbolt.org/z/fW56WvbKf](https://godbolt.org/z/fW56WvbKf)

src/kokkos_abstraction.hpp
```diff
@@ -214,470 +280,677 @@ inline void kokkos_dispatch(ParallelScanDispatch, Args &&...args) {
...
+struct SequenceOfOnes<N, VAL, std::integer_sequence<size_t, ones...>> {
+  using value =
+      typename SequenceOfOnes<N - 1, VAL,
+                              std::integer_sequence<size_t, VAL, ones...>>::value;
+};
+
+template <size_t N, size_t VAL = 1>
+using sequence_of_ones = SequenceOfOnes<N - 1, VAL, std::integer_sequence<size_t, VAL>>;
+
+} // namespace meta
```
acreyes: These aren't specific to kokkos_abstractions and might have a better home somewhere else in `utils`?

lroberts36: Yeah, I think it would be better if they live somewhere in `utils`. I have put some template related stuff in `concepts_lite.hpp` in the past, but this might be enough to make a new header in utils called `type_lists.hpp` or something.

lroberts36: I ended up adding some type list functionality in `utils/type_list.hpp` in #1143 for dealing with grid variable types. I am not sure how much overlap there is with what you are doing there, but at least that could be a place to put this stuff (at least once #1143 is in develop).

acreyes: Nice! There is some overlap but I think your use of tuples looks nicer than all the recursive templates I used

src/kokkos_abstraction.hpp
```diff
@@ -214,470 +286,561 @@ inline void kokkos_dispatch(ParallelScanDispatch, Args &&...args) {
...
 
-template <typename Function, typename R, typename T, typename Index, typename... FArgs>
-class FlatFunctor<Function, R (T::*)(Index, Index, Index, FArgs...) const> {
-  int NjNi, Ni, kl, jl, il;
+ public:
+  using LaunchBounds = typename LB::value;
+  using Rank = std::integral_constant<size_t, LB::NumInds::value / 2>;
+  using Function = typename TL::template type<LaunchBounds::n_types>;
+  using Args = typename TL::template continuous_sublist<LaunchBounds::n_types + 1,
+                                                        TL::n_types - 1>;
```
acreyes:
```
/Users/runner/work/parthenon/parthenon/src/utils/type_list.hpp:69:39: error: 'ContinuousSublist' is a private member of 'parthenon::TypeList<int, unsigned long, (lambda at /Users/runner/work/parthenon/parthenon/src/prolong_restrict/pr_loops.hpp:69:69)>'
  using continuous_sublist = decltype(ContinuousSublist<Start, End>());
                                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/runner/work/parthenon/parthenon/src/kokkos_abstraction.hpp:367:38: note: in instantiation of template type alias 'continuous_sublist' requested here
  using Args = typename TL::template continuous_sublist<LaunchBounds::n_types + 1,
```
@lroberts36 it seems the macOS CI compiler takes issue with the `continuous_sublist` typedef `decltype`ing a private member function. Is there an obvious workaround besides making `ContinuousSublist` public?

lroberts36: Hm, I don't know an obvious way to fix this aside from what you suggest (i.e. making `ContinuousSublist` public). I think that would be fine, but maybe we should rename `ContinuousSublist` to `ContinuousSublistImpl`.

src/kokkos_abstraction.hpp
```diff
@@ -214,470 +286,558 @@ inline void kokkos_dispatch(ParallelScanDispatch, Args &&...args) {
...
+  static constexpr bool team_mbr = std::is_same_v<team_mbr_t, base_type<Index>>;
+  using TL = TypeList<Index, Args...>;
+
+ public:
+  using IndexND = typename TL::template continuous_sublist<0, Rank + team_mbr - 1>;
+  using FArgs = typename TL::template continuous_sublist<Rank + team_mbr>;
+};
+
+template <size_t Rank, typename F>
+using function_signature = FunctionSignature<Rank, decltype(&base_type<F>::operator())>;
```
lroberts36: This feels like it belongs in `utils` somewhere, possibly in `type_list`

src/kokkos_abstraction.hpp
```diff
@@ -171,14 +175,66 @@ static struct LoopPatternTPTTRTVR {
...
+    : std::true_type {
+
+  static constexpr bool IsTPTTR =
+      std::is_same<Pattern, LoopPatternTPTTR>::value; // inner TeamThreadRange
+  static constexpr bool IsTPTVR =
+      std::is_same<Pattern, LoopPatternTPTVR>::value; // inner ThreadVectorRange
+  static constexpr bool IsTPTTRTVR = std::is_same<Pattern, LoopPatternTPTTRTVR>::value;
+
+  static constexpr std::size_t Nvector = IsTPTVR || IsTPTTRTVR;
+  static constexpr std::size_t Nthread = IsTPTTR || IsTPTTRTVR;
```
lroberts36: I am probably missing something, but does this result in thread and vector range loop sizes of only zero or one? Is that what we want here?

Yurlungur (commented):
This is cool! One concern I have about the simd case---are we sure it still effectively vectorizes? My experience in the past has been that it can be difficult to get the compiler to recognize vectorization if there's too much obfuscating the raw for loop.

I also worry a bit that this machinery is harder to interpret/extend without deep template knowledge. That said, the tag dispatch all still exists so something can be hacked in, it just won't be part of the generic machinery.

src/kokkos_abstraction.hpp
```diff
@@ -214,470 +286,558 @@ inline void kokkos_dispatch(ParallelScanDispatch, Args &&...args) {
...
+    for (int i = 1; i < Rank; i++) {
+      inds[i] = idx;
+      inds[i - 1] /= strides[i - 1];
+      for (int j = 0; j < i; j++) {
+        inds[i] -= inds[j] * strides[j];
+      }
+    }
+    for (int i = 0; i < Rank; i++) {
+      inds[i] += ranges[i].s;
+    }
```
lroberts36: FWIW, if vectorization is an issue this sort of loop flattening is provided by `utils/indexer.hpp` but the index calculation loops are unrolled explicitly at compile time.

acreyes:
> This is cool! One concern I have about the simd case---are we sure it still effectively vectorizes? My experience in the past has been that it can be difficult to get the compiler to recognize vectorization if there's too much obfuscating the raw for loop.

That's a valid concern. I don't think the raw for loop is any more obfuscated to the compiler than the previous code. The innermost loop is its own function specialization
```cpp
  template <typename Function, std::size_t... Is>
  inline void dispatch_simd(std::integer_sequence<size_t, Is...>, Function &function) {
#pragma omp simd
    for (int i = mdrange.lower[Rank - 1]; i <= mdrange.upper[Rank - 1]; i++) {
      function(indices[Is]..., i);
    }
  }
```
I tried compiling the `kokkos_abstraction.cpp` tests with `-fopt-info-vec-missed` and don't see any misses with that block of code, where the outer loops are being passed over for vectorization. `-fopt-info-vec` does show some successful vectorizations there, but I imagine this is all very compiler/hardware dependent.

> I also worry a bit that this machinery is harder to interpret/extend without deep template knowledge. That said, the tag dispatch all still exists so something can be hacked in, it just won't be part of the generic machinery.

I agree with this, `par_dispatch` overloads can always be added in the same way. This does simplify the user experience  in that `DEFAULT_LOOP_PATTERN` just works so long as it is handled by the generic overload. I was finding that I would compile on a new machine with a different kokkos back end and some seemingly random loops suddenly wouldn't compile.

Yurlungur:
> I tried compiling the `kokkos_abstraction.cpp` tests with `-fopt-info-vec-missed` and don't see any misses with that block of code, where the outer loops are being passed over for vectorization. `-fopt-info-vec` does show some successful vectorizations there, but I imagine this is all very compiler/hardware dependent.

:+1: @lroberts36 or @jdolence we should probably try running riot through an intel optimization report with this branch as a sanity check.
> 
> I agree with this, `par_dispatch` overloads can always be added in the same way. This does simplify the user experience in that `DEFAULT_LOOP_PATTERN` just works so long as it is handled by the generic overload. I was finding that I would compile on a new machine with a different kokkos back end and some seemingly random loops suddenly wouldn't compile.

Yeah I've experience this as well---and this is a nice fix/cleanup for that problem.

Yurlungur (approved):
Would like to see a review from, e.g., @pgrete and an optimization report for our internal code before merge. But just going to go ahead and approve now.

acreyes:
What is the expected behavior for an `InnerLoopPatternSimdFor` nested in a device `par_for_outer` like done [here](https://github.com/parthenon-hpc-lab/parthenon/blob/9e2c7eafd85b8efb1ba40089a5b10da2c4f15e12/src/solvers/solver_utils.hpp#L250)? 

There is a comment
```cpp
// Translate to a non-Kokkos plain C++ innermost loop (single index)
// decorated with #pragma omp simd
// IMPORTANT: currently only supported on CPUs
struct InnerLoopPatternSimdFor {};
constexpr InnerLoopPatternSimdFor inner_loop_pattern_simdfor_tag;
```
that I interpret to mean that pattern shouldn't be supported. Previously the simd overloads of `par_for_inner` were `KOKKOS_FORCEINLINE_FUNCTION` and would compile just fine. I can add the `KOKKOS` macros to the simd patterns in the new machinery in this PR to match the old behavior. Alternatively I can more directly enforce `DevExecSpace == HostExecSpace` for simd loops.

lroberts36 (commented):
Very cool! It is great to be moving toward a more generic `par_dispatch` system, I feel like I run into edge cases all the time. That being said, I share the concerns of @Yurlungur about the template complexity required for this. I think I would be challenged to add a new type of loop pattern, for instance. As I was going through the review, I started trying to take some of your approach and see if I could distill it a little bit so that the loop patterns were easier to parse. An attempt at this is in #1156. Perhaps we should discuss what the pros and cons of different choices for the code structure are some during the next Parthenon sync. 

I will also try and look at vectorization downstream in the next few days, although I tend to agree that if it used to vectorize then it should still vectorize given the limited changes to the inner most simd loop (although of course there is never a guarantee).

src/kokkos_abstraction.hpp
acreyes: this is really helpful!

src/kokkos_abstraction.hpp
acreyes: I think that's how they were before. A single thread/vector range for the innermost loop(s) and a team outer loop that flattens everything else.

lroberts36: I was misreading. It sounds like `Nvector` is the number of loop dimensions included in the vector loop level, but I was thinking it was the amount of work in the vector loop.

Yurlungur:
> What is the expected behavior for an `InnerLoopPatternSimdFor` nested in a device `par_for_outer` like done [here](https://github.com/parthenon-hpc-lab/parthenon/blob/9e2c7eafd85b8efb1ba40089a5b10da2c4f15e12/src/solvers/solver_utils.hpp#L250)?
> 
> There is a comment
> 
> ```c++
> // Translate to a non-Kokkos plain C++ innermost loop (single index)
> // decorated with #pragma omp simd
> // IMPORTANT: currently only supported on CPUs
> struct InnerLoopPatternSimdFor {};
> constexpr InnerLoopPatternSimdFor inner_loop_pattern_simdfor_tag;
> ```
> 
> that I interpret to mean that pattern shouldn't be supported. Previously the simd overloads of `par_for_inner` were `KOKKOS_FORCEINLINE_FUNCTION` and would compile just fine. I can add the `KOKKOS` macros to the simd patterns in the new machinery in this PR to match the old behavior. Alternatively I can more directly enforce `DevExecSpace == HostExecSpace` for simd loops.

I think we should enforce `DevExecSpace == HostExecSpace` for these---running them on device as a default inner loop pattern will produce poor performance and it won't be intuitive to a user why that's happening. Better IMO to produce an error message at compile time. Curious what others think though.

lroberts36:
> > What is the expected behavior for an `InnerLoopPatternSimdFor` nested in a device `par_for_outer` like done [here](https://github.com/parthenon-hpc-lab/parthenon/blob/9e2c7eafd85b8efb1ba40089a5b10da2c4f15e12/src/solvers/solver_utils.hpp#L250)?
> > There is a comment
> > ```c++
> > // Translate to a non-Kokkos plain C++ innermost loop (single index)
> > // decorated with #pragma omp simd
> > // IMPORTANT: currently only supported on CPUs
> > struct InnerLoopPatternSimdFor {};
> > constexpr InnerLoopPatternSimdFor inner_loop_pattern_simdfor_tag;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > that I interpret to mean that pattern shouldn't be supported. Previously the simd overloads of `par_for_inner` were `KOKKOS_FORCEINLINE_FUNCTION` and would compile just fine. I can add the `KOKKOS` macros to the simd patterns in the new machinery in this PR to match the old behavior. Alternatively I can more directly enforce `DevExecSpace == HostExecSpace` for simd loops.
> 
> I think we should enforce `DevExecSpace == HostExecSpace` for these---running them on device as a default inner loop pattern will produce poor performance and it won't be intuitive to a user why that's happening. Better IMO to produce an error message at compile time. Curious what others think though.

One argument for letting `DevExecSpace != HostExexSpace` is that users could use only `par_for` loops in a hierarchical setup like
```c++
parthenon::par_for_outer(
      outer_loop_pattern, "unit test Nested 4D", exec_space, scratch_size_in_bytes,
      scratch_level, 0, Nb, 0, N - 1,
      KOKKOS_LAMBDA(parthenon::team_mbr_t team_member, const int b, const int k,
                    const int j) {
        parthenon::par_for_inner(InnerLoopPatternSimdFor(), team_member, 0, Nl, [&](const int l){
          parthenon::par_for_inner(InnerLoopPatternTVR(), team_member, 0, N - 1, 0, N - 1,
                                   [&](const int j, const int i) { 
                                        v(b, l, k, j, i)++; 
           });
        });
      });
```
rather than write out the raw loop. I am not sure that I think that use case is particularly important to support though.

acreyes:
> One argument for letting `DevExecSpace != HostExexSpace` is that users could use only `par_for` loops in a hierarchical setup like
> 
> ```c++
> parthenon::par_for_outer(
>       outer_loop_pattern, "unit test Nested 4D", exec_space, scratch_size_in_bytes,
>       scratch_level, 0, Nb, 0, N - 1,
>       KOKKOS_LAMBDA(parthenon::team_mbr_t team_member, const int b, const int k,
>                     const int j) {
>         parthenon::par_for_inner(InnerLoopPatternSimdFor(), team_member, 0, Nl, [&](const int l){
>           parthenon::par_for_inner(InnerLoopPatternTVR(), team_member, 0, N - 1, 0, N - 1,
>                                    [&](const int j, const int i) { 
>                                         v(b, l, k, j, i)++; 
>            });
>         });
>       });
> ```
> 
> rather than write out the raw loop. I am not sure that I think that use case is particularly important to support though.

I see the utility in that. There could be a separate `parthenon::raw_for()` that would effectively do the same thing as the simd-inner but doesn't suggest that there is any performance to be had from its use.

lroberts36:
Testing vectorization using this branch in Riot with gcc/9.4.0 on a skylake-platinum node with -O3 -march=skylake-avx512 I get that with the old version of par_dispatch the simd for loop is vectorized in 191 places and with the new version it vectorizes in 158 places. Not sure what to make of this difference. It is hard to determine more from the gcc vectorization report. Trying to run, I get a segfault during initialization.

src/utils/indexer.hpp
```diff
@@ -120,35 +134,94 @@ struct Indexer {
...
   }
 
-  std::array<int, sizeof...(Ts)> N;
-  std::array<int, sizeof...(Ts)> start;
-  std::array<int, sizeof...(Ts)> end;
+  Kokkos::Array<int, sizeof...(Ts)> N;
+  Kokkos::Array<int, sizeof...(Ts)> start;
+  Kokkos::Array<int, sizeof...(Ts)> end;
+  std::size_t _size;
+};
```
acreyes, line 207: I switched these to `Kokkos::Array` to avoid an endless stream of `__host__` function called from `__device__ __host__` function warnings from `std::array`'s allocator and `operator[]`. I'm not familiar with the nuance between the two outside of being able to capture from device code or not.

lroberts36: I am not aware of any functional difference between them (aside from capturability), so this looks good to me. In #1156, I had gone to filling C-style arrays as a way to deal with filling indices in the hierarchical loops into a single array.

lroberts36:
Running on a Volta GPU with 
```
Currently Loaded Modules:
  1) nvhpc/23.5   2) cuda/11.4.2 ...
```
everything compiles (but with lots of warnings because this version of CUDA apparently doesn't like return statements only in an `if constexpr(...){return a; } else {return b;}` block). The test `par_for loops` and many of the regression tests segfault though. A bit surprising since it looks like the GitHub tests are succeeding. The output from the `par_for loops` test is: 
```
****************************************** Beginning Precheck ******************************************

Number of GPUs detected per node: 1
Number of GPUs per node, requested in tests: 1

Number of nodes detected: 1
Number of MPI procs, requested in tests: 4
Number of MPI procs per node: 4

******************************************  Ending Precheck   ******************************************

Constructing a list of tests
Done constructing a list of tests
Updating test list for fixtures
Added 0 tests to meet fixture requirements
Checking test dependency graph...
Checking test dependency graph end
test 13
    Start 13: par_for loops

13: Test command: /vast/home/lfroberts/parthenon/build-temp/tst/unit/unit_tests "par_for loops"
13: Working Directory: /vast/home/lfroberts/parthenon/build-temp/tst/unit
13: Test timeout computed to be: 1500
13: Filters: par_for loops
13: 
13: ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
13: unit_tests is a Catch v2.13.8 host application.
13: Run with -? for options
13: 
13: -------------------------------------------------------------------------------
13: par_for loops
13:   1D loops
13:   LoopPatternFlatRange
13:   IndexRange launch bounds
13: -------------------------------------------------------------------------------
13: /vast/home/lfroberts/parthenon/tst/unit/kokkos_abstraction.cpp:222
13: ...............................................................................
13: 
13: /vast/home/lfroberts/parthenon/tst/unit/kokkos_abstraction.cpp:222: FAILED:
13: due to unexpected exception with message:
13:   cudaFuncGetAttributes(&attr, func) error( cudaErrorInvalidDeviceFunction):
13:   invalid device function /vast/home/lfroberts/parthenon/external/Kokkos/core/
13:   src/Cuda/Kokkos_Cuda_KernelLaunch.hpp:140
13: 
13: -------------------------------------------------------------------------------
13: par_for loops
13:   2D loops
13:   LoopPatternFlatRange
13:   IndexRange launch bounds
13: -------------------------------------------------------------------------------
13: /vast/home/lfroberts/parthenon/tst/unit/kokkos_abstraction.cpp:222
13: ...............................................................................
13: 
13: /vast/home/lfroberts/parthenon/tst/unit/kokkos_abstraction.cpp:222: FAILED:
13: due to unexpected exception with message:
13:   cudaFuncGetAttributes(&attr, func) error( cudaErrorInvalidDeviceFunction):
13:   invalid device function /vast/home/lfroberts/parthenon/external/Kokkos/core/
13:   src/Cuda/Kokkos_Cuda_KernelLaunch.hpp:140
13: 
13: -------------------------------------------------------------------------------
13: par_for loops
13:   2D loops
13:   LoopPatternMDRange
13:   IndexRange launch bounds
13: -------------------------------------------------------------------------------
13: /vast/home/lfroberts/parthenon/tst/unit/kokkos_abstraction.cpp:222
13: ...............................................................................
13: 
13: /vast/home/lfroberts/parthenon/tst/unit/kokkos_abstraction.cpp:227: FAILED:
13:   REQUIRE( par_for_comp<Rank>(Sequence<Rank - 1>()) == true )
13: with expansion:
13:   false == true
13: 
13: -------------------------------------------------------------------------------
13: par_for loops
13:   3D loops
13:   LoopPatternFlatRange
13:   IndexRange launch bounds
13: -------------------------------------------------------------------------------
13: /vast/home/lfroberts/parthenon/tst/unit/kokkos_abstraction.cpp:222
13: ...............................................................................
13: 
13: /vast/home/lfroberts/parthenon/tst/unit/kokkos_abstraction.cpp:222: FAILED:
13: due to unexpected exception with message:
13:   cudaFuncGetAttributes(&attr, func) error( cudaErrorInvalidDeviceFunction):
13:   invalid device function /vast/home/lfroberts/parthenon/external/Kokkos/core/
13:   src/Cuda/Kokkos_Cuda_KernelLaunch.hpp:140
13: 
13: -------------------------------------------------------------------------------
13: par_for loops
13:   3D loops
13:   LoopPatternMDRange
13:   IndexRange launch bounds
13: -------------------------------------------------------------------------------
13: /vast/home/lfroberts/parthenon/tst/unit/kokkos_abstraction.cpp:222
13: ...............................................................................
13: 
13: /vast/home/lfroberts/parthenon/tst/unit/kokkos_abstraction.cpp:227: FAILED:
13:   REQUIRE( par_for_comp<Rank>(Sequence<Rank - 1>()) == true )
13: with expansion:
13:   false == true
13: 
13: -------------------------------------------------------------------------------
13: par_for loops
13:   3D loops
13:   LoopPatternTPTTRTVR
13:   IndexRange launch bounds
13: -------------------------------------------------------------------------------
13: /vast/home/lfroberts/parthenon/tst/unit/kokkos_abstraction.cpp:222
13: ...............................................................................
13: 
13: /vast/home/lfroberts/parthenon/tst/unit/kokkos_abstraction.cpp:222: FAILED:
13: due to unexpected exception with message:
13:   cudaFuncGetAttributes(&attr, func) error( cudaErrorInvalidDeviceFunction):
13:   invalid device function /vast/home/lfroberts/parthenon/external/Kokkos/core/
13:   src/Cuda/Kokkos_Cuda_KernelLaunch.hpp:140
13: 
13: -------------------------------------------------------------------------------
13: par_for loops
13:   3D loops
13: -------------------------------------------------------------------------------
13: /vast/home/lfroberts/parthenon/tst/unit/kokkos_abstraction.cpp:272
13: ...............................................................................
13: 
13: /vast/home/lfroberts/parthenon/tst/unit/kokkos_abstraction.cpp:272: FAILED:
13: due to a fatal error condition:
13:   SIGSEGV - Segmentation violation signal
13: 
13: ===============================================================================
13: test cases:  1 | 1 failed
13: assertions: 13 | 6 passed | 7 failed
13: 
1/2 Test #13: par_for loops ....................***Exception: SegFault  1.21 sec
```

acreyes:
I think this error:
```
13:   cudaFuncGetAttributes(&attr, func) error( cudaErrorInvalidDeviceFunction):
13:   invalid device function /vast/home/lfroberts/parthenon/external/Kokkos/core/
13:   src/Cuda/Kokkos_Cuda_KernelLaunch.hpp:140
```
comes from compiling for the wrong architecture, at least thats what the [cuda api](https://docs.nvidia.com/cuda/archive/11.4.2/cuda-runtime-api/group__CUDART__TYPES.html#group__CUDART__TYPES) says.

For what it's worth I've been testing on a V100 with cuda 12.4 & gcc 11.4

lroberts36:
I am also running on a V100 with the build command 
```bash
cmake -DKokkos_ENABLE_CUDA=ON -DKokkos_ARCH_VOLTA70=On -DCMAKE_CXX_COMPILER=$PWD/../external/Kokkos/bin/nvcc_wrapper ...
```
which I think should be the correct architecture. FWIW, I saw similar errors when I was debugging #1156.

Edit: I think this is an issue between CUDA versions. Compiling with the modules `1) nvhpc/24.7   2) cuda/12.3.1` and `GNU 8.5.0` but using the same `cmake` command above the tests seem to pass.

acreyes:
It looks like the CI is using 11.6(?) and seems to work like you said. This seems consistent with this [nvidia thread](https://forums.developer.nvidia.com/t/cudaerrorinvaliddevicefunction/228883) that reports a similar issue that goes away in 11.7. It sounds like you were able to work around this in #1156, with standalone functions if I recall. That can probably be replicated here.

lroberts36:
> It sounds like you were able to work around this in #1156, with standalone functions if I recall. That can probably be replicated here.

It was switching to standalone functions for the loops (which also ended up looking cleaner in #1156 anyway) and/or moving to single parameter packs.

lroberts36:
The other thing with Cuda 11.4 is I get a huge number of compiler warnings about having no return from functions that only return from `if constexpr(...) {... return a;} else {return b;}` blocks. These go away for newer versions.

tst/unit/kokkos_abstraction.cpp
```diff
@@ -33,238 +39,293 @@ using parthenon::ParArray3D;
...
+        dev(std::forward<decltype(args)>(args)...) +=
+            increment_data(std::forward<decltype(args)>(args)...);
+      };
+      if constexpr (bound_type == lbounds::integer) {
+        parthenon::par_for(Pattern(), "unit test ND integer bounds", exec_space,
+                           int_bounds[Ids]..., functor);
+      } else {
+        parthenon::par_for(Pattern(), "unit test ND IndexRange bounds", exec_space,
+                           bounds[Ids]..., functor);
+      }
```
acreyes: this was the culprit for the 
```
error( cudaErrorInvalidDeviceFunction)
```
errors. For whatever reason having the two `KOKKOS_CLASS_LAMBDA`s in the two `if constexpr` branches here was causing errors. Oddly enough I'm doing the same thing in the `par_reduce` tests and it works fine...

acreyes:
I was able to recreate the issues with the `nvidia/cuda:11.4.3-devel-ubuntu20.04` docker container. There weren't any problems with the `par_dispatch` but instead it was a `KOKKOS_CLASS_LAMBDA` problem in the tests. All of the missing return warnings should be resolved now too.

pgrete:
Out of curiosity: Is anyone working on machines/cluster where Cuda >= 11.7 is not available?
Even with 11.6 there are bugs in the compiler (capturing function parameters).
So I'm wondering from the point of view of bumping our cuda required version.

src/solvers/solver_utils.hpp
```diff
@@ -247,9 +247,8 @@ TaskStatus SetToZero(const std::shared_ptr<MeshData<Real>> &md) {
         IndexRange kb = cb.GetBoundsK(IndexDomain::interior, te);
         const int nvars = pack.GetUpperBound(b, var()) - pack.GetLowerBound(b, var()) + 1;
         for (int c = 0; c < nvars; ++c) {
-          parthenon::par_for_inner(
-              parthenon::inner_loop_pattern_simdfor_tag, member, kb.s, kb.e, jb.s, jb.e,
-              ib.s, ib.e,
+          parthenon::seq_for(
+              kb.s, kb.e, jb.s, jb.e, ib.s, ib.e,
```
lroberts36: Ah, this looks like a potential bug in the original code. I think this should be `PARTHENON_DEFAULT_INNER_LOOP_PATTERN`.

acreyes: Ha, that makes a lot more sense. Maybe its better to fix it in a separate PR

lroberts36: I may have done this because the right inner loop pattern didn't always exist previously for the default.

src/kokkos_abstraction.hpp
```diff
@@ -20,20 +20,24 @@
...
+#include "impl/Kokkos_Tools_Generic.hpp"
+#include "loop_bounds.hpp"
 #include "parthenon_array_generic.hpp"
-#include "utils/error_checking.hpp"
+#include "utils/concepts_lite.hpp"
+#include "utils/indexer.hpp"
 #include "utils/instrument.hpp"
 #include "utils/multi_pointer.hpp"
 #include "utils/object_pool.hpp"
+#include "utils/type_list.hpp"
```
lroberts36, line 38: This PR might be a good place to split all of the `using` statements below out into a separate file, e.g. `kokkos_types.hpp`.

src/kokkos_abstraction.hpp
```diff
@@ -214,476 +296,374 @@ inline void kokkos_dispatch(ParallelScanDispatch, Args &&...args) {
...
+  static constexpr std::size_t Rank = Translator::Rank;
+
+  using TeamPattern =
+      LoopPatternTeam<Pattern,
+                      std::integral_constant<size_t, Rank>>; // false_type unless we use
+                                                             // an outer team policy
+  static constexpr bool is_ParFor =
+      std::is_same<Tag, dispatch_impl::ParallelForDispatch>::value;
+  static constexpr bool is_ParScan =
+      std::is_same<Tag, dispatch_impl::ParallelScanDispatch()>::value;
```
lroberts36: Why are there parenthesis after `ParallelScanDispatch` and not after `ParallelForDispatch`?

src/kokkos_abstraction.hpp
```diff
@@ -168,17 +172,79 @@ static struct LoopPatternTPTVR {
...
+// thread/vecctor
+static struct LoopPatternTeamCollapse {
+} loop_pattern_team_collapse;
 // Used to catch undefined behavior as it results in throwing an error
 static struct LoopPatternUndefined {
 } loop_pattern_undefined_tag;
+// Translates to a Kokkos::TeamPolicy that collapse Nteams outer loops
+// with Nthread & Nvector inner loop collapses
+template <size_t Nteam, std::size_t Nthread, std::size_t Nvector>
+struct LoopPatternCollapse {};
```
lroberts36: Is `LoopPatternTPTTR` just `LoopPatternCollapse<N - 1, 1, 0>`?

src/kokkos_abstraction.hpp
```diff
@@ -168,17 +172,79 @@ static struct LoopPatternTPTVR {
...
+
+// trait to track if pattern requests any type of hierarchial parallelism
+template <typename Pattern, typename, typename T = void>
+struct LoopPatternTeam : std::false_type {
```
lroberts36: I would suggest using a different name for this type trait. At first glance it appears that this is another kind of loop pattern that a user can define, rather than a template class for getting the properties of some other loop pattern. 

Personally, I like having the loop pattern types themselves just contain this information, e.g. 
```c++
static struct LoopPatternTPTVR {
  static constexpr bool is_team_pattern = true;
  static constexpr std::size_t Nvector = 1;
  static constexpr std::size_t Nthread = 0; 
} loop_pattern_tptvr_tag;
``` 
I think this gets you the same information as the type traits without a lot of extra template specializations, but maybe I am missing something.

src/kokkos_abstraction.hpp
lroberts36: Actually, why do you want to be able to specify `Nteam` in the loop pattern? Wouldn't it work just as well to specify `Nthread` and `Nvector` and then figure out `Nteam` from the rank of the loop?

src/kokkos_abstraction.hpp
```diff
@@ -214,476 +296,374 @@ inline void kokkos_dispatch(ParallelScanDispatch, Args &&...args) {
...
+  static constexpr std::size_t Rank = Translator::Rank;
+  using Function = typename TL::template type<func_idx>;
+  using Args = typename TL::template continuous_sublist<func_idx + 1>;
+};
 
-template <typename Function, typename R, typename T, typename Index, typename... FArgs>
-class FlatFunctor<Function, R (T::*)(Index, Index, Index, FArgs...) const> {
-  int NjNi, Ni, kl, jl, il;
-  Function function;
+enum class LoopPattern { flat, md, simd, outer, collapse, undef };
```
lroberts36: It would be nice to have a comment here briefly describing what each of the enumerated loop patterns is. 

Also, I think the name `*LoopPattern*` is maybe overused overused since there are the original `LoopPatternMDRange` etc., the new `LoopPatternTeam` and `LoopPatternCollapse`, and the `LoopPattern` enum. It may be confusing that the `LoopPattern` enum doesn't have values that correspond to the loop patterns above.

src/kokkos_abstraction.hpp
```diff
@@ -214,476 +296,374 @@ inline void kokkos_dispatch(ParallelScanDispatch, Args &&...args) {
...
-        k += kl;
-        Kokkos::parallel_for(Kokkos::ThreadVectorRange<>(team_member, il, iu + 1),
-                             [&](const int i) { function(n, k, j, i); });
-      });
-}
+  template <std::size_t... Is>
+  using sequence = std::integer_sequence<std::size_t, Is...>;
+
+  template <typename ExecSpace, std::size_t... OuterIs, std::size_t... InnerIs>
+  inline void dispatch_impl(LoopPatternTag<LP::flat>, sequence<OuterIs...>,
```
lroberts36: I would add a comment before each of these `dispatch_impl` overloads saying which is which. I know the `LoopPatternTag<...>` is there, but I think that is hard to see if you don't know what you are looking for.

src/kokkos_abstraction.hpp
```diff
@@ -214,476 +296,374 @@ inline void kokkos_dispatch(ParallelScanDispatch, Args &&...args) {
...
+            !always_false<Tag, Pattern>,
+        "Loop pattern & tag combination not recognized in DispatchType::GetPatternTag");
+
+    constexpr bool isSimdFor =
+        std::is_same_v<LoopPatternSimdFor, base_type<decltype(pattern_tag)>>;
+    auto bound_arr = bound_translator().GetIndexRanges(std::forward<Bounds>(bounds)...);
+    if constexpr (isSimdFor) {
+      static_assert(!isSimdFor || (isSimdFor && std::is_same_v<ExecSpace, HostExecSpace>),
+                    "SimdFor pattern only supported in HostExecSpace");
+      SimdFor(std::make_index_sequence<Rank - 1>(), function, bound_arr);
```
lroberts36: Is it possible to have a `dispatch_impl` overload for the simd for pattern instead of the `if` block here, just so that everything is uniform?

src/kokkos_abstraction.hpp
```diff
@@ -214,476 +296,374 @@ inline void kokkos_dispatch(ParallelScanDispatch, Args &&...args) {
...
-      });
-}
+template <class Function, class... Bounds>
+struct seq_for_impl<Function, TypeList<Bounds...>> {
+  KOKKOS_INLINE_FUNCTION void execute(Bounds &&...bounds, Function function) {
+    using bound_translator = LoopBoundTranslator<Bounds...>;
+    constexpr std::size_t Rank = bound_translator::Rank;
+    const auto bound_arr =
+        bound_translator().GetIndexRanges(std::forward<Bounds>(bounds)...);
+    SimdFor(std::make_index_sequence<Rank - 1>(), function, bound_arr);
```
lroberts36: I think we should have a `SequentialFor(std::make_index_sequence<Rank - 1>(), function, bound_arr)` call here instead of `SimdFor(...)`.

src/utils/indexer.hpp
```diff
@@ -97,18 +103,27 @@ struct Indexer {
...
   auto GetIdxArray(int idx) const {
-    return get_array_from_tuple(
-        GetIndicesImpl(idx, std::make_index_sequence<sizeof...(Ts)>()));
+    return GetIndicesArrayImpl(idx, std::make_index_sequence<sizeof...(Ts)>());
+  }
+
+  KOKKOS_FORCEINLINE_FUNCTION
+  auto GetIdxKArray(int idx) const {
+    return GetIndicesKArrayImpl(idx, std::make_index_sequence<sizeof...(Ts)>());
+  }
```
lroberts36: I would just call this `GetIdxArray` and have it always return a kokkos array. I don't think there is anywhere in the code that that won't work (although maybe there will be somewhere the needs to be changed from `std::array` to `auto`).

lroberts36 (commented):
Thanks for making this work on for older versions of CUDA. Another round of comments below.

src/kokkos_abstraction.hpp
acreyes: True. Originally I was inspired by the openACC `#pragma acc loop gang/worker/vector collapse(n)` directives. I thought it might be useful to be able to specify exactly how to distribute the work for a specific loop, e.g.
```cpp
parthenon::par_for(LoopPatternCollapse<3,0,4>(),  "kernel", exec_space, ...)
```
and from that view all the `TP***` patterns are just special cases. But I also don't have use case that would justify supporting it.

lroberts36: Wouldn't you get the same thing from
```c++
template <std::size_t NMiddle, std::size_t NInner> 
struct LoopPatternCollapse {
  static constexpr bool is_team_pattern = true;
  static constexpr std::size_t Nvector = NInner;
  static constexpr std::size_t Nthread = NMiddle;
};

using LoopPatternTPTVR = LoopPatternCollapse<0, 1>;
constexpr auto loop_pattern_tptvr_tag = LoopPatternTPTVR();

using LoopPatternTPTTR = LoopPatternCollapse<1, 0>;
constexpr auto loop_pattern_tpttr_tag = LoopPatternTPTTR();

using LoopPatternTPTTRVR = LoopPatternCollapse<1, 1>;
constexpr auto loop_pattern_tpttrvr_tag = LoopPatternTPTTRVR();

Comment (acreyes) [reply src/kokkos_abstraction.hpp]: I see what you mean now. Yes that is a lot neater

File: src/kokkos_abstraction.hpp (hunk 11 above)
Comment (acreyes) [reply src/kokkos_abstraction.hpp]: oops my LSP autocomplete probably snuck that in there. Thanks!

File: src/kokkos_abstraction.hpp (hunk 12 above)
Comment (acreyes) [reply src/kokkos_abstraction.hpp]: thanks, its a lot cleaner this way

File: src/kokkos_abstraction.hpp (hunk 16 above)
Comment (acreyes) [reply src/kokkos_abstraction.hpp]: I also did the same for the inner loops

File: src/kokkos_abstraction.hpp (hunk 14 above)
Comment (acreyes) [reply src/kokkos_abstraction.hpp]: I changed it to `PatternTag` and changed `LoopPatternTeam` to `UsesHierarchialPar` to be more explicit about what that trait is being used for.

File: src/utils/indexer.hpp (hunk 18 above)
Comment (acreyes) [reply src/utils/indexer.hpp]: There were a handful of places across a few files that needed updating. I only changed the ones the compiler needed, but there is some mixed usage now in those files/classes that are now using both `std::array` & `Kokkos::Array`. Not sure if people have strong feelings about that, but I was worried it might start to creep to a lot more code if I started pulling on that thread more.

Comment (lroberts36) [reply src/utils/indexer.hpp]: Maybe the solution (for another PR) is to define `template <class T, std::size_t N> using array = Kokkos::array<T, N>` in `kokkos_types.hpp` and search and replace `std::array` with `array` everywhere in the code (since I believe the kokkos array should be a drop in replacement).

Comment (acreyes) [reply src/utils/indexer.hpp]: I think it's missing the iterator stuff that is in `std::array`

Comment (lroberts36) [reply src/utils/indexer.hpp]: good point, although I don't think we are using `std::array` iterators anywhere (could be wrong though).

File: src/bvals/neighbor_block.cpp [hunk 19]
```diff
@@ -42,7 +42,7 @@ NeighborBlock::NeighborBlock()
       offsets(0, 0, 0), ownership(true) {}
 
 NeighborBlock::NeighborBlock(Mesh *mesh, LogicalLocation loc, LogicalLocation origin_loc,
-                             int rank, int gid, std::array<int, 3> offsets_in, int bid,
+                             int rank, int gid, Kokkos::Array<int, 3> offsets_in, int bid,
```
Yurlungur, line 46: No objections to this, but is this necessary? I thought `std::array` was mostly constexpr and worked on device?

src/kokkos_types.hpp
Yurlungur, line 1: :+1: for splitting this out.

src/loop_bounds.hpp
```diff
@@ -0,0 +1,135 @@
...
+      } else if constexpr (std::is_integral_v<base_type<Bnd0>>) {
+        using Bnd1 = typename TL::template type<1>;
+        static_assert(std::is_integral_v<base_type<Bnd1>>,
+                      "integer launch bounds need to come in (start, end) pairs");
+        return 2 + GetNumBounds(typename TL::template continuous_sublist<2>());
+      }
+    }
+    // should never get here but makes older cuda compilers happy
+    return 0;
+  }
```
Yurlungur: I was able to unpack waht this is doing by staring at it, but it's pretty opaque. Might be worth an explanatory comment.

src/loop_bounds.hpp
```diff
@@ -0,0 +1,135 @@
...
+
+template <class... Bound_ts>
+struct LoopBoundTranslator<TypeList<Bound_ts...>>
+    : public LoopBoundTranslator<Bound_ts...> {};
+
+template <class F, class = void>
+struct is_functor : std::false_type {};
+
+template <class F>
+struct is_functor<F, void_t<decltype(&F::operator())>> : std::true_type {};
```
Yurlungur: should this be moved into `utils/concepts_lite`? It might be more broadly useful in concepts land.

src/kokkos_abstraction.hpp
```diff
@@ -156,42 +53,79 @@ static struct LoopPatternFlatRange {
 // a 1:1 indices matching
 static struct LoopPatternMDRange {
 } loop_pattern_mdrange_tag;
+// Translates to a Kokkos::TeamPolicy that collapse Nthread & Nvector inner loop collapses
+template <std::size_t num_thread, std::size_t num_vector>
+struct LoopPatternCollapse : std::true_type {
```
Yurlungur: what's going on with this inheritence?

Also I don't hate it, but `LoopPatternCollapse` isn't a very descriptive name. Maybe something like `LoopPatternThreadsVecs`?

src/kokkos_abstraction.hpp
```diff
@@ -156,42 +53,79 @@ static struct LoopPatternFlatRange {
...
+  static constexpr std::size_t Nthread = num_thread;
+  static constexpr std::size_t Nvector = num_vector;
+};
+
 // Translates to a Kokkos::TeamPolicy with a single inner
 // Kokkos::TeamThreadRange
-static struct LoopPatternTPTTR {
-} loop_pattern_tpttr_tag;
+using LoopPatternTPTTR = LoopPatternCollapse<1, 0>;
+constexpr auto loop_pattern_tpttr_tag = LoopPatternTPTTR();
```
Yurlungur: oh I like this.

tst/unit/kokkos_abstraction.cpp
```diff
@@ -33,405 +41,513 @@ using parthenon::ParArray3D;
...
+template <std::size_t Ni>
+using Sequence = std::make_index_sequence<Ni>;
+
+template <std::size_t>
+struct ParArrayND_impl {};
+template <>
+struct ParArrayND_impl<0> {
+  template <typename T>
+  using type = parthenon::ParArray0D<T>;
+};
```
Yurlungur: It might be worth finding a name for this and moving it into the source. You might not be the first one to want the ability to select the dimension via a `size_t`.

tst/unit/kokkos_abstraction.cpp
Yurlungur, line 1: Just for explanatory power. I think this test file should contain some calls to `par_*` the way a user would actually call the code, rather than unpacked from fold expressions. I at least frequently look in the source file or test code to see how to call a method. And both the source and tests are totally useless for that as they're written right now.

Yurlungur (changes requested):
Thanks @acreyes  and @lroberts36 ! This is much easier to read than previously and a big improvement over previous functionality. 

I still have some readability concerns though... I think they could be addressed by some of the comments below and additionally by adding some documentation to our docs. I think we should ensure the docs we have on `par_for` are still accurate. And we should also add a description of how to add a new loop pattern if desired.

src/bvals/neighbor_block.cpp
Yurlungur: Nevermind I saw the discussion below. :+1:

acreyes:
> Thanks @acreyes and @lroberts36 ! This is much easier to read than previously and a big improvement over previous functionality.
> 
> I still have some readability concerns though... I think they could be addressed by some of the comments below and additionally by adding some documentation to our docs. I think we should ensure the docs we have on `par_for` are still accurate. And we should also add a description of how to add a new loop pattern if desired.

I added some new docs for `par_for` that describes the loop pattern tagging as well as how to add new patterns.
The only previous docs I could find are in the developing guide and focuses on `mb->par_for`.

tst/unit/kokkos_abstraction.cpp
acreyes: I've added a `TEST_CASE("DEFAULT loop patterns", "[default]")` at the bottom that does this for `par_for`, `par_for_outer` & `par_reduce` on 3D loops.

src/kokkos_abstraction.hpp
acreyes: Thanks for catching that. It was left over from a previous implementation, and is now handled by the `UsesHierarchialPar` trait.

I've changed it to `LoopPatternTeamThreadVec`

src/loop_bounds.hpp
acreyes: I added some comments and split the functionality into separate types for each of the loop bound types so it's more explicit about what each of the cases are doing.

tst/unit/kokkos_abstraction.cpp
acreyes: There is a function buried in there for getting one of these that could be what actually gets exposed.
```cpp
template <std::size_t ND, typename T, typename... Args>
auto ParArrayND(Args &&...args) {
  static_assert(ND <= 8, "ParArrayND supoorted up to ND=8");
  return typename ParArrayND_impl<ND>::template type<T>(std::forward<Args>(args)...);
}
```

Yurlungur (approved):
Oops didn't realize our docs on patterns was so sparse! Thanks for filling that in. My other comments were minor nitpicks. From my perspective this is good to go.

tst/unit/kokkos_abstraction.cpp
Yurlungur: Ah I see. Yes maybe worth naming that internal type something more useful so we can expose it if we want to. `ParArray` without the `ND` maybe... or something like that.

pgrete:
I'm rerunning the Cuda test as it failed with some (unexpected) host to device mem copies.

pgrete:
> I'm rerunning the Cuda test as it failed with some (unexpected) host to device mem copies.

The test repeatedly failed. Any idea where those extra copies come from?

acreyes:
> > I'm rerunning the Cuda test as it failed with some (unexpected) host to device mem copies.
> 
> The test repeatedly failed. Any idea where those extra copies come from?

I'll check, but that is unexpected. I think it passed some time back in september so at least there should be a pretty recent diff to use

doc/sphinx/src/par_for.rst
```diff
@@ -0,0 +1,132 @@
...
+   :widths: 40 60
+   :header-rows: 1
+
+   * - Tag
+     - Execution Policy
+   * - ``loop_pattern_flatrange_tag``
+     - Flattens all of the loops into a single ``Kokkos::RangePolicy``
+   * - ``loop_pattern_simdfor_tag``
+     - Maps to two C-style loops. The innermost gets decorated with a ``#pragma omp simd`` and the remaining
+       loops are flattened into a single C-style for looop. Only supported on CPU.
```
pgrete:
```suggestion
       loops are flattened into a single C-style for loop. Only supported on CPU.
```

doc/sphinx/src/par_for.rst
```diff
@@ -0,0 +1,132 @@
...
+   * - ``loop_pattern_mdrange_tag``
+     - Maps all the loop bounds onto a ``Kokkos::MDRangePolicy``
+   * - ``LoopPatternTeamThreadVec<Nt, Nv>()``
+     - Maps onto a hierarchial parrallel loop. The ``Nv`` inner loops are flattened onto a ``VectorRange`` policy,
```
pgrete:
```suggestion
     - Maps onto a hierarchical parallel loop. The ``Nv`` inner loops are flattened onto a ``VectorRange`` policy,
```

src/kokkos_abstraction.hpp
```diff
@@ -171,42 +53,81 @@ static struct LoopPatternFlatRange {
...
+    : std::true_type {
+  static constexpr std::size_t Nthread = num_thread;
+  static constexpr std::size_t Nvector = num_vector;
+};
+
+template <>
+struct UsesHierarchialPar<OuterLoopPatternTeams> : std::true_type {
+  static constexpr std::size_t Nvector = 0;
+  static constexpr std::size_t Nthread = 0;
+};
```
pgrete: Can you comment on this trait?
I'm not sure I follow the default values for `Nvector` and `Nthread`.

src/kokkos_abstraction.hpp
```diff
@@ -229,476 +150,451 @@ inline void kokkos_dispatch(ParallelScanDispatch, Args &&...args) {
...
+struct DispatchType {
+  using Translator = LoopBoundTranslator<Bounds...>;
+  static constexpr std::size_t Rank = Translator::Rank;
+
+  using HierarchialPar = UsesHierarchialPar<Pattern>;
+
+  static constexpr bool is_ParFor =
+      std::is_same<Tag, dispatch_impl::ParallelForDispatch>::value;
+  static constexpr bool is_ParScan =
+      std::is_same<Tag, dispatch_impl::ParallelScanDispatch>::value;
```
pgrete: No special handling/logic required for par_reduces below?

src/kokkos_abstraction.hpp
```diff
@@ -229,476 +150,451 @@ inline void kokkos_dispatch(ParallelScanDispatch, Args &&...args) {
...
+  inline void dispatch_impl(LoopPatternTag<PT::md>, sequence<OuterIs...>,
+                            sequence<InnerIs...>, std::string name, ExecSpace exec_space,
+                            Kokkos::Array<IndexRange, Rank> bound_arr, Function function,
+                            Args &&...args, const int scratch_level,
+                            const std::size_t scratch_size_in_bytes) {
+    static_assert(sizeof...(InnerIs) == 0);
+    kokkos_dispatch(
+        Tag(), name,
+        Kokkos::MDRangePolicy<Kokkos::Rank<Rank>>(exec_space, {bound_arr[OuterIs].s...},
+                                                  {(1 + bound_arr[OuterIs].e)...}),
```
pgrete: Are the block sizes here forwarded (i.e., the old `{1, 1, 1, 1, iu + 1 - il}`)?

src/kokkos_abstraction.hpp
```diff
@@ -229,476 +150,451 @@ inline void kokkos_dispatch(ParallelScanDispatch, Args &&...args) {
...
-    int k = (idx - n * NkNjNi) / NjNi;
-    int j = (idx - n * NkNjNi - k * NjNi) / Ni;
-    int i = idx - n * NkNjNi - k * NjNi - j * Ni;
-    n += nl;
-    k += kl;
-    j += jl;
-    i += il;
-    function(n, k, j, i, std::forward<FArgs>(fargs)...);
+// builds a functor that uses inner hierarchial parrallelism used by both par_disp_inner &
+// par_dipsatch for LoopPatternCollapse
```
pgrete:
```suggestion
// par_dispatch for LoopPatternCollapse
```

src/kokkos_abstraction.hpp
```diff
@@ -229,476 +150,451 @@ inline void kokkos_dispatch(ParallelScanDispatch, Args &&...args) {
...
+                            sequence<InnerIs...>, std::string name, ExecSpace exec_space,
+                            Kokkos::Array<IndexRange, Rank> bound_arr, Function function,
+                            Args &&...args, const int scratch_level,
+                            const std::size_t scratch_size_in_bytes) {
+    const auto idxer =
+        MakeIndexer(Kokkos::Array<IndexRange, sizeof...(OuterIs)>{bound_arr[OuterIs]...});
+    using HierarchialPar = typename dispatch_type::HierarchialPar;
+    constexpr std::size_t Nvector = HierarchialPar::Nvector;
+    constexpr std::size_t Nthread = HierarchialPar::Nthread;
+    constexpr std::size_t Nouter = Rank - Nvector - Nthread;
```
pgrete, line 541: What exactly is `Nouter` here (given the `Rank - Nvector - Nthread` formula)?
I also tried to follow the ` MakeCollapse<Rank, Nouter` trail below but I `Nouter` seems to become `Nteam` which is then not used anymore.
I'm probably missing sth here.

src/kokkos_types.hpp
```diff
@@ -0,0 +1,252 @@
...
+
+template <typename ND>
+struct ScratchPadND_impl {
+  static_assert(ND::value <= 6, "ScratchPad only supported up to ND=6");
+};
+
+template <std::size_t ND, typename T, typename State = empty_state_t>
+using ParArray = typename ParArrayND_impl<std::integral_constant<std::size_t, ND>,
+                                          State>::template type<T>;
+
```
pgrete, line 165: and below: this is a new interface, isn't it? Might be worth to briefly add this to the doc along the ParArray#D.

pgrete (commented):
I really like this! Thanks for putting in all the effort.

I have to admit that mentally parsing the new machinery is more challenging than the old verbose one, but I think it's a way cleaner approach!
I'd like to do some downstream performance testing early next week and understand/track down the additional host/device copies before I finally approve.

src/kokkos_types.hpp
acreyes: yes, I've added some new documentation for it

src/kokkos_abstraction.hpp
acreyes: This one covers all the various `TPT[RTV]R` patterns. These are always 1 or 0 loops for either the vector or thread range loops in the inner pattern. `Nouter` is all the remaining loops that become flattened into an outer team policy loop.

src/kokkos_abstraction.hpp
acreyes: good catch, I think I had meant to come back and add this in and forgot. This is in now

src/kokkos_abstraction.hpp
acreyes: `par_reduce` should work with every pattern except for the `SimdFor` one. However `SimdFor` only works for `par_for`, which is why that is the only check that is done.

edit: maybe I take that back. The `TPT[RTV]R` I think could in principle work with `par_reduce` but certainly not the way they're currently written. I'll duplicate the `SimdFor` check for the `Hierarchical` ones

src/kokkos_abstraction.hpp
acreyes: The `par_dispatch_impl::dispatch_impl` tagged dispatches  abstract the kernel launch over an outer flattened loop and an inner flattened loop.  The inner loop flattening is used in the `TPT[RTV]R` patterns (hence the thread/vector) and also in the `SimdFor` pattern for the innermost vectorized loop.  The default then is zero for all those loop patterns that don't have any vector/thread inner loops.

doc/sphinx/src/par_for.rst
acreyes: ha, I seem to have made this typo repeatedly....

acreyes:
> I'd like to do some downstream performance testing early next week and understand/track down the additional host/device copies before I finally approve.

👍

I believe I've tracked down the source of the HtoD copies. The `Indexer` struct is used to flatten/reconstruct the multidimensional indices and holds some `Kokkos::Array<int.ND>`s for that purpose. For some reason the lambda capture of this guy triggers a mem copy.  It can be constructed inside the kernel instead and that seems to solve it.

I don't understand the behavior though, and it also seems to be related to the Kokkos version. 4.0.1 doesn't have the copies, but starting at least in 4.2 the copies show up.

Even stranger the same pattern is used for the `LoopPatternFlatRange` kernels, but doesn't result in any mem copies, at least according to Nsight.

src/kokkos_abstraction.hpp
pgrete: I see. This makes sense (I was probably confused because I associated sth different based on the naming and personal habits).

src/kokkos_abstraction.hpp
```diff
@@ -486,11 +489,16 @@ struct par_dispatch_impl<Tag, Pattern, Function, TypeList<Bounds...>, TypeList<A
                             Args &&...args, const int scratch_level,
                             const std::size_t scratch_size_in_bytes) {
     static_assert(sizeof...(InnerIs) == 0);
-    kokkos_dispatch(
-        Tag(), name,
-        Kokkos::MDRangePolicy<Kokkos::Rank<Rank>>(exec_space, {bound_arr[OuterIs].s...},
-                                                  {(1 + bound_arr[OuterIs].e)...}),
-        function, std::forward<Args>(args)...);
+    constexpr std::size_t Nouter = sizeof...(OuterIs);
+    Kokkos::Array<int, Nouter> tiling{(OuterIs, 1)...};
```
pgrete: Is this working as expected?
If I infer the intent correctly, this should create an array initialized to `1` everywhere.
My compiler complains with a warning 
```
/p/project/coldcluster/pgrete/athenapk/external/parthenon/src/kokkos_abstraction.hpp(493): warning #174-D: expression has no effect
      Kokkos::Array<int, Nouter> tiling{(OuterIs, 1)...};
```

AFAIK default init doesn't work for arrays, so we might need sth like
```
    std::array<int, Nouter> tiling;
    tiling.fill(1);
```

pgrete:
I now ran some more detailed tests.
Compile times increase (as expected given the additional template magic), here tested for AthenaPK on 48 cores
- CUDA: 5m29s -> 6m7s
- HIP: 2m18s -> 2m40s
so nothing too dramatic (from my point of view).

However, performance is a concern.
I tested small(ish) and large blocks with 128x256x256 and 32x64x64 cells respectively on A100 and MI250X and our flux kernels (that use hierarchical parallelism with scratch memory) are up to 22% slower with the new layout (whereas the flat kernels remain effectively identical in performance).
I'm not exactly sure where this difference comes from but I suspect that the additional logic result in additional register usage, which limits the occupancy of the kernels.

Maybe we can discuss the performance implications during the sync today.

src/kokkos_abstraction.hpp
acreyes: This was working for me, but better to avoid the warning.

The warning makes sense since `(OuterIs, 1)` will always just evaluate to 1.

fglines-nv:
I looked into the performance issues in AthenaPK, I verified that there are performance issues in the flux kernels with this PR but only for the X1 flux, not X2 and X3. It's definitely due to increased register pressure.

| Kernel | Baseline Time | PR 1142 Time |
| ------- | --------------- | --------------- |
| x1 flux  | 1.75 ms          | 1.98 ms           |
| x2        | 1.42 ms          | 1.43 ms           |
| x3        | 1.19 ms          | 1.20 ms           |

| Kernel | Baseline Regs | PR 1142 Regs |
| ------- | --------------- | --------------- |
| x1 flux  | 76                  |  82                  |
| x2 flux  | 82                  | 85                   |
| x3 flux  | 83                  | 87                    |

That jump from 76->82 registers is enough to push the kernel from running 6 blocks per SM to 5 blocks per SM, hence fewer warps in the pipeline doing loads and thus the ~20% drop in performance. The x2 and x3 kernels could potentially gain 20% if a few registers could be optimized away.

You'd see this PR affect other high register kernels the same way, the higher the count the more they'd be impacted by this PR. Generally, the hierarchal kernels I've seen in Parthenon codes are more complex and high register count. Sometimes higher than this.

I'll take a look at the arithmetic in this PR and see if we can reduce register usage without changing the interface.

fglines-nv:
The merge from develop broke this PR. Currently trying to fix it

Yurlungur:
@acreyes @fglines-nv let us know when this is ready for discussion again

acreyes:
> The merge from develop broke this PR. Currently trying to fix it

Should be fixed now

>@acreyes @fglines-nv let us know when this is ready for discussion again

I removed some extra member variables from the `Indexer` class that we thought might have contributed to the extra register usage, but @fglines-nv saw that it didn't make any difference in the athenaPK kernel in question.

We have an attempt at a reproducer [here](https://github.com/acreyes/par_for_perf) to compare the same kernel with all combinations of the old/new abstractions but can't seem to reproduce the extra registers on simpler kernels. It would be worth discussing whether we want to just revert to the explicit overloads for the `par_for_inner`s

fglines-nv:
> Should be fixed now

Thanks @acreyes. I'm getting errors now:
```
:0: : block: [81,0,0], thread: [0,32,0] Assertion `Kokkos::Array: index 6 >= 6` failed.
:0: : block: [81,0,0], thread: [0,33,0] Assertion `Kokkos::Array: index 6 >= 6` failed.
:0: : block: [81,0,0], thread: [0,34,0] Assertion `Kokkos::Array: index 6 >= 6` failed.
:0: : block: [81,0,0], thread: [0,35,0] Assertion `Kokkos::Array: index 6 >= 6` failed.
...
```
It occurs with both `advection_test` and with AthenaPK.  I suspect it's coming from the kernel in `parthenon/src/bvals/comms/boundary_communication.cpp:95`.  Note you might need to compile with Debug symbols include this assertion. Care to take a look at is @acreyes if you have the time?


> We have an attempt at a reproducer [here](https://github.com/acreyes/par_for_perf) to compare the same kernel with all combinations of the old/new abstractions but can't seem to reproduce the extra registers on simpler kernels. It would be worth discussing whether we want to just revert to the explicit overloads for the `par_for_inner`s

I tried reproducing the extra register pressure by inserting all of the calculate fluxes code (EOS, Riemann solver, reconstruction) into your example. It did not work, unfortunately, the register pressure is still the same across the 4 test cases, although I was not able replicate a `VariablePack` in the simple example.

Once AthenaPK is working with the new commit, I can try the line-by-line approach like we discussed

acreyes:
> > Should be fixed now
> 
> Thanks @acreyes. I'm getting errors now:
> 
> ```
> :0: : block: [81,0,0], thread: [0,32,0] Assertion `Kokkos::Array: index 6 >= 6` failed.
> :0: : block: [81,0,0], thread: [0,33,0] Assertion `Kokkos::Array: index 6 >= 6` failed.
> :0: : block: [81,0,0], thread: [0,34,0] Assertion `Kokkos::Array: index 6 >= 6` failed.
> :0: : block: [81,0,0], thread: [0,35,0] Assertion `Kokkos::Array: index 6 >= 6` failed.
> ...
> ```
> 
> It occurs with both `advection_test` and with AthenaPK. I suspect it's coming from the kernel in `parthenon/src/bvals/comms/boundary_communication.cpp:95`. Note you might need to compile with Debug symbols include this assertion. Care to take a look at is @acreyes if you have the time?

thanks for catching that, should be fixed now. I was able to run the advection example

fglines-nv:
I did some testing
in AthenaPK of register pressure and timing with @acreyes' recent changes and
past PRs.The good news is that between
this PR and parthenon develop, there is only a small dip in performance from my
limited testing of GLMMD+PPM+HLLD. The other news is that all the changes
@acreyes made since last year haven't affected the register pressure in the
`CalculateFluxes` but register pressure has gone up and down in other kernels
in `hydro.cpp` within AthenaPK. Particularly Prolongation/restriction
operations (using hierarchal parallelism) that get compiled in
`hydro.cpp`

We didn't document which hydro+recon+Riemann we
were testing last year @pgrete. I tested the linear wave 3D
with 128^3 with glmmhd+ppm+hlld and got these results:

Parthenon Commit | Commit Description | Zone-cycles/s |  
-- | -- | -- | --
635698 | Current PR | 9.00e+07 | ProlongationRestriction   mostly less than 52983a, CalculateFluxes the same as 279b12
52983a | Merge with Develop | FAILS | ProlongationRestriction strictly increased   from 279b12, Same CalculateFluxes as 279b12
b34762 | Strip out extra   end&size | FAILS | Same as  279b12
279b12 | 2024 commit (last   testing) | 9.08e+07 | ProlongationRestriction/CalculateFluxes   ops use more/less from baseline
d82ec6 | Parthenon Develop   at time of testing | 9.12e+07 | No difference in   regs from baseline
a0757d | Parthenon in   AthenaPK main | 9.13e+07 | (Baseline)

AthenaPK Commit:
commit 9686f5c47211a5e329724f87b9d8b323fc68053e

Kokkos Commit:
commit 1b1383c6001f3bfe9fe309ca923c2d786600cc79 

I'm looking at ncu reports of 635698 vs d82ec6 line by line next

fglines-nv:
I did some exploration of the register pressure in different kernels in AthenaPK with and without this PR using `ncu`, attempting to compare the two commits line-by-line. The final take-away is due to the heuristics in how `nvcc` choses to use registers, it's a bit of a lost cause to chase after small register changes.  If we're concerned about register pressure to affect occupancy, we should try exposing `__launch_bounds__` (via `Kokkos::LaunchBounds`). A bit disappointing, but I think we should merge in this PR if we like the unification.

<details>
  <summary>Expand for way more details, large NCU screenshots</summary>

I did tests of the 3d linear wave with VL2 integration, Euler hydro, PLM reconstruction, HLLE Riemann solver on an H200 with Cuda 12.9 since for the z-flux the Calculate fluxes kernel goes from 80->82 registers with this PR and thus 6->5 blocks in flight per SM, which should impact performance. Despite the change in occupancy, the total time for these kernels is not greatly impacted. (The two kernels come from changing ghost zones in the two integration steps, I believe)
"Main":
image
"This PR":
image

The "live-registers" determined line-by-line in the "Source" tab by `ncu` will be a bit off from the actual used registers due to "sort-of-hidden" registers used by CUDA, by the runtime, by the driver, etc. The registers in the above screenshots (which match that from the compiler reported by `-Xptxas -v`) are trustworthy for occupancy calculations, the "live-registers" will be a bit under.

For the first x1-flux kernel, the "live-registers" used in main is actually higher than this commit on certain lines, even though main uses less registers overall:

"Main":
image
"This PR":
image

The "live-register" highwatermark is also on different lines and is higher for main, despite main using fewer registers for the determination of occupancy that really matters.

"Main":
image

"This PR":
image

So trying to tackle the live-registers line-by-line seems too difficult to control. We could instead invoke [`__launch_bounds__`](https://docs.nvidia.com/cuda/cuda-c-programming-guide/#launch-bounds) via [`Kokkos::LaunchBounds`](https://kokkos.org/kokkos-core-wiki/API/core/Execution-Policies.html#common-arguments-for-all-execution-policies) (see also  [Trilinos docs](https://docs.trilinos.org/dev/packages/kokkos/doc/html/structKokkos_1_1LaunchBounds.html) ) in another PR.
</details>

acreyes:
> I did some exploration of the register pressure in different kernels in AthenaPK with and without this PR using `ncu`, attempting to compare the two commits line-by-line. The final take-away is due to the heuristics in how `nvcc` choses to use registers, it's a bit of a lost cause to chase after small register changes. If we're concerned about register pressure to affect occupancy, we should try exposing `__launch_bounds__` (via `Kokkos::LaunchBounds`). A bit disappointing, but I think we should merge in this PR if we like the unification.
> 
Amazing! Thanks @fglines-nv for digging into this. 

From my naive looking kokkos kernels it seems like they're always launched with 128 threads per block. Would setting that explicitly do anything for this or would it require more tailoring to the kernel/meshblock size?

pgrete (approved):
As discussed on Thursday in the PR review, let's get this in! 
Thanks again @acreyes for this nice cleanup.

pgrete:
All check pass (except for the formatting one because of the fork source).
I checked out locally, fixed the formatting and will force merge now.

pgrete:
Houston, we have a merge!

PR #1143, opened 2024-07-29
Small: Quality of life changes for downstream codes

lroberts36:
## PR Summary

- Access tensor indices of `ParArray3D` in pack using `VariableState`.
- Include radiation constant in `PhysicalConstants`
- Add a `TypeList` wrapper for lists of types that allows for concatenation, pulling out types at specific indices, etc.
- Allow for solvers to work on a non-base container

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [ ] Docs build

### DIFF
--- a/src/interface/make_pack_descriptor.hpp
+++ b/src/interface/make_pack_descriptor.hpp
@@ -137,8 +137,8 @@ inline auto MakePackDescriptor(StateDescriptor *psd, const std::vector<Uid_t> &v
   return typename SparsePack<>::Descriptor(base_desc);
 }
 
-template <class... Types, class... Args>
-inline auto MakePackDescriptorFromTypeList(TypeList<Types...>, Args &&...args) {
+template <template <class...> class TL, class... Types, class... Args>
+inline auto MakePackDescriptorFromTypeList(TL<Types...>, Args &&...args) {
   return MakePackDescriptor<Types...>(std::forward<Args>(args)...);
 }
 
@@ -146,7 +146,6 @@ template <class TL, class... Args>
 inline auto MakePackDescriptorFromTypeList(Args &&...args) {
   return MakePackDescriptorFromTypeList(TL(), std::forward<Args>(args)...);
 }
-
 } // namespace parthenon
 
 #endif // INTERFACE_MAKE_PACK_DESCRIPTOR_HPP_
--- a/src/interface/make_pack_descriptor.hpp
+++ b/src/interface/make_pack_descriptor.hpp
@@ -137,13 +137,13 @@ inline auto MakePackDescriptor(StateDescriptor *psd, const std::vector<Uid_t> &v
   return typename SparsePack<>::Descriptor(base_desc);
 }
 
-template <class... Types, class... Args> 
-inline auto MakePackDescriptorFromTypeList(TypeList<Types...>, Args&&... args) { 
+template <class... Types, class... Args>
+inline auto MakePackDescriptorFromTypeList(TypeList<Types...>, Args &&...args) {
   return MakePackDescriptor<Types...>(std::forward<Args>(args)...);
 }
 
-template <class TL, class... Args> 
-inline auto MakePackDescriptorFromTypeList(Args&&... args) { 
+template <class TL, class... Args>
+inline auto MakePackDescriptorFromTypeList(Args &&...args) {
   return MakePackDescriptorFromTypeList(TL(), std::forward<Args>(args)...);
 }
 
--- a/src/interface/make_pack_descriptor.hpp
+++ b/src/interface/make_pack_descriptor.hpp
@@ -33,6 +33,7 @@
 #include "interface/state_descriptor.hpp"
 #include "interface/variable.hpp"
 #include "mesh/mesh.hpp"
+#include "utils/type_list.hpp"
 
 namespace parthenon {
 
@@ -134,6 +135,16 @@ inline auto MakePackDescriptor(StateDescriptor *psd, const std::vector<Uid_t> &v
 
   impl::PackDescriptor base_desc(psd, var_ids, selector, options);
   return typename SparsePack<>::Descriptor(base_desc);
+}
+
+template <class... Types, class... Args> 
+inline auto MakePackDescriptorFromTypeList(TypeList<Types...>, Args&&... args) { 
+  return MakePackDescriptor<Types...>(std::forward<Args>(args)...);
+}
+
+template <class TL, class... Args> 
+inline auto MakePackDescriptorFromTypeList(Args&&... args) { 
+  return MakePackDescriptorFromTypeList(TL(), std::forward<Args>(args)...);
 }
 
 } // namespace parthenon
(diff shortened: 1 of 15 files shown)


src/interface/variable_state.hpp
```diff
@@ -61,7 +61,10 @@ struct VariableState : public empty_state_t {
   int sparse_id;
   int vector_component = NODIR;
   bool initialized = true;
-
+  
+  int t{0};
+  int u{0};
+  int v{0};
```
Yurlungur: Maybe should be initialized to some signal var indicating they haven't been set, like -1?

Yurlungur (approved):
Neat trick.

src/interface/variable_state.hpp
pgrete: :+1: on the default init.

Unrelated to this PR and a more general performance question: Has anyone looked at the impact on performance when putting more and more info in the packs?

pgrete (approved):

src/interface/variable_state.hpp
lroberts36: yeah, probably so. Also maybe would be better to store as an array? Not totally sure.

lroberts36: The message from @pgrete didn't show up for me before. I have not looked carefully at the performance impact, but at least when I put variable state in initially I didn't see any slow down. Of course, there is now more state stored there...

src/interface/make_pack_descriptor.hpp
```diff
@@ -136,6 +137,16 @@ inline auto MakePackDescriptor(StateDescriptor *psd, const std::vector<Uid_t> &v
...
+template <class... Types, class... Args>
+inline auto MakePackDescriptorFromTypeList(TypeList<Types...>, Args &&...args) {
+  return MakePackDescriptor<Types...>(std::forward<Args>(args)...);
+}
+
+template <class TL, class... Args>
+inline auto MakePackDescriptorFromTypeList(Args &&...args) {
+  return MakePackDescriptorFromTypeList(TL(), std::forward<Args>(args)...);
+}
+
```
acreyes:
```suggestion
template <template <class...> class TL, class... Types, class... Args>
inline auto MakePackDescriptorFromTypeList(TL<Types...>, Args &&...args) {
  return MakePackDescriptor<Types...>(std::forward<Args>(args)...);
}

template <class TL, class... Args>
inline auto MakePackDescriptorFromTypeList(Args &&...args) {
  return MakePackDescriptorFromTypeList(TL(), std::forward<Args>(args)...);
}
```
Would a template template parameter like this work? It could provide more flexibility for a downstream code to use their own type lists

src/interface/variable_state.hpp
```diff
@@ -62,6 +62,9 @@ struct VariableState : public empty_state_t {
   int vector_component = NODIR;
   bool initialized = true;
 
+  int t{-1};
+  int u{-1};
+  int v{-1};
```
Yurlungur, line 67: :+1:

src/utils/type_list.hpp
Yurlungur, line 1: We have some similar machinery in `singularity-eos` too... it's funny how this kind of stuff gets repeatedly implemented.

Yurlungur (approved):

src/interface/make_pack_descriptor.hpp
lroberts36: yeah, that is a nice way to do it. suggestion committed.

src/interface/variable_state.hpp
lroberts36: If this is a performance issue, there are definitely things that can be removed and some things that can be moved to the sparse pack itself. That being said, having easy access to this information simplifies a lot of code.

src/solvers/mg_solver.hpp
```diff
@@ -386,20 +394,20 @@ class MGSolver {
...
-    auto decorate_task_name = [partition, level](const std::string &in, auto b) {
-      return std::make_tuple(in + "(p:" + std::to_string(partition) +
-                                 ", l:" + std::to_string(level) + ")",
-                             1, b);
-    };
+//    auto decorate_task_name = [partition, level](const std::string &in, auto b) {
+//      return std::make_tuple(in + "(p:" + std::to_string(partition) +
+//                                 ", l:" + std::to_string(level) + ")",
+//                             1, b);
+//    };
```
pgrete, line 401: Any particular reason to keep this commented (rather than reintroducing or removing)?

src/solvers/mg_solver.hpp
```diff
@@ -76,13 +77,18 @@ class MGSolver {
...
+    internal_types_tl::IterateTypes(
+        [&names](auto t) { names.push_back(decltype(t)::name()); });
+    return names;
   }
 
   MGSolver(StateDescriptor *pkg, MGParams params_in, equations eq_in = equations(),
-           std::vector<int> shape = {})
-      : params_(params_in), iter_counter(0), eqs_(eq_in) {
+           std::vector<int> shape = {}, const std::string &container = "base")
+      : params_(params_in), iter_counter(0), eqs_(eq_in), container_(container) {
```
pgrete, line 91: Just thinking out loud here (independent of this PR): We are still sharing comm buffer of each variable across containers, aren't we?
Do we have any control over ensuring that only one communication piece is running/using the comm buffers in a task list at any given time?

src/utils/type_list.hpp
```diff
@@ -0,0 +1,91 @@
...
+
+#include <string>
+#include <tuple>
+#include <type_traits>
+#include <utility>
+#include <vector>
+
+namespace parthenon {
+template <class... Args>
+struct TypeList {
```
pgrete, line 31: Would you mind adding a minimum amount of documentation, like two lines here what TypeList are about and two lines in the general developer doc highlighting that they are available and their use case?

pgrete (approved):
LGTM (no implementation related  comment/question)

src/solvers/mg_solver.hpp
lroberts36: It is only for debugging multigrid. I kept it because I figured this wouldn't be the last time I spent debugging and didn't want to have to remember how to re-write it...

src/solvers/mg_solver.hpp
lroberts36: We don't have any control over that and I agree that if you had a task list that had independent groups of tasks that worked on different containers but also communicated you could possibly run into issues where one container grabs messages meant for another container. That is similar to some of the issues I had to fix for multi-grid in this PR. I am not sure what the best approach to dealing with this is. Making buffers for each container doesn't seem ideal and it could be a little tricky to build buffers and MPI tag maps for multiple containers in the current framework (since now buffers are built right after remesh, not at the point when containers are first asked for). Maybe we should put up a warning issue about this.

src/utils/type_list.hpp
lroberts36: done

pgrete: Perfect, thanks!

PR #1144, opened 2024-07-30
Fix some restarts w/non-CC fields

c-prather:
When restarting, Parthenon reads fields into a temporary space allocated to be as large as the largest field.  However, if the largest field was e.g. face-centered it would not be calculated correctly, resulting in overflowing the temporary.

Additionally, I think this would have made sparse non-CC fields read wrong.

I just simplify both with the new, better `VarInfo` utilities.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [ ] Docs build


Yurlungur (approved):
good catch! I must have missed this when I was doing the non-CC I/O rewrite

pgrete (approved):
Nice! This also reads a lot better.

PR #667, opened 2022-04-25
Add parallel scan

brryan:
## PR Summary

`Kokkos::parallel_scan` would be really useful for relating swarm indices to zone indices on device in a parallel way. This PR adds `parthenon::par_scan` to the existing `parthenon::par_for` and `parthenon::par_reduce`. 

## PR Checklist

- [x] Code passes cpplint
- [x] New features are documented.
- [x] Adds a test for any bugs fixed. Adds tests for new features.
- [x] Code is formatted
- [x] Changes are summarized in CHANGELOG.md
- [x] CI has been triggered on [Darwin](https://re-git.lanl.gov/eap-oss/parthenon/-/pipelines) for performance regression tests.
- [x] (@lanl.gov employees) Update copyright on changed files

PR #1145, opened 2024-07-31
Fix remaining swarm D->H->D copies

brryan:
## PR Summary

There are some outstanding places in the particle infrastructure where we copy arrays to host to perform sorting logic. This is almost certainly responsible for a just uncovered major performance bottleneck downstream. Should be fixable with prefix sums on device. This PR aims to do that.

- [x] Initialize device data in `Swarm` constructor with kernel launch rather than device/host copies
- [x] Switch some `Swarm` method names from camel case to uppercase
- [x] Rewrite `Defragment` method to use kernel launches rather than device/host copies to defragment the memory pool of particles
- [x] Rewrite `AddEmptyParticles` method to use kernel launches rather than device host copies to provide a list of empty particle indices
- [x] Remove probably unnecessary check on non-`null_ptr`-ness of user swarm boundary functions that was not interacting correctly with recent changes to the forest infrastructure.
- [x] Remove device host copies in `swarm_comms.cpp` ~~(can possibly defer this to another PR if anyone wants to merge this before middle of next week)~~
- [x] Remove copy constructor from `AddRegion` to prevent segfault if reference is not properly stored when making a task collection.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [ ] Docs build

### DIFF
--- a/example/particles/particles.cpp
+++ b/example/particles/particles.cpp
@@ -229,7 +229,7 @@ TaskStatus DepositParticles(MeshBlock *pmb) {
   return TaskStatus::complete;
 }
 
-TaskStatus CreateSomeParticles(MeshBlock *pmb, const double t0) {
+TaskStatus CreateSomeParticles(MeshBlock *pmb, const Real t0) {
   PARTHENON_INSTRUMENT
 
   auto pkg = pmb->packages.Get("particles_package");
@@ -340,8 +340,7 @@ TaskStatus CreateSomeParticles(MeshBlock *pmb, const double t0) {
   return TaskStatus::complete;
 }
 
-TaskStatus TransportParticles(MeshBlock *pmb, const StagedIntegrator *integrator,
-                              const double t0) {
+TaskStatus TransportParticles(MeshBlock *pmb, const Real t0, const Real dt) {
   PARTHENON_INSTRUMENT
 
   auto swarm = pmb->meshblock_data.Get()->GetSwarmData()->Get("my_particles");
@@ -349,8 +348,6 @@ TaskStatus TransportParticles(MeshBlock *pmb, const StagedIntegrator *integrator
   const auto orbiting_particles = pkg->Param<bool>("orbiting_particles");
 
   int max_active_index = swarm->GetMaxActiveIndex();
-
-  Real dt = integrator->dt;
 
   auto &t = swarm->Get<Real>("t").Get();
   auto &x = swarm->Get<Real>(swarm_position::x::name()).Get();
@@ -469,101 +466,33 @@ TaskStatus TransportParticles(MeshBlock *pmb, const StagedIntegrator *integrator
 // Custom step function to allow for looping over MPI-related tasks until complete
 TaskListStatus ParticleDriver::Step() {
   TaskListStatus status;
-  integrator.dt = tm.dt;
+
+  PARTHENON_REQUIRE(integrator.nstages == 1,
+                    "Only first order time integration supported!");
 
   BlockList_t &blocks = pmesh->block_list;
   auto num_task_lists_executed_independently = blocks.size();
 
   // Create all the particles that will be created during the step
   status = MakeParticlesCreationTaskCollection().Execute();
+  PARTHENON_REQUIRE(status == TaskListStatus::complete, "Task list failed!");
 
-  // Loop over repeated MPI calls until every particle is finished. This logic is
-  // required because long-distance particle pushes can lead to a large, unpredictable
-  // number of MPI sends and receives.
-  bool particles_update_done = false;
-  while (!particles_update_done) {
-    status = MakeParticlesUpdateTaskCollection().Execute();
-
-    particles_update_done = true;
-    for (auto &block : blocks) {
-      // TODO(BRR) Despite this "my_particles"-specific call, this function feels like it
-      // should be generalized
-      auto swarm = block->meshblock_data.Get()->GetSwarmData()->Get("my_particles");
-      if (!swarm->finished_transport) {
-        particles_update_done = false;
-      }
-    }
-  }
+  // Transport particles iteratively until all particles reach final time
+  status = IterativeTransport();
+  // status = MakeParticlesTransportTaskCollection().Execute();
+  PARTHENON_REQUIRE(status == TaskListStatus::complete, "Task list failed!");
 
   // Use a more traditional task list for predictable post-MPI evaluations.
   status = MakeFinalizationTaskCollection().Execute();
+  PARTHENON_REQUIRE(status == TaskListStatus::complete, "Task list failed!");
 
   return status;
-}
-
-// TODO(BRR) This should really be in parthenon/src... but it can't just live in Swarm
-// because of the loop over blocks
-TaskStatus StopCommunicationMesh(const BlockList_t &blocks) {
-  PARTHENON_INSTRUMENT
-
-  int num_sent_local = 0;
-  for (auto &block : blocks) {
-    auto sc = block->meshblock_data.Get()->GetSwarmData();
-    auto swarm = sc->Get("my_particles");
-    swarm->finished_transport = false;
-    num_sent_local += swarm->num_particles_sent_;
-  }
-
-  int num_sent_global = num_sent_local; // potentially overwritten by following Allreduce
-#ifdef MPI_PARALLEL
-  for (auto &block : blocks) {
-    auto swarm = block->meshblock_data.Get()->GetSwarmData()->Get("my_particles");
-    for (int n = 0; n < block->neighbors.size(); n++) {
-      NeighborBlock &nb = block->neighbors[n];
-      // TODO(BRR) May want logic like this if we have non-blocking TaskRegions
-      // if (nb.snb.rank != Globals::my_rank) {
-      //  if (swarm->vbswarm->bd_var_.flag[nb.bufid] != BoundaryStatus::completed) {
-      //    return TaskStatus::incomplete;
-      //  }
-      //}
-
-      // TODO(BRR) May want to move this logic into a per-cycle initialization call
-      if (swarm->vbswarm->bd_var_.flag[nb.bufid] == BoundaryStatus::completed) {
-        swarm->vbswarm->bd_var_.req_send[nb.bufid] = MPI_REQUEST_NULL;
-      }
-    }
-  }
-
-  MPI_Allreduce(&num_sent_local, &num_sent_global, 1, MPI_INT, MPI_SUM, MPI_COMM_WORLD);
-#endif // MPI_PARALLEL
-
-  if (num_sent_global == 0) {
-    for (auto &block : blocks) {
-      auto &pmb = block;
-      auto sc = pmb->meshblock_data.Get()->GetSwarmData();
-      auto swarm = sc->Get("my_particles");
-      swarm->finished_transport = true;
-    }
-  }
-
-  // Reset boundary statuses
-  for (auto &block : blocks) {
-    auto &pmb = block;
-    auto sc = pmb->meshblock_data.Get()->GetSwarmData();
-    auto swarm = sc->Get("my_particles");
-    for (int n = 0; n < pmb->neighbors.size(); n++) {
-      auto &nb = block->neighbors[n];
-      swarm->vbswarm->bd_var_.flag[nb.bufid] = BoundaryStatus::waiting;
-    }
-  }
-
-  return TaskStatus::complete;
 }
 
 TaskCollection ParticleDriver::MakeParticlesCreationTaskCollection() const {
   TaskCollection tc;
   TaskID none(0);
-  const double t0 = tm.time;
+  const Real t0 = tm.time;
   const BlockList_t &blocks = pmesh->block_list;
 
   auto num_task_lists_executed_independently = blocks.size();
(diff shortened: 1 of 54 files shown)


src/interface/swarm_container.cpp
```diff
@@ -36,36 +36,6 @@ void SwarmContainer::Initialize(const std::shared_ptr<StateDescriptor> resolved_
...
-  auto &bcs = pmb->pmy_mesh->mesh_bcs;
-  // Check that, if we are using user BCs, they are actually enrolled, and unsupported BCs
-  // are not being used
-  for (int iFace = 0; iFace < 6; iFace++) {
-    if (bcs[iFace] == BoundaryFlag::user) {
-      if (pmb->pmy_mesh->forest.GetTreePtr(pmb->loc.tree())->SwarmBndryFnctn[iFace] ==
-          nullptr) {
-        msg << (iFace % 2 == 0 ? "i" : "o") << "x" << iFace / 2 + 1
-            << " user boundary requested but provided function is null!";
-        PARTHENON_FAIL(msg);
```
brryan, line 58: @lroberts36 With the removal of boundary functions from the  mesh, this check no longer passes on a forest with two trees in the X1 direction and user swarm boundary functions enrolled for `ix1` and `ox1` -- basically each tree has only one of those two BCs but this check tries to find both BCs on each tree. I think it is fine to just remove this check because each tree will only call the boundaries appropriate for that tree (for example a tree with only block boundaries on its `ox1` side should not call user `ox1` boundary conditions on the swarms). I wanted to check with you though that this is reasonable to do.

tst/unit/test_swarm.cpp
```diff
@@ -212,8 +212,8 @@ TEST_CASE("Swarm memory management", "[Swarm]") {
 
   // Check that data was moved during defrag
   x_h = swarm->Get<Real>(swarm_position::x::name()).Get().GetHostMirrorAndCopy();
-  REQUIRE(x_h(2) == 1.2);
-  REQUIRE(x_h(4) == 1.1);
+  REQUIRE(x_h(2) == 1.1);
+  REQUIRE(x_h(4) == 1.2);
```
brryan, line 216: This change is because the previous defragmentation operation was flipping the ordering of particles that got moved, whereas the new method preserves the ordering during the move.

src/interface/swarm_container.cpp
lroberts36: Yes, I think this is reasonable. As you say, trees that share a boundary with another tree will not have physical boundary conditions defined on those boundaries.

lroberts36 (approved):
LGTM, glad to see there was a nice way to update the empty indices entirely on device.

src/interface/swarm.cpp
```diff
@@ -251,129 +246,153 @@ void Swarm::setPoolMax(const std::int64_t nmax_pool) {
...
-          max_active_index_ -= 1;
+  int &max_active_index = max_active_index_;
+
+  auto &mask = mask_;
+  auto &marked_for_removal = marked_for_removal_;
+
+  // Update mask, count number of removed particles
+  int num_removed = 0;
+  parthenon::par_reduce(
+      PARTHENON_AUTO_LABEL, 0, max_active_index,
```
pgrete, line 327: Is the last `max_active_index` inclusive or exclusive?
I'm asking with respect to the differences between par_for bounds in Kokkos and Parthenon.

pgrete (approved):
Thanks for taking care of this.
Looks good (though I should knowledge on parsing `parallel_scan`s` ;) )

brryan:
OK I fixed up the swarm communication logic to also function on device. Those specific changes are shown in #1154. I used a combination of parallel `for`s/`reduce`s and atomic opartions.

src/tasks/tasks.hpp
```diff
@@ -544,6 +544,8 @@ class TaskRegion {
 
  public:
   TaskRegion() = delete;
+  TaskRegion(const TaskRegion &) = delete; // Prevent copying TaskRegions during AddRegion
+                                           // calls which is a segfault
```
brryan, line 548: I ran into a nasty bug where if you mistakenly do
```c++
TaskRegion my_region = tc.AddRegion(N);
```
rather than storing the reference
```c++
TaskRegion &my_region = tc.AddRegion(N);
```
you get a very unintuitive segfault. I think we always want to return and store the reference from this function -- any reason not to delete the copy constructor to enforce this?

src/interface/swarm.cpp
brryan: `max_active_index` is inclusive so this loop should be written correctly. 

`Kokkos::parallel_scan(label, N, ...` loops from `0` to `N-1` (exclusive) so the scans here should also be correct if I am reading them right.

Adding `parthenon::par_scan` would be a good change to make in a future PR

brryan:
OK this is ready for review/merge and should cure the current host/device copy performance bottlenecks in swarms.

example/particles/particles.cpp
```diff
@@ -577,40 +506,161 @@ TaskCollection ParticleDriver::MakeParticlesCreationTaskCollection() const {
   return tc;
 }
 
-TaskCollection ParticleDriver::MakeParticlesUpdateTaskCollection() const {
+// TODO(BRR) To be used in the future, currently there appears to be a bug in iterative
+// tasking
```
Yurlungur: Might be worth adding an issue so we don't lose track of this.

src/interface/swarm.cpp
```diff
@@ -251,129 +245,153 @@ void Swarm::setPoolMax(const std::int64_t nmax_pool) {
...
+// indices into empty elements of the swarm pool, where N is the number of empty indices
+void Swarm::UpdateEmptyIndices() {
+  auto &mask = mask_;
+  auto &empty_indices = empty_indices_;
+
+  // Associate scratch memory
+  auto &empty_indices_scan = scratch_a_;
+
+  // Calculate prefix sum of empty indices
+  Kokkos::parallel_scan(
```
Yurlungur: :+1: for `parallel_scan`

Yurlungur (approved):

Yurlungur:
@astrobarker now that you're back you should be paying attention to swarm development in parthenon.

example/particles/particles.cpp
brryan: Yes you're right -- written up in #1155

src/interface/swarm.cpp
acreyes: I think `parthenon::par_scan` is already in from #667?

brryan: You're right -- thanks for the reminder! Migrated this PR to `parthenon::par_scan`.

example/particles/particles.cpp
```diff
@@ -340,8 +340,7 @@ TaskStatus CreateSomeParticles(MeshBlock *pmb, const double t0) {
   return TaskStatus::complete;
 }
 
-TaskStatus TransportParticles(MeshBlock *pmb, const StagedIntegrator *integrator,
-                              const double t0) {
+TaskStatus TransportParticles(MeshBlock *pmb, const Real t0, const Real dt) {
```
pgrete: Here `dt` (and in other places time integrator dependent quantities) should remain `double` (or `auto` when getting times from `tm`).
If we ever fix single precision runs (Rea -> float) some quantities (like time or global reductions) should remain at higher precision (or might cause issues in the long run).

example/particles/particles.cpp
```diff
@@ -469,101 +466,33 @@ TaskStatus TransportParticles(MeshBlock *pmb, const StagedIntegrator *integrator
...
+
+  PARTHENON_REQUIRE(integrator.nstages == 1,
+                    "Only first order time integration supported!");
 
   BlockList_t &blocks = pmesh->block_list;
   auto num_task_lists_executed_independently = blocks.size();
 
   // Create all the particles that will be created during the step
   status = MakeParticlesCreationTaskCollection().Execute();
+  PARTHENON_REQUIRE(status == TaskListStatus::complete, "Task list failed!");
```
pgrete: How about being more specic here (and for the other REQUIRES below), e.g., "Task list failed!" -> "MakeParticlesCreation task list failed!"?

example/particles/particles.cpp
```diff
@@ -469,101 +466,33 @@ TaskStatus TransportParticles(MeshBlock *pmb, const StagedIntegrator *integrator
...
-  }
-
-  return TaskStatus::complete;
-}
-
 TaskCollection ParticleDriver::MakeParticlesCreationTaskCollection() const {
   TaskCollection tc;
   TaskID none(0);
-  const double t0 = tm.time;
+  const Real t0 = tm.time;
```
pgrete: See above.

example/particles/particles.cpp
```diff
@@ -577,40 +506,161 @@ TaskCollection ParticleDriver::MakeParticlesCreationTaskCollection() const {
   return tc;
 }
 
-TaskCollection ParticleDriver::MakeParticlesUpdateTaskCollection() const {
+// TODO(BRR) To be used in the future, currently there appears to be a bug in iterative
```
pgrete: I'm not a super big fan of extended commented code that may get fixed on some time scale (while the other parts of the code base keep moving).
Would you mind removing the commented code from the PR and stash/paste it in the issue you already opened?

example/particles/particles.cpp
```diff
@@ -577,40 +506,161 @@ TaskCollection ParticleDriver::MakeParticlesCreationTaskCollection() const {
...
+          }
+        },
+        Kokkos::Sum<int>(num_unfinished_block));
+    num_unfinished_local += num_unfinished_block;
+  }
+
+  int num_unfinished_global = num_unfinished_local;
+#ifdef MPI_PARALLEL
+  MPI_Allreduce(&num_unfinished_local, &num_unfinished_global, 1, MPI_INT, MPI_SUM,
+                MPI_COMM_WORLD);
```
pgrete: In principle, could work with just a single ` num_unfinished` var using an `MPI_IN_PLACE`.

example/particles/particles.cpp
```diff
@@ -577,40 +506,161 @@ TaskCollection ParticleDriver::MakeParticlesCreationTaskCollection() const {
...
+}
+
+TaskCollection ParticleDriver::IterativeTransportTaskCollection(bool *done) const {
   TaskCollection tc;
   TaskID none(0);
-  const double t0 = tm.time;
   const BlockList_t &blocks = pmesh->block_list;
+  const int nblocks = blocks.size();
+  const Real t0 = tm.time;
+  const Real dt = tm.dt;
```
pgrete: See above.

example/particles/particles.cpp
```diff
@@ -577,40 +506,161 @@ TaskCollection ParticleDriver::MakeParticlesCreationTaskCollection() const {
...
+    auto &t = swarm->Get<Real>("t").Get();
+
+    auto swarm_d = swarm->GetDeviceContext();
+
+    int num_unfinished_block = 0;
+    parthenon::par_reduce(
+        PARTHENON_AUTO_LABEL, 0, max_active_index,
+        KOKKOS_LAMBDA(const int n, int &num_unfinished) {
+          if (swarm_d.IsActive(n)) {
+            if (t(n) < tf) {
```
pgrete, line 528: On older CUDA version (<= 11.6 IIRC) this might cause issue as `tf` passed as parameter is not correctly captured.
The dangerous part is that the compiler does not complain but just uses nonsense data.
Given that some older CUDA version are still floating around I suggest a `const auto tf_ = tf` outside of the kernel.

src/interface/swarm_comms.cpp
```diff
@@ -443,18 +356,37 @@ void Swarm::UnloadBuffers_() {
...
-    // construct map from buffer index to swarm index (or just return vector of
-    // indices!)
     const int particle_size = GetParticleDataSize();
     auto swarm_d = GetDeviceContext();
 
+    // Change meaning of neighbor_received_particles from particles per neighbor to
+    // cumulative particles per neighbor
+    int val_prev = 0;
+    for (int n = 0; n < nbmax; n++) {
+      double val_curr = neighbor_received_particles_h(n);
```
pgrete: This is an `int` isn't it?
```suggestion
      auto val_curr = neighbor_received_particles_h(n);
```

pgrete (commented):
Alright, I'm done reviewing (some comments are probably worth addressing).
I still want to check downstream (tomorrow at the latest) before final approval.

example/particles/particles.cpp
brryan: That's good to know -- I actually assumed that function arguments like this would always be correctly captured by lambdas inside the function. Fixed (except as a reference, `const auto &tf = tf_;`, which I again assume is fine).

example/particles/particles.cpp
brryan: Ah yes you're right, I forget we were enforcing `double` for timestep-related quantities.

It was easy for me to mistakenly think this was just a place where I had typed `double` instead of `Real` out of muscle memory in the past. One way to avoid this would be to have a separate type throughout Parthenon for this, `TimingReal` or something. Maybe that's overkill but it would be self-documenting. Just mentioning it in case that seems like a good idea. 

Also, there's a problem here actually which is that these time values are always being used to compare against current particle times, which are stored as a `ParticleVariable` and always have `Real` type. I think it would be quite invasive to enforce that a time variable (which is currently enrolled optionally by a user) always has `double` type. I'm inclined now to just fail if `Real != double` but particles are enabled, or at least throw a warning, because we are currently always doing time comparisons at `Real` precision in this example, and particles implementations generally even after I switched these variables back to `double` in function signatures. Do you have a preference here?

example/particles/particles.cpp
brryan: Good idea, fixed :+1:

example/particles/particles.cpp
brryan: :+1:

example/particles/particles.cpp
brryan: Yeah you are right about this, no reason the code can't be saved in an issue. I will do that.

I'm hopeful we can switch to the iterative tasking framework soon, because it's a really elegant solution for particle transport loops like this one.

brryan: There's a similar (maybe-)for-the-future code block in `swarm_comms.cpp` that I've also deleted here and moved to an issue.

example/particles/particles.cpp
brryan: Yes that's much cleaner, thanks for pointing this out, two `int`s removed.

example/particles/particles.cpp
brryan: :+1:

src/interface/swarm_comms.cpp
brryan: Good catch! Scary that that just compiled without warnings and worked. Warning/erroring on implicit casts would probably cause a whole bunch of headaches though. Anyway, this is fixed now.

brryan (commented):
Thanks @pgrete! Code is much improved. 

Only major question is what to do for this PR about the `double` versus `Real` time comparison issue, because particle time variables are locked in to `Real` now.

example/particles/particles.cpp
pgrete: I see now that this is a more subtle/complex issue.

While a `TimeReal` would in the long run potentially make sense, I agree that this is overkill here.

Keeping it Real (tm) (the data stored in the particle) is probably not that big of an issue. It's just important the the info from the time integrator is properly passed on.
I think adding a warning should be fine for now.

pgrete:
I did some quick downstream tests and the results look good.
- for swarm_container.cpp::165::Receive [region] 
orig: 2.02s
new: 1.25s
- for swarm_container.cpp::155::Send
orig: 0.97s
new: 0.53s

Feel free to hit merge whenever.

example/particles/particles.cpp
brryan: Warning added when state descriptors are created if swarms are present and `Real != double` :+1:

PR #1148, opened 2024-07-31
Add `GetPackDimension` to `StateDescriptor`

c-prather:
Sometimes when initializing a package, you want to know how many variables another package has declared (e.g., a `Flux` package might need the total number of `WithFluxes` variables declared so far).  The `Mesh` doesn't exist yet, so you have to ask the other packages directly, and have them tally up variable lists.  Thus, `GetPackDimension`.

This is the last of the patches I've been carrying in KHARMA's version of Parthenon vs. `develop`.  It works for my purposes downstream, ~~but I'm pretty sure it would fail for `Overridable` variables and maybe other corner cases.  Happy to try to make it more robust if it seems useful, or keep it downstream if no one else needs this kind of info at package init time (or maybe there's some cleaner way to get it? Create a temporary `resolved_state` or something?)~~

I've checked and these functions work fine for the `StateDescriptor` returned by `CreateResolvedStateDescriptor`, which I think already resolves any `Overridable` variables down to a single entry each.  So, if anyone else needs this sort of crazy thing, the code in here at least has a decent chance of being robust for whatever use case they're throwing at it.  Since I think the overall function has a chance of being broadly useful, I'm going to remove WIP, add a changelog entry, and ask for reviews/opinions.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/src/interface/state_descriptor.cpp
+++ b/src/interface/state_descriptor.cpp
@@ -13,6 +13,7 @@
 
 #include <iomanip>
 #include <iostream>
+#include <numeric>
 #include <sstream>
 #include <string>
 #include <unordered_map>
@@ -571,24 +572,16 @@ int StateDescriptor::GetPackDimension(const std::vector<std::string> &req_names,
                                       const Metadata::FlagCollection &flags,
                                       const std::vector<int> &sparse_ids) {
   std::vector<std::string> names = GetVariableNames(req_names, flags, sparse_ids);
-  int nvar = 0;
+  int dimension = 0;
   for (auto name : names) {
     const auto &meta = metadataMap_[VarID(name)];
-    int var_len = 0;
-    if (meta.Shape().size() < 1) {
-      var_len = 1;
-    } else if (meta.Shape().size() == 1) {
-      var_len = meta.Shape()[0];
-    } else if (meta.Shape().size() == 2) {
-      var_len = meta.Shape()[0] * meta.Shape()[1];
-    } else if (meta.Shape().size() == 3) {
-      var_len = meta.Shape()[0] * meta.Shape()[1] * meta.Shape()[2];
-    } else {
-      PARTHENON_THROW("Computing total lengths of tensors of rank >3 is not supported!");
-    }
-    nvar += var_len;
+    // if meta.Shape().size() < 1, then 'accumulate' will return the initialization value,
+    // which is 1. Otherwise, this multiplies all elements present in 'Shape' to obtain
+    // total length
+    dimension += std::accumulate(meta.Shape().begin(), meta.Shape().end(), 1,
+                                 [](auto a, auto b) { return a * b; });
   }
-  return nvar;
+  return dimension;
 }
 int StateDescriptor::GetPackDimension(const std::vector<std::string> &req_names,
                                       const std::vector<int> &sparse_ids) {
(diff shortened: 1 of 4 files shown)


src/interface/state_descriptor.cpp
```diff
@@ -566,4 +566,37 @@ StateDescriptor::GetVariableNames(const Metadata::FlagCollection &flags) {
...
+  int nvar = 0;
+  for (auto name : names) {
+    const auto &meta = metadataMap_[VarID(name)];
+    int var_len = 0;
+    if (meta.Shape().size() < 1) {
+      var_len = 1;
+    } else {
+      var_len = meta.Shape()[0];
+    }
+    nvar += var_len;
```
pgrete: Could this just be
`nvar += meta.Shape().size() < 1 ? 1 : meta.Shape()[0];` ?

src/interface/state_descriptor.cpp
```diff
@@ -566,4 +566,37 @@ StateDescriptor::GetVariableNames(const Metadata::FlagCollection &flags) {
...
+                                      const std::vector<int> &sparse_ids) {
+  std::vector<std::string> names = GetVariableNames(req_names, flags, sparse_ids);
+  int nvar = 0;
+  for (auto name : names) {
+    const auto &meta = metadataMap_[VarID(name)];
+    int var_len = 0;
+    if (meta.Shape().size() < 1) {
+      var_len = 1;
+    } else {
+      var_len = meta.Shape()[0];
```
pgrete: Does this work for tensor variables?

c-prather: Indeed it did not!  Should be fixed w/regression test now.

src/interface/state_descriptor.cpp
c-prather: This was a product of special-casing different `TopologicalElements` originally.  Since now we have to special-case tensors due to below, I'm inclined to keep the extra verbosity to keep it instantly readable.

src/interface/state_descriptor.cpp
Yurlungur: Can you not do this with an (almost) one-liner rather than all this branching?
```C++
if (meta.Shape().size() > 3) {
  PARTHENON_THROW("Computing total lengths of tensors of rank >3 is not supported!");
} else {
  // if meta.Shape().size() < 1, then var_len will be set to the initialization value, which is 1.
  var_len = std::accumulate(meta.Shape().begin(), meta.Shape.end(), 1, [](auto a, auto b){ return a*b;});
}
```

src/interface/state_descriptor.cpp
Yurlungur: I am going to push back against the revised branchy code. See my review.

Yurlungur (approved):
Minor comments below.

src/interface/state_descriptor.cpp
c-prather: Sure! Since the `accumulate` version is fully general, I just removed the error and made it a single line.  I don't think the branching matters since this function is called at most a few times, but having a general version we don't have to think about again will be nice.

Yurlungur: :+1: I don't think the branching matters for performance but the accumulate one-liner is much easier to read IMO.

pgrete (approved):

PR #1152, opened 2024-08-10
Bugfix: Fix memory leak 

adamdempsey90:
## PR Summary

Testing in a downstream code caught a bad memory leak related to the tasks that would eventually crash the node. I tracked this down to the use of `abi::__cxa_demangle` for the task graph output machinery.  `abi::__cxa_demangle` for one reason or another explicitly calls `malloc` for the name it returns. So we have to call `free` on the name after it's used. 

I can confirm that this change fixes the memory leak.

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build


pdmullen (approved):
Holy cow I cannot believe you were able to track this down...

Yurlungur (approved):
Wow nice catch!

PR #1156, opened 2024-08-21
WIP: Testing some `par_dispatch` stuff

lroberts36:
## PR Summary
This is just an attempt to see if some the template magic in #1142 could be written in a little different way. Basically just copies the ideas there but structures the code differently. Seems to be working both on cpu and on device.

- All loops can be called with either sets of integers (as currently supported) or with sets of `IndexRange`s defining the loop bounds.  This is enabled by `LoopBoundTranslator`.
- All loop patterns should work for any rank of loop (e.g. except for `LoopPatternTPTTR` which requires at least rank 2)
- Things are set up to try different types of work partitioning in hierarchical loops like TPTTR
- If an unsupported loop pattern for a particular loop is requested, should automatically fail through to a pattern that does support that type of loop (generally `FlatRange`).
- Adds `ThreadVectorRange` as an option for `par_for_inner`

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build


lroberts36:
Testing vectorization using this branch in Riot with `gcc/9.4.0` on a `skylake-platinum` node  with `-O3 -march=skylake-avx512` I get that with the old version of `par_dispatch` the simd for loop is vectorized in 191 places and with the new version it vectorizes in 200 places. Not sure what to make of this difference. It is hard to determine more from the gcc vectorization report.

The run time difference for a 3D AMR test problem between these two branches is negligible, the old one wins by maybe a couple of percent but I think the run to run variance is larger than that.

lroberts36:
no longer needed now that #1142 is in

PR #1159, opened 2024-08-23
Add more global time step control

adamdempsey90:
## PR Summary

This was originally in a downstream PR, but some functionality was not available, so I've moved it upstream. 

This adds some extra timestep controllers that are not package / meshblock / processor dependent.
- `dt_min`: if `dt <= dt_min` for `dt_min_count_max` cycles, then `PARTHENON_FAIL`. Defaults to 10 bad cycles in a row.
- `dt_max`: if `dt >= dt_max`, for `dt_max_count_max` cycles, then `PARTHENON_FAIL`. Defaults to 1 bad cycle in a row.
- `dt_user`: user settable timestep suggestion (`dt = min(dt, dt_user)`)
- `dt_force`: Force the timestep to `dt_force` regardless of other controllers
- `dt_init`: Set the first timestep to this value but allow it to be smaller if it wants to be. To force this value, the user can set `dt_init_force = true`. 
- `dt_floor`: sets `dt = std::max(dt , dt_floor)`
- `dt_ceil` : sets `dt = std::min(dt, dt_ceil)`
- `dt_factor`: `dt / dt_previous <= dt_factor`. Previously, this was hard coded to 2.

The main purpose of `dt_min` and `dt_max` is to stop simulations from going off the rails and cycling on `1e-20` timesteps or blowing up.  

All of these options are "off" by default. 

I have tested all options in a downstream code.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [ ] Docs build

### DIFF
--- a/src/driver/driver.cpp
+++ b/src/driver/driver.cpp
@@ -205,33 +205,36 @@ void EvolutionDriver::InitializeBlockTimeSteps() {
 // \brief function that loops over all MeshBlocks and find new timestep
 
 void EvolutionDriver::SetGlobalTimeStep() {
-  // Check if user wants to force the value
   if (dt_force > 0.0) {
+    // Check if user wants to force the value
     tm.dt = dt_force;
+  } else if ((tm.ncycle == 0) && (dt_init_force) && (dt_init > 0.0)) {
+    // Check if user wants to force the first cycle value
+    tm.dt = dt_init;
   } else {
+    if (tm.dt < 0.1 * std::numeric_limits<Real>::max()) {
+      // don't allow dt to grow by more than 2x
+      // consider making this configurable in the input
+      tm.dt *= dt_factor;
+    }
     // Check for special first cycle value
     if (tm.ncycle == 0) {
       tm.dt = std::min(tm.dt, dt_init);
     }
-    // don't allow dt to grow by more than 2x
-    // consider making this configurable in the input
-    if (tm.dt < 0.1 * std::numeric_limits<Real>::max()) {
-      tm.dt *= 2.0;
-    }
     // Allow the meshblocks to vote
-    Real big = std::numeric_limits<Real>::max();
     for (auto const &pmb : pmesh->block_list) {
       tm.dt = std::min(tm.dt, pmb->NewDt());
-      pmb->SetAllowedDt(big);
+      pmb->SetAllowedDt(std::numeric_limits<Real>::max());
     }
     // Allow the user to vote
     tm.dt = std::min(tm.dt, dt_user);
-  }
-
+    // Force timestep to be in the allowable range
+    tm.dt = std::max(dt_floor, std::min(tm.dt, dt_ceil));
 #ifdef MPI_PARALLEL
-  PARTHENON_MPI_CHECK(MPI_Allreduce(MPI_IN_PLACE, &tm.dt, 1, MPI_PARTHENON_REAL, MPI_MIN,
-                                    MPI_COMM_WORLD));
+    PARTHENON_MPI_CHECK(MPI_Allreduce(MPI_IN_PLACE, &tm.dt, 1, MPI_PARTHENON_REAL,
+                                      MPI_MIN, MPI_COMM_WORLD));
 #endif
+  }
 
   // Check that we have not gone off the rails
   if (tm.dt <= dt_min) {
@@ -256,6 +259,8 @@ void EvolutionDriver::SetGlobalTimeStep() {
   }
 
   // Limit timestep if it would take us past desired endpoint
+  // This comes after the bounds checking so that we don't fail when epsilon
+  // away from tlim
   if (tm.time < tm.tlim && (tm.tlim - tm.time) < tm.dt) {
     tm.dt = tm.tlim - tm.time;
   }
--- a/src/driver/driver.cpp
+++ b/src/driver/driver.cpp
@@ -205,15 +205,27 @@ void EvolutionDriver::InitializeBlockTimeSteps() {
 // \brief function that loops over all MeshBlocks and find new timestep
 
 void EvolutionDriver::SetGlobalTimeStep() {
-  // don't allow dt to grow by more than 2x
-  // consider making this configurable in the input
-  if (tm.dt < 0.1 * std::numeric_limits<Real>::max()) {
-    tm.dt *= 2.0;
-  }
-  Real big = std::numeric_limits<Real>::max();
-  for (auto const &pmb : pmesh->block_list) {
-    tm.dt = std::min(tm.dt, pmb->NewDt());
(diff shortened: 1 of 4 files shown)


src/driver/driver.cpp
```diff
@@ -205,25 +205,60 @@ void EvolutionDriver::InitializeBlockTimeSteps() {
...
-  if (tm.dt < 0.1 * std::numeric_limits<Real>::max()) {
-    tm.dt *= 2.0;
-  }
-  Real big = std::numeric_limits<Real>::max();
-  for (auto const &pmb : pmesh->block_list) {
-    tm.dt = std::min(tm.dt, pmb->NewDt());
-    pmb->SetAllowedDt(big);
+  // Check if user wants to force the value
+  if (dt_force > 0.0) {
+    tm.dt = dt_force;
```
lroberts36, line 210: The `AllReduce` could be skipped if this clause is chosen.

src/driver/driver.cpp
```diff
@@ -205,25 +205,60 @@ void EvolutionDriver::InitializeBlockTimeSteps() {
...
+  } else {
+    // Check for special first cycle value
+    if (tm.ncycle == 0) {
+      tm.dt = std::min(tm.dt, dt_init);
```
lroberts36, line 222: Maybe for full configurability we should have `dt_init_max` and `dt_init_force`, since it may be confusing if your simulation starts with a different time step than what you requested with `dt_init`

src/driver/driver.cpp
```diff
@@ -205,25 +205,60 @@ void EvolutionDriver::InitializeBlockTimeSteps() {
...
+    }
+    // don't allow dt to grow by more than 2x
+    // consider making this configurable in the input
+    if (tm.dt < 0.1 * std::numeric_limits<Real>::max()) {
+      tm.dt *= 2.0;
```
lroberts36: Maybe this PR is the place to add a parameter controlling the max `dt` increase.

src/driver/driver.cpp
```diff
@@ -205,25 +205,60 @@ void EvolutionDriver::InitializeBlockTimeSteps() {
...
+  } else {
+    dt_min_count = 0;
+  }
+  if (tm.dt >= dt_max) {
+    if (++dt_max_count >= dt_max_count_max) {
+      std::stringstream msg;
+      msg << "Timesetep has risen above maximum (parthenon/time/dt_max=" << dt_max
+          << ") for more than " << dt_max_count_max << " steps";
+      PARTHENON_FAIL(msg);
+    }
```
lroberts36, line 256: Why not just limit the step to `dt_max`?

lroberts36 (approved):
LGTM, definitely will be useful for me in a couple of places downstream. Just a few questions below that are non-blocking.

src/driver/driver.cpp
adamdempsey90: I debated whether to add new parameters called `dt_floor` and `dt_ceil` such that `dt = st::max(dt_floor, std::min(dt, dt_ceil))`, which does what you're suggesting. Those are separate from `dt_min` and `dt_max`. 

Sims may just blow up in timestep, and `dt_max` is meant to catch that and die. It's not really meant to limit your dead simulation to a large timestep value -- but `dt_ceil` could do that.

src/driver/driver.cpp
adamdempsey90: Sure, I can add that.

src/driver/driver.cpp
adamdempsey90: I can add `dt_init_force` as a `bool` and then force `dt_init` if it's true.

src/driver/driver.cpp
adamdempsey90: Yes, will update

src/driver/driver.cpp
```diff
@@ -205,25 +205,60 @@ void EvolutionDriver::InitializeBlockTimeSteps() {
...
+    // don't allow dt to grow by more than 2x
+    // consider making this configurable in the input
+    if (tm.dt < 0.1 * std::numeric_limits<Real>::max()) {
+      tm.dt *= 2.0;
+    }
+    // Allow the meshblocks to vote
+    Real big = std::numeric_limits<Real>::max();
+    for (auto const &pmb : pmesh->block_list) {
+      tm.dt = std::min(tm.dt, pmb->NewDt());
+      pmb->SetAllowedDt(big);
```
brryan:
```suggestion
    for (auto const &pmb : pmesh->block_list) {
      tm.dt = std::min(tm.dt, pmb->NewDt());
      pmb->SetAllowedDt(std::numeric_limits<Real>::max());
```
`big` only appears once so not sure it needs a variable

src/driver/driver.cpp
```diff
@@ -205,25 +205,60 @@ void EvolutionDriver::InitializeBlockTimeSteps() {
...
+    }
+    // Allow the user to vote
```
brryan:
```suggestion
    // Allow the user to enforce maximum timestep
```
Just because this isn't a vote in the sense that it can't be overruled in certain regimes.

brryan (approved):
LGTM

src/driver/driver.cpp
adamdempsey90: Fixed

src/driver/driver.cpp
adamdempsey90: Fixed

src/driver/driver.cpp
adamdempsey90: Done with new parameter `dt_factor`

src/driver/driver.cpp
lroberts36: Maybe it makes sense to rename as `dt_max_allowed` and `dt_min_allowed` so that it is clear that they aren't a floor and ceiling? I agree that there probably isn't much use for the floor/ceil since hitting those generally implies something isn't going so well in the sim. Just thought maybe we should add them for completeness.

adamdempsey90: Went ahead and added `dt_floor` and `dt_ceil`

Yurlungur (approved):
This is a great improvement! Have definitely felt the lack of these features more than once before. Can we make sure to add to the changelog? Also would be good to have these in the docs for reference.

adamdempsey90:
> This is a great improvement! Have definitely felt the lack of these features more than once before. Can we make sure to add to the changelog? Also would be good to have these in the docs for reference.

Done. I took the opportunity to also alphabetize the parameters.

Yurlungur:
> > This is a great improvement! Have definitely felt the lack of these features more than once before. Can we make sure to add to the changelog? Also would be good to have these in the docs for reference.
> 
> Done. I took the opportunity to also alphabetize the parameters.

Nice!

PR #1161, opened 2024-08-26
Allow for accessing flux `Metadata` from `Metadata::WithFluxes` `Metadata`

lroberts36:
## PR Summary

This PR adds a `std::shared_ptr<Metadata> flux_metadata` member to `Metadata` that is initialized in the `Metadata` constructor when the flag `Metadata::WithFluxes` is set. A new `Metadata` constructor is added that takes a list of flags for both the original field and the flux field, as well as separate prolongation and restriction operations for the base field and the flux field (but all of the old constructors behavior is unchanged). The easier way to set non-default properties of the flux `Metadata` is to get access to it through the method `Metadata::GetSPtrFluxMetadata()` and `Set` and `Unset` flags there and register refinement operations, e.g.
```c++
Metadata m({Metadata::Face, Metadata::WithFluxes, Metadata::Vector}, {1, 2});
auto pm_flux = m.GetSPtrFluxMetadata(); 
pm_flux.Unset(Metadata::OneCopy); 
pm_flux.RegisterRefinementOps<OpPro, OpRes, OpInternalPro>();  
```

This PR also adds a flag `Metadata::CellMemAligned`, which causes a field to have memory allocated with the same size as a cell variable no matter what its topological type is. This is set by default for fluxes associated with cell centered fields (to ensure backwards compatibility). Obviously, having this flag set means that face, edge, and nodal fields will be missing one layer of ghosts on some sides, but having it set can allow for vectorization in some downstream codes. 

Finally, (sorry) this PR includes a couple of small changes to sparse pack cache checking and to boundary communication. These are just performance enhancements that got merged into this branch by mistake and it was easier to leave them in than jump through a bunch of git hoops. If people are strongly opposed to including them here, let me know. 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [ ] Docs build


Yurlungur (approved):
LGTM! Should we have a test?

lroberts36:
> LGTM! Should we have a test?

Let's have a Sam based test first. Once that passes I can add one to the unit tests.

lroberts36:
@Yurlungur: Maybe this isn't the "right" way to do things. It seems like we will have to add a second flag for many options just so that it can pass through to the fluxes. Maybe we should really be making a `Metadata` constructor that takes two vectors of metadata flags and uses the second one to internally construct a second `Metadata` object for the flux field. What do you think?

Also, I added a flag `Metadat::CellMemAligned` in this PR that allows you to explicitly choose what the allocated size of an array is (so that you can do the Riot vectorization tricks more easily in the FLD solver).

Yurlungur:
> @Yurlungur: Maybe this isn't the "right" way to do things. It seems like we will have to add a second flag for many options just so that it can pass through to the fluxes. Maybe we should really be making a `Metadata` constructor that takes two vectors of metadata flags and uses the second one to internally construct a second `Metadata` object for the flux field. What do you think?

I think this is a nice idea... though I would suggest it be optional.

> Also, I added a flag `Metadat::CellMemAligned` in this PR that allows you to explicitly choose what the allocated size of an array is (so that you can do the Riot vectorization tricks more easily in the FLD solver).

:+1:

Yurlungur (approved):

lroberts36:
> > @Yurlungur: Maybe this isn't the "right" way to do things. It seems like we will have to add a second flag for many options just so that it can pass through to the fluxes. Maybe we should really be making a `Metadata` constructor that takes two vectors of metadata flags and uses the second one to internally construct a second `Metadata` object for the flux field. What do you think?
> 
> I think this is a nice idea... though I would suggest it be optional.
> 

It would have to be optional for backward compatibility, but the idea would be to just have a set of default Metadata (which we effectively have now) and allow that to be overridden with by the new constructor.

src/bvals/comms/boundary_communication.cpp
```diff
@@ -213,7 +213,7 @@ TaskStatus ReceiveBoundBufs(std::shared_ptr<MeshData<Real>> &md) {
       [&all_received](auto pbuf) { all_received = pbuf->TryReceive() && all_received; });
 
   int ibound = 0;
-  if (Globals::sparse_config.enabled) {
+  if (Globals::sparse_config.enabled && all_received) {
```
lroberts36, line 216: This changes to allocating fields only after all boundary buffers are received. This should have no impact on code behavior, but can have a performance impact in some cases. The change is unrelated to the rest of the PR.

src/interface/sparse_pack_base.hpp
```diff
@@ -138,7 +138,7 @@ struct PackDescriptor {
   // default constructor needed for certain use cases
   PackDescriptor()
       : nvar_groups(0), var_group_names({}), var_groups({}), with_fluxes(false),
-        coarse(false), flat(false), identifier("") {}
+        coarse(false), flat(false), identifier(""), nvar_tot(0) {}
```
lroberts36, line 141: This just counts the total number of variables included in the pack at the time of `PackDescriptor` construction so it doesn't have to be redone every time the cache is checked to see if a given pack is stale. Unrelated to the bulk of this PR.

src/interface/sparse_pack_base.cpp
```diff
@@ -67,19 +67,16 @@ SparsePackBase::GetAllocStatus(T *pmd, const PackDescriptor &desc,
...
     for (int i = 0; i < nvar; ++i) {
       for (const auto &[var_name, uid] : desc.var_groups[i]) {
-        if (uid_map.count(uid) > 0) {
-          const auto pv = uid_map.at(uid);
-          astat.push_back(pv->GetAllocationStatus());
-        } else {
-          astat.push_back(-1);
-        }
+        astat[idx++] =
+            uid_map.count(uid) > 0 ? (uid_map.at(uid))->GetAllocationStatus() : -1;
```
lroberts36, line 79: Try to limit branching and preallocate the `std::vector` for `SparsePack` cache checking. Unrelated to the rest of the PR.

pdmullen (approved):

Yurlungur (approved):
This all looks good to me.

PR #1167, opened 2024-09-06
Store block gid and neighbor refinement levels in sparse packs

brryan:
## PR Summary

For certain discrete methods we need to know the refinement level of each cell across a face. This information is not currently available inside compute kernels. This PR add this data to sparse packs, along with the gid of each current block in the pack, which is also needed for certain algorithms.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


src/outputs/parthenon_xdmf.cpp
```diff
@@ -211,7 +211,6 @@ void genXDMF(std::string hdfFile, Mesh *pm, SimTime *tm, IndexDomain domain, int
   if (swarm_xdmf && all_swarm_info.all_info.size() > 0) {
     std::string sfilename_aux = hdfFile + ".swarm.xdmf";
     std::ofstream pxdmf;
-    hsize_t dims[H5_NDIM] = {0}; // zero-initialized
```
brryan, line 214: unused variable warning

brryan:
There's currently no testing for this, I wasn't sure where the best place for unit tests might be. The sparse pack unit test doesn't initialize the necessary data in `MeshBlock`s, and I wasn't sure how to get `MeshBlock`s out of the forest machinery in `test_forest.cpp`

src/interface/sparse_pack_base.cpp
```diff
@@ -165,6 +165,10 @@ SparsePackBase SparsePackBase::Build(T *pmd, const PackDescriptor &desc,
   pack.bounds_ = bounds_t("bounds", 2, nblocks, nvar + 1);
   pack.bounds_h_ = Kokkos::create_mirror_view(pack.bounds_);
 
+  // This array stores refinement levels of current block and all neighboring blocks.
+  pack.block_props_ = block_props_t("block_props", nblocks, 2, 3, 3, 3);
```
brryan: Having a `1,3,3,3` block of memory just to store the current block's `gid` is wasteful but we would need to go to a `4,4,4` matrix to store all possible neighbor `gid`s and I don't know that that information is needed anywhere currently.

brryan:
This would obviate #1010

src/interface/sparse_pack_base.cpp
```diff
@@ -183,6 +187,19 @@ SparsePackBase SparsePackBase::Build(T *pmd, const PackDescriptor &desc,
       coords_h(b) = pmbd->GetBlockPointer()->coords_device;
     }
 
+    // This block's refinement level stored in central (0, 1,1,1) element of block
+    // properties
+    pack.block_props_h_(blidx, 0, 1, 1, 1) = pmbd->GetBlockPointer()->loc.level();
+    // This block's gid stored in central (1, 1, 1, 1) element
+    pack.block_props_h_(blidx, 1, 1, 1, 1) = pmbd->GetBlockPointer()->gid;
+    for (auto &neighbor : pmbd->GetBlockPointer()->neighbors) {
+      // Multiple refined neighbors will write to the same index but they will always have
```
lroberts36:
```suggestion
      // Multiple refined neighbors may write to the same index but they will always have
```

src/interface/sparse_pack_base.cpp
lroberts36: One option would be to make this a `ParArray2D<int>` with shape `(nblock, 27 + 1)` array and flatten the index space in the `GetLevel` calls. 

If it was flattened, another option would be to fold this into `bounds_` by changing its shape to `(2, nblocks, nvar + 1 + 27)`. This would eliminate the added `deep_copy` but keeps the same amount of extra storage as you have now.

lroberts36: At least as the `gid`s are assigned currently, I think we could store one of the neighbor block gids and recover the rest since they should be in z-order within  the coarse parent block. Not sure this matters as you point out.

src/outputs/parthenon_xdmf.cpp
lroberts36: Ah, glad someone finally fixed this.

lroberts36 (approved):
This LGTM, but a few comments below about how we might save some space or remove the extra deep copy.

src/interface/sparse_pack_base.cpp
pdmullen: Thanks for also adding the `gid` to the sparse pack @brryan! 😸 Now that we are storing block refinement levels in packs, it seems only natural to also store the `gid`. 

This MR should then supersede #1010.  One of the original arguments in favor of using `VariableState` for `gid` was to avoid added bloat to the `SparsePack`.  To accommodate the `gid` here in SparsePacks, it seems that we are being a bit superfluous with the ParArray memory.... what do you think of @lroberts36 idea to shift to a flattened space?

pdmullen (approved):

src/interface/sparse_pack_base.cpp
brryan: Fixed :+1:

src/interface/sparse_pack_base.cpp
brryan: I followed @lroberts36 suggestion to make this a `ParArray2D`. This reduces memory footprint but not number of deep copies. I decided the code would get too ugly with smooshing bounds and gid and block properties together. Maybe we can revisit in the future especially if we keep adding data to the sparse pack. 

Still only storing the `gid` of the current meshblock, not neighbors of the current meshblock.

src/outputs/parthenon_xdmf.cpp
brryan: 🫡

brryan:
> There's currently no testing for this, I wasn't sure where the best place for unit tests might be. The sparse pack unit test doesn't initialize the necessary data in `MeshBlock`s, and I wasn't sure how to get `MeshBlock`s out of the forest machinery in `test_forest.cpp`

Still untested but I did experiment in the `fine_advection` example that the GID and Level functions I added both don't segfault and return not-obviously-wrong data on both host and device.

src/interface/sparse_pack_base.cpp
lroberts36: Yeah, I don't have a sense for whether or not these deep copies matter. My guess is they are only likely to have an impact in codes that employ sparse fields (since the packs are likely to be rebuilt more often there). I agree that it is cleaner to separate them out for now.

brryan: Good point about this only being an issue in codes with sparse fields. Hopefully the smallness of the data makes this issue negligible.

PR #1172, opened 2024-09-10
make parthenon manager play nicely with other MPI libraries

Yurlungur:
## PR Summary

A downstream code wants to interface with PetSC, which also interacts with MPI via MPI init and finalize. This MR makes that interaction *and only that interaction* safe from Parthenon's perspective. 

I haven't thought fully how we interact with other libraries so I don't know if we are safe to interop with, e.g., PetSC. Also in the grand scheme of things, what this downstream code wants to do with PetSC parthenon can do natively. So it might be an intermediate step.

That said, I think this at a minimum just makes parthenon a little safer and more robust.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


lroberts36 (approved):
LGTM. I guess it is still possible that some other code calls `MPI_Initialize` after Parthenon calls it, but I think this is the best we can do from our perspective.

Yurlungur:
> LGTM. I guess it is still possible that some other code calls `MPI_Initialize` after Parthenon calls it, but I think this is the best we can do from our perspective.

Yeah if all relevant codes have guards like this, then order won't matter. If only we are careful, then it will.

pdmullen (approved):

PR #1173, opened 2024-09-11
Add sanity check so ParameterInput is not allowed to be different on different MPI ranks

Yurlungur:
## PR Summary

Multiple times we've been hit by the issue that `ParameterInput` is stateful because `GetOrAdd` modifies the object. If `ParameterInput` is different on different MPI ranks, then HDF5 output will hang, because writing to params is a collective action.

This is a minimal fix that at least helps debugging when we hit this issue. I add the ability to compute the hash of `ParameterInput`. Then, before output in HDF5 I check that this hash is the same on all MPI ranks with one MPI broadcast and one MPI reduce.

I added tests for the hashing machinery in the unit tests. I couldn't test the MPI bit in the unit tests but I did test it by hand by modifying one of the examples to add a param on only rank zero and got the desired behavior.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/src/outputs/output_utils.cpp
+++ b/src/outputs/output_utils.cpp
@@ -30,6 +30,7 @@
 #include "mesh/mesh_refinement.hpp"
 #include "mesh/meshblock.hpp"
 #include "outputs/output_utils.hpp"
+#include "parameter_input.hpp"
 
 namespace parthenon {
 namespace OutputUtils {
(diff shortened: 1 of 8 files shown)


src/outputs/output_utils.cpp
```diff
@@ -316,11 +317,42 @@ std::size_t MPISum(std::size_t val) {
...
+  if (Globals::my_rank == 0) {
+    PARTHENON_REQUIRE_THROWS(
+        pinput_same_accross_ranks,
+        "Parameter input object must be the same on every rank, otherwise I/O may "
+        "be\n\t\t"
+        "unable to write it safely. If you reached this error message, look to make "
+        "sure\n\t\t"
+        "that your calls to functions that look like pin->GetOrAdd are all called\n\t\t"
+        "exactly the same way on every MPI rank.");
+  }
```
lroberts36: Don't think it matters, but isn't this reduction unnecessary since you could do the `PARTHENON_REQUIRE_THROWS` on each rank?

lroberts36 (approved):
LGTM, this will definitely prevent some bug hunts in the future! 

I don't have a sense for how well the Boost hash combine function works so I worry a little about possible hash collisions. That being said, I don't see a better way to do things without a lot more work.

src/outputs/output_utils.cpp
Yurlungur: yes that's true. Saving one reduce is probably worth it. I will change it.

Yurlungur:
> I don't have a sense for how well the Boost hash combine function works so I worry a little about possible hash collisions. That being said, I don't see a better way to do things without a lot more work.

Yeah hash collisions are possible. But I figured this is better than nothing.

c-prather (approved):
LGTM.  I am not worried about collisions at all, as the two `ParameterInput`s will be highly correlated with a single mistaken change, not drawn randomly.  Since the combine op is non-associative, you'd need to find a parameter that hashes to the *inverse* of another, not just the same thing. i.e. a genuine collision in the underlying function.

I could imagine someday synchronizing automatically to rank 0 rather than just erroring, but that sounds like a lot of code to maintain for what amounts to hand-holding

Yurlungur:
> I could imagine someday synchronizing automatically to rank 0 rather than just erroring, but that sounds like a lot of code to maintain for what amounts to hand-holding

I considered implementing that---I don't actually think it would be too difficult, but I was concerned this might not be the desired behavior. I think it's better to die and make the developer/user figure out what's wrong.

PR #1174, opened 2024-09-12
[Breaking] Linear Solver Updates

lroberts36:
## PR Summary

This PR adds a number of solver related things:
- Adds a CG solver. I am not sure how useful this is since it seems to work poorly with AMR and multigrid preconditioning, but I think it is at least nice to have in there for testing.
- Adds simple diagonal preconditioners for CG and BiCGSTAB
- Adds the option to use a user defined task for prolongation during multigrid solves instead of the prolongation facilities provided by the Parthenon infrastructure. This more easily allows for physical boundary aware prolongation operators. 
- Adds a number of different boundary aware prolongation operators (constant, bilinear, and Kwak) to the `poisson_gmg` example.
- Adds explicit setting of boundary fluxes and fluxes at refinement boundaries in the `poisson_gmg` example.
- Adds a base class for solvers to simplify switching between solvers for downstream codes
- Switches to stage based solvers that can include multiple separate variables as part of the state vector. **This is a breaking change for any code that is using solvers downstream.**

To facilitate these changes to the solvers and examples, two small changes are made to the rest of the infrastructure:
- Extends the neighbor information that @brryan put into `SparsePack`s to work with multigrid neighbors and include physical boundary information as well. *This is a breaking change, since now physical boundaries are denoted with a large, negative level in the array containing neighbor information in `SparsePack`*
- Adds a mechanism for getting a flat index from a multi-d `Indexer`.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [ ] Docs build


example/poisson_gmg/poisson_equation.hpp
```diff
@@ -164,9 +184,128 @@ class PoissonEquation {
     return TaskStatus::complete;
   }
 
+  template <class... var_ts>
+  parthenon::TaskID Prolongate(parthenon::TaskList &tl, parthenon::TaskID depends_on,
```
Yurlungur: Not sure I love moving this into the poisson example. Any way we can keep at least a default prolongation type available publicly?

lroberts36: The default Parthenon prolongation is used by default, the user defined prolongation within the equation class only gets called if you set `solver/interior_prolongation=User`.

Yurlungur: Will the default parthenon prolongation work for multigrid?

lroberts36: The default will be the same as it was before this PR, `refinement_ops::ProlongateSharedLinear`, which works for multigrid (although it may not be the optimal prolongation operator). I think what I said before was unclear.

lroberts36: The switch between the Parthenon prolongation infrastructure and the user defined prolongation infrastructure is here: https://github.com/parthenon-hpc-lab/parthenon/blob/d554036717275c584991fd33742fca68f37c76b7/src/solvers/mg_solver.hpp#L488

I think it makes more sense to allow for a separate prolongation implementation in this case because it is easier if the prolongation operator is explicitly physical boundary aware.

Yurlungur: Yeah that makes sense. I assume bounds aware prolongation/restriction ops have to be implemented downstream... doesn't make much sense to implement them upstream.

doc/sphinx/src/solvers.rst
Yurlungur, line 1: :+1: for more detailed docs.

example/poisson_gmg/poisson_driver.cpp
```diff
@@ -68,54 +61,53 @@ TaskCollection PoissonDriver::MakeTaskCollection(BlockList_t &blocks) {
...
     if (use_exact_rhs) {
-      auto copy_exact = tl.AddTask(get_rhs, TF(solvers::utils::CopyData<exact, u>), md);
-      auto comm = AddBoundaryExchangeTasks<BoundaryType::any>(copy_exact, tl, md, true);
-      PoissonEquation eqs;
-      eqs.do_flux_cor = flux_correct;
-      get_rhs = eqs.Ax<u, rhs>(tl, comm, md);
+      auto copy_exact = tl.AddTask(
+          copy_rhs, TF(solvers::utils::between_fields::CopyData<exact, u>), md);
+      copy_exact = tl.AddTask(
+          copy_exact, TF(solvers::utils::CopyData<parthenon::TypeList<u>>), md, md_u);
```
Yurlungur: this pattern is a bit weird. Would prefer if the task ids had different names.

example/poisson_gmg/poisson_driver.cpp
```diff
@@ -68,54 +61,53 @@ TaskCollection PoissonDriver::MakeTaskCollection(BlockList_t &blocks) {
...
-      auto setup = bicgstab_solver->AddSetupTasks(tl, zero_u, i, pmesh);
-      solve = bicgstab_solver->AddTasks(tl, setup, pmesh, i);
-    } else if (solver == "MG") {
-      auto setup = mg_solver->AddSetupTasks(tl, zero_u, i, pmesh);
-      solve = mg_solver->AddTasks(tl, setup, pmesh, i);
-    } else {
-      PARTHENON_FAIL("Unknown solver type.");
-    }
+    auto zero_u = tl.AddTask(copy_rhs, TF(solvers::utils::SetToZero<u>), md);
+    zero_u = tl.AddTask(zero_u, TF(solvers::utils::SetToZero<u>), md_u);
```
Yurlungur: why is `SetToZero` called twice?

src/solvers/solver_utils.hpp
```diff
@@ -177,7 +179,79 @@ TaskStatus CopyData(const std::shared_ptr<MeshData<Real>> &md) {
...
+        const int nvars = pack_in.GetUpperBound(b) - pack_in.GetLowerBound(b) + 1;
+        for (int c = 0; c < nvars; ++c) {
+          Real *in = &pack_in(b, te, c, kb.s, jb.s, ib.s);
+          Real *out = &pack_out(b, te, c, kb.s, jb.s, ib.s);
+          parthenon::par_for_inner(DEFAULT_INNER_LOOP_PATTERN, member, 0,
+                                   npoints_inner - 1,
+                                   [&](const int idx) { out[idx] = in[idx]; });
+        }
+      });
+  return TaskStatus::complete;
```
Yurlungur, line 213: :+1: for evil riot-style fused loop.

src/solvers/solver_utils.hpp
```diff
@@ -177,7 +179,79 @@ TaskStatus CopyData(const std::shared_ptr<MeshData<Real>> &md) {
...
+  int nblocks = md->NumBlocks();
+  using TE = parthenon::TopologicalElement;
+  TE te = TE::CC;
+  static auto desc = [&] {
+    if constexpr (isTypeList<TL>::value) {
+      return parthenon::MakePackDescriptorFromTypeList<TL>(md.get());
+    } else {
+      return parthenon::MakePackDescriptor<TL>(md.get());
+    }
+  }();
```
Yurlungur, line 227: oh this is a neat trick for getting an automatic type.

src/solvers/solver_utils.hpp
```diff
@@ -177,7 +179,79 @@ TaskStatus CopyData(const std::shared_ptr<MeshData<Real>> &md) {
...
+        auto cb = GetIndexShape(pack(b, te, 0), ng);
+        const auto &coords = pack.GetCoordinates(b);
+        IndexRange ib = cb.GetBoundsI(IndexDomain::interior, te);
+        IndexRange jb = cb.GetBoundsJ(IndexDomain::interior, te);
+        IndexRange kb = cb.GetBoundsK(IndexDomain::interior, te);
+        const int nvars = pack.GetUpperBound(b) - pack.GetLowerBound(b) + 1;
+        for (int c = 0; c < nvars; ++c) {
+          parthenon::par_for_inner(
+              parthenon::inner_loop_pattern_simdfor_tag, member, kb.s, kb.e, jb.s, jb.e,
+              ib.s, ib.e, [&](int k, int j, int i) { pack(b, te, c, k, j, i) = 0.0; });
```
Yurlungur, line 245: Is there a reason this can't use a fused loop?

src/interface/sparse_pack_base.hpp
```diff
@@ -64,6 +64,8 @@ class SparsePackBase {
   using block_props_h_t = typename block_props_t::HostMirror;
   using coords_t = ParArray1DRaw<ParArray0D<Coordinates_t>>;
 
+  static constexpr int bnd_flag = -2000;
```
Yurlungur: what is this?

src/solvers/cg_solver.hpp
```diff
@@ -0,0 +1,282 @@
...
+
+    // 7. Check and print out residual
+    auto get_res = DotProduct<FieldTL>(correct_r, itl, &residual, md_r, md_r);
+
+    auto print = itl.AddTask(
+        TaskQualifier::once_per_region, get_res,
+        [&](CGSolver *solver, Mesh *pmesh) {
+          Real rms_res = std::sqrt(solver->residual.val / pmesh->GetTotalCells());
+          if (Globals::my_rank == 0 && solver->params_.print_per_step)
+            printf("%i %e\n", solver->iter_counter, rms_res);
```
Yurlungur: I find reading output like this can be difficult. I like to add tab offsets.
```suggestion
           printf("\t%i %e\n", solver->iter_counter, rms_res);
```

Yurlungur (approved):
LGTM. This seems to be a pretty significantly cleanup along with the new features and better performance.

src/solvers/solver_utils.hpp
lroberts36: yeah, this should possibly move to `MakePackDescriptor` so that `MakePackDescriptor` automatically determines if you passed it a `TypeList` and we don't need the second overload.

Yurlungur: good idea. That can wait though.

src/solvers/solver_utils.hpp
lroberts36: it is over `IndexDomain::interior`

src/interface/sparse_pack_base.hpp
lroberts36: Sparse pack stores the refinement level of each blocks neighbors in every direction, which Ben R. added for swarms. Now, if the level is set to -2000 (which I changed to `physical_bnd_flag`), it denotes that a given side of the block abuts a physical boundary. This is necessary for the boundary aware prolongation routines. This number is negative and large enough it will never collide with an actual refinement level.

example/poisson_gmg/poisson_driver.cpp
lroberts36: why is this weird? It just copies from one field to another on the base stage then copies that field from base stage to another stage. happy to change the task name though.

example/poisson_gmg/poisson_driver.cpp
lroberts36: leftover from when both staged and unstaged solvers were available. removed.

example/poisson_gmg/poisson_driver.cpp
Yurlungur: I just found the fact that both tasks had the same name confusing. Not a big deal

src/solvers/solver_utils.hpp
Yurlungur: I take it we care about not overwiting the ghosts?

lroberts36: yeah, since this is a general utility function I think it is better to just set the interior. If we ever see it as a bottleneck we can revisit.

c-prather (approved):
I've been using this as-is in KHARMA for ages now, it works for me.  The code looks great, I like the simpler interface.

PR #1175, opened 2024-09-12
WIP: Add boundary comm inner loop specializations for performance and allow setting vector length

lroberts36:
## PR Summary

This PR adds various inner loop specializations to avoid calling reductions, coordinate transformations, and masks in the inner boundary buffer setting loops where possible. I found this made a little bit of a difference for the multigrid communication, not sure that it really matters elsewhere.

It also exposes a flag `parthenon/mesh/comm_vector_length` which sets the vector length of the outer team policy. We have always used the Kokkos default value here (which is one). I am not sure if changing this value has any impact, but I at least wanted to be able to play around with it. 

As an aside, once #1142 is in, we may want to switch these loops to `par_for`s instead of raw Kokkos loops. 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [ ] Docs build

### DIFF
--- a/src/mesh/mesh.cpp
+++ b/src/mesh/mesh.cpp
@@ -77,6 +77,8 @@ Mesh::Mesh(ParameterInput *pin, ApplicationInput *app_in, Packages_t &packages,
       default_pack_size_(pin->GetOrAddInteger("parthenon/mesh", "pack_size", -1)),
       // private members:
       num_mesh_threads_(pin->GetOrAddInteger("parthenon/mesh", "num_threads", 1)),
+      comm_vector_length_(
+          pin->GetOrAddInteger("parthenon/mesh", "comm_vector_length", 1)),
       use_uniform_meshgen_fn_{true, true, true, true}, lb_flag_(true), lb_automatic_(),
       lb_manual_(), nslist(Globals::nranks), nblist(Globals::nranks),
       nref(Globals::nranks), nderef(Globals::nranks), rdisp(Globals::nranks),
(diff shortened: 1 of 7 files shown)


src/mesh/mesh.cpp
```diff
@@ -77,6 +77,8 @@ Mesh::Mesh(ParameterInput *pin, ApplicationInput *app_in, Packages_t &packages,
       default_pack_size_(pin->GetOrAddInteger("parthenon/mesh", "pack_size", -1)),
       // private members:
       num_mesh_threads_(pin->GetOrAddInteger("parthenon/mesh", "num_threads", 1)),
+      comm_vector_length_(
+          pin->GetOrAddInteger("parthenon/mesh", "comm_vector_length", 1)),
```
Yurlungur, line 85: do you have a sense for what the right default size is here?

Yurlungur (approved):
LGTM! This is a good change that should hopefully help with performance.

src/mesh/mesh.cpp
lroberts36: Not at all, in fact I am not sure if it makes any difference. That being said, the Kokkos default is one, so I just put that here.

pdmullen (approved):

lroberts36:
This PR is out of date with @bprather changes in #1196, as well as some other changes. It probably is still worth including at some point, but the required changes are pretty simple to just put in a new PR.

Yurlungur:
@lroberts36 what is the status of this MR?

lroberts36:
> @lroberts36 what is the status of this MR?

I am going to close it without merging and create an issue to add these changes in the future.

PR #1179, opened 2024-09-21
Make a global variable for whether simulation is a restart

brryan:
## PR Summary

It would be convenient downstream to be able to access `is_restart` as a global variable (when determining whether input parameters are set or not set appropriately only during initialization) -- right now whether we're restarted is hard to access outside driver classes or `ParthenonManager`. This PR adds `Globals::is_restart` set at the same time as `Globals::my_rank`. 

Along the way I also added a `soft_equiv` function to `utils/robust.hpp` which would also be useful downstream. 

I also renamed `tst/unit/kokkos_abstraction.cpp` -> `tst/unit/test_kokkos_abstraction.cpp` for consistency. Stop me if there's a reason to not do that.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


src/globals.cpp
```diff
@@ -30,8 +30,9 @@ namespace Globals {
 int nghost;
 
 // all of these global variables are set at the start of main():
-int my_rank; // MPI rank of this process
-int nranks;  // total number of MPI ranks
+int my_rank;     // MPI rank of this process
+int nranks;      // total number of MPI ranks
+bool is_restart; // Whether this simulation is restarted from a checkpoint file
```
adamdempsey90, line 35: Slight preference for `Globals::restart` over `Globals::is_restart`

src/utils/robust.hpp
```diff
@@ -60,6 +60,24 @@ KOKKOS_INLINE_FUNCTION auto ratio(const A &a, const B &b) {
...
+    return true;
+  } else {
+    if (std::fabs(ref) < std::numeric_limits<Real>::min() && pass_on_small) {
+      // Optionally return true if reference value is close to zero
+      return true;
+    } else {
+      return false;
+    }
+  }
+}
```
adamdempsey90: Is this performant enough to be used on device?

adamdempsey90 (approved):

src/utils/robust.hpp
Yurlungur: I assume this is mostly intended for testing, so performance probably not too important.

That said, is branching actually necessary here? Can this not be written as a compound boolean? The latter almost certainly is performant enough to be used on device, and compound expressions are lazily evaluated. So could do...

```suggestion
// Return true equivalence if value and reference differ by less than precision
// Optionally return true if reference value is close to zero
KOKKOS_FORCEINLINE_FUNCTION
bool SoftEquiv(const Real &val, const Real &ref,
                const Real eps = 10. * std::numeric_limits<Real>::epsilon(),
                const bool pass_on_small = true) {
  return ((std::fabs(val - ref) < eps * std::fabs(ref)) 
    || ((std::fabs(ref) <std::numeric_limits<Real>::min()) && pass_on_small));
}
```
Also changed name to parthenon style ;)

Yurlungur (approved):
LGTM

src/globals.cpp
brryan: `bool is_*` is a fairly common pattern in parthenon, and since this variable won't be appearing regularly in complex expressions I'm inclined to leave as is unless a majority pushes back on this

src/utils/robust.hpp
brryan: Good point @Yurlungur, I adopted your changes. This is is actually a use case for this in a compute intensive loop so this is a worthwhile change. Thanks also for catching the name style issue.

Yurlungur: As discussed on MM to make this more readable, you could also give the bools names, something like:
```C++
// Return true equivalence if value and reference differ by less than precision
// Optionally return true if reference value is close to zero
KOKKOS_FORCEINLINE_FUNCTION
bool SoftEquiv(const Real &val, const Real &ref,
                const Real eps = 10. * std::numeric_limits<Real>::epsilon(),
                const bool pass_on_small = true) {
  const bool is_close = std::fabs(val - ref) < eps * std::fabs(ref);
  const bool is_small = std::fabs(ref) <std::numeric_limits<Real>::min();
  return (is_close || (is_small && pass_on_small));
}
```
which is a bit more readable I think.

Also it just occurred to me: Should we check that both the value and the reference are small for `pass_on_small`? If so we could do:
```C++
// Return true equivalence if value and reference differ by less than precision
// Optionally return true if reference value is close to zero
KOKKOS_FORCEINLINE_FUNCTION
bool SoftEquiv(const Real &val, const Real &ref,
                const Real eps = 10. * std::numeric_limits<Real>::epsilon(),
                const bool pass_on_small = true) {
  constexpr Real tiny = std::numeric_limits<Real>::min();
  const bool is_close = std::abs(val - ref) < eps * std::abs(ref);
  const bool ref_is_small = std::abs(ref) < tiny;
  const bool val_is_small = std::abs(val) < tiny;
  const bool is_small = ref_is_small && val_is_small;
  return (is_close || (is_small && pass_on_small));
}
```
also I *think* that in C++ `fabs` and `abs` are templated to have identical behavior for identical inputs? Not sure?

src/globals.cpp
Yurlungur: I don't feel strongly. I think I have a slight preference for `bool is_*`.

src/utils/robust.hpp
brryan: You're right about `fabs` -- that's just my C99 instincts. Also agree with the `const bool is_` approach, nice suggestion.

I think the logic behind `pass_on_small` is that if the reference value is `0`, the comparison isn't well-defined, so the person using this function can decide either to let those cases all pass or all fail. That isn't quite the same behavior as checking that both `val` and `ref` are small, and I think the former is more useful.

Yurlungur: Yeah I'm fine either way regarding the behavior for `pass_on_small` my main logic was if the reference value is tiny, the correct answer is the test value should be too. Either way though.

PR #1184, opened 2024-10-02
Fix swarm block neighbor indexing in 1D, 2D

brryan:
## PR Summary

For some cases, the current block of particles was being computed incorrectly. This was because the `swarm_comms.cpp` initialization machinery would overwrite the region of `neighbor_indices_h` belonging to the currently active meshblock. I fixed this, and also changed the index we use to compute neighbor indices for unused dimensions in `SwarmDeviceContext` to improve consistency.

This should fix the downstream issue @pdmullen.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Current develop
 
 ### Added (new features/APIs/variables/...)
+- [[PR 1184]](https://github.com/parthenon-hpc-lab/parthenon/pull/1184) Fix swarm block neighbor indexing in 1D, 2D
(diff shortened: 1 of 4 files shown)


Yurlungur (approved):
Pretty trivial. LGTM. Is there a unit test we could add to catch this?

brryan:
> Pretty trivial. LGTM. Is there a unit test we could add to catch this?

I don't think this is easy to do as a unit test because ultimately we need to call `Mesh::Initialize()` to set up the data structures used by the swarm to compute neighbor blocks, but I added a check to the `particles` example that makes sure initialized particles report that they are on the current meshblock.

pdmullen (approved):

PR #1177, opened 2024-09-13
[Breaking Change] Make Mesh-Level Boundary Conditions Usable without "User" flag (now without bitrot)

Yurlungur:
## NOTICE

This is my second attempt at #989 . The code has changed so much since my original implementation that it was easier to start from scratch rather than trying to merge with develop. Some notes:

- Block boundaries are used by forest of octrees. I now leave those alone.
- At @pgrete 's request, reflecting boundaries are no longer automatically registered. You may register them with the BC infrastructure however.
- This MR depends on #1176 . Please merge that one first. The diff will also be a lot easier to read after that change.

## API Breaking change

When registering user boundary functions, the mechanism changes from assigning the function pointer to an array. You now must call `ApplicationInput::RegisterBoundaryCondition` and give the custom boundary condition a name. This also eliminates the need for the `user` boundary flag in the input file, although that flag is implicitly still supported.

## PR Summary

A discussion on the new downstream code headed by @brryan @pdmullen indicated a desire to set mesh-level boundary conditions without using the `user` flag. Here I implement this feature. The changes I made are:
1. `ApplicationInput` now has a registration function where you register either a grid or particle boundary condition, with a function pointer (or a class in the case of swarm bounds) and a name.
2. ALL boundary conditions (except periodic) use this mechanism. Reflecting boundaries are, for example, registered in the constructor for `ApplicationInput` with the `"reflecting"` name.
3. `BoundaryFlag`s are dramatically simplified, indicating (essentially) only whether or not boundaries are periodic.
4. The mesh no longer uses flags to set function pointers. Instead, it just pulls the functions from `ApplicationInput`.
5. For (sort of) backwards compatibility, if you register a function without naming it, it's named "user."
6. Swarm boundaries are now decoupled from mesh boundaries. Although by default, they are set to the same value as the mesh.

The advantage of this is that you can now call (in your main function)
```C++
pman->app_input.RegisterBoundaryCondition(parthenon::BoundaryFace::inner_x1, "my_bc", &my_bc_func);
```
and then in the input file just use
```
<parthenon/mesh>
ix1_bc = my_bc
```

I think the refactored code is also significantly cleaner and easier to read. Also I put a link to the parthenon paper in the readme.

## PR Checklist

- [x] Code passes cpplint
- [x] New features are documented.
- [x] Adds a test for any bugs fixed. Adds tests for new features.
- [x] Code is formatted
- [x] Changes are summarized in CHANGELOG.md
- [x] Change is breaking (API, behavior, ...)
[...]

PR #1185, opened 2024-10-05
Bugfix to particle defragmentation

brryan:
## PR Summary

There was a bug in the defragmentation routine acting on particle lists. The issue was that empty indices weren't being updated after defragmentation. This PR fixes that.

As an aside: this produced the surprising effect that simulation correctness depended on frequency of simulation output, even when no particle fields were being output. This is because particle defragmentation is called during every output cycle. Not saying we can't do this, just mentioning it because it was unexpected. Shouldn't be a correctness issue if defragmentation actually works :)

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/example/particles/particles.cpp
+++ b/example/particles/particles.cpp
@@ -62,6 +62,11 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
       destroy_particles_frac >= 0. && destroy_particles_frac <= 1.,
       "Fraction of particles to destroy each timestep must be between 0 and 1");
 
+  Real defrag_threshold = pin->GetOrAddReal("Particles", "defrag_threshold", 0.9);
+  pkg->AddParam<>("defrag_threshold", defrag_threshold);
+  PARTHENON_REQUIRE(defrag_threshold >= 0. && defrag_threshold <= 1.,
+                    "Defragmentation threshold must be between 0 and 1 inclusive.");
+
   std::string deposition_method =
       pin->GetOrAddString("Particles", "deposition_method", "per_particle");
   if (deposition_method == "per_particle") {
@@ -609,6 +614,9 @@ TaskCollection ParticleDriver::MakeFinalizationTaskCollection() const {
     auto &sc1 = pmb->meshblock_data.Get();
     auto &tl = async_region1[i];
 
+    auto pkg = pmb->packages.Get("particles_package");
+    const auto defrag_threshold = pkg->Param<Real>("defrag_threshold");
+
     auto destroy_some_particles = tl.AddTask(none, DestroySomeParticles, pmb.get());
 
     auto sort_particles = tl.AddTask(destroy_some_particles,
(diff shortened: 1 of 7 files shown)


src/interface/swarm.cpp
```diff
@@ -354,15 +354,18 @@ void Swarm::Defrag() {
   auto &mask = mask_;
 
   const int &num_active = num_active_;
+  auto empty_indices = empty_indices_;
   parthenon::par_scan(
-      "Set empty indices prefix sum", 0, nmax_pool_ - num_active_ - 1,
-      KOKKOS_LAMBDA(const int nn, int &update, const bool &final) {
-        const int n = nn + num_active;
+      "Set empty indices prefix sum", num_active, nmax_pool_ - 1,
```
brryan, line 359: Not a change to behavior, this is just clearer IMO

src/interface/swarm.cpp
```diff
@@ -354,15 +354,18 @@ void Swarm::Defrag() {
...
+      KOKKOS_LAMBDA(const int n, int &update, const bool &final) {
         const int val = mask(n);
         if (val) {
           update += 1;
         }
-        if (final) scan_scratch_toread(n) = update;
+        if (final) {
+          scan_scratch_toread(n) = update;
+          empty_indices(n - num_active) = n;
```
brryan, line 367: The bugfix

example/particles/particles.cpp
```diff
@@ -62,6 +62,11 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
       destroy_particles_frac >= 0. && destroy_particles_frac <= 1.,
       "Fraction of particles to destroy each timestep must be between 0 and 1");
 
+  Real defrag_threshold = pin->GetOrAddReal("Particles", "defrag_threshold", 0.9);
```
brryan, line 65: Useful to be able to control this at runtime.

pdmullen (approved):
LGTM! The defrag in outputs happens even when particles aren't being dumped, which is also likely undesirable.

example/particles/particles.cpp
Yurlungur: Should we create a `parthenon/particles` input block and have this read in there? We could store this in a specialized `parthenon/particles` package or in the mesh.

(Actually I think this input block already exists... I think we now use it for boundaries... at least with #1177 )

Yurlungur (approved):
Regarding defragging in output. We don't *have* to do this. I could make the output machinery not use it. But it was convenient.

example/particles/particles.cpp
brryan: Defrag doesn't actually get called by default in tasklists or by the particles infrastructure itself, so I think it's reasonable to leave this up to downstream applications since they will already be deciding when/where to call Defrag when they write their tasking.

PR #1188, opened 2024-10-10
Fix: Bounds in output for Metadata::None variables

AstroBarker:
In updating `phoebus` to use up-to-date `parthenon` I ran into an issue where a `Metadata::None` variable was failing to output due to view bounds accesses. There seems to have been an off-by-one error in the relevant code block in `output_utils.hpp:313`. I've updated the bounds there to correctly output `Metadata::None` variables. 

To test this, I modified the advection example to have an optional `Metadata::None` variable (enabled in the input deck `<Advection>/test_metadata_none`). I set the output_hdf5 test to use this variable.

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build

### DIFF
--- a/example/advection/advection_package.cpp
+++ b/example/advection/advection_package.cpp
@@ -196,6 +196,20 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
   m = Metadata({Metadata::Cell, Metadata::OneCopy}, std::vector<int>({1}));
   pkg->AddField("my_derived_var", m);
 
+  // Create a Metadata::None variable for IO testing purposes.
+  // Only load if test_metadata_none is specified in the Advection block
+  auto test_metadata_none =
+      pin->GetOrAddBoolean("Advection", "test_metadata_none", false);
+  pkg->AddParam<bool>("test_metadata_none", test_metadata_none);
+  if (test_metadata_none) {
+    const int nx1 = pin->GetOrAddInteger("parthenon/mesh", "nx1", 1);
+    const int nx2 = pin->GetOrAddInteger("parthenon/mesh", "nx2", 1);
+    const int nx3 = pin->GetOrAddInteger("parthenon/mesh", "nx3", 1);
+    std::vector<int> test_shape = {nx1 + 1, nx2 + 1, nx3 + 1, 3};
+    m = Metadata({Metadata::OneCopy, Metadata::None}, test_shape);
+    pkg->AddField("metadata_none_var", m);
+  }
+
   // List (vector) of HistoryOutputVar that will all be enrolled as output variables
   parthenon::HstVar_list hst_vars = {};
   // Now we add a couple of callback functions
@@ -281,6 +295,7 @@ AmrTag CheckRefinement(MeshBlockData<Real> *rc) {
 void PreFill(MeshBlockData<Real> *rc) {
   auto pmb = rc->GetBlockPointer();
   auto pkg = pmb->packages.Get("advection_package");
+  const bool test_metadata_none = pkg->Param<bool>("test_metadata_none");
   bool fill_derived = pkg->Param<bool>("fill_derived");
 
   if (fill_derived) {
@@ -300,6 +315,25 @@ void PreFill(MeshBlockData<Real> *rc) {
         PARTHENON_AUTO_LABEL, 0, num_vars - 1, kb.s, kb.e, jb.s, jb.e, ib.s, ib.e,
         KOKKOS_LAMBDA(const int n, const int k, const int j, const int i) {
           v(out + n, k, j, i) = 1.0 - v(in + n, k, j, i);
+        });
+  }
+
+  // Fill the metadata::None var with index gymnastics.
+  if (test_metadata_none) {
+    IndexRange ib = pmb->cellbounds.GetBoundsI(IndexDomain::interior);
+    IndexRange jb = pmb->cellbounds.GetBoundsJ(IndexDomain::interior);
+    IndexRange kb = pmb->cellbounds.GetBoundsK(IndexDomain::interior);
+
+    // packing in principle unnecessary/convoluted here and just done for demonstration
+    std::vector<std::string> vars({"metadata_none_var"});
+    PackIndexMap imap;
+    const auto &v = rc->PackVariables(vars, imap);
+
+    const int ivar = imap.get("metadata_none_var").first;
+    pmb->par_for(
+        PARTHENON_AUTO_LABEL, 0, 2, kb.s, kb.e, jb.s, jb.e, ib.s, ib.e,
+        KOKKOS_LAMBDA(const int n, const int k, const int j, const int i) {
+          v(ivar, n, k, j, i) = n + k * j * i;
         });
   }
 }
--- a/example/advection/advection_package.cpp
+++ b/example/advection/advection_package.cpp
@@ -202,9 +202,9 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
       pin->GetOrAddBoolean("Advection", "test_metadata_none", false);
   pkg->AddParam<bool>("test_metadata_none", test_metadata_none);
   if (test_metadata_none) {
-    const int nx1 = pin->GetOrAddInteger("parthenon/mesh", "nx1", 1);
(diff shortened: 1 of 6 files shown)


pdmullen (approved):
Nice find! Output bugs are always tricky to track down.

Yurlungur (approved):
Thanks for the fix! This breakage is definitely on me.

Yurlungur:
Looks like tests are failing @AstroBarker ?

AstroBarker:
> Looks like tests are failing @AstroBarker ?

Looks like `output_hdf5` so probably it's the gold files

Yurlungur:
almost certainly. run locally and see what the error is. might need to update the gold files.

AstroBarker:
> almost certainly. run locally and see what the error is. might need to update the gold files.

Yeah they pass locally for me only if I update the goldfiles. I added a new output for the `Metadata::None` variable in the hdf5 tests. What's the process for updating the proper gold files?

Yurlungur:
Run `parthenon/tst/regression/gold_standard/make_tarball.sh`. Then create a release from the generated tarball. You will also need to update the tarball version and hash in the `CMake`

example/advection/advection_package.cpp
```diff
@@ -302,6 +317,25 @@ void PreFill(MeshBlockData<Real> *rc) {
...
+    PackIndexMap imap;
+    const auto &v = rc->PackVariables(vars, imap);
+
+    const int ivar = imap.get("metadata_none_var").first;
+    pmb->par_for(
+        PARTHENON_AUTO_LABEL, 0, 2, kb.s, kb.e, jb.s, jb.e, ib.s, ib.e,
+        KOKKOS_LAMBDA(const int n, const int k, const int j, const int i) {
+          v(ivar, n, k, j, i) = n + k * j * i;
+        });
+  }
```
pgrete: I'm not sure what/why this is happening here.
If I read the the field correctly, then it's a vector with 3 components and extents larger than the mesh size by one.
In this `par_for` the block indices are used (without any global offset), so every block will have the same indices set to the same values and a large fraction of the index space will remain 0.
Is that intentional?

pgrete (commented):
In addition to the comment on filling the content of the new var: Should we also add a proper check that this is working, i.e., read the data of the variable to confirm that the data written is looking what it's supposed to look like?

example/advection/advection_package.cpp
AstroBarker: Good catch, I didn't pay much mind to the contents of the fields, only that its shape reproduced the crash. This should be fixed to be filled now, I think.

example/advection/advection_package.cpp
```diff
@@ -302,6 +317,28 @@ void PreFill(MeshBlockData<Real> *rc) {
...
+    const auto &v = rc->PackVariables(vars, imap);
+
+    const int ivar_lo = imap.get("metadata_none_var").first;
+    const int ivar_hi = imap.get("metadata_none_var").second;
+    pmb->par_for(
+        PARTHENON_AUTO_LABEL, 0, 2, 0, nx3, 0, nx2, 0, nx1,
+        KOKKOS_LAMBDA(const int n, const int k, const int j, const int i) {
+          v(ivar_lo, n, k, j, i) = n + k * j * i;
+          v(ivar_lo + 1, n, k, j, i) = 1 + n + k * j * i;
+          v(ivar_hi, n, k, j, i) = 2 + n + k * j * i;
```
pgrete: i just looked at the updated loop.
While I understand the fix going from mesh to meshbock, I now don't get the `ivar_lo`/`ivar_hi` indexing.
What's supposed to happen here? I could see this working if `ivar_lo`/`ivar_hi replace the `n` index.

tst/regression/test_suites/output_hdf5/parthinput.advection
```diff
@@ -64,7 +65,8 @@ file_type = hdf5
 dt = 1.0
 variables = advected, one_minus_advected, & # comments are ok
             one_minus_advected_sq, & # on every (& characters are ok in comments)
-            one_minus_sqrt_one_minus_advected_sq # line
+            one_minus_sqrt_one_minus_advected_sq, & # line
+            metadata_none_var
```
pgrete, line 69: Given that this is now in the output file, how about reading the content to make sure that the data written is actually there.
I'm not sure we have that kind of test for `Meshdata::None` already.

example/advection/advection_package.cpp
pgrete: Any comment on this? ping @AstroBarker 

I'm happy to merge this without adding test to read the data (which would be great), but I'd like to understand this loop before merging as from my current understanding it might seem like being an unintended bug.

AstroBarker: I've updated the loop, removing the `ivar_lo/hi` stuff, which I agree was not necessary in the loop, and added a section to the hdf5 test that checks the contents of this variable. Sorry for the delay!

pgrete: Great, thanks!

pgrete (approved):

pgrete:
Formatter does not work on forks. I'm force merging.

pgrete:
I don't see the force merge button any more? Am I missing sth?

Yurlungur:
> I don't see the force merge button any more? Am I missing sth?

I already clicked auto merge. It seems like not all the "long" tests triggered. I will try to disable/re-enable.

pgrete:
> > I don't see the force merge button any more? Am I missing sth?
> 
> I already clicked auto merge. It seems like not all the "long" tests triggered. I will try to disable/re-enable.

Yes, that's what I did earlier today and all tests passed (except for the formatting one that, as we know, does not report results for PRs from forks).
Previously, there was a force merge button but I don't see it any more (maybe because of the changed branched protection rules?)

Yurlungur:
> > > I don't see the force merge button any more? Am I missing sth?
> > 
> > 
> > I already clicked auto merge. It seems like not all the "long" tests triggered. I will try to disable/re-enable.
> 
> Yes, that's what I did earlier today and all tests passed (except for the formatting one that, as we know, does not report results for PRs from forks). Previously, there was a force merge button but I don't see it any more (maybe because of the changed branched protection rules?)

If you want to force merge, disable auto-merge. Then there should be an option.

pgrete:
> > > > I don't see the force merge button any more? Am I missing sth?
> > > 
> > > 
> > > I already clicked auto merge. It seems like not all the "long" tests triggered. I will try to disable/re-enable.
> > 
> > 
> > Yes, that's what I did earlier today and all tests passed (except for the formatting one that, as we know, does not report results for PRs from forks). Previously, there was a force merge button but I don't see it any more (maybe because of the changed branched protection rules?)
> 
> If you want to force merge, disable auto-merge. Then there should be an option.

That is basically my question. I don't see the button any more.
So if you still see it, feel free to go ahead and force merge (as the tests including extended ones passed earlier today).

Yurlungur:
> > > > > I don't see the force merge button any more? Am I missing sth?
> > > > 
> > > > 
> > > > I already clicked auto merge. It seems like not all the "long" tests triggered. I will try to disable/re-enable.
> > > 
> > > 
> > > Yes, that's what I did earlier today and all tests passed (except for the formatting one that, as we know, does not report results for PRs from forks). Previously, there was a force merge button but I don't see it any more (maybe because of the changed branched protection rules?)
> > 
> > 
> > If you want to force merge, disable auto-merge. Then there should be an option.
> 
> That is basically my question. I don't see the button any more. So if you still see it, feel free to go ahead and force merge (as the tests including extended ones passed earlier today).

Shoot you're right. I don't see it either. Maybe we submit a new MR on a branch and see if that goes through? Should figure out how to make fork merges work though...

pgrete:
Alright, so it wasn't just me and it was related to the branch protection rules.
I recently enabled
image
because of my accidental push to `develop`.
So I temp disabled it, now force merged, and then re-enabled it as a quick workaround for this PR.

In the long run it'd definitely be nice to have the formatter also trigger for forks.
I looked into this in the past and was not able to figure out a solution within a reasonable time (and thus tabled it).
So from my point of view, I'm happy to continue working with a manual work around (unless more and more PRs are coming from forks) and spent the time debugging CI on other (CI) pieces ;)

Yurlungur:
Aah I see. Ok sounds good.

PR #1199, opened 2024-10-25
Consolidate buffer packing functions with less atomics

alexrlongne:
## PR Summary

* Profiling with NSIGHT systems revealed that the `LoadBuffer_` function in `swarm_comms.cpp` took 20% of the GPU time in the particles example that used much more particles (1e6 per block). NSIGHT compute showed that the expensive kernel in `LoadBuffer_` had many stalls, we speculated this was due to waiting on the `atomic_fetch_add` function in `LoadBuffer_`. This PR reworks `LoadBuffer_` and `CountParticlesToSend_` to remove atomics and instead sort the particles by buffer ID and use their sorted order to determine their size and index into that buffer.

* Removing atomics reduces the runtime by about 20% in the million particle example and more in a 10 million particle version. A case where the particles are photons may involves even more cell block crossings as c*delta_t could be large compared to a block size (in the particles example the particle's path length is equal to a block size).  

* The reason for the atomic in both cases was that the loops were over particles where the particle buffer was determined and that was then used to atomically update a buffer specific field.

* As a rider, this PR also updates some sampling functions in the particle example to use less transcendentals, which has shown improvement in a Parthenon downstream code. I can separate into another PR if desired.

## Changes
+ Fold in buffer sizing to the load buffer function
+ Use a sort and a discontinuity check instead of an atomic_fetch_add inside of LoadBuffer_ to get particle index into buffer
+ Reduce transcendental functions call in particle sourcing in particle example

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build

### DIFF
--- a/example/particles/particles.cpp
+++ b/example/particles/particles.cpp
@@ -288,17 +288,18 @@ TaskStatus CreateSomeParticles(MeshBlock *pmb, const double t0) {
           } while (r > 0.5);
 
           // Randomly sample direction perpendicular to origin
-          Real theta = acos(2. * rng_gen.drand() - 1.);
-          Real phi = 2. * M_PI * rng_gen.drand();
-          v(0, n) = sin(theta) * cos(phi);
-          v(1, n) = sin(theta) * sin(phi);
-          v(2, n) = cos(theta);
+          const Real mu = 2.0 * rng_gen.drand() - 1.0;
+          const Real phi = 2. * M_PI * rng_gen.drand();
+          const Real stheta = std::sqrt(1.0 - mu * mu);
+          v(0, n) = stheta * cos(phi);
+          v(1, n) = stheta * sin(phi);
+          v(2, n) = mu;
           // Project v onto plane normal to sphere
           Real vdN = v(0, n) * x(n) + v(1, n) * y(n) + v(2, n) * z(n);
-          Real NdN = r * r;
-          v(0, n) = v(0, n) - vdN / NdN * x(n);
-          v(1, n) = v(1, n) - vdN / NdN * y(n);
-          v(2, n) = v(2, n) - vdN / NdN * z(n);
+          Real inverse_NdN = r * r;
+          v(0, n) = v(0, n) - vdN * inverse_NdN * x(n);
+          v(1, n) = v(1, n) - vdN * inverse_NdN * y(n);
+          v(2, n) = v(2, n) - vdN * inverse_NdN * z(n);
 
           // Normalize
           Real v_tmp = sqrt(v(0, n) * v(0, n) + v(1, n) * v(1, n) + v(2, n) * v(2, n));
@@ -335,11 +336,12 @@ TaskStatus CreateSomeParticles(MeshBlock *pmb, const double t0) {
           z(n) = minx_k + nx_k * dx_k * rng_gen.drand();
 
           // Randomly sample direction on the unit sphere, fixing speed
-          Real theta = acos(2. * rng_gen.drand() - 1.);
-          Real phi = 2. * M_PI * rng_gen.drand();
-          v(0, n) = vel * sin(theta) * cos(phi);
-          v(1, n) = vel * sin(theta) * sin(phi);
-          v(2, n) = vel * cos(theta);
+          const Real mu = 2.0 * rng_gen.drand() - 1.0;
+          const Real phi = 2. * M_PI * rng_gen.drand();
+          const Real stheta = std::sqrt(1.0 - mu * mu);
+          v(0, n) = vel * stheta * cos(phi);
+          v(1, n) = vel * stheta * sin(phi);
+          v(2, n) = vel * mu;
 
           // Create particles at the beginning of the timestep
           t(n) = t0;
--- a/example/particles/particles.cpp
+++ b/example/particles/particles.cpp
@@ -296,7 +296,7 @@ TaskStatus CreateSomeParticles(MeshBlock *pmb, const double t0) {
           v(2, n) = mu;
           // Project v onto plane normal to sphere
           Real vdN = v(0, n) * x(n) + v(1, n) * y(n) + v(2, n) * z(n);
-          Real inverse_NdN = r * r;
+          Real inverse_NdN = 1. / (r * r);
           v(0, n) = v(0, n) - vdN * inverse_NdN * x(n);
           v(1, n) = v(1, n) - vdN * inverse_NdN * y(n);
           v(2, n) = v(2, n) - vdN * inverse_NdN * z(n);
--- a/src/interface/swarm_comms.cpp
+++ b/src/interface/swarm_comms.cpp
@@ -221,7 +221,11 @@ void Swarm::LoadBuffers_() {
   auto swarm_d = GetDeviceContext();
   auto pmb = GetBlockPointer();
   const int particle_size = GetParticleDataSize();
+  vbswarm->particle_size = particle_size;
   const int nneighbor = pmb->neighbors.size();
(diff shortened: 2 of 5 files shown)


src/interface/swarm_comms.cpp
```diff
@@ -280,7 +344,7 @@ void Swarm::Send(BoundaryCommSubset phase) {
   auto swarm_d = GetDeviceContext();
 
   // Query particles for those to be sent
-  CountParticlesToSend_();
+  //CountParticlesToSend_();
```
alexrlongne: I'll remove this commented out function call and remove the old function.

src/interface/swarm_comms.cpp
```diff
@@ -236,42 +240,102 @@ void Swarm::LoadBuffers_() {
...
   auto &z = Get<Real>(swarm_position::z::name()).Get();
 
-  // Zero buffer index counters
-  auto &buffer_counters = buffer_counters_;
-  pmb->par_for(
-      PARTHENON_AUTO_LABEL, 0, NMAX_NEIGHBORS - 1,
-      KOKKOS_LAMBDA(const int n) { buffer_counters[n] = 0; });
+  if(max_active_index_ >= 0) {
+    // Make an n particle sized array of index, buffer pairs (with SwarmKey struct)
+    ParArray1D<SwarmKey> buffer_sorted("buffer_sorted", max_active_index_+1);
```
alexrlongne: This PR reuses the `SwarmKey` struct that used to sort and count particles in a cell. I can make a new version of that struct with more appropriate member names if you'd like.

brryan: What about renaming `cell_idx_1d_` in `SwarmKey` to something generic like `sort_idx_`? Then that one struct could be sensibly shared with the cell sorting and this buffer sorting.

I also think we want to get this `ParArray1D` declaration out of the `LoadBuffers_` call, because that will be a dynamic memory allocation every cycle which can be pretty slow. I would mimic what we do with the persistent memory for cell sorts, by adding `ParArray1D<SwarmKey> buffer_sorted_` right below `ParArray1D<SwarmKey> cell_sorted_` in `swarm.hpp`, and then also duplicating the work of initially allocating and resizing `buffer_sorted_` like we do `cell_sorted_`.

src/interface/swarm_comms.cpp
```diff
@@ -236,42 +240,102 @@ void Swarm::LoadBuffers_() {
...
+    pmb->par_for(
+        PARTHENON_AUTO_LABEL, 0, max_active_index_, KOKKOS_LAMBDA(const int n) {
+          if(swarm_d.IsActive(n)) {
+            bool on_current_mesh_block = true;
+            const int m =
+            swarm_d.GetNeighborBlockIndex(n, x(n), y(n), z(n), on_current_mesh_block);
+            buffer_sorted(n) = SwarmKey(m, n);
+          }
+          else {
+            buffer_sorted(n) = SwarmKey(-1, n);
```
brryan:
```suggestion
            buffer_sorted(n) = SwarmKey(this_block_, n);
```
This is a `Swarm` member we use to store the `-1` magic number as much as possible

src/interface/swarm_comms.cpp
```diff
@@ -236,42 +240,102 @@ void Swarm::LoadBuffers_() {
...
+          if (m >= 0 && n ==0 ) {
+            buffer_start(m) = 0;
+          }
+          else if (m >= 0 && m != buffer_sorted(n-1).cell_idx_1d_) {
+            buffer_start(m) = n;
+          }
+
+          // end checks (used to to size particle buffers)
+          if (m >= 0 && n == max_active_index ) {
+            num_particles_to_send(m) = n +1;
```
brryan: @alexrlongne in parthenon we use `clang-format-12` for formatting (I just have this hooked up to vscode so I always format on save) but just mentioning because the CI will fail otherwise.

src/interface/swarm_comms.cpp
brryan: Ditto for the `buffer_start` just below this comment (although it won't need to be resized as the swarm pool increases)

src/interface/swarm_comms.cpp
```diff
@@ -236,42 +240,102 @@ void Swarm::LoadBuffers_() {
...
+          }
+          else if (m >= 0 && m != buffer_sorted(n+1).cell_idx_1d_ ) {
+            num_particles_to_send(m) = n +1;
```
brryan: This is slick

brryan (commented):
@alexrlongne This looks great, I think with just formatting, cleanup of the `SwarmKey` `cell_idx_1d` name, and removing the dynamic memory allocs, this is ready to go.

src/interface/swarm_comms.cpp
alexrlongne: Thanks! Glad this worked out with minimal overhead.

src/interface/swarm_comms.cpp
alexrlongne: Thanks, the pipeline's passing for that now so I'll mark this as resolved.

src/interface/swarm_comms.cpp
alexrlongne: Fixed, thanks!

src/interface/swarm_comms.cpp
alexrlongne: I've added these to the class data and changed the `cell_idx_1d_` name to `sort_idx_`. Thanks!

src/interface/swarm_comms.cpp
alexrlongne: Done.

example/particles/particles.cpp
```diff
@@ -288,17 +288,18 @@ TaskStatus CreateSomeParticles(MeshBlock *pmb, const double t0) {
...
+          v(0, n) = stheta * cos(phi);
+          v(1, n) = stheta * sin(phi);
+          v(2, n) = mu;
           // Project v onto plane normal to sphere
           Real vdN = v(0, n) * x(n) + v(1, n) * y(n) + v(2, n) * z(n);
-          Real NdN = r * r;
-          v(0, n) = v(0, n) - vdN / NdN * x(n);
-          v(1, n) = v(1, n) - vdN / NdN * y(n);
-          v(2, n) = v(2, n) - vdN / NdN * z(n);
+          Real inverse_NdN = r * r;
```
brryan:
```suggestion
          Real inverse_NdN = 1. / (r * r);
```

src/interface/swarm_comms.cpp
```diff
@@ -156,72 +156,15 @@ void Swarm::SetupPersistentMPI() {
...
-}
-
 void Swarm::LoadBuffers_() {
   auto swarm_d = GetDeviceContext();
   auto pmb = GetBlockPointer();
   const int particle_size = GetParticleDataSize();
+  vbswarm->particle_size = particle_size;
   const int nneighbor = pmb->neighbors.size();
+  // Fence to make sure particles aren't currently being transported locally
+  // TODO(BRR) do this operation on device.
```
brryan:
```suggestion
```
Pretty sure that's a leftover comment I should have removed in an earlier cleanup.

brryan (approved):
Looks great! Just one more query about r^2 vs r^-2

example/particles/particles.cpp
alexrlongne: Whoops sorry! I guess the orbiting particles option is not in the tests right now. Applying...

pdmullen (approved):

PR #1209, opened 2024-11-13
Ordered history output

brryan:
## PR Summary

When populating history outputs, we index through a `Dictionary` of packages, which is an unordered map, and so the order in which we retrieve the packages isn't guaranteed. We've seen this cause inconsistencies in history outputs between platforms, and a more significant concern is that the order of the numbers in a history output could change between simulation runs. 

This is a simple fix that just always creates history outputs in alphabetical order of the name of the packages. 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - [[PR 1161]](https://github.com/parthenon-hpc-lab/parthenon/pull/1161) Make flux field Metadata accessible, add Metadata::CellMemAligned flag, small perfomance upgrades
 
 ### Changed (changing behavior/API/variables/...)
+- [[PR 1209]](https://github.com/parthenon-hpc-lab/parthenon/pull/1209) Ordered history output
 - [[PR 1206]](https://github.com/parthenon-hpc-lab/parthenon/pull/1206) Leapfrog fix
 - [[PR1203]](https://github.com/parthenon-hpc-lab/parthenon/pull/1203) Pin Ubuntu CI image
 - [[PR1177]](https://github.com/parthenon-hpc-lab/parthenon/pull/1177) Make mesh-level boundary conditions usable without the "user" flag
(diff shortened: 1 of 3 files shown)


pdmullen (approved):

Yurlungur (approved):
Nice fix. Alternative approach might be to change the map to an ordered one? I think lexicographic key ordering would then just happen for free?

brryan:
> Nice fix. Alternative approach might be to change the map to an ordered one? I think lexicographic key ordering would then just happen for free?

I think that's right, and we could do that too, except we would want to ensure that map ordering of `Dictionary` doesn't change in the future for the sake of this history output. This issue of history output columns switching midway through a run is potentially disastrous.

In that sense since this PR is probably free from a performance perspective, I'm slightly inclined to just do the sorting here since it's critical that we're sorted at this stage. And even if we have an ordered map later on, this sorting will still just be an extra check.

Yurlungur:
> > Nice fix. Alternative approach might be to change the map to an ordered one? I think lexicographic key ordering would then just happen for free?
> 
> I think that's right, and we could do that too, except we would want to ensure that map ordering of `Dictionary` doesn't change in the future for the sake of this history output. This issue of history output columns switching midway through a run is potentially disastrous.
> 
> In that sense since this PR is probably free from a performance perspective, I'm slightly inclined to just do the sorting here since it's critical that we're sorted at this stage. And even if we have an ordered map later on, this sorting will still just be an extra check.

Yeah no objections to doing it this way. I think in the end the difference will be in the weeds.

adamdempsey90:
I think it might be useful to document somewhere that the packages are not iterated through in the order that they were added, which one might naively expect.

adamdempsey90 (approved):

Yurlungur:
> I think it might be useful to document somewhere that the packages are not iterated through in the order that they were added, which one might naively expect.

+1

brryan:
> I think it might be useful to document somewhere that the packages are not iterated through in the order that they were added, which one might naively expect.

Good idea, done

PR #1192, opened 2024-10-17
Coalesced Buffer Communication

lroberts36:
## PR Summary

Coalesced buffer communication, see the included docs for a description. This came from the combined buffer communication work we did for the TACC Hackathon.

## PR Checklist

- [x] Code passes cpplint
- [x] New features are documented.
- [ ] Adds a test for any bugs fixed. Adds tests for new features.
- [x] Code is formatted
- [x] Changes are summarized in CHANGELOG.md
- [ ] Change is breaking (API, behavior, ...)
  - [ ] Change is *additionally* added to CHANGELOG.md in the breaking section
  - [ ] PR is marked as breaking
  - [ ] Short summary API changes at the top of the PR (plus optionally with an automated update/fix script)
- [x] CI has been triggered on [Darwin](https://re-git.lanl.gov/eap-oss/parthenon/-/pipelines) for performance regression tests.
- [ ] Docs build
- [ ] (@lanl.gov employees) Update copyright on changed files

PR #1214, opened 2024-11-25
Fix swarm memory management unit test failure

brryan:
## PR Summary

Right now multiple branches are failing in the CI with 
```
38/51 Test #38: Swarm memory management ....................................................***Failed    0.01 sec
Filters: Swarm memory management
*** The MPI_Comm_dup() function was called before MPI_INIT was invoked.
*** This is disallowed by the MPI standard.
*** Your MPI job will now abort.
[9b3082cbec2a:02162] Local abort before MPI_INIT completed completed successfully, but am not able to aggregate error messages, and not able to guarantee that all other processes were killed!
```

This should be fixed.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [ ] Docs build


pgrete:
~~The CI machine is currently heavily loaded (I already contacted the user in question), so I'm wondering if this is an actual bug or sth weird popping up with a machine under heavy load.~~

Nevermind. This happened also for the MacOS runner, which is not shared...

BenWibking:
So, err... does adding `printf` cause the error to go away? I tried re-triggering the CI actions, and I still don't see a problem...

Update: I triggered the macOS CI again...let's see what happens.

brryan:
> So, err... does adding `printf` cause the error to go away? I tried re-triggering the CI actions, and I still don't see a problem...
> 
> Update: I triggered the macOS CI again...let's see what happens.

So it just passed the macOS CI... so maybe this is an intermittent failure? Because it happens to this test on multiple CI workflows, even if only occasionally, I'll assume this is *not* fixed and I'll keep looking at this. Maybe it's some issue with how we're initializing the hierarchy of objects needed to create swarms...

pgrete:
@brryan is this still an issue or can we close PR/issue?

lroberts36:
This issue still exists in #1192, so I am not sure this should really be closed. The underlying issue is that `Swarm memory management` constructs a `Mesh` object within a catch2 test. If MPI is enabled, the mesh constructor makes a bunch of MPI communicators but MPI is never initialized. I am not sure why this issue isn't breaking the unit test all of the time. 

I have updated the PR with one possible solution that initializes MPI for all catch2 unit tests, but it does slow down the unit tests a little bit to have them each call MPI_init and MPI_finalize.

tst/CMakeLists.txt
```diff
@@ -1,6 +1,10 @@
 
 add_library(catch2_define catch2_define.cpp)
 target_link_libraries(catch2_define PUBLIC Catch2::Catch2 Kokkos::kokkos)
+if (ENABLE_MPI)
+  target_link_libraries(catch2_define PUBLIC MPI::MPI_CXX)
```
Yurlungur, line 5: do we want to rely on the C++ bindings for MPI? Aren't they completely un-maintained?

Yurlungur (approved):

tst/CMakeLists.txt
lroberts36: I just copied this from `src/CMakeLists.txt`:
```
if (ENABLE_MPI)
  target_link_libraries(parthenon PUBLIC MPI::MPI_CXX)
endif()
```

After looking around, I think this is not actually the C++ bindings 

> Added in version 3.9: Additionally, the following [IMPORTED](https://cmake.org/cmake/help/latest/prop_tgt/IMPORTED.html#prop_tgt:IMPORTED) targets are defined:

> `MPI::MPI_<lang>`

> Target for using MPI from `<lang>`.

pdmullen (approved):

Yurlungur:
I will rewrite this to make a separate executable for MPI unit tests.

lroberts36:
> I will rewrite this to make a separate executable for MPI unit tests.

I/Claude figured out how to only initialize MPI if we are running catch tests with an `[MPI]` tag. The unit tests now appear to run in about 0.01 seconds on a skylake node (aside from the swarm test, which takes half a second). I think this is cleaner than creating a second catch2 executable (which I tried initially).

Yurlungur (approved):
Nice LGTM

PR #1231, opened 2025-03-08
Add support for uint8_t data types in hdf5

adamdempsey90:
## PR Summary

In artemis, we store the contents of a binary file in params and read/write it in restart files. Previously, we've read that into a `std::vector<char>` variable, but we recently ran into a portability error when moving a restart file between two machines that (I think) define `char` differently. A solution to that problem is to use a more portable type like `uint8_t` instead of `char`. This adds support for that type to the hdf5 machinery. 

I've tested this in artemis and it works on read/write of the restart file.  

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build

### DIFF
--- a/src/outputs/parthenon_hdf5_types.hpp
+++ b/src/outputs/parthenon_hdf5_types.hpp
@@ -115,6 +115,7 @@ static hid_t getHDF5Type(const uint64_t *) { return H5T_NATIVE_UINT64; }
 static hid_t getHDF5Type(const float *) { return H5T_NATIVE_FLOAT; }
 static hid_t getHDF5Type(const double *) { return H5T_NATIVE_DOUBLE; }
 static hid_t getHDF5Type(const char *) { return H5T_NATIVE_CHAR; }
+static hid_t getHDF5Type(const uint8_t *) { return H5T_NATIVE_B8; }
 
 // On MacOS size_t is "unsigned long" and uint64_t is != "unsigned long".
 // Thus, size_t is not captured by the overload above and needs to selectively enabled.
--- a/src/outputs/parthenon_hdf5_types.hpp
+++ b/src/outputs/parthenon_hdf5_types.hpp
@@ -1,6 +1,6 @@
 //========================================================================================
 // Parthenon performance portable AMR framework
-// Copyright(C) 2020-2024 The Parthenon collaboration
+// Copyright(C) 2020-2025 The Parthenon collaboration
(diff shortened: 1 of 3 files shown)


src/outputs/parthenon_hdf5_types.hpp
```diff
@@ -115,6 +115,7 @@ static hid_t getHDF5Type(const uint64_t *) { return H5T_NATIVE_UINT64; }
 static hid_t getHDF5Type(const float *) { return H5T_NATIVE_FLOAT; }
 static hid_t getHDF5Type(const double *) { return H5T_NATIVE_DOUBLE; }
 static hid_t getHDF5Type(const char *) { return H5T_NATIVE_CHAR; }
+static hid_t getHDF5Type(const uint8_t *) { return H5T_NATIVE_B8; }
```
pgrete, line 118: Out of curiosity, why not using [H5T_NATIVE_UINT8](https://support.hdfgroup.org/documentation/hdf5/latest/group___p_d_t_c9x.html#ga1f71fcd3ede098948127632369435561)?

pgrete (approved):
LGTM

Out of curiosity, are you at liberty to say which machines those were?
I'm not sure I came across a machine  so far where the standard char def differed.
I'm asking from the point of view to make sure that the fix actually fixes the issue (and doesn't mask something else)

adamdempsey90:
> LGTM
> 
> Out of curiosity, are you at liberty to say which machines those were? I'm not sure I came across a machine so far where the standard char def differed. I'm asking from the point of view to make sure that the fix actually fixes the issue (and doesn't mask something else)

This was moving from our Venado machine with Grace-Hopper nodes to our Chicoma machine with AMD Rome - V100 nodes. Reading the Venado generated restart on Chicoma triggered the "types are not the same" error for this `std::vector<char>` param. 

I can't regenerate new Venado restarts at the moment (machine is unavailable for now) to test if this actually fixes the problem, but nevertheless this seems like a good idea.

src/outputs/parthenon_hdf5_types.hpp
adamdempsey90: I think the HDF5 docs I was reading pointed to the `B8` version instead of `UINT8`.

lroberts36 (approved):

Yurlungur (approved):

PR #1234, opened 2025-03-24
Tridiagonal Solver

lroberts36:
## PR Summary

This PR implements a tridiagonal solver within the Parthenon solver framework. *This is meant for testing only* and will only work for one-dimensional, single block runs. Additionally, it will only work for matrices that extend only one zone away in each direction. The advantage of this solver is that it is direct, so it sometimes allows for disentangling issues related to iterative solver convergence from downstream code implementation issues.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [ ] Docs build


src/bvals/comms/boundary_communication.cpp
```diff
@@ -366,18 +366,18 @@ TaskStatus ProlongateBounds(std::shared_ptr<MeshData<Real>> &md) {
 
   auto [rebuild, nbound] = CheckReceiveBufferCacheForRebuild<bound_type, false>(md);
 
-  if (rebuild) {
```
Yurlungur: Is this stuff no longer necessary?

Yurlungur (commented):
Looks reasonable. Also looks like it might be incomplete? Is everything in here you want? And should it be tested in the CI?

src/bvals/comms/boundary_communication.cpp
lroberts36: That should be removed from this PR since it isn't relevant to the solver (it got included in the PR by mistake), but I *think* the change may actually be necessary. There was a bug I ran into a few weeks ago that was fixed by this change, but I never was able to fully wrap my head around what was going on.

lroberts36:
@Yurlungur:
> Looks reasonable. Also looks like it might be incomplete? Is everything in here you want? And should it be tested in the CI?

The solver as written is complete (it is very simple), but the changes to `bvals` stuff should not be included here. I have put the PR in WIP status until I remove that stuff. I guess it would be trivial to add a test to the solver CI as well.

src/solvers/tridiag_solver.hpp
```diff
@@ -0,0 +1,316 @@
...
+class TridiagSolver : public SolverBase {
+  using FieldTL = typename equations::IndependentVars;
+
+  std::vector<std::string> sol_fields;
+  // Name of user defined container that should contain information required to
+  // calculate the matrix part of the matrix vector product
+  std::string container_base;
+  // User defined container in which the solution will reside, only needs to contain
+  // sol_fields
+  // TODO(LFR): Also allow for an initial guess to come in here
```
pgrete, line 54: Future todo or this PR?

src/solvers/tridiag_solver.hpp
```diff
@@ -0,0 +1,316 @@
...
+
+    const int b = 0;
+    const int k = kb.s;
+    const int j = jb.s;
+
+    // Since this needs to be sequential, we launch an outer loop of size one. Obviously
+    // this would be really inefficient on device
+    parthenon::par_for(
+        DEFAULT_LOOP_PATTERN, "DotProduct", DevExecSpace(), 0, 0,
+        KOKKOS_LAMBDA(const int) {
```
pgrete, line 200: Could this be replaced by a `parallel_scan` (in principle)?

src/solvers/tridiag_solver.hpp
```diff
@@ -0,0 +1,316 @@
...
+    auto pack_u = desc.GetPack(md_u.get());
+    auto pack_base = desc.GetPack(md_base.get());
+    auto pack_r = desc.GetPack(md_r.get());
+    auto pack_rhs = desc.GetPack(md_rhs.get());
+
+    parthenon::par_for(
+        DEFAULT_LOOP_PATTERN, "PrintSolution", DevExecSpace(), 0, 0, kb.s, kb.e, jb.s,
+        jb.e, ib.s - 1, ib.e + 1,
+        KOKKOS_LAMBDA(const int b, const int k, const int j, const int i) {
+          printf("row %i: %e %e %e  %e \n", i, pack_u(b, 0, k, j, i),
```
pgrete, line 246: Maybe `Kokkos::printf`?

src/solvers/tridiag_solver.hpp
```diff
@@ -0,0 +1,316 @@
...
+    sol = tl.AddTask(sol, utils::SetToZero<FieldTL>, md_r);
+    auto Ax_check = eqs_.Ax(tl, sol, md_base, md_u, md_r);
+    if (print_solution_)
+      Ax_check = tl.AddTask(Ax_check, PrintSolution, md_base, md_u, md_r, md_rhs);
+
+    return Ax_check;
+  }
+
+  Real GetSquaredResidualSum() const { return 0.0; }
+  int GetCurrentIterations() const { return 1; }
```
pgrete, line 304: This is hardcoded because it's not needed for the direct, single block solve, isn't it?

pgrete (approved):
LGTM

other minor comments:
- how about adding a comment somewhere at the top of the file that this is just for testing purpose (and already mention the restrictions in addition to the already implemented hard checks)?
- how hard/useful would it be to add some kind of (simple) test that makes use of the solver (motivated by the comment in the description that this was useful to test some part of the infrastructure downstream)?

Yurlungur (approved):

PR #1237, opened 2025-04-03
[Breaking] Sparse Pack Updates

lroberts36:
## PR Summary
This PR makes some changes to sparse packs and moves sparse and swarm pack related code to a new directory (since `interface/` has a *lot* of stuff in it). The two actual changes to packs are:
- Allow for passing a block selection functor to `SparsePack::Descriptor::GetPack` that determines which blocks in a `MeshData` object to include in a given pack. This makes creating a pack which contains only blocks that are on a physical boundary straightforward, for instance. A similar capability is also provided by the previously existing vector `include_block`, but I think this functor approach is a bit cleaner. 
- Update `SparsePackCache` to cache multiple packs created by the same pack descriptor but made from different block lists. 

**BREAKING:** This is a possibly breaking change for some downstream codes if they use `#include 'interface/sparse_pack.hpp'` (or explicitly include other headers that got moved). 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [ ] Docs build


src/pack/block_selector.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+namespace parthenon {
+
+// A block selector function can be passed SparsePack::Descriptor::GetPack to
+// select a subset of the blocks contained in a MeshData object based on a
+// chosen criteria
+using block_selector_func_t = std::function<bool(MeshBlockData<Real> *)>;
+
+// Predefined block selector functions
+namespace GetBlockSelector {
```
Yurlungur, line 47: It's a bit weird to have a namespace have a `GetPrefix` but I think I see the logic here and it works.

src/pack/sparse_pack_base.cpp
```diff
@@ -355,25 +355,21 @@ template <class T>
...
-      if (alloc_status[i] != alloc_status_in[i])
-        return BuildAndAdd(pmd, desc, include_block);
-    }
-    auto &include_status = std::get<2>(cache_tuple);
-    if (include_status.size() != include_block.size())
-      return BuildAndAdd(pmd, desc, include_block);
-    for (int i = 0; i < include_block.size(); ++i) {
-      if (include_status[i] != include_block[i])
+    auto &desc_pack_map = pack_map[desc.identifier];
+    if (desc_pack_map.count(include_block)) {
```
Yurlungur, line 359: Don't maps now have a `contains`?

Yurlungur (approved):
This is a nice cleanup, along with the new feature. :+1:

src/pack/block_selector.hpp
lroberts36: Yeah, that also felt a bit weird to me, but I think it makes the eventual calls looks cleaner though.

Yurlungur: :+1: yeah I see the logic. No need to change it.

src/pack/sparse_pack_base.cpp
lroberts36: I think `contains` is only in C++20.

Yurlungur: Ah ok. Boo.

src/mesh/meshblock.hpp
```diff
@@ -187,9 +187,17 @@ class MeshBlock : public std::enable_shared_from_this<MeshBlock> {
...
+    // TODO(LFR): Should we only return true if this is set to user?
     return boundary_flag[bf] != BoundaryFlag::block;
   }
 
+  bool IsPhysicalBoundary() const {
+    bool is_bound;
+    for (int bf = 0; bf < 6; ++bf)
+      is_bound = is_bound || IsPhysicalBoundary(static_cast<BoundaryFace>(bf));
+    return is_bound;
+  }
```
pdmullen:
```suggestion
    bool is_bound = IsPhysicalBoundary(static_cast<BoundaryFace>(0));
    for (int bf = 1; bf < 6; ++bf)
      is_bound = is_bound || IsPhysicalBoundary(static_cast<BoundaryFace>(bf));
    return is_bound;
  }
```

src/pack/sparse_pack.hpp
```diff
@@ -427,6 +399,67 @@ inline std::ostream &operator<<(std::ostream &os, const SparsePack<Vars...> &sp)
...
+    T *pmd, const block_selector_func_t &block_selector, std::vector<bool> &include_block,
+    bool only_fine_two_level_composite_blocks) const {
+  PARTHENON_REQUIRE(include_block.size() == pmd->NumBlocks() ||
+                        (include_block.size() == 0),
+                    "Must specify inclusion status for all blocks.");
+
+  // LFR: For multi-grid, we want to select only fine blocks on two-level composite
+  // grids by default since only the fine grid cells are "active" (the coarse level
+  // blocks just provide necessary boundary information during comms)
+  block_selector_func_t fbc_selector{};
```
pdmullen, line 444: Will this ever have to be threaded through `SwarmPack`s too?

pdmullen (approved):
LGTM.  Tested against select problems in `artemis+jaybenne`.

src/mesh/meshblock.hpp
lroberts36: nice catch, thanks!

src/pack/sparse_pack.hpp
lroberts36: Probably, if we want to be able to select subsets of blocks.

lroberts36: Probably we will eventually want to allow for selecting subsets of blocks in `MeshData` in `SwarmPack`s, but that is probably for another time.

PR #1240, opened 2025-04-11
fix cell mem aligned io

Yurlungur:
## PR Summary

Currently the code segfaults if you request fluxes in an output file. This is because (e.g., cell variable) fluxes are face fields but, because they default to `Metadata::CellMemAligned`, they have a different number of ghost zones than expected. The I/O packing/unpacking logic miscounts, and invalid memory is accessed. This MR resolves this issue. 

Because I wanted it, I also snuck in a minor quality of life improvement to the `movie2d.py` script. Mainly, it now supports a colorbar and the ability to choose a colormap.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -76,9 +76,9 @@ include(cmake/Format.cmake)
 include(cmake/Lint.cmake)
 
 # regression test reference data
-set(REGRESSION_GOLD_STANDARD_VER 25 CACHE STRING "Version of gold standard to download and use")
+set(REGRESSION_GOLD_STANDARD_VER 26 CACHE STRING "Version of gold standard to download and use")
 set(REGRESSION_GOLD_STANDARD_HASH
-  "SHA512=314dc8312366d81ba33d1fde25812e9a7697b2f529de29e22662df0d458f1c4bc5b5bb4e649888170f66ffec0df1be20a9cf401944531c1c1ad835e26eaad28f"
+  "SHA512=c3f3a2da823a7e65a5800c902d26b3891f0a129812031ceae46e5c394f9c5bea36e26f645007422dca46dfdc7c446271c1ff5c0cec851bc40a186aa874beb643"
   CACHE STRING "Hash of default gold standard file to download")
 option(REGRESSION_GOLD_STANDARD_SYNC "Automatically sync gold standard files." ON)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -78,7 +78,7 @@ include(cmake/Lint.cmake)
 # regression test reference data
 set(REGRESSION_GOLD_STANDARD_VER 26 CACHE STRING "Version of gold standard to download and use")
 set(REGRESSION_GOLD_STANDARD_HASH
-  "SHA512=c3f3a2da823a7e65a5800c902d26b3891f0a129812031ceae46e5c394f9c5bea36e26f645007422dca46dfdc7c446271c1ff5c0cec851bc40a186aa874beb643"
+  "SHA512=4c61a0b78a0d68c3175e506d9fc5bc46ec6e91009413125c358801e230f4d882daf0c4cff6a3017f134d7289471a93f677956c2865dd3253264a15ae96b664b6"
   CACHE STRING "Hash of default gold standard file to download")
 option(REGRESSION_GOLD_STANDARD_SYNC "Automatically sync gold standard files." ON)
 
--- a/scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
+++ b/scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
@@ -1,5 +1,5 @@
 # =========================================================================================
-# (C) (or copyright) 2020-2023. Triad National Security, LLC. All rights reserved.
+# (C) (or copyright) 2020-2025. Triad National Security, LLC. All rights reserved.
 #
 # This program was produced under U.S. Government contract 89233218CNA000001 for Los
 # Alamos National Laboratory (LANL), which is operated by Triad National Security, LLC
--- a/scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
+++ b/scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
@@ -66,6 +66,10 @@ parser.add_argument(
     help="Tensor components of field to plot. Mutally exclusive with --vector-component.",
 )
 parser.add_argument(
+    "--colorbar", type=str, default=None, help="Add a colorbar with the specified label"
+)
+parser.add_argument("--colormap", type=str, default="plasma", help="Colormap to use")
+parser.add_argument(
     "--swarm",
     type=str,
     default=None,
@@ -195,6 +199,8 @@ def plot_dump(
     q,
     time_title,
     output_file: Path,
+    colormap="viridis",
+    colorbar=None,
     with_mesh=False,
     block_ids=[],
     xi=None,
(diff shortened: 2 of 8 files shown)


scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
```diff
@@ -65,6 +65,10 @@ def maybe_float(string):
     default=None,
     help="Tensor components of field to plot. Mutally exclusive with --vector-component.",
 )
+parser.add_argument(
+    "--colorbar", type=str, default=None, help="Add a colorbar with the specified label"
+)
```
lroberts36, line 70: Nice!

src/outputs/output_utils.hpp
```diff
@@ -135,6 +136,8 @@ struct VarInfo {
           bool is_vector, const IndexShape &cellbounds)
       : label(label), num_components(num_components), nx_(nx),
         tensor_rank(metadata.Shape().size()), where(metadata.Where()),
+        is_mem_aligned(metadata.IsSet(Metadata::CellMemAligned) &&
+                       metadata.IsSet(Metadata::Flux)),
```
lroberts36: Should this be 
```c++
is_mem_aligned(metadata.IsSet(Metadata::CellMemAligned) &&
                       !metadata.IsSet(Metadata::Cell))
```
?

src/outputs/output_utils.cpp
```diff
@@ -47,6 +47,11 @@ Triple_t<int> VarInfo::GetNumKJI(const IndexDomain domain) const {
     nx2 = std::max(nx2, cellbounds.ncellsj(domain, el));
     nx1 = std::max(nx1, cellbounds.ncellsi(domain, el));
   }
+  if (is_mem_aligned && DomainTouchesOuterGhosts(domain)) {
+    nx1 -= 1;
+    if (nx2 > 1) nx2 -= 1;
+    if (nx3 > 1) nx3 -= 1;
+  }
```
lroberts36, line 54: I think there should be `TopologicalOffsetI(te)` calls here replacing the ones.

CMakeLists.txt
```diff
@@ -76,9 +76,9 @@ include(cmake/Format.cmake)
 include(cmake/Lint.cmake)
 
 # regression test reference data
-set(REGRESSION_GOLD_STANDARD_VER 25 CACHE STRING "Version of gold standard to download and use")
+set(REGRESSION_GOLD_STANDARD_VER 26 CACHE STRING "Version of gold standard to download and use")
 set(REGRESSION_GOLD_STANDARD_HASH
-  "SHA512=314dc8312366d81ba33d1fde25812e9a7697b2f529de29e22662df0d458f1c4bc5b5bb4e649888170f66ffec0df1be20a9cf401944531c1c1ad835e26eaad28f"
+  "SHA512=c3f3a2da823a7e65a5800c902d26b3891f0a129812031ceae46e5c394f9c5bea36e26f645007422dca46dfdc7c446271c1ff5c0cec851bc40a186aa874beb643"
```
pgrete: I see the gold files have been updated, but I'm not sure I see the code that now requires the new files to pass.

src/outputs/output_utils.hpp
pgrete: Is this only tied to flux fields (based on this code) or more general for `CellMemAligned` (based on the comment in line 60 above)?

I tried to make a connection to the metadata doc (`By default, a flux field for a cell-centered field is built with Metadata::CellMemAligned flag set for backwards compatability.`) but I guess I'm not entirely aware of what `CellMemAligned` actually means.

src/outputs/output_utils.cpp
pgrete, line 54: Potentially general question: What happens to the outermost layer of face centered fields, i.e., the ones marking the domain boundary?
This is probably also related to the question of what the difference is (both internally and for outputs) between true face centered fields and face centered fields serving as fluxes for cells.

pgrete (commented):
Thanks for fixing this. I have a couple of questions/comments but "blocking" is probably only the one regarding the integrated test.

In addition to the inline comments, I'm also wondering what this means for edges (fluxes of faces) and node fields.

CMakeLists.txt
Yurlungur: Oops you're right... A boundary flux should now be output in the hdf5 output test. I must have forgotten to commit that. Will fix it.

src/outputs/output_utils.hpp
Yurlungur: `CellMemAligned` is for face fields only, I think. It basically drops 1 ghost zone on the upper faces of the block so that the memory footprint for each face is the same as for cells. This is how fluxes worked in Athena++ before @lroberts36 turned fluxes into face-centered vars and unified that machinery.

I don't think you can add `CellMemAligned` to edge- or node-centered data.

You *can* make a non-flux face var `CellMemAligned` if you want to. And there might be reasons to do so.

src/outputs/output_utils.cpp
Yurlungur: Just to be clear, there are still outer ghosts, just one fewer. So if you had 2 you only have 1. That said I don't think the physical boundary code as ever been applied to `CellMemAligned` fields. That should probably be outside the purview of this MR.

Yurlungur:
Thanks for the review, @pgrete ! Will address your inline comments. Regarding what this means for edges and nodes... @lroberts36 should correct me if I'm wrong, but I think `CellMemAligned` is only valid for face-centered fields.

Edge and node centered data output should work.

src/outputs/output_utils.hpp
lroberts36: > I don't think you can add `CellMemAligned` to edge- or node-centered data.

> You _can_ make a non-flux face var `CellMemAligned` if you want to. And there might be reasons to do so.

It does work with face-, edge-, and node-centered data and doesn't require `Metadata::Flux`. But if you automatically make fluxes associated with a field, the associated flux field has `Metadata::CellMemAligned` and `Metadata::Flux` set automatically for backwards compatibility.

src/outputs/output_utils.cpp
lroberts36: `CellMemAligned` just means that the underlying array has the same shape as a cell-centered field `(nz + 2 * nghost, ny + 2 * nghost, nx + 2 * nghost)` instead of `(nz + 2 * nghost + 1, ny + 2 * nghost + 1, nx + 2 * nghost + 1)`. For instance, that means that for a face-field in the x-direction, there is no storage for the last ghost face. This means we cannot set `Metadata::FillGhost` for `Metadata::CellMemAligned` fields. This is enforced at https://github.com/parthenon-hpc-lab/parthenon/blob/a0fdccf7007c5b83dec85d0deec5252b6c15882b/src/interface/metadata.cpp#L294

CMakeLists.txt
Yurlungur: Should be resolved now.

src/outputs/output_utils.cpp
Yurlungur: I dug into this... These functions I modified always explicitly ask for the "padded" shape, which includes the empty pieces in the memory footprint. So I don't think subtracting `TopologicalOffset` is correct.

src/outputs/output_utils.hpp
Yurlungur: Fixed.

Yurlungur:
Thanks for the reviews @pgrete @lroberts36 I think I've addressed your comments.

lroberts36 (approved):

pgrete (approved):
Thanks for the replies. I'll probably include some of the info in the docs for future reference.

Looks good now!

PR #1243, opened 2025-04-16
Increase MPI timeout counter by 10x

pgrete:
## PR Summary

We recently got bit by this again on a buy CPU cluster.
IIRC other groups have also adapted this in the past so that I think increasing the default is a good idea (that also does no harm).
Any objection getting this in before the next release?

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 - [[PR 1161]](https://github.com/parthenon-hpc-lab/parthenon/pull/1161) Make flux field Metadata accessible, add Metadata::CellMemAligned flag, small perfomance upgrades
 
 ### Changed (changing behavior/API/variables/...)
+- [[PR 1243]](https://github.com/parthenon-hpc-lab/parthenon/pull/1243) Increase MPI timeout counter by 10x
 - [[PR 1223]](https://github.com/parthenon-hpc-lab/parthenon/pull/1223) Update to v4 artifacts
 - [[PR 1216]](https://github.com/parthenon-hpc-lab/parthenon/pull/1216) Move to Kokkos 4.5
 - [[PR 1191]](https://github.com/parthenon-hpc-lab/parthenon/pull/1191) Update Kokkos version to 4.4.1
--- a/src/utils/communication_buffer.hpp
+++ b/src/utils/communication_buffer.hpp
@@ -327,8 +327,8 @@ bool CommBuffer<T>::TryReceive() noexcept {
       *comm_type_ == BuffCommType::sparse_receiver) {
 #ifdef MPI_PARALLEL
     (*nrecv_tries_)++;
-    PARTHENON_REQUIRE(*nrecv_tries_ < 1e8,
-                      "MPI probably hanging after 1e8 receive tries.");
+    PARTHENON_REQUIRE(*nrecv_tries_ < 1e9,
+                      "MPI probably hanging after 1e9 receive tries.");
 
     TryStartReceive();
 


Yurlungur (approved):
No objections from me. Hopefully the unified comms machinery can help with this long term...

lroberts36 (approved):
LGTM. I do think we should move to a timer based solution though, which may be pretty easy to add to the tasking infrastructure though, rather than directly in the comms stuff (i.e. set a maximum amount of time a `TaskCollection` is allowed to run for).

c-prather:
I propose we just remove this counter, even without a replacement in place yet.  I have yet to hear a report of an actual MPI hang being prevented by this thing, and there was nothing magic to the number I bumped it to last time(s) we hit this.  The same number of iterations might be seconds on a fast CPU cluster, and hours on older GPU hardware, making this method functionally useless no matter how much we tweak it.  If we remove it, the worst case scenario is Parthenon wastes a few hours of allocation time hung, but smart job management and supervision really limit that impact.

I have incidentally had Parthenon hang and waste entire jobs, but always at initialization, filesystem access, that sort of thing, not an MPI hang caught by this code.  So, as @lroberts36 mentions, the real solution is to implement timer watchdogs for any task, and set default maybe ~hour time limits for a bunch of simple things.  If we don't care about thread safety, it's just a matter of kicking off a timer and implementing a callback, right?

pgrete:
I agree that the current solution is not ideal and that a timer based solution is very much preferable.
IIRC we even have one already in place for some other routine.

However, the current solution did help us in the past on Frontier (where even short jobs were 10k+ GPU hours).
Thus, I'm in favor of keeping this suboptimal solution for now (instead of removing it) until a clean one is in place.

lroberts36:
I just started on adding `TaskRegion` timeouts, https://github.com/parthenon-hpc-lab/parthenon/pull/1244, let me know if you think this sort of strategy will work for you.

pgrete:
I think that strategy overall looks great (as is much more powerful).

Given subtle (past) challenges with threading and the ThreadPool (isn't it disabled by default at the moment), I'd feel better if that kind of change only makes it develop after the next release (so that it can mature/be tested more broadly).

So what shall we do with the current PR? We can just keep the number updated downstream and not merge this (though from my point of view it wouldn't hurt).

lroberts36:
> I think that strategy overall looks great (as is much more powerful).
> 
> Given subtle (past) challenges with threading and the ThreadPool (isn't it disabled by default at the moment), I'd feel better if that kind of change only makes it develop after the next release (so that it can mature/be tested more broadly).

Yeah, it seemed like a simple enough change but the threading logic makes it a bit complicated to determine whether or not this has some unintended consequence. I am currently playing around with testing this. It would probably be good to have some unit tests for `ThreadPool`. 

> So what shall we do with the current PR? We can just keep the number updated downstream and not merge this (though from my point of view it wouldn't hurt).

I agree that we should just merge this for now.

lroberts36:
FWIW, it is trivial to add timing to `SerialPool` in `check_task_returns`. After looking at this a bit, I am not sure why `*::wait()` and `*::check_task_returns` are both needed though...

c-prather:
I'm happy merging this for the release (though maybe we should consider 100x, for the breathing room?). I was under the impression it hadn't provided much value to anyone, but if it's useful that's great.

Also agree with merging timers after release. I don't think there's any great risk of bugs with such simple implementations, but I do think setting any default limits is going to screw with downstreams (e.g., say that mostly you want `TaskRegions` to complete in <1 second, but your file output region takes >10 minutes, that sort of thing).  And I would suggest we set some default limits with good error messages, rather than rely on users to lay out all of their own timings.

pgrete:
Alright, thanks for the feedback. I'm going to pull the trigger (keeping it at the 10x for now as this has been sufficient on all cases we encountered in the past -- and it's also within standard int limits).

PR #1244, opened 2025-04-16
Add timeout capability to `TaskCollection`

lroberts36:
## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [ ] Docs build


Yurlungur:
@lroberts36 is this ready for review?

lroberts36:
> @lroberts36 is this ready for review?

Yes, it is ready for review.

src/mesh/mesh.cpp
```diff
@@ -80,6 +80,8 @@ Mesh::Mesh(ParameterInput *pin, ApplicationInput *app_in, Packages_t &packages,
       nbnew(), nbdel(), step_since_lb(), gflag(), packages(packages),
       resolved_packages(ResolvePackages(packages)),
       default_pack_size_(pin->GetOrAddInteger("parthenon/mesh", "pack_size", -1)),
+      task_collection_timeout_in_seconds(pin->GetOrAddInteger(
+          "parthenon/mesh", "task_collection_timeout_in_seconds", 60 * 60 * 24 * 365)),
```
Yurlungur: 1 year lol?

src/mesh/mesh.cpp
```diff
@@ -641,70 +643,17 @@ void Mesh::BuildTagMapAndBoundaryBuffers() {
...
-  for (auto &partition : partitions) {
-    auto &md = mesh_data.Add(md_name, partition, fields);
-    if (multilevel) {
-      ApplyBoundaryConditionsOnCoarseOrFineMD(md, true);
-      ProlongateBoundaries(md);
-    }
-    ApplyBoundaryConditionsOnCoarseOrFineMD(md, false);
+  for (int i = 0; i < num_partitions; i++) {
+    auto &md = mesh_data.Add(md_name, partitions[i], fields);
+    auto bound = AddBoundaryExchangeTasks(TaskID(0), region[i], md, multilevel);
```
Yurlungur, line 664: nice cleanup

src/driver/driver.hpp
```diff
@@ -134,15 +134,15 @@ TaskListStatus ConstructAndExecuteBlockTasks(T *driver, Args... args) {
   for (auto &pmb : driver->pmesh->block_list) {
     tr[i++] = driver->MakeTaskList(pmb.get(), std::forward<Args>(args)...);
   }
-  TaskListStatus status = tc.Execute();
+  TaskListStatus status = tc.Execute(driver->pmesh->task_collection_timeout_in_seconds);
```
Yurlungur, line 164: Are these functions the only place this needs to be expressed? It seems like it?

Yurlungur (approved):
Given that the default timeout is longer than any code will ever run, I think it's probably worth documenting this mechanism in the docs.

Also should the default timeout time be something shorter? Like a week or a day or something? Short enough that it might actually trigger, but long enough that nobody will hit it accidentally?

Approving now, though, as these are minor nitpicks

src/driver/driver.hpp
lroberts36: Yeah, I think so. I grepped the code for `Execute()` and IIRC this was the only place I found it (aside from where I added it during the internal Mesh boundary communication).

src/mesh/mesh.cpp
lroberts36: ha, yeah, I wanted a timeout that no one would hit and I thought a year was funny. 

We should probably have a discussion about what people think is a good default choice. I will make an issue.

Yurlungur:
We should just pick a timeout so this can get merged and we can change it later.

src/mesh/mesh.cpp
```diff
@@ -78,6 +78,8 @@ Mesh::Mesh(ParameterInput *pin, ApplicationInput *app_in, Packages_t &packages,
                                      "enable a multigrid mesh")),
       nbnew(), nbdel(), step_since_lb(), gflag(), packages(packages),
       resolved_packages(ResolvePackages(packages)),
+      task_collection_timeout_in_seconds(pin->GetOrAddInteger(
+          "parthenon/mesh", "task_collection_timeout_in_seconds", 60 * 60 * 24 * 365)),
```
lroberts36:
```suggestion
          "parthenon/mesh", "task_collection_timeout_in_seconds", 60 * 5)),
```

src/tasks/tasks.hpp
```diff
@@ -493,8 +493,8 @@ class TaskCollection {
     regions.emplace_back(num_lists);
     return regions.back();
   }
-  TaskListStatus Execute() {
-    static Pool_t pool(1);
+  TaskListStatus Execute(std::size_t timeout_in_seconds = 60 * 60 * 24 * 365) {
```
lroberts36:
```suggestion
  TaskListStatus Execute(std::size_t timeout_in_seconds = 60 * 5) {
```

src/tasks/thread_pool.hpp
```diff
@@ -66,17 +67,22 @@ class ThreadQueue {
     complete = true;
     cv.notify_all();
   }
-  void wait_for_complete() {
-    std::unique_lock<std::mutex> lock(mutex);
-    waiting = true;
-    if (queue.empty() && nwaiting == nworkers) {
+  bool wait_for_complete(std::chrono::seconds max_time) {
+    bool timeout{false};
```
pgrete, line 71: Is it worth to work with `std::cv_status` variables here?

src/tasks/thread_pool.hpp
```diff
@@ -66,17 +67,22 @@ class ThreadQueue {
...
+      }
+      timeout = !complete_cv.wait_for(lock, max_time, [this]() { return complete; });
       complete = false;
       waiting = false;
-      return;
     }
-    complete_cv.wait(lock, [this]() { return complete; });
-    complete = false;
-    waiting = false;
+    if (!timeout) signal_kill();
```
pgrete: Why do we sigkill if timeout is false?

pgrete (commented):
I just fixed the param doc and two typos.
After quick clarification of the question, I think we can merge (assuming the tests pass now)

src/tasks/thread_pool.hpp
lroberts36: I think this is a bug, it should be `if (timeout)`. It ends up not mattering currently, since we are limited to one thread right now.

src/tasks/thread_pool.hpp
lroberts36: Changed.

lroberts36: Actually, using `std::cv_status` requires a different call signature. I am going to leave this as is.

tst/unit/test_tasklist.cpp
```diff
@@ -48,3 +48,14 @@ TEST_CASE("Task Object Lifecycle", "[TaskList][AddTask]") {
...
 }
+
+TEST_CASE("TaskCollection timeout", "[TaskList][AddTask]") {
+  GIVEN("A TaskCollection") {
+    parthenon::TaskCollection tc;
+    parthenon::TaskRegion &region = tc.AddRegion(1);
+    region[0].AddTask(TaskID(0), []() { return TaskStatus::incomplete; });
+    const std::size_t timeout_in_seconds = 4;
+    tc.Execute(timeout_in_seconds);
+    REQUIRE(true);
```
pgrete: I missed this yesterday.
How does this unit test check the timeout capabilities?
I'd have imagined sth like a task waiting for more than the timeout and and then check for the timeout return.
Given subtle issues in the past we had with tasking/threading, I think that it's important to have a test for this new feature.

lroberts36: If the task list timeout didn't work, `tc.Execute()` would run forever since the task always says it is `incomplete` and the task would be repeatedly called (similar to what would happen for a `TryReceive` call that never gets data from a neighbor). I changed this to get the status of the `TaskCollecion` and make sure the task collection failed. 

Remember that when we compile with `SerialPool`, a task that runs for a long time without returning will *not* trigger the timeout. The only way to have that behavior I think is to spin off another thread (similar to what happens when we compile with `ThreadPool` or in the Athena++ code you linked to the other day).

pgrete: Ah, right, thanks for clarifying.

pgrete (approved):

PR #1247, opened 2025-04-23
cleanup output and add option of outputting a specific stage

Yurlungur:
## PR Summary

While working on 4th-order hydro with Sam, it seemed useful to be able to output specific stages other than the base stage into an HDF5 file. This MR adds the `MeshData` object you want to read from as an input as an optional argument to the input block. 

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build


src/parameter_input.cpp
```diff
@@ -848,87 +848,6 @@ std::string ParameterInput::SetString(const std::string &block, const std::strin
...
-            << "Parameter name 'dt' not found in block '" << pb->block_name << "'";
-        PARTHENON_FAIL(msg);
-      }
-      next_time -= static_cast<Real>(atof(pl->param_value.c_str()));
-      msg << next_time;
-      // AddParameter(pb, "next_time", msg.str().c_str(), "# Updated during run time");
-      SetReal(pb->block_name, "next_time", next_time);
-    }
-    pb = pb->pnext;
-  }
```
lroberts36, line 884: Why is this being removed in this PR?

lroberts36 (approved):
LGTM.

src/parameter_input.cpp
pgrete: IIRC we never called these functions and they're leftover from Athena++ (where at least one of them was also not used).
If I'm not mistaken, they were supposed to support logic to adjust the time when restarting from intermediate outputs, which I think we solved with some updated ordering of when we write restart outputs (always last so that new non-restart outputs have already been written and are not overwritten again when restarting).

pgrete (approved):
I think this makes sense and seems like a clean implementation.

src/parameter_input.cpp
Yurlungur: Yeah I think that's right. I deleted them because they're completely unused.

As an aside, something I want to add to another MR maybe is documentation for how to change output cadence upon restart from the command line. This is totally possible to do, but requires setting both `parthenon/output/dt` and `parthenon/output/next_time`.

PR #1248, opened 2025-04-24
Fix edge case where meshblocks can be improperly de-refined

Yurlungur:
## PR Summary

The last week or so, I've been chasing down an AMR bug in riot. With a Rayleigh-Taylor instability, I was seeing the solution become asymmetric. I chased down the problem to the mesh refining incorrectly. Here's a plot of asymmetry in the fluid solution, with the meshblocks labeled by gid right before the problem appears:
rt_amr_broken_before
On the next time step, a collection of blocks at the top derefines, while the symmetric block at the bottom does not. In particular, blocks 87, 88, 89, and 90 get coarsened, while blocks 2, 3, 4, and 5 do not.
rt_amr_broken_mid
One step later, the mesh re-refines:
rt_amr_broken_after
At this point, the top blocks have experienced restriction and prolongation, while the bottom blocks have not. This introduces asymmetric truncation error.

The problem turned out to be in `mesh-amr_loadbalance.cpp`. The code should only de-refine if all children of a parent block want to de-refine. However, the code that checks for this is quite brittle. It compares logical locations, rather than actual parent associativity. When we moved to a forest of octrees, rather than a single tree, we needed to add a comparison for tree as well as location within a tree, and that check was missing. 

In this MR, I move to a slightly less brittle approach suggested by @lroberts36 which checks for "same level neighbor" rather than computing by hand. That said, this code is *still* quite brittle and we may wish to move away from the current logic, which assumes morton ordering, to logic that just tracks which blocks want to de-refine via, e.g., an unordered set.

While I was at it, I also couldn't resist cleaning up the memory management, but replacing the `new` and `delete` calls with a `std::vector` and adding some explanatory comments.

I also added a little documentation and added the script I used to check for symmetry.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/doc/sphinx/src/outputs.rst
+++ b/doc/sphinx/src/outputs.rst
@@ -167,6 +167,17 @@ immediately prior to restart files being written with the optional
 callbacks (if provided) will be called in that order before restart files are
 written.
 
+Changing output cadence from command line when restarting from file
+--------------------------------------------------------------------
+
+When restarting from a restart file, you can change any paremeter
+input argument. To change the output cadence in this way, however,
+special care is required, as the code stores a ``next_time`` and a
+``next_n`` for the next time to output and next iteration to
+output. Fortunately, these can be overwritten on the command line in
+the usual way. To set the next time to output when restarting, run
+with ``parthenon/output*/next_time=some_number``.
+
 Postprocessing/native analysis
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
--- a/doc/sphinx/src/outputs.rst
+++ b/doc/sphinx/src/outputs.rst
@@ -177,6 +177,8 @@ special care is required, as the code stores a ``next_time`` and a
 output. Fortunately, these can be overwritten on the command line in
 the usual way. To set the next time to output when restarting, run
 with ``parthenon/output*/next_time=some_number``.
+Note that this is independent of updating ``parthenon/output*/dt=some_number``.
+In other words, to change the cadence and apply the change immediately both ``dt`` (or ``dn``) and ``next_time`` (or ``next_n`` need to be updated simultaneously.
 
 Postprocessing/native analysis
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/doc/sphinx/src/outputs.rst
+++ b/doc/sphinx/src/outputs.rst
@@ -170,7 +170,7 @@ written.
 Changing output cadence from command line when restarting from file
 --------------------------------------------------------------------
 
-When restarting from a restart file, you can change any paremeter
+When restarting from a restart file, you can change any parameter
 input argument. To change the output cadence in this way, however,
 special care is required, as the code stores a ``next_time`` and a
 ``next_n`` for the next time to output and next iteration to
--- a/scripts/python/packages/parthenon_tools/parthenon_tools/compute_asymmetry.py
+++ b/scripts/python/packages/parthenon_tools/parthenon_tools/compute_asymmetry.py
@@ -1,0 +1,65 @@
+#!/usr/bin/env python
+# ========================================================================================
+#  (C) (or copyright) 2025. Triad National Security, LLC. All rights reserved.
+#
+#  This program was produced under U.S. Government contract 89233218CNA000001 for Los
+#  Alamos National Laboratory (LANL), which is operated by Triad National Security, LLC
+#  for the U.S. Department of Energy/National Nuclear Security Administration. All rights
+#  in the program are reserved by Triad National Security, LLC, and the U.S. Department
+#  of Energy/National Nuclear Security Administration. The Government is granted for
+#  itself and others acting on its behalf a nonexclusive, paid-up, irrevocable worldwide
+#  license in this material to reproduce, prepare derivative works, distribute copies to
+#  the public, perform publicly and display publicly, and to permit others to do so.
+# ========================================================================================
+
+import sys
+import numpy as np
+import h5py
+from argparse import ArgumentParser
+
+
+def compute_asymmetry(f, varname):
+    "Computes the asymmetry of var with varname in hdf5 output file object f"
+    xlocs = f["Locations/x"][:]
+    ylocs = f["Locations/y"][:]
+    iylocs = -np.flip(f["Locations/y"][:], axis=1)
+    matches = np.zeros((ylocs.shape[0]), dtype=int)
+    for b in range(ylocs.shape[0]):
+        for bb in range(ylocs.shape[0]):
+            if np.all(np.abs(ylocs[b] - iylocs[bb]) <= 1e-10) and np.all(
+                np.abs(xlocs[b] - xlocs[bb]) <= 1e-10
+            ):
+                matches[b] = bb
+
+    var = f[varname][:]
+    var_diff = np.zeros_like(var)
+    for b in range(ylocs.shape[0]):
+        bb = matches[b]
+        if np.any(ylocs[b] >= 0):
+            var_diff[b] = var[b] - np.flip(var[bb], axis=-2)
+            var_diff[bb] = var[bb] - np.flip(var[b], axis=-2)
+
+    if np.any(np.abs(var_diff) > 1):
+        print("var_diff > 1e-8!", np.where(np.abs(var_diff) > 1))
+
+    return var_diff
+
+
+parser = ArgumentParser(
+    prog="compute_asymmetry.py",
+    description="compute asymmetry in X2 of a field and save it to the output file. Assumes mesh is symmetric about 0 in X2.",
+)
+parser.add_argument("field", type=str, help="Variable to compute")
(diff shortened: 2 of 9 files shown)


Yurlungur:
@pgrete any idea what's going on with the CI here? This looks like a local issue with the CI machine?
https://github.com/parthenon-hpc-lab/parthenon/actions/runs/14647195866/job/41104266637?pr=1248

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env python
```
Yurlungur, line 1: This shebang was just missing before.

Yurlungur:
Thanks for the additional cleanup @lroberts36 ! Looks like builds are now failing? 
https://github.com/parthenon-hpc-lab/parthenon/actions/runs/14650074098/job/41113472608?pr=1248

pgrete:
> @pgrete any idea what's going on with the CI here? This looks like a local issue with the CI machine? https://github.com/parthenon-hpc-lab/parthenon/actions/runs/14647195866/job/41104266637?pr=1248

For whatever reason some lock file was not cleaned up. A manual removal seems to have fixed the issue.

src/mesh/mesh-amr_loadbalance.cpp
```diff
@@ -558,35 +546,32 @@ void Mesh::UpdateMeshBlockTree(int &nnew, int &ndel) {
...
+          for (std::int64_t oj = 0; oj <= lj; oj++) {
+            for (std::int64_t oi = 0; oi <= 1; oi++) {
               if (r < tnderef) {
-                if ((lderef[n].lx1() + i) == lderef[r].lx1() &&
-                    (lderef[n].lx2() + j) == lderef[r].lx2() &&
-                    (lderef[n].lx3() + k) == lderef[r].lx3() &&
-                    lderef[n].level() == lderef[r].level())
+                if (lderef[n].GetSameLevelNeighbor(oi, oj, ok) == lderef[r]) {
                   rr++;
+                }
```
Yurlungur, line 574: this change here was the location of the bug.

Yurlungur (commented):
Thanks @lroberts36 for the additional cleanup. Original bug is below. Luke's changes are non-functional cleanup that better isolate the tree machinery to prevent something like this from being re-introduced.

lroberts36 (approved):
LGTM

Yurlungur:
@pgrete this error looks like the CI machine didn't see any GPUs?

https://github.com/parthenon-hpc-lab/parthenon/actions/runs/14716288427/job/41300312436#step:6:1

doc/sphinx/src/outputs.rst
```diff
@@ -167,6 +167,17 @@ immediately prior to restart files being written with the optional
 callbacks (if provided) will be called in that order before restart files are
 written.
 
+Changing output cadence from command line when restarting from file
+--------------------------------------------------------------------
+
+When restarting from a restart file, you can change any paremeter
```
pgrete:
```suggestion
When restarting from a restart file, you can change any parameter
```

doc/sphinx/src/outputs.rst
```diff
@@ -167,6 +167,17 @@ immediately prior to restart files being written with the optional
...
+input argument. To change the output cadence in this way, however,
+special care is required, as the code stores a ``next_time`` and a
+``next_n`` for the next time to output and next iteration to
+output. Fortunately, these can be overwritten on the command line in
+the usual way. To set the next time to output when restarting, run
+with ``parthenon/output*/next_time=some_number``.
+
```
pgrete, line 182:
```suggestion
Note that this is independent of updating ``parthenon/output*/dt=some_number``.
In other words, to change the cadence and apply the change immediately both ``dt`` (or ``dn``) and ``next_time`` (or ``next_n`` need to be updated simultaneously.

```

Suggestion to be more specific.

scripts/python/packages/parthenon_tools/parthenon_tools/compute_asymmetry.py
```diff
@@ -0,0 +1,65 @@
...
+    var = f[varname][:]
+    var_diff = np.zeros_like(var)
+    for b in range(ylocs.shape[0]):
+        bb = matches[b]
+        if np.any(ylocs[b] >= 0):
+            var_diff[b] = var[b] - np.flip(var[bb], axis=-2)
+            var_diff[bb] = var[bb] - np.flip(var[b], axis=-2)
+
+    if np.any(np.abs(var_diff) > 1):
+        print("var_diff > 1e-8!", np.where(np.abs(var_diff) > 1))
```
pgrete: It's 1e-10, isn't it?
Also any reason to not be even more stringent here?

src/mesh/mesh-amr_loadbalance.cpp
```diff
@@ -542,11 +530,11 @@ void Mesh::UpdateMeshBlockTree(int &nnew, int &ndel) {
   }
 
   // allocate memory for the location arrays
-  LogicalLocation *lref{}, *lderef{}, *clderef{};
-  if (tnref > 0) lref = new LogicalLocation[tnref];
+  std::vector<LogicalLocation> lref, lderef, clderef;
+  if (tnref > 0) lref.resize(tnref);
```
pgrete, line 534: :+1:

src/mesh/mesh-amr_loadbalance.cpp
```diff
@@ -558,35 +546,32 @@ void Mesh::UpdateMeshBlockTree(int &nnew, int &ndel) {
...
-          (lderef[n].lx3() & 1LL) == 0LL) {
+      if (lderef[n].IsLowerLeftCornerOfParent()) {
         int r = n, rr = 0;
-        for (std::int64_t k = 0; k <= lk; k++) {
-          for (std::int64_t j = 0; j <= lj; j++) {
-            for (std::int64_t i = 0; i <= 1; i++) {
+        // offsets from "bottom left" corner of parent block
+        for (std::int64_t ok = 0; ok <= lk; ok++) {
+          for (std::int64_t oj = 0; oj <= lj; oj++) {
+            for (std::int64_t oi = 0; oi <= 1; oi++) {
```
pgrete, line 570: I cannot say that I have fully internalized the updated mesh structure yet, but double checking for my own understanding:
Why are `lk` and `lj` tied to `symmetry` but `i` is not?
I see that our grids are always at least 1D (which I'd understand overall a different behavior), but what determines the use of `symmetry`?
IIRC I had a similar question in the past and it was tied to being periodic but I'm not that sure any more.

pgrete (approved):
LGTM (as far as I was able to follow).

pgrete:
> @pgrete this error looks like the CI machine didn't see any GPUs?
> 
> https://github.com/parthenon-hpc-lab/parthenon/actions/runs/14716288427/job/41300312436#step:6:1

I think the machine is currently under heavy use. I just retriggerd the job (expecting that it'll go through).

src/mesh/mesh-amr_loadbalance.cpp
Yurlungur: @lroberts36 should jump in but... as I understand it, `symmetry` is totally synonymous with dimension. It's essentially, "are we plane-symmetric in this direction?"

scripts/python/packages/parthenon_tools/parthenon_tools/compute_asymmetry.py
Yurlungur: Actually I don't think that print statement should be there at all... It's from when I was debugging. I will remove it, but also will also upload another script, `report_asymmetry` which I think is useful.

src/mesh/mesh-amr_loadbalance.cpp
lroberts36: Yes, `symmetry(dir)` being true means that we have translational symmetry in direction `dir`. Previously in many places in the code we just checked if the size of a block was one in a given direction, but this created some issues with multigrid when multiple block sizes were allowed. I think we could also probably remove the `symmetry` stuff and replace it with `ndim` checks.

Yurlungur: I prefer checking `symmetry`. It's more self-descriptive.

lroberts36: Maybe it should be changed to `TranslationalSymmetry` or something because this isn't the first time someone has been confused about what that call implies.

Yurlungur: A reasonable suggestion for a later MR.

lroberts36: yeah, when I wrote "Maybe it should be changed..." I was thinking the change should occur at some (possibly infinite) time in the future

PR #1253, opened 2025-05-19
Add support for uint64 swarm vars and add default ids

pgrete:
## PR Summary

What it says on the tin.

I was about to add some logic to enforce globally unique ids (which can be added dynamically) based on MPI logic and then thought about a simpler way(which kind of requires a larger pool of numbers, i.e., uint64_t without being too restrictive.
Thus, I added support for that data type first.

When we add another data type to particles, we probably want to start working with additional template magic as I was touching way more code than anticipated.

As it stands right now, the new `swarm.id` field is added by default for all Swarms.
Is this what we want? I kind of lean towards it as it adds little overhead and potentially solves some hassle down the road.
UPDATE: Seems that we want it. The old behavior (i.e., no `id` field) be recovered by adding a swarm with the `Metadata::NoPersistentParticleIds` flag

Another item, I'm currently unsure about is the `id` field for postprocessing tools (like visit or Paraview).
Do those tools require a field named `id` (rather than `swarm.id` or sth else)? We could add an override to the xdmf output.
UPDATE: A default "id" field (with this exact name) is now always added to the xdmf file either pointing to the real unique ids or some automatically generated field (only existing in the output) as before

Last big question: What do we do about backwards compatibility? I imagine this could get a little bit ugly given that the underlying datatype changed.
UPDATE: Should not be an issue any more by adding `Metadata::NoPersistentParticleIds` to the swarm.

Anything else I missed?

## Breaking Changes
A unique particle id field is added by default to each swarm.
If this is not desired (and to recover the old behavior), the field can be disabled by passing the `Metadata::NoPersistentParticleIds`flag.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


Yurlungur:
> As it stands right now, the new swarm.id field is added by default for all Swarms.
Is this what we want? I kind of lean towards it as it adds little overhead and potentially solves some hassle down the road.

I don't think it hurts much, but may be hard to interpret in some cases, e.g., what about Monte Carlo packets, where many may be created and destroyed over the course of a simulation? Peristent IDs for these may be a pain. I say it's fine to always make them exist by default, but I would want some warning labels in the docs/comments. Others should chime in. @pdmullen? 

> Another item, I'm currently unsure about is the id field for postprocessing tools (like visit or Paraview).
Do those tools require a field named id (rather than swarm.id or sth else)? We could add an override to the xdmf output.

The field can be named whatever we like so long as the XML calls it `id` and points at it appropriately. But honestly, I never got the XDMF working for particles. I still don't understand why.

Yurlungur (approved):
LGTM. Though we may want to keep the ability to automatically set the IDs to the global ID in output somehow?

pdmullen:
> > As it stands right now, the new swarm.id field is added by default for all Swarms.
> > Is this what we want? I kind of lean towards it as it adds little overhead and potentially solves some hassle down the road.
> 
> I don't think it hurts much, but may be hard to interpret in some cases, e.g., what about Monte Carlo packets, where many may be created and destroyed over the course of a simulation? Peristent IDs for these may be a pain. I say it's fine to always make them exist by default, but I would want some warning labels in the docs/comments. Others should chime in. @pdmullen?
> 
> > Another item, I'm currently unsure about is the id field for postprocessing tools (like visit or Paraview).
> > Do those tools require a field named id (rather than swarm.id or sth else)? We could add an override to the xdmf output.
> 
> The field can be named whatever we like so long as the XML calls it `id` and points at it appropriately. But honestly, I never got the XDMF working for particles. I still don't understand why.

I'd echo @Yurlungur's thoughts.  For Monte Carlo, we are constantly destroying and creating new particles... The gears are still spinning in my head re: if there are any edge cases where the current changeset leads to issues given the above.  Meanwhile, I'd also like to test this changeset against our downstream code `jaybenne`.

Also, XDMF outputs *do* now work for swarms... I added this functionality when `SwarmPack`s were introduced.  There is already a check in there for an ID field: 

https://github.com/parthenon-hpc-lab/parthenon/blob/develop/src/outputs/parthenon_xdmf.cpp#L252

We might want to change occurences of "id" to `id::name()`

Yurlungur:
> Also, XDMF outputs do now work for swarms... I added this functionality when SwarmPacks were introduced. There is already a check in there for an ID field:

They do??? What was I doing wrong before? Have you tried this with both visit and paraview?

pdmullen:
> > Also, XDMF outputs do now work for swarms... I added this functionality when SwarmPacks were introduced. There is already a check in there for an ID field:
> 
> They do??? What was I doing wrong before? Have you tried this with both visit and paraview?

https://github.com/parthenon-hpc-lab/parthenon/pull/1079
I had only tested with VisIt at that time.

Yurlungur:
> > > Also, XDMF outputs do now work for swarms... I added this functionality when SwarmPacks were introduced. There is already a check in there for an ID field:
> > 
> > 
> > They do??? What was I doing wrong before? Have you tried this with both visit and paraview?
> 
> #1079 I had only tested with VisIt at that time.

I completely forgot about this lol. Thanks again for finding the magic incantation that worked.

pgrete:
Alright, following the discussion I made some updates to the PR.
1. I added a `NoPersistentParticleIds` Metadata flag
2. Adding said flag to particles reverts to the old default behavior (i.e., no "additional" id field)
3. Reintroduced writing default "id"s if the internal field has been disabled
4. Overwrite the internal id variable name in the xdmf file with the common "id" one.
5. Updated the docs to reflect those changes
6. Using the logic in the particle tracer example for testing

src/outputs/outputs.cpp
```diff
@@ -292,6 +294,8 @@ Outputs::Outputs(Mesh *pm, ParameterInput *pin, SimTime *tm) {
         }
 #ifdef ENABLE_HDF5
         op.write_xdmf = pin->GetOrAddBoolean(op.block_name, "write_xdmf", true);
+        // TODO(reviewers) Why are do we forcibly disallow writing particle xdmf for
+        // restarts?
         op.write_swarm_xdmf =
             (restart) ? false
```
pgrete: Any reason for doing this? I'd like to be able to generate particle xdmf for restart outputs (as those are sometime the only type I write).

Yurlungur: @pdmullen ? My guess is it's fine to write particle xdmf for restarts but that since it's an extra file it's just a lot of output?

pdmullen: You have my blessing.  

If I had to guess (it has been a while...), I disabled this because I wanted the imprint of my swarm XDMF MR to be minimal since it requires a bunch of ugly geometry writes.  That code is so ugly (but the only path I could find to make VisIt happy...)

pgrete: I was not suggesting to enable it by default I was suggest to allow enabling it.
Right now the `write_swarm_xdmf` parameter is ignored for restart outputs (i.e., it's impossible to write particle xdmf for restart files.

pdmullen: All good by me :)

pgrete:
@mfournier01 discovered and correctly isolated a casting issue for the uint64 type (as we're always communicating packed `Real`s).

With the latest commit I now use a `std::memcpy` and added static asserts for `int` and `uint64` in the packing and unpacking routines.

@pdmullen were you able to test downstream and/or are there any other changes required?

pdmullen:
> @mfournier01 discovered and correctly isolated a casting issue for the uint64 type (as we're always communicating packed `Real`s).
> 
> With the latest commit I now use a `std::memcpy` and added static asserts for `int` and `uint64` in the packing and unpacking routines.
> 
> @pdmullen were you able to test downstream and/or are there any other changes required?

Thanks for the bug fix (i.e., thanks @mfournier01 !)  @pgrete, I'll have an answer for you re: downstream by no later than tomorrow COB.

pdmullen:
> > @mfournier01 discovered and correctly isolated a casting issue for the uint64 type (as we're always communicating packed `Real`s).
> > With the latest commit I now use a `std::memcpy` and added static asserts for `int` and `uint64` in the packing and unpacking routines.
> > @pdmullen were you able to test downstream and/or are there any other changes required?
> 
> Thanks for the bug fix (i.e., thanks @mfournier01 !) @pgrete, I'll have an answer for you re: downstream by no later than tomorrow COB.

Sorry for the delay on this... something pertinent came up.

doc/sphinx/src/particles.rst
```diff
@@ -256,17 +260,14 @@ per-swarm list will be output for that swarm.
...
-   variable. However, when Parthenon outputs a swarm, it automatically
-   generates an ID variable even if one is not present or
-   requested. If a variable named ``id'' is available **and** the user
-   requests it be output, Parthenon will use it. Otherwise, Parthenon
-   will generate an ``id`` variable just for output and write it to
-   file.
+   general. Therefore, if swarms do not automatically contain a unique ID swarm
+   variable (because the ``Metadata::NoPersistentParticleIds`` has been passed when
+   creating the swarm), Parthenon will generate an ``id`` variable just for output
+   (i.e., it is still not available within the simulation itself) and write it to file.
```
lroberts36, line 266: Does the use still have to request that `id` be output?

example/particle_tracers/particle_tracers.cpp
```diff
@@ -146,10 +148,10 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
 
   // Add swarm of tracer particles
   std::string swarm_name = "tracers";
-  Metadata swarm_metadata({Metadata::Provides, Metadata::None});
+  Metadata swarm_metadata(
+      {Metadata::Provides, Metadata::None, Metadata::NoPersistentParticleIds});
```
lroberts36, line 154: Why is the `id` field being removed here when previously the example explicitly created an `id` field?

src/outputs/outputs.cpp
```diff
@@ -256,10 +256,12 @@ Outputs::Outputs(Mesh *pm, ParameterInput *pin, SimTime *tm) {
...
-            // Always output x, y, and z for swarms so that they work with vis tools.
-            std::vector<std::string> coords = {swarm_position::x::name(),
-                                               swarm_position::y::name(),
-                                               swarm_position::z::name()};
+            // Always output id, x, y, and z for swarms so that they work with vis tools.
+            // Note, it's fine to add the id by default (even though it might not actually
+            // exist) because only variables that do exists are actually being written.
+            std::vector<std::string> coords = {
+                swarm_position::id::name(), swarm_position::x::name(),
+                swarm_position::y::name(), swarm_position::z::name()};
```
lroberts36, line 264: This looks like it answers my question above.

src/pack/swarm_default_names.hpp
```diff
@@ -28,6 +29,9 @@
 
 namespace swarm_position {
 
+// TODO(review) technically not a position, but it feels overkill to add a separate
+// namespace
+SWARM_VARIABLE(std::uint64_t, swarm, id);
```
lroberts36, line 34: I agree this is the natural place to keep the type.

lroberts36 (approved):
LGTM

example/particle_tracers/particle_tracers.cpp
pgrete: I made this change to test the `Metadata::NoPersistentParticleIds` feature.
The particle_tracer regression test results in particles being in the same order, i.e., the automatically generated ids match, and it also tests accessing "id" via that string rather than `swarm.id`.
Given that this might be confusing I now added comments both here and in the test.

PR #1258, opened 2025-06-12
add "corehdf" file that automatically outputs every field

Yurlungur:
## PR Summary

This is a debugging tool that @jdolence and I thought would be helpful and I'm now using. The idea is that sometimes you want absolutely everything, and you don't want to have to enumerate it. In that case, you can request a `corehdf` as your output type, and every variable parthenon is aware of will be output from the specified stage. 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


src/raycasting/raycasting.cpp
```diff
@@ -0,0 +1,74 @@
+//========================================================================================
```
pdmullen: Does this belong in this PR?

src/outputs/outputs.hpp
```diff
@@ -190,7 +190,19 @@ class PHDF5Output : public OutputType {
...
                         hid_t file, const HDF5::H5P &pl, size_t offset,
                         hsize_t max_blocks_global) const;
-  const bool restart_; // true if we write a restart file, false for regular output files
+  std::string FilePostfix_() const {
+    if (mode_ == DumpOutputMode::DUMP) {
+      return ".phdf";
+    } else if (mode_ == DumpOutputMode::RESTART) {
+      return ".rhdf";
+    } else if (mode_ == DumpOutputMode::CORE) {
+      return ".corehdf";
```
pdmullen: Don't feel strongly at all... but I would have chosen `".chdf"` for consistency with above.

src/outputs/outputs.cpp
```diff
@@ -164,7 +168,8 @@ Outputs::Outputs(Mesh *pm, ParameterInput *pin, SimTime *tm) {
       op.analysis_flag = pin->GetOrAddBoolean(op.block_name, "analysis_output", false);
 
       // read single precision output option
-      const bool is_hdf5_output = (op.file_type == "rst") || (op.file_type == "hdf5");
+      const bool is_hdf5_output = (op.file_type == "rst") || (op.file_type == "hdf5") ||
+                                  (op.file_type == "corehdf");
```
pdmullen: "core" or "corehdf5" (to either distinguish or be consistent with `hdf5` output option, respectively)? Don't feel strongly.

src/defs.cpp
```diff
@@ -0,0 +1,39 @@
+//========================================================================================
```
pdmullen: Is this purely associated with the raycasting stuff too?

src/raycasting/raycasting.cpp
Yurlungur: oops no it does not. THanks for the catch.

src/outputs/outputs.hpp
Yurlungur: will change that.

src/outputs/outputs.cpp
Yurlungur: I think it should be `corehdf` to disambiguate from coredumps which are already a thing. That was my reasoning.

src/defs.cpp
Yurlungur: yes will remove.

example/stochastic_subgrid/stochastic_subgrid_package.cpp
```diff
@@ -308,7 +308,7 @@ void DoLotsOfWork(MeshBlockData<Real> *rc) {
 
         // surprisingly, this seems to be almost free
         if (num_iter > 0) {
-          Kokkos::atomic_increment(&hist(num_iter - N_min));
+          Kokkos::atomic_inc(&hist(num_iter - N_min));
```
Yurlungur, line 315: This makes a kokkos warning go away

Yurlungur:
@pdmullen please re-review. Also should I add a regression test? I lean against, as the code path is basically already tested... but I could be convinced to do so.

src/outputs/outputs.cpp
Yurlungur: oh wait I missed the 5. Sure, will make it `corehdf5`.

pdmullen (approved):

pdmullen:
Should:

https://github.com/parthenon-hpc-lab/parthenon/blob/acaa7587a8dc1fdcab2203ab49b01a7def7287c2/src/outputs/outputs.cpp#L159

default to `true` for `corehdf5`?

Yurlungur:
> Should:
> 
> https://github.com/parthenon-hpc-lab/parthenon/blob/acaa7587a8dc1fdcab2203ab49b01a7def7287c2/src/outputs/outputs.cpp#L159
> 
> default to `true` for `corehdf5`?

I lean towards leaving this because keeping defaults consistent probably has some value. Could be convinced to switch it up though.

jdolence (approved):
LGTM

jdolence:
> > Should:
> > https://github.com/parthenon-hpc-lab/parthenon/blob/acaa7587a8dc1fdcab2203ab49b01a7def7287c2/src/outputs/outputs.cpp#L159
> > 
> > default to `true` for `corehdf5`?
> 
> I lean towards leaving this because keeping defaults consistent probably has some value. Could be convinced to switch it up though.

PR #1259, opened 2025-06-16
Add default num partitions per rank

Yurlungur:
## PR Summary

This has been a long time coming and @pgrete 's talk spurred me into action. Untested right now, hence WIP.

MR adds the ability to set number of partitions per rank in the input file. Default is 1 partition per rank.

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build


lroberts36 (approved):
LGTM, thanks!

src/mesh/mesh.cpp
```diff
@@ -79,14 +79,29 @@ Mesh::Mesh(ParameterInput *pin, ApplicationInput *app_in, Packages_t &packages,
...
       nref(Globals::nranks), nderef(Globals::nranks), rdisp(Globals::nranks),
       ddisp(Globals::nranks), bnref(Globals::nranks), bnderef(Globals::nranks),
       brdisp(Globals::nranks), bddisp(Globals::nranks) {
+  // pack size
+  bool pack_size_exists = pin->DoesParameterExist("parthenon/mesh", "pack_size");
+  bool num_partitions_exists =
+      pin->DoesParameterExist("parthenon/mesh", "partitions_per_rank");
+  if (pack_size_exists && num_partitions_exists) {
+    PARTHENON_THROW("Both pack_size and partitions_per_rank set! Set only one");
+  }
```
pgrete: I've been thinking back and forth about this -- particularly with respect to backwards compatibility (do we have a way to remove parameters?)
How about allowing both inputs with `num_packs`  taking precedence and just issue a warning if both are present (and in disagreement)?
I'm not sure if my concern is just academic (i.e., how many simulation will actually be restarted and add the new parameter along the existing one).

src/mesh/mesh.cpp
```diff
@@ -79,14 +79,29 @@ Mesh::Mesh(ParameterInput *pin, ApplicationInput *app_in, Packages_t &packages,
...
       num_mesh_threads_(pin->GetOrAddInteger("parthenon/mesh", "num_threads", 1)),
       use_uniform_meshgen_fn_{true, true, true, true}, lb_flag_(true), lb_automatic_(),
       lb_manual_(), nslist(Globals::nranks), nblist(Globals::nranks),
       nref(Globals::nranks), nderef(Globals::nranks), rdisp(Globals::nranks),
       ddisp(Globals::nranks), bnref(Globals::nranks), bnderef(Globals::nranks),
       brdisp(Globals::nranks), bddisp(Globals::nranks) {
+  // pack size
+  bool pack_size_exists = pin->DoesParameterExist("parthenon/mesh", "pack_size");
+  bool num_partitions_exists =
+      pin->DoesParameterExist("parthenon/mesh", "partitions_per_rank");
```
pgrete: How common is the (external) use of "partitions" versus "pack"?
In AthenaPK discussion, we typically talk about packs so I'm leaning towards `packs_per_rank`.

src/mesh/mesh.cpp
```diff
@@ -79,14 +79,29 @@ Mesh::Mesh(ParameterInput *pin, ApplicationInput *app_in, Packages_t &packages,
...
+  if (pack_size_exists && num_partitions_exists) {
+    PARTHENON_THROW("Both pack_size and partitions_per_rank set! Set only one");
+  }
+  if (pack_size_exists) {
+    use_pack_size_ = true;
+    default_pack_size_ = pin->GetInteger("parthenon/mesh", "pack_size");
```
pgrete: ~This may also need a default init value, doesn't it?~

Never mind. I see the logic with the alternative option being the new default/preferred way.

pgrete (commented):
Should we also add the new param to the docs?

src/mesh/mesh.cpp
Yurlungur: Ahh, I see your point. Yes good suggestion.

src/mesh/mesh.cpp
Yurlungur: I don't care either way. I went with partition since that seemed to mimic the internal language within the code more. I can change it to `packs` though and maybe add some documentation.

src/mesh/mesh.cpp
Yurlungur: Done.

src/mesh/mesh.cpp
Yurlungur: Done.

Yurlungur:
@pgrete I think your comments are addressed. Please re-review.

pgrete (approved):
I just made some minor direct edits:
1. change two leftover places from partitions to packs in the parameter name
2. renamed the internal variable to packs from partitions
3. (most relevant) changed the logic so that packs_per_rank takes precedent when both are set (my logic is here that for backwards compatibility only `pack_size` will be present and no new default is being set. Thus, if someone add `packs_per_rank`, this is done deliberately and thus should be followed.

I tested the logic in AthenaPK and it seems to work.
Please double check those changes and then merge (I already approve so that two approvals are present).

Yurlungur:
> I just made some minor direct edits:
> 
>     1. change two leftover places from partitions to packs in the parameter name
> 
>     2. renamed the internal variable to packs from partitions
> 
>     3. (most relevant) changed the logic so that packs_per_rank takes precedent when both are set (my logic is here that for backwards compatibility only `pack_size` will be present and no new default is being set. Thus, if someone add `packs_per_rank`, this is done deliberately and thus should be followed.
> 
> 
> I tested the logic in AthenaPK and it seems to work. Please double check those changes and then merge (I already approve so that two approvals are present).

Sounds good. Everything looks good to me. Thanks @pgrete !

Issue #1262, opened 2025-06-18
ParameterInput wishlist


Yurlungur:
- [x] Move next output to something like last output and store it internally and in files, not in `ParameterInput`
- [ ] Replace all `ParameterInput` internal nonsense with TOML dictionary
- [ ] For backwards compatiblity the char find-replace
- [ ] Add logic for docstrings in `ParameterInput`
- [ ] Compilation
- [ ] Check which variables are read to print out list of orphans
- [ ] Includes

PR #1266, opened 2025-06-19
Refactor output logic to not stash next output time/iteration in Params

Yurlungur:
## PR Summary

Previously, Parthenon stashed `next_time` and `next_n` in `ParameterInput` so that it was updated continuously and available on restart. This meant that to update output cadence in parthenon, `next_time` and `dt` both had to be updated. This was also problematic from an internal code cleanliness point of view, as `ParameterInput` had hidden state. 

In this MR, I move the "next output" logic into the params of a new `Outputs` package. This stashes the next time logic in params in a file, rather than in the `ParameterInput` string in the `Info` attribute. This is in preparation for the input file updates suggested by @bprather and @adamdempsey90 , as well as the docstring logic I plan to add soon. See #1262 .

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


src/interface/state_descriptor.hpp
```diff
@@ -110,6 +111,60 @@ class StateDescriptor {
...
+    params_.Add(name, val, mutability);
+  }
+  template <typename T, typename... Args>
+  void AddParamFromInput(Params::Mutability mutability, const std::string &block,
+                         const std::string &name, const T &default_val,
+                         ParameterInput *pinput, Args &&...args) {
+    T val = pinput->template GetOrAdd<T>(block, name, default_val,
+                                         std::forward<Args>(args)...);
+    params_.Add(name, val, mutability);
+  }
```
Yurlungur, line 132: I added this set of functions in this MR but they're not really relevant for the bulk of the changes. These are just syntactic sugar.

Yurlungur:
Output test fails because files contain additional param metadata. We will fix by adding a new gold file for this and several other MRs.

Yurlungur:
@bprather @pgrete reminder to please review.

Yurlungur:
I've updated the gold files for regression tests, so this should pass now. @pgrete @bprather please review.

Alternatively, @adamdempsey90 @pdmullen @AstroBarker . I'd like to get this in, as it's ready and has been for about 2 weeks.

src/outputs/output_parameters.hpp
```diff
@@ -53,21 +56,20 @@ struct OutputParameters {
...
-  bool write_swarm_xdmf;
-  // TODO(felker): some of the parameters in this class are not initialized in constructor
-  OutputParameters()
-      : block_number(0), next_time(0.0), dt(-1.0), next_n(0), dn(-1), file_number(0),
-        include_ghost_zones(false), cartesian_vector(false),
-        single_precision_output(false), sparse_seed_nans(false),
-        hdf5_compression_level(5), write_xdmf(false), write_swarm_xdmf(false) {}
+  Real last_time;
+  Real next_time = 0.0;
+  Real dt = 1.0;
```
lroberts36: did you intend to change the default `dt` from -1 to 1?

src/outputs/outputs.cpp
```diff
@@ -101,237 +102,255 @@ OutputType::OutputType(OutputParameters oparams)
...
+  for (int iinput = 0; iinput < block_names.size(); ++iinput) {
+    OutputParameters op; // define temporary OutputParameters struct
+    op.block_name = block_names[iinput];
+    op.block_number = block_numbers[iinput];
+    op.contiguous_block_index = iinput;
+
+    Real dt = 0.0; // default value == 0 means that initial data is written by default
+    int dn = -1;
+    if (tm != nullptr) {
+      dn = pin->GetOrAddInteger(op.block_name, "dn", -1.0);
```
lroberts36:
```suggestion
      dn = pin->GetOrAddInteger(op.block_name, "dn", -1);
```

src/outputs/outputs.cpp
```diff
@@ -101,237 +102,255 @@ OutputType::OutputType(OutputParameters oparams)
...
+
+      // If this is a dn controlled output (dn >= 0), soft disable dt based triggering
+      // (-> dt = -1), otherwise setting dt to tlim ensures a final output is also
+      // written for temporal drivers.
+      const auto tlim = dn >= 0 ? -1 : tm->tlim;
```
lroberts36:
```suggestion
      const auto tlim = dn >= 0 ? -1.0 : tm->tlim;
```

src/outputs/output_parameters.hpp
```diff
@@ -38,7 +38,10 @@ enum class DumpOutputMode { DUMP, RESTART, CORE };
 //! \struct OutputParameters
 //  \brief  container for parameters read from <output> block in the input file
 struct OutputParameters {
-  int block_number;
+  OutputParameters() = default;
```
lroberts36, line 41: It feels to me like a lot of the stuff in `Outputs::Outputs` should be done in the constructor of `OutputParameters`.

src/outputs/output_parameters.hpp
```diff
@@ -53,21 +56,20 @@ struct OutputParameters {
...
-  OutputParameters()
-      : block_number(0), next_time(0.0), dt(-1.0), next_n(0), dn(-1), file_number(0),
-        include_ghost_zones(false), cartesian_vector(false),
-        single_precision_output(false), sparse_seed_nans(false),
-        hdf5_compression_level(5), write_xdmf(false), write_swarm_xdmf(false) {}
+  Real last_time;
+  Real next_time = 0.0;
+  Real dt = 1.0;
+  int last_n;
+  int next_n = 0;
```
lroberts36: Are `last_n`, `last_time`, `next_n`, and `next_time` the only things that change for a given output block after initialization? If so, I would either move them to separate class or document that.

src/outputs/outputs.cpp
```diff
@@ -101,237 +102,255 @@ OutputType::OutputType(OutputParameters oparams)
...
+  // loop over "parthenon/output" blocks located in params
+  auto pkg = pm->packages.Get("Outputs");
+  const auto &block_names = pkg->Param<std::vector<std::string>>("block_names");
+  const auto &block_numbers = pkg->Param<std::vector<int>>("block_numbers");
+  auto *pactive = pkg->MutableParam<std::vector<bool>>("active");
+  auto *pfile_numbers = pkg->MutableParam<std::vector<int>>("file_numbers");
+  auto *plast_times = pkg->MutableParam<std::vector<Real>>("last_times");
+  auto *plast_ns = pkg->MutableParam<std::vector<int>>("last_ns");
+  for (int iinput = 0; iinput < block_names.size(); ++iinput) {
+    OutputParameters op; // define temporary OutputParameters struct
```
lroberts36: Could the output parameters object be created for each block in the outputs package and then stashed in params?

src/outputs/outputs.cpp
```diff
@@ -101,237 +102,255 @@ OutputType::OutputType(OutputParameters oparams)
...
-        pfirst_type_ = pnew_type;
-      } else {
-        plast->pnext_type = pnew_type;
-      }
-      plast = pnew_type;
+    // Append type as tail node in singly linked list
+    if (pfirst_type_ == nullptr) {
+      pfirst_type_ = pnew_type;
+    } else {
+      plast->pnext_type = pnew_type;
```
lroberts36: Is this an opportunity to replace the linked list with a vector?

src/outputs/output_parameters.hpp
lroberts36: I guess `file_number` changes too.

src/outputs/ascent.cpp
```diff
@@ -209,16 +209,7 @@ void AscentOutput::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm,
...
-  pin->SetInteger(output_params.block_name, "file_number", output_params.file_number);
-  if (output_params.dt > 0.0) {
-    output_params.next_time += output_params.dt;
-    pin->SetReal(output_params.block_name, "next_time", output_params.next_time);
-  }
-  if (output_params.dn > 0) {
-    output_params.next_n += output_params.dn;
-    pin->SetInteger(output_params.block_name, "next_n", output_params.next_n);
-  }
+  UpdateNextOutput_(pm, tm);
```
lroberts36, line 212: Could you just call `ptype->UpdateNextOutput_(pm, tm)` after `WriteOutputFile(...)` in `Outputs::MakeOutputs` so this doesn't need to be called within every type of output?

src/outputs/parthenon_hdf5.cpp
```diff
@@ -104,6 +108,13 @@ void PHDF5Output::WriteOutputFileImpl(Mesh *pm, ParameterInput *pin, SimTime *tm
   // open HDF5 file
   // Define output filename
   auto filename = GenerateFilename_(pin, tm, signal);
+  if (signal == SignalHandler::OutputSignal::none) {
+    // After file has been opened with the current number, already advance output
+    // parameters so that for restarts the file is not immediatly overwritten again.
+    // Only applies to default time-based data dumps, so that writing "now" and "final"
+    // outputs does not change the desired output numbering.
+    UpdateNextOutput_(pm, tm);
+  }
```
lroberts36, line 117: This looks like it was the previous behavior, but I think that only some of the output types deal with the signal. Should all output types only update the next output time if the signal handler is `none`?

lroberts36 (approved):
LGTM, some non-blocking suggestions below

src/outputs/output_parameters.hpp
Yurlungur: Oops nope. Will revert thanks.

src/outputs/output_parameters.hpp
Yurlungur: That's a fair thought. I'm not sure I agree, but I started an issue. #1278

src/outputs/output_parameters.hpp
Yurlungur: :+1: Added a comment and moved them into a separate block of the class. Agree they could be split out and enforced but I think that's a larger refactor.

src/outputs/outputs.cpp
Yurlungur: Sure---is there a reason to do that differently than how I've currently done it? Params currently doesn't know how to dump this struct so it wouldn't be possible to dump it to outputs.

src/outputs/ascent.cpp
Yurlungur: ...yes. Good suggestion.

src/outputs/parthenon_hdf5.cpp
Yurlungur: Agreed. I'll change it. (Assuming that doesn't break tests...)

src/outputs/outputs.cpp
lroberts36: This comment was predicated on moving `OutputParameters` initialization to the `OutputParameters` constructor. I mostly just thought that would be cleaner.

How hard would it be to make POD classes dumpable from params?

src/outputs/ascent.cpp
Yurlungur: oh actually no, this doesn't quite work. Because different output types handle this differently, and it needs to be updated at different times, specifically for restarts.

src/outputs/outputs.cpp
Yurlungur: I could always serialize arbitrary classes and dump that... but it wouldn't be human readable.

src/outputs/parthenon_hdf5.cpp
Yurlungur: Ah actually I think it doesn't matter, because if the signal handler isn't none, the simulation is going to end, so it doesn't matter and output signal doesn't need to be updated. The reason for the signal check in parthenon_hdf5 is because it's BEFORE the output is actually written. So I don't think adding checks to the other outputs will actually change behavior in any way. So I lean towards leaving it, as that's fewer LOC.

src/outputs/outputs.cpp
Yurlungur: Yes. Done. While I'm at it, I am also deleting the vestigal `OutputData` linked list. That was supposed to be used for slices, but I think in practice when we implement slicing, we'll want to do it differently than Athena++ did it, so best to just start from scratch.

src/outputs/outputs.hpp
```diff
@@ -46,20 +46,6 @@ namespace parthenon {
...
-
-struct OutputData {
-  std::string type; // one of (SCALARS,VECTORS) used for vtk outputs
-  std::string name;
-  ParArrayND<Real> data; // array containing data (usually shallow copy/slice)
-  // ptrs to previous and next nodes in doubly linked list:
-  OutputData *pnext, *pprev;
-
-  OutputData() : pnext(nullptr), pprev(nullptr) {}
-};
```
Yurlungur, line 61: This is vestigal. It was for slices in Athena++, but I don't think we want it. When we implement slicing, we'll almost certainly do it differently. So I'm deleting this code.

src/outputs/outputs.cpp
```diff
@@ -348,111 +361,8 @@ Outputs::Outputs(Mesh *pm, ParameterInput *pin, SimTime *tm) {
...
-
-// destructor - iterates through singly linked list of OutputTypes and deletes nodes
-
-Outputs::~Outputs() {
-  OutputType *ptype = pfirst_type_;
-  while (ptype != nullptr) {
-    OutputType *ptype_old = ptype;
-    ptype = ptype->pnext_type;
-    delete ptype_old;
-  }
```
Yurlungur, line 390: moving the collection of OutputTypes from a linked list to a std::vector<std::shared_ptr> means we can get rid of all this code by being a little clever.

src/outputs/outputs_package.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+      // these are used for book-keeping
+      block_names.push_back(block_name);
+      block_numbers.push_back(atoi(outn.c_str()));
+
+      // These will be updated later or restarted from
+      active.push_back(false);
+      file_numbers.push_back(0);
+      // JMM: Limits to indicate these haven't been set yet
+      last_times.push_back(std::numeric_limits<Real>::lowest());
```
acreyes, line 81: does it make sense to use `parthenon/time/start_time` here?

src/outputs/outputs.cpp
```diff
@@ -86,252 +86,265 @@ namespace parthenon {
...
+                             "dt and dn are enabled for the same output block, which "
+                             "is not supported. Please set at most one value >= 0.");
+
+    // set time of last output, time between outputs
+    op.last_time = (*plast_times)[iinput];
+    op.last_n = (*plast_ns)[iinput];
+    if (tm != nullptr) {
+      op.dt = dt;
+      op.dn = dn;
+      // TODO(JMM): Should this be a check for pmesh->is_restart instead?
```
acreyes: I can see use case where `parthenon/time/start_time` >0 to initialize from an analytic self-similar solution where it would be nice to have the output interval start at the initial time.

src/parameter_input.hpp
```diff
@@ -118,8 +120,35 @@ class ParameterInput {
...
+    if constexpr (std::is_same_v<T, int>) {
+      return GetOrAddInteger(block, name, value, std::forward<Args>(args)...);
+    } else if constexpr (std::is_same_v<T, Real>) {
+      return GetOrAddReal(block, name, value, std::forward<Args>(args)...);
+    } else if constexpr (std::is_same_v<T, bool>) {
+      return GetOrAddBoolean(block, name, value, std::forward<Args>(args)...);
+    } else {
+      PARTHENON_THROW("Unknown type\n");
+    }
+  }
```
acreyes: One frustrating issue I could see in a downstream would be
```
std::size_t my_default = GetMyDefault();
...
auto value = GetOrAdd(..., my_default);
```

acreyes (approved):
Looks good!

src/outputs/outputs_package.cpp
Yurlungur: I thought about that. The reason I decided to go with a signaling number, even though it's kind of a gross hack, is because we want to make sure that we have a "first" output. I.e., when the simulation starts, an initial dump is created. 

With the current logic, the process is the following:
1. Simulation starts at start time
2. the outputs package decides to set `last_output` to lowest Real
3. outputs.cpp detects that `last_output` is this dummy value, and sets the next output time to the current time (i.e., start time)
4. An output is performed, the next output will be `start_time + dt`.

Alternatively, one could set `last_output` to `start_time - dt` but I think that's actually a bit more confusing and fraught.

src/outputs/outputs.cpp
Yurlungur: It does that in the current model. See my answer above.

src/parameter_input.hpp
Yurlungur: Good point. I'll add a special case for size_t and uint64. It's a bit dangerous but probably fine and worth the risk.

src/outputs/outputs_package.cpp
Yurlungur: I'll add a comment explaining this

src/parameter_input.hpp
acreyes: Could something like this work? https://godbolt.org/z/dzaW8xnev
Would also let you add the string overload

Yurlungur: Do we want `char`? If so, sure. I can make that change.

acreyes: more the suggestion was to use `is_integral_v<T>`

Yurlungur: I know, but `is_integral_v<T>` returns true for `char` types.

Yurlungur: Done.

Yurlungur:
@pgrete what is going on here? it looks like the test passed successfully, but the job aborted for some reason
https://github.com/parthenon-hpc-lab/parthenon/actions/runs/16056062529/job/45310641665?pr=1266

pgrete:
> @pgrete what is going on here? it looks like the test passed successfully, but the job aborted for some reason https://github.com/parthenon-hpc-lab/parthenon/actions/runs/16056062529/job/45310641665?pr=1266

It seems to repeatedly fail (also for the schedule nightly pipeline).
There's no error message on the host and enough resources are definitely available.
I'll have to investigate more closely next week.

PR #1268, opened 2025-06-20
Add build provenance to output

AstroBarker:
## PR Summary
This PR exposes build information into the source and output. I have exposed:
- Git hash
- Git branch
- Compiler and version
- Build timestamp
- CMake Optimization level
- Build architecture
- Kokkos configuration

The Git hash and branch are integrated in a way to creates a make target and fills in `src/provenance.cpp.in` with the necessary information. It is checked at make time without the need to reconfigure or rebuild the whole code.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


cmake/Provenance.cmake
```diff
@@ -0,0 +1,119 @@
...
+#========================================================================================
+# Originally based on code by Jonathan Hamberg
+# https://gitlab.com/jhamberg/cmake-examples/-/tree/master/cmake
+set(CURRENT_LIST_DIR ${CMAKE_CURRENT_LIST_DIR})
+if (NOT DEFINED pre_configure_dir)
+    set(pre_configure_dir ${CMAKE_SOURCE_DIR}/src)
+endif ()
+
+if (NOT DEFINED post_configure_dir)
+    set(post_configure_dir ${CMAKE_SOURCE_DIR}/src/generated)
```
lroberts36: Should we be generating this in `CMAKE_SOURCE_DIR` or in `CMAKE_BINARY_DIR`? The binary dir seems cleaner if possible so that we aren't putting untracked files in the source tree. I guess another option would be to add `src/generate/*` to `.gitignore`

cmake/Provenance.cmake
```diff
@@ -0,0 +1,119 @@
...
+    add_custom_target(ParthenonAlwaysCheckGit COMMAND ${CMAKE_COMMAND}
+        -DRUN_CHECK_GIT_VERSION=1
+        -Dpre_configure_dir=${pre_configure_dir}
+        -Dpost_configure_file=${post_configure_dir}
+        -DGIT_HASH_CACHE=${GIT_HASH_CACHE}
+        -P ${CURRENT_LIST_DIR}/Provenance.cmake
+        BYPRODUCTS ${post_configure_file}
+        )
+
+        add_library(git_version ${CMAKE_SOURCE_DIR}/src/generated/provenance.cpp)
```
lroberts36:
```suggestion
    add_library(git_version ${CMAKE_SOURCE_DIR}/src/generated/provenance.cpp)
```

cmake/Provenance.cmake
```diff
@@ -0,0 +1,119 @@
...
+# This is used to run this function from an external cmake process.
+if (RUN_CHECK_GIT_VERSION)
+    CheckGitVersion()
+endif ()
+
+# Other information:
+# Compiler
+# build options
+# timestamp
+string(TIMESTAMP PARTH_BUILD_TIMESTAMP "%Y-%m-%d %H:%M:%S %Z")
```
lroberts36, line 113: Does this give the timestamp when `cmake ..` is called or when `cmake --build . ` is called? I think the former, but I guess enforcing the latter would mean forcing a rebuild of `provenance.cpp` every time you call `cmake --build .`, as well as relinking everything.

cmake/Provenance.cmake
```diff
@@ -0,0 +1,119 @@
...
+        # Set che GIT_HASH_CACHE variable the next build won't have
+        # to regenerate the source file.
+        CheckGitWrite(${GIT_HASH})
+
+        configure_file(${pre_configure_file} ${post_configure_file} @ONLY)
+    endif ()
+
+endfunction()
+
+function(CheckGitSetup)
```
lroberts36, line 86: I haven't completely wrapped my head around `CheckGitSetup` but I *think* it is triggered every time you call `cmake --build .` but only triggers a rebuild of `provenance.cpp` if the git hash has changed?

lroberts36 (approved):
LGTM, this will be a nice feature to have! 

The only other thing I can think of is that it would be nice if we could run `my_parthenon_code --parthenon-version` and it would spit out this information and quit. Maybe for another PR though

cmake/Provenance.cmake
Yurlungur: :+1: I agree. I think it should go in `CMAKE_BINARY_DIR` but `.gitignore` would be an acceptable solution too

cmake/Provenance.cmake
```diff
@@ -0,0 +1,119 @@
...
+    )
+
+    CheckGitRead(GIT_HASH_CACHE)
+    if (NOT EXISTS ${post_configure_dir})
+        file(MAKE_DIRECTORY ${post_configure_dir})
+    endif ()
+
+    if (NOT EXISTS ${post_configure_dir}/provenance.hpp)
+        file(COPY ${pre_configure_dir}/provenance.hpp DESTINATION ${post_configure_dir})
+    endif()
```
Yurlungur, line 68: I don't fully understand why this is necessary... shouldn't it already be in our `src` directory? Or is this a way of exposing things downstream?

Yurlungur (changes requested):
LGTM. I'd like to add the ability to have a downstream code register a git hash and branch with this machinery too, and I think that it would be straightforward to build on top of what you have here. For another time though.

Only blocking change is @lroberts36 's comment about either the .gitignore or the binary directory for `post_configure_dir`.

cmake/Provenance.cmake
AstroBarker: It is indeed the former. Not as accurate as it could be perhaps, but avoids lots of relinkings/buildings.

cmake/Provenance.cmake
AstroBarker: Yep!

cmake/Provenance.cmake
AstroBarker: Generated files now live in `$CMAKE_BINARY_DIR/generated`

cmake/Provenance.cmake
AstroBarker: This ensures that there is a header file in `$CMAKE_BINARY_DIR/generated` so that the `cpp` created there compiles.

Yurlungur (approved):
Thanks for pulling this through!

PR #1280, opened 2025-07-03
Output header to history files on each invocation, even restarts

c-prather:
This was briefly discussed at the meeting, but for everyone:

When restarting existing simulations parthenon does not print the header on `.hst` history output files.

This was inadequate in two cases: restarts which used a new directory printed header-less histories, and restarts with new packages loaded added a bunch of unlabeled columns.

This outputs the header once each time Parthenon is invoked, which will always print headers when necessary, at the cost of sometimes printing unnecessary headers for routine restarts.  Given it's easy to tell e.g. numpy to ignore these lines, that does not seem like too bad a price, when the alternative is a bunch of logic for whether a file exists and whether its existing columns (which we'd need to read and parse) match the new set of columns.

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build

### DIFF
--- a/src/outputs/history.cpp
+++ b/src/outputs/history.cpp
@@ -30,6 +30,7 @@
 #include <sstream>
 #include <stdexcept>
 #include <string>
+#include <unordered_set>
 #include <vector>
 
 #include "coordinates/coordinates.hpp"
@@ -180,11 +181,11 @@ void HistoryOutput::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm,
       PARTHENON_FAIL(msg);
     }
 
-    // Write the header lines once per invocation of Parthenon -- output variables may
-    // have changed, or we may be restarting in a new directory with no existing
-    // output file
-    static bool first_invocation = true;
-    if (first_invocation) {
+    // Write the header lines once per distinct history file, each once
+    // per invocation of parthenon -- once a header has been output for
+    // the given file_id, ignore it.
+    static std::unordered_set<std::string> ids_output;
+    if (!ids_output.count(output_params.file_id)) {
       int iout = 1;
       std::fprintf(pfile, "#  History data\n"); // descriptor is first line
       std::fprintf(pfile, "# [%d]=time    ", iout++);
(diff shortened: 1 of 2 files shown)


Yurlungur (approved):
LGTM

src/outputs/history.cpp
```diff
@@ -179,8 +179,11 @@ void HistoryOutput::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm,
       PARTHENON_FAIL(msg);
     }
 
-    // If this is the first output, write header
-    if (output_params.file_number == 0) {
+    // Write the header lines once per invocation of Parthenon -- output variables may
```
pgrete: This pattern results in the header info only being written for the first history output, doesn't it?

c-prather: Not sure what you mean...  The old pattern wrote it once per simulation, only at output 0, so any restarts would not re-write (since output number was re-read at restart to be nonzero).  The new invocation has a static variable which is not stored persistently.

pgrete: With "first history output" I was referring to the first history output block as we support having multiple distinct history outputs.
Thus, only the first one will get a header (and will get a header on every execution) but all other history files will never receive a header.

AstroBarker (approved):
Looks good to me, @pgrete's concern aside.

src/outputs/history.cpp
c-prather: OH now I understand sorry, I completely forgot about that case.  I'll just switch to a static `vector<bool>` for each output.

c-prather:
Updated.  It bothers me somewhat that we don't have a way of telling different outputs apart without consulting a string (and a user-controlled string which is easy to mistakenly repeat, at that), but that's for another time.

src/outputs/history.cpp
```diff
@@ -180,8 +181,11 @@ void HistoryOutput::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm,
       PARTHENON_FAIL(msg);
     }
 
-    // If this is the first output, write header
-    if (output_params.file_number == 0) {
+    // Write the header lines once per distinct history file, each once
+    // per invocation of parthenon -- once a header has been output for
+    // the given file_id, ignore it.
+    static std::unordered_set<std::string> ids_output;
```
Yurlungur, line 187: excellent use of unordered set

Yurlungur (approved):

Yurlungur:
@pgrete @bprather is this ready for merge?

pgrete (approved):
Just tested and works!

PR #1283, opened 2025-07-07
docstrings and provenance

Yurlungur:
## PR Summary

As we discussed in the parthenon workshop, this MR implements "docstrings" in a similar way to that suggested by @acreyes . Every `Get` function in `ParameterInput` now accepts an optional "docstring" argument.

Whenever the ParameterInput class is interacted with, a query record is updated, which stores the default value, type, allowed values, limited provenance information, and docstring for a given parameter. These are updated and validated. So, for example, Parthenon now warns if any parameters in the input deck are unused.

With the `-p` flag, you can have parthenon output this information in CSV format, which can be read by the docs using a `csv-table` in `sphinx`. The `-p` flag optionally takes a regex which you can use to specify which blocks you'd like to output. Blocks are, by default, separated by an empty line, which makes for a cleaner separation in the table and makes the table a little more readable. My thinking with the CSV format is that this is easy to parse and portable, so downstream codes can process it manually if they'd like to. For example, Artemis could use this to automatically generate the docs in the format it already imposes by hand.

I went through parthenon and, in an incomplete way, documented the most common parameters with these docstrings. As a proof of concept, I also added cmake targets to generate these tables for two of the executables we build. This should get integrated into the CI for the docs build so that the tables are automatically generated and used.

I am pretty happy with how this came out, though it will need to be hand-merged with #1277 (sorry @bprather ). I would like feedback, as this should meet everyone's needs ideally.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


doc/sphinx/_static/custom.css
```diff
@@ -0,0 +1,15 @@
+/* Enable word wrap in table cells */
+/*
+.rst-content table.docutils td {
+  white-space: normal !important;
+  word-wrap: break-word;
+  word-break: break-word;
+}
+*/
```
Yurlungur, line 13: I originally used this version, but it messed up our hand-written ultrawide tables, so I changed it to the one below.

doc/sphinx/src/generated/.gitignore
Yurlungur, line 1: The csv files that get generated in the doc build end up in doc/sphinx/src/generated. A local gitignore prevents them from getting committed.

example/diffusion/CMakeLists.txt
```diff
@@ -25,4 +25,28 @@ if( "diffusion-example" IN_LIST DRIVER_LIST OR NOT PARTHENON_DISABLE_EXAMPLES)
...
+    VERBATIM
+    )
+  add_custom_command(
+    OUTPUT ${DOC_GEN_PATH}/diffusion-table.csv
+    COMMAND ${CMAKE_COMMAND} -E env
+    bash -c
+    "${CMAKE_CURRENT_BINARY_DIR}/diffusion-example -i ${CMAKE_CURRENT_LIST_DIR}/parthinput.diffusion -p '^(?!.*parthenon/).*' > ${DOC_GEN_PATH}/diffusion-table.csv"
+    DEPENDS diffusion-example parthinput.diffusion
+    VERBATIM
+    )
```
Yurlungur: I considered making a cmake macro to do this automatically as a tool for downstream and more examples but I'm a little torn. It's a little opaque and it requires a lot of manual nonsense so I'm not sure how generalizable it is.

src/argument_parser.hpp
```diff
@@ -126,12 +133,16 @@ class ArgParse {
...
   bool analysis_flag = false;
   int res_flag = 0;
-  int narg_flag = 0;
+  int param_flag = 0;
   int mesh_flag = 0;
   int wtlim = 0;
   int exit_flag = 0;
+
+ private:
+  char default_params_regex_[1024] = "(.*)"; // wildcard
```
Yurlungur: This felt a little evil, but also the cleanest solution

src/driver/driver.cpp
```diff
@@ -42,6 +42,7 @@ Kokkos::Timer Driver::timer_LBandAMR;
 
 void Driver::PreExecute() {
   if (Globals::my_rank == 0) {
+    pinput->CheckOrphans();
```
Yurlungur: This warns on missing parameters

src/driver/driver.cpp
```diff
@@ -65,6 +66,19 @@ void Driver::PostExecute(DriverStatus status) {
...
+                            ? OutputSignal::analysis
+                            : OutputSignal::none;
+  int perf_cycle_offset = pinput->GetOrAddInteger(
+      "parthenon/time", "perf_cycle_offset", 0,
+      "don't measure performance for some number of initial cycles");
+
+  if (Globals::output_params_and_exit && Globals::my_rank == 0) {
+    pinput->OutputParameterTable(std::cout, std::regex(Globals::params_block_regex));
+    return DriverStatus::complete;
+  }
```
Yurlungur: The ordering here is a little tricky. This has to be done after setup (so that pgens are captured) but before any actual looping. I chose this conditional logic here.

src/globals.cpp
```diff
@@ -44,6 +46,11 @@ Real receive_boundary_buffer_timeout;
 // timeouts for tasks
 Real current_task_runtime_sec;
 
+// logic for parsing and outputting params
+bool output_params_and_exit = false;
+// cpplint says global strings are bad
+char params_block_regex[PARAMS_STRING_LEN];
```
Yurlungur: This is gross but it seemed like the best option.

Yurlungur:
We'll see if the docs build through the CI. If they do, I will share here. Here's what they look like on my machine:
image

doc/sphinx/src/generated/.gitignore
Yurlungur: I have changed the model here. This comment is no longer relevant.

Yurlungur:
The parameter tables generated during the CI can in principle change between branches, so I decided to put them under version control, despite the fact that they are generated by the CI.

The CI now *checks* that they are up to date and asks the developer to run a command if they are not. We could in principle make this a par-hermes command, but I didn't do that here. You have to type a 1 line make command, that has no dependencies.

Yurlungur:
Automatically generated tables now live
https://parthenon-hpc-lab.github.io/parthenon/jmm/docstrings/src/0_input_parameters_ref.html

.github/workflows/docs.yml
```diff
@@ -21,14 +21,40 @@ jobs:
...
+            make param-docs
+            cd ..
+        - name: Check generated files for this branch are up to date
+          run: |
+            if git diff --exit-code -- doc/sphinx/src/generated/; then
+              echo "Generated doc files up to date."
+            else
+              echo "Generated doc files not up to date. Run make param-docs and commit changes."
+              exit 1
+            fi
```
Yurlungur, line 52: We now build the code and run the executable with -p to generate the docs tables. Those docs tables are stored in version control because they may vary between branches. To ensure the correct version is stored, the CI compares against the committed version.

You can build them yourself and update them with `make param-docs`.

doc/sphinx/conf.py
```diff
@@ -50,5 +50,10 @@
 # so a file named "default.css" will overwrite the builtin "default.css".
 html_static_path = ["_static"]
 
+
+def setup(app):
+    app.add_css_file("custom.css")  # Sphinx >= 1.8
```
Yurlungur, line 55: ensures that the csv tables are properly word-wrapped

src/mesh/mesh.cpp
```diff
@@ -1257,4 +1262,48 @@ Mesh::GetBCsFromNames_(const std::array<std::string, BOUNDARY_NFACES> &names) co
...
+                       "number of cells on base mesh in x2 direction"),
+       pin->GetInteger("parthenon/mesh", "nx3",
+                       "number of cells on base mesh in x3 direction")},
+      {false, pin->GetInteger("parthenon/mesh", "nx2") == 1,
+       pin->GetInteger("parthenon/mesh", "nx3") == 1});
+
+  RegionSize base_block_size = GetBaseMeshBlockSize(pin, mesh_size);
+
+  return std::make_pair(mesh_size, base_block_size);
+}
```
Yurlungur: these functions reduce duplicated code. They also are exposed downstream so downstream codes can get these region sizes without having to spin up all this nonsense. That's useful in problem generators.

Yurlungur:
I think tests should all pass. This is now ready for review. @pgrete @pdmullen @jdolence @AstroBarker @acreyes @adamdempsey90 @bprather

doc/sphinx/src/0_input_parameters_ref.rst
```diff
@@ -0,0 +1,48 @@
...
+terminal.
+
+Every call to ``ParameterInput::Get*`` optionally takes a "docustring"
+as a final argument, which is the details column listed here. Default
+values are recorded if they are available.
+
+In Parthenon
+--------------
+
+An incomplete list of all input parameters provided by Parthenon is tabulated below
```
lroberts36, line 20: Why is the list incomplete?

src/parameter_input.hpp
```diff
@@ -150,9 +252,10 @@ class ParameterInput {
     }
   }
 
-  template <typename T, typename... Args>
-  std::vector<T> GetVector(const std::string &block, const std::string &name,
-                           Args &&...args) {
+  template <typename T>
+  std::vector<T>
+  GetVector(const std::string &block, const std::string &name,
+            const std::optional<std::string> &docstring = std::optional<std::string>{}) {
```
lroberts36, line 327: Is the `docstring` unused for `GetVector` and `GetOrAddVector`?

src/parameter_input.hpp
```diff
@@ -245,7 +352,108 @@ class ParameterInput {
...
+        }
+      }
+      // Allowed values are checked after a query, so this function
+      // will be called twice: once with no allowed values and once
+      // with them. This check ensures that validation for allowed
+      // values only happens if they're both active.
+      if ((allowed_vals.size() > 0) && (record.allowed_values.size() > 0)) {
+        PARTHENON_REQUIRE_THROWS(allowed_vals.size() == record.allowed_values.size(),
+                                 "Allowed values must be consistently shaped");
+        std::size_t i = 0;
```
lroberts36, line 487: should `allowed_vals` be sorted before this check? Maybe a little tricky since `Container_t` appears to be an arbitrary container (which may or may not be sortable).

src/parameter_input.hpp
```diff
@@ -245,7 +352,108 @@ class ParameterInput {
...
+        for (const auto &allowed : allowed_vals) {
+          PARTHENON_REQUIRE_THROWS(allowed == std::any_cast<T>(record.allowed_values[i]),
+                                   "Allowed values must be consistent");
+        }
+      } else if (allowed_vals.size() > 0) {
+        for (const auto &allowed : allowed_vals) {
```
lroberts36, line 494: sort `allowed_vals` first?

src/parameter_input.cpp
```diff
@@ -883,6 +944,25 @@ void ParameterInput::CheckDesired(const std::string &block, const std::string &n
...
+  }
+  std::stringstream msg;
+  msg << "The following input parameters are set but unused:\n";
+  for (const auto &[b, p] : orphans) {
+    msg << b << "/" << p << "\n";
+  }
+  msg << std::endl;
+  PARTHENON_WARN(msg);
+}
+
```
lroberts36, line 990: Nice!

src/parameter_input.cpp
```diff
@@ -914,6 +994,70 @@ void ParameterInput::ParameterDump(std::ostream &os) {
...
+  std::map<std::string, std::map<std::string, std::string>> csvblocks;
+  for (InputBlock *pb = pfirst_block; pb != nullptr; pb = pb->pnext) {
+    const std::string &block_name = pb->block_name;
+    if (std::regex_match(block_name, block_regex)) {
+      auto &csvlines = csvblocks[block_name];
+      for (InputLine *pl = pb->pline; pl != nullptr; pl = pl->pnext) {
+        const std::string &param_name = pl->param_name;
+        auto record_key = std::make_pair(block_name, param_name);
+        // This ensures the code doesn't crash for orphan parameters
+        /* clang-format off */
```
lroberts36:
```suggestion
        /* clang-format off */
        // This ensures the code doesn't crash for orphan parameters
```

src/parameter_input.hpp
```diff
@@ -94,31 +167,60 @@ class ParameterInput {
   void LoadFromFile(IOWrapper &input);
   void ModifyFromCmdline(int argc, char *argv[]);
   void ParameterDump(std::ostream &os);
+  // TODO(JMM): Make this more general?
+  void OutputParameterTable(std::ostream &os, const std::regex &block_regex) const;
```
c-prather: Can we get an overload without the regex?

src/parameter_input.hpp
```diff
@@ -24,20 +24,93 @@
...
 #include "utils/hash.hpp"
 #include "utils/string_utils.hpp"
+#include "utils/utils.hpp"
 
 namespace parthenon {
 
+struct QueryRecord {
+  // TODO(JMM): Update this with more provenance information
+  enum class OriginType { None, Input, Default, SetInCode };
+  OriginType origin_type = OriginType::Input;
```
c-prather, line 74: Why is input the default?  Will we get entries for "restart file" and "command line"?  Is there a way to query this for a given parameter that I'm missing?

src/driver/driver.cpp
c-prather: Can we get an option to skip this (in code or as a parameter)?  I imagine this will spit a lot of nonsense for some downstreams for a while.

src/parthenon_manager.cpp
```diff
@@ -227,6 +229,14 @@ void ParthenonManager::ParthenonInitPackagesAndMesh(
...
   }
 
+  if (arg.param_flag) {
+    Globals::output_params_and_exit = true;
+    // JMM: cpplint doesn't like strcopy. snprintf is safer because of
+    // max buffer size parameter
+    std::snprintf(Globals::params_block_regex, Globals::PARAMS_STRING_LEN, "%s",
+                  arg.params_regex);
+  }
+
```
lroberts36: Why not just add a couple of parameters to `pinput` representing these? That would avoid the globals and allow triggering parameter output directly from the parameter file.

lroberts36 (approved):
Nice, just a couple of small comments below. Thanks for going through and documenting all the parameters in addition to adding the docstring functionality.

c-prather (approved):
LGTM (FWIW, given I don't plan to use this downstream).

The provenance stuff is half-baked here (see questions) but I understand now you're saving it for after merging with the TOML parameters, which makes good sense to me.  It'll make the merge easier, and will be easier to implement on top of that code than it would here anyway.

doc/sphinx/src/0_input_parameters_ref.rst
Yurlungur: Two reasons:
1. I didn't actually document everything.
2. It's generated from just one example. To capture everything, I'd have to run it through ALL the examples and process it to remove duplicates. Which I could do, but I figured this was a good start.

src/parameter_input.hpp
Yurlungur: oops... not on purpose. I'll fix that.

pdmullen:
Will leave review shortly... but one quick question came up during a meeting: 

What happens with

```
const bool do_that = pin->GetOrAdd("block", "do_that", true);
const bool do_this = do_that;
pin->GetOrAdd("block", "do_this", do_this);
```

src/parameter_input.hpp
Yurlungur: in practice, it's always a `std::vector`, and I could enforce that. Sorting this is a good idea.

Yurlungur: I can use SFINAE/concepts to check if its sortable and sort. To sort though, I need a copy, or I need to make it not a const reference. Will implement.

lroberts36: I guess the other option is to just do a nested iteration through both lists. I assume that there will never be a huge list of allowed values (?), so the O(N^2) shouldn't really matter.

src/driver/driver.cpp
Yurlungur: Sure

src/parthenon_manager.cpp
Yurlungur: sure that's a good idea

src/parameter_input.hpp
Yurlungur: Sure

src/parameter_input.hpp
Yurlungur: Input is shorthand for  file, restart, and CLI. This is half baked for now, let's fix it in a later MR on top of this + the TOML stuff.

Yurlungur:
> Will leave review shortly... but one quick question came up during a meeting:
> 
> What happens with
> 
> ```
> const bool do_that = pin->GetOrAdd("block", "do_that", true);
> const bool do_this = do_that;
> pin->GetOrAdd("block", "do_this", do_this);
> ```

That will depend on the code path the code takes. It won't default to true, it'll default to whatever `do_that` was set to in the input deck. I don't see a way around this completely, and this isn't a complete replacement for hand-written documentation.

My suggestion for cases like this is to specify in the docstring how this works.

lroberts36:
> > Will leave review shortly... but one quick question came up during a meeting:
> > What happens with
> > ```
> > const bool do_that = pin->GetOrAdd("block", "do_that", true);
> > const bool do_this = do_that;
> > pin->GetOrAdd("block", "do_this", do_this);
> > ```
> 
> That will depend on the code path the code takes. It won't default to true, it'll default to whatever `do_that` was set to in the input deck. I don't see a way around this completely, and this isn't a complete replacement for hand-written documentation.
> 
> My suggestion for cases like this is to specify in the docstring how this works.

We could overload the default argument to also take a `parameter_pointer("block", "other parameter")`, then the ordering wouldn't matter. Obviously this would require some changes under the hood, might not work with TOML, and would definitely be for another PR.

src/parameter_input.hpp
Yurlungur: Done

Yurlungur:
> Will leave review shortly... but one quick question came up during a meeting:
> 
> What happens with
> 
> ```
> const bool do_that = pin->GetOrAdd("block", "do_that", true);
> const bool do_this = do_that;
> pin->GetOrAdd("block", "do_this", do_this);
> ```

That will depend on the code path the code takes. It won't default to true, it'll default to whatever `do_that` was set to in the input deck. I don't see a way around this completely, and this isn't a complete replacement for hand-written documentation.

My suggestion for cases like this is to specify in the docstring how this works.

> > > Will leave review shortly... but one quick question came up during a meeting:
> > > What happens with
> > > ```
> > > const bool do_that = pin->GetOrAdd("block", "do_that", true);
> > > const bool do_this = do_that;
> > > pin->GetOrAdd("block", "do_this", do_this);
> > > ```
> > 
> > 
> > That will depend on the code path the code takes. It won't default to true, it'll default to whatever `do_that` was set to in the input deck. I don't see a way around this completely, and this isn't a complete replacement for hand-written documentation.
> > My suggestion for cases like this is to specify in the docstring how this works.
> 
> We could overload the default argument to also take a `parameter_pointer("block", "other parameter")`, then the ordering wouldn't matter. Obviously this would require some changes under the hood, might not work with TOML, and would definitely be for another PR.

Yeah that's a good idea. I think it *would* work under the hood and with TOML. I'll make an issue. #1285

src/parthenon_manager.cpp
Yurlungur: Done.

src/parameter_input.hpp
Yurlungur: nah the sort was easy:
```C++
  template <typename T, template <class...> class Container_t, class... extra>
  void CheckAndUpdateQueries_(const std::string &block, const std::string &name,
                              const std::optional<T> &defval,
                              Container_t<T, extra...> allowed_vals,
                              const std::optional<std::string> &docstring) {
    if constexpr (is_sortable_v<decltype(allowed_vals)>) {
      if (allowed_vals.size() > 0) {
        std::sort(std::begin(allowed_vals), std::end(allowed_vals));
      }
    }
// blah...
```
where I added 
```C++
template <typename T, typename = void>
struct is_sortable : std::false_type {};

template <typename T>
struct is_sortable<T, std::void_t<decltype(std::sort(std::begin(std::declval<T &>()),
                                                     std::end(std::declval<T &>())))>>
    : std::true_type {};

template <typename T>
constexpr bool is_sortable_v = is_sortable<T>::value;
```
to a utility header.

Yurlungur: Note now the collection of allowed values gets copied in by value. But by the same argument as the O(N^2) search, I don't think copying by value is so bad here.

And hopefully non-sortable collections are automatically sorted somehow... e.g., sets.

Yurlungur:
> > Will leave review shortly... but one quick question came up during a meeting:
> > What happens with
> > ```
> > const bool do_that = pin->GetOrAdd("block", "do_that", true);
> > const bool do_this = do_that;
> > pin->GetOrAdd("block", "do_this", do_this);
> > ```
> 
> That will depend on the code path the code takes. It won't default to true, it'll default to whatever `do_that` was set to in the input deck. I don't see a way around this completely, and this isn't a complete replacement for hand-written documentation.
> 
> My suggestion for cases like this is to specify in the docstring how this works.
> 
> > > > Will leave review shortly... but one quick question came up during a meeting:
> > > > What happens with
> > > > ```
> > > > const bool do_that = pin->GetOrAdd("block", "do_that", true);
> > > > const bool do_this = do_that;
> > > > pin->GetOrAdd("block", "do_this", do_this);
> > > > ```
> > > 
> > > 
> > > That will depend on the code path the code takes. It won't default to true, it'll default to whatever `do_that` was set to in the input deck. I don't see a way around this completely, and this isn't a complete replacement for hand-written documentation.
> > > My suggestion for cases like this is to specify in the docstring how this works.
> > 
> > 
> > We could overload the default argument to also take a `parameter_pointer("block", "other parameter")`, then the ordering wouldn't matter. Obviously this would require some changes under the hood, might not work with TOML, and would definitely be for another PR.
> 
> Yeah that's a good idea. I think it _would_ work under the hood and with TOML. I'll make an issue. #1285

This turned out to be really easy. #1286

doc/sphinx/_static/custom.css
pdmullen: Can you leave this tidbit as a comment in the code?

doc/sphinx/src/0_input_parameters_ref.rst
```diff
@@ -0,0 +1,48 @@
...
+===========================
+
+This reference is automatically generated with Parthenon's
+``ParameterInput`` class. To generate it, pass the ``-p`` flag into a
+parthenon-based executable (in addition to the other flags you would
+normally pass), and optionally a ``regex`` to specify which blocks
+you'd like to output. Parthenon will print a valid CSV file to the
+terminal.
+
+Every call to ``ParameterInput::Get*`` optionally takes a "docustring"
```
pdmullen: docustring --> docstring?

pdmullen (approved):
Nice!

Yurlungur:
Sounds like folks are happy with this, so I'm going to attempt to merge it, especially since #1277 depends on it.

doc/sphinx/src/0_input_parameters_ref.rst
Yurlungur: done

doc/sphinx/_static/custom.css
Yurlungur: done

PR #1292, opened 2025-07-15
Fix some more sanitizer issues

lroberts36:
## PR Summary

Remove a `shared_ptr` cycle that was causing address sanitizer issues and attempt a fix of the `MPI_Comm` destructor in `ReductionBase`.

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build


Yurlungur (approved):

c-prather (approved):
LGTM.  Freeing the MPI comm in the reductions destructor has never worked very well, if this reliably passes the tests it does better than the original code.

src/utils/hash.hpp
```diff
@@ -82,9 +83,26 @@ struct std::hash<std::tuple<Ts...>> {
...
+  }
+};
+
+template <class T>
+struct WeakPtrEqual {
+  bool operator()(const std::weak_ptr<T> &lhs, const std::weak_ptr<T> &rhs) const {
+    return !lhs.owner_before(rhs) && !rhs.owner_before(lhs);
+  }
+};
+
```
pgrete, line 103: Is this executed somewhere?

src/utils/reductions.hpp
```diff
@@ -68,7 +68,11 @@ struct ReductionBase {
     // Store the communicator in a shared_ptr, so that
     // MPI_Comm_free is called, but only when the last
     // copy of this ReductionBase is destroyed.
-    pcomm = std::shared_ptr<MPI_Comm>(new MPI_Comm, MPI_Comm_disconnect);
+    pcomm = std::shared_ptr<MPI_Comm>(new MPI_Comm, [](MPI_Comm *comm) {
+      auto *temp = comm;
+      PARTHENON_MPI_CHECK(MPI_Comm_disconnect(comm));
+      delete temp;
```
pgrete, line 74: What's the rationale behind this logic?
Is is about making sure that the disconnect is successful?
We could also check if `comm == MPI_COMM_NULL` following the call.

pgrete (approved):

src/utils/reductions.hpp
lroberts36: The rational is that the constructor of the shared pointer call `new MPI_Comm`, but the user defined destructor never called `delete` on that object (`MPI_Comm_disconnect` doesn't call `delete` I think). The sanitizer was complaining about this and changing to this logic fixed the sanitizer issue.

pgrete:
Given that this touches active code it should probably get a changelog doesn't it?

PR #1296, opened 2025-07-28
[WIP] Add mutability maybe restart and made it so we can read old restart files

### LINKED ISSUE #1293: Backwards compatibility for `next_time` in old restarts
```
      if (pin->DoesParameterExist(block_name, "next_time")) {
        std::stringstream msg;
        msg << "You have used the next_time parameter in the " << block_name
            << " output block. This parameter is deprecated. Instead change"
            << " the output cadence with dt." << std::endl;
        PARTHENON_THROW(msg);
      }
      if (pin->DoesParameterExist(block_name, "next_n")) {
        std::stringstream msg;
        msg << "You have used the next_n parameter in the " << block_name
            << " output block. This parameter is deprecated. Instead change"
            << " the output cadence with dn." << std::endl;
        PARTHENON_THROW(msg);
```

The change introduced in https://github.com/parthenon-hpc-lab/parthenon/pull/1266 results in hard fails for existing data on disk.

We probably want to keep this a `throw` for new simulations but for restarts I think we should maybe just warn (once) and then internally convert to the new structure.
Am I missing sth @Yurlungur ?

PS: With regard to the "convert", do we have an option to remove/delete a line from `pinput` so that it doesn't show up again for subsequent restarts?

Yurlungur:
## PR Summary

This resolves #1293 . As @pgrete reported, #1266 breaks restarts from old runs, where `next_n` or `next_time` were present in the input file. To resolve, I do three things:
1. I add `Mutability::MaybeRestart` which doesn't error out if a param that should be restarted from isn't present. Instead a warning is printed and the code continues. (I do this by changing the error handling in HDF5 to be a bit more exception-ey and then using a try catch block. One could instead check for existence and then read if it exists. This seemed easier. I could be convinced to drop try-catch though.)
2. I modify the outputs logic to warn if `next_n` or `next_time` are present (upon restart, it still errors out if it's a new simulation) and compute the new values for `last_n` and `last_time` from them if they exist. To make it so this error doesn't keep happening, I also delete these from the parameter input object so that on new restart the new format is present, not the old one. Note (sorry @bprather ) this touches the inputs again so it needs to be fused with TOML, again. It's a pretty minor fix though. The TOML version will be much simpler.

This is WIP because I think these changes are kind of gross and I want feedback. But this is basically ready for review, and I can update the changelog and merge it if people like it.

One more outstanding question regarding this machinery: Should the output version be bumped? That would require an upgold. 

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build

### DIFF
--- a/src/interface/params.cpp
+++ b/src/interface/params.cpp
@@ -11,10 +11,12 @@
 // the public, perform publicly and display publicly, and to permit others to do so.
 //========================================================================================
 
+#include <sstream>
 #include <string>
 
 #include "utils/error_checking.hpp"
 
+#include "globals.hpp"
 #include "kokkos_abstraction.hpp"
 #include "parthenon_arrays.hpp"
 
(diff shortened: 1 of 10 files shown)


pgrete (commented):
Thanks for the quick fix.
As mentioned in https://github.com/parthenon-hpc-lab/parthenon/issues/1293#issuecomment-3131205728 I'm in favor of not introducing a new flag.
Otherwise, the fix looks good and removing an entry from `pinput` is also very welcome (we carry around a couple of legacy parameters that we can now purge from files).

pgrete:
> One more outstanding question regarding this machinery: Should the output version be bumped? That would require an upgold.

I'm leaning towards "no" as we didn't change the internal structure of the hdf5 file and the overhead involved may not be worth the (expected few) incompatibilities with postprocessing tools (that digest `next_t`)

pgrete:
FYI I'm fixing more stuff (against this branch) here #1297

Yurlungur:
Pre-empted by #1297

PR #1294, opened 2025-07-25
Optionally allow `Metadata::Restart` fields to be absent on disk

pgrete:
## PR Summary

We (@mfournier01) recently came across an issue where a field was could selectively enabled in a simulation upon restart.
Once enabled the field information is required for subsequent restarts (thus adding a `Metadata::Restart` flag).
However, this resulted in an attempt to read the data when first enabled, which naturally fails.
This is a quick and dirty attempt (not up to date with current `develop` to solicit feedback for a clean solution).
Thus, comments would be welcome.
I could, for example, also imagine to not introduce a new metadata flag and just warn the user if the variable is not present (to keep the logic simple).

## PR Checklist

- [ ] Code passes cpplint
- [ ] New features are documented.
- [ ] Adds a test for any bugs fixed. Adds tests for new features.
- [ ] Code is formatted
- [ ] Changes are summarized in CHANGELOG.md
- [ ] Change is breaking (API, behavior, ...)
  - [ ] Change is *additionally* added to CHANGELOG.md in the breaking section
  - [ ] PR is marked as breaking
  - [ ] Short summary API changes at the top of the PR (plus optionally with an automated update/fix script)
- [ ] CI has been triggered on [Darwin](https://re-git.lanl.gov/eap-oss/parthenon/-/pipelines) for performance regression tests.
- [ ] Docs build
- [ ] (@lanl.gov employees) Update copyright on changed files

PR #1297, opened 2025-07-29
Backward compatibility fixes (`last_/next_` output package, restart with new `Restart` vars, ,`is_restart`) 

### LINKED ISSUE #1235: Cleanup: Two independent checks for `is_restart` exist
I just noticed that theres a `pmesh->is_restart` and `Globals::is_restart`, which are set at different places (with different logic).
I think we should decide to only have one (so that future changes/logic can rely on a single code path).

### LINKED ISSUE #1293: Backwards compatibility for `next_time` in old restarts

### LINKED ISSUE #1298: Provenance handling during package init and (mesh) defaults
In #1286 following safety check was introduced.
```
            // JMM: Forbid setting a default value after requesting but
            // allow requesting without a default if a default has
            // already been set.  I know this is unpleasantly stateful,
            // but we do this in a few places in the code.
            std::stringstream msg;
            msg << "Input parameter " << block << "/" << name
                << " called previously without a default value and now called with one."
                << " If a default value is used, the first call must always set one."
                << std::endl;
            PARTHENON_THROW(msg);
```

This now lead to issues in AthenaPK because in some Package Init (i.e., before the Mesh is being constructed and setting `GetOrAddString("parthenon/mesh","refinement","none") calls 
```
  PARTHENON_REQUIRE_THROWS(
      pin->GetString("parthenon/mesh", "refinement") != "adaptive",
      "Tracers/swarms currently only supported on non-adaptive meshes.");
```
which triggers the failsafe above (both for fresh sims as well for restarts).

What's the best use case/proper fix here (from a big picture point of view)?
The easy one would be to just add a default `none` in the package, too.
Is that desirable? Should downstream codes handle some "common/shared" (mostly mesh related) input parameters in a more central way?

ping @Yurlungur @bprather

pgrete:
## PR Summary

Fixes
- allow override of default (in code) input parameters when parameters are set from the command line
- allows restarts with new `Metadata::Restart` variables/fields/params.  Will add manual fixing machinery in separate PR, see issue #1300
- allow restarts with missing output package (transition from `next_xxx` input to `last_xxx` output package)
- allows to enable a custom `id` field for swarms
- potential future `last_xxx` entries (when output was soft disabled with negative `dt` or `dn`) -- edge case
- remove `pmesh->is_restart` in favor of `Globals::is_restart`
- make `Globals::is_restart` work properly

Adds
- Interface to remove variables from `pinput`

### "API" breaking
- `is_restart` `Mesh` member has been removed. Please use `Globals::is_restart` now.

Fixes #1235 #1293 #1298

Supersedes #1294 #1296

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/src/globals.cpp
+++ b/src/globals.cpp
@@ -1,6 +1,6 @@
 //========================================================================================
-// Athena++ astrophysical MHD code
-// Copyright(C) 2014 James M. Stone <jmstone@princeton.edu> and other code contributors
+// Parthenon performance portable AMR framework
+// Copyright(C) 2020-2025 The Parthenon collaboration
 // Licensed under the 3-clause BSD License, see LICENSE file for details
 //========================================================================================
 // (C) (or copyright) 2020-2025. Triad National Security, LLC. All rights reserved.
@@ -32,8 +32,9 @@ namespace Globals {
 int nghost;
 
 // all of these global variables are set at the start of main():
-int my_rank; // MPI rank of this process
-int nranks;  // total number of MPI ranks
+int my_rank;     // MPI rank of this process
+int nranks;      // total number of MPI ranks
+bool is_restart; // Whether this simulation is restarted from a checkpoint file
 
 // sparse configuration values that are needed in various places
 SparseConfig sparse_config;
--- a/src/globals.cpp
+++ b/src/globals.cpp
@@ -32,9 +32,8 @@ namespace Globals {
 int nghost;
 
 // all of these global variables are set at the start of main():
-int my_rank;     // MPI rank of this process
-int nranks;      // total number of MPI ranks
-bool is_restart; // Whether this simulation is restarted from a checkpoint file
+int my_rank; // MPI rank of this process
+int nranks;  // total number of MPI ranks
 
 // sparse configuration values that are needed in various places
 SparseConfig sparse_config;
--- a/src/globals.hpp
+++ b/src/globals.hpp
@@ -38,7 +38,6 @@ struct SparseConfig {
 };
 
 extern int my_rank, nranks, nghost;
-extern bool is_restart;
 
 extern SparseConfig sparse_config;
 
--- a/src/globals.hpp
+++ b/src/globals.hpp
@@ -1,6 +1,6 @@
 //========================================================================================
-// Athena++ astrophysical MHD code
-// Copyright(C) 2014 James M. Stone <jmstone@princeton.edu> and other code contributors
+// Parthenon performance portable AMR framework
+// Copyright(C) 2020-2025 The Parthenon collaboration
 // Licensed under the 3-clause BSD License, see LICENSE file for details
 //========================================================================================
 // (C) (or copyright) 2020-2025. Triad National Security, LLC. All rights reserved.
@@ -38,6 +38,7 @@ struct SparseConfig {
 };
 
 extern int my_rank, nranks, nghost;
+extern bool is_restart;
 
 extern SparseConfig sparse_config;
 
--- a/src/interface/params.cpp
+++ b/src/interface/params.cpp
@@ -58,25 +58,19 @@ void Params::ReadFromHDF5AllParamsOfType(const std::string &prefix,
                                          const HDF5::H5G &group) {
   for (auto &[key, pparam] : myParams_) {
     if (std::type_index(pparam->type()) == std::type_index(typeid(T)) &&
-        IsRestartable(key)) {
-      auto mutability = myMutable_.at(key);
+        GetMutability(key) == Mutability::Restart) {
       auto typed_ptr = std::any_cast<T>(pparam.get());
       auto &val = *typed_ptr;
       std::string fullpath = prefix + "/" + key;
-      if (mutability == Mutability::Restart) {
+      try {
         HDF5::HDF5ReadAttribute(group, fullpath, val);
         Update(key, val);
-      } else if (mutability == Mutability::MaybeRestart) {
-        try {
-          HDF5::HDF5ReadAttribute(group, fullpath, val);
-          Update(key, val);
-        } catch (std::runtime_error e) {
-          if (Globals::my_rank == 0) {
-            std::stringstream ss;
-            ss << "Failed to load parameter " << fullpath
-               << " from the restart file! Using default value." << std::endl;
-            PARTHENON_WARN(ss);
-          }
+      } catch (std::runtime_error e) {
+        if (Globals::my_rank == 0) {
+          std::stringstream ss;
+          ss << "Failed to load parameter " << fullpath
+             << " from the restart file! Using default value." << std::endl;
+          PARTHENON_WARN(ss);
         }
       }
     }
--- a/src/interface/params.cpp
+++ b/src/interface/params.cpp
@@ -66,6 +66,7 @@ void Params::ReadFromHDF5AllParamsOfType(const std::string &prefix,
         HDF5::HDF5ReadAttribute(group, fullpath, val);
         Update(key, val);
       } catch (std::runtime_error e) {
+        // TODO(JMM/PG) Add failed load list of "fail/needs fix" list
         if (Globals::my_rank == 0) {
           std::stringstream ss;
           ss << "Failed to load parameter " << fullpath
--- a/src/interface/params.cpp
+++ b/src/interface/params.cpp
@@ -11,10 +11,12 @@
 // the public, perform publicly and display publicly, and to permit others to do so.
 //========================================================================================
 
+#include <sstream>
 #include <string>
 
 #include "utils/error_checking.hpp"
 
+#include "globals.hpp"
 #include "kokkos_abstraction.hpp"
 #include "parthenon_arrays.hpp"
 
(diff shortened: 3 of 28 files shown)


src/interface/params.cpp
```diff
@@ -58,25 +58,19 @@ void Params::ReadFromHDF5AllParamsOfType(const std::string &prefix,
...
   for (auto &[key, pparam] : myParams_) {
     if (std::type_index(pparam->type()) == std::type_index(typeid(T)) &&
-        IsRestartable(key)) {
-      auto mutability = myMutable_.at(key);
+        GetMutability(key) == Mutability::Restart) {
       auto typed_ptr = std::any_cast<T>(pparam.get());
       auto &val = *typed_ptr;
       std::string fullpath = prefix + "/" + key;
-      if (mutability == Mutability::Restart) {
+      try {
```
Yurlungur: I think we should discuss this particular item a bit more.... I am concerned that loading from a restart with incomplete data without erroring out will cause a lot of hard to find bugs down the line. It seems like it would be really easy for someone to trigger this without thinking too hard about it.

Yurlungur (commented):
Thanks for chasing this down! Why did you choose to go with `pmesh->is_restart` rather than `Globals`? To my mind actually globals makes more sense. I'd prefer to move things *out* of mesh, which has kind of turned into a fortran common block.

src/globals.cpp
```diff
@@ -1,6 +1,6 @@
 //========================================================================================
-// Athena++ astrophysical MHD code
-// Copyright(C) 2014 James M. Stone <jmstone@princeton.edu> and other code contributors
+// Parthenon performance portable AMR framework
+// Copyright(C) 2020-2025 The Parthenon collaboration
```
Yurlungur, line 3: Do we not need both of these? I think we can't get rid of the original CC?

src/globals.hpp
```diff
@@ -1,6 +1,6 @@
 //========================================================================================
-// Athena++ astrophysical MHD code
-// Copyright(C) 2014 James M. Stone <jmstone@princeton.edu> and other code contributors
+// Parthenon performance portable AMR framework
+// Copyright(C) 2020-2025 The Parthenon collaboration
```
Yurlungur, line 3: I think we need both of these right?

src/parameter_input.cpp
```diff
@@ -934,6 +934,28 @@ std::string ParameterInput::SetString(const std::string &block, const std::strin
   return value;
 }
 
+void ParameterInput::RemoveParameter(const std::string &block, const std::string &name) {
```
Yurlungur, line 937: Sorry @bprather . I think this gets a lot of simpler with TOML though.

Yurlungur (approved):
Given the discussion on Matrix and in issues, I'm happy to approve now. Thanks for cleaning all this up!

Yurlungur:
I pushed the following new logic
```C++
void Driver::PreExecute() {
  std::string check_orphans = pinput->GetOrAddString(
      "parthenon/job", "check_orphans", "initially",
      std::vector<std::string>{"always", "initially", "never"},
      "Print a warning if any parameters are in the input deck but not used in the code. "
      "By default this check is performed for new runs, but can also be enabled for "
      "restarts or completely disabled.");
  // Output a text file of all parameters at this point
  // Optionally also dump to console
  DumpInputParameters();

  if (Globals::my_rank == 0) {
    if ((check_orphans == "always") ||
        (!Globals::is_restart && (check_orphans == "initially"))) {
      pinput->CheckOrphans();
    }
    std::cout << "# Variables in use:\n" << *(pmesh->resolved_packages) << std::endl;
    std::cout << std::endl;
    std::cout << "Setup complete, executing driver...\n" << std::endl;
  }

  timer_main.reset();
}
```

src/globals.cpp
pgrete: Good question. I looked at the file and there was effectively nothing from the original left (apart from maybe the filename). I'm happy to revert in case want/need to be safe than sorry.

pgrete:
Uh oh
```
 	 44 - Test deleting parameters from ParameterInput (SEGFAULT)
```
This seem legit.

pgrete:
I think I fixed it. There was an issue in case the initial pointer in the list was deleted.

pgrete:
I have no clue why it's failing now.
I can reproduce the failed test locally (with MPI):
```
$ mpirun -np 2 /home/pgrete/src/parthenon/build-vscode/example/diffusion/diffusion-example -i /home/pgrete/src/parthenon/tst/regression/test_suites/diffusion/parthinput.diffusion
...
cycle=262 time=1.9989013671875000e-02 dt=1.0986328125000416e-05 zone-cycles/wsec_step=2.34e+04 wsec_total=5.45e+00 wsec_step=2.19e-02
cycle=263 time=2.0000000000000000e-02 dt=2.1972656250000833e-05 zone-cycles/wsec_step=2.07e+04 wsec_total=5.48e+00 wsec_step=2.48e-02

Driver completed.
time=2.00e-02 cycle=263
tlim=2.00e-02 nlim=-1

walltime used = 5.48e+00
zone-cycles/wallsecond = 2.46e+04
[13:22:15][pgrete@haerke: ~/src/parthenon/build-vscode]
$ echo $?
1
```
Any idea?

src/globals.cpp
Yurlungur: I think we should actually have both rather than revert. So add the second but keep the first.

Yurlungur:
> I have no clue why it's failing now. I can reproduce the failed test locally (with MPI):
> 
> ```
> $ mpirun -np 2 /home/pgrete/src/parthenon/build-vscode/example/diffusion/diffusion-example -i /home/pgrete/src/parthenon/tst/regression/test_suites/diffusion/parthinput.diffusion
> ...
> cycle=262 time=1.9989013671875000e-02 dt=1.0986328125000416e-05 zone-cycles/wsec_step=2.34e+04 wsec_total=5.45e+00 wsec_step=2.19e-02
> cycle=263 time=2.0000000000000000e-02 dt=2.1972656250000833e-05 zone-cycles/wsec_step=2.07e+04 wsec_total=5.48e+00 wsec_step=2.48e-02
> 
> Driver completed.
> time=2.00e-02 cycle=263
> tlim=2.00e-02 nlim=-1
> 
> walltime used = 5.48e+00
> zone-cycles/wallsecond = 2.46e+04
> [13:22:15][pgrete@haerke: ~/src/parthenon/build-vscode]
> $ echo $?
> 1
> ```
> 
> Any idea?

That doesn't look like a failure to me? Is it just returning the wrong exit code?

acreyes:
> > I have no clue why it's failing now. I can reproduce the failed test locally (with MPI):
> > ```
> > $ mpirun -np 2 /home/pgrete/src/parthenon/build-vscode/example/diffusion/diffusion-example -i /home/pgrete/src/parthenon/tst/regression/test_suites/diffusion/parthinput.diffusion
> > ...
> > cycle=262 time=1.9989013671875000e-02 dt=1.0986328125000416e-05 zone-cycles/wsec_step=2.34e+04 wsec_total=5.45e+00 wsec_step=2.19e-02
> > cycle=263 time=2.0000000000000000e-02 dt=2.1972656250000833e-05 zone-cycles/wsec_step=2.07e+04 wsec_total=5.48e+00 wsec_step=2.48e-02
> > 
> > Driver completed.
> > time=2.00e-02 cycle=263
> > tlim=2.00e-02 nlim=-1
> > 
> > walltime used = 5.48e+00
> > zone-cycles/wallsecond = 2.46e+04
> > [13:22:15][pgrete@haerke: ~/src/parthenon/build-vscode]
> > $ echo $?
> > 1
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Any idea?
> 
> That doesn't look like a failure to me? Is it just returning the wrong exit code?

It looks like these `DiffusionDriver::final_rms*` [members](https://github.com/parthenon-hpc-lab/parthenon/blob/5203c721ab7642ed2fe7d4fafa9bb249aee80c15/example/diffusion/diffusion_driver.hpp#L40) are uninitialized and not being set after the driver executes. Unlike in the [poisson example](https://github.com/parthenon-hpc-lab/parthenon/blob/develop/example/diffusion/main.cpp#L52)

Yurlungur:
Good catch. That's probably a bug from when this code was written. Zero-initializing them should solve the issue yeah?

Also, we should probably add some memory sanitizer logic to catch this kind of thing...

acreyes:
> Good catch. That's probably a bug from when this code was written. Zero-initializing them should solve the issue yeah?
> 
> Also, we should probably add some memory sanitizer logic to catch this kind of thing...

yeah, although can we just get rid of the checks? Are the residuals not being checked inside the solvers?

lroberts36:
> yeah, although can we just get rid of the checks? Are the residuals not being checked inside the solvers?

That is just copy pasted from the `poisson_gmg` example, where the driver copies the value of the residuals from the solvers to the driver. The check is there in main.cpp since the solver could exit based on hitting the maximum number of iterations, which would signal that the test was failing, and there would be a non-zero return from main.

I think in `diffusion/diffusion_driver.hpp` those members can be removed, while in `poisson_gmg/poisson_driver.hpp` they should be zero initialized.

lroberts36 (approved):

pgrete:
Thanks for the fix @acreyes

PR #1305, opened 2025-08-07
compute asymmetry tweaks

Yurlungur:
## PR Summary

These are minor bug fixes and code tweaks while I continued to chase down asymmetry in riot. The main things I changed, which I think are open to discussion:
1. I added the ability to register fluxes as `ForceRemeshComm` and have them safely moved during load balancing and remeshing. Though this is not the default.
2. I added the ability to request the code write outputs *before* remeshing instead of *after.* This is mostly a debugging feature. It's *not* desirable if you are doing restarts, because then the restart is a different mesh configuration than you expect.
3. I significantly modified my `compute_asymmetry` and `report_asymmetry` scripts to be much faster and more robust.
4. I added several quality of life improvements to `movie2d.py`, mainly bounds on the x, y, and z axes

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build

### DIFF
--- a/doc/sphinx/src/generated/diffusion-parth-table.csv
+++ b/doc/sphinx/src/generated/diffusion-parth-table.csv
@@ -75,7 +75,7 @@ block,parameters,type,default,description
 "parthenon/time","ncycle_out","int","1","cadence of outputs to stdout"
 "parthenon/time","ncycle_out_mesh","int","0","cadence of outputs describing mesh"
 "parthenon/time","nlim","int","-1","maximum number of iterations, only limiting if non-negative"
-"parthenon/time","output_before_amr","bool","0","Set to true to generate outputs in a step BEFORE modifying the mesh at the end of the step. By default outputs happen AFTER remeshing if remeshing happens."
+"parthenon/time","output_before_amr","bool","0","Set to true to generate outputs in a step BEFORE modifying the mesh at the end of the step. By default outputs happen AFTER remeshing if remeshing happens. WARNING: this will make restarts not bitwise-exact."
 "parthenon/time","perf_cycle_offset","int","0","don't measure performance for some number of initial cycles"
 "parthenon/time","recv_bdry_buf_timeout_sec","Real","-1",""
 "parthenon/time","start_time","Real","0","physical time at which to start the simulation"
--- a/doc/sphinx/src/generated/diffusion-parth-table.csv
+++ b/doc/sphinx/src/generated/diffusion-parth-table.csv
@@ -75,6 +75,7 @@ block,parameters,type,default,description
 "parthenon/time","ncycle_out","int","1","cadence of outputs to stdout"
 "parthenon/time","ncycle_out_mesh","int","0","cadence of outputs describing mesh"
 "parthenon/time","nlim","int","-1","maximum number of iterations, only limiting if non-negative"
+"parthenon/time","output_before_amr","bool","0","Set to true to generate outputs in a step BEFORE modifying the mesh at the end of the step. By default outputs happen AFTER remeshing if remeshing happens."
 "parthenon/time","perf_cycle_offset","int","0","don't measure performance for some number of initial cycles"
 "parthenon/time","recv_bdry_buf_timeout_sec","Real","-1",""
 "parthenon/time","start_time","Real","0","physical time at which to start the simulation"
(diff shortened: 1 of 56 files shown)


doc/sphinx/src/generated/diffusion-parth-table.csv
```diff
@@ -75,6 +75,7 @@ block,parameters,type,default,description
 "parthenon/time","ncycle_out","int","1","cadence of outputs to stdout"
 "parthenon/time","ncycle_out_mesh","int","0","cadence of outputs describing mesh"
 "parthenon/time","nlim","int","-1","maximum number of iterations, only limiting if non-negative"
+"parthenon/time","output_before_amr","bool","0","Set to true to generate outputs in a step BEFORE modifying the mesh at the end of the step. By default outputs happen AFTER remeshing if remeshing happens."
```
lroberts36: Not sure if we want to limit the length of docstrings, but if we are willing to go longer it might be useful to mention here that setting this to true will make restarts not bitwise exact (since I think the code will take one extra step with the previous refinement structure before refining, right?)

src/parthenon_manager.cpp
```diff
@@ -325,9 +325,8 @@ void ParthenonManager::RestartPackages(Mesh &rm, RestartReader &resfile) {
 
   // make sure we have all sparse variables that are in the restart file
   PARTHENON_REQUIRE_THROWS(
-      num_sparse == sparse_info.num_sparse,
+      num_sparse <= sparse_info.num_sparse,
```
lroberts36, line 334: Why this change?

lroberts36 (approved):

src/parthenon_manager.cpp
Yurlungur: This was required to restart from core dump file as opposed to rst file, because the core dump file contains EVERYTHING so there might be more variables.

I could put it back if you feel strongly. But I found this to be a useful capability for debugging.

doc/sphinx/src/generated/diffusion-parth-table.csv
Yurlungur: yeah that's right. I can add that comment.

src/parthenon_manager.cpp
lroberts36: I think I would just be satisfied with a comment in the code.

Yurlungur: done

Yurlungur:
@pgrete @pdmullen can I get one more review?

pgrete:
@pdmullen IIRC you (were) volunteered given that you're more familiar with the infrastructure. If you're currently unavailable I could review tomorrow.

scripts/python/packages/parthenon_tools/parthenon_tools/compute_asymmetry.py
```diff
@@ -18,28 +18,69 @@
...
-        if np.any(ylocs[b] >= 0):
-            for d in range(var_diff.shape[1]):
-                sign = -1 if (var.shape[1] == 3) and d == 1 else 1
-                var_diff[b, d] = var[b, d] - sign * np.flip(var[bb, d], axis=-2)
-                var_diff[bb, d] = var[bb, d] - sign * np.flip(var[b, d], axis=-2)
+
+    # indices for blocks with y > 0 and their partners with y < 0
+    mask = np.any(ylocs >= 0, axis=1)
+    blk_indices = np.nonzero(mask)[0]
+    partner_indices = matches[blk_indices]
```
pdmullen, line 51: It might be worth reporting to user the failure mode where the grid structure itself is not symmetric across the symmetry plane, but maybe I missed that somewhere else?

pdmullen (approved):

scripts/python/packages/parthenon_tools/parthenon_tools/compute_asymmetry.py
Yurlungur: Good suggestion! But I'm actually not sure how to check for this, because you could have any number of blocks bisected at the midplane (i.e., if the base mesh has an odd number of blocks in the X2 direction), so it's not as simple as checking block count... I think this will somewhat be automatically captured because if the mesh structure is different, the asymmetry that gets computed will be outrageously large...

pdmullen: :+1:

Yurlungur: Maybe I'm missing something though, this certainly *feels* like something one could check for.

PR #1307, opened 2025-08-14
Fix imports in parthenon_tools + phdf_diff logic

acreyes:
## PR Summary

When installing `parthenon_tools` I found I needed to have the directory in my `PYTHONPATH` or none of the internal import statements would work, getting things like
```
python -m parthenon_tools.phdf_diff build/tst/regression/outputs/restart/gold.out0.00002.rhdf build/tst/regression/outputs/restart/silver.out0.00002.rhdf
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/Users/adamreyes/Documents/research/repos/github/parthenon-hpc/parthenon/scripts/python/packages/parthenon_tools/parthenon_tools/phdf_diff.py", line 604, in <module>
    ret = compare(
        files,
    ...<6 lines>...
        relative,
    )
  File "/Users/adamreyes/Documents/research/repos/github/parthenon-hpc/parthenon/scripts/python/packages/parthenon_tools/parthenon_tools/phdf_diff.py", line 392, in compare
    from phdf import phdf
ModuleNotFoundError: No module named 'phdf'
```

This PR changes the imports to absolute,
```
from .phdf import phdf
```

Additionally I've inverted the tolerance checks in `phdf_diff` so that NaNs in any of the data will result in a test failure.

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build

### DIFF
--- a/.github/workflows/ci-extended.yml
+++ b/.github/workflows/ci-extended.yml
@@ -44,7 +44,7 @@ jobs:
 
       - name: Install parthenon_tools
         run: |
-          pip install --user scripts/python/packages/parthenon_tools
+          pip install --user --break-system-packages scripts/python/packages/parthenon_tools
 
       - name: Setup cache for gold standard
         uses: actions/cache@v3
@@ -154,7 +154,7 @@ jobs:
 
       - name: Install parthenon_tools
         run: |
-          pip install --user scripts/python/packages/parthenon_tools
+          pip install --user --break-system-packages scripts/python/packages/parthenon_tools
 
       - name: Setup cache for gold standard
         uses: actions/cache@v3
--- a/.github/workflows/ci-extended.yml
+++ b/.github/workflows/ci-extended.yml
@@ -42,10 +42,6 @@ jobs:
         with:
           submodules: 'true'
 
-      - name: Install parthenon_tools
-        run: |
-          pip install --user --break-system-packages scripts/python/packages/parthenon_tools
-
       - name: Setup cache for gold standard
         uses: actions/cache@v3
         with:
@@ -82,8 +78,6 @@ jobs:
           export ASAN_OPTIONS=abort_on_error=1:fast_unwind_on_malloc=1
           export UBSAN_OPTIONS=print_stacktrace=0
           export LSAN_OPTIONS=detect_leaks=0
-          # make sure parthenon tools is in our python path
-          export PYTHONPATH="$PYTHONPATH:$(python -m site --user-site)"
           ctest -L performance -LE perf-reg
 
       # run regression tests
@@ -96,8 +90,6 @@ jobs:
           export ASAN_OPTIONS=abort_on_error=1:fast_unwind_on_malloc=1
           export UBSAN_OPTIONS=print_stacktrace=0
           export LSAN_OPTIONS=detect_leaks=0
-          # make sure parthenon tools is in our python path
-          export PYTHONPATH="$PYTHONPATH:$(python -m site --user-site)"
           ctest -L regression -L ${{ matrix.parallel }} -LE perf-reg --timeout 3600
 
       # Test Ascent integration (only most complex setup with MPI and on device)
@@ -152,10 +144,6 @@ jobs:
         with:
           submodules: 'true'
 
-      - name: Install parthenon_tools
-        run: |
-          pip install --user --break-system-packages scripts/python/packages/parthenon_tools
-
       - name: Setup cache for gold standard
         uses: actions/cache@v3
         with:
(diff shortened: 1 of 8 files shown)


Yurlungur (approved):
LGTM.

Yurlungur:
@acreyes is there some reason this is WIP?

Yurlungur:
I'd like to get this in as I'm about to muck about more with parthenon tools and it would be nice to base that on these fixes.

acreyes:
sorry it is only WIP because I had initially thought it was going to be easy, and just needed to build up the momentum to iterate on convincing the various CIs to install `parthenon_tools`.

It also makes me realize it might be potentially breaking for any downstream CI that is using the regression tests from parthenon without installing the tools

pgrete:
Would this new change preclude the option to manually point to the tools?
I typically have sth like 
```
        sys.path.insert("MY/PATH/TO/scripts/python/packages/parthenon_tools/parthenon_tools")

        try:
            import phdf
        except ModuleNotFoundError:
            print("Couldn't find module to load Parthenon hdf5 files.")
            return False
```
as this allows me to use the "current" version (when switching between different submodules etc).

Yurlungur:
> Would this new change preclude the option to manually point to the tools? I typically have sth like
> 
> ```
>         sys.path.insert("MY/PATH/TO/scripts/python/packages/parthenon_tools/parthenon_tools")
> 
>         try:
>             import phdf
>         except ModuleNotFoundError:
>             print("Couldn't find module to load Parthenon hdf5 files.")
>             return False
> ```
> 
> as this allows me to use the "current" version (when switching between different submodules etc).

What about doing something like this in each of our parthenon_tools modules?

```python
try:
  from file import thing
except:
  from parthenon_tools.file import thing
```
then it will try your python path first and then if that doesn't work try the module?

acreyes:
> Would this new change preclude the option to manually point to the tools? I typically have sth like
> 
> ```
>         sys.path.insert("MY/PATH/TO/scripts/python/packages/parthenon_tools/parthenon_tools")
> 
>         try:
>             import phdf
>         except ModuleNotFoundError:
>             print("Couldn't find module to load Parthenon hdf5 files.")
>             return False
> ```
> 
> as this allows me to use the "current" version (when switching between different submodules etc).

As an example you would be switching out the branch of `parthenon` in a downstream code, but want to always point to the tools in `develop` when working locally?  I have a virtual env that has `pip install -e MY/PATH/TO/scripts/python/packages/parthenon_tools`

Yurlungur:
you can also use the `--develop` flag with pip install

Yurlungur:
er wait `setup.py develop` sorry

Yurlungur:
> > Would this new change preclude the option to manually point to the tools? I typically have sth like
> > ```
> >         sys.path.insert("MY/PATH/TO/scripts/python/packages/parthenon_tools/parthenon_tools")
> > 
> >         try:
> >             import phdf
> >         except ModuleNotFoundError:
> >             print("Couldn't find module to load Parthenon hdf5 files.")
> >             return False
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > as this allows me to use the "current" version (when switching between different submodules etc).
> 
> As an example you would be switching out the branch of `parthenon` in a downstream code, but want to always point to the tools in `develop` when working locally? I have a virtual env that has `pip install -e MY/PATH/TO/scripts/python/packages/parthenon_tools`

For me a more common use case is I both want to install the tools for myself, but also let a user just runt he script directly without installing. I think to support both modes, we need the try/except thing

pgrete:
> > > Would this new change preclude the option to manually point to the tools? I typically have sth like
> > > ```
> > >         sys.path.insert("MY/PATH/TO/scripts/python/packages/parthenon_tools/parthenon_tools")
> > > 
> > >         try:
> > >             import phdf
> > >         except ModuleNotFoundError:
> > >             print("Couldn't find module to load Parthenon hdf5 files.")
> > >             return False
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > as this allows me to use the "current" version (when switching between different submodules etc).
> > 
> > 
> > As an example you would be switching out the branch of `parthenon` in a downstream code, but want to always point to the tools in `develop` when working locally? I have a virtual env that has `pip install -e MY/PATH/TO/scripts/python/packages/parthenon_tools`
> 
> For me a more common use case is I both want to install the tools for myself, but also let a user just runt he script directly without installing. I think to support both modes, we need the try/except thing

Yes, that was indirectly the motivation for my question above.
I think we should still support both versions.
How about editing the CI tests so that there's one case (e.g., the short tests) where we install and another case (the extended ones) where we directly import?

Yurlungur:
I went ahead and pushed that change, but if you don't like it go ahead and revert @acreyes

acreyes:
> I went ahead and pushed that change, but if you don't like it go ahead and revert @acreyes

I don't like it, but I don't see a better way to support what you want without users installing the python package.

pgrete (approved):
Tested downstream and works!

PR #1310, opened 2025-08-18
Fix logic causing issues for restarts with varying output blocks

### LINKED ISSUE #1309: Questions on output params logic
@Yurlungur I'm trying to track down some weird behavior for restart with the updated output logic.

I run a simulation, stop, and restart with an additional output block.
The new output block looks like:
```
<parthenon/output3>
file_type = openpmd
id = half
dt = 0.1 
last_time = 17.00
coarsening_factor = 2
backend_config=/e/project1/jureap19/turb_test/openpmd.toml
single_precision_output = true
```
and I restart from
```
# Swarm	Value	metadata

cycle=33510 time=1.7000447028474159e+01 dt=5.2336490814341909e-04 zone-cycles/wsec_step=0.00e+00 wsec_total=2.60e-01 wsec_step=7.65e+00 zone-cycles/wsec=0.00e+00 wsec_AMR=0.00e+00
```

Now the new output block is being written every cycle...
Any clue why?
I also noticed that in the outputs the Output `Params` look weird:
```
  int32_t   /data/35/Params~Outputs~block_numbers                                 attr   = {1, 2, 3}
  int32_t   /data/35/Params~Outputs~file_numbers                                  attr   = {343, 35}
  int32_t   /data/35/Params~Outputs~last_ns                                       attr   = {0, 0}
  double    /data/35/Params~Outputs~last_times                                    attr   = {17.1, 17}
```
So block numbers got an additional entry (`output1` is `hst`), but the other vector only contain two entries.

Looking at the code, I also noticed that block numbers are not `Mutability::Restart` whereas the other ones are.
Is it possible that the info is being overridden rather than processed (from an updated `pinput`)?

More generally, is there a potential danger that the order of the entries in the vector are being overridden/associated incorrectly on restart when input blocks are dynamically enabled/disabled/added/removed (because there's an early `continue` when the output is soft disabled but results are being pushed to `Mutability::Restart` vectors further down following the `continue`).

Finally, note to myself: setup a complex downstream regression tests that encompasses a typical workflow.

pgrete:
## PR Summary

Fixes #1309 by replacing the vector with unique params.

Note that the current tests are failing because the variables in the `bronze` (dynamically added) and `silver` output differ.
The former contains the `advection.phi_fine_restricted` variable as the it's a regex match for `advection.phi` triggered in the logic to add vars by name, whereas it's absent in `silver` because the silver output is a `restart` output, which currently ignores the `variables` parameter altogether.
I always wanted to support adding additional variables to restart output.
Any objections to this?

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build

### DIFF
--- a/src/outputs/outputs.cpp
+++ b/src/outputs/outputs.cpp
@@ -103,20 +103,26 @@ Outputs::Outputs(Mesh *pm, ParameterInput *pin, SimTime *tm) {
 
   // loop over "parthenon/output" blocks located in params
   auto pkg = pm->packages.Get("Outputs");
-  const auto &block_names = pkg->Param<std::vector<std::string>>("block_names");
-  const auto &block_numbers = pkg->Param<std::vector<int>>("block_numbers");
-  auto *pactive = pkg->MutableParam<std::vector<bool>>("active");
-  auto *pfile_numbers = pkg->MutableParam<std::vector<int>>("file_numbers");
-  auto *plast_times = pkg->MutableParam<std::vector<Real>>("last_times");
-  auto *plast_ns = pkg->MutableParam<std::vector<int>>("last_ns");
-  for (int iinput = 0; iinput < block_names.size(); ++iinput) {
+  // loop over input block names.  Find those that start with "parthenon/output", read
+  // parameters, and construct a vector of output types.
+  // PG: It could be discussed if we should work based on a vector that is initialized in
+  // the outpus packages (as before), but I don't think it's bad practice to work on
+  // `pinput` again here as we're actually processing (potentially even modifying)
+  // `pinput`.
+  for (InputBlock *pib = pin->pfirst_block; pib != nullptr; pib = pib->pnext) {
+    if (pib->block_name.compare(0, 16, "parthenon/output") != 0) {
+      continue;
+    }
     std::shared_ptr<OutputType> pnew_type; // the new output we will create
     bool restart = false;                  // we track restart outputs separately so we
                                            // need this temp variable to check
     OutputParameters op;                   // define temporary OutputParameters struct
-    op.block_name = block_names[iinput];
-    op.block_number = block_numbers[iinput];
-    op.contiguous_block_index = iinput;
+    op.block_name = pib->block_name;
+    const auto outn_str = pib->block_name.substr(16); // 16 because counting starts at 0!
+    op.block_number = atoi(outn_str.c_str());
+    auto *pfile_number = pkg->MutableParam<int>(outn_str + "/file_number");
+    auto *plast_time = pkg->MutableParam<Real>(outn_str + "/last_time");
+    auto *plast_n = pkg->MutableParam<int>(outn_str + "/last_n");
 
     Real dt = 0.0; // default value == 0 means that initial data is written by default
     int dn = -1;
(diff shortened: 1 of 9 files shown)


src/outputs/outputs.cpp
```diff
@@ -103,20 +103,26 @@ Outputs::Outputs(Mesh *pm, ParameterInput *pin, SimTime *tm) {
...
-  auto *pfile_numbers = pkg->MutableParam<std::vector<int>>("file_numbers");
-  auto *plast_times = pkg->MutableParam<std::vector<Real>>("last_times");
-  auto *plast_ns = pkg->MutableParam<std::vector<int>>("last_ns");
-  for (int iinput = 0; iinput < block_names.size(); ++iinput) {
+  // loop over input block names.  Find those that start with "parthenon/output", read
+  // parameters, and construct a vector of output types.
+  // PG: It could be discussed if we should work based on a vector that is initialized in
+  // the outpus packages (as before), but I don't think it's bad practice to work on
+  // `pinput` again here as we're actually processing (potentially even modifying)
+  // `pinput`.
```
Yurlungur, line 111: Is there anything guaranteeing that the input deck upon restart is a superset of the input deck from the original run? Maybe we should check for this? 

I'm worried about the case where an output is present in the restart but not the input deck, like if a block has been deleted somehow. Is that possible? I guess not since it'll at the very least be in the original input deck stored in the restart. So maybe it's fine. 

I guess we will also implicitly check for this because `MutableParam<T>("name")` will fail

src/outputs/outputs.cpp
```diff
@@ -103,20 +103,26 @@ Outputs::Outputs(Mesh *pm, ParameterInput *pin, SimTime *tm) {
...
+  for (InputBlock *pib = pin->pfirst_block; pib != nullptr; pib = pib->pnext) {
+    if (pib->block_name.compare(0, 16, "parthenon/output") != 0) {
+      continue;
+    }
```
Yurlungur, line 115: note we will need to unify this with the TOML stuff when it goes through

src/outputs/outputs_package.cpp
```diff
@@ -86,14 +72,12 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
           PARTHENON_THROW(msg);
         }
       }
+      // It should be safe here to jsut use outn as output blocks are unique
```
Yurlungur:
```suggestion
      // It should be safe here to just use outn as output blocks are unique
```

Yurlungur (approved):
Thanks for cleaning this up and sorry for the bug introduced by the refactor.

src/outputs/outputs.cpp
pgrete: I think it's save to assume because the infrastructure you wrote to "remove" lines from `pinput` was just recently added (so practically no use case/functionality for this existed before).

pgrete:
Looks like an gold update is needed. Will do once we have a second approval.

src/outputs/outputs.cpp
Yurlungur: That does mean future us could cause this... but fair enough.

pgrete:
Should we get this merged with just one review?
Given that the changes are not backwards compatible, I'd like to fix `develop` as quickly as possible before there's more output produced with the current version.

Yurlungur:
Let's just merge it.

pgrete:
All (extended) tests pass with the new gold. Force merging now.

PR #1317, opened 2025-09-16
Make VariableExists logic in HDF5 restart reader work

pgrete:
## PR Summary

The logic inside `VariableExits` didn't properly work as an exception was thrown that was not caught.
Instead of catching the exception the return status of the HDF5 lib function is now being used.
Also the check has to be updated to `>0` as negative values are returned in the function fails (e.g., because the hierarchy could not be traversed).

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [ ] Docs build

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 
 ### Fixed (not changing behavior/API/variables/...)
 - [[PR 1317]](https://github.com/parthenon-hpc-lab/parthenon/pull/1317) Make VariableExits logic in HDF5 restart reader work
+- [[PR 1288]](https://github.com/parthenon-hpc-lab/parthenon/pull/1288) Fix restriction bug for non-cartesian coordinates and make poisson_gmg example work for curvilinear coordinates
 - [[PR 1307]](https://github.com/parthenon-hpc-lab/parthenon/pull/1307) Fix imports in parthenon tools + phdf_diff logic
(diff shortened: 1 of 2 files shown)


Yurlungur (commented):
Nice fix. You mean `VariableExists` right?

Yurlungur (approved):

lroberts36 (approved):
Kind of gross that HDF5 throws an error here, but the fix LGTM. Maybe we are running into the issue [shown in the example here](https://support.hdfgroup.org/documentation/hdf5/latest/group___h5_o.html#gab0fef18d97844c4f83d412c5a22def7b)? [Edit: I now see your comment above mentions this.]

pgrete:
> Nice fix. You mean `VariableExists` right?

Yes, fixed the typo.

pgrete:
> Kind of gross that HDF5 throws an error here, but the fix LGTM. Maybe we are running into the issue [shown in the example here](https://support.hdfgroup.org/documentation/hdf5/latest/group___h5_o.html#gab0fef18d97844c4f83d412c5a22def7b)? [Edit: I now see your comment above mentions this.]

We recently introduced the "throw" behavior in #1297 and, yes, from my point of view some of the HDF5 interfaces could be made significantly more straightforward and/or provide more functionality that would not need to be duplicated in all codes.

Yurlungur:
Adding the throw behavior is on me... Sorry. I do think it is better as far as getting a useful error message from HDF5 for what it's worth... HDF5's default error handling is awful.

PR #1319, opened 2025-09-17
Add common scratch variable utilities

acreyes:
## PR Summary

I mentioned this capability I had implemented downstream in the last sync. This is essentially how I had done it downstream, using the regex `variable_names::base_t` to alias the same `metadata::overrideable` fields with different type aliases that can each have their own shapes across different packages. 

There are a few use cases where I want to have a meshblock field, but don't care for it to persist beyond the scope of a single task and this feature frees up the possibility for that memory to be reused in another package/task.

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build


src/pack/scratch_variables.hpp
```diff
@@ -0,0 +1,176 @@
...
+  using TT = TopologicalType;
+  if (tt == TT::Face) {
+    return Metadata::Face;
+  } else if (tt == TT::Edge) {
+    return Metadata::Edge;
+  } else if (tt == TT::Node) {
+    return Metadata::Node;
+  }
+  return Metadata::Cell;
+}
```
Yurlungur: I'm kind of shocked this doesn't exist elsewhere. I would prefer to move it to a more central location.

src/pack/scratch_variables.hpp
```diff
@@ -0,0 +1,176 @@
...
+  using TT = TopologicalType;
+  if (tt == TT::Face) {
+    return "face";
+  } else if (tt == TT::Edge) {
+    return "edge";
+  } else if (tt == TT::Node) {
+    return "node";
+  }
+  return "cell";
+}
```
Yurlungur: This one too...

src/pack/scratch_variables.hpp
```diff
@@ -0,0 +1,176 @@
...
+};
+
+namespace impl {
+template <typename... Ts>
+void AddScratch(ScratchVariableList<Ts...>, StateDescriptor *pkg) {
+  using SL = ScratchVariableList<Ts...>;
+  (
+      [&] {
+        auto m = Metadata(
+            {TopologicalTypeToMetaData(SL::TT), Metadata::Derived, Metadata::Overridable},
```
Yurlungur: Nice use of `Metadata::Overridable`

src/pack/scratch_variables.hpp
Yurlungur, line 1: Should this file (and maybe others...) be added to the `parthenon/package.hpp` file?

src/pack/scratch_variables.hpp
```diff
@@ -0,0 +1,176 @@
...
+  // in debug mode each scratch variable has its own unique name
+  impl::AddScratch(SL(), pkg);
+#else
+  auto m = Metadata(
+      {TopologicalTypeToMetaData(SL::TT), Metadata::Derived, Metadata::Overridable});
+  for (const auto var : SL::GetVarNames()) {
+    pkg->AddField(var, m);
+  }
+#endif
+}
```
Yurlungur: I guess to make this opt-in it makes the most sense for this to be here, not in `StateDescriptor` but I could see adding this directly as a method of `StateDescriptor` making sense...

src/pack/scratch_variables.hpp
```diff
@@ -0,0 +1,176 @@
...
+
+  static const auto GetVarNames() {
+    std::array<std::string, n_vars> vars;
+    auto base = "scratch_" + TopologicalTypeToString(TT) + "_";
+    for (int i = 0; i < n_vars; i++) {
+      vars[i] = base + std::to_string(i);
+    }
+    return vars;
+  }
+};
```
Yurlungur: I could imagine a way of distributing this accross the code so you can automatically add to this list somehow with maybe a static shared list. Similar to the way we can add custom metadata flags. But that probably breaks the typing logic...

Yurlungur (approved):
This is pretty cool!

src/pack/scratch_variables.hpp
acreyes: the downside is that nobody will actually provide the fields, so when multiple state descriptors request scratch then you get spammed with warnings

src/pack/scratch_variables.hpp
acreyes: I can see the use of something like that. You're right that it would break the typing logic. The logic is just to hold the lower bound of the different variables. If the state descriptor were in charge instead of this type then that could be handled when new types are registered.

src/pack/scratch_variables.hpp
acreyes: done

src/pack/scratch_variables.hpp
acreyes: done

src/pack/scratch_variables.hpp
acreyes: done

src/pack/scratch_variables.hpp
acreyes: I ended up moving this to `StateDescriptor` so that the scratch variables can be registered dynamically rather than all at once

src/pack/scratch_variables.hpp
Yurlungur: Hmm. Maybe shouldn't warn there...

doc/sphinx/src/sparse_packs.rst
```diff
@@ -42,6 +42,45 @@ which inherits from ``parthenon::variable_names::base_t`` [5]_. Rather than writ
...
+17 indexable fields across the 4 types, but only 11 fields will be allocated.
+These types can then be packed and accessed as normal in type-based sparse packs. 
+An example of usage can be found in `the unit
+test <https://github.com/parthenon-hpc-lab/parthenon/blob/develop/tst/unit/test_scratch_variables.cpp>`
+
+It should not be expected that values in these fields will persist outside the local
+scope of the tasks where they are packed. For debugging purposes the cmake build can
+be configured with ``-DPARTHENON_DEBUG_SCRATCH=ON``, in which case the fields will
+be registered with unique names, ``scratch_First`` as an example from the above
+snippet.
```
lroberts36, line 81: I worry a little bit about packages that could be executed by the task list without dependencies on one another. If they use shared scratch variables, this does introduces an implicit dependence. Obviously, this is really just on the downstream code to be careful about this but it would be good to have at least a warning about possible interactions with tasking here.

src/interface/state_descriptor.hpp
```diff
@@ -232,6 +234,32 @@ class StateDescriptor {
...
+    TopologicalType tt = T::type;
+    if constexpr (!debug_scratch_variables()) {
+      // tally how many scratch variables of this topologicaltype have already been added
+      int lower_bound = 0;
+      if (num_scratch_.find(tt) != num_scratch_.end()) {
+        lower_bound = num_scratch_.at(tt);
+      }
+      num_scratch_[tt] = T::update_bounds(lower_bound);
+      auto m = Metadata(
+          {TopologicalTypeToMetaData(T::type), Metadata::Derived, Metadata::Overridable});
```
lroberts36: Why not `Metadata::OneCopy` as well?

src/pack/scratch_variables.hpp
```diff
@@ -0,0 +1,115 @@
...
+  return pattern.str();
+}
+
+template <TopologicalType TT, int... NCOMPS>
+struct ScratchVariable : public parthenon::variable_names::base_t<true, NCOMPS...> {
+  using base_t = parthenon::variable_names::base_t<true, NCOMPS...>;
+  template <typename... Ts>
+  KOKKOS_INLINE_FUNCTION ScratchVariable(Ts &&...args)
+      : base_t(std::forward<Ts>(args)...) {}
+  static constexpr TopologicalType type = TT;
```
lroberts36: I would maybe vote for `topo_type` or something instead of `type`. `T::type` makes me think it is something like 
```c++
using type = something;
```
when I read `T::type` elsewhere in the code.

doc/sphinx/src/building.rst
```diff
@@ -49,6 +49,7 @@ General list of cmake options:
 || PARTHENON\_USE\_SYSTEM\_PACKAGES         || OFF                           || Option || If ON, attempt to link to system dependencies for Kokkos and Ascent if possible. If OFF, will avoid doing so by default.                                    |
 || PARTHENON\_IMPORT\_KOKKOS                || OFF/ON                        || Option || If ON, attempt to link to an external Kokkos library. Else build from source. Default is ON if PARTHENON\_USE\_SYSTEM\_PACKAGES and OFF otherwise.          |
 || BUILD\_SHARED\_LIBS                      || OFF                           || Option || If installing Parthenon, whether to build as shared rather than static                                                                                      |
+|| PARTHENON_DEBUG_SCRATCH                  || OFF                           || Option || If ON, will register scratch variables with unique persistent fields.                                                                                       |
```
lroberts36:
```suggestion
|| PARTHENON_DEBUG_SCRATCH                  || OFF                           || Option || If ON, will register scratch variables with unique persistent fields per package.                                                                           |
```

lroberts36 (approved):

lroberts36 (approved):

doc/sphinx/src/sparse_packs.rst
acreyes: good point. I thought about the possibility of having a lock, maybe in `MeshData`, that prevents one from having more than one pack that references scratch data checked out at a time. Not sure of other consequences of having something like that.

pgrete (approved):
LGTM
Love the doc and unit tests!

Issue #1228, opened 2025-02-07
Common field registry and virtual fields


pgrete:
I was wondering if there's need to have some machinery to create virtual fields (or if this is too much of an overhead upstream and should be handled downstream) -- similar to the discussion about a common name registry we talked about during the sync yesterday.

With virtual fields, I imagine some `Metadata::Derived` field that is also `Metadata::Virtual/Unallocated` that translates to the field being specified by a callback function (e.g., to get a temperature from the conserved variables).
The main advantages over existing machinery that I see is that it saves memory.
IIRC this is also a feature that is currently being implemented (or maybe already is) in ADIOS2 so that new fields are available for direct access in post processing tools without requiring additional disk space.

Yurlungur:
:+1: I think this would be great to have. We pondered adding this to riot but never found a great non-intrusive mechanism for doing so.

PR #1322, opened 2025-09-18
Pack on virtual fields

acreyes:
## PR Summary

This PR allows for type-based packs to pack "virtual" fields that don't exist in memory, but can be evaluated from their own set of dependent variables in the pack by calling a provided `evaluate` function. 

This seems related to #1228, but doesn't provide that capability for outputs as mentioned for ADIOS2. 

Also adds `SubPacks` to provide stencil (0, 1, 2, & 3D) views into a pack centered on a meshblock and ijk index. This can simplify the need to pass around extra indices in the `evaluate` functions for virtual fields. 

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build


src/pack/pack_utils.hpp
```diff
@@ -144,6 +151,117 @@ struct any_nonautoflux : public base_t<true> {
...
+} // namespace impl
+
+template <typename, auto...>
+struct TUV_t {};
+
+template <typename T, int... tuv>
+struct TUV_t<T, tuv...> : impl::TUV_t<T, TopologicalElement::CC, tuv...> {};
+
+template <typename T, TopologicalElement te, int... tuv>
+struct TUV_t<T, te, tuv...> : impl::TUV_t<T, te, tuv...> {};
```
Yurlungur: with the variadics is this going to get correctly resolved?

src/pack/pack_utils.hpp
```diff
@@ -144,6 +151,117 @@ struct any_nonautoflux : public base_t<true> {
...
+
+// Concept to state that a typed-field is dependent on other
+// fields, and is not itself an actual indexable field.
```
Yurlungur, line 156: should the concepts stuff be in concepts_lite or should it be here?

src/utils/type_list.hpp
```diff
@@ -95,6 +95,27 @@ auto ConcatenateTypeLists(TypeList<Args1...>, TypeList<Args2...>, Args...) {
...
+  using TL = TypeList<Ts...>;
+  using type = std::conditional_t<
+      TL::template IsIn<V>(), TL,
+      typename UnionTypeLists<TypeList<Ts..., V>, TypeList<Vs...>>::type>;
+};
+
+template <typename T, typename U, typename V, typename... Args>
+struct UnionTypeLists<T, U, V, Args...> {
+  using type = typename UnionTypeLists<typename UnionTypeLists<T, U>::type, V>::type;
+};
```
Yurlungur, line 122: I'm surprised there isn't functionality for this in the standard

src/pack/pack_utils.hpp
```diff
@@ -144,6 +151,117 @@ struct any_nonautoflux : public base_t<true> {
...
+  static auto get(T, Ts...) {
+    return union_type_lists_t<decltype(AllDependentVariables::get(T())),
+                              decltype(AllDependentVariables::get(Ts()...))>();
+  }
+};
+} // namespace impl
+
+// Get a TypeList of all the non-dependent variables that make up the requested types.
+template <typename... Ts>
+using all_dependent_variables_t = decltype(impl::AllDependentVariables::get(Ts()...));
```
Yurlungur: The name and the comment disagree. Is this all dependent variables or all independent variables?

Yurlungur (commented):
This is very cool and I like it a lot. It definitely needs a bit of documentation though. Though I suppose we also don't have documentation for type-based variables, which is sorely needed...

Yurlungur:
Also changelog and mark non-wip :)

src/pack/pack_utils.hpp
acreyes: The types should be renamed. It is all the independent variables pulled out from the dependent variables

src/pack/pack_utils.hpp
acreyes: I think so, since the `TE` & `int`s are different types. Happy to add some static asserts like [this](https://godbolt.org/z/dGndKv359) to the unit tests to demonstrate. Unless there is a different edge case you're thinking of?

acreyes:
> This is very cool and I like it a lot. It definitely needs a bit of documentation though. Though I suppose we also don't have documentation for type-based variables, which is sorely needed...

Happy to add both in this PR.

lroberts36:
This is very cool! I am wondering though about how complex the template stuff needs to be. Why can't we just do something like
```c++
class my_derived_var_t : public variable_names::dependent_var_base_t<...> { 
  ...
  template <class pack_t>
  KOKKOS_INLINE_FUNCTION
  Real evaluate(const pack_t &pack, int b, TopologicalElemet te, int k, int j, int i) {
    return pack(b, var1_t(), te, k, j, i) + pack(b, var2_t(), te, k, j, i);
  }
};
```
and 
```c++
template <class TIn, REQUIRES(IncludesType<TIn, Ts...>::value && std::is_base_of_v<dependent_var_base_t, TIn>)>
  KOKKOS_INLINE_FUNCTION Real operator()(const int b, const TE el, const TIn &t,
                                          const int k, const int j, const int i) const {
    return t.evaluate(*this, b, te, k, j, i);
  }
```
I think there are obviously some circular dependencies that could crop up here (e.g. `var2_t` is a dependent variable that depends on `my_derived_var_t`), but assuming the downstream user is responsible this allows things like stencil operations etc. (Caveat: I could be missing some obvious reason this doesn't work since I didn't code it up.)

acreyes:
> This is very cool! I am wondering though about how complex the template stuff needs to be. Why can't we just do something like
> 
> ```c++
> class my_derived_var_t : public variable_names::dependent_var_base_t<...> { 
>   ...
>   template <class pack_t>
>   KOKKOS_INLINE_FUNCTION
>   Real evaluate(const pack_t &pack, int b, TopologicalElemet te, int k, int j, int i) {
>     return pack(b, var1_t(), te, k, j, i) + pack(b, var2_t(), te, k, j, i);
>   }
> };
> ```
> 
> and
> 
> ```c++
> template <class TIn, REQUIRES(IncludesType<TIn, Ts...>::value && std::is_base_of_v<dependent_var_base_t, TIn>)>
>   KOKKOS_INLINE_FUNCTION Real operator()(const int b, const TE el, const TIn &t,
>                                           const int k, const int j, const int i) const {
>     return t.evaluate(*this, b, te, k, j, i);
>   }
> ```
> 
> I think there are obviously some circular dependencies that could crop up here (e.g. `var2_t` is a dependent variable that depends on `my_derived_var_t`), but assuming the downstream user is responsible this allows things like stencil operations etc. (Caveat: I could be missing some obvious reason this doesn't work since I didn't code it up.)

That is much simpler and I think would work just fine. My thought was that it would be nice to make it easy to unit test the `evaluate` methods downstream separate from the packing infrastructure. I think it would be simple enough to mock something that works with the template in your example. This could probably thread later with the sub-pack (#1265 ) infrastructure.

pgrete:
I just got the chance to look at this PR.
It's definitely a nice feature and the user facing interface looks clean and easy to use/write.

Quite of few of our targeted use cases would involve either stencil based ops and/or other information (say an adiabatic index).
Thus the structure @lroberts36 suggested seems more flexible in that regard.

How big of a lift do you think this would be (asking from the point of view whether we should get this interface in as is [pending  review comments] or spend "just a little" more time on getting a more flexible version in first place).

acreyes:
> I just got the chance to look at this PR. It's definitely a nice feature and the user facing interface looks clean and easy to use/write.
> 
> Quite of few of our targeted use cases would involve either stencil based ops and/or other information (say an adiabatic index). Thus the structure @lroberts36 suggested seems more flexible in that regard.
> 
> How big of a lift do you think this would be (asking from the point of view whether we should get this interface in as is [pending review comments] or spend "just a little" more time on getting a more flexible version in first place).

I think luke's suggestion would be pretty easy to implement and would be a better starting place for this kind of feature.

src/pack/pack_utils.hpp
acreyes: It seemed specialized enough to the pack types that I put it here, but if the convention is to have all concepts in concepts_lite I'm happy to move it

Yurlungur:
@acreyes do we still want to merge this? I'm still pretty excited about it

acreyes:
> @acreyes do we still want to merge this? I'm still pretty excited about it

Yeah it is something I want to use as well. I think  @pgrete wanted to review. Looking at #1415 I think that the `SubPack`s I added here could be made to behave more like the pack views introduced there, and might be something I want to do before merging this assuming that goes through first.

Yurlungur:
Sounds good. @pgrete can you review?

Yurlungur (approved):

doc/sphinx/src/sparse_packs.rst
```diff
@@ -85,6 +85,83 @@ snippet.
...
    dependent on each other. Issues resolved by setting ``-DPARTHENON_DEBUG_SCRATCH=ON`` can be an indication of this issue.
 
+Slicing into ``SparsePack``\ s with ``SubPack``\ s
+--------------------------------------------------
+
+A `SubPack` provdies a view into a slice of a `SparsePack` along a given dimension(s).
+`SubPack`\ s are constructed with a block + `kji` index to give a slice into the 
+fields at the meshblock + cell index. When a `SubPack` is constructed with the
+`Axis` template parameters then the `SubPack` also providies a view into slices
+of the `SparsePack` along the provided axes offset from the provided `ijk` indices.
```
lroberts36, line 95: I think a lot of this has overlap with the `pack_view`s in my loop abstraction PR. We should think about if there is a way to combine things, but maybe after this is merged.

src/pack/make_pack_descriptor.hpp
```diff
@@ -62,7 +64,8 @@ inline auto MakePackDescriptor(MT *pmd, const std::vector<std::string> &vars,
                             options);
 }
 
-template <class... Ts, class MT>
+template <class... Ts, class MT,
+          REQUIRES(!variable_names::dependent_variable_v<Ts> && ...)>
```
lroberts36:
```suggestion
          requires(!variable_names::dependent_variable_v<Ts> && ...)>
```

src/pack/make_pack_descriptor.hpp
```diff
@@ -71,7 +74,8 @@ inline auto MakePackDescriptor(MT *pmd, const std::vector<MetadataFlag> &flags =
       MakePackDescriptor(pmd, vars, use_regex, flags, options)));
 }
 
-template <class... Ts, class MT>
+template <class... Ts, class MT,
+          REQUIRES(!variable_names::dependent_variable_v<Ts> && ...)>
```
lroberts36:
```suggestion
          requires(!variable_names::dependent_variable_v<Ts> && ...)>
```

src/pack/make_pack_descriptor.hpp
```diff
@@ -110,6 +114,14 @@ inline auto MakePackDescriptorFromTypeList(Args &&...args) {
   return MakePackDescriptorFromTypeList(TL(), std::forward<Args>(args)...);
 }
 
+template <class... Ts, class MT,
+          REQUIRES(variable_names::dependent_variable_v<Ts> || ...)>
```
lroberts36:
```suggestion
          requires(variable_names::dependent_variable_v<Ts> || ...)>
```

doc/sphinx/src/sparse_packs.rst
lroberts36: Probably it is mostly orthogonal after I have looked some more. I could change `pack_view`s to support virtual variables, but that would be separate from `SubPacks`.

acreyes: I was thinking of redoing this on top of the `pack_view`s. My first look at what you added seemed to me like it would be pretty straightforward.

lroberts36 (approved):
LGTM! I think one of my comments was free floating instead of part of the review.

doc/sphinx/src/sparse_packs.rst
lroberts36: `pack_views` are pretty tightly coupled to the loop abstraction stuff, so maybe not a replacement unless everyone switched to the loop abstraction.

PR #1323, opened 2025-09-18
Pull out sparse dealloc into a templated function, move it out of update

Yurlungur:
## PR Summary

This MR is motivated by a desire in riot initialization to call sparse-dealloc manually in the process of problem generation. Basically we want to initialize things, then check if we can de-allocate, and then move on to the next block or pack in our pgen. This is less efficient than deallocating naturally in the task list, but it could save memory and let us initialize on a smaller memory footprint.

To make this possible, I pull out the implementation of `SparseDealloc` from `src/interface/update.hpp`, and introduce a new folder at the top level of `src` called `sparse` where I put two new functions in `sparse_management.{hpp,cpp}`. One checks if sparse fields are zero, and the other deallocates them. 

This change is non-breaking, as the original tasking functions still live in `update.hpp` and now call `sparse_management`. I think this is a little cleaner and more general. And it also opens up the door down the line to moving more sparse functionality (such as sparse pools) into this folder down the line. (I didn't do that now as I wanted a non-breaking MR that we could move in quickly to support something we're trying to achieve with riot.)

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


src/interface/mesh_data.hpp
```diff
@@ -536,6 +536,15 @@ std::vector<Uid_t> UidIntersection(MeshData<T> *md1, MeshData<T> *md2, Args &&..
...
 
+template <typename T>
+MeshBlockData<Real> *GetBlockDataPointer(T *rc, std::size_t b) {
+  if constexpr (std::is_same_v<T, MeshBlockData<Real>>) {
+    return rc;
+  } else {
+    return rc->GetBlockData(b).get();
+  }
+}
+
```
lroberts36: Shouldn't this live in `meshblock_data.hpp`?

src/interface/update.hpp
```diff
@@ -312,6 +312,7 @@ TaskStatus FillDerived(T *rc) {
   return TaskStatus::complete;
 }
 
+// TODO(JMM): Should this move into sparse_management?
```
lroberts36: Yes, I think so.

lroberts36 (approved):
LGTM

src/interface/mesh_data.hpp
Yurlungur: Yeah... I was lazy. I can put it in `meshblock_data.hpp`.

Yurlungur: Done.

src/interface/update.hpp
Yurlungur: Done.

Yurlungur:
Ok I moved all the sparse stuff out of update and into `sparse/sparse_management.hpp`. I did however leave `using` statements in `update.hpp` to not break downstream. That's my preference for now but if people feel strongly I can remove them and then downstream codes might have to change an include line or a namespace line.

Yurlungur:
> To make this possible, I pull out the implementation of SparseDealloc from src/interface/update.hpp, and introduce a new folder at the top level of src called sparse where I put two new functions in sparse_management.{hpp,cpp}. One checks if sparse fields are zero, and the other deallocates them.

I got rid of the zero check function. It was too painful to thread that through with CUDA. Apparently cuda does not like `auto` return types for *any* function that calls a host device lambda.

Yurlungur:
Ping @pdmullen for one more review

pdmullen (approved):

PR #1325, opened 2025-09-18
Add Kokkos::fence() in Swarm UpdateEmptyIndices.

RyanWollaeger:
## PR Summary

* This change was motivated by invalid particle behavior observed with HIP/ROCm 6.4.2 on RZAdams, and this change seems to reliably fix the errors for a particular Artemis+Jaybenne simulation.
* An Artemis+Jaybenne simultion showed particles reaching invalid state, which may happen if inactive particles are not masked; evidently the empty index list update can happen before the prior prefix sum is finished.

## Changes

+ Add Kokkos::fence() between prefix sum and list update.
+ Add || || at similar locations in Defrag (not tested).

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build


RyanWollaeger:
@pdmullen @alexrlongne @adamdempsey90 @Yurlungur

src/interface/swarm.cpp
```diff
@@ -342,6 +342,9 @@ void Swarm::UpdateEmptyIndices() {
         }
       });
 
+  // this is evidently needed for some HIP/ROCm/AMD platforms
+  Kokkos::fence();
+
```
adamdempsey90: I'm kind of surprised this is required because I would assume the `par_scan` and subsequent `par_for` are on the same "stream" and so should be sequential. But, if this is required, then can we try the less expensive `DevExecSpace()::fence`?

Do you know for certain that `empty_indices` is bad inside the following `par_for`, or is something else on host using it after this function returns and the `Kokkos::fence` is just needed at the end of this function?

RyanWollaeger: I did not directly observe the bad `empty_indices`, and also haven't tried putting the fence at the end of the routine. Can try that pretty quickly.

RyanWollaeger: On a bit of a side note, it looks like `src/kokkos_abstraction.hpp` has been significantly changed relative to the version in Artemis' submodule commit. The current one on `develop` seems to give me a compile error in some sort of simd check, which could indicate something else is wrong with my HIP/ROCm build. For now, I'll revert `kokkos_abstraction.hpp` and test relocating the fence.

RyanWollaeger: @adamdempsey90 - you're right - this works at the end of the routine (checked again that it is in fact needed there, at least). Will move this, convert it to `DevExecSpace()::fence()`, and remove the other ones in this change set.

adamdempsey90: Okay, that suggests to me that it needs to be a real `Kokkos:fence` and not the `DevExecSpace` one, though. Might be worth testing both.

RyanWollaeger: I'll also test it out for more cycles - just checked the first few after moving it.

RyanWollaeger: Hm - the error can actually still happen, at least with the updated location (time step 11 instead of 1) - let me retry with the original location (between the par_scan and the par_for).

RyanWollaeger: It definitely gets past it (on step 32 now, consistent with the 1st few runs). Of course this could fail much later (at step 100, e.g.). Will try with the `DevExecSpace` version next, then out to 64 steps.

RyanWollaeger: Looks like the way to do this is(?):
```c++
auto pmb = GetBlockPointer();
// ...
pmb->exec_space.fence();
```

RyanWollaeger: N/m - had the syntax wrong for `DevExecSpace`

RyanWollaeger: The `DevExecSpace` fence seems to be working - at 37 time steps and counting.

src/CMakeLists.txt
```diff
@@ -24,32 +24,32 @@ endif()
 # See `kokkos_abstraction.hpp` for available tags and what they translate to.
 if (Kokkos_ENABLE_CUDA OR Kokkos_ENABLE_HIP)
   set(PAR_LOOP_LAYOUT "MANUAL1D_LOOP" CACHE STRING
-    "Default loop layout for parallel_for wrapper")
+    "Default loop layout for parallel_for wrapper" FORCE)
```
RyanWollaeger, line 27: Was getting SIMD loop settings cached when trying to toggle `Kokkos_ENABLE_HIP` after an initial configuration pass, resulting in a compile-time error. This fixes it (but might be off-base of course).

src/parthenon_manager.hpp
```diff
@@ -45,6 +45,7 @@ class ParthenonManager {
   ParthenonInitPackagesAndMesh(std::optional<forest::ForestDefinition> forest_def = {});
   ParthenonStatus ParthenonFinalize();
 
+  bool IsRestart() { return (arg.restart_filename == nullptr ? false : true); }
```
RyanWollaeger, line 48: Not sure why this was removed - is it faulty? (Notwithstanding changes in Artemis, downstream, this is needed there...)

Yurlungur: You should now use `Globals::is_restart`. We had multiple ways of accessing whether or not a simulation was a restart and we tried to consolidate.

src/interface/swarm.cpp
Yurlungur: I echo @adamdempsey90 's surprise but 🤷 . Thanks for the fix!

src/CMakeLists.txt
Yurlungur: Surprised that's necessary but reasonable. Thanks for the fix.

Yurlungur (changes requested):
See my comment about `Globals::is_restart` can you switch to that downstream? Otherwise, thanks for all the fixes!

src/CMakeLists.txt
pgrete: Hmm, I'd not be surprised if this is required in more places as I don't think we cleanly allow to reconfigure. I'm not even sure what the cmake rules/best practices are for this.
I typically delete the CMakeCache files and dirs for a clean build.

Bottom line, I'm not sure we should make this change because it may convey a false sense of correctness.

RyanWollaeger:
Thanks @Yurlungur ! Will try getting `is_restart` in Artemis and then revert that change here. Retesting my simulation now without the fence just to be sure the other latest Parthenon changes don't also fix my particle errors. I noticed the simulation is sped up with the head of Parthenon now... which is good.

src/interface/swarm.cpp
```diff
@@ -342,6 +342,9 @@ void Swarm::UpdateEmptyIndices() {
         }
       });
 
+  // this is evidently needed for some HIP/ROCm/AMD platforms
+  DevExecSpace().fence();
+
```
pgrete, line 347: Here (and below for the fences), I do not understand how this happens because the way we're using the Kokkos programming model at the moment, all kernels are executed sequentially in order they are submitted.
Thus, there's either a bug in Kokkos or (more likely) I'm really afraid that adding fences hides a deeper down design issue (e..g, the fences results in different order in the tasks that make the issue go away).
I have no doubt that there is an issue but I'd feel much better if I'd understand where it is and fix it without adding fences (because otherwise I bet a beer at the next dev meeting that we'll spend way more time debugging the hidden issue down the road).

I  may also be wrong and happy to be convinced otherwise on wy the fences at those locations specifically fix a general issue.

RyanWollaeger: Ah - okay - I typically don't delete the CMakeCache file between builds, and sometimes do a second configuration pass by toggling parameters through `ccmake`.

RyanWollaeger: Concerning the fences, that is what surprised @adamdempsey90 and @Yurlungur, and it certainly seems like it could conceal another issue. In my most recent run with the latest Parthenon `develop`, commenting this fence out seems to not cause the issue (over the first 64 steps), so it might be that it is a problem only with the submodule commit of parthenon in Artemis and has been fixed in a different way. I could close this PR in favor of updating Artemis' submodule and restart check - is that reasonable?

adamdempsey90: If I recall correctly, the kokkos documentation says that they have the right to execute kernels asynchronously if they want without telling you. As of today, they don't do that, but it helps me sleep at night knowing we've added a fence between two data-dependent kernels so that our future selves have less potential bugs to deal with.

src/CMakeLists.txt
RyanWollaeger: Meant to reply with my first comment here - that seems reasonable - I suppose a downstream app might set these PAR_LOOP variables incorrectly, which these FORCEs would quietly reset given the Kokkos flags. That occurred to me (but didn't seem unreasonable if these loop parameters should always be these values for these Kokkos flags).

src/interface/swarm.cpp
RyanWollaeger: It doesn't seem to cause a performance hit in my sim, AFAICT

Yurlungur: IIRC Cuda does guarantee the kernels will be non-blocking but execution will be ordered on the same stream. Does HIP actually make the same guarantee?

src/CMakeLists.txt
Yurlungur: > I suppose a downstream app might set these PAR_LOOP variables incorrectly, which these FORCEs would quietly reset given the Kokkos flags.

A downstream code may indeed wish to use a different loop pattern... so I think @pgrete is right we shouldn't use `FORCE`. I'm not sure what the right thing to do here is regarding the cmake cache... Perhaps just user beware.

src/interface/swarm.cpp
RyanWollaeger: I couldn't find definitive information on this - will ask around. At least in these Frontier slides, it is indicated it is guaranteed: https://www.olcf.ornl.gov/wp-content/uploads/2019/10/CAAR_HIP_on_Frontier.pdf

RyanWollaeger: However, I think @adamdempsey90 did not observe the errors on Frontier that I observed on RZAdams...

pgrete: a) Glad this seems to be resolved, in which case I'd be in favor of not making any changes.
b) IIRC the programming model guarantees that kernels in the same execution space are (like a cuda stream) are executed in order but can overlap between execution spaces (like different streams or host/device `par_for`s).
c) if we have data dependencies, I think we should provide combined `Tasks` for downstream codes to enroll or make warning statement that downstream codes have to manually execute certain task in order/with the correct dependencies (rather than relying on fences).

Yurlungur: I agree with that.

pgrete: > However, I think @adamdempsey90 did not observe the errors on Frontier that I observed on RZAdams...

One more piece of info.
Sth is currently off on Frontier (potentially more broader on some combination of Cray/Slingshot/AMD GPUs).
I'm in contact with folks at OLCF as even a simple advection tests with AMR currently needs `export  MPICH_SMP_SINGLE_COPY_MODE=NONE` (i.e., disabling intra-node GPU-to-GPU transfer) to circumvent `memory access error`s.

RyanWollaeger:
Going to go ahead and close this - thanks for all the feedback @Yurlungur  @pgrete @adamdempsey90 - if I see new particle errors, I can retry this fence and then maybe think more about the task dependency as @pgrete noted.

src/interface/swarm.cpp
RyanWollaeger: Oh interesting... will keep that in mind - I was toying around a lot with `MPICH_GPU` env variables initially.

PR #1326, opened 2025-09-19
Allow nmax_pool to be set by metadata

Yurlungur:
## PR Summary

This MR threads to userspace the initial size of the reservation of a swarm pool (i.e., the number of particles it can hold). This doesn't change functionality, as the pool grows greedily, but it can be useful because it will require fewer malloc calls as particle counts grow and the pool grows greedily. You can now set it with

```C++
metadata.SetInitialSwarmPoolReservation(n);
```

I also change this reservation size internally to be a `std::size_t` instead of an `int`, which I think it should be anyway. As suggested/requested by the Artemis/Jaybenne team. Ping @adamdempsey90.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/src/interface/metadata.hpp
+++ b/src/interface/metadata.hpp
@@ -375,6 +375,9 @@ class Metadata {
   std::shared_ptr<Metadata> flux_metadata;
 
  public:
+  // TODO(JMM): I'm not sure where this should go...
+  static constexpr const std::size_t SWARM_DEFAULT_NMAX_POOL = 3;
+
   // Static routines
   static MetadataFlag AddUserFlag(const std::string &name);
   static bool FlagNameExists(const std::string &flagname);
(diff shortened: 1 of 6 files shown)


RyanWollaeger:
Awesome! Thanks @Yurlungur !

adamdempsey90 (approved):

src/interface/metadata.hpp
```diff
@@ -375,6 +375,9 @@ class Metadata {
   std::shared_ptr<Metadata> flux_metadata;
 
  public:
+  // TODO(JMM): I'm not sure where this should go...
+  static constexpr const std::size_t SWARM_DEFAULT_NMAX_POOL = 3;
+
```
pgrete, line 380: Yeah, good question. Eventually, we could consider separating concerns between dense, sparse, swarms, etc. but, on the other hand, there's also quite some overlap.
I think it's fine to keep it here for now.

pgrete (approved):
Changelog and we're good to go!

PR #1150, opened 2024-08-08
Reduce memory consumption for buffer pool

pgrete:
## PR Summary

Tries to reduce the memory footprint of the object pool by precalculating the number of buffers required for each type.
Also fixes an original overflow for very large buffer sizes (and the default 200 number of buffers).

## PR Checklist

- [x] Code passes cpplint
- [ ] New features are documented.
- [ ] Adds a test for any bugs fixed. Adds tests for new features.
- [x] Code is formatted
- [x] Changes are summarized in CHANGELOG.md
- [ ] Change is breaking (API, behavior, ...)
  - [ ] Change is *additionally* added to CHANGELOG.md in the breaking section
  - [ ] PR is marked as breaking
  - [ ] Short summary API changes at the top of the PR (plus optionally with an automated update/fix script)
- [ ] CI has been triggered on [Darwin](https://re-git.lanl.gov/eap-oss/parthenon/-/pipelines) for performance regression tests.
- [x] Docs build
- [ ] (@lanl.gov employees) Update copyright on changed files

PR #1330, opened 2025-10-16
Try to decrease comm buffer memory footprint... part 2

Yurlungur:
## PR Summary

We've been fighting with memory footprint recently and were looking for ways to reduce it. This is (sort of) a follow on to #1150. When more comm buffers need to be allocated, either because of remeshing, or because of sparsity, they are allocated in chunks. Currently, the chunk size is large, and tuned to the total number of blocks, to minimize the number of allocations for dense simulations. And buffers are never deallocated.

This is fine probably for dense simulations where the block count stays roughly constant, but isn't ideal when (a) the block count might change a lot or (b) when we have sparse variables active. This MR attempts to resolve both those issues in a user tunable way. It makes the following changes:

1. The chunk size is now tuned to the number of **allocated/active** comm buffers expected, and is sparse aware.
2. The chunk size is also user set-able, via a new input parameter `parthenon/mesh/comm_buffer_granularity`. If it's a positive value, the user-set number takes precedence. Otherwise the behavior above is used.
3. It is now possible to de-allocate comm buffers. A new mesh function `TryReallocCommBufferPools` checks how many buffers are used. If less than `parthenon/mesh/comm_buffer_reset_fraction` of allocated buffers are in use, **and** the chunk size is not so large that reallocation wouldn't put us back where we currently are, all buffer pools are cleared and rebuilt, hopefully with a smaller number of comm buffers. I chose 0.8 as this factor, i.e., 20% of all allocated buffers are unused. I don't know if that's a good choice. We might also want a larger safety factor on the chunk size.
4. You can call `TryReallocCommBufferPools` by hand. Or, if you're using an evolution driver, you can set `parthenon/time/comm_buffer_reset_cadence` to a positive number. Then the call will be made at that cadence in cycles.

There are two additional points of complexity here:
1. The buffer pool cannot just be cleared, as the reference counting means there are circular pointers... buffers must own a pointer to the pool. I resolve this by having the pool walk through all the buffers it owns and set their "pool" pointers by hand.
2. Clearing out the buffer pool does not actually free comm buffers. The reason is that the comm buffers are views under the hood, and Kokkos does its own reference counting of those views. The views thus need to be freed separately by clearing and rebuilding the comm buffer cache objects.

Thanks to @lroberts36 for helping me work through that complexity.

My hope is that this will help significantly reduce the number of comm buffers we actually need in simulations where the mesh (and sparsity) is changing a lot. I ran sparse advection as a simple test, and it improved memory overhead by a non-trivial amount. With all these changes and a buffer pool reallocation every 10 cycles, I see about 0.4GB of memory usage on my laptop on this branch, compared to 0.35GB of memory usage on develop, which is a 10% decrease. I expect this matters more the more sparsity is in a problem.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


Yurlungur:
Marking WIP for now because tests are failing and I haven't had time to do, e.g., changelog. But I think this is ready for review.

src/bvals/comms/build_boundary_buffers.cpp
```diff
@@ -147,13 +148,27 @@ void BuildBoundaryBufferSubset(std::shared_ptr<MeshData<Real>> &md,
           buf_map[r_key] = CommBuffer<buf_pool_t<Real>::owner_t>(
               tag, receiver_rank, sender_rank, comm, get_resource_method,
               use_sparse_buffers);
-        // Register this buffer with the combined buffers (must happen after CommBuffer is
-        // created)
-        pmesh->pcoalesced_comms->AddRecvBuffer(pmb, nb, v, BTYPE);
```
Yurlungur, line 152: this machinery needed to be split out because I wanted to call the build boundary buffers function on the base container, which correctly counts up comm buffers for everything. I think it's cleaner split out anyway though.

src/driver/driver.cpp
```diff
@@ -182,6 +182,9 @@ DriverStatus EvolutionDriver::Execute() {
 
       timer_LBandAMR.reset();
       pmesh->LoadBalancingAndAdaptiveMeshRefinement(pinput, app_input);
+      if ((buffer_reset_cadence_ > 0) && (tm.ncycle % buffer_reset_cadence_ == 0)) {
```
Yurlungur, line 185: I was very tempted to write 
```C++
      if ((buffer_reset_cadence_ > 0) && !(tm.ncycle % buffer_reset_cadence_)) {
```
but I resisted. (This time.)

tst/regression/test_suites/sparse_advection/parthinput.sparse_advection
```diff
@@ -57,6 +57,7 @@ nlim = -1
 tlim = 1.0
 integrator = rk2
 ncycle_out_mesh = -10000
+comm_buffer_reset_cadence = 10
```
Yurlungur, line 60: to ensure the new buffer machinery is tested

Yurlungur:
Ok this is not going to work as submitted. We can't have weak pointers on device in the buffer pool objects. I'll think about what to do and discuss with @lroberts36

Yurlungur:
> Ok this is not going to work as submitted. We can't have weak pointers on device in the buffer pool objects. I'll think about what to do and discuss with @lroberts36

This issue is resolved. However this MR appears to break coalesced comms. I don't understand why.

Yurlungur:
OK I believe this works now. It's ready for review. @pgrete @pdmullen @lroberts36 @jdolence @chadmeyer

src/bvals/comms/bnd_id.hpp
```diff
@@ -81,6 +81,8 @@ struct BndId {
   BndId() = default;
   KOKKOS_DEFAULTED_FUNCTION
   BndId(const BndId &) = default;
+  KOKKOS_DEFAULTED_FUNCTION
+  ~BndId() = default;
```
Yurlungur, line 85: This appears to be necessary on AMD systems. I'm not sure why we didn't need it until now.

src/bvals/comms/bnd_info.hpp
```diff
@@ -82,6 +82,8 @@ struct BndInfo {
   BndInfo(const BndInfo &) = default;
   BndInfo(MeshBlock *pmb, const NeighborBlock &nb, std::shared_ptr<Variable<Real>> v,
           CommBuffer<buf_pool_t<Real>::owner_t> *combuf, IndexRangeType idx_range_type);
+  KOKKOS_DEFAULTED_FUNCTION
+  ~BndInfo() = default;
```
Yurlungur, line 86: and same here.

src/mesh/mesh.cpp
```diff
@@ -1184,6 +1173,68 @@ void Mesh::DoStaticRefinement(ParameterInput *pin) {
...
+  boundary_comm_map.clear();
+  pcoalesced_comms->clear();
+
+  // JMM: Estimates about chunk size, pre-allocation, etc., are tuned to
+  // all blocks on a rank, so we use the base partition.
+
+  // TODO(JMM): I don't like that we need to add base with the base
+  // partition here. We shouldn't have to, but I don't know how to fix
+  // it without a refactor.
+  BuildBoundaryBuffers(mesh_data.Add("base", GetBasePartition()));
```
Yurlungur, line 1187: This might deserve an issue. I don't think it's intended behavior actually.

Yurlungur:
@lroberts36 I did not go with the weak pointer strategy you suggested, as smart pointers can't be captured inside device lambdas, even weak pointers.

doc/sphinx/src/boundary_communication.rst
```diff
@@ -324,7 +324,7 @@ pools for various buffer sizes.
 
    template <typename T>
    using buf_pool_t = ObjectPool<BufArray1D<T>>
-   std::unordered_map<int, buf_pool_t<Real>> pool_map;
+   std::unordered_map<int, std::shared_ptr<buf_pool_t<Real>>> pool_map;
```
lroberts36:
```suggestion
   std::unordered_map<int, buf_pool_t<Real>> pool_map;
```

I think this change got reverted

src/driver/driver.hpp
```diff
@@ -128,6 +128,13 @@ class EvolutionDriver : public Driver {
         "Set to true to generate outputs in a step BEFORE modifying the mesh at the end "
         "of the step. By default outputs happen AFTER remeshing if remeshing happens. "
         "WARNING: this will make restarts not bitwise-exact.");
+    buffer_reset_cadence_ = pin->GetOrAddInteger(
+        "parthenon/time", "comm_buffer_reset_cadence", -1,
+        "If set to a positive number N, the comm buffer object pool will be reset "
+        "every N remeshing events if the expected number of required buffers has "
```
lroberts36: Doesn't this just reset every `N` time steps (not remeshing events)?

src/utils/object_pool.hpp
```diff
@@ -80,6 +81,26 @@ class ObjectPool {
   void AddFreeObjectToPool(const T &in) { available_.push(in); }
   void AddFreeObjectToPool(T &&in) { available_.emplace(in); }
 
+  // JMM: Note that calling this while, e.g., comm buffers still exist
+  // will make the memory a comm buffer points to
+  // invalid. Fortunately, comm buffers always call `Allocate`
+  // beforehand.
```
lroberts36: I don't think this is true. Calling `ObjectPool::Clear()` will just remove all of the `Kokkos::views` from the object pool. If they are still referenced elsewhere (e.g. in `Mesh::boundary_comm_map`), the memory should still be valid since the view reference count never dropped to zero. As long as the `ObjectPool` object hasn't been destroyed, eventually calling the destructor of the object that has been released from the pool should still work fine, since a call to `IsValid` will return `false` and the destructor won't do anything to the object pool it points at.

src/mesh/mesh.cpp
```diff
@@ -1184,6 +1173,68 @@ void Mesh::DoStaticRefinement(ParameterInput *pin) {
...
+  }
+  if (realloc) {
+    // The buffer pool must be cleared out, since otherwise, buffers
+    // will just be reference-count-freed
+    for (auto &[k, pool] : pool_map) {
+      pool.Clear();
+    }
+    // We need to clear the caches because comm buffers are kokkos
+    // views, which are reference counted, and they're stored in the
+    // caches.
```
lroberts36: `BvarsSubCache_t::buf_vec` is a `std::vector<CommBuffer<buf_pool_t<Real>::owner_t> *>`, so we need to clear the caches because these pointers are no longer valid.

lroberts36 (approved):
LGTM, thanks for doing this! Just a couple of comments about comments.

src/driver/driver.hpp
Yurlungur: oops yes

src/utils/index_split.cpp
```diff
@@ -64,7 +64,7 @@ void IndexSplit::Init(MeshData<Real> *md, const int kbe, const int jbe) {
   // equivalent to NSMS in Kokkos
   // TODO(JMM): I'm not sure if this is really the best way to do
   // this. Based on discussion on Kokkos slack.
-#ifdef KOKKOS_ENABLE_CUDA
+#ifdef PARTHENON_ENABLE_GPU
```
Yurlungur, line 67: without this the wrong defaults are being chosen on AMD systems

Yurlungur:
Thanks @lroberts36 I think I addressed all your concerns

doc/sphinx/src/boundary_communication.rst
```diff
@@ -338,6 +338,65 @@ Note that every stage shares the same ``CommBuffer``\ s, but we keep
...
+Memory for comm buffers in the buffer pools is allocated in "chunks"
+of some fixed size. Ideally this chunk size is large enough to
+minimize individual ``malloc`` calls but not over-allocate by too
+much. This can be set via the ``comm_buffer_granuarity`` setting in
+the ``parthenon/mesh`` block of the input deck. E.g.,
+
+.. code::
+
+   <parthenon/mesh>
+   comm_buffer_granularity = 200
```
pgrete: Adding a comment here, but it's more general: I was a little confused bu the term "granularity" in combination with "heuristic" because I interpreted it in a way that positive value also refer to some "granularity heuristic" factor rather than "granularity = chunk size".
If this line of thought is reasonable, I suggest to to rename the parameter to `comm_buffer_chunk_size` (or sth along those lines).
Alternatively, I could imagine to be more specific in the doc, i.e, have sth like "Here, 200 means that 200 buffers are allocated in a single call. The default, ..."

doc/sphinx/src/boundary_communication.rst
```diff
@@ -338,6 +338,65 @@ Note that every stage shares the same ``CommBuffer``\ s, but we keep
...
+
+The default, ``-1``, tells parthenon to make a heuristic choice based
+on the number of meshblocks per rank. This chunk size will always be
+at least 1 buffer and never more than the total number of buffers it
+is possible for a given mesh to require.
+
+In general a smaller chunk size will require more ``malloc`` calls
+(which are expensive) but will imply a smaller memory footprint, as
+the code is less likely to over-allocate beyond the number of buffers
```
pgrete: Maybe add a note at the bottom that this is only relevant for AMR and/or sims with sparse variables as buffers are being reused (and thus for static/dense meshes any "overhead" is only present in the very first cycle).

example/diffusion/CMakeLists.txt
```diff
@@ -24,6 +24,14 @@ if( "diffusion-example" IN_LIST DRIVER_LIST OR NOT PARTHENON_DISABLE_EXAMPLES)
           parthenon_app_inputs.cpp
   )
   target_link_libraries(diffusion-example PRIVATE Parthenon::parthenon)
+
+
+  # Suppresses annoying ABI notes. See:
+  # https://stackoverflow.com/questions/52020305/what-exactly-does-gccs-wpsabi-option-do-what-are-the-implications-of-supressi
+  target_compile_options(diffusion-example PRIVATE
+    $<$<AND:$<CXX_COMPILER_ID:GNU>,$<COMPILE_LANGUAGE:CXX>>:-Wno-psabi>
+  )
```
pgrete, line 33: Out of curiosity, how did this pop up (and in what environment)?
And why only for these three examples?

I haven't come across this (but recently ran into issue with inconsistent python stacks on Frontier that resulted in some segfaults difficult to track down )

src/utils/index_split.cpp
pgrete: huh, interesting.
This has been around for some time, how did we miss this?
I should probably check what this does to performance on Frontier.

src/kokkos_abstraction.hpp
```diff
@@ -37,8 +37,19 @@
...
+
 namespace parthenon {
 
+// namespaced constexpr version of above macro
+#ifdef PARTHENON_ENABLE_GPU
+constexpr bool ENABLE_GPU = true;
+#else
+constexpr bool ENABLE_GPU = true;
+#endif // PARTHENON_ENABLE_GPU
+
```
pgrete, line 52: 👍

src/bvals/comms/build_boundary_buffers.cpp
```diff
@@ -61,6 +64,7 @@ void BuildBoundaryBufferSubset(std::shared_ptr<MeshData<Real>> &md,
     if (pmb->gid == nb.gid && nb.offsets.IsCell()) buf_size = 0;
 
     nbufs[buf_size] += 1; // relying on value init of int to 0 for initial entry
+    nbufs_allocated[buf_size] += v->IsAllocated();
```
pgrete, line 67: Where is `nbufs_allocated` ever being updated/reset/decremented?

src/mesh/mesh.cpp
```diff
@@ -1184,6 +1173,67 @@ void Mesh::DoStaticRefinement(ParameterInput *pin) {
...
+    for (int gmg_level = GetGMGMinLevel(); gmg_level <= GetGMGMaxLevel(); ++gmg_level) {
+      const auto grid_id = GridIdentifier::two_level_composite(gmg_level);
+      for (auto &partition : GetDefaultBlockPartitions(grid_id)) {
+        auto &mdg = mesh_data.Add("base", partition);
+        BuildBoundaryBuffers(mdg);
+        BuildGMGBoundaryBuffers(mdg);
+        if (do_coalesced_comms) RegisterCoalescedCommsGMG(mdg);
+      }
+    }
+  }
```
pgrete, line 1201: This is now new because of the cleanup (separating RegisterCoal... calls rather than the registration at the end of the `BuildBoundaryBufferSubset` function), isn't it?

src/utils/object_pool.hpp
```diff
@@ -80,6 +81,22 @@ class ObjectPool {
...
+  void Clear() {
+    // normalize by moving everything into the stack
+    for (auto &[k, v] : inuse_) {
+      v.first.pool_ = nullptr;
+      Free(v.first);
+    }
+    // walk through the stack, disable reference counting, kill the
+    // buffer
+    while (!available_.empty()) {
+      available_.top().pool_ = nullptr;
```
pgrete, line 93: Just to double check: there's not other place in the code where the pool objects may live so that they're free here for sure?

pgrete (approved):
LGTM and thanks for the doc.

Comments are mostly for clarification (so I'm approving).

doc/sphinx/src/boundary_communication.rst
Yurlungur: :+1: I will change the name

doc/sphinx/src/boundary_communication.rst
Yurlungur: :+1:

example/diffusion/CMakeLists.txt
Yurlungur: gcc 11.4.0, my Ubuntu laptop. ;) Here's the error I get

```bash
In file included from /home/jonahm/programming/parthenon/example/poisson_gmg/poisson_package.cpp:33:
/home/jonahm/programming/parthenon/example/poisson_gmg/poisson_equation.hpp: In instantiation of ‘auto poisson_package::PoissonEquation<var_t, D_t>::GetEffectiveInverseDx2(const coords_t&, int, int, int) [with parthenon::CoordinateDirection dir = parthenon::X1DIR; coords_t = parthenon::UniformCartesian; var_t = poisson_package::u; D_t = poisson_package::D]’:
/home/jonahm/programming/parthenon/example/poisson_gmg/poisson_equation.hpp:125:64:   required from ‘parthenon::TaskStatus poisson_package::PoissonEquation<var_t, D_t>::SetDiagonal(std::shared_ptr<parthenon::MeshData<double> >&, std::shared_ptr<parthenon::MeshData<double> >&) [with var_t = poisson_package::u; D_t = poisson_package::D]’
/home/jonahm/programming/parthenon/src/solvers/bicgstab_solver.hpp:128:37:   required from ‘parthenon::TaskID parthenon::solvers::BiCGSTABSolver<equations_t, preconditioner_t>::AddSetupTasks(parthenon::TaskList&, parthenon::TaskID, int, parthenon::Mesh*) [with equations_t = poisson_package::PoissonEquation<poisson_package::u, poisson_package::D>; preconditioner_t = parthenon::solvers::MGSolver<poisson_package::PoissonEquation<poisson_package::u, poisson_package::D>, parthenon::solvers::ProlongationBlockInteriorZeroDirichlet>]’
/home/jonahm/programming/parthenon/src/solvers/bicgstab_solver.hpp:121:10:   required from here
/home/jonahm/programming/parthenon/example/poisson_gmg/poisson_equation.hpp:92:61: note: parameter passing for argument of type ‘std::pair<double, double>’ when C++17 is enabled changed to match C++14 in GCC 10.1
   92 |     return std::make_pair(Ap / (dxp * Vol), Am / (dxm * Vol));
      |  
```

It seems to show up for any calls where `std::make_pair` is the return argument with a move constructor. The reason we don't see it in parthenon proper, is we already have this line in parthenon for the parthenon shared library object.  I considered making that line "`PUBLIC`" scope in `CMake` but I figured that might hide issues people actually want to see in downstream code. So I just disabled it for the two examples where it shows up for me.

src/utils/index_split.cpp
Yurlungur: It probably doesn't impact you. It only matters for `index_split`, which I think only riot uses.

src/bvals/comms/build_boundary_buffers.cpp
Yurlungur: It's local to `build_boundary_buffers.cpp` so whenever that function is called. In practice, though, it's only set/updated once per pool shape. The chunk size is set at initialization from the initial mesh.

src/mesh/mesh.cpp
Yurlungur: Correct, yes. I moved the coalesced comms logic out of `BuildBoundaryBuffers` so now it has to be called separately.

src/utils/object_pool.hpp
Yurlungur: This function is defined inside the pool, so this frees the weak_t members for sure. *But* the buffers are owned by value in `BndInfo` and friends. That's one reason it's so important that the comm caches are cleared.

Yurlungur:
Thanks, @pgrete!

PR #1331, opened 2025-10-22
Use  for sizes

pgrete:
## PR Summary

I ran into an issues when trying to write a conserved variable vector for a 16384^3 mesh (turns out it was more related to the local mesh size rather than the total).
This motivated going a little more through the code base and fix some type mismatches.
There are still plenty left (which should be addressed in the context of #175) .

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build


src/outputs/output_utils.hpp
```diff
@@ -75,11 +75,11 @@ struct VarInfo {
...
   Triple_t<IndexRange> GetPaddedBoundsKJI(const IndexDomain domain) const;
 
-  int Size() const;
+  size_t Size() const;
   // Includes topological element shape
-  int TensorSize() const;
+  size_t TensorSize() const;
   // Size of region that needs to be filled with 0s if not allocated
-  int FillSize(const IndexDomain domain) const;
+  size_t FillSize(const IndexDomain domain) const;
```
pgrete, line 82: Those `int`s were the original issue.

Those propagated into the openpmd output for calculating the size of the output buffer vector and resulted in an overflow when multiplied by the number of local blocks.
Note, the hdf5 output is fine as the results from these calls are directly being put into a `size_t` variable.

src/mesh/mesh.cpp
```diff
@@ -861,6 +861,14 @@ void Mesh::Initialize(bool init_problem, ParameterInput *pin, ApplicationInput *
   }
 #endif
 
+  PARTHENON_REQUIRE_THROWS(
+      nbtotal < std::numeric_limits<int>::max(),
+      "Congratulations. You're the first one to run a simulation with more blocks than "
+      "max `int`. Many loops in parthenon still use `int` indices over blocks or use "
+      "`int` for pack indices and partitions, so who knows what happens next. Please get "
+      "in contact with the dev team before proceeding (or proceed on your own risk and "
+      "remove this statement).");
```
Yurlungur, line 859: Hah. Good error message.

Yurlungur (approved):
Thanks for the fix! This is long overdue.

I think we should use `std::size_t` instead of `size_t` just for clarity/C++ consistency.

src/outputs/restart_hdf5.hpp
```diff
@@ -134,13 +134,13 @@ class RestartReaderHDF5 : public RestartReader {
...
     const auto &shape = m.Shape();
-    const int rank = shape.size();
+    const auto rank = shape.size();
     const bool is_vector = m.IsSet(Metadata::Vector);
     std::size_t total_count = count;
     for (int i = 0; i < CHUNK_MAX_DIM; ++i) {
       h5_offset[i] = h5_count[i] = 0;
     }
-    for (int i = 0; i < rank; ++i) {
+    for (auto i = 0; i < rank; ++i) {
```
lroberts36, line 143: Why `auto` here instead of `size_t`?

src/pack/pack_utils.hpp
```diff
@@ -102,7 +102,7 @@ struct var_base_t {
   KOKKOS_INLINE_FUNCTION
   static int ndim() { return sizeof...(NCOMP); }
```
lroberts36:
```suggestion
  static size_t ndim() { return sizeof...(NCOMP); }
```

lroberts36 (approved):
Thanks for going through and fixing all these, I know I am guilty of using `int` in place of `size_t` in many places!

src/outputs/restart_hdf5.hpp
pgrete: Updated.

pgrete:
I think I got all of them now. Does anyone wants to have another look given the extended size of the changeset now?

pgrete:
> I think I got all of them now. Does anyone wants to have another look given the extended size of the changeset now?

Merging now (as the openpmd PR depends on this).

pgrete:
Why is the WIP check pending?

Yurlungur:
> Why is the WIP check pending?

That workflow sometimes fails. I think we should just disable it and force merge.

pgrete:
> > Why is the WIP check pending?
> 
> That workflow sometimes fails. I think we should just disable it and force merge.

Done.

I cannot even find where we specified that workflow/check.
I see it in the branch protection rules on Github but nothing in the repo itself. Maybe I'm missing sth.

PR #1332, opened 2025-10-24
Add control over whether to include/exclude an output on final signal

pgrete:
## PR Summary

I came across the "issue" that a large number of output blocks contributed significantly to the wallclock time on gracefully exiting the simulation.
I didn't need all these outputs as I only needed the restart output.
This quick hack allows to select which outputs to include.
I don't think it hurts (as it's a small fix) but raises a potentially separate question/discussion: should we differentiate between simulation that shutdown because of running into the wallclock limit `-t ...` versus simulations that shutdown because they're "complete" (by running into `tlim` or `nlim`).

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Current develop
 
 ### Added (new features/APIs/variables/...)
+- [[PR 1331]](https://github.com/parthenon-hpc-lab/parthenon/pull/1331) Add control over whether to include/exclude an output on final signal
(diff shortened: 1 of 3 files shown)


Yurlungur (approved):
This hack looks totally reasonable to me... Broadly I think your question is a good point though. I think I *always* want restarts when we hit wallclock, but not everything else.

acreyes (approved):
LGTM. I too would only want restarts on wall clock limits

pgrete:
Thanks for the quick reviews.
Seems like there's agreement that it's worth to separate the "final" cases.
New issue here #1334 to keep track (I think it'd be a pretty small lift, but one that breaks default behavior).

PR #1337, opened 2025-11-12
Add task list based profiling

lroberts36:
## PR Summary

This PR adds the capability to time groupings of tasks in a task list for profiling. For instance, it was used to create the the by multigrid level timing plot shown below. For an example of using the task list timing, see the unit test. 

Additionally, this PR starts on adding some unit tests for the tasking infrastructure itself.
[mg_timing.pdf](https://github.com/user-attachments/files/23511376/mg_timing.pdf)

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [ ] Docs build


src/tasks/task_timing.hpp
```diff
@@ -0,0 +1,113 @@
...
+class TimingAccumulatorDictionary {
+  std::map<std::string, std::shared_ptr<TimingAccumulator>> dict_;
+
+ public:
+  std::shared_ptr<TimingAccumulator> GetOrAddAndRegister(const std::string &label,
+                                                         TaskList &tl);
+
+  std::shared_ptr<TimingAccumulator> Get(const std::string &label) {
+    PARTHENON_REQUIRE(dict_.count(label) > 0, "Asking for non-existent timing region.");
+    return dict_[label];
```
Yurlungur, line 99: Why not just do this?
```suggestion
    return dict_.at(label);
```

src/tasks/task_timing.hpp
```diff
@@ -0,0 +1,113 @@
...
+  explicit TimingAccumulatorGuard(std::shared_ptr<TimingAccumulator> timing_accumulator)
+      : tidc(timing_accumulator) {
+    tidc->StartCollectingTasks();
+  }
+  ~TimingAccumulatorGuard() { tidc->StopCollectingTasks(); }
+  std::shared_ptr<TimingAccumulator> tidc;
+};
+
+class TaskList;
+class TimingAccumulatorDictionary {
```
Yurlungur, line 90: So I was going to suggest that this could be replaced with
```C++
using TimingAccumulatorDictionary = std::map<std::string, std::shared_ptr<TimingAccumulator>, TimingAccumulatorAllocator>;
```
but I see that you have the constructor for `TimingAccumulator` as private, so I'm not sure that's possible actually. Also it seems writing a new allocator class is sort obnoxious.

src/tasks/task_timing.cpp
```diff
@@ -0,0 +1,109 @@
...
+}
+
+std::shared_ptr<TimingAccumulator>
+TimingAccumulatorDictionary::GetOrAddAndRegister(const std::string &label, TaskList &tl) {
+  if (dict_.count(label) == 0) dict_[label] = TimingAccumulator::create();
+  tl.RegisterTimingAccumulator(dict_[label]);
+  return dict_[label];
+}
+
+void TimingAccumulatorDictionary::WriteToJSON(const std::string &filename) {
```
Yurlungur, line 56: This is essentially just a 3-column table. Is json really the optimal format for this? Why not just a 2D ascii array?

Yurlungur (commented):
Looks like some of the new tests you added are failing. Otherwise LGTM.

src/tasks/task_timing.hpp
lroberts36: I thought about this, but decided against it for the reasons you mentioned.

src/tasks/task_timing.cpp
lroberts36: It is a ragged array. Each timing region has a vector of start and stop times for the tasks in the region, so the data could look something like
```
"region1" : [(0, 0.1), (0.5, 0.65), (0.8, 0.81)]
"region2": [(0.1, 0.2)]
...
```
Since in practice there can be a large number of start and stop time pairs. It would be better to dump as a binary format readable by python, but that would require significantly more code than the json implementation.

lroberts36: I guess you are suggesting to repeat keys to make it a 3-column table?

Yurlungur: Ah ok. I see.

Yurlungur: Nah I didn't think about it clearly enough. Your design is fine.

pdmullen (approved):

src/tasks/task_timing.hpp
lroberts36: I like having the descriptive error message and I think this isn't performance critical.

lroberts36:
> Looks like some of the new tests you added are failing. Otherwise LGTM.

@Yurlungur: Apparently `std::this_thread::sleep_for` actually should be called `std::this_thread::sleep_for_at_least`. I just removed the upper limit checks from the unit test since its primary purpose is to show how to time task lists.

lroberts36:
@par-hermes format

Yurlungur (approved):

PR #1339, opened 2025-11-22
[Trivial] Correct documentation and usage of numlevel

adamdempsey90:
## PR Summary

`numlevel` is only read and used with AMR, but the docstring reads as if you need it with SMR. This has caused some confusion, so I'm correcting it.

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build


adamdempsey90:
@par-hermes format

Yurlungur (approved):

Yurlungur:
Thanks for the fix!

pgrete (commented):
https://github.com/parthenon-hpc-lab/parthenon/blob/1e9a566abaf6fbd38ba3f55dac7464de0c7fb5c7/src/mesh/mesh.cpp#L1128 inside a
`if (pib->block_name.compare(0, 27, "parthenon/static_refinement") == 0) {`
conditional leads me to believe that we are using the max level info also for static mesh refinement.
I also think that this makes from a logical point of view in cases where I want to have some regions statically refined and other adaptively.
Overall, there should be a global max level (which is the current parameter), and then there's the level for the static regions individually as well as for the adaptive criteria.

adamdempsey90:
> inside a
> `if (pib->block_name.compare(0, 27, "parthenon/static_refinement") == 0) {`
> conditional leads me to believe that we are using the max level info also for static mesh refinement.


`max_level_ref_` is hard coded to 63 when `refinement == static`. Also, if you do set `numlevel` for SMR, you'll get a warning that the parameter is never used.

> I also think that this makes from a logical point of view in cases where I want to have some regions statically refined and other adaptively.

AMR + `static_refinement` blocks currently does not work (I would love for it to though).

adamdempsey90:
> > inside a
> > `if (pib->block_name.compare(0, 27, "parthenon/static_refinement") == 0) {`
> > conditional leads me to believe that we are using the max level info also for static mesh refinement.
> 
> `max_level_ref_` is hard coded to 63 when `refinement == static`. Also, if you do set `numlevel` for SMR, you'll get a warning that the parameter is never used.
> 
> > I also think that this makes from a logical point of view in cases where I want to have some regions statically refined and other adaptively.
> 
> AMR + `static_refinement` blocks currently does not work (I would love for it to though).

This might just be an artemis thing though.

tst/regression/test_suites/advection_convergence/parthinput.advection
```diff
@@ -20,7 +20,6 @@ problem_id = advection
 
 <parthenon/mesh>
 refinement = none
-numlevel = 2
```
pgrete, line 23: AMR is done for this test case (enabled via command line parameters for some steps).
So this needs to either stay or added to the test cases with AMR in `advection_convergence.py`

src/mesh/mesh.cpp
```diff
@@ -211,8 +212,9 @@ Mesh::Mesh(ParameterInput *pin, ApplicationInput *app_in, Packages_t &packages,
 
   // SMR / AMR:
   if (adaptive) {
-    max_level_ref_ = pin->GetOrAddInteger("parthenon/mesh", "numlevel", 1,
-                                          "maximum level of refinement globally");
+    max_level_ref_ =
+        pin->GetOrAddInteger("parthenon/mesh", "numlevel", 1,
+                             "maximum level of refinement globally when AMR is on");
```
pgrete, line 217: This change might need a rebuilt of the autogenerated parameter docs as the string changed.

pgrete (changes requested):
Thanks for pointing out the default value for static meshes.

At some point point, allowing both SMR and AMR would be nice.

Yurlungur:
@adamdempsey90 what's the status of this MR?

tst/regression/test_suites/advection_convergence/parthinput.advection
pgrete: I now fixed this.

pgrete (approved):

PR #1344, opened 2025-12-04
Remove unecessary solver communication

lroberts36:
## PR Summary
Add `Metadata::CommunicateOne` for only communicating a single layer of ghosts cells and cut out communications between coarse blocks in the two-level composite grid.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [ ] Docs build

### DIFF
--- a/src/utils/loop_utils.hpp
+++ b/src/utils/loop_utils.hpp
@@ -77,6 +77,7 @@ inline auto func_caller(F func, Args &&...args) -> typename std::enable_if<
 // need to be a template parameter, it could just be a function argument]
 template <BoundaryType bound = BoundaryType::any, class F>
 inline void ForEachBoundary(std::shared_ptr<MeshData<Real>> &md, F func) {
+  int fine_level = md->grid.logical_level;
   for (int block = 0; block < md->NumBlocks(); ++block) {
     auto &rc = md->GetBlockData(block);
     auto pmb = rc->GetBlockPointer();
@@ -125,8 +126,10 @@ inline void ForEachBoundary(std::shared_ptr<MeshData<Real>> &md, F func) {
       } else if constexpr (bound == BoundaryType::gmg_same) {
         if (v->IsSet(Metadata::FillGhost)) {
           for (auto &nb : *gmg_same) {
-            if (func_caller(func, pmb, rc, nb, v) == LoopControl::break_out) {
-              return;
+            if (pmb->loc.level() == fine_level || nb.loc.level() == fine_level) {
+              if (func_caller(func, pmb, rc, nb, v) == LoopControl::break_out) {
+                return;
+              }
             }
           }
         }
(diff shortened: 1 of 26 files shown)


lroberts36:
@par-hermes format

src/utils/loop_utils.hpp
```diff
@@ -125,8 +126,10 @@ inline void ForEachBoundary(std::shared_ptr<MeshData<Real>> &md, F func) {
       } else if constexpr (bound == BoundaryType::gmg_same) {
         if (v->IsSet(Metadata::FillGhost)) {
           for (auto &nb : *gmg_same) {
-            if (func_caller(func, pmb, rc, nb, v) == LoopControl::break_out) {
-              return;
+            if (pmb->loc.level() == fine_level || nb.loc.level() == fine_level) {
```
pgrete, line 129: Just double checking:
Given that `fine_level` above it taken from the`MeshData`, is that `MeshData` being passed in constructed in a way that it only contains blocks at the same level, and, thus, this new check ensures that the ghost buffer (of either the receiving block or sending neighbor) is also at the same level of the `MeshData` object?
Or does `fine_level` from `MeshData` contain the "global" finest level and `MeshData` contains blocks from arb. levels?
In the latter case, I'm not sure I see how this change is tied to "same level" boundaries (unless the "same level" only applies to the finest level and not at each level individually).

pgrete (approved):
LGTM (just some clarification question as I'm [still] not too familiar with the MG internals).

src/utils/loop_utils.hpp
lroberts36: The `MeshData` object knows what type of grid the blocks in it come from (from the member `MeshData::grid`, which is a `GridIdentifier` object). If it is a two-level composite grid, the `MeshData` will contain (some partition of) all the leaf blocks at `MeshData::grid.logical_level` of the grid (which is what `fine_level` is referring to), all the blocks in the interior of the tree at `fine_level`, and all leaf blocks at the next coarser level (`fine_level - 1`). The coarser level blocks only serve as boundary conditions for the finer level blocks and are not updated during smoothing operations on a two-level composite grid. Therefore, only boundaries where at least one of the blocks is on the `fine_level` need to be communicated. Obviously, it doesn't hurt to communicate between blocks on the coarser level, but it is potentially a performance hit.

pdmullen (approved):
LGTM, but quick question:

If I wanted the solvers to under-the-hood only communicate a single layer of ghosts, but then after coming out of the solve I want the solution to be available over `IndexDomain::entire`, is that an impossibility as things stand in this PR?

lroberts36:
> If I wanted the solvers to under-the-hood only communicate a single layer of ghosts, but then after coming out of the solve I want the solution to be available over `IndexDomain::entire`, is that an impossibility as things stand in this PR?

Yes, that is impossible with this PR. It might take a little bit of plumbing to get what you want to be available, I would suggest leaving it for later (given that you can get your desired behavior by just not including the metadata flag, at the cost of a little performance).

PR #1345, opened 2025-12-04
Coalesce and speed up dot products in solvers

lroberts36:
## PR Summary

Move variable size check outside of loop and allow for combined reductions as used in BiCGSTAB.

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build

### DIFF
--- a/src/solvers/solver_utils.hpp
+++ b/src/solvers/solver_utils.hpp
@@ -514,16 +514,13 @@ TaskStatus DotProductLocal(const std::shared_ptr<MeshData<Real>> &md_a,
   auto pack_a = desc.GetPack(md_a.get());
   auto pack_b = desc.GetPack(md_b.get());
   Real gsum(0);
+  const int nvars = pack_a.GetUpperBoundHost(0) - pack_a.GetLowerBoundHost(0) + 1;
   parthenon::par_reduce(
       parthenon::loop_pattern_mdrange_tag, "DotProduct", DevExecSpace(), 0,
-      pack_a.GetNBlocks() - 1, kb.s, kb.e, jb.s, jb.e, ib.s, ib.e,
-      KOKKOS_LAMBDA(const int b, const int k, const int j, const int i, Real &lsum) {
-        const int nvars = pack_a.GetUpperBound(b) - pack_a.GetLowerBound(b) + 1;
-        // TODO(LFR): If this becomes a bottleneck, exploit hierarchical parallelism and
-        //            pull the loop over vars outside of the innermost loop to promote
-        //            vectorization.
-        for (int c = 0; c < nvars; ++c)
-          lsum += pack_a(b, te, c, k, j, i) * pack_b(b, te, c, k, j, i);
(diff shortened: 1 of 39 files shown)


src/solvers/solver_utils.hpp
```diff
@@ -552,10 +549,112 @@ TaskID DotProduct(TaskID dependency_in, TaskList &tl, AllReduce<Real> *adotb,
...
+
+  value_type &operator[](std::size_t i) { return data_[i]; }
+  const value_type &operator[](std::size_t i) const { return data_[i]; }
+
+  // Contiguous container requirements
+  KOKKOS_INLINE_FUNCTION
+  std::size_t size() const { return N; }
+
+  value_type *data() { return data_; }
+};
```
lroberts36: Maybe this should live somewhere else?

pdmullen (approved):

src/solvers/solver_utils.hpp
Yurlungur: `utils/summable_array.hpp`?

Yurlungur (approved):
LGTM.

src/utils/summable_array.hpp
```diff
@@ -0,0 +1,69 @@
...
+} // namespace parthenon
+
+namespace Kokkos { // reduction identity must be defined in Kokkos namespace
+template <>
+struct reduction_identity<parthenon::summable_array_t<parthenon::Real, 2>> {
+  KOKKOS_FORCEINLINE_FUNCTION static parthenon::summable_array_t<parthenon::Real, 2>
+  sum() {
+    return parthenon::summable_array_t<parthenon::Real, 2>();
+  }
+};
```
pgrete, line 68: This mean the currently supported arrays dims in Parthenon are limited to 2, correct?
Do we want this to be more flexible for downstream use, i.e., provide default implementations for dims up to some `N`?
Alternatively (or on top), it'd be good to also add to the docs that this is available now (and downstream codes would use it either directly or by providing a custom impl for desired `N`). The latter could also be combined/demonstrated with a unit test.

pgrete (commented):
@lroberts36 do you want to fix the merge conflict (so that I don't mess it up)?

lroberts36:
@par-hermes format

pgrete:
Looks like legit compile errors -- just not sure why they show up now.

Also the warning
```
/__w/parthenon/parthenon/example/poisson_gmg/poisson_equation.hpp:122:35: warning: capture host side class data member by this pointer in device or host device lambda function may result in invalid memory access if this pointer is not accessible on device side [-Wgpu-maybe-wrong-side]
  122 |             auto [idx2p, idx2m] = GetEffectiveInverseDx2<X1DIR>(coords, k, j, i);
```
may be relevant.

lroberts36:
@pgrete: Yeah, I think this is real. I am hoping to fix it later today.

src/utils/summable_array.hpp
lroberts36: This is now updated to the reduction identity being templated on the size and type of the array so that it works for anything. I also added a unit test of the summable array.

lroberts36:
@par-hermes format

PR #1346, opened 2025-12-04
Allow for user defined restriction between multigrid levels

lroberts36:
## PR Summary

This PR just adds the option to override the default restriction machinery when communicating. In partitular, this is useful for communicating between multigrid levels. Doing the restriction in a unified kernel, as shown in `RestrictionCombined`, seems to be significantly more performant than the default machinery when restricting over the entire block.

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build

### DIFF
--- a/example/poisson_gmg/poisson_package.cpp
+++ b/example/poisson_gmg/poisson_package.cpp
@@ -96,7 +96,8 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
   std::shared_ptr<parthenon::solvers::SolverBase> psolver;
   using prolongator_t = parthenon::solvers::ProlongationBlockInteriorZeroDirichlet;
   using restrictor_t = parthenon::solvers::RestrictionCombined;
-  using preconditioner_t = parthenon::solvers::MGSolver<PoissEq, prolongator_t, restrictor_t>;
+  using preconditioner_t =
+      parthenon::solvers::MGSolver<PoissEq, prolongator_t, restrictor_t>;
   if (solver == "MG") {
     psolver = std::make_shared<parthenon::solvers::MGSolver<PoissEq, prolongator_t>>(
         "base", "u", "rhs", pin, "poisson/solver_params", PoissEq(pin, "poisson"));
--- a/example/poisson_gmg/poisson_package.cpp
+++ b/example/poisson_gmg/poisson_package.cpp
@@ -95,7 +95,8 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
 
   std::shared_ptr<parthenon::solvers::SolverBase> psolver;
   using prolongator_t = parthenon::solvers::ProlongationBlockInteriorZeroDirichlet;
-  using preconditioner_t = parthenon::solvers::MGSolver<PoissEq, prolongator_t>;
+  using restrictor_t = parthenon::solvers::RestrictionCombined;
+  using preconditioner_t = parthenon::solvers::MGSolver<PoissEq, prolongator_t, restrictor_t>;
   if (solver == "MG") {
     psolver = std::make_shared<parthenon::solvers::MGSolver<PoissEq, prolongator_t>>(
         "base", "u", "rhs", pin, "poisson/solver_params", PoissEq(pin, "poisson"));
--- a/src/bvals/comms/boundary_communication.cpp
+++ b/src/bvals/comms/boundary_communication.cpp
@@ -48,7 +48,8 @@ using namespace loops;
 using namespace loops::shorthands;
 
 template <BoundaryType bound_type>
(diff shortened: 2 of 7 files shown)


lroberts36:
@par-hermes format

Yurlungur:
@Yurlungur and @pdmullen to review

src/bvals/comms/boundary_communication.cpp
```diff
@@ -48,7 +48,8 @@ using namespace loops;
 using namespace loops::shorthands;
 
 template <BoundaryType bound_type>
-TaskStatus SendBoundBufs(std::shared_ptr<MeshData<Real>> &md) {
+TaskStatus SendBoundBufsWithRestrictOption(std::shared_ptr<MeshData<Real>> &md,
+                                           bool do_restriction) {
```
Yurlungur, line 52: Why the name change? Ah, because the tasking system doesn't support default arguments?

src/solvers/internal_prolongation.hpp
```diff
@@ -124,17 +124,11 @@ class ProlongationBlockInteriorZeroDirichlet {
 
     using TE = parthenon::TopologicalElement;
 
-    int nblocks = md->NumBlocks();
-    std::vector<bool> include_block(nblocks, true);
-    for (int b = 0; b < nblocks; ++b) {
-      include_block[b] =
-          md->grid.logical_level == md->GetBlockData(b)->GetBlockPointer()->loc.level();
-    }
```
Yurlungur, line 132: is this no longer necessary?

Yurlungur (approved):
I'm not going to object---we need this in. But at a high level, I wonder why we can't combine this with the custom prolongation/restriction machinery we have for mesh fields? It's a bit confusing to be able to set this in different ways depending on whether it's used in solvers or not.

example/poisson_gmg/poisson_package.cpp
```diff
@@ -95,7 +95,9 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
 
   std::shared_ptr<parthenon::solvers::SolverBase> psolver;
   using prolongator_t = parthenon::solvers::ProlongationBlockInteriorZeroDirichlet;
-  using preconditioner_t = parthenon::solvers::MGSolver<PoissEq, prolongator_t>;
+  using restrictor_t = parthenon::solvers::RestrictionCombined;
```
pdmullen, line 98: I was surprised to see this style of enrollment, given we typically invoke

```m.RegisterRefinementOps<MyProlong, MyRestrict>();```

for non-multigrid-specialized prolongation/restriction.

Do we want these custom multigrid restriction routines to be tied to fields and their Metadata and enrolled as above? or keep as is in this PR?

**Review by pdmullen — APPROVED** [review]

File: src/solvers/internal_prolongation.hpp [hunk 4]
```diff
@@ -208,6 +202,65 @@ class ProlongationBlockInteriorZeroDirichlet {
...
+
+// This is just an empty class without a Restrict method, which will trigger
+// calling the default restriction machinery in multigrid
+class RestrictionDefault {
+ public:
+  RestrictionDefault() = default;
+  RestrictionDefault(parthenon::ParameterInput *pin, const std::string &label) {}
+};
+
+class RestrictionCombined {
```
pdmullen, line 214: Can we leave a comment that this is just a demo of enrolling custom multigrid restriction?

And it does not, for example, work for all cases? e.g., curvilinear geometries.

example/poisson_gmg/poisson_package.cpp
Yurlungur: I had hte same thought.

src/bvals/comms/boundary_communication.cpp
lroberts36: yeah, that was the main reason

src/solvers/internal_prolongation.hpp
lroberts36: Yeah, this was redundant. Only selecting this set of blocks is the default behavior internally.

src/solvers/internal_prolongation.hpp
lroberts36, line 214:
```suggestion
class RestrictionCombined {
// This Restriction class is an example that is only correct for uniform cartesian coordinates
```

example/poisson_gmg/poisson_package.cpp
lroberts36: I would prefer to keep it tied to the solver. It is simpler and these are the restriction/prolongation operations between multigrid levels, which are really a separate thing from the boundary prolongation and restriction and prolongation and restriction during AMR operations.

src/solvers/internal_prolongation.hpp
lroberts36: added

PR #1349, opened 2025-12-10
[Trivial] thread font size and blocks through movie2d

Yurlungur:
## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build

### DIFF
--- a/scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
+++ b/scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
@@ -200,6 +200,13 @@ parser.add_argument(
     default=None,
 )
 parser.add_argument("--ylabel", type=str, help="Label for y axis", default=None)
+
+parser.add_argument(
+    "--blocks",
+    type=bool,
+    default=True,
+    help="Whether or not to plot block boundaries. Default is true.")
+
 parser.add_argument("field", type=str, help="field to plot")
 parser.add_argument("files", type=str, nargs="+", help="files to plot")
 
@@ -494,7 +501,7 @@ def main():
                         args.colorbar,
                         args.colorbounds,
                         args.symmetrizecolors,
-                        True,
+                        args.blocks,
                         data.gid,
                         data.xig,
                         data.yig,
@@ -525,7 +532,7 @@ def main():
                         args.colorbar,
                         args.colorbounds,
                         args.symmetrizecolors,
-                        True,
+                        args.blocks,
                         components=components,
                         swarmx=swarmx,
                         swarmy=swarmy,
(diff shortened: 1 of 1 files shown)


scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
```diff
@@ -334,7 +361,7 @@ def plot_dump(
                 0.5 * (xf[i, 0] + xf[i, -1]),
                 0.5 * (yf[i, 0] + yf[i, -1]),
                 str(block_ids[i]),
-                fontsize=8,
+                fontsize=int((3.0 / 4.0) * fontsize),
```
pgrete: This increases the default size from 8 to 9, doesn't it? Just checking this was intentional.

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
```diff
@@ -200,6 +200,28 @@ def maybe_float(string):
...
+    default=True,
+    action="store_true",
+    help="Whether or not to plot block boundaries. Default is true.",
+)
+parser.add_argument(
+    "--no-blocks",
+    dest="blocks",
+    action="store_false",
+    help="Whether or not to plot block boundaries. Default is true.",
+)
```
pgrete: I haven't come across this pattern yet and the resulting help text
```
  --blocks              Whether or not to plot block boundaries. Default is true.
  --no-blocks           Whether or not to plot block boundaries. Default is true.
```
confused me a little.

Are both cases really needed or is one argument enough. What happens if I provide both? Which one takes precedence?

In principle, I'd be in favor of a single argument only.
If both arguments are required for some reason, I'd be great to get more specific help messages, maybe sth like.
```
  --blocks              Plot block boundaries (default)
  --no-blocks           Disable plotting block boundaries.
```

pgrete (approved):
I'm already approving to not hold up this simple PR but I'm curious about those `--(no-)blocks` arguments.

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
Yurlungur: Not intentional. Fixed.

scripts/python/packages/parthenon_tools/parthenon_tools/movie2d.py
Yurlungur: I initially wanted `--blocks=true` or `--blocks=false` however, for whatever reason the way `argparse` parsed booleans seemed bizarre. But I think you're right we only need one switch. Now it's just

```python
parser.add_argument(
    "--no-blocks",
    dest="blocks",
    action="store_false",
    help="Disables the plotting of block boundaries. They are plotted by default.",
)
```

lroberts36 (approved):

PR #1356, opened 2025-12-18
make object pool map

Yurlungur:
## PR Summary

@swjones and I are going to use @lroberts36 object pools for more things and I found that the map of pools based on shapes is a useful construct, so I am factoring it out of mesh so I can reuse it. This small MR does that.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/doc/sphinx/src/boundary_communication.rst
+++ b/doc/sphinx/src/boundary_communication.rst
@@ -302,6 +302,44 @@ On host a lot of this functionality could be replicated with
 be able to exist on device (even though the reference counting doesn’t
 work there).
 
+It is often convenient to hold multiple pools, one for each different
+size of object that we might want to pool. (For example for
+differently shaped comm buffers.) We implement that via the
+``ObjectPoolMap<T>`` type, which owns an arbitrary number of
+pools. The main functionality that it provides is it can construct a
+pool with an allocation strategy that allocates buffers in chunks of
+contiguous blocks of memory, reducing the number of malloc calls
+required. For example, the code
+
+.. code:: cpp
+
+  ObjectPoolMap<ParArray1D<Real>> pool_map;
+  pool_map.AddPool(buf_len, nbufs);
+
+tells the pool map that you would like a pool of objects of size
+``buf_len`` that allocates ``nbufs`` at once when it runs out of
+memory, minimizing allocation costs. Objects can be explicitly added
+to the pool with
+
+.. code:: cpp
+
+  pool_map.AddFreeObjectsToPool(buf_len, nbufs);
+
+A pool for objects of a given size may be requested with
+``pool_map.GetPool(buf_len)``, a ``weak_t`` buffer may be requested
+with ``pool_map.GetBuffer(buf_len)`` and an ``owning_t`` may be
+requested with ``GetOwningBuffer``. The existence of a pool for
+objects may be checked with ``pool_map.Contains(buf_len)`` and all
+pools may be cleared and their buffers deallocated with
+``pool_map.Clear()``. Finally, the underlying map may be accessed for,
+e.g., iteration, with ``pool_map.GetMap()``.
+
+.. note::
+
+   The pool map is bounds checking. It will not add a new pool of
+   objects of a given shape unless they are requested. If you request
+   a buffer from a non-existent pool, the code will throw an error.
+
 Sparse boundary communication implementation
 --------------------------------------------
 
--- a/src/utils/object_pool.hpp
+++ b/src/utils/object_pool.hpp
@@ -275,6 +275,11 @@ bool UsingSameResource(const T &lhs, const U &rhs) {
   return lhs.GetKey() == rhs.GetKey();
 }
 
+/*
+  TODO(JMM): Currently the key type here is always size_t. It can
+  easily be extended to vector types such as std::tuple if needed,
+  but I didn't bother since we don't need that right now.
+ */
 template <typename T>
 class ObjectPoolMap {
  public:
--- a/src/utils/object_pool.hpp
+++ b/src/utils/object_pool.hpp
@@ -273,6 +273,51 @@ bool UsingSameResource(const T &lhs, const U &rhs) {
   return lhs.GetKey() == rhs.GetKey();
 }
 
+template <typename T>
+class ObjectPoolMap {
+ public:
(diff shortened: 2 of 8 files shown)


doc/sphinx/src/boundary_communication.rst
```diff
@@ -302,6 +302,44 @@ On host a lot of this functionality could be replicated with
 be able to exist on device (even though the reference counting doesn’t
 work there).
 
+It is often convenient to hold multiple pools, one for each different
+size of object that we might want to pool. (For example for
+differently shaped comm buffers.) We implement that via the
+``ObjectPoolMap<T>`` type, which owns an arbitrary number of
```
lroberts36, line 308: Have you considered `ObjectWaterPark<T>`?

src/utils/object_pool.hpp
```diff
@@ -273,6 +275,74 @@ bool UsingSameResource(const T &lhs, const U &rhs) {
...
+      auto label = "pool buffer of size " + std::to_string(shape) + " x " +
+                   std::to_string(chunk_size);
+      T chunk(label, pool_size);
+      for (std::size_t i = 1; i < chunk_size; ++i) {
+        pool->AddFreeObjectToPool(T(chunk, std::make_pair(i * shape, (i + 1) * shape)));
+      }
+      return T(chunk, std::make_pair(static_cast<std::size_t>(0), shape));
+    };
+    map_.emplace(shape, ObjectPool<T>(allocation_strategy));
+  }
```
lroberts36, line 335: I think this allocation strategy is pretty specific to a `BufArray1D`, so maybe we should limit what `T` can be (or at least add a comment that `T` is pretty constrained).

lroberts36 (approved):
This LGTM, only one real comment below.

src/utils/object_pool.hpp
Yurlungur: Yeah I agree. I will think about how to best enforce this.

Yurlungur: Done. I added a concept that this implements a view and I added a static assert specifically in the allocation strategy methods. Down the line maybe we can generalize.

Yurlungur:
@pgrete and/or @swjones please take a look when you get the chance

Yurlungur:
Reminder to @pgrete and @swjones please review.

src/utils/object_pool.hpp
```diff
@@ -1,5 +1,5 @@
 //========================================================================================
-// Parthenon performance portable AMR framework
+// parthenon performance portable AMR framework
```
pgrete: On purpose or did this happen when updating the copyright year (which should now be 2026)?

src/utils/object_pool.hpp
```diff
@@ -273,6 +275,87 @@ bool UsingSameResource(const T &lhs, const U &rhs) {
...
+        std::is_pointer_v<std::remove_reference_t<data_t>> &&
+            !std::is_pointer_v<std::remove_pointer_t<std::remove_reference_t<data_t>>>,
+        "Underlying view must be 1D");
+    if (map_.count(shape) > 0) return;
+    // This lambda is called whenever a buffer is requested but no
+    // buffers remain in the pool
+    auto allocation_strategy = [shape, chunk_size](ObjectPool<T> *pool) {
+      const auto pool_size = shape * chunk_size;
+      auto label = "pool buffer of size " + std::to_string(shape) + " x " +
+                   std::to_string(chunk_size);
```
pgrete: Very nice. I wonder if it makes sense (foreshadowing some debugging here) to have an internal, `private` counter in the map object that is used as additional info in the label so that the labels are unique.
A single counter for all `shape`s should be sufficient for unique names.

On a different topic: are our pools thread safe? Not that it matters right now, but might become unknowingly relevant at some point.

src/utils/object_pool.hpp
```diff
@@ -273,6 +275,87 @@ bool UsingSameResource(const T &lhs, const U &rhs) {
...
+    };
+    map_.emplace(shape, ObjectPool<T>(allocation_strategy));
+  }
+  // TODO(JMM): This assumes the pool is of a 1D Kokkos view-like object
+  void AddFreeObjectsToPool(const std::size_t shape, const std::size_t nobjs) {
+    static_assert(
+        std::is_pointer_v<std::remove_reference_t<data_t>> &&
+            !std::is_pointer_v<std::remove_pointer_t<std::remove_reference_t<data_t>>>,
+        "Underlying view must be 1D");
+    auto &pool = map_.at(shape);
```
pgrete: Should there also be a 
```
if (!Contains(shape)) {
```
as above or do we rely on the error handling from `.at()`?

tst/unit/test_object_pool.cpp
```diff
@@ -0,0 +1,75 @@
...
+#include "utils/object_pool.hpp"
+
+using parthenon::Real;
+using array_t = parthenon::ParArray1DRaw<Real>;
+using pool_t = parthenon::ObjectPool<array_t>;
+using pool_map_t = parthenon::ObjectPoolMap<array_t>;
+
+SCENARIO("Object pools", "[ObjectPool]") {
+  GIVEN("An object pool map containing several sizes") {
+    const std::size_t NSIZES = 2;
```
pgrete, line 35: 👍 for test unit test

pgrete (approved):
LGTM

src/utils/object_pool.hpp
Yurlungur: Not on purpose. Will change it.

src/utils/object_pool.hpp
Yurlungur: > I wonder if it makes sense (foreshadowing some debugging here) to have an internal, private counter in the map object that is used as additional info in the label so that the labels are unique.

This is a good idea...

> On a different topic: are our pools thread safe?

I don't know, and I suspect not, because I don't think `std::map` is thread safe. But if we added an internal counter they also wouldn't be.

Yurlungur: I added an internal counter as a static variable in the lambda, which automatically makes it a counter per combination of shape/chunk size. Probably no more "not thread safe" than anything else in this machinery but something to be aware of.

src/utils/object_pool.hpp
Yurlungur: Actually we should just call `GetPool`. That's better.

PR #1360, opened 2026-01-09
Fix comm buffer init for mesh data

### LINKED ISSUE #1359: "out of bounds" ("pool buffer" related?) access on Mesh init
Here's a smallish input file

```
<parthenon/job>
problem_id = advection

<parthenon/mesh>
refinement = adaptive
numlevel = 2

nx1 = 64
x1min = -0.5
x1max = 0.5
ix1_bc = periodic
ox1_bc = periodic

nx2 = 64
x2min = -0.5
x2max = 0.5
ix2_bc = periodic
ox2_bc = periodic

nx3 = 1
x3min = -0.5
x3max = 0.5
ix3_bc = periodic
ox3_bc = periodic

pack_size = 1

<parthenon/meshblock>
nx1 = 16
nx2 = 16
nx3 = 1

<parthenon/time>
nlim = -1
tlim = 1.0
integrator = rk2
ncycle_out_mesh = -10000

<Advection>
cfl = 0.45
vx = 1.0
vy = 1.0
vz = 1.0
profile = hard_sphere

refine_tol = 0.3    # control the package specific refinement tagging function
derefine_tol = 0.03
compute_error = false
num_vars = 1 # number of variables
vec_size = 1 # size of each variable
fill_derived = false # whether to fill one-copy test vars

```
that triggers the issue using the `advection_example` (in compiled in debug mode)
```
:0: : block: [1,0,0], thread: [0,9,0] Assertion `Kokkos::View ERROR: out of bounds access label=("") with indices [18] but extents [18]` failed.
:0: : block: [0,0,0], thread: [0,9,0] Assertion `Kokkos::View ERROR: out of bounds access label=("") with indices [18] but extents [18]` failed.
:0: : block: [0,0,0], thread: [0,10,0] Assertion `Kokkos::View ERROR: out of bounds access label=("") with indices [20] but extents [18]` failed.
:0: : block: [0,0,0], thread: [0,11,0] Assertion `Kokkos::View ERROR: out of bounds access label=("") with indices [22] but extents [18]` failed.
:0: : block: [0,0,0], thread: [0,12,0] Assertion `Kokkos::View ERROR: out of bounds access label=("") with indices [24] but extents [18]` failed.
```

In AthenaPK (where I originally stumbled across this) the access error is more specific
```
:0: : block: [1,0,0], thread: [0,97,0] Assertion `Kokkos::View ERROR: out of bounds access label=("pool buffer") with indices [3104] but extents [1998]` failed.
```
pointing to the "pool buffer".
Other failure modes include 
```
### PARTHENON ERROR
  Message:     Receiving buffer is never available for write.
  File:        /workspaces/parthenon/src/utils/communication_buffer.hpp
  Line number: 313
```
Which one is shown is also intermittent (and, for example, depends on whether running in a debugger or not).

Backtraces sth along 
```
#6  0x0000555555981fa8 in parthenon::CommBuffer<parthenon::ObjectPool<Kokkos::View<double*, Kokkos::LayoutRight, Kokkos::CudaSpace> >::owner_t>::IsAvailableForWrite (this=0x55559093ed88) at /workspaces/parthenon/src/utils/communication_buffer.hpp:313
#7  0x000055555596ce83 in parthenon::CheckSendBufferCacheForRebuild<(parthenon::BoundaryType)2, true>(std::shared_ptr<parthenon::MeshData<double> >)::{lambda(auto:1, std::shared_ptr<parthenon::MeshBlockData<double> >, parthenon::NeighborBlock&, std::shared_ptr<parthenon::Variable<double> >)#1}::operator()<parthenon::MeshBlock*>(parthenon::MeshBlock*, std::shared_ptr<parthenon::MeshBlockData<double> >, parthenon::NeighborBlock&, std::shared_ptr<parthenon::Variable<double> >) const (__closure=0x7fffffffba80, pmb=0x55558fa5aad0, rc=..., nb=..., v=...) at /workspaces/parthenon/src/bvals/comms/bvals_utils.hpp:159
#8  0x000055555598238a in parthenon::loops::func_caller<parthenon::CheckSendBufferCacheForRebuild<(parthenon::BoundaryType)2, true>(std::shared_ptr<parthenon::MeshData<double> >)::{lambda(auto:1, std::shared_ptr<parthenon::MeshBlockData<double> >, parthenon::NeighborBlock&, std::shared_ptr<parthenon::Variable<double> >)#1}, parthenon::MeshBlock*&, std::shared_ptr<parthenon::MeshBlockData<double> >&, parthenon::NeighborBlock&, std::shared_ptr<parthenon::Variable<double> > cons--Type <RET> for more, q to quit, c to continue without paging--
t&>(parthenon::CheckSendBufferCacheForRebuild<(parthenon::BoundaryType)2, true>(std::shared_ptr<parthenon::MeshData<double> >)::{lambda(auto:1, std::shared_ptr<parthenon::MeshBlockData<double> >, parthenon::NeighborBlock&, std::shared_ptr<parthenon::Variable<double> >)#1}, parthenon::MeshBlock*&, std::shared_ptr<parthenon::MeshBlockData<double> >&, parthenon::NeighborBlock&, std::shared_ptr<parthenon::Variable<double> > const&) (func=...) at /workspaces/parthenon/src/utils/loop_utils.hpp:68
```
so failing on the host or 
```
#21 0x00005555560ff10e in parthenon::SendBoundBufs<(parthenon::BoundaryType)2> (md=...) at /workspaces/athenapk/external/parthenon/src/bvals/comms/boundary_communication.cpp:103
#22 0x000055555574e3b5 in parthenon::TaskList::AddUserTask<parthenon::TaskStatus (&)(std::shared_ptr<parthenon::MeshData<double> >&), std::shared_ptr<parthenon::MeshData<double> >&>(parthenon::TaskID&, std::optional<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > const&, int, parthenon::TaskStatus (&)(std::shared_ptr<parthenon::MeshData<double> >&), std::shared_ptr<parthenon::MeshData<double> >&)::{lambda()#1}::operator()() (__closure=0x5555c60edfe0) at /workspaces/athenapk/external/parthenon/src/tasks/tasks.hpp:470
```
so failing on device in this kernel https://github.com/parthenon-hpc-lab/parthenon/blob/01c61ebffcb388ca349d23ffe3fa692b93dc7b1e/src/bvals/comms/boundary_communication.cpp#L103

Summary of what I found so far:
- Can be triggered running even on a single rank
- Can be triggered on host and device builds
- It is intermittent (i.e., sometimes it happens and happens to various degrees, sometime it doesn't)
- When not compiled in debug mode (and by accident running past the of of bounds issue) surfaces at multilevel boundaries, e.g., 
Image

- It does not seem to happen with static mesh refinement (though I'm not certain given the intermittent nature)
- It can safely be triggered by using `pack_size=1` (but it more intermittent with `pack_size=-1`)
- Happens with and without coaslesced comms
- Given both failure modes (the out of bounds and proper `PARTHENON ERROR`, I suspect some issue in the boundary comm buffer handling, but it's not clear to me whether this is
  - an indexing issue when accessing the buffer
  - an indexing/logical issue when creating/assigning the buffers
  - some race condition with regard to creating/using the buffer
  - or something completely else...

Any additional eyes/input would be appreciated.

pgrete:
## PR Summary

This fixes the issue seen in #1359 *but I'm not sure if there isn't sth wrong more fundamentally*.
Thus, additional comments/eyes on the code around the caching of comm buffer and `MeshData` would be great, particularly looking at @lroberts36 and @Yurlungur 

As far as I was able to follow the issue, it stems from calling `BuildBoundaryBuffers(mesh_data.Add("base", GetBasePartition()));` with the "base partition",  which contains *all* block, i.e, a single `MeshData` object.
In subsequent places (e.g., in the following `RegisterCoalescedComms`) or in the `BuildTagMapAndBoundaryBuffers`, the comm related routine are called via the default partitions, so there was a misalignment between those.

Does this mean our comm patterns are tied to `MeshData` itself and not to blocks (and then collected/summarized in `MeshData`)?
Is this what we want?
Are there other place that might be affected by different default use between "all blocks" and "partitioned blocks"?

For example, I noticed in the swarms
```c++
// Eliminate any cached SwarmPacks, as they will need to be rebuilt following SetPoolMax
pmb->meshblock_data.Get()->ClearSwarmCaches();
pm->mesh_data.Add("base", pm->GetBasePartition())->ClearSwarmCaches();
for (auto &partition : pm->GetDefaultBlockPartitions()) {
  pm->mesh_data.Add("base", partition)->ClearSwarmCaches();
```
that the caches are cleaned separately.

Fixes #1359

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build

### DIFF
--- a/src/mesh/mesh.hpp
+++ b/src/mesh/mesh.hpp
@@ -265,22 +265,35 @@ class Mesh {
   using comm_buf_t = CommBuffer<buf_pool_t<Real>::owner_t>;
   std::unordered_map<int, buf_pool_t<Real>> pool_map;
 
-  struct comm_buf_map_t {
-    std::size_t epoch{0};
-    using map_t = std::unordered_map<channel_key_t, comm_buf_t>;
-    map_t m;
-    auto &operator[](const channel_key_t &k) { return m[k]; }
-    auto &at(const channel_key_t &k) { return m.at(k); }
-    auto count(const channel_key_t &k) const { return m.count(k); }
+  class comm_buf_map_t {
+   public: 
+    using map_t = std::unordered_map<channel_key_t, comm_buf_t>; 
+    using key_type = map_t::key_type;
+    using mapped_type = map_t::mapped_type;
+  
+   private:
+    // On initial meshing and after remeshing, the comm buffer map is cleared and
+    // rebuilt. The member epoch_ stores the number of times the comm buffers have
+    // been built so that various boundary cache objects that point to the comm 
+    // buffers can easily check if they point to buffers from an old mesh
+    // configuration that have been cleared.
+    std::size_t epoch_{1};
+    map_t m_;
 
-    auto begin() noexcept { return m.begin(); }
-    auto end() noexcept { return m.end(); }
-    auto begin() const noexcept { return m.begin(); }
-    auto end() const noexcept { return m.end(); }
+   public:
+    auto &operator[](const key_type &k) { return m_[k]; }
+    auto &at(const key_type &k) { return m_.at(k); }
+    auto count(const key_type &k) const { return m_.count(k); }
 
+    auto begin() noexcept { return m_.begin(); }
+    auto end() noexcept { return m_.end(); }
+    auto begin() const noexcept { return m_.begin(); }
+    auto end() const noexcept { return m_.end(); }
+    
+    auto GetCurrentEpoch() const {return epoch_; } 
     void clear() {
-      m.clear();
-      epoch++;
+      m_.clear();
+      epoch_++;
     }
   };
 
--- a/src/mesh/mesh.hpp
+++ b/src/mesh/mesh.hpp
@@ -264,7 +264,26 @@ class Mesh {
   using channel_key_t = std::tuple<int, int, std::string, int, int>;
   using comm_buf_t = CommBuffer<buf_pool_t<Real>::owner_t>;
   std::unordered_map<int, buf_pool_t<Real>> pool_map;
-  using comm_buf_map_t = std::unordered_map<channel_key_t, comm_buf_t>;
+
+  struct comm_buf_map_t {
+    std::size_t epoch{0};
+    using map_t = std::unordered_map<channel_key_t, comm_buf_t>;
+    map_t m;
+    auto &operator[](const channel_key_t &k) { return m[k]; }
+    auto &at(const channel_key_t &k) { return m.at(k); }
+    auto count(const channel_key_t &k) const { return m.count(k); }
+
+    auto begin() noexcept { return m.begin(); }
+    auto end() noexcept { return m.end(); }
+    auto begin() const noexcept { return m.begin(); }
+    auto end() const noexcept { return m.end(); }
+
+    void clear() {
+      m.clear();
+      epoch++;
+    }
+  };
+
   comm_buf_map_t boundary_comm_map;
   TagMap tag_map;
 
--- a/src/mesh/mesh.hpp
+++ b/src/mesh/mesh.hpp
@@ -266,15 +266,15 @@ class Mesh {
   std::unordered_map<int, buf_pool_t<Real>> pool_map;
 
   class comm_buf_map_t {
-   public: 
-    using map_t = std::unordered_map<channel_key_t, comm_buf_t>; 
+   public:
+    using map_t = std::unordered_map<channel_key_t, comm_buf_t>;
     using key_type = map_t::key_type;
     using mapped_type = map_t::mapped_type;
-  
+
    private:
     // On initial meshing and after remeshing, the comm buffer map is cleared and
     // rebuilt. The member epoch_ stores the number of times the comm buffers have
-    // been built so that various boundary cache objects that point to the comm 
+    // been built so that various boundary cache objects that point to the comm
     // buffers can easily check if they point to buffers from an old mesh
     // configuration that have been cleared.
     std::size_t epoch_{1};
@@ -289,8 +289,8 @@ class Mesh {
     auto end() noexcept { return m_.end(); }
     auto begin() const noexcept { return m_.begin(); }
     auto end() const noexcept { return m_.end(); }
-    
-    auto GetCurrentEpoch() const {return epoch_; } 
+
+    auto GetCurrentEpoch() const { return epoch_; }
     void clear() {
       m_.clear();
       epoch_++;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 - [[PR 1355]](https://github.com/parthenon-hpc-lab/parthenon/pull/1355) Allow disabling format and lint targets
 
 ### Fixed (not changing behavior/API/variables/...)
-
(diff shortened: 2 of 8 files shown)


Yurlungur (approved):
Oof... I think we do want comm buffers associated with partitions, which are synonymous with mesh data but are not the base partitions. So this is probably a bug wherever the pattern is repeated.

lroberts36:
@pgrete, @Yurlungur: Ok, I think I understand what is going on here. `BuildBoundaryBuffers(md)` actually does two things:

 *First*, it builds the boundary buffers for all the blocks contained in `md`. This piece should build all of the required buffers whether it is given one partition that contains all the blocks or if it is called on multiple partitions, the union of which contains all the blocks on a given rank. The buffer pool logic that @Yurlungur added recently is written with the base partition in mind to give a better guess at the number of boundary buffers that are required. It should still work with any partitioning though, just not be optimal. 

*Second*, it clears all of the boundary caches in the `MeshData` it is passed. This is where the bug comes in. Since we moved to doing `BuildBoundaryBuffers` on the base partition, we were never calling `MeshData::GetBVarsCache().clear()` on the default block partitions. Most of the time this probably was fine because all of the boundary caches were stale after a remesh and things got rebuilt, but there are probably some edge cases where that isn't the case.

I made some changes to the PR to explicitly check that we don't need to reinitialize the caches in the boundary comm routines and also to clear the caches in the default block partitions.

lroberts36:
@par-hermes format

lroberts36:
I ended up removing the explicit clearing of the cache, since we can just check in place now.

lroberts36:
@par-hermes format

Yurlungur:
this is a cleaner solution. Though I think the `epoch` field needs a comment describing its intent, as it's not obvious without a bigger view of the code base.

lroberts36:
@par-hermes format

src/mesh/mesh.hpp
```diff
@@ -264,7 +264,39 @@ class Mesh {
...
+    using key_type = map_t::key_type;
+    using mapped_type = map_t::mapped_type;
+
+   private:
+    // On initial meshing and after remeshing, the comm buffer map is cleared and
+    // rebuilt. The member epoch_ stores the number of times the comm buffers have
+    // been built so that various boundary cache objects that point to the comm
+    // buffers can easily check if they point to buffers from an old mesh
+    // configuration that have been cleared.
+    std::size_t epoch_{1};
```
pgrete, line 278: Why the different default value from the one used in `BvarsSubCache_t` above?

pgrete (commented):
I cannot approve (as I opened the PR).
@lroberts36 @Yurlungur if you approve we could merge.

Yurlungur (approved):
with the merge conflict now resolved, I approve

src/mesh/mesh.hpp
lroberts36: Because we want them to not be consistent by default, but only after `InitializeBufferCache` has been successfully called.

lroberts36 (approved):
Approving to get this in.

PR #1361, opened 2026-01-09
Bump formatters to clang-format-20 and black 25.12

pgrete:
## PR Summary

In addition to an outdated `clang-format` (original version 11) we now also faced issue with black and python files in AthenaPK.
That sparked an update of the default versions to sth more recent (clang-format 18 is now 2+ years old).
~IIRC the formatting changes are now robust across many clang-format version~ so that installing a local "old" version should not be required any more for development.
[edit]
Changes seem to be less robust.
clang-format 16/17 are different from 18/19 and 20/21 again agree with 16/17.
[/edit]
However, there might be an impact for downstream codes and their tooling/formatting.

Therefore, I'm specifically asking for approval/feedback from many downstream codes.

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build

### DIFF
--- a/.github/workflows/apply-formatting.yml
+++ b/.github/workflows/apply-formatting.yml
@@ -8,8 +8,8 @@ jobs:
     runs-on: ubuntu-latest
     if: startsWith(github.event.comment.body, '@par-hermes format')
     steps:
-    - uses: pgrete/cpp-py-formatter/command@v0.3.0
+    - uses: pgrete/cpp-py-formatter/command@v0.4.0
       with:
         botName: par-hermes
-        clangFormatVersion: 11
+        clangFormatVersion: 18
         token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/apply-formatting.yml
+++ b/.github/workflows/apply-formatting.yml
@@ -11,5 +11,5 @@ jobs:
     - uses: pgrete/cpp-py-formatter/command@v0.4.0
       with:
         botName: par-hermes
-        clangFormatVersion: 18
+        clangFormatVersion: 20
         token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/check-formatting.yml
+++ b/.github/workflows/check-formatting.yml
@@ -5,7 +5,7 @@ jobs:
     name: Check Python and C++ Formatting
     runs-on: ubuntu-latest
     steps:
-    - uses: pgrete/cpp-py-formatter/check@v0.3.0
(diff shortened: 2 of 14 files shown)


adamdempsey90 (approved):
artemis is already on these versions, so this is fine with me.

pgrete:
> artemis is already on these versions, so this is fine with me.

Which ones are you using?

And which ones other downstream codes?

I'm asking as I just noticed that changes seem to be less robust. clang-format 16/17 are different from 18/19. 20/21 again agree with 16/17. Kokkos on 16 as far as I can tell, so given that 16 agrees with the more recent version I might be inclined to skip 18/19 (and either do 17, which is older but also default/present in ubuntu 24 or directly 20, which is newer).

Yurlungur (approved):
Long overdue. I'd be supportive of a newer clang format even

Yurlungur:
I think Spiner and singularity-eos are on 19

adamdempsey90:
> > artemis is already on these versions, so this is fine with me.
> 
> Which ones are you using?
> 
> And which ones other downstream codes?
> 
> I'm asking as I just noticed that changes seem to be less robust. clang-format 16/17 are different from 18/19. 20/21 again agree with 16/17. Kokkos on 16 as far as I can tell, so given that 16 agrees with the more recent version I might be inclined to skip 18/19 (and either do 17, which is older but also default/present in ubuntu 24 or directly 20, which is newer).

For artemis, we have local python virtual environments that install clang-format 18.1.8 and the most recent black version at the time of venv creation (on the machine I checked that's 25.9). But, there shouldn't be any conflict with what parthenon chooses.

acreyes:
> For artemis, we have local python virtual environments that install clang-format 18.1.8 and the most recent black version at the time of venv creation (on the machine I checked that's 25.9). But, there shouldn't be any conflict with what parthenon chooses.

We do the same. With `uv` as the dependency manager we also track the lock file so all developers end up on the same versions. I think in a similar vein you can pin an exact version in the `pyproject.toml` ore requirements file so everyone gets the same versions reliably when making the venv.

We're on 20.1.0

AstroBarker:
Phoebus is on a rather old version still. If I were updating it right now I'd go 20+. I seem to recall weirdness with 18/19 as well. There is an option to dump the entire config (i.e., including defaults) to a file which I think has given me some forward stability (although certainly _not_ backwards stability as they add new config options...) in another project, but they may well find a way to break that too.

pgrete:
Thanks for all the quick comments.
Based on all those, it seems that a bump to version 20 is the way to go.

PR #1363, opened 2026-01-15
Generalize multigrid hierarchy

lroberts36:
## PR Summary

- Generalizes the geometric multigrid (GMG) hierarchy to support an initial
  sequence of **leaf-only GMG grids** before transitioning to **two-level
  composite grids**, rather than forcing an immediate move to composite grids.

- Introduces a new runtime parameter `parthenon/mesh/base_block_coarsenings` 
  which controls how many leaf GMG grids (with increasing in-block coarsening) are
  included at the start of the hierarchy.

- Refactors GMG grid identification to explicitly track:
  - grid type (`leaf` vs `two_level_composite`),
  - in-block coarsening level (`block_coarsenings`),
  - multigrid level index,
  - and whether a grid participates in the GMG hierarchy.

- Preserves and clarifies **two-level composite grid semantics**, where a composite 
  grid at logical level ℓ includes:
  - level ℓ blocks (leaf + internal parents of finer leaves), and
  - level ℓ−1 leaf blocks to provide boundary conditions.

- Updates GMG hierarchy construction logic to cleanly separate:
  - leaf-only GMG stages,
  - tree traversal using two-level composite grids,
  - and subsequent in-block coarsening at the root.

- Extends boundary communication infrastructure to allow **multiple 
  communication buffers per communication channel**, enabling concurrent
  communication on multiple `MeshData` stages without MPI lock contention.

  - Implements the above communication changes across:
    `tag_map.*`, `bvals_utils.hpp`, `mesh_data.hpp`, and `build_boundary_buffers.cpp`.

- Adds a more extensive **solver testing infrastructure** for GMG-based solvers.

---

## Things that are left for future PRs:

- Update **coalesced communication infrastructure** to correctly handle the new “multiple buffers per communication channel” capability. Currently, the code will error out if you try to use coalesced comms with multiple communication buffers. 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


Yurlungur:
since this depends on all those other MRs... the diff is hard to read. I think I'll wait to review until those are in

lroberts36:
> since this depends on all those other MRs... the diff is hard to read. I think I'll wait to review until those are in

@Yurlungur: Yeah, no need to review while it is in WIP. I am still working on it and I put it up mostly because I wanted to easily be able to see the diff.

tst/unit/test_index_split.cpp
```diff
@@ -100,7 +100,7 @@ TEST_CASE("IndexSplit", "[IndexSplit]") {
     BlockList_t block_list = MakeBlockList(pkg, NBLOCKS, N, NDIM);
 
     MeshData<Real> mesh_data("base");
-    mesh_data.Initialize(block_list, nullptr, NDIM);
+    mesh_data.Initialize(block_list, nullptr);
```
lroberts36, line 103: This was just a previously existing bug where the `NDIM` being passed here was assumed to be the multigrid level.

Yurlungur:
@Yurlungur and @pdmullen to review

src/bvals/comms/bnd_id.cpp
```diff
@@ -43,7 +43,8 @@ namespace parthenon {
 BndId BndId::GetSend(MeshBlock *pmb, const NeighborBlock &nb,
                      std::shared_ptr<Variable<Real>> v, BoundaryType b_type,
                      int partition, int start_idx) {
-  auto [send_gid, recv_gid, vlabel, loc, extra_id] = SendKey(pmb, nb, v, b_type);
+  // TODO: This needs to be fixed for unique buffer ids
```
Yurlungur: reminder

src/bvals/comms/bnd_info.cpp
```diff
@@ -107,6 +107,18 @@ GetFluxCorrectionElements(const std::shared_ptr<Variable<Real>> &v,
...
+  return nb.loc.level() < pmb->loc.level() || nb.block_coarsenings > pmb->block_coarsenings; 
+}
+
+bool NeighborIsFiner(MeshBlock *pmb, const NeighborBlock &nb) { 
+  return nb.loc.level() > pmb->loc.level() || nb.block_coarsenings < pmb->block_coarsenings; 
+}
+
+bool NeighborIsSame(MeshBlock *pmb, const NeighborBlock &nb) { 
+  return nb.loc.level() == pmb->loc.level() && nb.block_coarsenings == pmb->block_coarsenings;
+}
```
Yurlungur: In general, an elegant way to plug the new coarse level stuff in with the block infrastructure logic.

src/bvals/comms/build_boundary_buffers.cpp
```diff
@@ -116,21 +114,27 @@ void BuildBoundaryBufferSubset(std::shared_ptr<MeshData<Real>> &md,
...
-            tag, sender_rank, receiver_rank, comm, get_resource_method,
-            use_sparse_buffers);
+      for (int id = 0; id <= (BTYPE == BoundaryType::gmg_restrict_send); ++id) {
+        auto s_key = SendKey(pmb, nb, v, BTYPE, id);
+        const int tag = pmesh->tag_map.GetTag(pmb, nb, id);
+        if (buf_map.count(s_key) == 0)
+          buf_map[s_key] = CommBuffer<buf_pool_t<Real>::owner_t>(
+              tag, sender_rank, receiver_rank, comm, get_resource_method,
+              use_sparse_buffers);
+      }
```
Yurlungur: Is this change because a block has one buffer per boundary per coarsened level potentially?

src/interface/mesh_data.hpp
```diff
@@ -197,7 +197,18 @@ class MeshData {
...
+  int partition{-1};
+
+  void SetBoundBufferId(BoundaryType btype, int id) { 
+    bound_buffer_ids_[btype] = id;
+  }
+  
+  int GetBoundBufferId(BoundaryType btype) const { 
+    if (bound_buffer_ids_.count(btype)) return bound_buffer_ids_.at(btype);
+    return 0;
+  }
```
Yurlungur: This strikes me as a bit odd... is 0 always the correct default here? Can this not hide a bug in the base infrastructure if, for example, a buffer isn't constructed properly?

src/bvals/comms/bvals_utils.hpp
```diff
@@ -41,30 +41,36 @@ inline auto &GetSenderGid(Mesh::channel_key_t &key) { return std::get<0>(key); }
...
-                                   BoundaryType btype) {
+                                   BoundaryType btype, int id) {
   const int sender_id = pmb->gid;
   const int receiver_id = nb.gid;
   const int location_idx = nb.offsets.GetIdx();
   int other = (nb.gid == pmb->gid && (btype == BoundaryType::gmg_restrict_recv ||
                                       btype == BoundaryType::gmg_restrict_send))
                   ? 1
                   : 0;
+  other += 2 * id;
```
Yurlungur: why 2?

src/solvers/bicgstab_solver.hpp
```diff
@@ -162,7 +162,7 @@ class BiCGSTABSolver : public SolverBase, BiCGSTABSolverCounter {
 
     iter_counter = 0;
     bool multilevel = pmesh->multilevel;
-
+    auto timer_guard_total = TimingAccumulatorGuard(solver_timings.GetOrAddAndRegister("BiCGSTAB: Total", tl));
```
Yurlungur: there seems to be some timing stuff mixed in to this MR. Is that part of another MR or did you just sneak it in?

src/mesh/meshblock.hpp
```diff
@@ -432,12 +433,16 @@ class MeshBlock : public std::enable_shared_from_this<MeshBlock> {
                                  kb.s, kb.e, jb.s, jb.e, ib.s, ib.e, function,
                                  std::forward<Args>(args)...);
   }
+  
+  bool IsLeafLL() const { return is_leaf_ll_; }
```
Yurlungur, line 457: what does `LL` stand for?

src/solvers/solver_utils.hpp
```diff
@@ -521,7 +522,8 @@ TaskStatus DotProductLocal(const std::shared_ptr<MeshData<Real>> &md_a,
       pack_a.GetNBlocks() - 1, 0, nvars - 1, kb.s, kb.e, jb.s, jb.e, ib.s, ib.e,
       KOKKOS_LAMBDA(const int b, const int c, const int k, const int j, const int i,
                     Real &lsum) {
-        lsum += pack_a(b, te, c, k, j, i) * pack_b(b, te, c, k, j, i);
+        //const auto vol = pack_a.GetCoordinates(b).CellVolume(k, j, i);
+        lsum += pack_a(b, te, c, k, j, i) * pack_b(b, te, c, k, j, i);// * vol * vol;
```
Yurlungur: commented out because you found it's ok without?

src/utils/loop_utils.hpp
```diff
@@ -69,6 +69,31 @@ inline auto func_caller(F func, Args &&...args) -> typename std::enable_if<
   return LoopControl::cont;
 }
 
+inline auto &GetNeighborsOnCoarserGMGGrid(MeshBlock *pmb, const GridIdentifier &grid) { 
+  if (grid.type() == GridType::two_level_composite && 
+      pmb->loc.level() != grid.logical_level()) { 
+    // This is a boundary block on a two-level composite grid, its
+    // data is up to date but it needs to send a message to itself
+    // on the next coarser grid for synchronization
```
Yurlungur: I do not understand this comment.

src/mesh/mesh-gmg.cpp
```diff
@@ -85,90 +87,261 @@ void Mesh::SetMeshBlockNeighbors(
...
-    } else if (grid_id.type == GridType::two_level_composite &&
-               pmb->loc.level() == grid_id.logical_level - 1) {
+    } else if (grid_id.type() == GridType::two_level_composite &&
+               pmb->loc.level() == grid_id.logical_level() - 1) {
       pmb->gmg_composite_finer_neighbors = all_neighbors;
     }
   }
 }
 
+// Terminology and GridIdentifier conventions for Parthenon GMG on a forest-of-octrees AMR mesh
```
Yurlungur: Wish I'd looked here first lol. This is helpful, but I think it needs to be in the docs, not just in a comment.

Yurlungur (commented):
There must be something I'm missing here. I was expecting to see some data structure for the coarser levels internal to a block, but there doesn't seem to be anything. Do you just create blocks of different sizes, or the same block in the same position many times? What's happening?

Yurlungur:
> There must be something I'm missing here. I was expecting to see some data structure for the coarser levels internal to a block, but there doesn't seem to be anything. Do you just create blocks of different sizes, or the same block in the same position many times? What's happening?

Maybe we can discuss and go through the code together next week.

src/bvals/comms/build_boundary_buffers.cpp
lroberts36: This change creates two buffers per communication channel on the multigrid restriction/prolongation boundary type. This is required because when we communicate from a coarser to a finer MG level, we have to restrict and communicate the fine solution and residual, which are stored in different `MeshBlockData` stages. Previously, when we only allowed for two-level composite grids in the hierarchy, restriction from one multigrid level only communicated with the next coarser level. This meant we could just do two communications in sequence using the same buffers. Now, when you go from a grid that covers the whole mesh to two-level composite grids of internally coarsened buffers, restriction from the grid that covers the entire mesh concurrently sends data to many coarser MG levels. This means that the first communication on the fine grid doesn't finish until all of the coarser grids finish with their receives, which also rely on communication from intermediate grids. If you just try to reuse buffers and wait until they are free, such a setup will just hang. Going to two sets of buffers that can be in flight concurrently fixes this.

src/bvals/comms/bvals_utils.hpp
lroberts36: we are just combining multiple things into the tag. The first bit stores the switch between boundary types, the rest stores `id`.

src/solvers/bicgstab_solver.hpp
lroberts36: I just snuck this in here. Forgot to add it in a previous PR.

src/mesh/meshblock.hpp
lroberts36: `LogicalLocation`

src/solvers/solver_utils.hpp
lroberts36: It is essentially just a choice you can make. For the way we discretize the diffusion equation, including the volume terms makes the effective matrix symmetric (or at least close to it). This might allow for using CG, but in practice I haven't found this to be better than using BiCGSTAB generally. I will remove the commented out bit or add a flag for this.

src/utils/loop_utils.hpp
lroberts36: The same block can be present in multiple multigrid grids (this is not new in this PR). If you use `packsize = 1`,  a block can start tasks on a coarser level before tasks on a finer level have finished if there is no communication from the block on the coarser grid to itself on the finer grid. Sending a dummy message works well for this. You could also do this through task dependencies, but it would be really gross. Obviously, this involves no MPI communication.

src/bvals/comms/bvals_utils.hpp
Yurlungur: might be useful to add a comment to this effect, or name the 2 something descriptive like NBTYPES or similar.

src/mesh/meshblock.hpp
Yurlungur: After we discussed, I think I understand the meaning here. Might be worth adding a comment above, though, so somebody could get to this function by grepping LogicalLocation or similar

src/utils/loop_utils.hpp
Yurlungur: Can you change "message" to "dummy message" in the comment? I think that would clarify.

src/mesh/mesh-gmg.cpp
Yurlungur: I think a high-level description as we talked about it on the call might help. I struggled to understand this set of comments before we talked.

src/solvers/mg_solver.hpp
```diff
@@ -53,7 +54,9 @@ struct MGParams {
     residual_tolerance =
         pin->GetOrAddReal(input_block, "residual_tolerance", residual_tolerance);
     do_FAS = pin->GetOrAddBoolean(input_block, "do_FAS", do_FAS);
-    smoother = pin->GetOrAddString(input_block, "smoother", smoother);
+    std::string smoother = pin->GetOrAddString(input_block, "smoother", "SRJ2");
+    presmoother = pin->GetOrAddString(input_block, "presmoother", smoother); 
+    postsmoother = pin->GetOrAddString(input_block, "postsmoother", smoother);
```
Yurlungur: You might consider adding docstrings to these. You might also consider adding SRJ1, SRJ2, and SRJ3 explicitly as allowed options so the doc system will pick that up.

Yurlungur (approved):
After discussing in a call, I think I understand this well enough to approve. My comments are mostly for clarity and non-blocking. I'll admit I don't *fully* understand all the details of the logic still, though.

src/interface/mesh_data.hpp
lroberts36: boundary buffer id 0 is the default communication channel that we currently use now. If `MeshData::SetBoundBufferId(btype, id)` is never called for a given boundary type `btype` on a given `MeshData` object, then it should communicate on the default set of buffers for consistency with the previous version of the code. 

The more likely bug would be that someone does
```c++
my_md->SetBoundBufferId(btype, 3); 
```
then communicates with that `MeshData` and we have not created a set of boundary buffers for that boundary type with an `id = 3`. I have added `Mesh::SetNumberOfCommChannels(BoundaryType)`, `Mesh::LockCommChannelNumbers(BoundaryType)`, and `Mesh::GetNumberOfCommChannels(BoundaryType)` for a more explicit and safer interface to setting the number of comm channels. 

Now, `MeshData::SetBoundBufferId(BoundaryType, int)` explicitly checks that the set value is in the range allowed by the mesh.

src/bvals/comms/bvals_utils.hpp
lroberts36: done

src/solvers/mg_solver.hpp
lroberts36: done

src/solvers/solver_utils.hpp
lroberts36: added flag

src/bvals/comms/bnd_id.cpp
lroberts36: This is just to make coalesced comms work. I am punting for now, have a PARTHENON_FAIL detect if you use multiple stages and coalesced comms.

src/mesh/mesh-gmg.cpp
lroberts36: Added something to the docs.

lroberts36:
@par-hermes format

src/bvals/comms/bnd_id.cpp
```diff
@@ -43,7 +43,8 @@ namespace parthenon {
 BndId BndId::GetSend(MeshBlock *pmb, const NeighborBlock &nb,
                      std::shared_ptr<Variable<Real>> v, BoundaryType b_type,
                      int partition, int start_idx) {
-  auto [send_gid, recv_gid, vlabel, loc, extra_id] = SendKey(pmb, nb, v, b_type);
+  // TODO(LFR): This needs to be fixed for unique buffer ids
```
pdmullen, line 46: Is this comment alluding to something still being broken within in this PR?

src/bvals/comms/bnd_info.cpp
```diff
@@ -405,7 +424,7 @@ ProResInfo ProResInfo::GetSend(MeshBlock *pmb, const NeighborBlock &nb,
   ProResInfo out(pmb, nb, v);
   if (!out.allocated) return out;
 
-  if (nb.origin_loc.level() < pmb->loc.level()) {
+  if (NeighborIsCoarser(pmb, nb)) {
```
pdmullen, line 427: This looks like change of behavior from before since it was looking at `nb.origin_loc.level()` and is now looking at `nb.loc.level()`.

src/solvers/bicgstab_solver.hpp
```diff
@@ -62,6 +63,8 @@ struct BiCGSTABParams {
     mg_params = MGParams(pin, input_block);
     relative_residual =
         pin->GetOrAddBoolean(input_block, "relative_residual", relative_residual);
+    volume_weight = pin->GetOrAddBoolean(input_block, "volume_weight", volume_weight,
```
pdmullen, line 80: When should a user invoke this?

src/solvers/mg_solver.hpp
```diff
@@ -52,13 +52,22 @@ struct MGParams {
     max_iters = pin->GetOrAddInteger(input_block, "max_iterations", max_iters);
     residual_tolerance =
         pin->GetOrAddReal(input_block, "residual_tolerance", residual_tolerance);
-    do_FAS = pin->GetOrAddBoolean(input_block, "do_FAS", do_FAS);
-    smoother = pin->GetOrAddString(input_block, "smoother", smoother);
+    do_FAS = pin->GetOrAddBoolean(
+        input_block, "do_FAS", do_FAS,
+        "Use the full approximation scheme in multigrid, required for amr.");
```
pdmullen, line 57: thanks for adding the docstrings here.

tst/regression/test_suites/poisson_gmg/parthinput.poisson
```diff
@@ -71,7 +71,7 @@ exterior_D = 1.0
 <poisson/solver_params>
 precondition = true
 max_iterations = 15
-residual_tolerance = 1.e-12
+residual_tolerance = 1.e-13
```
pdmullen: Caught my eye... how come this change?

src/mesh/mesh.hpp
```diff
@@ -299,6 +307,42 @@ class Mesh {
   comm_buf_map_t boundary_comm_map;
   TagMap tag_map;
 
+  // Sets the number of communication buffers that can be in-flight concurrently
+  // for a given boundary type. This *must* be called before build boundary buffers
+  // is called internally, so use beyond the defaults with care
+  void SetNumberOfCommChannels(BoundaryType bound, std::size_t n_channels) {
+    // TODO(LFR): Fix this, there is no fundamental issue just requires work
+    PARTHENON_REQUIRE(!do_coalesced_comms || n_channels == 1,
```
pdmullen, line 315: Do I understand correctly that because `SetNumberOfCommChannels` is called with `n_channels=2` above,  then multigrid is currently incompatible with `do_coalesced_comms`? 

Artemis with multigrid self-gravity has not worked for us with `do_coalesced_comms=true` in the past, but from this PR description, I thought that maybe this PR was partially addressing that.  

Nevertheless, I pulled this PR into artemis and indeed hit this `PARTHENON_REQUIRE` when trying `do_coalesced_comms=true`.

pdmullen (approved):

src/bvals/comms/bnd_id.cpp
pdmullen: Just read your response to Jonah's similar comment above.

lroberts36: Yeah, this is related to the coalesced comms stuff that I am punting on. I want to leave the TODO so I don't forget to change things here eventually.

src/bvals/comms/bnd_info.cpp
lroberts36: Ah, sorry, that is a little confusing. It should always be true that `nb.origin_loc.level() == nb.loc.level()`, since a logical coordinate transformation doesn't change the refinement level.

src/solvers/bicgstab_solver.hpp
lroberts36: If you want your system to look closer to being symmetric, you should use this. If you write evolve the point value of your potential (as opposed to the cell integrated potential), you can show that by switching to a volume weighted dot product in CG or BiCGSTAB you are essentially solving the volume integrated version of the diffusion equation (which is the version that results in a symmetric matrix). I originally implemented it to see if I could get CG to work, but the MG preconditioner seems to break symmetry anyway.

Often, it seems like this gives a little better convergence than no volume weighting. But of course there is the extra cost of getting the volume.

src/mesh/mesh.hpp
lroberts36: I have never put much work into making sure multigrid works with coalesced comms (although I thought it did just sort of work previously, slightly surprised to hear that you were having problems in Artemis).

This PR definitely breaks coalseced comms for multigrid, since I didn't update the coalesced comms to know about multiple communication channels. I started on it, but it seemed like it would be too time consuming so I punted. As you see, multigrid requires having two channels for the gmg_restriction boundaries, so it triggers this check. This seemed better than having it silently fail.

pdmullen: Thanks!

src/solvers/bicgstab_solver.hpp
pdmullen: Thanks!

src/bvals/comms/bnd_id.cpp
pdmullen: Thanks!

src/mesh/mesh-gmg.cpp
```diff
@@ -93,91 +98,285 @@ void SetMeshBlockNeighbors(Mesh *pmesh, GridIdentifier grid_id, BlockList_t &blo
...
+// This comment was written with the help of generative AI:
+// Terminology and GridIdentifier conventions for Parthenon GMG on a forest-of-octrees AMR
+// mesh
+//
+// The GMG hierarchy is constructed over two independent notions of coarsening:
+//
+//   (1) AMR/tree refinement, indexed by logical_level (Athena++ / Parthenon convention):
+//         - Larger logical_level => finer AMR blocks.
+//         - For a domain of size L in a given direction, the physical block size is
+//               L_block = L / 2^logical_level
```
pgrete: That only applies to domains that can be split into blocks in powers of two in each direction, doesn't it?
Or how does this translate to an L=[1,1,1] mesh with nx1,2,3=[30,30,30] and block size [10,10,10]?

src/mesh/mesh.cpp
```diff
@@ -617,14 +625,13 @@ void Mesh::BuildTagMapAndBoundaryBuffers() {
...
   if (multigrid) {
     for (int gmg_level = GetGMGMinLevel(); gmg_level <= GetGMGMaxLevel(); ++gmg_level) {
-      const auto grid_id = GridIdentifier::two_level_composite(gmg_level);
-      for (auto &partition : GetDefaultBlockPartitions(grid_id)) {
+      for (auto &partition : GetMultigridBlockPartitions(gmg_level)) {
         auto &md = mesh_data.Add("base", partition);
         tag_map.AddMeshDataToMap<BoundaryType::gmg_same>(md);
         tag_map.AddMeshDataToMap<BoundaryType::gmg_prolongate_send>(md);
-        tag_map.AddMeshDataToMap<BoundaryType::gmg_restrict_send>(md);
+        tag_map.AddMeshDataToMap<BoundaryType::gmg_restrict_send>(md, 2);
```
pgrete: Any potential danger (in the long run) to work with a magic number here?

pgrete (approved):
I now skimmed the PR and didn't spot anything that might impact our code at the moment.
Happy to approve and many thanks for the documentation!

src/mesh/mesh.cpp
lroberts36: oh, good catch, that is a bug. It should be explicitly checking how many have been registered

src/mesh/mesh-gmg.cpp
lroberts36: Well, technically each base block is a tree now. So that if `L` is the size of one dimension of a tree, then the block sizes are as written. I will add the word tree here to make things clearer.
```suggestion
//         - For a tree domain of size L in a given direction, the physical block size is
//               L_block = L / 2^logical_level
```

src/mesh/mesh.cpp
lroberts36: fixed

PR #1369, opened 2026-03-09
Parser for ASCII tables and string broadcasting

Yurlungur:
## PR Summary

In riot we have a need to be able to read ascii tables. I decided to put in something relatively generic, which basically reads a 2D table of primitive types in numpy txt format. I also add functionality to read a string from a file on one rank and then broadcast it to all ranks before parsing. I'm using this with the intent to avoid thrashing the filesystem when reading a table. But it could also be used for, e.g., the input decks if we run into filesystem problems at scale.

The catch2 unit testing machinery *does* support tests with MPI (though I can't tell if the tests are actually run on multiple ranks through ctest), but it wasn't threaded all the way through. So minor changes were needed to expose things like MPI ranks to the test code.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [ ] Docs build

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -24,4 +24,5 @@ detail.  Why is this change required?  What problem does it solve?-->
   - [ ] Short summary API changes at the top of the PR (plus optionally with an automated update/fix script)
 - [ ] CI has been triggered on [Darwin](https://re-git.lanl.gov/eap-oss/parthenon/-/pipelines) for performance regression tests.
 - [ ] Docs build
+- [ ] Any contribution that was created or modified with the assistance of generative AI must have a comment disclosing this such as `// This file was made in part with generative AI.`
 - [ ] (@lanl.gov employees) Update copyright on changed files
--- a/src/utils/string_utils.cpp
+++ b/src/utils/string_utils.cpp
@@ -22,6 +22,7 @@
 #include "config.hpp"
 #include "error_checking.hpp"
 #include "globals.hpp"
+#include "kokkos_types.hpp"
 #include "parthenon_mpi.hpp"
 
 namespace parthenon {
@@ -105,14 +106,14 @@ std::string BroadcastFileString(const std::string &filename) {
 }
 
 template <typename T>
-Table2D<T> ParseAsciiTable(std::istream &in) {
-  Table2D<T> out;
+HostArray2D<T> ParseAsciiTable(std::istream &in) {
+  std::vector<T> data;
 
   std::string line;
-  std::size_t line_no = 0;
+  std::size_t rows = 0;
+  std::size_t cols = 0;
 
   while (std::getline(in, line)) {
-    ++line_no;
 
     // Strip comments...
     if (auto pos = line.find('#'); pos != std::string::npos) {
@@ -128,44 +129,57 @@ Table2D<T> ParseAsciiTable(std::istream &in) {
     std::size_t row_count = 0;
 
     while (iss >> value) {
-      out.data.push_back(value);
+      data.push_back(value);
       ++row_count;
     }
 
     if (!iss.eof()) {
-      PARTHENON_THROW("ASCII parser error on line: " + std::to_string(line_no) +
+      PARTHENON_THROW("ASCII parser error on line: " + std::to_string(rows) +
                       "! Incorrect type.");
     }
     if (row_count == 0) continue; // should not happen after trim, but safe
 
-    if (out.rows == 0) {
-      out.cols = row_count;
-      if (out.cols == 0) { // table is empty. We can just return.
-        return out;
+    if (rows == 0) {
+      cols = row_count;
+      if (cols == 0) { // table is empty. We can just return.
+        break;
       }
-    } else if (row_count != out.cols) {
+    } else if (row_count != cols) {
       PARTHENON_THROW("Parsed ASCII table is ragged.");
     }
-    ++out.rows;
+    ++rows;
+  }
+
+  // JMM: Thought about doing this by just copying the data, but doing
+  // it this way safeguards against HostArray2D having a different
+  // layout than row-major ordering.
+  HostArray2D<T> out("Parsed ascii table", rows, cols);
+  {
+    std::size_t idx = 0;
+    for (std::size_t row = 0; row < rows; ++row) {
+      for (std::size_t col = 0; col < cols; ++col) {
+        out(row, col) = data[idx++];
+      }
+    }
   }
 
   return out;
 }
-template Table2D<double> ParseAsciiTable(std::istream &);
-template Table2D<float> ParseAsciiTable(std::istream &);
-template Table2D<int> ParseAsciiTable(std::istream &);
-template Table2D<std::size_t> ParseAsciiTable(std::istream &);
+template HostArray2D<double> ParseAsciiTable<double>(std::istream &);
+template HostArray2D<float> ParseAsciiTable<float>(std::istream &);
+template HostArray2D<int> ParseAsciiTable<int>(std::istream &);
+template HostArray2D<std::size_t> ParseAsciiTable<std::size_t>(std::istream &);
 
 template <typename T>
-Table2D<T> ParseAsciiTable(const std::string &filename) {
+HostArray2D<T> ParseAsciiTable(const std::string &filename) {
   std::string str = BroadcastFileString(filename);
   std::istringstream stream(str);
   return ParseAsciiTable<T>(stream);
 }
-template Table2D<double> ParseAsciiTable(const std::string &);
-template Table2D<float> ParseAsciiTable(const std::string &);
-template Table2D<int> ParseAsciiTable(const std::string &);
-template Table2D<std::size_t> ParseAsciiTable(const std::string &);
+template HostArray2D<double> ParseAsciiTable<double>(const std::string &);
+template HostArray2D<float> ParseAsciiTable<float>(const std::string &);
+template HostArray2D<int> ParseAsciiTable<int>(const std::string &);
+template HostArray2D<std::size_t> ParseAsciiTable<std::size_t>(const std::string &);
 
 } // namespace string_utils
 } // namespace parthenon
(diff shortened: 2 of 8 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -24,4 +24,5 @@ detail.  Why is this change required?  What problem does it solve?-->
   - [ ] Short summary API changes at the top of the PR (plus optionally with an automated update/fix script)
 - [ ] CI has been triggered on [Darwin](https://re-git.lanl.gov/eap-oss/parthenon/-/pipelines) for performance regression tests.
 - [ ] Docs build
+- [ ] Any contribution that was created or modified with the assistance of generative AI must have a comment disclosing this such as `// This file was made in part with generative AI.`
```
Yurlungur, line 27: This is required by LANL rules.

src/utils/string_utils.cpp
```diff
@@ -66,5 +71,97 @@ std::vector<std::string> UnpackStrings(const std::string &pack, char delimiter)
   return unpack;
 }
 
+std::string BroadcastFileString(const std::string &filename) {
```
Yurlungur, line 75: I was torn about how to deal with this. I considered doing the broadcast on the data after parsing, but this seemed more general. And I Decided since it's only at startup, the overhead of reading the whole file once and possibly reading more bytes because, e.g., a double is represented by 14 chars, was sufficiently negligible the simplicity is worth it.

tst/unit/test_ascii_table_parsing.cpp
```diff
@@ -0,0 +1,148 @@
...
+       << "1 2.0 -4567.89e-1\n"
+       << "3 4.0 5.0\n"
+       << std::endl;
+    std::string teststring = ss.str();
+
+    const std::string filename = "testfile.txt";
+    if (parthenon::Globals::my_rank == 0) {
+      std::ofstream out(filename);
+      out << teststring;
+    }
```
Yurlungur, line 122: This generates a new file every time this code is run. Not ideal but it's tiny so I figured it's fine.

.github/PULL_REQUEST_TEMPLATE.md
pgrete: Out of curiosity: where does LANL stand wrt the copyright of generated code?

src/utils/string_utils.cpp
pgrete: Couldn't we just use the existing `IOWrapper::Read_all` call?

src/utils/string_utils.hpp
```diff
@@ -11,15 +11,34 @@
...
 #include <string>
 #include <vector>
 
+#include "basic_types.hpp"
+
 namespace parthenon {
 namespace string_utils {
 
+// Return type for ParseAsciiTable
+// could alternatively be a HostArray2D
```
pgrete: How about going in that direction right now (or any direction that allows for data to be transparently transferred to device)?
I imagine that this might be common use case and the `std::vector` in this struct might cause hidden issue when accessing a captured Table2D on device?

tst/unit/test_ascii_table_parsing.cpp
```diff
@@ -0,0 +1,148 @@
...
+SCENARIO("We can parse a simple ASCII table", "[AsciiTableParser][StringUtils]") {
+  GIVEN("A gross but valid ascii table") {
+    std::stringstream ss;
+    ss << "# header comment line\n"
+       << "                # weird header comment line\n"
+       << "\n" // empty line
+       << "0\t0.0\t123.0e4\n"
+       << "1 2.0 \t -4567.89e-1     " << std::endl // trailing whitespace
+       << "\t     3 4.0 5   \t  # inine comment" << std::endl
+       << std::endl;
```
pgrete, line 40: 👍

tst/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 
 add_library(catch2_define catch2_define.cpp)
-target_link_libraries(catch2_define PUBLIC Catch2::Catch2 Kokkos::kokkos)
+target_link_libraries(catch2_define PUBLIC Catch2::Catch2 Kokkos::kokkos parthenon)
```
pgrete, line 3: Is this required now because of the globals?
And if so, could we potentially get around it (as this feels like entire Parthenon "leaking" into individual unit tests).

pgrete (commented):
Overall looks good and all comments are comments for discussion.

.github/PULL_REQUEST_TEMPLATE.md
Yurlungur: As far as I am aware, LANL considers the CC to belong to the person who used the generative model. That said, the current rule as far as it applies to us, is disclosure.

src/utils/string_utils.cpp
Yurlungur: ...Yes we could. I'll clean that up.

tst/CMakeLists.txt
Yurlungur: Yes it's because of `Globals::my_rank` and `Globals::nranks`. The way to get around this would be to refactor the tested functions to compute the rank and size internally or to pass them into the relevant functions.

I decided to just leave them as global, but that's the alternative. And I'd be willing to switch to that.

src/utils/string_utils.hpp
Yurlungur: My intent was that a table2d would be copied to device later. But I'm fine switching it to a ParArray2D now.

pgrete: I see. But in that case we'd need an additional function to copy from `std::vector` to sth device supported, don't we?
If the basis is a `ParArray2D` then we already have said function in place.

Yurlungur: You can just wrap an-unmanaged view around the vector and then deep copy it to device. But I agree.

src/utils/string_utils.cpp
Yurlungur: Ah actually no. I *did not* use MPI_Read_all deliberately. The reason is I do not want to open the file in parallel---I want to have a single rank read the file and broadcast. Given the expected size of an ASCII file, I think that's preferable, both from a code simplicity and performance perspective. That said, if you feel strongly, I can be convinced to go another way.

The point of reading in serial and broadcasting, rather than having every rank access the file is to avoid throttling the filesystem.

c-prather: I still occasionally run CPU jobs w/many ranks which potentially read from spinning disks or slow parallel filesystems, so I would appreciate the single-rank read.  I was for a while looking into this for initial conditions, even, so that new problem initializations would be easier to write.

src/utils/string_utils.hpp
adamdempsey90: Spiner!

Yurlungur: Indeed, that's what I originally had in mind. :)

src/utils/string_utils.cpp
```diff
@@ -66,5 +71,97 @@ std::vector<std::string> UnpackStrings(const std::string &pack, char delimiter)
...
+
+  if (Globals::my_rank == 0) {
+    std::ifstream in(filename);
+    if (!in) {
+      PARTHENON_THROW("Failed to open file " + filename);
+    }
+
+    // allocates memory for string upfront
+    in.seekg(0, std::ios::end);
+    strlen = static_cast<std::uint64_t>(in.tellg());
```
lroberts36: Potentially dangerous since `tellg` can return a `std::streampos` that is `-1` as an error code.

src/utils/string_utils.cpp
```diff
@@ -66,5 +71,97 @@ std::vector<std::string> UnpackStrings(const std::string &pack, char delimiter)
...
+    str.reserve(strlen);
+    in.seekg(0, std::ios::beg);
+
+    str.assign((std::istreambuf_iterator<char>(in)), std::istreambuf_iterator<char>());
+  }
+
+#ifdef MPI_PARALLEL
+  MPI_Bcast(&strlen, 1, MPI_UNSIGNED_LONG, 0, MPI_COMM_WORLD);
```
lroberts36: Possibly more portable to use `MPI_UINT64_T` here

src/utils/string_utils.hpp
lroberts36: Yeah, I lean towards having it return a host-side `ParArray2D`.

tst/catch2_define.cpp
```diff
@@ -49,6 +51,9 @@ bool HasMPITests(const T &config) {
 #endif // CATCH2_MPI_PARALLEL
 
 int main(int argc, char *argv[]) {
+  parthenon::Globals::my_rank = 0; // overwritten below as needed
+  parthenon::Globals::nranks = 1;
+
   // With Catch2 >2.13.4 catch_discover_tests() is used to discover tests by calling the
```
lroberts36, line 57: 👍

tst/CMakeLists.txt
lroberts36: Leaving them as globals seems fine to me.

lroberts36 (approved):
LGTM

src/utils/string_utils.cpp
Yurlungur: Good point

src/utils/string_utils.cpp
Yurlungur: Good point. Fixed.

src/utils/string_utils.hpp
Yurlungur: Ok I've made it a `HostArray2D` now. It's still host-side, not device, side, but it's transparently a Kokkos type that the downstream code can decide how it.

I think I still prefer a vector, since it's easier to manipulate into other forms but this is more parthenonic so lets go with it.

Yurlungur:
@pgrete this is ready for re-review

pgrete (approved):
LGTM Thanks for the updates. 
Last question/suggestion: Is it worth moving the `BroadcastFileString` function to the IOWrapper namespace/file? It feels more natural there.

Yurlungur:
> LGTM Thanks for the updates. Last question/suggestion: Is it worth moving the `BroadcastFileString` function to the IOWrapper namespace/file? It feels more natural there.

Isn't `IOWrapper` a class? I don't mind moving the function but I think that's actually a little weird given what `IOWrapper` is structurally?

pgrete:
> > LGTM Thanks for the updates. Last question/suggestion: Is it worth moving the `BroadcastFileString` function to the IOWrapper namespace/file? It feels more natural there.
> 
> Isn't `IOWrapper` a class? I don't mind moving the function but I think that's actually a little weird given what `IOWrapper` is structurally?

ah, yes, never mind then.

PR #1375, opened 2026-03-25
Add uint64 support to SwarmPacks

pdmullen:
## PR Summary

PR #1253 introduced the usage of `std::uint64_t` for particle `id`s (if invoked).  This PR extends that change to `SwarmPacks`, wherein we add `std::uint64_t` compatibility.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [ ] Docs build

### DIFF
--- a/example/particle_leapfrog/particle_leapfrog.cpp
+++ b/example/particle_leapfrog/particle_leapfrog.cpp
@@ -3,7 +3,7 @@
 // Copyright(C) 2021-2024 The Parthenon collaboration
 // Licensed under the 3-clause BSD License, see LICENSE file for details
 //========================================================================================
-// (C) (or copyright) 2020-2024. Triad National Security, LLC. All rights reserved.
+// (C) (or copyright) 2020-2026. Triad National Security, LLC. All rights reserved.
 //
 // This program was produced under U.S. Government contract 89233218CNA000001 for Los
 // Alamos National Laboratory (LANL), which is operated by Triad National Security, LLC
--- a/example/particle_leapfrog/particle_leapfrog.cpp
+++ b/example/particle_leapfrog/particle_leapfrog.cpp
@@ -232,6 +232,10 @@ TaskStatus TransportParticles(MeshData<Real> *md, const StagedIntegrator *integr
   auto pack_v_map = desc_v.GetMap();
   parthenon::PackIdx spi_v(pack_v_map["v"]);
 
+  // Make a SwarmPack containing ids (of type uint64_t)
+  static auto desc_id = MakeSwarmPackDescriptor<swarm_position::id>(swarm_name);
+  auto pack_id = desc_id.GetPack(md);
+
   parthenon::par_for(
       DEFAULT_LOOP_PATTERN, "TestSwarmPack", DevExecSpace(), 0,
       pack_pos.GetMaxFlatIndex(), KOKKOS_LAMBDA(const int idx) {
--- a/example/particle_leapfrog/particle_leapfrog.cpp
+++ b/example/particle_leapfrog/particle_leapfrog.cpp
@@ -257,6 +257,10 @@ TaskStatus TransportParticles(MeshData<Real> *md, const StagedIntegrator *integr
           pack_pos(b, swarm_position::x(), n) += pack_v(b, iv + 0, n) * 0.5 * dt;
           pack_pos(b, swarm_position::y(), n) += pack_v(b, iv + 1, n) * 0.5 * dt;
           pack_pos(b, swarm_position::z(), n) += pack_v(b, iv + 2, n) * 0.5 * dt;
+
+          // id
+          PARTHENON_REQUIRE(pack_id(b, swarm_position::id(), n) >= 0,
+                            "Issue with SwarmPack containing uint64 IDs!");
         }
       });
 
--- a/src/interface/meshblock_data.hpp
+++ b/src/interface/meshblock_data.hpp
@@ -370,17 +370,20 @@ class MeshBlockData {
 
   template <typename TYPE>
   SwarmPackCache<TYPE> &GetSwarmPackCache() {
-    if constexpr (std::is_same<TYPE, int>::value) {
-      return swarm_pack_int_cache_;
-    } else if constexpr (std::is_same<TYPE, Real>::value) {
+    if constexpr (std::is_same<TYPE, Real>::value) {
       return swarm_pack_real_cache_;
+    } else if constexpr (std::is_same<TYPE, int>::value) {
+      return swarm_pack_int_cache_;
+    } else if constexpr (std::is_same<TYPE, std::uint64_t>::value) {
+      return swarm_pack_uint64_cache_;
     }
-    PARTHENON_THROW("SwarmPacks only compatible with int and Real types");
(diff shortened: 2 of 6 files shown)


Yurlungur (approved):
LGTM. Some tests would be welcome. But we don't really have any for this codepath anyway.

pdmullen:
> LGTM. Some tests would be welcome. But we don't really have any for this codepath anyway.

I did add a quick `SwarmPack` creation on `id` in the `particle-leapfrog` example to ensure that this type compiles.  Let me know if you want anything more for this PR.

Yurlungur:
> > LGTM. Some tests would be welcome. But we don't really have any for this codepath anyway.
> 
> 
> 
> I did add a quick `SwarmPack` creation on `id` in the `particle-leapfrog` example to ensure that this type compiles.  Let me know if you want anything more for this PR. 

I think that's sufficient for now.

src/interface/meshblock_data.hpp
```diff
@@ -370,17 +370,20 @@ class MeshBlockData {
...
-    if constexpr (std::is_same<TYPE, int>::value) {
-      return swarm_pack_int_cache_;
-    } else if constexpr (std::is_same<TYPE, Real>::value) {
+    if constexpr (std::is_same<TYPE, Real>::value) {
       return swarm_pack_real_cache_;
+    } else if constexpr (std::is_same<TYPE, int>::value) {
+      return swarm_pack_int_cache_;
+    } else if constexpr (std::is_same<TYPE, std::uint64_t>::value) {
+      return swarm_pack_uint64_cache_;
     }
```
lroberts36, line 379: [Written with the help of generative AI] 

I think this could be made quite a bit easier to maintain by making the supported `SwarmPack` scalar types a single `TypeList`, and then deriving the cache storage from that list instead of manually keeping the members / `if constexpr` branches / clear calls in sync.

In other words, instead of repeating:

```cpp
SwarmPackCache<int> swarm_pack_int_cache_;
SwarmPackCache<Real> swarm_pack_real_cache_;
SwarmPackCache<std::uint64_t> swarm_pack_uint64_cache_;
```

and then updating `GetSwarmPackCache<TYPE>()` and `ClearSwarmCaches()` every time a new scalar type is added, we could define the supported types once:

```cpp
using SwarmPackScalarTypes = parthenon::TypeList<int, Real, std::uint64_t>;
```

and then build the cache tuple from that typelist automatically.

Something along these lines:

```cpp
template <typename TypeList>
struct SwarmPackCacheTuple;

template <typename... Ts>
struct SwarmPackCacheTuple<parthenon::TypeList<Ts...>> {
  using type = std::tuple<SwarmPackCache<Ts>...>;
};

using SwarmPackScalarTypes = parthenon::TypeList<int, Real, std::uint64_t>;
using SwarmPackCaches = typename SwarmPackCacheTuple<SwarmPackScalarTypes>::type;

SwarmPackCaches swarm_pack_caches_;
```

Then `GetSwarmPackCache<TYPE>()` can just use `TypeList::IsIn<TYPE>()` / `TypeList::GetIdx<TYPE>()`:

```cpp
template <typename TYPE>
SwarmPackCache<TYPE> &GetSwarmPackCache() {
  static_assert(SwarmPackScalarTypes::template IsIn<TYPE>(),
                "SwarmPacks only compatible with supported scalar types");
  return std::get<SwarmPackScalarTypes::template GetIdx<TYPE>()>(
      swarm_pack_caches_);
}
```

And `ClearSwarmCaches()` can stay very small, e.g.

```cpp
void ClearSwarmCaches() {
  std::apply(
      [](auto &...caches) { (caches.clear(), ...); }, swarm_pack_caches_);
}
```

The main advantage is that adding/removing a supported scalar types in both `MeshData` and `MeshBlockData` becomes a one-line change to `SwarmPackScalarTypes`, and the storage/access pattern updates automatically.

lroberts36 (approved):
LGTM. The comment below is non-blocking if you want to ignore it

example/particle_leapfrog/particle_leapfrog.cpp
```diff
@@ -232,6 +232,10 @@ TaskStatus TransportParticles(MeshData<Real> *md, const StagedIntegrator *integr
   auto pack_v_map = desc_v.GetMap();
   parthenon::PackIdx spi_v(pack_v_map["v"]);
 
+  // Make a SwarmPack containing ids (of type uint64_t)
+  static auto desc_id = MakeSwarmPackDescriptor<swarm_position::id>(swarm_name);
+  auto pack_id = desc_id.GetPack(md);
+
```
pgrete, line 238: Any chance the compiler throws these out as they're not used anywhere?
For simplicity (and checking access), one could add a trivial check in the following kernel that the ids are >= 0.

src/interface/meshblock_data.hpp
pgrete: As far as I can tell, this would basically mirror what we do for dumping various types in `Params` in outputs.

pgrete (approved):
LGTM ( both comments are non blocking).

This is a good reminder to finally start using SwarmPacks in AthenaPK (otherwise, I'd have stumbled across this already earlier when I introduced the uint64 ids).

src/interface/meshblock_data.hpp
lroberts36: Yeah, I think you could get rid of things like 
```c++
  WriteToHDF5AllParamsOfTypeOrVec<bool>(prefix, group);
  WriteToHDF5AllParamsOfTypeOrVec<int32_t>(prefix, group);
  WriteToHDF5AllParamsOfTypeOrVec<int64_t>(prefix, group);
  WriteToHDF5AllParamsOfTypeOrVec<uint32_t>(prefix, group);
  WriteToHDF5AllParamsOfTypeOrVec<uint64_t>(prefix, group);
  WriteToHDF5AllParamsOfTypeOrVec<float>(prefix, group);
  WriteToHDF5AllParamsOfTypeOrVec<double>(prefix, group);
  WriteToHDF5AllParamsOfTypeOrVec<uint8_t>(prefix, group);
```
with a type list approach. Doesn't really matter though.

example/particle_leapfrog/particle_leapfrog.cpp
pdmullen: Addressed in [6fa276d](https://github.com/parthenon-hpc-lab/parthenon/pull/1375/commits/6fa276d3086b1ef67b61dd8ea511f5f0344ee482).

src/interface/meshblock_data.hpp
pdmullen: This is an excellent suggestion.  Would you be opposed to me attempting this proposed changeset on #1376 instead, since it already contains a pretty significant refactor?

lroberts36: If you want this one to go in without the change, I would almost make it a separate PR later. We could combine it with a couple changes to `TypeList` that allow you to check for uniqueness of all the types in the list and make some helper traits for applying the type list to things. Your call

pdmullen: I think it belongs in #1376.  Let me know if you disagree: [5047be3](https://github.com/parthenon-hpc-lab/parthenon/pull/1376/commits/5047be30f166b660e057f17a3b829d2585e6fa01).

PR #1376, opened 2026-03-26
Refactor SwarmPacks

pdmullen:
Diff will look funny because it depends on #1375.

## PR Summary
The first implementation of `SwarmPacks` was header only and did not exactly mirror `SparsePacks`.  I have long disliked this, as it was just an artifact of me trying to get things working as soon as possible.  I have corrected this in this MR, while also moving `SparsePack` and `SwarmPack` files to unique directories.  This change is breaking for anyone that directly included files from `pack/*.hpp`.  Now downstream users need to include via `pack/swarm_pack/*.hpp` or `pack/sparse_pack/*.hpp`.  This change isn't required, but I think it promotes nice separation between the two (very similar!) packing files.

We also here adopt @lroberts36's changeset recommendation from #1375 wherein we make it user for future developers to add new types to `SwarmPack`s. 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


Yurlungur:
Should we wait for #1375 to review?

pdmullen:
> Should we wait for #1375 to review?

Should be ready now.

src/pack/swarm_pack/swarm_pack_types.hpp
```diff
@@ -0,0 +1,47 @@
...
+#include "utils/type_list.hpp"
+
+namespace parthenon {
+
+#define PARTHENON_SWARM_PACK_TYPES(X)                                                    \
+  X(Real)                                                                                \
+  X(int)                                                                                 \
+  X(std::uint64_t)
+
+using SwarmPackTypes = TypeList<int, Real, std::uint64_t>;
```
lroberts36:
```suggestion
using SwarmPackTypes = TypeList<Real, int, std::uint64_t>;
```

src/interface/mesh_data.hpp
```diff
@@ -512,17 +517,13 @@ class MeshData {
...
   template <typename TYPE>
   SwarmPackCache<TYPE> &GetSwarmPackCache() {
-    if constexpr (std::is_same<TYPE, int>::value) {
-      return swarm_pack_int_cache_;
-    } else if constexpr (std::is_same<TYPE, Real>::value) {
-      return swarm_pack_real_cache_;
-    }
-    PARTHENON_THROW("SwarmPacks only compatible with int and Real types");
+    static_assert(SwarmPackTypes::template IsIn<TYPE>(),
+                  "Unsupported type encountered in SwarmPack");
```
Yurlungur: :+1:

lroberts36 (approved):
LGTM, glad the type list stuff worked cleanly.

src/pack/swarm_pack/swarm_pack.hpp
```diff
@@ -65,12 +68,14 @@ class SwarmPack : public SwarmPackBase<TYPE> {
...
+    // accessed on device via instance of types in the type list Ts...
+    // The pack will be created and accessible on the device
     template <class T>
-    SwarmPack GetPack(T *pmd) const {
-      return SwarmPack(SwarmPackBase<TYPE>::GetPack(pmd, *this));
-    }
+    SwarmPack GetPack(T *pmd) const;
 
-    SparsePackIdxMap GetMap() const {
+    SwarmPackIdxMap GetMap() const {
```
pgrete: Just double checking: We cannot use the "Sparse" version any more as the latter is hardcoded to a single (`Real`) type, correct?

Or in other words, the `SwarmPacks` are now more "powerful"/flexible compared to the sparse ones as they're based on the type list?
I'm thinking ahead/wondering about future support of multiple `MeshData` types.

src/pack/swarm_pack/swarm_pack_types.hpp
```diff
@@ -0,0 +1,48 @@
...
+
+namespace parthenon {
+
+#define PARTHENON_SWARM_PACK_TYPES(X)                                                    \
+  X(Real)                                                                                \
+  X(int)                                                                                 \
+  X(std::uint64_t)
+
+using SwarmPackTypes = TypeList<Real, int, std::uint64_t>;
+
```
pgrete, line 32: Should this list somehow be (more closely) tied to the types supported by swarms themselves?

pgrete (approved):
LGTM

src/pack/swarm_pack/swarm_pack.hpp
pdmullen: `SparsePackIdxMap` and `SwarmPackIdxMap` are both of type `std::unordered_map<std::string, std::size_t>`.  

So this change was actually "fixing" a "bug" wherein I failed to switch the return type to the swarm version for the swarm code in #1037.  Perhaps we could move this out of both `SparsePack` and `SwarmPack` and have simply `PackIdxMap`, but that broaches the topic of more general unification between the two codebases.

What do you think?

pgrete: I have too little insight into the internal of the "sparse" packs, but if the machinery is practically identical, I think it makes sense to have one pack index map machinery.

pdmullen: Addressed in [0dc5184](https://github.com/parthenon-hpc-lab/parthenon/pull/1376/commits/0dc51840e88966a8e54bf92de73aa7665f46512f).  But more general unification b/w `SparsePack`s and `SwarmPack`s should be considered in the future.

src/pack/swarm_pack/swarm_pack_types.hpp
pdmullen: Partially addressed by [ccb74cb](https://github.com/parthenon-hpc-lab/parthenon/pull/1376/commits/ccb74cb5a48b59d116abf067c8902e3217b4ba12) and [962a259](https://github.com/parthenon-hpc-lab/parthenon/pull/1376/commits/962a259a553f441755415c9e6571e6983c4175f3), wherein I explicitly check that the `Swarm::Add()` types exist in the `SwarmPackTypes` list.  

But unfortunately, as it stands, adding a new `ParticleVariable` type to `Swarm` is a much bigger lift than adding a type to `SwarmPack`.  I hope you'll agree, however, that this is outside the scope of this PR.

src/pack/default_names.hpp
```diff
@@ -10,14 +10,23 @@
...
+#include "pack/sparse_pack/sparse_pack.hpp"
+#include "pack/swarm_pack/swarm_pack.hpp"
+
+#define VARIABLE(ns, varname)                                                            \
+  struct varname : public parthenon::variable_names::base_t<false> {                     \
+    template <class... Ts>                                                               \
+    KOKKOS_INLINE_FUNCTION varname(Ts &&...args)                                         \
+        : parthenon::variable_names::base_t<false>(std::forward<Ts>(args)...) {}         \
+    static std::string name() { return #ns "." #varname; }                               \
+  }
```
Yurlungur: I'm shocked we didn't have this somewhere already. :+1:

Though I do note that since it's a macro in a header it could end up in any downstream code and possibly conflict with a name macro they're using (that might be different). Perhaps we should use `PAR_VARIABLE`?

src/pack/swarm_pack/swarm_pack_types.hpp
Yurlungur: I think it's outside the scope of this PR, but is it actually that big of a lift?

src/pack/swarm_pack/swarm_pack_types.hpp
```diff
@@ -0,0 +1,48 @@
...
+#include "basic_types.hpp"
+#include "utils/type_list.hpp"
+
+namespace parthenon {
+
+#define PARTHENON_SWARM_PACK_TYPES(X)                                                    \
+  X(Real)                                                                                \
+  X(int)                                                                                 \
+  X(std::uint64_t)
```
Yurlungur, line 29: Clever. And a little evil. Everywhere this is called in header files that include it, I had trouble tracking it down. Not something that needs to be changed necessarily. Just a comment.

src/pack/swarm_pack/swarm_pack_cache.cpp
```diff
@@ -0,0 +1,64 @@
...
+  return BuildAndAdd(pmd, desc);
+}
+
+#define INSTANTIATE_GET(TYPE)                                                            \
+  template SwarmPackBase<TYPE> &SwarmPackCache<TYPE>::Get<MeshData<Real>>(               \
+      MeshData<Real> *, const impl::SwarmPackDescriptor<TYPE> &);                        \
+  template SwarmPackBase<TYPE> &SwarmPackCache<TYPE>::Get<MeshBlockData<Real>>(          \
+      MeshBlockData<Real> *, const impl::SwarmPackDescriptor<TYPE> &);
+PARTHENON_SWARM_PACK_TYPES(INSTANTIATE_GET)
+#undef INSTANTIATE_GET
```
Yurlungur, line 45: Is this pattern actually saving space? I don't hate it... But it feels like the number of lines wouldn't necessarily actually be shorter (and would be much more legible) without the template magic here. I agree this scales better, though, if we add more types.

Yurlungur (approved):
Sorry for the delay. Looks good and thanks for the refactor. Some nitpicks below. My biggest nit is about the macro pattern. I see the appeal, and I'm certainly guilty of patterns like this. But I don't like how the pattern is spread out across so many files. And I'm not sure it's that much shorter in line count than a simpler pattern.

src/pack/swarm_pack/swarm_pack_types.hpp
pdmullen: This is not something I wish to pursue in the near term.

src/pack/default_names.hpp
pdmullen: I guess it was my hope that downstream codes will just use this one instead.  Happy to change, though.

src/pack/swarm_pack/swarm_pack_cache.cpp
pdmullen: I also don't love the macros.  But the idea here was how do we enable the addition (or deletion) of a type to `SwarmPack`s  by simply changing this `swarm_pack_types.hpp` file.  If we get rid of these macros, then we must instantiate all template realizations in the cpp files for all `SwarmPack` related functions.  Maybe this could go away if we go back to header only (maybe...), but then we miss out on the mirror symmetry between `SparsePack` and `SwarmPack` codebases which this PR strives for.

Yurlungur: Yeah I get it---this certainly generalizes better. I'm just wondering if it's saving us space vs explicit instantiations, given we only have 3 types. That said, if we're imagining adding more types in the fullness of time, then maybe it's best to keep the macros.

src/pack/default_names.hpp
Yurlungur: Yeah I mean all the current downstreams almost certainly would. But they could use this one if it were named `PAR_VARIABLE` too.

I'm just a little worried about playing nice down the line. `VARIABLE` is just such a generic name, I wouldn't be super surprised if a parthenon-based code wanted to slurp in another library that has a macro named `VARIABLE` down the line or something like that.

src/pack/swarm_pack/swarm_pack_cache.cpp
pdmullen: Currently we have ~6 functions it is used for.  For `MeshData<Real>` and `MeshBlockData<Real>` template realizations, that amounts to 6*2*3 = 36 instantiations (yikes!). 

This "removes" type, dividing the number by (currently) three (with some added macro boilerplate).  So I think it does reduce lines of code pretty substantially.

But yeah as you mention, the biggest gain here is that you don't have to go and write 12 new instantiations if you add a type to `SwarmPack`s in the future.

src/pack/default_names.hpp
pdmullen: Totally happy to change. 

`PAR_VARIABLE`
`PARTHENON_VARIABLE`
`PARVARIABLE`

and

`PAR_SWARM_VARIABLE`
`PARTHENON_SWARM_VARIABLE`
`PARSWARMVARIABLE`
...

what do y'all think?

src/pack/swarm_pack/swarm_pack_cache.cpp
Yurlungur: Ok yeah, that's quite a bit of code reduction actually. I stand corrected! Thanks for walking me through it.

src/pack/default_names.hpp
Yurlungur: What about `PAR_VARIABLE` and `PAR_SWARMVAR`?

pdmullen: That works.  I'd also throw out `PAR_VAR` and `PAR_SWARMVAR`.

Yurlungur: :+1: I'd be in favor of that.

PR #1378, opened 2026-03-27
MeshData Swarm Tasks

pdmullen:
## PR Summary
This PR adds MeshData equivalents for many Swarm operations including communication and particle removal.  These tasks allow us to more naturally write downstream drivers in terms of MeshData.  This more modern usage of swarms and SwarmPacks is exemplified in a modernized particle_tracers example.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


example/particle_tracers/particle_tracers.cpp
```diff
@@ -36,159 +36,187 @@
...
-// particles.
+// ****************************************************//
+// Define variables/names to be used by this example. *//
+// ****************************************************//
+
+static const char swarm_name[] = "tracers";
+
+namespace field {
+VARIABLE(field, advected);
+VARIABLE(field, deposition);
```
Yurlungur: Note my comment in #1376 that maybe this should be `PAR_VARIABLE`,

example/particle_tracers/particle_tracers.cpp
```diff
@@ -36,159 +36,187 @@
...
-  const Real &dx_i = pmb->coords.Dxc<1>(0);
-  const Real &dx_j = pmb->coords.Dxc<2>(0);
-  const Real &dx_k = pmb->coords.Dxc<3>(0);
+  std::array<Real, 3> dx;
+  dx.fill(std::numeric_limits<Real>::max());
+  for (auto &pmb : pm->block_list) {
+    const auto &reg = pmb->block_size;
+    dx[0] = std::min(dx[0], (reg.xmax_[0] - reg.xmin_[0]) / reg.nx_[0]);
+    dx[1] = std::min(dx[1], (reg.xmax_[1] - reg.xmin_[1]) / reg.nx_[1]);
+    dx[2] = std::min(dx[2], (reg.xmax_[2] - reg.xmin_[2]) / reg.nx_[2]);
```
Yurlungur: why this change?

example/particle_tracers/particle_tracers.cpp
```diff
@@ -36,159 +36,187 @@
...
 
   const Real advected_mean = 1.0;
   const Real advected_amp = 0.5;
-  PARTHENON_REQUIRE(advected_mean > advected_amp, "Cannot have negative densities!");
-  pkg->AddParam<>("advected_mean", advected_mean);
-  pkg->AddParam<>("advected_amp", advected_amp);
+  adv_pkg->AddParam("advected_mean", advected_mean);
+  adv_pkg->AddParam("advected_amp", advected_amp);
+  PARTHENON_REQUIRE(advected_mean > advected_amp,
+                    "Require advected_mean > advected_amp!");
```
Yurlungur: I think the error message here was more clear before.

example/particle_tracers/particle_tracers.cpp
Yurlungur, line 1: The vast majority of changes in this file seem pretty superficial. Why were all these changes made?

example/particle_tracers/particle_tracers.hpp
```diff
@@ -30,28 +30,35 @@ using namespace parthenon;
 
 namespace tracers_example {
 
-class ParticleDriver : public MultiStageDriver {
+class ParticleDriver : public EvolutionDriver {
```
Yurlungur, line 33: Does the multistage driver no longer work? Why remove it? If there's some reason it doesn't work for this use case we should either fix it or remove it completely, as this is a pretty basic use-case.

src/interface/swarm_container.cpp
Yurlungur, line 1: :+1:

Yurlungur (approved):
I'm not going to block this MR, but I'm a little concerned about the rewrite of the particle example to drop the multistage driver. Is this because the multistage driver is for some reason not appropriate? If so we should discuss that. Maybe it should be refactored.

example/particle_tracers/particle_tracers.cpp
pdmullen: Let's continue discussion there.

example/particle_tracers/particle_tracers.cpp
pdmullen: This is now `EstimateTimestepMesh`.  So now we loop through the block list rather than looking at a single block.  I can revert to accessing pmb->coords.Dxc<1>(0) inside the loop if you like.  But I felt that a demonstration of working with `pmb->block_size` was a nice addition (plus we do it downstream in places).

example/particle_tracers/particle_tracers.cpp
pdmullen: I didn't want to name it a density field.

example/particle_tracers/particle_tracers.hpp
pdmullen: Many downstream apps (including those invoking particles) actually invoke `EvolutionDriver`.  `MultiStageDriver` would work fine, and it is tested elsewhere (e.g., `advection`, `fine_advection`, etc.), but I thought it a fresh take on the example that has symmetry with downstream implementations.

example/particle_tracers/particle_tracers.cpp
pdmullen: Can you be more specific? I was really careful to try and do some very specific things here:

- Always use `SparsePack`s instead of legacy packs.
- Always use type based packs with type based variables for both fields and swarms.
- Use `SwarmPack`s where possible.
- Move to `MeshData` level tasks in the driver.
- Exemplify `EvolutionDriver` useage which is used in many contemporary parthenon-based codes.
- Get this example ready for AMR extensions.

This changeset took me quite a bit of time to get working, so your knee jerk reaction is disappointing...  I thought I was making a more modern demonstration of Parthenon advection + particles, using many of the paradigms that I thought we agreed were what we foresaw as "contemporary" parthenon practices.  Let me know where I went wrong in this.

example/particle_tracers/particle_tracers.cpp
Yurlungur: Maybe something like "advected field must be everywere positive" then?

example/particle_tracers/particle_tracers.hpp
Yurlungur: Ok. If that's the reasoning, then I don't object. :+1:

example/particle_tracers/particle_tracers.cpp
pdmullen: Sounds good :+1:

example/particle_tracers/particle_tracers.cpp
Yurlungur: I don't like using `block_size` to reproduce arithmetic the code provides elsewhere, that's all. Since you're looping through blocks anyway, coords is available, and it seems to me to be preferred and more general if, e.g., curvilinear coordinates are active. 

That said, maybe I'm missing something and we do want to show off `pmb->block_size` works?

example/particle_tracers/particle_tracers.cpp
Yurlungur: I'm mostly referring to things like moving two lines for `AddParam` to one line... or changing an if/then to a ternary. It's broadly all cleanup, which I think is good. It just made it hard for me to tell what was necessary modernization and what was aesthetic/style. 

You don't need to change anything back---sorry. It's a nice cleanup.

lroberts36 (approved):

PR #1377, opened 2026-03-26
Extend Initialization Hierarchy

pdmullen:
Depends on #1375 and #1376.

## PR Summary
We here provide yet another function hook to initialization.  A downstream user may want to have one last pass at the initial conditions after the mesh has resolved during initialization.  This will be important, for example, when initializing particles with AMR, as we do not want to continually be sourcing particles via `ProblemGenerator`,  but to instead wait until the Mesh has finalized before we place particles.  `PostInitialization` is the name that makes the most sense for this new hook.  But that one is already taken.  So we rename the old `PostInitialization` routines to `PostProblemGenerator`.  This makes this changeset a breaking change.

## PR Checklist

- [x] Code passes cpplint
- [x] New features are documented.
- [x] Adds a test for any bugs fixed. Adds tests for new features.
- [x] Code is formatted
- [x] Changes are summarized in CHANGELOG.md
- [x] Change is breaking (API, behavior, ...)
  - [x] Change is *additionally* added to CHANGELOG.md in the breaking section
  - [x] PR is marked as breaking
  - [x] Short summary API changes at the top of the PR (plus optionally with an automated update/fix script)
- [ ] CI has been triggered on [Darwin](https://re-git.lanl.gov/eap-oss/parthenon/-/pipelines) for performance regression tests.
- [x] Docs build
- [x] Any contribution that was created or modified with the assistance of generative AI must have a comment disclosing this such as `// This file was made in part with generative AI.`
- [x] (@lanl.gov employees) Update copyright on changed files

PR #1373, opened 2026-03-24
WIP: Particle AMR Exploration

pdmullen:
WARNING: This PR is an exploration of generative AI's utility on developing AMR for particles. AI was let loose on the problem, with frequent berating from me whenever/wherever necessary.  We should discuss how we want to proceed with this style of work, because a PR of this type may represent a first for Parthenon development.  Even this PR description was largely generated by AI.  

## PR Summary

This PR adds particle support for adaptive mesh refinement (AMR) in Parthenon. The main
new core behavior is remapping particles onto the correct post-remesh owning blocks after
refine and derefine events. The PR also fixes a same-rank swarm communication bug exposed
by hostile AMR particle tests, invalidates stale swarm-pack caches after remesh, and
hardens exact-face ownership / cell-mapping behavior so particle edge cases are more
consistent with Parthenon's existing shared-face field ownership model.

The work also adds package-level final-initialization hooks so package-owned state can be
initialized exactly once after startup AMR has converged, without requiring downstream
codes to wire one-off logic through `main.cpp`.

## Core Parthenon Changes

- Refactor swarm-pack implementation so it mirrors sparse-pack layout more closely:
  - `src/pack/swarm_pack/swarm_pack.hpp`
  - `src/pack/swarm_pack/swarm_pack.cpp`
  - `src/pack/swarm_pack/swarm_pack_base.hpp`
  - `src/pack/swarm_pack/swarm_pack_base.cpp`
  - `src/pack/swarm_pack/swarm_pack_cache.hpp`
  - `src/pack/swarm_pack/swarm_pack_cache.cpp`
  - `src/pack/swarm_pack/swarm_pack_descriptor.hpp`
  - `src/pack/swarm_pack/swarm_pack_descriptor.cpp`
  - `src/pack/swarm_pack/make_swarm_pack_descriptor.hpp`
  - `src/pack/swarm_pack/make_swarm_pack_descriptor.cpp`
- Reorganize pack-family files into:
  - `src/pack/sparse_pack/`
  - `src/pack/swarm_pack/`
- Extend `SwarmPackCache` support to include `uint64_t` so swarm IDs can participate in
  cached `MeshData` / `MeshBlockData` swarm packs alongside `Real` and `int`.
- Add a dedicated swarm AMR remesh path in:
  - `src/mesh/swarm_amr_remesh.hpp`
  - `src/mesh/swarm_amr_remesh.cpp`
- Add `src/mesh/amr_particle_ownership.hpp` to define geometric ownership lookup helpers
[...]

PR #1382, opened 2026-04-07
Support Particle AMR

pdmullen:
## PR Summary
This is an agentic PR.  This PR summary itself employed generative AI.

This PR adds swarm/particle support across AMR ownership changes in Parthenon, covering refinement, derefinement, and same-level off-rank redistribution. The new remesh path packs full particle records on device and exchanges them with sparse point-to-point MPI, keeping swarm payloads on direct device buffers rather than bouncing through host memory. It also fixes a pre-existing same-rank swarm send bug exposed during AMR validation and updates `particle_tracers` to serve as a minimal working AMR particle example. Coverage was added with two small unit tests for refine/derefine remapping and an AMR tracer regression/demo that verifies particle positions remain correct through remeshing.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


example/particle_tracers/particle_tracers.cpp
```diff
@@ -113,6 +116,9 @@ std::shared_ptr<StateDescriptor> Initialize(ParameterInput *pin) {
   const Real advected_amp = 0.5;
   adv_pkg->AddParam("advected_mean", advected_mean);
   adv_pkg->AddParam("advected_amp", advected_amp);
+  adv_pkg->AddParam("refine_tol", pin->GetOrAddReal("Background", "refine_tol", 1.25));
+  adv_pkg->AddParam("derefine_tol",
+                    pin->GetOrAddReal("Background", "derefine_tol", 1.05));
```
Yurlungur, line 121: I don't object to the refinement being set here in code... But is there a reason to do it this way, rather than just using the built in AMR criteria for the magnitude of a field?

plan_histories/1382.md
Yurlungur, line 1: Nice! :D 

FWIW the rule for plan histories I came up with was for singularity-eos, not parthenon. But excited to see it here. Hopefully it's a useful way of working with the agents.

src/interface/swarm.hpp
```diff
@@ -247,6 +248,32 @@ class Swarm {
...
+    for (const auto &var : GetVariableVector<T>()) {
+      count += var->NumComponents();
+    }
+    return count;
+  }
+
+  int GetRecordSize() const {
+    return GetComponentCount<Real>() * sizeof(Real) +
+           GetComponentCount<int>() * sizeof(int) +
+           GetComponentCount<std::uint64_t>() * sizeof(std::uint64_t);
```
Yurlungur: There's probably a way of summing over the allowed types with a fold expression or something... not worth it with only 3 though.

src/mesh/swarm_amr_remesh.hpp
```diff
@@ -0,0 +1,68 @@
...
+// This file was made in part with generative AI.
+
+#include <memory>
+#include <vector>
+
+#include "mesh/mesh.hpp"
+
+namespace parthenon {
+
+struct SwarmRemeshContext {
```
Yurlungur, line 26: Am I understanding correctly that this is just a little convenience container to group the relevant data together?

src/mesh/swarm_amr_remesh.cpp
```diff
@@ -0,0 +1,686 @@
...
+namespace parthenon {
+
+namespace {
+
+// Each old MeshBlock produces one local send plan. The payload buffer is contiguous and
+// rank-partitioned so it can be concatenated into one rank-partitioned send buffer later.
+struct BlockSendPlan {
+  ParArray1D<char> buffer;
+  std::vector<int> send_counts;
+  std::vector<int> send_displs;
```
Yurlungur: I had a little trouble figuring out that `displs` was short for `displacements`. I don't feel strongly about this, but I would prefer one of these names:
- `send_offsets`
- `send_displacements`
- `send_offsts`

src/mesh/swarm_amr_remesh.cpp
```diff
@@ -0,0 +1,686 @@
...
+// need MPI messages. This helper computes the possible peer sets from AMR topology alone.
+//
+// The distinction between "possible" and "actual" matters:
+// - topology tells us which rank pairs might communicate after remesh
+// - particle counts tell us which of those pairs actually have nonzero payloads
+//
+// We use the possible peer sets only for the fixed-size count exchange. That avoids an
+// all-rank handshake while still guaranteeing that every posted receive has a matching
+// send, even when the actual count later turns out to be zero.
+struct RemeshPeerSets {
```
Yurlungur: The comment clearly explains what a `PeerSet` is, but I think it would be more in-line with parthenon language and AMR literature to call this `RemeshNeighborPairs` or something similar.

src/mesh/swarm_amr_remesh.cpp
```diff
@@ -0,0 +1,686 @@
...
+KOKKOS_INLINE_FUNCTION void PackValue(const ParArray1D<char> &buffer, int &offset,
+                                      const T &value) {
+  static_assert(std::is_trivially_copyable_v<T>,
+                "Swarm remesh packing requires trivially copyable values.");
+  const auto bytes = Kokkos::bit_cast<std::array<unsigned char, sizeof(T)>>(value);
+  for (std::size_t i = 0; i < sizeof(T); ++i) {
+    buffer(offset + i) = static_cast<char>(bytes[i]);
+  }
+  offset += sizeof(T);
+}
```
Yurlungur: This seems like a cleaner way to do buffer packing than our old buffer packing routines. I wonder if we should merge this with those? (Doesn't have to be in this MR)

src/mesh/swarm_amr_remesh.cpp
```diff
@@ -0,0 +1,686 @@
...
+
+//----------------------------------------------------------------------------------------
+// Unpack a trivially-copyable value from a byte buffer and advance the running offset.
+//
+// As above, keep this implementation device-safe by rebuilding a fixed byte array and
+// then bit-casting it back to the requested type.
+template <typename T>
+KOKKOS_INLINE_FUNCTION T UnpackValue(const ParArray1D<char> &buffer, int &offset) {
+  static_assert(std::is_trivially_copyable_v<T>,
```
Yurlungur: Same comment as above

src/mesh/swarm_amr_remesh.cpp
```diff
@@ -0,0 +1,686 @@
...
+    const Real y, const Real z) {
+  // old_loc is the logical location of the pre-remesh block. During refinement that
+  // block is replaced only by its daughters, so ownership can be resolved locally by
+  // asking which side of the block midpoint the particle lies on in each active
+  // direction.
+  const auto block = pmesh->GetBlockSize(old_loc);
+  const auto child_side = [&](const CoordinateDirection dir, const Real coord) {
+    if (dir > pmesh->ndim) return 0;
+    const Real midpoint = 0.5 * (block.xmin(dir) + block.xmax(dir));
+    return coord > midpoint ? 1 : 0;
```
Yurlungur:
```suggestion
    return coord > midpoint;
```

src/mesh/swarm_amr_remesh.cpp
```diff
@@ -0,0 +1,686 @@
...
+  };
+
+  // Convert the chosen child logical location back into the gid numbering used by the
+  // newly constructed leaf mesh.
+  const auto child = old_loc.GetDaughter(child_side(X1DIR, x), child_side(X2DIR, y),
+                                         child_side(X3DIR, z));
+  const auto it = new_gid_by_loc.find(child);
+  PARTHENON_REQUIRE(it != new_gid_by_loc.end(),
```
Yurlungur: Should this be `PARTHENON_DEBUG_REQUIRE`, just for performance? I guess though if this fails, we need the code to crash...

src/mesh/swarm_amr_remesh.cpp
```diff
@@ -0,0 +1,686 @@
...
+
+  // A swarm may exist on this block but currently contain no active particles. In that
+  // case there is nothing to communicate and the empty plan is returned immediately.
+  if (swarm->GetNumActive() == 0) return plan;
+
+  const int new_gid = context.NewGid(old_gid);
+  const auto &old_loc = context.OldLoc(old_gid);
+  const auto &new_loc = context.NewLoc(new_gid);
+  const bool refined = new_loc.level() > old_loc.level();
+  // The packed wire format stores one leading destination gid followed by the full swarm
```
Yurlungur: I think "packed wire format" is a format in protobuf. It's not a parthenon thing is it?

src/mesh/swarm_amr_remesh.cpp
```diff
@@ -0,0 +1,686 @@
...
+  // record for that particle in byte form.
+  const int record_size = sizeof(int) + swarm->GetRecordSize();
```
Yurlungur, line 115: are our gids ints? Maybe a question for @lroberts36 ... Maybe should be size_t?

src/mesh/swarm_amr_remesh.cpp
```diff
@@ -0,0 +1,686 @@
...
+    auto z_h = swarm->Get<Real>(swarm_position::z::name()).Get().GetHostMirrorAndCopy();
+    for (int n = 0; n <= swarm->GetMaxActiveIndex(); ++n) {
+      if (!mask_h(n)) continue;
+      const int dest_gid = GetRefinedDestinationGid(pmesh, old_loc, new_gid_by_loc,
+                                                    x_h(n), y_h(n), z_h(n));
+      const int dest_rank = context.NewRank(dest_gid);
+      source_indices_by_rank[dest_rank].push_back(n);
+      dest_gids_by_rank[dest_rank].push_back(dest_gid);
+    }
+  }
```
Yurlungur, line 185: Doing this logic on host, and copying the entire x,y,z position arrays seems like it might problematic from a performance perspective. Do we really want to do it this way? I think the preferred mechanism would be to copy the necessary topological data to device and build a GPU data structure containing the relevant information.

src/mesh/swarm_amr_remesh.cpp
```diff
@@ -0,0 +1,686 @@
...
+  int total_particles = 0;
+  for (int rank = 0; rank < nranks; ++rank) {
+    // MPI counts/displacements are expressed in units of the communicated datatype. Here
+    // that datatype is MPI_BYTE, so send_counts and send_displs are byte counts.
+    plan.send_displs[rank] = total_particles * record_size;
+    total_particles += source_indices_by_rank[rank].size();
+    const auto count =
+        static_cast<std::size_t>(source_indices_by_rank[rank].size()) * record_size;
+    PARTHENON_REQUIRE(count <= std::numeric_limits<int>::max(),
+                      "Swarm remesh send buffer too large for MPI.");
```
Yurlungur, line 196: I must be missing something here... `count` is a `size_t`. Why is `int` the MPI limit, not `size_t`?

src/mesh/swarm_amr_remesh.cpp
```diff
@@ -0,0 +1,686 @@
...
+//----------------------------------------------------------------------------------------
+// Derive the sparse rank communication pattern implied by the AMR topology.
+//
+// Every rank has the full old/new block topology, so it can independently determine
+// which remote ranks it may need to exchange remeshed particles with. This lets the
+// remesh path stay in the same point-to-point style as the rest of Parthenon instead of
+// requiring an all-rank communication operation.
+RemeshPeerSets
+BuildRemeshPeerSets(const Mesh *pmesh, const SwarmRemeshContext &context,
+                    const std::unordered_map<LogicalLocation, int> &new_gid_by_loc) {
```
Yurlungur: Is there some opportunity for unifying this with the code used by the mesh AMR logic? Obviously they both use the mesh code. But since this doesn't care about particles, it seems like it should be possible to reuse the same code?

src/mesh/swarm_amr_remesh.cpp
```diff
@@ -0,0 +1,686 @@
...
+
+//----------------------------------------------------------------------------------------
+#ifdef MPI_PARALLEL
+// Wait on a small batch of remesh point-to-point requests.
+void WaitAll(std::vector<MPI_Request> &reqs) {
+  if (!reqs.empty()) {
+    PARTHENON_MPI_CHECK(MPI_Waitall(reqs.size(), reqs.data(), MPI_STATUSES_IGNORE));
+  }
+}
+#endif
```
Yurlungur: Feels like this is something we should be able to reuse outside of swarms. Maybe it should be in utils. I bet there are places where a wait like this could be useful.

tst/unit/test_swarm_amr_remesh.cpp
Yurlungur, line 1: I like these test cases!

Yurlungur (changes requested):
This is pretty cool! And pretty clean now. Thanks again for splitting this MR into multiple smaller ones. The diff is a lot more manageable now.

I have mostly minor comments and one major one, below, about the way the coordinates are used to decide destination. You've thought about this more deeply than I have, so maybe there's no better way to do it than you've written here, but I'd like to reason it out a little bit.

Also doesn't have to be this PR, but I think there's opportunity here to unify some of the swarm and mesh logic... I get the sense there might be opportunities for code re-use. Also with the "normal" particle comms logic, I wonder if there's an opportunity to unify some of this stuff with that code path.

plan_histories/1382.md
Yurlungur: and obviously I took the idea from @lroberts36

example/particle_tracers/particle_tracers.cpp
pdmullen: We could probably get the magnitude refinement to work similarly here.  

The current AMR strategy for `particle_tracers` looks at the max of `advected` and the min of `advected`.  Then it finds the delta = max - min and then refines off of that.  

So definitely something in the same spirit could be achieved with included Parthenon default refinement routines, but I think this `particle_tracers` example is shaping up to be a nice demo of fields + particles working together in a "modern" (to us, at least) Parthenon implementation--- Perhaps demoing custom refinement is a nice addition.

plan_histories/1382.md
pdmullen: Cool.  What do you think about its placement?  Should we put our `CHANGELOG.md` and a subdirectory called `plan_histories` inside a folder called `dev` (or something)?

src/interface/swarm.hpp
pdmullen: Will take a look.

src/mesh/swarm_amr_remesh.hpp
pdmullen: Correct. Wrapping up everything needed for remesh.  That way I don't have to have a million arguments to the `RemeshSwarms` function.

src/mesh/swarm_amr_remesh.cpp
pdmullen: Will do!

src/mesh/swarm_amr_remesh.cpp
pdmullen: Will do!

src/mesh/swarm_amr_remesh.cpp
pdmullen: I am glad you had this reaction.  I also liked it.  Previous swarm logic packed everything has `MPI_DOUBLE`.  I thought AI did a nice job with this one.  I actually already spun up another branch that tries to switch normal swarm comm logic to this design.  A separate PR will come after this one.

src/mesh/swarm_amr_remesh.cpp
pdmullen: Same response as above 😸

src/mesh/swarm_amr_remesh.cpp
pdmullen: Will do!

src/mesh/swarm_amr_remesh.cpp
pdmullen: I worry that some corner cases may present themselves as we push the infrastructure.  Can we keep this `PARTHENON_REQUIRE` for now, leaving a comment about switching to `PARTHENON_DEBUG_REQUIRE` as the swarm remeshing capability matures?

src/mesh/swarm_amr_remesh.cpp
pdmullen: Not a Parthenon thing to my knowledge.  Happy to change language here if "packed wire format" is too evocative.

src/mesh/swarm_amr_remesh.cpp
pdmullen: Good idea/catch. Will look into this one!

example/particle_tracers/particle_tracers.cpp
Yurlungur: Yep that makes sense to me. :+1:

plan_histories/1382.md
Yurlungur: oh moving that stuff into dev is a good idea. I like that.

src/interface/swarm.hpp
Yurlungur: You don't have to... like I said, probably not worth it with 3 types. Was just an aside.

src/mesh/swarm_amr_remesh.cpp
pdmullen: I think some of the forest code uses wider integer types.  But for the `gid`s used in this remeshing pathway, I think `gid`s are universally `int`s (as the code stands today).

src/mesh/swarm_amr_remesh.cpp
Yurlungur: Nice. Sounds good!

src/mesh/swarm_amr_remesh.cpp
Yurlungur: Sounds good.

plan_histories/1382.md
pdmullen: Will do!

src/mesh/swarm_amr_remesh.cpp
Yurlungur: 🤷 it's probably fine. I'd just say something like "this format" or "our format" and leave it at that.

pdmullen: Will do!

src/mesh/swarm_amr_remesh.cpp
pdmullen: Happy to place in `utils`.  But I may wait (no pun intended) on converting other Parthenon wait calls (in, e.g., field logic) to this utility.

src/mesh/swarm_amr_remesh.cpp
pdmullen: Good question.  Maybe @lroberts36 has ideas.  Probably a grand sweep of unification would benefit the comms logic.  But that sounds like a pretty big undertaking.

src/mesh/swarm_amr_remesh.cpp
Yurlungur: Sounds good.

src/mesh/swarm_amr_remesh.cpp
pdmullen: `count` is indeed a `size_t`.  However, it will later need to be narrowed to `int` to be used in, e.g., `MPI_ISend`.  We first define count as a `size_t` so that we can check if it actually fits into `int` before passing it to MPI API.  So this sidestep via `size_t` is purely for the sake of the check.  Hope that makes sense.

Yurlungur: Why not send `size_t` with `MPI_ISend`? I'm sure you've thought about this and I'm just missing something obvious.

Yurlungur: It is indeed something obvious I missed, which is `MPI_Isend only accepts ints for count.

src/mesh/swarm_amr_remesh.cpp
pdmullen: Hope to have addressed in [cf08403](https://github.com/parthenon-hpc-lab/parthenon/pull/1382/commits/cf084039c321ad80207cf1b752734f44c76bce29).  Let me know what you think.

plan_histories/1382.md
pdmullen: Addressed in [fbe60f7](https://github.com/parthenon-hpc-lab/parthenon/pull/1382/commits/fbe60f707e129e3df9978fe720095af414c2e97d).

src/mesh/swarm_amr_remesh.cpp
```diff
@@ -181,15 +175,45 @@ BuildBlockSendPlan(const std::shared_ptr<Swarm> &swarm, const Mesh *pmesh,
...
+    const Real y_mid = 0.5 * (block.xmin(X2DIR) + block.xmax(X2DIR));
+    const Real z_mid = 0.5 * (block.xmin(X3DIR) + block.xmax(X3DIR));
+
+    auto child_gids_h = GetRefinedDestinationGids(pmesh, old_loc, new_gid_by_loc);
+    ParArray1D<int> child_gids("swarm_amr_remesh_child_gids", child_gids_h.size());
+    auto child_gids_mirror = child_gids.GetHostMirror();
+    for (int i = 0; i < child_gids_h.size(); ++i) {
+      child_gids_mirror(i) = child_gids_h[i];
+    }
+    child_gids.DeepCopy(child_gids_mirror);
```
Yurlungur: You can probably eliminate one host-side copy here by wrapping `child_grids_h` in a `Kokkos::view` with unmanaged memory.

dev/plan_histories/1382.md
```diff
@@ -0,0 +1,43 @@
...
+refinement, derefinement, and same-level redistribution, while keeping the data path
+compatible with device-resident execution and GPU-aware MPI.
+
+## Functional Plan
+
+- Add a dedicated swarm remesh phase that runs after the new leaf mesh and field AMR
+  transfers are already complete.
+- Build the remesh around block ownership changes:
+  - same-level remap sends every particle in an old block to one new block
+  - derefinement sends every particle in a child block to its parent block
```
lroberts36, line 15: How does this interact with the ownership model? A particle on a face may switch blocks on de-refinement if the neighboring block stays at the old level. That being said, I almost think ownership for particles should be based on the direction of travel (i.e. the block they are moving into owns them) rather than the which block owns the face fields.

src/mesh/swarm_amr_remesh.cpp
Yurlungur: Yep. Looks good! Thanks for the quick fix!

src/interface/swarm.hpp
pdmullen: Addressed in [d16fac6](https://github.com/parthenon-hpc-lab/parthenon/pull/1382/commits/d16fac6cbafaf1b54eb615fe048e661a803c43cf).

dev/CHANGELOG.md
```diff
@@ -0,0 +1,728 @@
+# Changelog
+
+## Current develop
+
+### Added (new features/APIs/variables/...)
```
lroberts36, line 7: Why is the changelog diff so huge?

src/interface/swarm.hpp
Yurlungur: Nice. :+1:

src/mesh/swarm_amr_remesh.cpp
pdmullen: This one was removed.  But I did eliminate this pattern where it existed elsewhere in the file in [243c751](https://github.com/parthenon-hpc-lab/parthenon/pull/1382/commits/243c75111acbf4f1be20014b8e79df3d7cd32b72).

src/mesh/swarm_amr_remesh.cpp
pdmullen: Addressed in [7ce81e0](https://github.com/parthenon-hpc-lab/parthenon/pull/1382/commits/7ce81e0c8114222d9dfef9791a914dbf118ad581).

src/mesh/swarm_amr_remesh.cpp
lroberts36: yeah, I didn't change the type of `gid`s with the forest updates.

src/mesh/swarm_amr_remesh.cpp
lroberts36: Would a better pattern be to just communicate all particles to each daughter and then after remesh do a pass over blocks and invalidate any particles that are not within a block? That gets rid of a lot of the logic here, but obviously means a little more communication. I also don't know exactly what machinery already exists for checking particles on a block and defragging.

src/mesh/swarm_amr_remesh.cpp
pdmullen: Moved to `utils` in [914c0c7](https://github.com/parthenon-hpc-lab/parthenon/pull/1382/commits/914c0c7af9cca3e8bebbfb338c894b66732484f0).

src/mesh/swarm_amr_remesh.cpp
pdmullen: Moved to `utils` in [914c0c7](https://github.com/parthenon-hpc-lab/parthenon/pull/1382/commits/914c0c7af9cca3e8bebbfb338c894b66732484f0).

src/mesh/swarm_amr_remesh.cpp
pdmullen: Moved to `utils` in [914c0c7](https://github.com/parthenon-hpc-lab/parthenon/pull/1382/commits/914c0c7af9cca3e8bebbfb338c894b66732484f0).

src/mesh/swarm_amr_remesh.cpp
pdmullen: Addressed in [7a6aab1](https://github.com/parthenon-hpc-lab/parthenon/pull/1382/commits/7a6aab14c615b393e492e1292ddef72347797176).

src/mesh/swarm_amr_remesh.cpp
lroberts36: Yeah, I think it would be nice to unify and the AMR remeshing code is a bit gross. That being said, we know it works. So probably not something we want to spend time on unless there is a performance concern.

src/mesh/swarm_amr_remesh.cpp
pdmullen: Renamed to `RemeshNeighborRanks` in [82f021e](https://github.com/parthenon-hpc-lab/parthenon/pull/1382/commits/82f021ebe3cc0c8f7d8ab954577f4324fd3ef85c).

plan_histories/1382.md
Yurlungur: Leaving this one unresolved in case other folks want to chime in, but I am happy with how this looks.

dev/CHANGELOG.md
lroberts36: ah, I get it.

Yurlungur (approved):
Thanks for the quick fixes! Looks good!

src/mesh/swarm_amr_remesh.cpp
pdmullen: Addressed in [d868b57](https://github.com/parthenon-hpc-lab/parthenon/pull/1382/commits/d868b5704059be3c100ca03b8dea31762543cfd1).

dev/plan_histories/1382.md
pdmullen: You may remember in my monster PR #1373 that I made a big kerfuffle about ownership and Parthenon behaving when particles live exactly on faces, edges, or nodes.  

In practice, I think the chances are near zero that we ever have a particle exactly on such a topological feature, unless the user sourced it there. 

That absolutely does not mean that we shouldn't be cognizant of that edge case, but I was hoping to split off that effort from this one (particularly since many downstream codes take explicit measures to avoid doing exactly this).  

Do you think ownership for exact face/edge/nodes should be addressed in this PR? or okay to split off?

lroberts36: yeah, I think I wouldn't have been concerned about ownership without the initial PR. It seems like it is not as big an issue for particles, since there is only one representation of a given particle. Totally fine to ignore it here.

pdmullen: Resolving this for now.  Made Issue #1383.

src/mesh/swarm_amr_remesh.cpp
pdmullen: Interesting idea.  From a performance standpoint I am not sure which is better.  

From the machinery available to us currently, the call chain required would be something like:
(1) Remesh everybody to daughter blocks.
(2) Loop through all particles on all daughter blocks and see if they are on or off block.  If off block, `MarkParticleForRemoval`.  
(3) `RemoveMarkedParticles` over all daughter blocks.
(4) `Defrag`

Right now the `Defrag` task is moreso put in the users hands.  This would tie it to remeshing too.  It feels a bit more involved to me, what do you think?

dev/plan_histories/1382.md
adamdempsey90: > In practice, I think the chances are near zero that we ever have a particle exactly on such a topological feature, unless the user sourced it there.

With snapping to edges/vertices/faces this should happen, especially in test problems that might say push particles along a specific coordinate direction. With trillions of particle steps in a simulation, nothing is off the table.

src/mesh/swarm_amr_remesh.cpp
lroberts36: That is essentially what I was thinking. It was just an idea that came to mind when looking through this, not sure if it is better. It will not change the number of messages but it certainly means more data transfer. I mostly thought it would simplify the code a bit, but maybe not enough to be worth it.

src/mesh/swarm_amr_remesh.cpp
```diff
@@ -0,0 +1,665 @@
...
+    }
+  } else {
+    // Refinement is the only case where particles from one old block fan out to several
+    // new leaf blocks. Compute that daughter ownership on device and only copy back the
+    // compact destination-gid result, rather than mirroring full coordinate arrays.
+    const auto block = pmesh->GetBlockSize(old_loc);
+    const Real x_mid = 0.5 * (block.xmin(X1DIR) + block.xmax(X1DIR));
+    const Real y_mid = 0.5 * (block.xmin(X2DIR) + block.xmax(X2DIR));
+    const Real z_mid = 0.5 * (block.xmin(X3DIR) + block.xmax(X3DIR));
+
```
lroberts36, line 149: Does this break if we move to non-uniform coordinates in the future? Maybe we expect `xmin` and `xmax` to be in some uniform system though. Probably a problem for future us.

src/mesh/swarm_amr_remesh.hpp
```diff
@@ -0,0 +1,68 @@
...
+  SwarmRemeshContext(int old_gid_first_in, int old_gid_last_in,
+                     const std::vector<int> &old_to_new_gid_in,
+                     const std::vector<LogicalLocation> &old_locs_in,
+                     const std::vector<LogicalLocation> &new_locs_in,
+                     const std::vector<int> &old_ranks_in,
+                     const std::vector<int> &new_ranks_in)
```
lroberts36, line 32: Doesn't matter much, but this is really not specific to swarms so a better name might be `RemeshContext`. At some point we may want to unify this with the field AMR stuff, rather than effectively doing it twice. Unfortunately, the field stuff is a bit of a mess.

lroberts36 (approved):
LGTM! Very glad to see we are now supporting particle AMR. 

I only reviewed the changes related to #1382, but it looked like #1377 and #1378 were included here.

Yurlungur:
@pdmullen do you have merge power so you can merge this when you're ready?

src/mesh/swarm_amr_remesh.hpp
pdmullen: Feel free to push back:

Since this utility is only used for Swarms, can we keep the name as is for now.  During a later unification, I absolutely agree that a better name will be `RemeshContext`.

src/mesh/swarm_amr_remesh.cpp
pdmullen: See PR #1391 .

src/mesh/swarm_amr_remesh.hpp
lroberts36: yeah, I am not worried either way

pgrete:
I just skimmed through the PR. Interesting solution to the remesh unit test!

There's nothing preventing a merge isn't it?

pgrete:
Another WIP hiccup. Force-merging now.

PR #1385, opened 2026-04-09
Refactor ParameterInput: Separate parsing from storage

lroberts36:
<!--Provide a general summary of your changes in the title above, for
example "Add AMR unit test for cell centered fields.".  Please avoid
non-descriptive titles such as "Addresses issue #8576".-->

## PR Summary

This PR relied on agentic AI and this summary was produced by generative AI.

## Motivation

The original `ParameterInput` implementation tightly coupled text file parsing with parameter storage, making it difficult to support alternative input formats (Python, TOML, etc.) without duplicating storage logic or hacking around the text parser.

This PR refactors the architecture to cleanly separate **parsing** concerns from **storage** concerns, enabling multiple input formats to coexist.

## Key Changes

### 1. Parser Separation with Public API

**New `AddParsedParameter` interface** for parsers to populate parameters:
```cpp
void AddParsedParameter(const std::string &block,
                       const std::string &name,
                       const ParamValue &value,
                       const std::string &comment = "# From parser");
```

- Any parser (text, future formats like Python/TOML) can use this generic interface
- Supports both typed values (future parsers) and `UnresolvedString` (text files)
- Enforces that parsing must not be finalized yet
- Lazy conversion: `UnresolvedString` values are converted on first access

**Explicit finalization:**
```cpp
void FinalizeParsing(); // Finalize parsing phase (GetOrAdd/Set still work)
```

Renamed from `MarkResolved()` for clarity - makes it explicit that only the *parsing phase* is finalized, not all modifications. Runtime operations (`GetOrAdd`, `Set`) continue to work after finalization.

### 2. Storage Refactor: Vector-of-Vectors with Index Maps                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                      
  Migrated from dual linked-list + map storage to **vector-of-vectors with index maps**:                                                                                                                                                                                                                                                              
  ```cpp                                                                                                                                                                                                                                                                                                                                              
  std::vector<Block> param_storage_;  // Each Block contains std::vector<Parameter>                                                                                                                                                                                                                                                                   
  std::unordered_map<std::string, size_t> block_index_;  // O(1) block lookup                                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                                                      
  struct Block {                                                                                                                                                                                                                                                                                                                                      
    std::string name;                                                                                                                                                                                                                                                                                                                                 
    std::vector<Parameter> params;                                                                                                                                                                                                                                                                                                                    
    std::unordered_map<std::string, size_t> param_index;  // O(1) parameter lookup within block                                                                                                                                                                                                                                                     
  };                                                                                                                                                                                                                                                                                                                                                  
``` 
                                                                                                                                                                                                                                                                                                                                                    
  Why vectors + index maps?                                                                                                                                                                                                                                                                                                                           
  - Preserves insertion order (critical for restart file compatibility via vectors)                                                                                                                                                                                                                                                                 
  - O(1) lookup performance via index maps at both block and parameter levels                                                                                                                                                                                                                                                                         
  - Pointer stability via indices instead of pointers (avoids invalidation on vector reallocation)                                                                                                                                                                                                                                                  
  - Maintains backward compatibility with existing restart files                                                                                                                                                                                                                                                                                      
  - Scales well for complex problems with many blocks/parameters                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                     
### 3. Internal Architecture Improvements

**Separated internal helper from parser API:**
- `AddParsedParameter()` (public) - For parsers, enforces `!parsing_finalized_`
- `AddParameter_()` (private) - For internal use (`GetOrAdd`, `Set`)

This prevents confusion about which interface to use:
- **Creating new input sources** → use `AddParsedParameter`
- **Runtime overrides** (restart, command line) → use `Set*`

**Clear documentation** distinguishing parser interface from runtime overrides:
- **Parser interface**: `AddParsedParameter` for external input sources (text files, future formats)
- **Runtime interface**: `Set*` for programmatic overrides (restart data, command line flags)
- **DO NOT** use `Set*` in parsers - use `AddParsedParameter` instead

**Existence checks don't trigger finalization:**
- `DoesParameterExist()` and `DoesBlockExist()` are const queries that work regardless of finalization state
- Only value retrieval (`Get`, `GetOrAdd`, `Set`) triggers automatic finalization

### 4. Type System

**ParamValue variant** supports both typed and unresolved values:
```cpp
using ParamValue = std::variant<
  UnresolvedString,      // Lazy conversion from text files
  bool, int, Real, std::string,
  std::vector<bool>, std::vector<int>,
  std::vector<Real>, std::vector<std::string>
>;
```

## Benefits

1. **Extensibility**: New input formats can be added without modifying core storage logic
2. **Type Preservation**: Future parsers can provide typed values directly without string conversion
3. **Backward Compatibility**:
   - Existing text file parsing unchanged
   - Restart files maintain same ordering
   - All existing query APIs (`Get*`, `GetOrAdd*`, `Set*`) unchanged
4. **Clear Separation**: Parser concerns isolated from storage and query operations
5. **Future-Ready**: Foundation for Python input support (separate PR), TOML, or other formats

## Testing

Added comprehensive test coverage (16 tests, 75 assertions):

- ✅ `AddParsedParameter` with typed scalars and vectors
- ✅ `UnresolvedString` lazy conversion
- ✅ Mixing `LoadFromStream` and `AddParsedParameter`
- ✅ Parameter override behavior (last value wins)
- ✅ `FinalizeParsing` enforcement (blocks further parsing)
- ✅ Parameter ordering preservation for restart compatibility
- ✅ Automatic block creation
- ✅ Parser vs runtime interface separation

All existing tests pass, including:
- Parameter hashing
- Required/desired parameter checking
- Parameter deletion
- Text file parsing with weird whitespace

## Breaking Changes

### What's Breaking

**Direct linked list access removed:**
```cpp
// OLD (no longer works):
InputBlock *pib = pin->pfirst_block;
while (pib != nullptr) {
  if (pib->block_name.compare(0, 27, "parthenon/static_refinement") == 0) {
    // ...
  }
  pib = pib->pnext;
}

// NEW (use this instead):
auto blocks = pin->GetBlockNamesWithPrefix("parthenon/static_refinement");
for (const auto &block_name : blocks) {
  // Use block_name with standard Get methods
  pin->GetReal(block_name, "x1min");
}
```

### Migration Guide

**For downstream codes:**
- Replace `pfirst_block`/`pnext` iteration with `GetBlockNamesWithPrefix()` or `GetBlockNames()`
- All standard query methods unchanged: `Get*`, `GetOrAdd*`, `Set*`, `DoesParameterExist`
- Text file parsing works exactly as before
- Restart file format unchanged

**For new parser developers:**
- Use `AddParsedParameter` to populate parameters from new input formats
- Call `FinalizeParsing()` when parsing is complete
- See `docs/parser_separation_summary.md` for detailed examples
- **Do not use `Set*` in parsers** - those are for runtime overrides

### Why This Breaking Change?

The dual linked-list + map storage was complex to maintain and made it difficult to add new parsers. The new vector-based storage:
- Simplifies implementation (single source of truth)
- Enables multiple input formats through unified interface
- Preserves all functional behavior (ordering, restart compatibility)
- Provides cleaner query APIs

## Documentation

- `docs/parser_separation_summary.md` - Detailed design document
  - Will be moved to `dev/plan_histories/` once Patrick's PRs merge into develop
- Enhanced inline documentation clarifying parser vs runtime interfaces
- Clear usage examples for `AddParsedParameter` workflow

## Follow-up Work

This PR lays the groundwork for:
- Python input support
- TOML/YAML parsers
- Rummy bytecode input

## PR Checklist

<!-- Note that some of these check boxes may not apply to all pull requests -->

- [x] Code passes cpplint
- [x] Code is formatted
  - [x] Short summary API changes at the top of the PR
- [x] Docs build


lroberts36:
@par-hermes format

docs/parser_separation_summary.md
```diff
@@ -0,0 +1,231 @@
...
+
+## Architecture
+
+### Three Layers
+
+```
+┌─────────────────────────────────────────────┐
+│  Parser Layer                               │
+│  - Text (.pin files)                        │
+│  - Python (pybind11)                        │
```
Yurlungur: Is python to be implemented in the future?

docs/parser_separation_summary.md
```diff
@@ -0,0 +1,231 @@
...
+pi.add_string("block", "name", "value")
+pi.add_int_vector("block", "name", [1, 2, 3])
+# ... similar for real_vector, bool_vector, string_vector
+
+# Mark complete (optional - done automatically on first Get)
+pi.mark_resolved()
+
+# Query parameters (normal API)
+val = pi.get_int("block", "name")
+```
```
Yurlungur: This is to be added in a separate MR right?

src/mesh/mesh.cpp
```diff
@@ -1106,79 +1106,76 @@ void Mesh::DoStaticRefinement(ParameterInput *pin) {
...
-            << "Refinement level must be larger than 0 (root level = 0)" << std::endl;
-        PARTHENON_FAIL(msg);
-      }
-      if (ref_lev > max_level_ref_) {
-        msg << "### FATAL ERROR in Mesh constructor" << std::endl
-            << "Refinement level exceeds the maximum level (specify "
-            << "'numlevel' parameter in <parthenon/mesh> input block if adaptive)."
-            << std::endl;
+  auto static_ref_blocks = pin->GetBlocksWithPrefix("parthenon/static_refinement");
+  for (const auto &block_name : static_ref_blocks) {
```
Yurlungur: This is a good change, but make sure to note in the changelog this is breaking.

src/CMakeLists.txt
```diff
@@ -10,6 +10,7 @@
 # license in this material to reproduce, prepare derivative works, distribute copies to
 # the public, perform publicly and display publicly, and to permit others to do so.
 #=========================================================================================
+# This file was made in part with generative AI.
```
Yurlungur: just for future reference? :P

src/utils/type_list.hpp
```diff
@@ -155,6 +156,23 @@ template <class... Ts>
 struct isTypeList<TypeList<Ts...>> : public std::true_type {};
 
 //>>>>>>> develop
```
Yurlungur, line 158: this comment looks like it shouldn't be here... (though you didn't add it)

src/parthenon_manager.cpp
```diff
@@ -136,6 +137,9 @@ ParthenonStatus ParthenonManager::ParthenonInitEnv(int argc, char *argv[]) {
   // Modify based on command line inputs
   pinput->ModifyFromCmdline(argc, argv);
 
+  // Mark parsing as complete - no more parameters can be added
+  pinput->MarkResolved();
```
Yurlungur: How does this interact with `GetOrAdd`?

Also this seems unnecessarily stateful. Why do we need a `MarkResolved` function at all?

src/parameter_input.cpp
```diff
@@ -259,48 +234,45 @@ void ParameterInput::LoadFromFile(IOWrapper &input) {
...
-  // if this is the first block in list, save pointer to it in class
-  if (pfirst_block == nullptr) {
-    pfirst_block = pib;
-  } else {
-    plast->pnext = pib; // link new node into list
+Block *ParameterInput::FindOrAddBlock_(const std::string &name) {
+  // Search vector to see if name exists, return if found
+  for (auto &block : param_storage_) {
+    if (block.name == name) return &block;
   }
```
Yurlungur: `std::find` maybe preferred? Here and elsewhere?

src/parameter_input.cpp
```diff
@@ -20,6 +20,10 @@
 // PURPOSE: Member functions of this class are used to read and parse the input file.
 //   Functionality is loosely modeled after FORTRAN namelist.
 //
+// NOTE: This describes the standard text-based input format. Parthenon also supports
+//   other input formats (e.g., Python scripts via -i script.py) that populate the
+//   same underlying parameter storage.
```
Yurlungur: Plans to...? May? I know you're going to merge that change momentarily but it still feels like we should be a bit more general here

src/parameter_input.cpp
```diff
@@ -259,48 +234,45 @@ void ParameterInput::LoadFromFile(IOWrapper &input) {
...
-
-  // Search singly linked list of InputBlocks to see if name exists, return if found.
-  while (pib != nullptr) {
-    if (name.compare(pib->block_name) == 0) return pib;
-    plast = pib;
-    pib = pib->pnext;
+//! \fn Block* ParameterInput::FindBlock_(const std::string & name)
+//  \brief find specified Block.  Returns pointer to block or nullptr.
+
+Block *ParameterInput::FindBlock_(const std::string &name) {
```
Yurlungur, line 239: I assume you just did pointers here for consistency with the old API?

src/parameter_input.cpp
```diff
@@ -350,177 +322,94 @@ bool ParameterInput::ParseLine(InputBlock *pib, std::string line, std::string &n
...
-  InputLine *pl;
 
   for (int i = 1; i < argc; i++) {
     input_text = argv[i];
     std::size_t equal_posn = input_text.find_first_of("=");     // first "=" character
     std::size_t slash_posn = input_text.rfind("/", equal_posn); // last "/" before "="
 
+    // skip if either "/" or "=" do not exist in input
+    if ((slash_posn == std::string::npos) || (equal_posn == std::string::npos)) continue;
+
```
Yurlungur, line 346: Why did this move?

src/parameter_input.cpp
```diff
@@ -350,177 +322,94 @@ bool ParameterInput::ParseLine(InputBlock *pib, std::string line, std::string &n
...
-                                       const std::string &name) {
-  InputLine *pl;
-  InputBlock *pb;
-  pb = GetPtrToBlock(block);
-  if (pb == nullptr) return 0;
-  pl = pb->GetPtrToLine(name);
-  return (pl == nullptr ? 0 : 1);
+bool ParameterInput::DoesParameterExist(const std::string &block,
+                                        const std::string &name) {
+  MarkResolved();
```
Yurlungur: Do we want this to automatically happen?

src/parameter_input.cpp
```diff
@@ -530,33 +419,7 @@ std::string ParameterInput::GetComment(const std::string &block,
...
-    PARTHENON_FAIL(msg);
-  }
-
-  std::string val = pl->param_value;
-
-  CheckAndUpdateQueries_<int>(block, name, docstring);
-
-  // Convert string to integer and return value
-  return stoi(val);
+  return Get<int>(block, name, docstring);
```
Yurlungur, line 422: beautiful

src/parameter_input.cpp
```diff
@@ -1111,14 +762,329 @@ void ParameterInput::OutputParameterTable(std::ostream &os,
...
+      if (i > 0) ss << ", ";
+      ss << (vec[i] ? "true" : "false");
+    }
+  } else if (std::holds_alternative<std::vector<std::string>>(value)) {
+    const auto &vec = std::get<std::vector<std::string>>(value);
+    for (size_t i = 0; i < vec.size(); ++i) {
+      if (i > 0) ss << ", ";
+      ss << vec[i];
+    }
+  }
```
Yurlungur, line 813: Its a little more memory intensive, but why not just store both the converted value and the unresolved string? Then you wouldn't need to do this conversion dance. You can always just access the unresolved string, because you held on to it.

In fact, if you do that, this resolution dance can go away. Just stash a string and a `std::optional<std::any>>` per param. Also eliminates the need for a `std::variant`.

src/parameter_input.cpp
```diff
@@ -1111,14 +762,329 @@ void ParameterInput::OutputParameterTable(std::ostream &os,
...
+  MarkResolved();
+  Parameter *param = FindParameter_(block, name);
+  if (param == nullptr) {
+    return std::nullopt; // Not in storage
+  }
+
+  // If it's an UnresolvedString, convert and cache
+  if (std::holds_alternative<UnresolvedString>(param->value)) {
+    T typed_val = ConvertParamValue<T>(param->value, block, name);
+    param->value = typed_val; // Cache the typed value in the variant
```
Yurlungur, line 953: see my comment above

src/parameter_input.hpp
```diff
@@ -280,44 +316,87 @@ class ParameterInput {
...
+    return def_value;
   }
-  template <typename T, typename... Args>
-  T Get(const ParameterRef &r, Args &&...args) {
-    return Get<T>(r.block_, r.name_, std::forward<Args>(args)...);
+
+  //! RUNTIME OVERRIDE: Programmatically force a parameter value
+  //! Use this for runtime overrides (restart data, command line flags, etc.)
+  //! NOT for parsers - parsers should use AddParsedParameter instead.
+  //! Marks parameter with SetInCode origin, distinguishing it from input file values.
```
Yurlungur, line 371: I am unreasonably amused by `//!` like a siren emoji.

tst/unit/test_parameter_input.cpp
Yurlungur, line 1: good additional tests :+1:

Yurlungur (changes requested):
Pretty nice cleanup. I have some suggestions for how I think this MR could be improved, and I'd like to see them implemented. But I don't feel strongly enough to block the MR so feel free to push back and I'll click approve.

docs/parser_separation_summary.md
lroberts36: sorry, just notice that I didn't pull this out into the other PR. Fix incoming

src/CMakeLists.txt
lroberts36: Another artifact of removing the python

src/utils/type_list.hpp
lroberts36: Yeah, it shouldn't be. I thought another PR removed it though, so I am a little confused how it got back in to develop (since as you say this PR makes no changes there).

src/parthenon_manager.cpp
lroberts36: My thinking was that you want to ensure that you have parsed all the parameter inputs to the code before you start calling `GetOrAdd` and things like that from the code. Otherwise it seems like a mess to track provenance, etc. So by cleanly delineating a parsing phase from a code phase you get that separation.

src/parameter_input.cpp
lroberts36: Yeah, I had to pull everything apart and didn't get every comment. Will fix.

src/parameter_input.cpp
lroberts36: I think that you don't want to be making copies of blocks, so returning it as a pointer suggests no copying. To be fair, this was a choice Claude made, I had the same reaction as you that it wasn't great, but then I came around to the pointer choice.

src/parameter_input.cpp
lroberts36: ah, that move was from the python PR. If you have command line args that are read by the python and args that are read here by parthenon, the old position triggers an error from the python args.

src/parameter_input.cpp
Yurlungur: I agree about not copying. I think the alternative would be a reference, but a pointer is fine.

src/parameter_input.cpp
Yurlungur: ah. Fine to leave this here then. Pre-emptive.

src/parameter_input.cpp
lroberts36: For the python parser for instance we only get the typed value, so we need to populate the string at some point ourselves. 

I think the variant makes sense since we only support a finite set of types.

src/parthenon_manager.cpp
Yurlungur: Ah ok, but `GetOrAdd` and provenance stuff still works?

tst/unit/test_parameter_input.cpp
lroberts36: Really I wish I had started by writing a big unit test suite and then made the changes.

src/parameter_input.hpp
lroberts36: haha

src/parameter_input.cpp
Yurlungur: Ah I see. I was still thinking of the string-based approach only.

Yurlungur (approved):
I have been successfully pushed back on.

src/mesh/mesh.cpp
lroberts36: oh, yeah, forgot about that, good point

tst/unit/test_parameter_input.cpp
Yurlungur: Possibly a lesson we should take to heart with Claude...

src/parthenon_manager.cpp
lroberts36: yeah, `GetOrAdd` and `Set` etc. definitely work. We could switch this to something like `MarkParsingResolved()` if that was clearer.

src/parameter_input.cpp
lroberts36: Well with a reference you could still very easily write `Block my_block = pin.FindBlock_(...)`.

src/parthenon_manager.cpp
Yurlungur: :+1: a more verbose name like that would make me feel better.

src/parameter_input.cpp
lroberts36: Good point, I think this is a `const` function aside from the `MarkResolved()` so it would be safe to do before resolution.

src/parameter_input.cpp
Yurlungur: True. Pointer it is then.

src/parthenon_manager.cpp
lroberts36: Went with `FinalizeParsing()`

src/parameter_input.cpp
lroberts36: removed the parsing resolution here.

lroberts36:
@par-hermes format

docs/parser_separation_summary.md
```diff
@@ -0,0 +1,180 @@
...
+### Rummy Bytecode
+```cpp
+void LoadFromRummy(const std::string& file) {
+  RummyDeck deck(file);
+  deck.compile();
+  for (auto& param : deck.parameters()) {
+    AddParsedParameter(param.suit(), param.card(), param.value());
+  }
+}
+```
```
pdmullen, line 156: My immediate concern from seeing this PR first show up was that it might be causing @c-prather or @adamdempsey90 more work for TOML or Rummy capability... especially since their PRs have been sitting in Parthenon for much longer (albeit they are still marked draft/WIP).  So thanks for including this bit in the docs to show how they might need to adapt their extensions to fit in with the new refactor.  Still, I'd appreciate their reviews on this PR so that we aren't stepping on toes too much.

src/parameter_input.hpp
```diff
@@ -280,44 +316,87 @@ class ParameterInput {
...
-      return GetOrAddString(block, name, value, std::forward<Args>(args)...);
-    } else if (std::is_integral_v<T>) {
-      return GetOrAddInteger(block, name, value, std::forward<Args>(args)...);
-    } else {
-      PARTHENON_THROW("Unknown type\n");
+  // === TEMPLATE INTERFACE (preferred for new code) ===
+
+  //! Get parameter value with compile-time type checking
+  template <typename T>
+  T Get(const std::string &block, const std::string &name,
```
pdmullen, line 326: I didn't know this existed before... (my first time really looking at ParameterInput).  Why don't we use these downstream? I almost prefer the templated versions over the named e.g., `GetString`?

pdmullen (approved):

src/parameter_input.hpp
Yurlungur: history I think lol... no reason not to use the templated versions.

adamdempsey90:
RE: using a vector of blocks instead of a parameter map. Why not just store the ordered block names in a vector (not the parameters themselves), but keep the map as the underlying storage? Maybe I'm misunderstanding

lroberts36:
> RE: using a vector of blocks instead of a parameter map. Why not just store the ordered block names in a vector (not the parameters themselves), but keep the map as the underlying storage? Maybe I'm misunderstanding

I originally had that design, but I didn't like having to track things consistently in two places. I would prefer the map, but it seemed much harder to ensure the backward compatible parameter ordering. Not super wedded to the vectors.

docs/parser_separation_summary.md
lroberts36: Agreed, my thinking with this PR was to try and separate concerns so that it was easier to implement multiple parsers without stepping on toes.

src/outputs/outputs.cpp
```diff
@@ -109,16 +109,14 @@ Outputs::Outputs(Mesh *pm, ParameterInput *pin, SimTime *tm) {
   // the outpus packages (as before), but I don't think it's bad practice to work on
   // `pinput` again here as we're actually processing (potentially even modifying)
   // `pinput`.
-  for (InputBlock *pib = pin->pfirst_block; pib != nullptr; pib = pib->pnext) {
-    if (pib->block_name.compare(0, 16, "parthenon/output") != 0) {
-      continue;
-    }
+  auto output_blocks = pin->GetBlocksWithPrefix("parthenon/output");
```
adamdempsey90: This should maybe be called `GetBlockNamesWithPrefix` to indicate that it is returning names. This is also similar to the other new function `GetBlockNames`

src/parameter_input.cpp
```diff
@@ -259,48 +233,43 @@ void ParameterInput::LoadFromFile(IOWrapper &input) {
...
+//----------------------------------------------------------------------------------------
+//! \fn Block* ParameterInput::FindOrAddBlock_(const std::string & name)
+//  \brief find or add specified Block.  Returns pointer to block.
+
+Block *ParameterInput::FindOrAddBlock_(const std::string &name) {
+  // Search vector to see if name exists, return if found
+  auto it = std::find_if(param_storage_.begin(), param_storage_.end(),
+                         [&name](const Block &b) { return b.name == name; });
+  if (it != param_storage_.end()) return &(*it);
+
```
adamdempsey90: I still think having a "fast path" with a map (in addition to the ordered vector of names) would (1) scale better as parthenon is used on more complicated problems and (2) simplify this repeated searching code.

src/parameter_input.cpp
```diff
@@ -259,48 +233,43 @@ void ParameterInput::LoadFromFile(IOWrapper &input) {
...
-  if (pfirst_block == nullptr) {
-    pfirst_block = pib;
-  } else {
-    plast->pnext = pib; // link new node into list
-  }
+const Block *ParameterInput::FindBlock_(const std::string &name) const {
+  auto it = std::find_if(param_storage_.begin(), param_storage_.end(),
+                         [&name](const Block &b) { return b.name == name; });
+  return (it != param_storage_.end()) ? &(*it) : nullptr;
+}
```
adamdempsey90, line 247: At any point do you check for duplicate block names?

src/parameter_input.cpp
adamdempsey90: Can the python parser be modified to also pass the `repr` of the parameter?

src/parameter_input.hpp
```diff
@@ -43,10 +44,18 @@
...
 
 namespace parthenon {
 
+//----------------------------------------------------------------------------------------
+// Supported parameter types - single source of truth
+//----------------------------------------------------------------------------------------
+using SupportedParamTypes =
+    TypeList<bool, int, Real, std::string, std::vector<bool>, std::vector<int>,
+             std::vector<Real>, std::vector<std::string>>;
+
```
adamdempsey90, line 59: I suppose I expected this to be the base types and not include the vector types (since those are "derived" from the supported base types).

adamdempsey90: Maybe along those lines, why just vector and not also something like `std::array`?  `std::array` might be nice if you want to enforce the length

src/parameter_input.hpp
adamdempsey90: You have to be careful though, since it only really supports a few types, so the other interface is safer

c-prather (approved):
The best PR is the one that actually goes in.

This should make the TOML PR much simpler, since it already has storage and origin tracking done properly (instead of abusing the parsed TOML object).

I will probably still advocate using the TOML parser for Athena-format text files as well when it goes in (by doing the substitutions < to [ etc.), but at least that's a question we can easily resolve either way now.

src/parameter_input.hpp
lroberts36: That is a good point, we should be `assert`ing or `requiring` that `T` in the allowed type list for a cleaner error.

src/parameter_input.cpp
lroberts36: very easily I think, it just seemed a little less coupled to pass the parameter only.

src/parameter_input.hpp
lroberts36: It is easier to keep all of the types in a single list. That being said it would be pretty easy to just define a `TypeList` of the base types, then build a `TypeList` of the vector versions of those automatically, and concatenate the two lists. Nevertheless, I think for the small list of types we have now, it is clearer just to write everything out. 

For `std::array`, we would need to think a little bit since every different size array would be a type in the list. We could move to a model like Jonah had suggested above where we store `std::any` instead of `std::variant`, but that seems like a significant departure from the existing `ParameterInput` model.

src/parameter_input.hpp
lroberts36: oh, actually should have looked at the code. We are `static_asserting` that the type is in the type list.

src/outputs/outputs.cpp
lroberts36: fixed.

src/parameter_input.cpp
lroberts36: Agreed. I switched back to a mixed representation of a vector holding the blocks but a map from names indexing into the vector (and similarly for parameters in the block).

src/parameter_input.cpp
lroberts36: I think we never need to explicitly check because we only add to the vector if the block name wasn't found. Since the internal storage isn't exposed outside the class, we can guarantee this invariant is preserved.

lroberts36:
> The best PR is the one that actually goes in.
> 
> This should make the TOML PR much simpler, since it already has storage and origin tracking done properly (instead of abusing the parsed TOML object).
> 
> I will probably still advocate using the TOML parser for Athena-format text files as well when it goes in (by doing the substitutions < to [ etc.), but at least that's a question we can easily resolve either way now.

@c-prather: Thanks for the review. Do the changes here substantially change your TOML parser or will it be reasonably easy to slot into this framework? Does it mean that there is some functionality you can't provide? One of my goals with this was to make it easy for all of the different parsers people had proposed to be in parthenon without impacting backward compatibility or parsing strategies other people wanted to use.

src/parameter_input.cpp
lroberts36: Thinking about this a little more, I like handling stringification internally in parthenon from a typed input. This guarantees a similar format and we don't have to trust the parser to give us a string that actually represent the stored value.

c-prather:
It looks like they just simplify it, frankly.  I'm quite happy to only be dealing with parsing and not provenance or retrieval -- I wasn't happy with mine or Jonah's implementation, compared to having a new data structure like this.

My last comment was just that the TOML parser ought to be a more reliable way to parse old-format files, too, rather than use the current buggy parser which it looks like you mostly kept.  It's quite delicate, and easily produces incorrect parsing rather than errors, which was part of my motivation for introducing the TOML parser at all.

I'll try to update my code atop this eventually, but I'm not sure I can do that soon, so this should go in first and I'll sort out eliminating duplicate functionality.

lroberts36:
> My last comment was just that the TOML parser ought to be a more reliable way to parse old-format files, too, rather than use the current buggy parser which it looks like you mostly kept. It's quite delicate, and easily produces incorrect parsing rather than errors, which was part of my motivation for introducing the TOML parser at all.

Oh, yeah, definitely. My thinking was that maybe some people relied on "features" of the old parser, so we should keep it as an option rather than force them to rework all their input files. 
 
> I'll try to update my code atop this eventually, but I'm not sure I can do that soon, so this should go in first and I'll sort out eliminating duplicate functionality.

👍

lroberts36:
@par-hermes format

dev/plan_histories/1385.md
```diff
@@ -0,0 +1,180 @@
+# Parameter Parser Separation - Implementation Summary
```
pgrete, line 1: Is this the "plan", which should live in `dev/plan_histories` or sth else?
Right now this file is kind of in the air in the root `docs` directory (so we should either mention/point to it from somewhere or move it to a different loc).

pgrete:
I skimmed the changes and comments. It seems that there's agreement that this can go in "as is" and that the other parser PRs will be updated following the approach here, correct?

pgrete:
@Yurlungur @lroberts36 I don't recall exactly but either one of you wanted to fix the conflict so that we can merge, isn't it?

(plus see me comment above wrt to the location of the "plan")

dev/plan_histories/1385.md
lroberts36: Yes, you are right. I forgot about this, it belongs in the plan directory.

PR #1391, opened 2026-04-23
WIP: Towards Curvilinear Swarms

pdmullen:
## PR Summary
Made in part with generative AI.

Swarm positions are now treated as native mesh coordinates x1, x2, and x3 rather than Cartesian x, y, and z. That change updates the default swarm variable names, the internal swarm coordinate accesses, output defaults, and
  the touched examples/tests so the swarm infrastructure is expressed in coordinate-system-native terms.

 On the remeshing side, the refinement branch now classifies particles against
  the actual daughter split interfaces returned by the mesh topology, expressed
  in x1/x2/x3. For the current uniform coordinate systems this is equivalent to
  the old midpoint logic, but it is clearer and better aligned with the
  generalized coordinate interpretation. We still emit a warning for non-
  UniformCartesian swarms because broader curvilinear swarm behavior remains
  mostly untested.

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build
- [ ] Any contribution that was created or modified with the assistance of generative AI must have a comment disclosing this such as `// This file was made in part with generative AI.`


src/mesh/swarm_amr_remesh.cpp
```diff
@@ -142,10 +142,13 @@ BuildBlockSendPlan(const std::shared_ptr<Swarm> &swarm, const Mesh *pmesh,
...
-    const Real x_mid = 0.5 * (block.xmin(X1DIR) + block.xmax(X1DIR));
-    const Real y_mid = 0.5 * (block.xmin(X2DIR) + block.xmax(X2DIR));
-    const Real z_mid = 0.5 * (block.xmin(X3DIR) + block.xmax(X3DIR));
+    const Real x1_split = pmesh->GetBlockSize(old_loc.GetDaughter(1, 0, 0)).xmin(X1DIR);
+    const Real x2_split =
+        pmesh->ndim > 1 ? pmesh->GetBlockSize(old_loc.GetDaughter(0, 1, 0)).xmin(X2DIR)
+                        : 0.0;
+    const Real x3_split =
+        pmesh->ndim > 2 ? pmesh->GetBlockSize(old_loc.GetDaughter(0, 0, 1)).xmin(X3DIR)
+                        : 0.0;
```
pdmullen, line 151: this is really the only major change.

CHANGELOG.md
```diff
@@ -37,6 +38,7 @@
 
 
 ### Incompatibilities (i.e. breaking changes)
+- [[PR 1391]](https://github.com/parthenon-hpc-lab/parthenon/pull/1391) Towards Curvilinear Swarms
```
Yurlungur, line 55: :+1:

scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
```diff
@@ -62,16 +64,16 @@ def id(self):
         return self.Get("swarm.id")
 
     @property
-    def x(self):
-        return self.Get("swarm.x")
+    def x1(self):
+        return self.Get("swarm.x1")
```
Yurlungur: It's really easy to keep backwards compatibility for the python with maybe a try-catch block and overloading both sets of names for these functions. I think we might want to do that because otherwise phdf won't be backwards compatible with old output/dump files, which we may want to mitigate.

Yurlungur (approved):
:+1: Good change and good use of AI.

One possible suggestion below regarding the python API.

scripts/python/packages/parthenon_tools/parthenon_tools/phdf.py
pdmullen: happy to do that.

pdmullen: Addressed in [9842147](https://github.com/parthenon-hpc-lab/parthenon/pull/1391/commits/9842147d6ec9b7530fb416cb749c125a6bde8daf).

pgrete:
@pdmullen will look into backwards compatibilty (especialyl for outputs)

pgrete (commented):
In general, the changes look good.
Given that OpenPMD outputs are now merged, the new changes need to be also implemented for those outputs (plus keeping hte logic to choose the restart reader).
Should I take a stab at this or do you @pdmullen ?
Also should we add safeguards again in functions that will definitely result in wrong results in non-cartesian coordinates (like `SortParticlesByCell`)?

pdmullen:
> In general, the changes look good. Given that OpenPMD outputs are now merged, the new changes need to be also implemented for those outputs (plus keeping hte logic to choose the restart reader). Should I take a stab at this or do you @pdmullen ? Also should we add safeguards again in functions that will definitely result in wrong results in non-cartesian coordinates (like `SortParticlesByCell`)?

I'm happy to take a stab at it!

pdmullen:
> In general, the changes look good. Given that OpenPMD outputs are now merged, the new changes need to be also implemented for those outputs (plus keeping hte logic to choose the restart reader). Should I take a stab at this or do you @pdmullen ? 

Can you take a look at [834fdea](https://github.com/parthenon-hpc-lab/parthenon/pull/1391/commits/834fdeac643450926a70187bac320b46afe33279) and scream at me if I messed anything up?

> Also should we add safeguards again in functions that will definitely result in wrong results in non-cartesian coordinates (like `SortParticlesByCell`)?

AFAIK, safeguards are not needed for `SortParticlesByCell` or other swarm functions in non-Cartesian coordinates—they should work correctly for `UniformCartesian`, `UniformCylindrical`, and `UniformSpherical` because all use uniform spacing in native coordinates and `Xtoijk` correctly uses `Dx()` which returns that uniform spacing. What would likely break the current design is a non-uniform mesh (e.g., logarithmically-spaced radial grid) where cell spacing varies, since `Xtoijk` assumes constant dx when computing, e.g., `i = floor((x1 - x1_min) / dx)`.  Hope I am not missing something obvious here.

src/outputs/parthenon_opmd.cpp
```diff
@@ -206,19 +208,29 @@ GetParticleRecordAndComponentNames(const std::string &vname, const int rank,
...
   std::string particle_record_component;
 
-  // special sauce to align "positions" with standard
-  if (vname == swarm_position::x::name()) {
+  // Map swarm positions to OpenPMD standard "position" record with x1/x2/x3 components
+  if (vname == swarm_position::x1::name()) {
     particle_record = "position";
-    particle_record_component = "x";
-  } else if (vname == swarm_position::y::name()) {
+    particle_record_component = "x1";
```
pgrete, line 214: I think we may need to be more specific wrt coordinate component naming as the standard has specific names depending on the coordinate system (x,y,z or r,t,z), see https://github.com/openPMD/openPMD-standard/blob/latest/STANDARD.md#naming-conventions and geometry parameter in https://github.com/openPMD/openPMD-standard/blob/latest/STANDARD.md#required-attributes-for-each-mesh-record 

Given that some pieces are still "reserved" my suggestion is to keep x,y,z for Cartesian, default to x1,x2,x3 for non-Cartesian with a warning that the output metadata is currently non-standard compliant and that people wanting to use it in production should get in touch to discuss next steps/a proper implementation.

pgrete (commented):
I still need to test this in practice, but I'm wondering what this make to restart old sims with the new interface with regard to also keep the old naming (so that outputs for a given simulations are consistent in the name being used).
This also applies to HDF5 outputs.
Otherwise, we might want/gave to bump the output format version to reflect those changes.

PR #1394, opened 2026-04-30
token scratch

Yurlungur:
## PR Summary

This MR is motivated by a desire to pre-allocate scratch memory for flat par_for loops. It basically copies how level 1 scratch works in Kokkos with hierarchical loops, where you request a certain amount of memory per thread using Kokkos's unique token interface. Then in the loop body, you can request multiple unmanaged views based on the pre-allocated pool.

The advantage of this over, say, using a variable array as scratch is that it has a smaller memory footprint, one "scratch" per thread vs one per cell, which could be a big difference. We're thinking about changing how our riot loop patterns work (as we discussed in the last call) and this machinery is useful for facilitating cleaner loops for riot and hopefully other codes. I think @pdmullen also mentioned it'll be useful for artemis.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


doc/plan_histories/token_scratch.md
```diff
@@ -0,0 +1,231 @@
...
+## Components
+
+### 1. `TokenScratchPool<MemorySpace, ExecutionSpace, TokenScope>`
+
+The main pool manager that allocates scratch memory for all tokens.
+
+**Template Parameters:**
+- `MemorySpace` - Kokkos memory space (e.g., `HostSpace`, device memory space)
+- `ExecutionSpace` - Kokkos execution space
+- `TokenScope` - (Optional) `UniqueTokenScope::Global` (default) or `UniqueTokenScope::Instance`
```
lroberts36: Slightly confusing placement of `(optional)` and `(default)`

doc/plan_histories/token_scratch.md
```diff
@@ -0,0 +1,231 @@
...
+  KOKKOS_LAMBDA(const member_type& team) {
+    // Each team gets its own token
+    auto scratch = pool.acquire();
+    auto shared_data = scratch.template allocate_view<double>(256);
+    
+    // Team-parallel work with shared scratch
+    Kokkos::parallel_for(Kokkos::TeamThreadRange(team, 256),
+      [&](int i) { shared_data(i) = ...; });
+  });
+```
```
lroberts36: I think there’s a mismatch between the `TeamPolicy` example and what the current implementation guarantees.

As written, `TokenScratchPool::acquire()` just calls `tokens_.acquire()` and doesn’t know anything about the team (e.g. it doesn't pass the `team_member`). In a `TeamPolicy` kernel, the outer lambda runs once per team member, so multiple threads in the same team can call `acquire()` independently and potentially get different token IDs (i.e., different backing buffers) depending on the concurrency of the execution space. That would make the scratch effectively per-thread rather than per-team in the example above.

This also lines up with how the pool is sized (`tokens_.size() * bytes_per_token`), which reflects execution-space concurrency rather than number of teams. For true per-team scratch, I’d expect something along the lines of a `PerTeam` acquisition (or using `AcquireTeamUniqueToken`) so that all threads in a team share the same buffer.

Maybe the example is not really the intended usage though? It looks like there are no hierarchical tests in the unit test suite.

doc/sphinx/src/par_for.rst
```diff
@@ -75,6 +75,27 @@ When ommitted the ``DEFAULT_LOOP_PATTERN`` is used.
...
+Token Scratch Utility
+---------------------
+
+For kernels launched with ``parthenon::par_for`` or directly with Kokkos, Parthenon provides
+``parthenon::TokenScratchPool`` in ``utils/token_scratch.hpp`` for temporary per-thread workspace.
+The pool pre-allocates a fixed number of bytes for each
+``Kokkos::Experimental::UniqueToken``, and ``acquire()`` returns a ``ScratchAllocator`` that can
+carve one or more unmanaged typed ``Kokkos::View`` objects from that token-local buffer. This is
+useful when a kernel needs short-lived scratch storage without performing repeated dynamic
+allocations inside the parallel region.
```
adamdempsey90, line 87: Will a View fence in the dtor? If it doesn't, then the View backing the memory could go out of scope before the kernel finishes because the kernel is async. It might be worth pointing out here if that is a possibility and the user might want to add an execution space fence before the calling routine exits.

src/utils/token_scratch.hpp
```diff
@@ -0,0 +1,193 @@
...
+//========================================================================================
+//! \class ScratchAllocator
+//! \brief RAII wrapper for per-token scratch allocation with type-safe view carving
+//!
+//! Provides stack-like allocation of typed views from a pre-allocated token-specific
+//! buffer. Handles alignment automatically and validates bounds. Token is automatically
+//! released when the allocator goes out of scope.
+//========================================================================================
+template <Kokkos::Experimental::UniqueTokenScope TokenScope =
+              Kokkos::Experimental::UniqueTokenScope::Global,
```
adamdempsey90: There is also `UniqueTokenScope::Instance` which provides a unique token for a stream rather than one per all streams. Not sure which is appropriate here, but I've used `Instance` for a similar capability in another code.

tst/unit/test_token_scratch.cpp
```diff
@@ -0,0 +1,404 @@
...
+    WHEN("We allocate views in parallel iterations") {
+      Kokkos::View<double *> results("results", n_iterations);
+
+      Kokkos::parallel_for(
+          "test_basic", n_iterations, KOKKOS_LAMBDA(const int i) {
+            auto scratch = pool.acquire();
+
+            // Allocate views
+            auto doubles = scratch.template allocate_view<double>(100);
+            auto ints = scratch.template allocate_view<int>(50);
```
adamdempsey90, line 49: In the unit tests, it might be worth verifying that the token id is what you expect given the execution space.

doc/sphinx/src/par_for.rst
Yurlungur: The view does not fence. Yes, good point. I will add a comment.

Yurlungur: Actually---wait no I think views must fence on scoping right? Otherwise it would never be safe to create a view outside of a kernel?

Yurlungur: I added a clarifying comment.

doc/plan_histories/token_scratch.md
Yurlungur: Resolved

doc/plan_histories/token_scratch.md
Yurlungur: Yeah---this doesn't work, and I didn't intend it to. I'm just removing it from the plan history doc. This was generated at the beginning of the effort to build the code and I made a bunch of changes after. 

This raises a few related issues, which maybe are or are not best discussed here:

1. SHOULD something akin to this be made to work? We could make it work by adding a constructor and token type for team scratch so we can use the same API for teams as for flat parallelism. IMO, given that Kokkos already supports something like this, the main benefit is we get to keep the same API.
2. On the other hand, a simpler solution would maybe to just have our allocator object optionally be constructible from a team member, rather than our pool, and then it just is a thin wrapper around the equivalent kokkos API. Might be something we should think about in our loops?
3. How should we handle things like this where the plan history generated by the robot isn't actually in sync with the final product? Updating it by hand as we have here is, I think, sensible... but it might also be useful, for transparency's sake, to show the original artifact for code archeology purposes if needed?

src/utils/token_scratch.hpp
Yurlungur: Maybe I'm misunderstanding---this is something we basically already support right? Because it's just a different choice of this template parameter?

tst/unit/test_token_scratch.cpp
Yurlungur: Good idea. Done.

Yurlungur:
@adamdempsey90 @lroberts36 thanks for the comments! I think they're addressed now. Please re-review.

src/utils/token_scratch.hpp
adamdempsey90: I guess i'm suggesting that Instance is the default template

doc/sphinx/src/par_for.rst
adamdempsey90: That would make sense. I haven't found the dtor implementation of the View to verify though.

src/utils/token_scratch.hpp
Yurlungur: I guess if we only use the default stream, it's identical right? I'll change the default.

Yurlungur: Done.

adamdempsey90 (approved):
Hooking this up to a memory pool would be the next step to increase performance.

Yurlungur:
> Hooking this up to a memory pool would be the next step to increase performance.

Agreed. Perhaps the object pools we already utilize could work for that.

Yurlungur:
I've been doing some performance tuning on an AMD system. For flat loops, this mechanism is very slow due to an unavoidable atomic to synchronize the tokens. I don't know whether or not we want to keep this. Certainly I think the design is useful but may just not be something we want to encourage. Curious folks thoughts.

@adamdempsey90 I think that may at least in part depend on whether or not you want this for artemis.

adamdempsey90:
> I've been doing some performance tuning on an AMD system. For flat loops, this mechanism is very slow due to an unavoidable atomic to synchronize the tokens. I don't know whether or not we want to keep this. Certainly I think the design is useful but may just not be something we want to encourage. Curious folks thoughts.
> 
> @adamdempsey90 I think that may at least in part depend on whether or not you want this for artemis.

We certainly need something like this for our dust coagulation routine in artemis. There's no other way to store potentially 100x100 matrices per thread. That being said, I can always use the ports of call memory pool + my own implementation of this in artemis.

Have you verified that the slowness is due to this token acquisition and not the cost of allocating the view? And is this only slow on AMD systems? It might be worth having even if it's only performant on CPU and CUDA systems

Yurlungur:
> We certainly need something like this for our dust coagulation routine in artemis. There's no other way to store potentially 100x100 matrices per thread. That being said, I can always use the ports of call memory pool + my own implementation of this in artemis.
> 
> Have you verified that the slowness is due to this token acquisition and not the cost of allocating the view? And is this only slow on AMD systems? It might be worth having even if it's only performant on CPU and CUDA systems

It's definitely the token acquisition. That said, I was doing something pretty dumb in my tests---I was basically always acquiring the token, whether or not I needed scratch. That was a pretty massive slowdown.  A very easy/simple fix for that is to just only acquire a token if the number of requested bytes is > 0. I can push that change.

In artemis, and in kernels that are very heavy the performance might not be noticable. It's probably most noticable where only a small amount of work is being done, in which case, the scratch should probably be replaced with another mechanism.

pgrete:
@adamdempsey90 do you want to try this in Artemis?
We'll probably want to wait merging until there's a downstream code usage.

Yurlungur:
@adamdempsey90 @lroberts36 do we still want this? Or should I close it?

adamdempsey90:
> @adamdempsey90 @lroberts36 do we still want this? Or should I close it?

I don't feel strongly either way. It's not much to implement this in artemis if we ever get to it. And if/when we do and if it's helpful, we can just upstream here. But I do believe that there are monster kernels out there that would easily hide the token acquisition cost.

Yurlungur:
>  But I do believe that there are monster kernels out there that would easily hide the token acquisition cost.

Definitely agreed. So maybe it's worth merging. @lroberts36 what do you think?

lroberts36:
> > But I do believe that there are monster kernels out there that would easily hide the token acquisition cost.
> 
> Definitely agreed. So maybe it's worth merging. @lroberts36 what do you think?

I think it makes sense to merge it. It works and we can see where it is a performance advantage and wire it into the loop abstraction

Yurlungur:
OK sounds good. Then I'm removing WIP. If somebody can give it a quick second review. (Thanks @adamdempsey90 for the first one) we can get this in. @lroberts36 you already reviewed once, if you could give it a quick skim and click approve, we can merge.

doc/sphinx/src/par_for.rst
```diff
@@ -75,6 +75,40 @@ When ommitted the ``DEFAULT_LOOP_PATTERN`` is used.
...
+
+.. code:: cpp
+
+   parthenon::TokenScratchPool<> pool(64 * 1024);
+
+   parthenon::par_for(
+       "token_scratch_example", 0, nblocks - 1, KOKKOS_LAMBDA(const int b) {
+         auto scratch = pool.acquire();
+         auto work = scratch.template allocate_view<double>(ni, nj);
```
lroberts36, line 96: Maybe cleaner to add a free function 
```c++
template <class T, class scratch_t, class... Args>
decltype(auto) allocate_view(scratch_t &scratch, Args&&...args) {
  return scratch.template allocate_view<T>(std::forward<Args>(args)...);
}
```
to get rid of the `template` noise.

lroberts36 (approved):
LGTM

doc/sphinx/src/par_for.rst
Yurlungur: Should be called `allocate_scratch_view` to disambiguate but otherwise I agree. Done.

pgrete:
Discussed during sync: Will merge when there's a downstream need.

PR #1401, opened 2026-05-26
Sparse Field Component Names

acreyes:
## PR Summary

Drops the sparse id from component labels for sparse fields when explicit sparse labels are provided, this gives `mass_fraction_hydrogen` rather than `mass_fraction_0_hydrogen`. Also uses the component names in the xdmf outputs for scalars.

Adds a new `sparse_label` member to metadata that can be specified when adding a sparse pool.

```c++
pkg->AddSparsePool("mass_fraction", meta_data, {1,2,3}, {"hydrogen", "aluminum", "helium"});
```

Some bug fixes to openpmd outputs:

* only 3-component vectors get the `x,y,z` hard coded suffix
* sparse field allocation on each level are broadcast so that mesh records are only written for levels that contain that field

This is potentially breaking if anyone downstream is  using the component labels with sparse id in post processing, but I think this aligns better with the behavior for vectors with component labels. Also changing behavior for scalar dense variables that provided a component name. Previously always used the `base_name`, now it would be `base_name_comp`. 

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build


Yurlungur (approved):
My one concern would be if we had sparse vector variables and we wanted something like "hydrogen vx" or similar. This doesn't work with that right?

acreyes:
> My one concern would be if we had sparse vector variables and we wanted something like "hydrogen vx" or similar. This doesn't work with that right?

I think this still works. The [restart test](https://github.com/parthenon-hpc-lab/parthenon/blob/acreyes/component-labels/example/sparse_advection/sparse_advection_package.cpp#L97) does this. You have to specify "hydrogen_vx" and "hydrogen_vy" as the component labels for each vector index separately. The difference is just that before you would have gotten `<base_name>_<sparse_id>_hydrogen_vx`. 

Here is what I get now in that restart test output:
```
   ATTRIBUTE "ComponentNames" {
      DATATYPE  H5T_STRING {
         STRSIZE H5T_VARIABLE;
         STRPAD H5T_STR_NULLTERM;
         CSET H5T_CSET_ASCII;
         CTYPE H5T_C_S1;
      }
      DATASPACE  SIMPLE { ( 14 ) / ( 14 ) }
      DATA {
      (0): "dense_A", "dense_B", "shape_shift_scalar", "shape_shift_vec_x",
      (4): "shape_shift_vec_y", "shape_shift_vec_z", "shape_shift_4_0",
      (7): "shape_shift_4_1", "shape_shift_4_2", "shape_shift_4_3",
      (10): "sparse_0", "sparse_1", "sparse_2", "sparse_3"
      }
   }
```
On `develop` you'll get `shape_shift_3_vec_x`.

Yurlungur:
ah ok sounds good

pgrete (commented):
The changes may also require an update here https://github.com/parthenon-hpc-lab/parthenon/blob/6663f3ac904f2a16e036de402e62ec1811758887/src/outputs/parthenon_opmd.cpp#L279 don't they?
So that for sparse vars the `base` label is removed rather than the `label`?

pdmullen (approved):
This looks good to me, @acreyes 😸 Thanks for the changeset, this is useful and I will definitely be using it!

My only thought (for another PR) is that does this PR somewhat conflate custom sparse ID labels and custom component labels? I wonder if we added some more constructors if we could get an API like this:

```                                                                                                                                                                      
  Metadata velocity_md({Metadata::Sparse, Metadata::Vector},                                                                                                                                                             
                       std::vector<int>{3},  // shape                                                                                                                                                                    
                       std::vector<std::string>{"x", "y", "z"});  // component labels                                                                                                                                    
                                                                                                                                                                                                                                                                                                                                                            
  pkg->AddSparsePool("velocity", velocity_md,                                                                                                                                                                            
                     std::vector<int>{0, 1, 2},  // sparse IDs                                                                                                                                                           
                     std::vector<std::string>{"hydrogen", "helium", "carbon"});  // sparse labels
``` 

Thoughts?

acreyes:
> This looks good to me, @acreyes 😸 Thanks for the changeset, this is useful and I will definitely be using it!
> 
> My only thought (for another PR) is that does this PR somewhat conflate custom sparse ID labels and custom component labels? I wonder if we added some more constructors if we could get an API like this:
> 
> ```
>   Metadata velocity_md({Metadata::Sparse, Metadata::Vector},                                                                                                                                                             
>                        std::vector<int>{3},  // shape                                                                                                                                                                    
>                        std::vector<std::string>{"x", "y", "z"});  // component labels                                                                                                                                    
>                                                                                                                                                                                                                                                                                                                                                             
>   pkg->AddSparsePool("velocity", velocity_md,                                                                                                                                                                            
>                      std::vector<int>{0, 1, 2},  // sparse IDs                                                                                                                                                           
>                      std::vector<std::string>{"hydrogen", "helium", "carbon"});  // sparse labels
> ```
> 
> Thoughts?

This distinction was necessary to get this working with the openpmd outputs. 

I also uncovered a bug in the openpmd sparse variable outputs. If there is a sparse field that isn't allocated at all on a given level the writer still adds a mesh record for this level/variable combination but doesn't point to any actual data. I think this is not standard compliant, or at least `openpmd_api` throws an error. @pgrete please take a look if this makes sense. I added a smoke test that checks that openpmd output from the sparse_advection example produces stand compliant outputs. 

This ended up being a bit bigger than the initial PR fixing a few openpmd things, I'm happy to split those off into a separate PR.

Super cool that the outputs are actually sparse though!
<img width="989" height="833" alt="image" src="https://github.com/user-attachments/assets/df803a52-8e1b-455d-87a2-fe6ed40f529d" />

src/interface/sparse_pool.hpp
pgrete, line 132: 👍

src/outputs/output_utils.hpp
```diff
@@ -148,38 +159,39 @@ struct VarInfo {
...
+
+    component_labels = [&]() -> std::vector<std::string> {
+      const bool has_labels = component_labels_.size() == num_components;
+
+      const auto suffix = [&](const int i) -> std::string {
+        // suffix only if explicit component labels or multi-component variable
+        if (num_components == 1 && !has_labels) return "";
+        return "_" + (has_labels ? component_labels_[i] : std::to_string(i));
+      };
+
```
pgrete, line 178: Haven't seen this init pattern before. Will keep it in mind.

src/outputs/parthenon_opmd.cpp
```diff
@@ -704,6 +706,32 @@ void OpenPMDOutput::WriteOutputFileImpl(Mesh *pm, ParameterInput *pin, SimTime *
...
+          sparse_level_has_data[s_idx * num_levels + level] = 1;
+      }
+    }
+#ifdef MPI_PARALLEL
+    PARTHENON_MPI_CHECK(MPI_Allreduce(MPI_IN_PLACE, sparse_level_has_data.data(),
+                                      static_cast<int>(sparse_level_has_data.size()),
+                                      MPI_INT8_T, MPI_MAX, MPI_COMM_WORLD));
+#endif
+  }
+#endif
```
pgrete, line 733: Interesting that this is required. My interpretation of the doc on collectives that `Mesh` and `Record` ops should be independent in our case (as we just use the ADIOS2 backend), see https://openpmd-api.readthedocs.io/en/0.17.1/details/mpi.html#collective-behavior
I'm happy to keep as is due to limited bandwidth in figuring out where/why this does not seem to be the case.

pgrete (approved):

PR #1403, opened 2026-06-04
Fourier Transforms

lkasselm:
## PR Summary

This PR adds infrastructure for performing distributed Fast Fourier Transforms (FFTs) on uniform meshes in Parthenon. It introduces two new utility classes, a generic spectral output type, and a self-contained example demonstrating correct usage.

Most of this PR (including the documentation) was created in collaboration with LLM chatbots. Claude code refactored the spectrum calculation into its own utility function and created the regression test. 

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


pgrete:
@par-hermes format

example/fourier_transform/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
+#=========================================================================================
+# (C) (or copyright) 2020. Triad National Security, LLC. All rights reserved.
```
Yurlungur: 2026 now. Also the CC should be the Parthenon collaboration, since you (presumably) don't work for Triad.

example/fourier_transform/fourier_driver.cpp
Yurlungur, line 1: This file should have a copyright notice

example/fourier_transform/fourier_driver.hpp
```diff
@@ -0,0 +1,53 @@
+//========================================================================================
+// (C) (or copyright) 2020-2021. Triad National Security, LLC. All rights reserved.
```
Yurlungur: dates wrong, probably shouldn't be Triad

example/fourier_transform/parthinput.example
```diff
@@ -0,0 +1,54 @@
+# ========================================================================================
+#  Athena++ astrophysical MHD code
+#  Copyright(C) 2014 James M. Stone <jmstone@princeton.edu> and other code contributors
+#  Licensed under the 3-clause BSD License, see LICENSE file for details
+# ========================================================================================
```
Yurlungur: since it's new code, the Stone CC is probably not necessary.

example/fourier_transform/parthinput.example
```diff
@@ -0,0 +1,54 @@
...
+#  (C) (or copyright) 2020. Triad National Security, LLC. All rights reserved.
```
Yurlungur: Should be 2026 and Parthenon

src/mesh/mesh.hpp
```diff
@@ -93,6 +96,26 @@ class Mesh {
...
+
+  std::unique_ptr<parthenon::UniformGridHelper> uniform_grid_helper;
+
+  UniformGridHelper *GetUniformGridHelper() {
+    if (!uniform_grid_helper) {
+      uniform_grid_helper = std::make_unique<UniformGridHelper>(this);
+      uniform_grid_helper->Initialize();
+    }
+    return uniform_grid_helper.get();
+  }
```
Yurlungur, line 119: Why tie this to the mesh rather than let it be free-floating?

src/utils/uniform_grid_helper.cpp
Yurlungur, line 1: File needs a copyright notice

src/mesh/uniform_grid_helper.cpp
```diff
@@ -0,0 +1,169 @@
+#include "uniform_grid_helper.hpp"
+#include "mesh/mesh.hpp"
+
+namespace parthenon { // Should this be in parthenon namespace?
```
Yurlungur: yeah probably

src/mesh/uniform_grid_helper.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+  initialized_ = true;
+
+  // debug: print local mesh box info
+  std::cout << "Rank " << parthenon::Globals::my_rank << " local mesh box: low = ("
+            << LocalMeshBox.low[0] << ", " << LocalMeshBox.low[1] << ", "
+            << LocalMeshBox.low[2] << "), high = (" << LocalMeshBox.high[0] << ", "
+            << LocalMeshBox.high[1] << ", " << LocalMeshBox.high[2] << "), size = ("
+            << LocalMeshBox.size[0] << ", " << LocalMeshBox.size[1] << ", "
+            << LocalMeshBox.size[2] << ")\n";
```
Yurlungur: Do you always want this output?

src/mesh/uniform_grid_helper.cpp
```diff
@@ -0,0 +1,169 @@
+#include "uniform_grid_helper.hpp"
+#include "mesh/mesh.hpp"
+
+namespace parthenon { // Should this be in parthenon namespace?
+
+UniformGridHelper::UniformGridHelper(Mesh *mesh) : mesh_(mesh) {}
+
+void UniformGridHelper::Initialize() {
+  if (initialized_) return;
```
Yurlungur: It may be better to just move initialize into the constructor and then you don't need this flag.

CMakeLists.txt
```diff
@@ -512,4 +530,4 @@ endif()
 #
 # If you add tests with other binaries, add them to the DEPENDS line
 add_custom_target(checkit COMMAND ${CMAKE_CTEST_COMMAND}
-                  DEPENDS unit_tests)
+                  DEPENDS unit_tests)
```
Yurlungur: missing newline

CMakeLists.txt
```diff
@@ -491,7 +493,23 @@ set(CMAKE_INSTALL_INCLUDEDIR "${CMAKE_INSTALL_INCLUDEDIR}/parthenon")
...
 add_subdirectory(src)
+
+# Now parthenon target exists — safe to link
+if (PARTHENON_ENABLE_HEFFTE)
+  if (TARGET Heffte)
+    target_link_libraries(parthenon PRIVATE Heffte)
+  else()
+    message(FATAL_ERROR "HeFFTe target not found after add_subdirectory")
+  endif()
+endif()
```
Yurlungur: It is good practice to hide all of the HeFFTe stuff inside a target guard so that if a parent code that uses parthenon loads HeFFTe itself with different configuration, parthenon will use the choices of the parent code. e.g.,

```
if (PARTHENON_ENABLE_HEFFTE)
  if (NOT TARGET Heffte)
    set(Heffte_ENABLE_CUDA ${Kokkos_ENABLE_CUDA} CACHE BOOL "Enable CUDA for HeFFTe" FORCE)
    add_subdirectory(external/heffte)
  endif()
endif()
```

src/CMakeLists.txt
```diff
@@ -1,4 +1,8 @@
 #=========================================================================================
+# Parthenon performance portable AMR framework
+# Copyright(C) 2020-2025 The Parthenon collaboration
```
Yurlungur: 2026

src/CMakeLists.txt
```diff
@@ -356,6 +366,17 @@ target_compile_definitions(parthenon PRIVATE
     FS_NAMESPACE=${CXX_FILESYSTEM_NAMESPACE}
     )
 
+find_package(CUDAToolkit REQUIRED)
```
Yurlungur: This should not be global, because we build for more architectures than just Cuda.

src/utils/FFTManager.hpp
```diff
@@ -0,0 +1,88 @@
...
+  std::unique_ptr<Impl> impl_; // owns backend-specific data
+
+  Mesh *mesh_;
+  bool initialized_ = false;
+
+  // Global mesh dimensions, stored during Initialize()
+  int Nx_ = 0, Ny_ = 0, Nz_ = 0;
+};
+
+} // namespace parthenon
```
Yurlungur: missing newline

src/utils/FFTManager.hpp
```diff
@@ -0,0 +1,88 @@
+#pragma once
```
Yurlungur: We use include guards of this style
```C++
#ifndef UTILS_FFTMANAGER_
#define UTILS_FFTMANAGER_

/* blah */

#endif // UTILS_FFTMANAGER_
```

src/utils/fft_manager.hpp
Yurlungur, line 1: This file needs a copyright notice

src/utils/FFTManager.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+std::size_t FFTManager::size_real_space_box() const {
+  const auto box = real_space_box();
+  return static_cast<std::size_t>(box.size[0]) * static_cast<std::size_t>(box.size[1]) *
+         static_cast<std::size_t>(box.size[2]);
+}
+
+FFTManager::~FFTManager() = default;
+
+} // namespace parthenon
```
Yurlungur: missing newline

src/utils/fft_manager.cpp
Yurlungur, line 1: This file is missing a copyright notice

src/outputs/spectrum.cpp
```diff
@@ -0,0 +1,212 @@
+//========================================================================================
+// Parthenon performance portable AMR framework
+// Copyright(C) 2023-2025 The Parthenon collaboration
+// Licensed under the 3-clause BSD License, see LICENSE file for details
+//========================================================================================
+// (C) (or copyright) 2020-2025. Triad National Security, LLC. All rights reserved.
```
Yurlungur: 2026 :)

src/outputs/spectrum.cpp
```diff
@@ -0,0 +1,212 @@
...
+    }
+    fout.close();
+  }
+
+  // advance output parameters
+  UpdateNextOutput_(pm, tm);
+
+} // void SpectralOutput::WriteOutputFile
+
+} // namespace parthenon
```
Yurlungur: missing newline

src/outputs/spectrum.cpp
```diff
@@ -0,0 +1,212 @@
...
+    if (!fout.is_open()) {
+      PARTHENON_FAIL("Could not open " + fname + " for writing");
+    }
+
+    // Write each bin's results to the file: val_sum, k_sum, count_sum
+    fout << "# Bin    val_sum    K_sum    Count\n";
+    for (int i = 0; i < num_bins; ++i) {
+      fout << i << " " << spectra_h(i, 0) << " " << spectra_h(i, 1) << " "
+           << spectra_h(i, 2) << "\n";
+    }
```
Yurlungur, line 61: Does this have the proper precision? Might be better to use `std::format` now that we're on C++20?

src/outputs/helicity_slice.cpp
```diff
@@ -0,0 +1,376 @@
+//========================================================================================
+// Parthenon performance portable AMR framework
+// Copyright(C) 2023-2025 The Parthenon collaboration
+// Licensed under the 3-clause BSD License, see LICENSE file for details
+//========================================================================================
+// (C) (or copyright) 2020-2025. Triad National Security, LLC. All rights reserved.
```
Yurlungur: 2026

src/outputs/helicity_slice.cpp
```diff
@@ -0,0 +1,376 @@
...
+
+  it.seriesFlush();
+  it.close();
+  series.close();
+
+  UpdateNextOutput_(pm, tm);
+
+} // void HelicitySliceOutput::WriteOutputFile
+
+} // namespace parthenon
```
Yurlungur: missing newline

src/outputs/helicity_slice.cpp
Yurlungur, line 1: I may have missed it, but I don't think I fully understand the helicity slice output. What is this for? I don't think I saw it documented in the docs

Yurlungur (changes requested):
Very cool! Thanks for adding this!

src/mesh/mesh.hpp
lkasselm: Because this way, they only have to be initialized once, exist for the entire lifetime of the mesh, and can be easily accessed from anywhere in a downstream application. It also makes sense conceptually, because (at least for now) the FFTs always operate on the entire mesh.

lkasselm: If memory footprint is a concern, we could also add a Destroy() method to these classes, so that users can delete them when they are not needed anymore.

doc/sphinx/src/fourier_transforms.rst
```diff
@@ -0,0 +1,314 @@
...
+
+   auto &md   = pmesh->mesh_data.Get();
+   auto cons = md->PackVariables(std::vector<std::string>{"cons"});
+
+   auto &mbb = uniformGridHelper->MeshBlockBox; // interior cell bounds within a meshblock
+
+   auto helper = uniformGridHelper->GetKernelHelper();
+
+   parthenon::par_for(
+       "GatherVelocity", 0, pmesh->GetNumMeshBlocksThisRank() - 1,
```
pgrete: While in practice it doesn't make a difference given that `pack_size=-1` is enforced, it's probably cleaner/clearer to use the number of blocks from the `MeshData` object, i.e., `md->NumBlocks()`

doc/sphinx/src/fourier_transforms.rst
```diff
@@ -0,0 +1,314 @@
+Fourier Transforms
+==================
+
+Parthenon provides infrastructure for performing distributed Fast Fourier Transforms (FFTs)
+on uniform meshes via the :cpp:class:`FFTManager` and :cpp:class:`UniformGridHelper` classes.
+These are built on top of `heFFTe <https://github.com/icl-utk-edu/heffte>`_ and support
+both CPU and GPU backends transparently.
+
+.. note::
+   FFT functionality requires ``pack_size = -1`` in the input file, meaning all meshblocks
```
pgrete: I personally prefer to use the new `num_packs=1` (rather than `pack_size=-1`) as it's clearer (though technically identical)

doc/sphinx/src/fourier_transforms.rst
```diff
@@ -0,0 +1,314 @@
...
+      parthenon::par_for(...,
+          KOKKOS_LAMBDA(...) {
+              // use output_kk[idx], not output(idx)
+              auto val = output_kk[idx] * some_kokkos_complex;
+          });
+
+   ``Kokkos::complex`` and ``std::complex`` have identical memory layouts, so the
+   reinterpret cast is safe. The cast must be done **before** the lambda — capturing
+   a ``ParArray1D<std::complex<Real>>`` and calling ``.data()`` inside the kernel
+   will not work on GPU.
```
pgrete: Are there any issues working with `Kokkos::complex<Real>` from the beginning and pass the data pointer of an `ParArray1D<Kokos::complex<Real>>` to heFFTe?

doc/sphinx/src/fourier_transforms.rst
```diff
@@ -0,0 +1,314 @@
...
+       "FourierSpaceKernel",
+       outbox.low[2], outbox.high[2],
+       outbox.low[1], outbox.high[1],
+       outbox.low[0], outbox.high[0],
+       KOKKOS_LAMBDA(const int kz_idx, const int ky_idx, const int kx_idx) {
+
+           const auto idx = fft_helper.FourierFlatIndex(kz_idx, ky_idx, kx_idx);
+
+           // integer wavevector components (negative frequencies unwrapped)
+           auto [kx, ky, kz] = fft_helper.Wavevector(kz_idx, ky_idx, kx_idx);
```
pgrete: Most of the Parthenon base is agnostic to x/y/z (so it's x1/x2/x3, ijk).
I see why xzy feels more natural/suitable here but I'm wondering if this should be changed for consistency (see also recent discussion on particle xyz positions).
Any thoughts @Yurlungur

example/fourier_transform/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+# for the U.S. Department of Energy/National Nuclear Security Administration. All rights
+# in the program are reserved by Triad National Security, LLC, and the U.S. Department
+# of Energy/National Nuclear Security Administration. The Government is granted for
+# itself and others acting on its behalf a nonexclusive, paid-up, irrevocable worldwide
+# license in this material to reproduce, prepare derivative works, distribute copies to
+# the public, perform publicly and display publicly, and to permit others to do so.
+#=========================================================================================
+
+get_property(DRIVER_LIST GLOBAL PROPERTY DRIVERS_USED_IN_TESTS)
+if(NOT PARTHENON_DISABLE_EXAMPLES)
```
pgrete: Probably needs another guard to only compile/include when heFFTe is available.

example/fourier_transform/fourier_driver.cpp
```diff
@@ -0,0 +1,123 @@
...
+    driver.Execute();
+  }
+  pman.ParthenonFinalize();
+  return 0;
+}
+
+// Initialize a simple test field. Note that FFTs only work on a uniform grid, no AMR.
+void FillTestField(MeshBlock *pmb, ParameterInput *pin) {
+  auto &rc = pmb->meshblock_data.Get();
+  auto field = rc->Get("test_field").data;
```
pgrete: We're trying to discourage the use of "per-block" kernels (in favor of `MeshData` over many blocks).
Just mentioning this here for future reference, not requesting to fix for the PR.

example/fourier_transform/fourier_driver.cpp
```diff
@@ -0,0 +1,123 @@
...
+  // Reduce across MPI ranks to get global maximum
+  Real max_error = local_max_error;
+#ifdef MPI_PARALLEL
+  PARTHENON_MPI_CHECK(MPI_Allreduce(&local_max_error, &max_error, 1, MPI_PARTHENON_REAL,
+                                    MPI_MAX, MPI_COMM_WORLD));
+#endif
+
+  if (parthenon::Globals::my_rank == 0) {
+    std::cout << "Max relative error after FFT round-trip: " << max_error << std::endl;
+  }
```
pgrete, line 150: Given that this already does a check, I suggest to add the test to the regression suite by just executing this example and parse the result for the `max_error`.
Let me know if you need further instruction on adding the test.

example/fourier_transform/fourier_driver.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * @brief Constructs a driver to demonstrate the use of Fourier transforms in Parthenon.
+ * The driver will compute the Fourier transform of a 3D array, compute the inverse
+ * Fourier transform to recover the original array, and compute the maximum error between
+ * the original and recovered arrays.
+ */
+class FourierDriver : public Driver {
+ public:
+  FourierDriver(ParameterInput *pin, ApplicationInput *fin, Mesh *pm)
```
pgrete: `ApplicationInput *fin`
Why `fin`? I'm confused by the naming.
It's practically `app_in` everywhere else.

example/fourier_transform/fourier_driver.hpp
```diff
@@ -0,0 +1,53 @@
...
+  // but each driver is expected to implement at least one of them.
+  /// TaskList MakeTaskList(MeshBlock *pmb);
+  template <typename T>
+  TaskCollection MakeTaskCollection(T &blocks);
+
+  /// `Execute` cycles until simulation completion.
+  DriverStatus Execute() override;
+
+ protected:
+  void FourierPostExecute(Real max_error);
```
pgrete: Is this called anywhere?

example/fourier_transform/parthinput.example
```diff
@@ -0,0 +1,54 @@
...
+
+nx3 = 16
+x3min = -2.0
+x3max = 2.0
+ix3_bc = periodic
+ox3_bc = periodic
+
+# How many meshblocks to use in a premade default kernel.
+# A value of <1 means use the whole mesh.
+pack_size = -1
```
pgrete: See above -> `num_packs=1`

src/mesh/uniform_grid_helper.hpp
```diff
@@ -0,0 +1,63 @@
+#pragma once
```
pgrete: See below on the usage of include guards

pgrete: Also needs copyright header.

src/mesh/uniform_grid_helper.hpp
```diff
@@ -0,0 +1,63 @@
...
+#include "parthenon_arrays.hpp"
```
pgrete: linting will probably fail.
We first use C++ header and then Parthenon headers.

src/mesh/uniform_grid_helper.cpp
pgrete: I think it's actually useful to see how all the rank local blocks have been rearranged to a dense large block.
So I'm fine with keeping it (as it's also only printed once).

src/mesh/mesh.hpp
pgrete: But there's no technical reason, is there?

I'm wondering with regard to leaking heFFTe dependencies into `Mesh` given that most of the Parthenon builds will not contain `heFFTe`, so I'm also leaning towards an implementation can selectively be included/instantiated.

src/mesh/mesh.hpp
```diff
@@ -50,10 +50,12 @@
 #include "mesh/forest/forest.hpp"
 #include "mesh/forest/forest_topology.hpp"
 #include "mesh/meshblock_pack.hpp"
+#include "mesh/uniform_grid_helper.hpp"
 #include "outputs/io_wrapper.hpp"
 #include "pack/sparse_pack/pack_descriptor.hpp"
 #include "parameter_input.hpp"
 #include "parthenon_arrays.hpp"
+#include "utils/FFTManager.hpp"
```
pgrete: depending on where these finally live, these should be included conditionally

src/mesh/uniform_grid_helper.cpp
```diff
@@ -0,0 +1,169 @@
...
+      (local_loc_max.at(1) - local_loc_min.at(1)) + 1,
+      (local_loc_max.at(2) - local_loc_min.at(2)) + 1,
+  };
+  const auto loc_max_vol = local_nlocs.at(0) * local_nlocs.at(1) * local_nlocs.at(2);
+
+  PARTHENON_REQUIRE_THROWS(
+      loc_max_vol == mesh_->GetNumMeshBlocksThisRank(),
+      "Block coverage on rank cannot be matched to a contiguous "
+      "array, which is required for FFTs. Try a different amount of "
+      "ranks (one block per rank, i.e. pack_size=-1, will always work).");
```
pgrete: The last line of the comment is mixed up.
One block per rank will always work, but this is independent of pack_size=-1 (which just guarantees that there's a single pack per rank -- which is not always guaranteed to work, for example, if blocks are unevenly distributed).

src/utils/uniform_grid_helper.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+void UniformGridHelper::GatherField(const std::string &var_name, int var_index,
+                                    parthenon::ParArray1D<Real> &output) {
+
+  // Check that var_name and var_index correspond to a valid variable in the mesh data and
+  // that output array is large enough to hold the gathered data.
+  auto &md = mesh_->mesh_data.Get();
+  auto vars = md->PackVariables(std::vector<std::string>{var_name});
+  PARTHENON_REQUIRE_THROWS(vars.GetDim(5) > 0, "GatherField: variable '" + var_name +
+                                                   "' not found in mesh data");
```
pgrete, line 121: Interesting. I somehow expected the pack creation to fail if some var is not present.

src/mesh/uniform_grid_helper.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+  IndexRange ib = md->GetBlockData(0)->GetBoundsI(IndexDomain::interior);
+  IndexRange jb = md->GetBlockData(0)->GetBoundsJ(IndexDomain::interior);
+  IndexRange kb = md->GetBlockData(0)->GetBoundsK(IndexDomain::interior);
+
+  const auto vi = var_index;
+  auto helper = GetKernelHelper();
+
+  parthenon::par_for(
+      "UniformGridHelper::GatherField", 0, mesh_->GetNumMeshBlocksThisRank() - 1, kb.s,
```
pgrete: See above on using the info on the number of blocks from `MeshData` rather than `Mesh`.

src/mesh/uniform_grid_helper.hpp
```diff
@@ -0,0 +1,63 @@
...
+// Assuming a uniform grid, this class helps gather information about the grid layout
+// across all meshblocks on a rank.
+class UniformGridHelper {
+
+ public:
+  explicit UniformGridHelper(Mesh *mesh);
+  void Initialize();
+
+  Box3D MeshBlockBox;
+  Box3D LocalMeshBox;
```
pgrete: variable names are typically lowercase with `_` rather than CamelCase

src/outputs/helicity_slice.cpp
pgrete: I think that's a leftover from AthenaPK downstream usage and should not be included upstream.

src/outputs/outputs.cpp
```diff
@@ -359,6 +359,8 @@ Outputs::Outputs(Mesh *pm, ParameterInput *pin, SimTime *tm) {
       pnew_type = std::make_shared<HistoryOutput>(op);
     } else if (op.file_type == "ascent") {
       pnew_type = std::make_shared<AscentOutput>(op);
+    } else if (op.file_type == "spectrum") {
+      pnew_type = std::make_shared<SpectralOutput>(op);
```
pgrete, line 366: @Yurlungur what are your thoughts wrt including a rather "specific" output type upstream by default?

I was wondering if the route to include it in the `fourier_transform` example via #1362 might not be a cleaner approach?

Alternatively, I could also imagine that calculating spectra could live in `utils` so that it can be reused for calculations (and independent of outputs, which would also allow downstream codes to define their own preferred output format for spectra with much reduced code as the spectra calculation is shared).

I'm leaning towards the latter option.

src/outputs/spectrum.cpp
```diff
@@ -0,0 +1,212 @@
...
+  // Initialize FFTManager and I/O arrays:
+  int n_comp = components.size(); // number of field components to transform
+  auto FFTManager = pm->GetFFTManager();
+  const auto fft_size_inbox = FFTManager->size_real_space_box();
+  parthenon::ParArray1D<Real> input("fft input", n_comp * fft_size_inbox);
+  parthenon::ParArray1D<std::complex<Real>> output(
+      "fft output", n_comp * FFTManager->size_fourier_space_box());
+  PARTHENON_REQUIRE_THROWS(
+      pm->DefaultNumPartitions() == 1,
+      "Only pack_size=-1 currently supported for heffte.") // pack size -1 means 1 pack
```
pgrete: See above on `num_packs=1`

src/utils/fft_manager.cpp
```diff
@@ -0,0 +1,122 @@
+#include "FFTManager.hpp"
+#include "heffte.h"
+#include "mesh/mesh.hpp"
+
+namespace parthenon {
+
+struct FFTManager::Impl {
+// @pgrete: Can the backend selection be made nicer?
+#if defined(KOKKOS_ENABLE_CUDA) || defined(KOKKOS_ENABLE_HIP)
```
pgrete, line 20: Probably needs another branch to check for sycl and raise an error that it's untested.

src/utils/fft_manager.cpp
```diff
@@ -0,0 +1,122 @@
...
+FFTManager::FFTManager(Mesh *mesh) : mesh_(mesh) {}
+
+void FFTManager::Initialize() {
+  if (initialized_) return;
+
+  auto UniformGridHelper = mesh_->GetUniformGridHelper();
+
+  auto mesh_size = mesh_->mesh_size;
+  Nx_ = mesh_size.nx(X1DIR);
+  Ny_ = mesh_size.nx(X2DIR);
```
pgrete: lowercase var names for style consistency

src/utils/fft_manager.cpp
```diff
@@ -0,0 +1,122 @@
...
+  using BackendTag = heffte::backend::default_backend<heffte::tag::cpu>::type;
+#endif
+
+  heffte::fft3d_r2c<BackendTag> fft_plan;
+  ParArray1D<std::complex<double>> workspace_;
+
+  Impl(const heffte::box3d<> &real_space_box, const heffte::box3d<> &fourier_space_box,
+       int r2c_direction, MPI_Comm comm)
+      : fft_plan(real_space_box, fourier_space_box, r2c_direction, comm),
+        workspace_("fft workspace", fft_plan.size_workspace()) {}
```
pgrete: Eventually, it'd be nice to pass some parameters as, for example, the comm patter is performance critical (order or magnitude at scale), but  we can skip that for now.

In an even more ideal world, we'd do a few tiny benchmarks and then auto initialize the best pattern, but that's definitely beyond scope of the PR.

src/utils/FFTManager.hpp
```diff
@@ -0,0 +1,88 @@
...
+
+#include "mesh/uniform_grid_helper.hpp"
+#include "parthenon_arrays.hpp"
+#include <complex>
+#include <memory>
```
pgrete: include order

src/utils/fft_manager.hpp
```diff
@@ -0,0 +1,88 @@
...
+  // Device-copyable kernel helper
+  // -----------------------------
+  struct KernelHelper {
+    parthenon::Box3D fourier_box;
+    parthenon::Box3D real_box;
+    int Nx, Ny, Nz;
+
+    // Flat index into the local Fourier-space array
+    KOKKOS_INLINE_FUNCTION
+    std::int64_t FourierFlatIndex(const int k, const int j, const int i) const {
```
pgrete, line 53: 👍 for directly using int64_t indices

src/CMakeLists.txt
```diff
@@ -356,6 +366,17 @@ target_compile_definitions(parthenon PRIVATE
...
+
+target_link_libraries(parthenon
+  PRIVATE
+    CUDA::cufft
+)
```
pgrete: Should not be required at all as it should come in via heFFTe

src/CMakeLists.txt
```diff
@@ -195,6 +199,8 @@ add_library(parthenon
   mesh/meshblock.cpp
   mesh/swarm_amr_remesh.cpp
   mesh/swarm_amr_remesh.hpp
+  mesh/uniform_grid_helper.cpp 
+  mesh/uniform_grid_helper.hpp
```
pgrete: These may need to be conditionally included.

CMakeLists.txt
```diff
@@ -75,6 +75,8 @@ else()
     OFF)
 endif()
 
+option(PARTHENON_ENABLE_HEFFTE "Enable HeFFTe FFT backend" ON)
```
pgrete: Default should probably be off.

CMakeLists.txt
```diff
@@ -491,7 +493,23 @@ set(CMAKE_INSTALL_INCLUDEDIR "${CMAKE_INSTALL_INCLUDEDIR}/parthenon")
 set(DOC_GEN_PATH "${CMAKE_SOURCE_DIR}/doc/sphinx/src/generated" CACHE STRING
   "Path to save generated data for docs.")
 
+# Add HeFFTe first if enabled
+if (PARTHENON_ENABLE_HEFFTE)
+  set(Heffte_ENABLE_CUDA ${Kokkos_ENABLE_CUDA} CACHE BOOL "Enable CUDA for HeFFTe" FORCE)
```
pgrete: Needs logic for HIP, too.

CMakeLists.txt
```diff
@@ -491,7 +493,23 @@ set(CMAKE_INSTALL_INCLUDEDIR "${CMAKE_INSTALL_INCLUDEDIR}/parthenon")
 set(DOC_GEN_PATH "${CMAKE_SOURCE_DIR}/doc/sphinx/src/generated" CACHE STRING
   "Path to save generated data for docs.")
 
+# Add HeFFTe first if enabled
+if (PARTHENON_ENABLE_HEFFTE)
```
pgrete: I suggest to use the pattern similar to what's currently done in the OpenPMD branch, i.e., also allow system-installed heffte libraries.
Similarly, going the route via `FetchContent` rather than `submodule` might be cleaner as it can be selectively included only when people ask for heffte to be included (rather than everyone pulling a heffte submodule right now).

pgrete (changes requested):
Thanks for PR. This definitely already looks pretty good.
Given that this is your first PR I was a little more verbose with respect to style/conventions etc so that these are items to have in mind right from the beginning for future PRs.

In addition to the inline comments, this PR also
- needs a changelog
- add heFFTe as optional dependency in the README

src/mesh/mesh.hpp
lkasselm: No, I think in principle they could be free-floating. But then they would have to be re-initialized everytime they are called in a new function right? Is there another way to make them more long-lived that doesnt involve tying them to the mesh? 

Also, the FFT initialisation is lazy and all the heFFTe stuff is hidden via pimpl, so I think dependency leaking shoudnt be a problem?

Yurlungur: Is the re-initialization upon free-floating instantiation a big deal, performance-wise? On the parthenon side, it seems to be a few small arrays. On HeFFTE's side, how big is the overhead? If the performance hit is small, I lean towards making it free-floating and just re-instantiating it when needed. (You could use things like static to cache locally in user code.) I prefer this because it's simpler from a code complexity and inter-dependency perspective. Dependencies leaking aren't just about functionality but about code comprehension for new developers and maintainers.

But on the other hand, if initialization is a performance hit, then we should be more careful and your current way of doing things might be appropriate.

doc/sphinx/src/fourier_transforms.rst
Yurlungur: I don't feel strongly but I think I agree that I lean towards x3, x2, x1. Also I notice the order of the structured binding return is backwards from the order passed in. Is that intentional? It's a bit unintuitive IMHO.

src/mesh/uniform_grid_helper.cpp
Yurlungur: :+1: Sounds good. Maybe worth revisiting if we change to free-floating or make initialization not lazy. Because then this codepath may be called many times.

src/mesh/mesh.hpp
pgrete: Arrays/views are probably not an issue in practice, but creating an FFT plan is expensive. So that'd have to be cached anyway.

Shouldn't it work to create a `static` FFTManager given that the only use case for now is to support one "configuration/plan" per run?

src/outputs/outputs.cpp
Yurlungur: I lean towards agreeing that I would prefer a functionality that lives in utils and lets downstreams pick their own way of outputting this data. 

However, I am not married to that. If an automated output is sufficiently broadly useful then that can motivate this current structure.

src/utils/fft_manager.cpp
Yurlungur: I agree, but I wonder: can we unit test the FFT machinery? Especially if it gets decoupled from the mesh object?

src/mesh/mesh.hpp
Yurlungur: I see. Ok then perhaps tying to the mesh makes sense because that conveys the intent that it's a "one-off" thing.

Static *should* work for performance however.

lkasselm: A limitation of a static FFTManager would be that there's no way to free the workspace memory when FFTs are no longer needed.

To partially adress the dependency concern, we could make the mesh tie-in conditional on a CMake flag: 

```
// mesh.hpp
#ifdef PARTHENON_ENABLE_FFT
  std::unique_ptr<parthenon::FFTManager> fft_manager;
  FFTManager *GetFFTManager() {
    if (!fft_manager) {
      fft_manager = std::make_unique<FFTManager>(this);
      fft_manager->Initialize();
    }
    return fft_manager.get();
  }

  std::unique_ptr<parthenon::UniformGridHelper> uniform_grid_helper;
  UniformGridHelper *GetUniformGridHelper() {
    if (!uniform_grid_helper) {
      uniform_grid_helper = std::make_unique<UniformGridHelper>(this);
      uniform_grid_helper->Initialize();
    }
    return uniform_grid_helper.get();
  }
#endif
```

lkasselm:
@par-hermes format

lkasselm:
@par-hermes format --amend

lkasselm:
@par-hermes format --amend

example/fourier_transform/fourier_driver.hpp
lkasselm: Nope. Removed.

example/fourier_transform/CMakeLists.txt
lkasselm: I put the guard into example/CMakeLists.txt.

src/utils/uniform_grid_helper.cpp
lkasselm: If I remember correctly, it didnt fail but gave me vars.GetDim(5) = 0.

src/outputs/outputs.cpp
lkasselm: So lets keep it like this for now and change it when #1362 is implemented?

src/mesh/mesh.hpp
Yurlungur: I think I would prefer it optional like this. Given that there *are* performance and memory reasons to tie it to the mesh, we can do that for now.

An alternative would be to make it a singleton class but not tied to the mesh. That feels more natural to me but is more complicated C++.

src/outputs/outputs.cpp
Yurlungur: That's fine with me.

lkasselm:
Should uniform_grid_helper live in mesh/ ? Or should we move it into utils/ ?

src/utils/fft_manager.cpp
lkasselm: Like this? Maybe should be a warning instead of an error? 

```
#if defined(KOKKOS_ENABLE_CUDA)
    using BackendTag = heffte::backend::default_backend<heffte::tag::gpu>::type;
#elif defined(KOKKOS_ENABLE_HIP)
    #warning "FFTManager has not been tested with HIP/ROCm. Proceed with caution."
    using BackendTag = heffte::backend::default_backend<heffte::tag::gpu>::type;
#elif defined(KOKKOS_ENABLE_SYCL)
    #warning "FFTManager has not been tested with SYCL. Proceed with caution."
    using BackendTag = heffte::backend::default_backend<heffte::tag::cpu>::type;
#else
    using BackendTag = heffte::backend::default_backend<heffte::tag::cpu>::type;
#endif
```

Yurlungur:
> Should uniform_grid_helper live in mesh/ ? Or should we move it into utils/ ?

My vote is utils but it doesn't matter much.

doc/sphinx/src/fourier_transforms.rst
lkasselm: Its not possible to pass a data pointer of an `ParArray1D<Kokos::complex<Real>>` directly to heFFTe. But I now put the cast into the FFTManager::Forward() and Backward() functions so that users can work with Kokkos type arrays the entire time.

pgrete (approved):

pgrete:
I just pushed some changes to the PR, in particular
- make the machinery work with `Real=float`
- separating the precision in the spectra calculation (so that these are always calculated in `SpecReal = double`, which is important for the big boxes in single precision output)
- Fixed compilation/linking when `heffte` is externally used
- added test case to the `CalcSpectrum` function (now within the `utils::fft` namespace), which implicitly tests the Hermitian prefactor in calculating the power
- added an interface to directly pass a data pointer (rather than field string) to `CalcSpec`
- added a more performance way for FFTs on GPU using heffte `options`

This PR should go in after #1412 (as it already contains those changes).

@Yurlungur would be great if you could review (briefly again) so that we get the second approval.

doc/sphinx/src/fourier_transforms.rst
```diff
@@ -0,0 +1,297 @@
...
+
+Both classes use the :cpp:struct:`Box3D` struct to describe spatial extents:
+
+.. code-block:: cpp
+
+   struct Box3D {
+       int low[3];    // lower bound in each dimension
+       int high[3];   // upper bound in each dimension
+       int size[3];   // size in each dimension: high - low + 1
+   };
```
Yurlungur, line 39: Can we re-use the `RegionSize` struct?

example/CMakeLists.txt
```diff
@@ -24,3 +24,6 @@ add_subdirectory(poisson)
 add_subdirectory(poisson_gmg)
 add_subdirectory(diffusion)
 add_subdirectory(sparse_advection)
+if(PARTHENON_ENABLE_FFT)
+  add_subdirectory(fourier_transform)
+endif()
```
Yurlungur: missing newline

src/outputs/outputs.hpp
```diff
@@ -115,6 +115,17 @@ using HstVec_list = std::vector<HistoryOutputVec>;
...
+//----------------------------------------------------------------------------------------
+//! \class SpectralOutput
+//  \brief derived OutputType class for Spectrum dumps
+
+class SpectralOutput : public OutputType {
+ public:
+  explicit SpectralOutput(const OutputParameters &oparams) : OutputType(oparams) {}
+  void WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm,
+                       const SignalHandler::OutputSignal signal) override;
+};
```
Yurlungur, line 127: should this also be guarded behind the enable FFT macro?

src/outputs/spectrum.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include "utils/calc_spectrum.hpp"
+#include "utils/error_checking.hpp"
+
+namespace parthenon {
+
+//----------------------------------------------------------------------------------------
+//! \fn void SpectralOutput::WriteOutputFile()
+//  \brief Writes a spectrum output file
+
+void SpectralOutput::WriteOutputFile(Mesh *pm, ParameterInput *pin, SimTime *tm,
```
Yurlungur, line 26: see above question about macro

src/outputs/spectrum.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+    std::ofstream fout(fname);
+    if (!fout.is_open()) {
+      PARTHENON_FAIL("Could not open " + fname + " for writing");
+    }
+
+    fout << "# Bin    val_sum    K_sum    Count\n";
+    for (int i = 0; i < static_cast<int>(num_bins); ++i) {
+      fout << std::format("{:d} {:.15e} {:.15e} {:.15e}\n", i, spectra_h(i, 0),
+                          spectra_h(i, 1), spectra_h(i, 2));
```
Yurlungur, line 60: :+1: nice

Yurlungur (approved):
LGTM. Minor nitpicks remaining but they're nonblocking

src/outputs/outputs.hpp
pgrete: I think this can stay (so that there's no accidental overwrite by another `SpectralOutput` type)

src/outputs/spectrum.cpp
pgrete: This should be safe as the entire file is only selectively included through the macro in `CMakeLists.txt`

doc/sphinx/src/fourier_transforms.rst
pgrete: Regionsize currently only holds effectively the `size` aspect (plus position related `Real` info), but we need the lower and upper index bounds here.

pgrete:
Current CI image doesn't ship rocfft (required by heffte).
I fixed this in #1414, which now needs to go in before we can merge this PR.

pgrete:
@lkasselm the fail in the CUDA test seems legit:
```
[21](https://github.com/parthenon-hpc-lab/parthenon/actions/runs/29504003750/job/87639925037?pr=1403#step:9:222)
Initialized UniformGridHelper
Mesh layout:
Rank 0 local mesh box: low = (0, 0, 0), high = (15, 15, 15), size = (16, 16, 16)
*** The MPI_Comm_rank() function was called before MPI_INIT was invoked.
*** This is disallowed by the MPI standard.
*** Your MPI job will now abort.
[67e8270543fe:30505] Local abort before MPI_INIT completed completed successfully, but am not able to aggregate error messages, and not able to guarantee that all other processes were killed!
```
see https://github.com/parthenon-hpc-lab/parthenon/actions/runs/29504003750/job/87639925037?pr=1403

Do you want to take a look at this?
We might need to check that MPI (communicators) are properly initialized (in proper order) and, similarly, also closed cleanly on termination.

pgrete:
> @lkasselm the fail in the CUDA test seems legit:
> 
> ```
> [21](https://github.com/parthenon-hpc-lab/parthenon/actions/runs/29504003750/job/87639925037?pr=1403#step:9:222)
> Initialized UniformGridHelper
> Mesh layout:
> Rank 0 local mesh box: low = (0, 0, 0), high = (15, 15, 15), size = (16, 16, 16)
> *** The MPI_Comm_rank() function was called before MPI_INIT was invoked.
> *** This is disallowed by the MPI standard.
> *** Your MPI job will now abort.
> [67e8270543fe:30505] Local abort before MPI_INIT completed completed successfully, but am not able to aggregate error messages, and not able to guarantee that all other processes were killed!
> ```
> 
> see https://github.com/parthenon-hpc-lab/parthenon/actions/runs/29504003750/job/87639925037?pr=1403
> 
> Do you want to take a look at this? We might need to check that MPI (communicators) are properly initialized (in proper order) and, similarly, also closed cleanly on termination.

Argh, it's the serial test that fails, so no MPI...
I'll fix this.

pgrete:
Wait what?!
How did this merge go through with one test failing?!

PR #1405, opened 2026-06-10
[small] Fix race condition in flux corr comm with sparse=off

### LINKED ISSUE #1364: Bitwise mismatch after single cycle
# What's going wrong
Follow up on the issue in #1359 I'm seeing sth more subtle in AthenaPK.
As discussed, here's the reproducer @Yurlungur @lroberts36 

For a pure hydro problem with 3D periodic bounds (so the code path does not include any reduction with sums etc on device), some values in cells (always touching block boundaries) do not agree after a single cycle when run on different number of ranks.

I can reliable (i.e., every single time) reproduce the issue on [JUWELS](https://apps.fz-juelich.de/jsc/hps/juwels/booster-overview.html) Booster on a single node (4x way A100 40G with NVLink3).

# What I tried
Things I've already tried (and probably more), which do *not* (reliably) help
- Changing software/library stack between
  - CUDA/13.0, GCC/14.3.0, OpenMPI/5.0.8
  - CUDA/12.6, GCC/13.3.0, OpenMPI/5.0.5
  - CUDA/12.2, GCC/12.3.0, OpenMPI/4.1.5
- Disabling `gdr_copy` UCX transport, see also related (and fixed) https://github.com/parthenon-hpc-lab/parthenon/issues/1260
- Different time integrators (changing to RK1 makes the issue show up in the second cycle)
- Running with multiple ranks on a single GPU versus multiple ranks on multiple GPU (on the same node) -- though it seems to happen less when running with one rank per GPU
- `parthenon/mesh/pack_size=1`

Things that seem to help
- `parthenon/mesh/do_coalesced_comms=true`
- Enabling `export CUDA_LAUNCH_BLOCKING=1`
- compile *with* sparse support `-DPARTHENON_DISABLE_SPARSE=OFF` (sparse is disabled by default in AthenapK)

So if this is not a lib issue, then my best guess at the moment is some very subtle race condition.

# Steps to reproduce

1. Get current AthenaPK `main` branch
2. Update Parthenon submodule to `pgrete/fix-buffer-init` branch to include #1360 
3. Configure/build:
```
cmake -Bbuild-a100-main-fixbuf -DKokkos_ENABLE_CUDA=ON -DKokkos_ARCH_AMPERE80=ON -DKokkos_ARCH_ZEN2=ON  -DPARTHENON_DISABLE_HDF5_COMPRESSION=ON -DPARTHENON_ENABLE_PYTHON_MODULE_CHECK=OFF -GNinja
```
4. Run baseline (on single rank)
```
srun -n1 /path/to/build-a100-main-fixbuf/bin/athenaPK -i cloud.in
```
with `cloud.in` being
```
<comment>
problem = Cloud in wind

<job>
problem_id = cloud

<parthenon/mesh>
refinement = adaptive
numlevel = 3
nghost = 2

nx1 = 64 
x1min = -4000
x1max = 4000
ix1_bc = periodic
ox1_bc = periodic

nx2 = 128
x2min = -3000
x2max = 13000
#ix2_bc = cloud_inflow_x2
ix2_bc = periodic
ox2_bc = periodic

nx3 = 64 
x3min = -4000
x3max = 4000
ix3_bc = periodic
ox3_bc = periodic

pack_size = -1      # number of MeshBlock per MeshBlockPack

<parthenon/meshblock>
nx1=32
nx2=32
nx3=32

<parthenon/time>
integrator = rk2
cfl = 0.20
tlim = 20.0
nlim = 2
ncycle_out_mesh = -500  # print mesh structure every 500 cyles and on refinement

<hydro>
fluid = euler
eos = adiabatic
riemann = hllc
reconstruction = plm
gamma = 1.666666666666667 # gamma = C_p/C_v
nscalars = 1

#dfloor = -1.0                            # unused, in [code units]
#pfloor = -1.0                            # unused, in [code units]
Tfloor = 8000                             # in K
He_mass_fraction = 0.24

<units>
#Units parameters
code_length_cgs = 3.085677580962325e+18  # 1 pc
code_mass_cgs = 1.98841586e+33           # 1 M_sun
code_time_cgs = 3.15576e+13              # 1 Myr


<refinement>
type                         = maxdensity
maxdensity_refine_above      = 1e-4        # in code units
maxdensity_deref_below       = 5e-5        # in code units

<problem/cloud>
r0_cgs = 1.68169e21                     # 545 pc
rho_cloud_cgs = 1e-26                   # g/cm^3
rho_wind_cgs = 3.33333e-29              # g/cm^3
T_wind_cgs  = 3e6                       # K
v_wind_cgs = 1.7e8                      # 1700 km/s
rescale_code_time_to_tcc = false        # if set, all dt and time above will be rescaled in units of t_cc

<parthenon/output2>
file_type = rst
#dt = 0.5
dn =  1
id = restart
hdf5_compression_level = 0
```
6. Store results in somewhere
7. Run again with more ranks (4 seems to reliably trigger the issue)
```
srun -n4 /path/to/build-a100-main-fixbuf/bin/athenaPK -i cloud.in
```
8. Compare results (at cycle 1)
```
python /path/to/athenapk/external/parthenon/scripts/python/packages/parthenon_tools/parthenon_tools/phdf_diff.py  -i -o -b -t 1e-20 parthenon.restart.00001.rhdf ref/parthenon.restart.00001.rhdf 
...
Ignoring metadata
Tolerance = 1e-20
Diff in cons                
    bkji: (  42,   4,  15,   0)
    zyx: (-1718.750000,1968.750000,31.250000)
    err_mag: 3.764949e-15
    f0: 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01
    f1: 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01 7.7689e-01
    err: 0.0000e+00 0.0000e+00 0.0000e+00 0.0000e+00 0.0000e+00 0.0000e+00 3.2196e-15 1.7764e-15 7.7716e-16 2.2204e-16 0.0000e+00 0.0000e+00 0.0000e+00 0.0000e+00 0.0000e+00 0.0000e+00 0.0000e+00 0.0000e+00 0.0000e+00 0.0000e+00 0.0000e+00 0.0000e+00 0.0000e+00 0.0000e+00 0.0000e+00 0.0000e+00 0.0000e+00 0.0000e+00 0.0000e+00 0.0000e+00 0.0000e+00 0.0000e+00
  cons                : differs

```

Note that the initial conditions written to disk (output 00000) agree.

### LINKED ISSUE #1400: Multi-rank MG solver diverges on 3D AMR Poisson problem (Dirichlet BCs)
**Summary:**

Multigrid solver diverges on multi-rank runs for a 3D AMR Poisson problem with Dirichlet ("zero") boundary conditions. The same binary and
input converge perfectly on 1 rank. The issue manifests in both pure MG and BiCGSTAB+MG-precond configurations.

**Environment:**
- Parthenon: develop @ 287313a77
- Kokkos: 5.1.0
- Compiler: GCC 13.3.0, CUDA 12.5.82
- Build: H100 GPUs
- Application: AthenaPK with self-gravity port (Poisson source defined via 4πG(ρ - ρ_mean), Dirichlet outer BCs)

**Reproducer:**

Production input: 256^3 root grid, meshblock 32^3 (512 base blocks), AMR with Jeans criterion to numlevel=12, Dirichlet gravity BCs, outflow
hydro BCs, M=0.5 turbulent initial conditions, collapse_be problem generator.

Solver parameters used:
- solver = MG
- max_iterations = 200
- residual_tolerance = 1.0e-10
- do_FAS = true     (default)
- max_coarsenings = INT_MAX  (default)
- smoother = SRJ2   (default)

**Issue 1: Default MG parameters diverge on multi-rank**

| Rank count | V-cycle 1 | V-cycle 2 | V-cycle 13 | V-cycle 200 |
|------------|-----------|-----------|------------|-------------|
| 1          | 3.20e-01  | 2.99e-02  | 8.51e-11 (converged) | --      |
| 4          | 3.04e+00  | 6.81e+00  | 4.08e+01   | 2.87e+111  |

Same binary, same input, only --ntasks differs. 4-rank residual grows monotonically by factor ~3× per V-cycle.

**Issue 1 -- Workaround found:**

Setting these solver parameters makes MG converge on cycle 0 on 4-rank:
- do_FAS = false
- max_coarsenings = 3
- smoother = SRJ1
- presmoother = SRJ1
- postsmoother = SRJ1

With these, cycles 0-32 run cleanly on 4-rank, KE/momentum/energy grow physically, and the first AMR refinement event (512 -> 533 blocks) is survived without issue.

**Issue 2: AMR remap on multi-rank fails for large refinement bursts**

Even with the working solver config above, the simulation explodes at the second AMR refinement event:

- cycle 32,  nbtotal=533,  KE=1350 (clean, first AMR survived)
- cycle 33,  nbtotal=673, KE=3.4e46  (explodes at AMR event creating +140 new fine-level blocks across ranks simultaneously)

The new blocks appear to start with bad cross-rank ghost data for the gravity field, producing infinite acceleration and CFL collapse.

**What seems to be the case:**

1. The bug is multi-rank specific. 1-rank converges cleanly on the same problem with default parameters.
2. The smoother divergence (Issue 1) and the AMR remap failure (Issue 2) appear to be distinct bugs in the multi-rank MG path -- one in the V-cycle itself with default params, one in the post-remap ghost initialization.

Reproducer data and full logs available — happy to share.

pgrete:
## PR Summary

Codex helped (Claude tried harder but didn't find the right spot) in just 3 min...

Prompt:
```
❯ The current folder contains the finite-volume, compressible                   
  magnetohydrodynamics code AthenaPK based on the performance portable,         
  adaptive mesh refinement framework parthenon (located in the                  
  `external/parthenon` folder). I am trying to find the root cause of the       
  following issue https://github.com/parthenon-hpc-lab/parthenon/issues/1364    
  Other Parthenon based code have not been observing this issue. The key        
  differences between AthenaPK and other codes are the use of a single large    
  state vector in AthenaPK (versus multiple separate fields for each variable   
  in other codes) and that, by default, the machinery for sparse variables is   
  disabling in AthenaPK (i.e., configured with `-DPARTHENON_DISABLE_SPARSE=ON). 
  Please have a close look at the AthenaPK and Parthenon codebases with a       
  specific emphasis on potential race conditions or index differences (as they  
  were done in the past for sparse signaling) around ghost zone communication.
```
Answer:
```
• I found a likely root cause and patched it in external/parthenon/src/bvals/comms/boundary_communication.cpp:154.

  The failing path is non-coalesced BoundaryType::flxcor_send with sparse disabled. The send buffer is filled by an async
  Kokkos kernel, then MPI can immediately send from that device buffer without a fence. Normal ghost sends had a fence for
  any/nonlocal; flux-correction sends did not. This matches the issue symptoms: boundary-touching diffs, fixed by
  CUDA/HIP_LAUNCH_BLOCKING, fixed by coalesced comms, and fixed by compiling sparse support because the sparse nonzero flag
  deep_copy accidentally synchronizes first. Issue context: https://github.com/parthenon-hpc-lab/parthenon/issues/1364

  Patch made:

  const bool coal_comm = pmesh->do_coalesced_comms;
  #ifdef MPI_PARALLEL
    if (!coal_comm && IsSender(bound_type)) Kokkos::fence();
  #endif

  This applies the pre-MPI fence to all non-coalesced sender boundary types, including flxcor_send, while leaving coalesced
  comms to use their existing fence in PackAndSend.
```

I guess the reproducer and mitigation strategies in the issue description contained enough info to quickly isolate the issue.

Fixes #1364
Fixes #1400

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build


Yurlungur (approved):
Is there some way we can add a test for this? I know it's a race condition, so that may be difficult.

pgrete:
> Is there some way we can add a test for this? I know it's a race condition, so that may be difficult.

Testing for this specifically is probably hard, but  in general I think it'd be worth to have some "bitwise identical" test (not run on every single commit) that ensure we get the same results for
- various key options ON/OFF
  -  incl MPI and sparse
  - single rank vs multiple rank
for
  - uniform (with different block sizes)
  - smr/amr

and probably more.

src/bvals/comms/boundary_communication.cpp
```diff
@@ -151,14 +153,13 @@ TaskStatus SendBoundBufsWithRestrictOption(std::shared_ptr<MeshData<Real>> &md,
   // Send buffers
   if (Globals::sparse_config.enabled)
     Kokkos::deep_copy(sending_nonzero_flags_h, sending_nonzero_flags);
+  const bool coal_comm = pmesh->do_coalesced_comms;
 #ifdef MPI_PARALLEL
-  if (bound_type == BoundaryType::any || bound_type == BoundaryType::nonlocal)
-    Kokkos::fence();
+  if (!coal_comm && IsSender(bound_type)) Kokkos::fence();
```
acreyes: This also covers now the `gmg_*_send`s, which is what fixes #1400 ?

acreyes (approved):
Super subtle! I don't think I would have noticed `BoundaryType::any`. Does that refer to `local`/`nonlocal` rather than any `BoundaryType` in usage?

src/bvals/comms/boundary_communication.cpp
lroberts36: I think `IsSender(bound_type)` should always be `true` here. This check was originally to prevent a fence for `BoundaryType::local` (which involves only same rank boundaries and therefore no MPI) and was added before fluxes were unified here. Probably should just check `if(!coal_comm) Kokoks::fence();`

lroberts36 (approved):

src/bvals/comms/boundary_communication.cpp
pgrete: I think so.

pgrete: Yes, good point.

PR #1406, opened 2026-06-16
Fix deadlock in parallel outputs after ParameterInput refactor

pdmullen:
This PR was made with the assistance of generative AI.

## PR Summary

MPI deadlock during parallel HDF5 output after Parthenon commit `fc16e67` (`ParameterInput` refactor). Different ranks could have the same parameter in different resolution states (`UnresolvedString` vs typed), causing `ParameterDump()` to produce different strings on different ranks, which deadlocked HDF5 collective metadata writes.    

Changes:                                                                                                                                                       
  1. `parameter_input.hpp`: Added `original_string` field to `Parameter`
  2. `parameter_input.cpp`:                                                                                                                                        
    - Preserve original string in `GetFromStorage_()` before type conversion                                                                                     
    - Use original string in `ParameterDump()` when available         

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 - [[PR 1351]](https://github.com/parthenon-hpc-lab/parthenon/pull/1351) Bump Kokkos 5 & C++20
 
 ### Fixed (not changing behavior/API/variables/...)
+- [[PR 1406]](https://github.com/parthenon-hpc-lab/parthenon/pull/1406) Fix deadlock in parallel outputs after ParameterInput refactor
 - [[PR 1405]](https://github.com/parthenon-hpc-lab/parthenon/pull/1405) Fixes race condition in flux correction communication when sparse variables are disabled
 - [[PR 1339]](https://github.com/parthenon-hpc-lab/parthenon/pull/1339) Fixes doc for `numlevel` usage in non-AMR sims
 - [[PR 1393]](https://github.com/parthenon-hpc-lab/parthenon/pull/1393) Fixes a bug where CMake does not detect HDF5 parallel IO for HDF5 2.0.0 and later.
--- a/src/parameter_input.cpp
+++ b/src/parameter_input.cpp
@@ -680,7 +680,9 @@ void ParameterInput::ParameterDump(std::ostream &os) {
     std::size_t max_len_name = 0;
     std::size_t max_len_value = 0;
     for (const auto &param : block.params) {
-      std::string value_str = ParamValueToString(param.value);
+      std::string value_str = param.original_string.has_value()
+                                  ? param.original_string.value().value
+                                  : ParamValueToString(param.value);
       max_len_name = std::max(max_len_name, param.name.length());
       max_len_value = std::max(max_len_value, value_str.length());
     }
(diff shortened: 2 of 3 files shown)


lroberts36 (approved):
Good catch! I wonder a little bit about what is going on generally though. If one rank calls `GetOrAdd` and another doesn't, don't we end up in a situation where if the parameter isn't in the input file we get different parameter input on different ranks? That seems like a problem beyond the `ParameterInput` refactor.

Yurlungur (approved):

Yurlungur:
> Good catch! I wonder a little bit about what is going on generally though. If one rank calls `GetOrAdd` and another doesn't, don't we end up in a situation where if the parameter isn't in the input file we get different parameter input on different ranks? That seems like a problem beyond the `ParameterInput` refactor.

I thought I had added logic to check for this. When ParameterInput is resolved, there's a hash that is computed accross ranks. Why did that not trigger here?

lroberts36:
> > Good catch! I wonder a little bit about what is going on generally though. If one rank calls `GetOrAdd` and another doesn't, don't we end up in a situation where if the parameter isn't in the input file we get different parameter input on different ranks? That seems like a problem beyond the `ParameterInput` refactor.
> 
> I thought I had added logic to check for this. When ParameterInput is resolved, there's a hash that is computed accross ranks. Why did that not trigger here?

Yes, I forgot about that. That will catch the issue I mentioned, but it doesn't hash the parameter values so it won't catch the issue at the level Patrick ran into it.

Yurlungur:
> > > Good catch! I wonder a little bit about what is going on generally though. If one rank calls `GetOrAdd` and another doesn't, don't we end up in a situation where if the parameter isn't in the input file we get different parameter input on different ranks? That seems like a problem beyond the `ParameterInput` refactor.
> > 
> > 
> > I thought I had added logic to check for this. When ParameterInput is resolved, there's a hash that is computed accross ranks. Why did that not trigger here?
> 
> Yes, I forgot about that. That will catch the issue I mentioned, but it doesn't hash the parameter values so it won't catch the issue at the level Patrick ran into it.

AAAH. Maybe we should also hash parameters... That would be a mostly straightforward generalization for scalar params... I guess it doesn't get everything though.

lroberts36:
> > > > Good catch! I wonder a little bit about what is going on generally though. If one rank calls `GetOrAdd` and another doesn't, don't we end up in a situation where if the parameter isn't in the input file we get different parameter input on different ranks? That seems like a problem beyond the `ParameterInput` refactor.
> > > 
> > > 
> > > I thought I had added logic to check for this. When ParameterInput is resolved, there's a hash that is computed accross ranks. Why did that not trigger here?
> > 
> > 
> > Yes, I forgot about that. That will catch the issue I mentioned, but it doesn't hash the parameter values so it won't catch the issue at the level Patrick ran into it.
> 
> AAAH. Maybe we should also hash parameters... That would be a mostly straightforward generalization for scalar params... I guess it doesn't get everything though.

At the end of the day, we just write the parameters as a big string to the hdf5 file. We could just broadcast and compare that string (although maybe it is getting bigger as parameter input gets more complicated). We could hash the string, but that is obviously not a perfect check.

pgrete (approved):

pgrete:
I wonder if there's a more general solution to ensure consistency. Let's get this in and discuss during the sync.

Issue #1409, opened 2026-07-09
Single precision todo/wish list


pgrete:
While we support single precision in principle, the current state fragile/untested:
To we we should (at least -- potentially more):
- [ ] Make it compile
- [ ] Fix Swarm comm (cast to `Real`, but should probably be bitmap)
- [ ] Add tests
- [ ] Ensure collectives are at higher prec (both on device and MPI)
- [ ] Identify pieces that should be at high prec (like time and location related variables)
- [ ] Allow restart from/to other precision

PR #1412, opened 2026-07-10
[trivial] Fix single precision compilation with Real=float

pgrete:
## PR Summary

Does what is says one the tin.
Focuses on compilation only.
Additional todos are listed here #1409 

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build

### DIFF
--- a/example/kokkos_pi/kokkos_pi.cpp
+++ b/example/kokkos_pi/kokkos_pi.cpp
@@ -244,8 +244,9 @@ result_t naiveKokkos(int n_block, int n_mesh, int n_iter, double radius) {
   // formulate result struct
   constexpr int niops = 8;
   constexpr int nfops = 11;
-  auto r = result_t{"Naive_Kokkos", (6.0 * sumArray(blocks, n_block) * dVol / radius3),
-                    time_basic, niops, nfops};
+  auto r = result_t{"Naive_Kokkos",
+                    (static_cast<Real>(6.0 * sumArray(blocks, n_block) * dVol / radius3)),
+                    static_cast<Real>(time_basic), niops, nfops};
 
   return r;
 }
--- a/src/basic_types.hpp
+++ b/src/basic_types.hpp
@@ -39,6 +39,8 @@ using Real = double;
 #endif
 #endif
 
+using LocReal = double; // precision for location related variables
+
 struct IndexRange {
   int s = 0; /// Starting Index (inclusive)
   int e = 0; /// Ending Index (inclusive)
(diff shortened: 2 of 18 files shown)


src/outputs/parthenon_opmd.cpp
```diff
@@ -805,7 +805,7 @@ void OpenPMDOutput::WriteOutputFileImpl(Mesh *pm, ParameterInput *pin, SimTime *
             if (pm->ndim == 3) {
               auto grid_spacing = std::vector<Real>{dx3, dx2, dx1};
               auto axis_labels = std::vector<std::string>{"z", "y", "x"};
-              auto global_offset = std::vector<Real>{
+              auto global_offset = std::vector<LocReal>{
```
Yurlungur, line 808: what is LocReal?

src/basic_types.hpp
```diff
@@ -39,6 +39,8 @@ using Real = double;
 #endif
 #endif
 
+using LocReal = double; // precision for location related variables
```
Yurlungur, line 42: ah this is `LocReal`. Will we ever change this precision? Should we not just use double where this is relevant? Or is naming it something like this useful so we know it's intentional?

Yurlungur (approved):
tests not passing. Otherwise LGTM.

src/outputs/parthenon_opmd.cpp
pgrete: I think it'd be useful to abstract location (and time) precision so that it's decoupled from `double` to make it (more) obvious that all position and time related quantities are always at a fixed precision, which will be useful for changing precision at runtime and, in the long run, for very deep hierarchies and large dynamical range (thinking about cosmology here).

One thing that just came to my mind is whether we want to fix `Real` within `pinput` internally to double (though transparently to the user) very much similar to the string handling we (original vs processed).

pgrete:
FYI the `/` filesystem is full on the CI machine, which is why the tests fail

Yurlungur:
> FYI the `/` filesystem is full on the CI machine, which is why the tests fail

lol :+1: ok let me know when I should re-attempt

pgrete:
> > FYI the `/` filesystem is full on the CI machine, which is why the tests fail
> 
> lol 👍 ok let me know when I should re-attempt

ready for re-review (and a second one). All pipelines pass

example/kokkos_pi/kokkos_pi.cpp
Yurlungur, line 1: we may want to remove this example at some point

Yurlungur (approved):

acreyes (approved):

PR #1413, opened 2026-07-15
submeshdata part 1

Yurlungur:
## PR Summary

In riot, we often have a pattern where we make new mesh data objects to hold, e.g., right-hand sides for source terms, so that order in the task list doesn't matter and source terms can be, in principle, computed concurrently. Right now, we pull all these out by hand in a step init function in the driver. This MR provides a general framework for automatically registering this kind of mesh data object in packages. Then, at least for the evolution driver, these meshdata objects are automatically constructed.

For riot this should reduce a lot of boiler plate and it's also more "parthenonic" support for the kind of thing I know @c-prather and @acreyes have extended the `StateDescriptor` type to do down stream.

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build

@wip ready for review


src/interface/mesh_data_descriptor.hpp
```diff
@@ -0,0 +1,45 @@
...
+#include "utils/unique_id.hpp"
+
+namespace parthenon {
+class Mesh;
+template <typename T>
+class MeshData;
+
+class SubMeshDataRequirements {
+ public:
+  std::vector<std::string> varnames;
```
acreyes, line 34: It might be nice to have a helper to set this from variable types, something like
```c++
submesh_requirements.varnames = SubMeshDataRequirements::VariableList<Density, Temperature>();
submesh_requirements.varnames = SubMeshDataRequirements::VariableList(TypeList<Density, Temperature>());
```

tst/catch2_define.cpp
```diff
@@ -75,12 +79,15 @@ int main(int argc, char *argv[]) {
...
-      std::cerr << "### FATAL ERROR in ParthenonInit" << std::endl
-                << "MPI Initialization failed." << std::endl;
-      return -1;
+    if (!already_initialized) {
+      int status = MPI_Init(&argc, &argv);
+      if (status != MPI_SUCCESS) {
+        std::cerr << "### FATAL ERROR in ParthenonInit" << std::endl
+                  << "MPI Initialization failed." << std::endl;
+        return -1;
+      }
```
acreyes, line 90: does this fix the mpi issues with the unit tests you discovered?

acreyes (approved):
This is super cool, I am definitely going to use it. I was doing something similar downstream to narrow the fields exchanged in boundary tasks. Tying it to the state descriptor is a great idea!

tst/catch2_define.cpp
Yurlungur: Nope, that's just because it was easier for me to read/inject print statement in this form. These lines are the MPI bug:

```C++
    const auto &info = test_case.getTestCaseInfo();
    const auto &tags = info.tags;
    // if the spec is empty, i.e., all tests are requested, we have to special case.
    const bool selected = has_filters ? test_spec.matches(test_case) : !info.isHidden();
    const bool is_mpi = std::find(tags.begin(), tags.end(), "MPI") != tags.end();
    if (selected && is_mpi) {
```

But once the other branch is merged, this diff should disappear.

src/interface/mesh_data_descriptor.hpp
Yurlungur: Good idea. I'll add something like that.

Yurlungur: Added non-static helper functions:

```C++
class SubMeshDataRequirements {
 public:
  std::vector<std::string> varnames;
  Metadata::FlagCollection flags;
  std::vector<int> sparse_ids;
  bool shallow = false;

  template <typename... Ts>
  void RegisterVariables() {
    std::vector<std::string> new_names = {Ts::name()...};
    RegisterVariables(new_names);
  }
  template <typename... Ts>
  void RegisterVariables(const TypeList<Ts...> &tl) {
    RegisterVariables<Ts...>();
  }
  void RegisterVariables(const std::vector<std::string> &new_names) {
    varnames.insert(varnames.end(), new_names.begin(), new_names.end());
  }
  void RegisterVariables(const std::string &new_name) { varnames.push_back(new_name); }
  template <typename... Args>
  void RegisterVariables(const std::string &new_name, Args... args) {
    varnames.push_back(new_name);
    RegisterVariables(std::forward<Args>(args)...);
  }

  std::vector<Uid_t> AddMDSubset(Mesh *pmesh, const std::string &name,
                                 const std::shared_ptr<MeshData<Real>> &base);
  const auto &GetUids() const { return uids_; }

 private:
  std::vector<Uid_t> uids_;
};
```

src/interface/mesh_data_descriptor.hpp
```diff
@@ -0,0 +1,66 @@
...
+  Metadata::FlagCollection flags;
+  std::vector<int> sparse_ids;
+  bool shallow = false;
+
+  template <typename... Ts>
+  void RegisterVariables() {
+    std::vector<std::string> new_names = {Ts::name()...};
```
acreyes, line 44: Should we guard on `Ts::regex()`?

Yurlungur: Yes. Done.

src/driver/driver.cpp
```diff
@@ -161,6 +161,18 @@ DriverStatus EvolutionDriver::Execute() {
         pmesh->PreStepUserDiagnosticsInLoop(pmesh, pinput, tm);
       }
 
+      { // Anonymous meshdata subsets from packages
+        // TODO(JMM): Currently this is always based on base. This is
+        // possibly not always desirable for shallow copies (which are
+        // nominally supported).
+        auto &base = pmesh->mesh_data.Get();
```
lroberts36, line 168: Did you want to add this for the base partition or the default partitions or both?

lroberts36 (approved):

src/driver/driver.cpp
Yurlungur: ah good call. I imagined it always being base, since the default partitions will get generated automatically when `GetOrAdd` is called later. Do you have thoughts on that? 

I could also make it more sophisticated with an optional target for the shallow copies...

Yurlungur: @lroberts36 thoughts?

Yurlungur:
@par-hermes format

Yurlungur:
So I chatted with @lroberts36 offline and I think he suggested a nice extension of this that is significantly more general. So I'm going to go after that. But I'd like to get this merged in now, and submit a revised MR that cleans this up.

src/driver/driver.cpp
Yurlungur: Will resolve in subsequent MR.

src/interface/state_descriptor.hpp
```diff
@@ -378,6 +392,25 @@ class StateDescriptor {
...
+  }
+
+  bool ContainsMeshDataSubset(const std::string &partial_name) {
+    return submeshdata_map_.count(partial_name) > 0;
+  }
+
+  std::vector<Uid_t> AddMeshDataSubset(Mesh *pmesh, const std::string &partial_name,
+                                       const std::shared_ptr<MeshData<Real>> &base) {
+    auto full_name = GetMeshDataSubsetFullname(partial_name);
+    return submeshdata_map_[partial_name].AddMDSubset(pmesh, full_name, base);
```
pgrete, line 408: Should we check that `partial_name` exists (as below where
```
  PARTHENON_REQUIRE(ContainsMeshDataSubset(partial_name),
                    "Package " + label() + " must contain a subset " + partial_name);
```

Yurlungur:
The failing test is the Fourier transform. @pgrete how should we resolve? Just force merge for now? Or should that test be disabled while you work on it?

src/interface/mesh_data_descriptor.cpp
```diff
@@ -0,0 +1,46 @@
...
+#include "mesh_data_descriptor.hpp"
+#include "utils/unique_id.hpp"
+#include "utils/utils.hpp"
+
+namespace parthenon {
+std::vector<Uid_t>
+MeshDataDescriptor::AddMDSubset(Mesh *pmesh, const std::string &name,
+                                const std::shared_ptr<MeshData<Real>> &base) {
+  std::vector<std::string> resolved_vars =
+      pmesh->GetVariableNames(varnames, flags, sparse_ids);
```
pgrete: Codex pointed out that there might be an issue here as `GetVariableNames` could return an empty vector, which, when passed to `mesh_data.Add` it is passed also empty to `Initialize` where it triggers a special case `    // special case when the list of vars is empty, copy everything`, which is probably not what we want given that there might be a typo in the `varnames` input here.

pgrete:
> The failing test is the Fourier transform. @pgrete how should we resolve? Just force merge for now? Or should that test be disabled while you work on it?

I spent almost three hours yesterday trying to track this down.
I can reproduce on the CI machine pointing towards some internal unpack function in `heFFTe` (so no Parthenon indexing issue etc).
Trying to disable Kokkos::complex align or changing the ROCM version (to the latest 7.14) release fixed git.
I just confirmed that this works on Frontier with ROCM 7.0.2 as expected (both single and multi-rank):
```
pgrete@frontier08768 /ccs/proj/ast146/pgrete/src/parthenon/build-jun26-hip-dev-d9d4a-fft-own $ ctest -R fourier
Test project /ccs/proj/ast146/pgrete/src/parthenon/build-jun26-hip-dev-d9d4a-fft-own
    Start 10: regression_test:fourier_transform
1/2 Test #10: regression_test:fourier_transform .......   Passed    4.18 sec
    Start 11: regression_mpi_test:fourier_transform
2/2 Test #11: regression_mpi_test:fourier_transform ...   Passed    9.91 sec

100% tests passed, 0 tests failed out of 2

Label Time Summary:
cuda          =   9.91 sec*proc (1 test)
mpi           =   9.91 sec*proc (1 test)
regression    =  14.08 sec*proc (2 tests)
serial        =   4.18 sec*proc (1 test)

Total Test time (real) =  14.17 sec

```
So I assume that there's an issue with the consumer/workstation GPU in the CI machine, for which heffte was probably never properly tested.
My suggested fix/workaround for now is to disable the test here `https://github.com/parthenon-hpc-lab/parthenon/blob/d9d4af66903fae2ad6e68b3a4dfaa9e2f53ce04a/.github/workflows/ci-extended.yml#L160` with a comment to this GitHub comment that the test should be reenabled when we run CI on a different AMD machine.

src/interface/mesh_data_descriptor.cpp
Yurlungur: Hmm. I do think we want to support that possibility. But I agree that might not entirely be intentional. I will add a special flag to mesh data descriptor that lets you specify "actually I want everything"

pgrete:
It's not clear why the jobs failed. I restarted the failed ones.

Yurlungur:
The WIP check is hanging for some reason, despite the MR not being WIP.

Yurlungur:
@wip ready for review

Yurlungur:
Ok I don't know why WIP is just hanging but I'm force merging.

PR #1416, opened 2026-07-18
[Trivial] remove virtual from packs

Yurlungur:
## PR Summary

I just noticed that SparsePackBase and SwarmPackBase both have a virtual destructor, which is unused. This adds a vtable to each class that depends on the base class. Is there a reason for this, @pdmullen @lroberts36 ? I think no right? 

## PR Checklist

- [ ] Code passes cpplint
- [ ] Code is formatted
- [ ] Docs build

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 
 
 ### Changed (changing behavior/API/variables/...)
+- [[PR 1416][(https://github.com/parthenon-hpc-lab/parthenon/pull/1416) Remove virtual tag from destructors in sparse and swarm pack base classes
 - [[PR 1401]](https://github.com/parthenon-hpc-lab/parthenon/pull/1401) Sparse Field Component Names
 - [[PR 1331]](https://github.com/parthenon-hpc-lab/parthenon/pull/1331) Use `size_t` for sizes
 - [[PR 1355]](https://github.com/parthenon-hpc-lab/parthenon/pull/1355) Allow disabling format and lint targets
--- a/src/pack/sparse_pack/sparse_pack_base.hpp
+++ b/src/pack/sparse_pack/sparse_pack_base.hpp
@@ -29,7 +29,7 @@ namespace parthenon {
 class SparsePackBase {
  public:
   SparsePackBase() = default;
-  virtual ~SparsePackBase() = default;
+  ~SparsePackBase() = default;
 
  protected:
   friend class SparsePackCache;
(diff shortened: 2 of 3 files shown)


pgrete (approved):
Anyone knows/remembers what the intention was to have them virtual in first place?

lroberts36 (commented):
I think this is not necessarily what we want to do. Every templated `SparsePack` inherits from `SparsePackBase`. I would have to look through the code again to see if we rely on polymorphism anywhere, but generally it is not a good idea to have a base class with a non-virtual destructor. I would guess removing virtual here probably works currently, but not sure if this is what we really want to do. You could move to including `SparsePackBase` by composition if you are worried about the cost of the virtual destructor or some specific behavior this causes on device.

Yurlungur:
> I think this is not necessarily what we want to do. Every templated `SparsePack` inherits from `SparsePackBase`. I would have to look through the code again to see if we rely on polymorphism anywhere, but generally it is not a good idea to have a base class with a non-virtual destructor. I would guess removing virtual here probably works currently, but not sure if this is what we really want to do. You could move to including `SparsePackBase` by composition if you are worried about the cost of the virtual destructor or some specific behavior this causes on device.

I hit this issue in riot, basically the virtual destructor prevents one from making kokkos views of sparse packs, at least in the naive way. When the view destructor gets called, it calls the destructor for the sparse pack, which is virtual and thus requires a vtable lookup, which fails.

The fix would be to initialize the views via placement new or something similar (see my MR in riot), but IMO this adds a lot of complexity for not much gain and is a pretty big footgun. Since `SparsePack`s are trivially copyable, they should also be trivially destructible, so I don't think the typical runtime polymorphism wisdom applies. If we never *delete* sparse packs through pointers to `SparsePackBase`, I don't think we need virtual, and my instinct would be we should avoid it, but I definitely could be missing something. I don't completely understand, e.g., the caching mechanism you use for sparse packs and pack descriptors.

Yurlungur:
@lroberts36 do you have further thoughts here?

lroberts36:
> @lroberts36 do you have further thoughts here?

I don't think sparse packs are trivially destructible since they contain managed `Kokkos::view`s. Making a view of `SparsePack`s or `SparsePackBase`s runs into all the same view of view problems as usual. But these are surmountable, while the vtable is obviously not I guess?

I think removing the virtual destructor is safe for the current version of the code because we do not use runtime polymorphism anywhere currently (including in the cacheing). There is no reason in principle not to go to composition here though, which would be safer but require some changes.

Yurlungur:
So how should we proceed in that case? We could either merge this or hold off for a larger rewrite?

lroberts36 (approved):

PR #1431, opened 2026-08-11
Add Loop Abstraction Reductions

lroberts36:
## PR Summary

[Partially made with generative AI]

Adds outer_reduce/inner_reduce to src/loop_abstraction, mirroring outer/inner but folding a single Kokkos reducer over an IndexSpace. Supported for all three loop tags (bvoi/bovi/boiv). Reductions always run on the Kokkos backend (on host-only builds DevExecSpace is a host space, so the Kokkos reduce still runs); there is no raw path currently.

### API

The reducer is carried on the index-space type, so the outer body has the same shape as loop abstraction for loops:

```c++
  using rist = la::ReductionIndexSpace<lt, it, Kokkos::Sum<Real>>;
  rist idx_space(/* ... */);

  auto result = la::outer_reduce(idx_space,
    KOKKOS_LAMBDA(const rist::idx_range_t &r, int b) {
      la::inner_reduce(r, [&](auto idx, auto &v) { v += /* value at idx */; });
    });
```

  - Build a reduction space with ReductionIndexSpace<lt, it, R> (hides the backend param) or idx_space.WithReducer<R>().
  - A region may interleave plain inner (e.g. scratch fill) with multiple inner_reduce calls that all join into one accumulator.

### Guarantees / guard rails

  - Reductions never touch ghost cells: halo-extended ranges are rejected at compile time, and the memory inner tag degenerates to logical_flat inside inner_reduce (plain inner is unaffected).
  - outer rejects reduction spaces and outer_reduce requires them (both static_assert).
  - Per-range reduction state is a [[no_unique_address]] member that is empty (zero bytes) for non-reduction spaces — ordinary ranges are byte-identical to before (checked with static_asserts).

### Incidental changes

  - Pack/flux/var views were decoupled from the IndexSpace type (templated on <loop_tag, inner_tag>, holding a const Indexer3D*) so the new reducer parameter costs them nothing.
  - New explicit-logical-ranges IndexSpace constructor for non-standard logical boxes.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [ ] Docs build


lroberts36:
@par-hermes format

lroberts36:
@par-hermes format

lroberts36:
@par-hermes format

lroberts36:
@par-hermes format

pdmullen (approved):

doc/sphinx/src/loop_abstraction.rst
```diff
@@ -179,6 +179,64 @@ scratch for the Kokkos backend for other paths), but the user-facing indexing is
...
+     KOKKOS_LAMBDA(const rist::idx_range_t &idx_range, int b) {
+       la::inner_reduce(idx_range, [&](auto idx, auto &v) {
+         v += /* something at idx */;
+       });
+     });
+
+An escape-hatch overload instead takes a caller-constructed reducer instance *last*
+(matching ``Kokkos::parallel_reduce(policy, functor, reducer)``) for reducing into a
+``View``, ``ScatterView``, or device memory; it returns void and its reducer type must
+match the space's ``reduction_t``.
```
Yurlungur:
```suggestion
An overload instead takes a caller-constructed reducer instance *last*, matching ``Kokkos::parallel_reduce(policy, functor, reducer)``. This returns void and its reducer type must match the space's ``reduction_t``. This overload is necessary for, e.g., reducing into a
``View``, ``ScatterView``, or device memory.
```

doc/sphinx/src/loop_abstraction.rst
```diff
@@ -179,6 +179,64 @@ scratch for the Kokkos backend for other paths), but the user-facing indexing is
...
+
+Because the reducer type lives on the index space, ``inner_reduce`` reuses its join op
+without the caller restating it, and a single ``outer_reduce`` region may contain
+several ``inner_reduce`` calls (interleaved with plain ``inner`` calls that only fill
+scratch) that all join into one accumulator. There is one reducer per region. The
+``inner_reduce`` body takes the usual index form plus a trailing reduction-value
+reference.
```
Yurlungur: As with standard ``Kokkos`` reduction operations, because the reducer type lives...

doc/sphinx/src/loop_abstraction.rst
```diff
@@ -179,6 +179,64 @@ scratch for the Kokkos backend for other paths), but the user-facing indexing is
...
+
+Two rules keep reductions off ghost cells:
+
+- **No reductions over halo ranges.** ``inner_reduce`` ``static_assert``\ s that the
+  range's halo is ``none_t``. Extend a range only for producer (scratch) ``inner``
+  loops and reduce over the base, halo-free range.
```
Yurlungur, line 228: This seems a little ambiguous to me. Can I have a "producer" inner loop inside an outer reduce, so long as the actual accumulation is only done in the inner reduce?

doc/sphinx/src/loop_abstraction.rst
```diff
@@ -179,6 +179,64 @@ scratch for the Kokkos backend for other paths), but the user-facing indexing is
...
+- **The** ``memory`` **inner tag degenerates to** ``logical_flat`` **-- but only for**
+  ``inner_reduce``. Under a reduction the ``memory`` tag iterates logical cells rather
+  than a contiguous memory span, so no swept ghost cell is folded in (the body still
+  gets a memory-relative flat index, so call sites are unchanged). This is scoped
+  strictly to ``inner_reduce``: a plain ``inner()`` inside an ``outer_reduce`` region
+  behaves exactly as under ``outer()`` and does **not** degenerate -- with the
+  ``memory`` tag it still sweeps whole contiguous spans, ghosts included. A
+  ``memory``-tag producer feeding an ``inner_reduce`` consumer is therefore fine; just
+  don't assume the producer stayed inside the logical set.
```
Yurlungur, line 237: ...maybe this answers my question

src/loop_abstraction/pack_view.hpp
Yurlungur, line 1: @lroberts36 can you check if the changes in jmm/partest are consistent with this? And if not can we unify those? We probably want to include them.

Yurlungur (approved):
Surprisingly simple to extend.

doc/sphinx/src/loop_abstraction.rst
lroberts36: Standard Kokkos reductions can only be on an inner loop right?

src/loop_abstraction/pack_view.hpp
lroberts36: I think these changes are orthogonal to what you and Josh were doing there. The changes here are cosmetic to reduce the amount of dependency on `IndexSpaceType` (passing `loop_tag` and `inner_tag` directly and using a pointer to the memory indexer directly rather than through the index space). None of the logic about building the views was changed.

doc/sphinx/src/loop_abstraction.rst
Yurlungur: Maybe I misunderstood what this paragraph was trying to say. I thought it meant we didn't need to call `Reducer.join` anywhere, which is also true of calling Kokkos.

lroberts36: A Kokkos reduction over both levels of a hierarchical loop requires
```c++
using team_policy = Kokkos::TeamPolicy<>;
using member_type = team_policy::member_type;

int sum = 0;
Kokkos::Sum<int> reducer(sum);

Kokkos::parallel_reduce(
    "hierarchical_reduce",
    team_policy(N, Kokkos::AUTO()),
    KOKKOS_LAMBDA(const member_type &team_member, int &lsum) {
      int team_sum = 0;

      Kokkos::parallel_reduce(
          Kokkos::TeamThreadRange(team_member, team_size),
          [&](const int k, int &inner_sum) {
            const int value = /* contribution from k */;
            inner_sum += value;
          },
          team_sum);

      Kokkos::single(
          Kokkos::PerTeam(team_member),
          [&]() {
            // For Kokkos::Sum<int>, equivalent to lsum += team_sum;
            reducer.join(lsum, team_sum);
          });
    },
    reducer);
```

A possible update to clarify this:
```suggestion
Because the reducer type lives on the index space, ``inner_reduce`` uses the 
reducer's merge operation to combine each inner team reduction into the enclosing
reduction without the caller restating it, and a single ``outer_reduce`` region may
contain several ``inner_reduce`` calls (interleaved with plain ``inner`` calls that
only fill scratch) that all join into one accumulator. There is one reducer per region.
The ``inner_reduce`` body takes the usual index form plus a trailing reduction-value
reference.
```

Yurlungur: Ah I see. Yes I think the clarification works.

lroberts36:
@par-hermes format

Yurlungur:
@lroberts36 is this ready to go?

lroberts36:
@yurlungur: Yeah, I think it is just waiting on the CI to pass to automerge.

PR #1434, opened 2026-08-12
Some XDMF fixes

Yurlungur:
## PR Summary

Recently @jdolence noticed that including nodal fields in *any* HDF5 output broke visit for *all* fields. The reason for this was what must have been a long persistent bug, where the x1,x2,x3 values were reset per block for the variables, and this polluted the geometry logic.

There's also been a persistent issue/complaint that nodal fields (even when displayed) displayed weirdly, often on the back side of a 3D image. This was because xdmf distinguishes between 2D and 3D meshes, while parthenon does not. To resolve this, I add logic to switch between 2DRectMesh and 3DRectMesh the same way we do for custom user coordinates. This resolves the issue so that now Visit and Paraview correctly plot nodal and cell fields in 2D and 3D.

To test all of this, I add a nodal field to the output of the fine advection example, which I fill by averaging in FillDerived. I also add some Python based xml parsing to the restart_fine regression test to ensure this continues to work (as long as we continue to care about xdmf at least... It's days may be numbered.)

For fun, I fill that field with @lroberts36 's new loop abstraction. So there's now a functioning, live example of how to use it.

## PR Checklist

- [x] Code passes cpplint
- [x] Code is formatted
- [x] Docs build

### DIFF
--- a/example/fine_advection/advection_package.cpp
+++ b/example/fine_advection/advection_package.cpp
@@ -253,31 +253,20 @@ TaskStatus FillDerived(MeshData<Real> *md) {
     la::outer(
         idx_space, KOKKOS_LAMBDA(const IST::idx_range_t &idx_range, int b) {
           const int nvar = pack.GetSize(b, Conserved::phi());
-          const Real weight = (ndim > 2) ? 0.125 : (ndim > 1) ? 0.25 : 0.5;
           for (int n = 0; n < nvar; ++n) {
             auto pvn = la::make_sparse_pack_view(idx_range, pack, n);
 
             la::inner(idx_range, [&](const auto kji) {
-              pvn(Nodal::phi(), kji) =
-                  pvn(Conserved::phi(), kji) + pvn(Conserved::phi(), kji - di);
+              pvn(Nodal::phi(), kji) = 0.125 * (
+                     pvn(Conserved::phi(), kji) 
+                  + pvn(Conserved::phi(), kji - di)
+                  + pvn(Conserved::phi(), kji - dj)
+                  + pvn(Conserved::phi(), kji - dk)
+                  + pvn(Conserved::phi(), kji - di - dj)
+                  + pvn(Conserved::phi(), kji - dj - dk)
+                  + pvn(Conserved::phi(), kji - dk - di)
+                  + pvn(Conserved::phi(), kji - di - dj - dk));
             });
-            idx_range.TeamBarrier();
-            if (ndim > 1) {
-              la::inner(idx_range, [&](const auto kji) {
-                pvn(Nodal::phi(), kji) +=
-                    (pvn(Conserved::phi(), kji) + pvn(Conserved::phi(), kji - dj));
-              });
-              idx_range.TeamBarrier();
-            }
-            if (ndim > 2) {
-              la::inner(idx_range, [&](const auto kji) {
-                pvn(Nodal::phi(), kji) +=
-                    (pvn(Conserved::phi(), kji) + pvn(Conserved::phi(), kji - dk));
-              });
-              idx_range.TeamBarrier();
-            }
-            la::inner(idx_range,
-                      [&](const auto kji) { pvn(Nodal::phi(), kji) *= weight; });
           }
         });
   }
--- a/example/fine_advection/advection_package.cpp
+++ b/example/fine_advection/advection_package.cpp
@@ -257,14 +257,18 @@ TaskStatus FillDerived(MeshData<Real> *md) {
             auto pvn = la::make_sparse_pack_view(idx_range, pack, n);
 
             la::inner(idx_range, [&](const auto kji) {
+              /* clang-format off */
               pvn(Nodal::phi(), kji) =
                   0.125 *
-                  (pvn(Conserved::phi(), kji) + pvn(Conserved::phi(), kji - di) +
-                   pvn(Conserved::phi(), kji - dj) + pvn(Conserved::phi(), kji - dk) +
+                  (pvn(Conserved::phi(), kji) +
+                   pvn(Conserved::phi(), kji - di) +
+                   pvn(Conserved::phi(), kji - dj) +
+                   pvn(Conserved::phi(), kji - dk) +
                    pvn(Conserved::phi(), kji - di - dj) +
                    pvn(Conserved::phi(), kji - dj - dk) +
                    pvn(Conserved::phi(), kji - dk - di) +
                    pvn(Conserved::phi(), kji - di - dj - dk));
+              /* clang-format on */
             });
           }
         });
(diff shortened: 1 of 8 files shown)


example/fine_advection/advection_package.cpp
```diff
@@ -222,6 +228,60 @@ TaskStatus FillDerived(MeshData<Real> *md) {
...
+              la::inner(idx_range, [&](const auto kji) {
+                pvn(Nodal::phi(), kji) +=
+                    (pvn(Conserved::phi(), kji) + pvn(Conserved::phi(), kji - dk));
+              });
+              idx_range.TeamBarrier();
+            }
+            la::inner(idx_range,
+                      [&](const auto kji) { pvn(Nodal::phi(), kji) *= weight; });
+          }
+        });
```
lroberts36, line 276: I think to average from the cell center to the node you want:
```suggestion
    la::outer(
        idx_space, KOKKOS_LAMBDA(const IST::idx_range_t &idx_range, int b) {
          const int nvar = pack.GetSize(b, Conserved::phi());
          for (int n = 0; n < nvar; ++n) {
            auto pvn = la::make_sparse_pack_view(idx_range, pack, n);

            la::inner(idx_range, [&](const auto kji) {
              pvn(Nodal::phi(), kji) = 0.125 * (
                     pvn(Conserved::phi(), kji) 
                  + pvn(Conserved::phi(), kji - di)
                  + pvn(Conserved::phi(), kji - dj)
                  + pvn(Conserved::phi(), kji - dk)
                  + pvn(Conserved::phi(), kji - di - dj)
                  + pvn(Conserved::phi(), kji - dj - dk)
                  + pvn(Conserved::phi(), kji - dk - di)
                  + pvn(Conserved::phi(), kji - di - dj - dk));
            });
          }
        });
```

src/outputs/parthenon_xdmf.cpp
```diff
@@ -191,11 +205,9 @@ void genXDMF(std::string hdfFile, Mesh *pm, SimTime *tm, IndexDomain domain, int
         }
         ndim = vinfo.FillShape<hsize_t>(domain, &(dims[1])) + 1;
         const int num_components = vinfo.num_components;
-        nx3 = dims[ndim - 3];
-        nx2 = dims[ndim - 2];
-        nx1 = dims[ndim - 1];
         writeXdmfSlabVariableRef(xdmf, vinfo.label, vinfo.component_labels, hdfFile, ib,
-                                 num_components, ndim, dims, nx3, nx2, nx1, output_coords,
+                                 num_components, ndim, dims, dims[ndim - 3],
+                                 dims[ndim - 2], dims[ndim - 1], output_coords || is_2d,
```
lroberts36, line 212: This looks a little suspect to me (e.g. `ndim - 3` seems like it is should give -1 for 2D), but I don't see how `ndim` is being set.

example/fine_advection/advection_package.cpp
Yurlungur: oh yes, you're right. But I need guards for 1D/2D yeah?

lroberts36: You don't need guards (or at least shouldn't). The `IndexSpaceType::GetDelta` is dimension aware, so if you ask for say `dk` in a 2D problem, you get the result `kji - dk == kji`.

src/outputs/parthenon_xdmf.cpp
Yurlungur: Nope this is correct.

`dims` is set here and it's the HDF5 dimension array so it's shaped backwards from the rest of parthenon convention. You have to subtract off from ndim.
https://github.com/parthenon-hpc-lab/parthenon/blob/develop/src/outputs/output_utils.hpp#L92

And ndim is always at least 3:
https://github.com/parthenon-hpc-lab/parthenon/blob/develop/src/outputs/output_utils.cpp#L119

example/fine_advection/advection_package.cpp
Yurlungur: Ah gotcha. That's much simpler.

Yurlungur:
@par-hermes format

src/outputs/parthenon_xdmf.cpp
lroberts36: ok, cool, I think I read it as `ndim` == number of dimensions in the problem.

lroberts36 (approved):

src/outputs/parthenon_xdmf.cpp
```diff
@@ -154,7 +166,7 @@ void genXDMF(std::string hdfFile, Mesh *pm, SimTime *tm, IndexDomain domain, int
       xdmf << StringPrintf("      <Topology TopologyType=\"%s\" Dimensions=\"%s\"/>\n",
                            mesh_type.c_str(), dimstring.c_str());
       xdmf << StringPrintf("      <Geometry GeometryType=\"%s\">\n",
-                           output_coords ? "X_Y_Z" : "VXVYVZ");
+                           output_coords ? "X_Y_Z" : (is_2d ? "VXVY" : "VXVYVZ"));
```
acreyes, line 171:
```suggestion
                           output_coords ? "X_Y_Z" : (is_2d ? "VXVY" : (is_1d ? "VX" : "VXVYVZ")));
```
I don't know if this is a thing

src/outputs/parthenon_xdmf.cpp
```diff
@@ -175,7 +187,9 @@ void genXDMF(std::string hdfFile, Mesh *pm, SimTime *tm, IndexDomain domain, int
       } else {
         BlockCoordRegularRef(xdmf, pm->nbtotal, ib, nx1, hdfFile, "x");
         BlockCoordRegularRef(xdmf, pm->nbtotal, ib, nx2, hdfFile, "y");
```
acreyes, line 191:
```suggestion
        if (!is_1d) BlockCoordRegularRef(xdmf, pm->nbtotal, ib, nx2, hdfFile, "y");
```
I'm guessing this would be necessary with the above `"VX"` suggestion

acreyes (approved):
LGTM! Does the change to "VXVY" geometry in 2D mean that visit will interpret the outputs as 2D data now?

Yurlungur:
> Does the change to "VXVY" geometry in 2D mean that visit will interpret the outputs as 2D data now?

Yep that's right!

src/outputs/parthenon_xdmf.cpp
Yurlungur: Unfortunately it's not, which is why I didn't include it. XDMF seems to assume 2D+.

Yurlungur: See here: https://www.xdmf.org/index.php/XDMF_Model_and_Format

src/outputs/parthenon_xdmf.cpp
Yurlungur: Yeah. Unfortunately VX doesn't exist... I will add a clarifying comment.

